PR #2455, opened 2020-10-27
Add non-numeric pressure check to fix box/relax

stanmoore1:
**Summary**

Add non-numeric pressure check to `fix box/relax` so that the minimizer doesn't hang when the pressure is NaN, etc. This check matches that already in `fix_nh.cpp`, see https://github.com/lammps/lammps/blob/master/src/fix_nh.cpp#L1061.

**Related Issue(s)**

None

**Author(s)**

Stan Moore (SNL), reported by Calvin Anderson (Miami University)

**Licensing**

**Backward Compatibility**

Yes

### DIFF
--- a/src/fix_box_relax.cpp
+++ b/src/fix_box_relax.cpp
@@ -735,12 +735,18 @@ void FixBoxRelax::couple()
     p_current[2] = tensor[2];
   }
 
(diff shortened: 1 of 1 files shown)


athomps (approved):
Looks good.

sjplimp (approved):

PR #2462, opened 2020-10-29
Preserve molecule templates when using "replicate"

akohlmey:
**Summary**

The replicate command works by creating a new instance of an Atom class and then populates it with correspondingly replicated copies of atoms and associated data from the original class and finally swaps the two and deletes the old one. Since the list molecule templates is stored as a member of the Atom class, it needs to be transferred in the process. Otherwise those "molecules" are 'lost' and the command will thus not work when using atom style "template".

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

Previously molecule templates would have to recreated by re-issuing molecule commands after using "replicate". This is no longer necessary and re-issuing molecule commands for the same molecules will raise an error.

**Implementation Notes**

Since the list of molecules is independent of the number of atoms, the list of molecules is simply transferred.

**Post Submission Checklist**


rbberger:
@akohlmey the testcases are getting unreadable. Please add some comments or use helper functions to document these endless lists of assertions please. Otherwise you'll be the only person able to read and maintain them.

akohlmey:
> @akohlmey the testcases are getting unreadable. Please add some comments or use helper functions to document these endless lists of assertions. Otherwise you'll be the only person able to read and maintain them.

This is not as easy to do as you may think. The principle is quite simple: there are a whole lot of settings and arrays associated with an atom style and this first checks that everything is reset properly to zero/non-zero for a given style and then it creates/reads/deletes some atoms and verifies the associated data is set/unset as needed. To keep the lists from growing even longer, only selected properties are checked on subsequent tests. Since this is all depended on what data is associated with an atom style, I have yet to find a way to make this more readable. The best way to make sense of these tests is to look at them while also looking at the Atom and AtomVec classes. That is what I do.

rbberger:
@akohlmey cleaned this up a bit and found an issue in the avec template class. Have another idea of getting rid of the assert/expects for the vectors, but that is for another day.

akohlmey:
@rbberger @sjplimp can we merge this now? I would like to start working on adding support for "molfrac" to "create_atoms" so that it can handle templates with multiple molecules. but before starting with that I need to have the components in those two PRs merged so I won't have to do work twice and won't have to resolve merge conflicts. thanks.

sjplimp (approved):

PR #2464, opened 2020-11-02
bug fix in pair_mesont_tpm.cpp

iafoss:
**Summary**

Fix bug with memory allocation.

### DIFF
--- a/src/USER-MESONT/pair_mesont_tpm.cpp
+++ b/src/USER-MESONT/pair_mesont_tpm.cpp
@@ -9,7 +9,7 @@
    the GNU General Public License.
 
    See the README file in the top-level LAMMPS directory.
-
+ 2
    Contributing author: Maxim Shugaev (UVA), mvs9t@virginia.edu
 ------------------------------------------------------------------------- */
 
@@ -29,6 +29,7 @@
 #include <cstring>
 #include <vector>
 #include <cmath>
+#include <array>
 
 #include <fstream>
 #include <sstream>
@@ -36,28 +37,17 @@
 
 using namespace LAMMPS_NS;
 
-//since LAMMPS is compiled with C++ 2003, define a substitution for std::array
-template<typename T, int N>
-class array2003{
-public:
-  T& operator[] (int idx) { return data[idx];};
-  const T& operator[] (int idx) const{ return data[idx];};
-private:
-  T data[N];
-};
-
-
 class MESONTList {
 public:
   MESONTList(const Atom* atom, const NeighList* nblist);
   ~MESONTList() {};
   //list of segments
-  const std::vector<array2003<int,2> >& get_segments() const;
+  const std::vector<std::array<int,2>>& get_segments() const;
   //list of triplets
-  const std::vector<array2003<int,3> >& get_triplets() const;
+  const std::vector<std::array<int,3>>& get_triplets() const;
   //list of neighbor chains [start,end] for segments
   //(use idx() to get real indexes)
-  const std::vector<std::vector<array2003<int,2> > >& get_nbs() const;
+  const std::vector<std::vector<std::array<int,2>>>& get_nbs() const;
   //convert idx from sorted representation to real idx
   int get_idx(int idx) const;
   //return list of indexes for conversion from sorted representation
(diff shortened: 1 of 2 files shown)


src/USER-MESONT/pair_mesont_tpm.cpp
```diff
@@ -510,6 +510,12 @@ void PairMESONTTPM::compute(int eflag, int vflag){
 
   // set per atom values and accumulators
   // reallocate per-atom arrays if necessary
+  if (eatom_s == nullptr)
```
akohlmey, line 513: This is likely the wrong fix. It should not be needed, if `maxeatom` is properly initialized.

iafoss: @akohlmey , Could you please tell me where the initial allocation of 'eatom' array is made? I will make sure that the variables added in the mesont package: 'eatom_s', 'eatom_b', 'eatom_t' are treated in the same way, because it seems that there is an issue with the initial allocation.

akohlmey: eatom is the wrong item to compare to.

the situation is quite clear: `eatom_s`, `eatom_b`, and `eatom_t` are initialized to `nullptr` in the pair style constructor, so the only case where they are not increased by this code segment is if `maxeatom` is larger or equal than `atom->nmax`. but that should not happen unless `maxeatom` is not initialized to 0 in the pair style constructor.

If you need some code to compare to, you could look at the "rho" array in  pair style eam.

BTW: I notice that your code deallocates and allocates `eatom`, that should *not* be done. This array is managed by the base class.

akohlmey: as background information: the allocation of eatom is triggered only if there is a "consumer" for it, which results in the eflag_atom (and eflag_either) variable become "1" instead of "0". So if you have problems accessing eatom, you have to prefix that with corresponding tests, so that it is not set when it is not requested and thus not allocated.

iafoss: Thank you so much for your clarification, I'll look more into it.

src/USER-MESONT/pair_mesont_tpm.cpp
```diff
@@ -508,58 +520,58 @@ void PairMESONTTPM::compute(int eflag, int vflag){
...
+      int idx = ntlist.get_idx(i);
+      for (int j = 0; j < 3; j++) f[idx][j] += f_sort[3*i+j];
+      buckling[idx] = b_sort[i];
   }
-
-  if (atom->nmax > maxvatom) {
-    maxvatom = atom->nmax;
-    memory->destroy(vatom);
-    memory->create(vatom,comm->nthreads*maxvatom,6,"pair:vatom");
+  if(eflag){
```
akohlmey, line 529: this is still not good enough. eflag is nonzero if either eflag_atom or eflag_global is set, but you can use eatom *only* if eflag_atom is set (and I would assume the same is true for the other eatom_* variants, otherwise only global accumulators are available.

src/USER-MESONT/pair_mesont_tpm.cpp
```diff
@@ -508,58 +520,58 @@ void PairMESONTTPM::compute(int eflag, int vflag){
...
-    energy_t += u_tt_sort[i];
-    vatom[idx][0] = s_sort[9*i+0]; //xx
-    vatom[idx][1] = s_sort[9*i+4]; //yy
-    vatom[idx][2] = s_sort[9*i+8]; //zz
-    vatom[idx][3] = s_sort[9*i+1]; //xy
-    vatom[idx][4] = s_sort[9*i+2]; //xz
-    vatom[idx][5] = s_sort[9*i+5]; //yz
-    for (int j = 0; j < 6; j++) virial[j] += vatom[idx][j];
-    buckling[idx] = b_sort[i];
+  int vflag_atom = vflag & 4;
```
akohlmey, line 560: this should not be needed and is shadowing an existing variable. vflag_atom (and vflag_global) are set by calling ev_init().

src/USER-MESONT/pair_mesont_tpm.cpp
iafoss: Thank you for pointing it out, I already realized it and wanted to separate the evaluation of per atom and global energies.

src/USER-MESONT/pair_mesont_tpm.cpp
iafoss: Thank you, I'll fix it.

iafoss:
@akohlmey , Thank you for your suggestions. I hope the latest version addresses the issues you mentioned.

akohlmey:
> @akohlmey , Thank you for your suggestions. I hope the latest version addresses the issues you mentioned.

yes. at least from a superficial glance. I currently don't have the time to check out and carefully test everything with valgrind.
one suggestion for a cosmetic change. the code would be more consistent with the usual formatting in LAMMPS, if there were more spaces in if and for statements, i.e. not `if(condition){` but `if (condition) {` and similar for for loops.

iafoss:
Done. I have one more question: should I keep backward compatibility with C++ 2003, or I can assume that the next versions of LAMMPS will be compiled with C++ 11? I just think about getting rid of "array2003" class that I introduced when converted my code to C++ 2003.

akohlmey:
> Done. I have one more question: should I keep backward compatibility with C++ 2003, or I can assume that the next versions of LAMMPS will be compiled with C++ 11? I just think about getting rid of "array2003" class that I introduced when converted my code to C++ 2003.

The 3 March 2020 version was the last version where the LAMMPS core code was requiring C++98. All releases since then require C++11. The requirements for packages could already be higher (KOKKOS has been requiring C++11 for a long time and some other packages, too). With the changes to the core LAMMPS code over the last 8 months, it would be impossible to compile LAMMPS with anything but C++11 or later.

akohlmey (approved):

PR #2467, opened 2020-11-05
Bond/react: molecule keyword + bugfixes

jrgissing:
**Summary**

add molecule keyword to optionally force reaction to be inter- or intramolecular

also update terminology from 'bonding atoms' to 'initiator atoms' for the atom pair that initiates a reaction, as bonds do not need to be formed or broken between 'initiator atoms'

bugfix for issue that can appear when using the delete atoms feature (previously, could hang when deleted atom was last on processor list)

bugfix for correctly discarding reaction sites for certain ringed structures (previously, an atom at the 'end' of a symmetric ring could be treated like an edge atom, in some cases)

bugfix when error-checking for reaction distance cutoff vs. pairwise cutoff (now doesn't check when bonding atoms are directly bonded, and also fixes false negatives when using pair_style hybrid)

**Related Issue(s)**

none

**Author(s)**

JG

**Licensing**

**Backward Compatibility**

updates map file section title from 'BondingIDs' to 'InitiatorIDs'. Old section title still works for backward compatibility, with a printed warning that it has been deprecated

**Implementation Notes**

**Post Submission Checklist**


rbberger (approved):

doc/src/fix_bond_react.rst
```diff
@@ -581,10 +593,14 @@ Default
...
-reset_mol_ids = yes, custom_charges = no
+reset_mol_ids = yes, custom_charges = no, molecule = off
 
 ----------
 
 .. _Gissinger:
 
 **(Gissinger)** Gissinger, Jensen and Wise, Polymer, 128, 211 (2017).
+
+.. _Gissinger:
```
akohlmey, line 604: there must not be the same anchor target twice. "make html" should print a warning about this.

akohlmey (changes requested):

doc/src/fix_bond_react.rst
rbberger: @akohlmey on it

akohlmey (approved):

PR #2468, opened 2020-11-05
Added electronic stopping as a friction term with a velocity cutoff w…

ManojWarrier:
**Summary**

**Related Issue(s)**

**Author(s)**

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/src/USER-ESFRICTION/fix_esfriction.cpp
+++ b/src/USER-ESFRICTION/fix_esfriction.cpp
@@ -1,0 +1,203 @@
+/* ----------------------------------------------------------------------
+   LAMMPS - Large-scale Atomic/Molecular Massively Parallel Simulator
+   http://lammps.sandia.gov, Sandia National Laboratories
+   Steve Plimpton, sjplimp@sandia.gov
+
+   Copyright (2003) Sandia Corporation.  Under the terms of Contract
+   DE-AC04-94AL85000 with Sandia Corporation, the U.S. Government retains
+   certain rights in this software.  This software is distributed under
+   the GNU General Public License.
+
+   See the README file in the top-level LAMMPS directory.
+------------------------------------------------------------------------- */
+
+/* ----------------------------------------------------------------------
+       Contributing Author: Harsh Hemani and Manoj Warrier
+       (Bhabha Atomc Research Center, Computational Analysis Division,
+        Visakhapatnam, India)
+       harshscience777@gmail.com, Manoj.Warrier@gmail.com
+---------------------------------------------------------------------- */
+
+#include <mpi.h>
+#include <cmath>
+#include <cstring>
+#include "fix_esfriction.h"
+#include "atom.h"
+#include "domain.h"
+#include "update.h"
+#include "error.h"
+#include "force.h"
+#include "respa.h"
+
+using namespace LAMMPS_NS;
+using namespace FixConst;
+
+/* ---------------------------------------------------------------------- */
+
+FixESFriction::FixESFriction(LAMMPS *lmp, int narg, char **arg) :
+  Fix(lmp, narg, arg)
+{
+  if (narg < 4) error->all(FLERR,"Illegal fix es-friction command");
+  // args -- 0: fix, 1: fix-id, 2: group-id
+  if (strcmp(update->unit_style,"metal") != 0) {
+    error->all(FLERR, "fix es-friction supports only metal units");
+  }
+  Z1 = utils::numeric(FLERR, arg[3], false, lmp); // atomic number of PKA
+  Z2 = utils::numeric(FLERR, arg[4], false, lmp); // atomic number of target
+  A1 = utils::numeric(FLERR, arg[5], false, lmp); // mass number of PKA
+  A2 = utils::numeric(FLERR, arg[6], false, lmp); // mass number of target
+  Ne_val = utils::numeric(FLERR, arg[7], false, lmp);  // no. of valance shell electrons
+  E_pka = utils::numeric(FLERR, arg[8], false, lmp);   // PKA energy
+  /* ES_cutoff is the energy cutoff for applying ES.
+     That is, atoms with energy below this will not experience ES */
+  ES_cutoff = utils::numeric(FLERR, arg[9], false, lmp);
+  // Compute total energy lost due to ES as per NRT model
+  double E_hat = get_ehat();
+  ES_loss_NRT = E_pka - E_hat;
+  // following computes coefficient of friction (beta) using LS model
+  double number_density = atom->natoms / 
+    (domain->xprd * domain->yprd * domain->zprd) * 1e30; // number-per-m3
+  double E_0 = h_cross*h_cross / (2*m_e) * 
+    pow( 3*pi*pi*Ne_val*(number_density), 2.0/3.0) * joules_to_ev; // eV
+  double v_0 = 5.931e5 * sqrt(E_0) * 1e-2; // ang/ps
+  // lambda is in eV-Angstrom-ps units
+  double lambda = 8.0 * pi * (e_charge * e_charge * 10) * bohr_radius * 
+    pow(Z1, 1.0/6.0) * Z1*Z2 / ( v_0 * pow( pow(Z1, 2.0/3.0) + 
+    pow(Z2, 2.0/3.0) , 3.0/2.0) ); 
+  beta = number_density * lambda / 1.0e30; //eV-ps-angstrom^-2 units
+  // Init energy lost due to ES as per LS model (calculated numerically)
+  ES_loss_LS = 0.0;
+  MPI_Comm_rank(world, &my_rank);
+  if (my_rank==0){
+    if(screen){
+      fprintf(screen, "E_pka = %g eV\n", E_pka);
+      fprintf(screen, "Energy loss due to ES (from NRT model) = %g\n", 
+        ES_loss_NRT);
+    }
+    if (logfile){
+      fprintf(logfile, "E_pka = %g eV\n", E_pka);
+      fprintf(logfile, "Energy loss due to ES (from NRT model) = %g\n", 
+        ES_loss_NRT);
+    }
+  }
+}
+
+/* ---------------------------------------------------------------------- */
+
+double FixESFriction::get_kenergy(double A, double vx, double vy, double vz)
+{
+  double gms_to_kg = 1e-3;  // grams to kg
+  double apps_to_mps = 100; // amstrong per picoseconds -to- m/s
+  return (0.5 * A * ( vx*vx + vy*vy + vz*vz ) * gms_to_kg * apps_to_mps * 
+    apps_to_mps * joules_to_ev / Nav); // in eV
+}
+
+/* ----------------------------------------------------------------------
+   Ehat is total energy lost due to ES based on NRT model 
+----------------------------------------------------------------------- */
+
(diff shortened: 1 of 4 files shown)


ManojWarrier:
The title comment should have been, "Added electronic stopping as a friction term with a energy cutoff ..." and not "... velocity cutoff ..."

akohlmey:
> The title comment should have been, "Added electronic stopping as a friction term with a energy cutoff ..." and not "... velocity cutoff ..."

There is an "Edit" button next to the title, where you can change the title of the pull request.

akohlmey:
Before getting into details, please see my [comment on PR 2391](https://github.com/lammps/lammps/pull/2391#issuecomment-698007379) which is also submitting a fix to model electron stopping effects in addition to a fix that is already included.
Also, we do not create a new USER package for a single command contribution unless that is an interface to a library (bundled or not bundled) or consists of a collection of related contributions or is a single complex multi-file contribution. Thus this fix should be added to the USER-MISC package unless there is an agreement to have (and maintain!!) a common package for **all** electron stopping models.
Including a new package requires multiple changes to the manual and the build system to work properly and people owning up to be responsible for the package. For inclusion into USER-MISC only the README in the USER-MISC folder needs to be updated and the added fix needs to be integrated into the documentation (when doing "make html" scripts will be run to tell you about it and where).

I will add some comments about the code separately, but looking from the user perspective we need that you coordinate with  @jstewa-snl and @rtoijala to have a consistent naming scheme for all those electron stopping fixes and either converge on a single documentation page for all three fixes or have the other two fixes referenced from the three individual pages and include a paragraph discussing (in a neutral way) the difference between the different methodology. The conclusion which is the superior method can be left to the user. If all include references to associated publications, there will be plenty of suitable information.

akohlmey:
Please note that no pull request can be merged with failing tests. In this case it looks like that "make spelling" in the "docs" folder is failing, most likely due to false positives from keywords or names. Those can be added to `doc/utils/sphinx-config/false_positives.txt`

src/USER-ESFRICTION/fix_esfriction.cpp
```diff
@@ -0,0 +1,203 @@
...
+     That is, atoms with energy below this will not experience ES */
+  ES_cutoff = utils::numeric(FLERR, arg[9], false, lmp);
+  // Compute total energy lost due to ES as per NRT model
+  double E_hat = get_ehat();
+  ES_loss_NRT = E_pka - E_hat;
+  // following computes coefficient of friction (beta) using LS model
+  double number_density = atom->natoms / 
+    (domain->xprd * domain->yprd * domain->zprd) * 1e30; // number-per-m3
+  double E_0 = h_cross*h_cross / (2*m_e) * 
+    pow( 3*pi*pi*Ne_val*(number_density), 2.0/3.0) * joules_to_ev; // eV
```
akohlmey, line 61: Please use the `MY_PI` constant from the `MathConst` namespace to have a consistent use and precision of the PI constant and not your own definition.

src/USER-ESFRICTION/fix_esfriction.cpp
```diff
@@ -0,0 +1,203 @@
...
+  double v_0 = 5.931e5 * sqrt(E_0) * 1e-2; // ang/ps
```
akohlmey, line 62: What does `5.931e5` stand for?

src/USER-ESFRICTION/fix_esfriction.cpp
```diff
@@ -0,0 +1,203 @@
...
+  // lambda is in eV-Angstrom-ps units
+  double lambda = 8.0 * pi * (e_charge * e_charge * 10) * bohr_radius * 
+    pow(Z1, 1.0/6.0) * Z1*Z2 / ( v_0 * pow( pow(Z1, 2.0/3.0) + 
+    pow(Z2, 2.0/3.0) , 3.0/2.0) ); 
+  beta = number_density * lambda / 1.0e30; //eV-ps-angstrom^-2 units
+  // Init energy lost due to ES as per LS model (calculated numerically)
+  ES_loss_LS = 0.0;
+  MPI_Comm_rank(world, &my_rank);
```
akohlmey, line 70: Please use `comm->me` instead of explicitly setting `my_rank`.

src/USER-ESFRICTION/fix_esfriction.cpp
```diff
@@ -0,0 +1,203 @@
...
+  if (my_rank==0){
+    if(screen){
```
akohlmey, line 72: Please use `utils::logmesg` and `fmt::format` to make this a single invocation.

src/USER-ESFRICTION/fix_esfriction.cpp
```diff
@@ -0,0 +1,203 @@
...
+}
+
+/* ----------------------------------------------------------------------
+   Ehat is total energy lost due to ES based on NRT model 
+----------------------------------------------------------------------- */
+
+double FixESFriction::get_ehat()
+{
+  double Emax = 25.0 * pow(Z1, 4.0/3.0) * A1;
+  if (E_pka >= Emax){
```
akohlmey, line 102: Please use `utils::logmesg` and `fmt::format` to simplify the code and protect this output with an if for `comm->me == 0` so that the warning is only printed once when running in parallel. Keep in mind that LAMMPS can be (and is) run across 1000s of MPI ranks.

akohlmey: You may also consider using `Error::warning()`

src/USER-ESFRICTION/fix_esfriction.cpp
```diff
@@ -0,0 +1,203 @@
...
+      fprintf(logfile, "Hence E_hat calculation (NRT) might be invalid.\n");
+    }
+  }
+  double eps, gofeps, kay;
+  double const1 = pow( (9.0 * pi*pi / 128.0), (1.0/3.0) );
+  double a = ( const1 * (bohr_radius /  10.0) ) / sqrt( pow(Z1, 2/3.0) + 
+    pow(Z2, 2/3.0) ) ;
+  eps = ( A2 / (A1+A2) ) * ( (a / ( Z1 * Z2 * e_charge * e_charge ) ) ) * 
+    E_pka;
+  gofeps = (3.4008 * pow(eps, 1/6.0)) + (0.40244 * pow(eps, 3.0/4.0)) + eps;
```
akohlmey, line 120: What do 3.4008, 0.40244, and 0.1337 represent?

src/USER-ESFRICTION/fix_esfriction.cpp
```diff
@@ -0,0 +1,203 @@
...
+      fprintf(screen, "Hence E_hat calculation (NRT) might be invalid.\n");
+    }
+    if (logfile){
+      fprintf(logfile, "E_pka >= Emax (%g > %g) in esfriction fix.\n", E_pka, 
+        Emax);
+      fprintf(logfile, "Hence E_hat calculation (NRT) might be invalid.\n");
+    }
+  }
+  double eps, gofeps, kay;
+  double const1 = pow( (9.0 * pi*pi / 128.0), (1.0/3.0) );
```
akohlmey, line 115: Please use `MathConst::MY_PI`

src/USER-ESFRICTION/fix_esfriction.cpp
```diff
@@ -0,0 +1,203 @@
...
+  eps = ( A2 / (A1+A2) ) * ( (a / ( Z1 * Z2 * e_charge * e_charge ) ) ) * 
+    E_pka;
+  gofeps = (3.4008 * pow(eps, 1/6.0)) + (0.40244 * pow(eps, 3.0/4.0)) + eps;
+  kay = 0.1337 * pow(Z1, 1/6.0) * sqrt(Z1/A1);
+  double Ehat = E_pka / (1.0 + kay * gofeps);
+  return Ehat;
+}
+
+/* ---------------------------------------------------------------------- */
+FixESFriction::~FixESFriction()
```
akohlmey, line 127: An empty destructor can be defined in the class definition directly. No need to have it here.

src/USER-ESFRICTION/fix_esfriction.cpp
```diff
@@ -0,0 +1,203 @@
...
+int FixESFriction::setmask()
+{
+  int mask = 0;
+  mask |= POST_FORCE;
+  return mask;
+}
+
+/* ---------------------------------------------------------------------- */
+
+void FixESFriction::init()
```
akohlmey, line 141: An empty init() method is already defined in fix.h. No need to override it with an empty method here. A re-implementation of methods is only required for methods that are pure (i.e. defined to be NULL) in the base class.

src/USER-ESFRICTION/fix_esfriction.cpp
```diff
@@ -0,0 +1,203 @@
...
+    world);
+  ES_loss_LS += ES_loss_step; // energy lost till now
+}
+
+/* ---------------------------------------------------------------------- */
+
+
+void FixESFriction::post_run()
+{
+  if(my_rank==0){
```
akohlmey, line 185: See my previous comments about compacting and simplifying output.

src/USER-ESFRICTION/fix_esfriction.h
```diff
@@ -0,0 +1,72 @@
...
+  double get_kenergy(double, double, double, double);
+
+  double const ev_to_joules = 1.6022e-19;
+  double const joules_to_ev = 1.0 / ev_to_joules;
+  double const Nav = 6.0221e23; // Avogadro's constant
+  double const bohr_radius = 5.2917721067e-1 ; // Angstroms
+  double const h_cross = 1.0546e-34; // planck's constant/(2pi) in J-s
+  double const m_e = 9.1090e-31; // electron mass in Kg
+  double const e_charge = 1.2; // sqrt(eV-nm)
+  double const pi = 3.14159265359;
```
akohlmey, line 43: Please do not define your own constants where avoidable, but use the ones defined in force.h as much as possible and try to derive from them what is not directly available. Beyond that, please use the "best" values available in full precision e.g. from the NIST database and document when and where you retrieved them.

akohlmey (changes requested):
This needs some changes as indicated, but please work on resolving the issue of having multiple implementations of the same process included that should have a consistent naming scheme and some guidance to the user on how the different approaches differ.

akohlmey:
Please also add a simple (and small/fast) example input and output and make certain it is "valgrind clean".

ManojWarrier:
Thanks for the comments. Ill work on them and submit a new pull request. This is the first time I am submitting via git. Please bear with me.

akohlmey:
No worries about doing this the first time. You are not the first. There is
no need for a new submission. You can commit changes to the same branch and
push it to your repo to update this pull request. That will make it easier
to track changes.

akohlmey:
@ManojWarrier there has been no update here for 3 weeks. Because of that, we are now going to merge the other pending electron stopping changes as is and you will have to adapt to the example set by that without the option to align all three implementations first.

akohlmey:
@ManojWarrier since there has been no reaction to previous comments I will now close this pull request. You are most welcome to resubmit, but please follow the example set by the existing two fixes and add your variant in a similar fashion so that users can see that a) there are three variants available and b) what the specific differences of those variants are. Thanks for your understanding.

PR #2470, opened 2020-11-06
Fix issue with Kokkos::finalize and library interface

stanmoore1:
**Summary**

`Kokkos::finalize()​` can only be called once in `main()`. Currently the LAMMPS class destructor calls the `Kokkos::finalize()` function. However this doesn't work with the library interface if a user calls `lammps_open()` and `lammps_close()` multiple times in a row, which corrupts the Kokkos memory. This PR removes the `Kokkos::finalize()` call out of `KokkosLMP` destructor, and adds a new library function to finalize Kokkos.

**Related Issue(s)**

None

**Author(s)**

Stan Moore (SNL), reported by Calvin Anderson (Miami University)

**Licensing**

**Backward Compatibility**

Yes


akohlmey:
If Kokkos initialization is to be treated like MPI initialization, we should do it all the way, i.e. also have a `lammps_kokkos_init()` function and - more importantly - have a check so that Kokkos initialization is indeed only run once. That is taking under consideration that somebody wants to call LAMMPS (as a library) from a code that is also using Kokkos and has it already initialized. Similarly, both the `lammps_kokkos_init() and the `lammps_kokkos_finalize()` function should be "safe" to be called, i.e. they should not call the low level functions if Kokkos is already initialized or already finalized or never initialized. Please see the LAMMPS class and the `lammps_mpi_init()`, `lammps_mpi_finalize()` library functions.

stanmoore1:
@akohlmey I'd like to get the Kokkos developers to push some of the logic down into the Kokkos library instead of keeping track of it in LAMMPS. But I suppose I can add logic into LAMMPS in the meantime.

stanmoore1:
@akohlmey I believe I addressed your feedback in the comment above. The calls to Kokkos `Init/Finalize` are as safe as I can make them. I'm not sure it makes sense to add a `lammps_kokkos_init()` function to the library interface, since currently this is already always called in the KokkosLMP constructor when LAMMPS is created, and there is no way to not initialize Kokkos. However, the Init function is safe and a no-op if an external caller already initialized Kokkos. Please let me know what you think.

src/accelerator_kokkos.h
```diff
@@ -43,6 +43,10 @@
 #include "modify.h"
 #include "neighbor.h"
 
+namespace Kokkos {
+  typedef int InitArguements;
```
akohlmey, line 47: Shouldn't this be "InitArguments"?

stanmoore1: Yeah, it is unused so I removed it.

src/KOKKOS/kokkos.cpp
```diff
@@ -299,9 +299,27 @@ KokkosLMP::KokkosLMP(LAMMPS *lmp, int narg, char **arg) : Pointers(lmp)
...
-  Kokkos::finalize();
+}
+
+/* ---------------------------------------------------------------------- */
+
+void KokkosLMP::initialize(Kokkos::InitArguments args)
+{
+  if (!Kokkos::is_initialized()) {
+    if (is_finalized)
+      error->all(FLERR,"Kokkos package already finalized, cannot re-initialize");
```
akohlmey, line 311: The Intel compilers don't like this:
```CMakeFiles/lammps.dir/mnt/lammps/workspace/dev/pull_requests/compilation_tests/lammps/src/KOKKOS/atom_kokkos.cpp.o
/mnt/lammps/workspace/dev/pull_requests/compilation_tests/lammps/src/KOKKOS/kokkos.cpp(311): error: a nonstatic member reference must be relative to a specific object
        error->all(FLERR,"Kokkos package already finalized, cannot re-initialize");
```

stanmoore1:
Thanks @akohlmey. We could also just make `KokkosLMP::initialize` non-static if that is simpler.

akohlmey:
> Thanks @akohlmey. We could also just make `KokkosLMP::initialize` non-static if that is simpler.

Whether static or not is a question for how the Kokkos lib behaves. If the initialization status is a global one - and it appears that is the case here - then the variable must be static, otherwise a class member.

On top of that, I have been wondering, what would happen, if somebody tries to create a LAMMPS instance with 1 GPU assigned to Kokkos and then wants to create another LAMMPS instance with 2 GPUs assigned to Kokkos or using OpenMP threads (to be executed sequentially, e.g. for testing or benchmarking)?

stanmoore1:
I mean make the function non-static, not the variable. Then we wouldn't need to pass in the `error` class pointer.

stanmoore1:
>On top of that, I have been wondering, what would happen, if somebody tries to create a LAMMPS instance with 1 GPU assigned to Kokkos and then wants to create another LAMMPS instance with 2 GPUs assigned to Kokkos or using OpenMP threads (to be executed sequentially, e.g. for testing or benchmarking)?

I don't think this is supported in Kokkos, but I will ask on Slack.

akohlmey:
> I mean make the function non-static, not the variable. Then we wouldn't need to pass in the `error` class pointer.

sorry, my bad. 

I think it is better to keep those function declarations "symmetric". That has fewer surprises.

stanmoore1:
>On top of that, I have been wondering, what would happen, if somebody tries to create a LAMMPS instance with 1 GPU assigned to Kokkos and then wants to create another LAMMPS instance with 2 GPUs assigned to Kokkos or using OpenMP threads (to be executed sequentially, e.g. for testing or benchmarking)?

I asked the Kokkos developers and they said we only get one Kokkos initialization, then the parameters (like number of GPUs/node) are locked in place forever. So subsequent attempts to re-initialize Kokkos will be ignored in LAMMPS, unless `Kokkos::finalize()` is called, then it will error out.

akohlmey:
> I asked the Kokkos developers and they said we only get one Kokkos initialization, then the parameters (like number of GPUs/node) are locked in place forever. So subsequent attempts to re-initialize Kokkos will be ignored in LAMMPS, unless `Kokkos::finalize()` is called, then it will error out.

Hmm... "ignored" is not what I was hoping for.  At least a warning would be very useful. But generally, I would prefer a clean error, when a code is not doing what I ask it to do.

This is a minor issue, though, and would probably most commonly apply to using the LAMMPS library interface for testing.

stanmoore1:
>Hmm... "ignored" is not what I was hoping for. At least a warning would be very useful. But generally, I would prefer a clean error, when a code is not doing what I ask it to do.

Say you are creating multiple LAMMPS instances in the library interface. If the params are the same, then I don't think it should error out or even give a warning. If the params are different then I think an error would be good.

akohlmey:
> > Hmm... "ignored" is not what I was hoping for. At least a warning would be very useful. But generally, I would prefer a clean error, when a code is not doing what I ask it to do.
> 
> Say you are creating multiple LAMMPS instances in the library interface. If the params are the same, then I don't think it should error out or even give a warning. If the params are different then I think an error would be good.

Exactly my thinking.

stanmoore1:
OK so we need to make the `Kokkos::InitArguments` static, and compare new to old settings in the `KokkosLMP` constructor?

akohlmey:
> OK so we need to make the `Kokkos::InitArguments` static, and compare new to old settings in the `KokkosLMP` constructor?

Something like that. And with "impossible" default values, so that you can tell whether they have been set before or not.

stanmoore1:
>Something like that. And with "impossible" default values, so that you can tell whether they have been set before or not.

Done, though I haven't tested with the library interface yet.

akohlmey:
@stanmoore1 did some checks with the library interface and make a few minor tweaks to retain the (unchanged) initialization settings when doing multiple initializations with different settings. As far as I am concerned this is now ready for merging.

stanmoore1:
Great, thank you!

sjplimp (approved):

PR #2472, opened 2020-11-09
types in gather_atoms updated from 0 and 1 to LAMMPS_INT and LAMMPS_D…

liviabp:
…OUBLE to be consistent with library.cpp

**Summary**

gather_atoms in lammps.py failed to extract information as type==1 no longer meant double

**Related Issue(s)**

Within library.cpp, the changes 0--> LAMMPS_INT and 1-->LAMMPS_DOUBLE had been done for gather atoms only, and not in e.g. gather_atoms_concat or in scatter_atoms, causing some inconsistency in the code. Shouldn't this be addressed, especially that LAMMPS_DOUBLE=2 ? 

**Author(s)**

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/python/lammps.py
+++ b/python/lammps.py
@@ -1186,10 +1186,10 @@ class lammps(object):
   def gather_atoms(self,name,type,count):
     if name: name = name.encode()
     natoms = self.get_natoms()
-    if type == 0:
+    if type == LAMMPS_INT:
       data = ((count*natoms)*c_int)()
       self.lib.lammps_gather_atoms(self.lmp,name,type,count,data)
-    elif type == 1:
+    elif type == LAMMPS_DOUBLE:
       data = ((count*natoms)*c_double)()
       self.lib.lammps_gather_atoms(self.lmp,name,type,count,data)
     else: return None


rbberger:
@liviabp Thanks for pointing out the inconsistency. Indeed the change in `lammps_gather_atoms` seems to have been done by mistake. Because the gather/scatter methods were using different "magic numbers", they shouldn't have been touched at all for the last release to ensure backwards compatibility. Oh well...

@akohlmey I traced this back to 68b533d1b8c33a8d0ec3bb603acb0b313a54fb46. Looks like a line of a newer version sneaked in. How do you want to proceed? Undo that change for now and then have a separate PR for changing the gather functions as intended?

akohlmey:
> @akohlmey I traced this back to [68b533d](https://github.com/lammps/lammps/commit/68b533d1b8c33a8d0ec3bb603acb0b313a54fb46). Looks like a line of a newer version sneaked in. How do you want to proceed? Undo that change for now and then have a separate PR for changing the gather functions as intended?

Your call. Mid term `lammps_gather_atoms()` and its variants can probably be eliminated, since we have more generic functionality to do the same, but also work on computes and fixes.

rbberger:
Ok, changing `lammps_gather_atoms` back to using 0 and 1 for now. That's the less intrusive change. No point in breaking compatibility for something that might even go away.

liviabp:
Many thanks for checking and clarifying!

PR #2473, opened 2020-11-09
Gather bonds from python

robeme:
**Summary**

A Python routine that extracts a bonded topology from a LAMMPS instance. It is also capable of extracting the total number of bonds. A small example to illustrate the principle of it is attached to examples/python/gather_bonds.py, which extracts and prints the number of bonds and the explicit bonds of "data.spce".

I haven't added it to the documentation yet, because I wasn't sure if this is useful or not. But it would probably not require too much documentation.

**Author(s)**

Robert Meißner, TU Hamburg, Germany, robert.meissner@tuhh.de

**Licensing**

**Backward Compatibility**

Yes

**Implementation Notes**

I had some ambiguities, since a lot has changed in the python interpreter of LAMMPS since I made the first implementation a year ago. I have tried to modify my original implementation as much as possible. But what I am most surprised about is why the variable natoms is always declared as "int" in the routines but returned as "double" in get_natoms(). For get_nbonds() I followed the structure of get_natoms() - I just don't know if it's thus correct. Before I just checked it like this: "if (lmp->atom->nbonds > MAXSMALLINT) return 0;" in get_nbonds().

The gather_bonds() function does not consider doubles as "type", since the bonded topology is rather made of integers which it returns. I hope I implemented it correctly by not considering double types and sticked to the desired structure of the code. But at least the example works, though.

**Post Submission Checklist**

- [x] The source code follows the LAMMPS formatting guidelines (hope so ...)

We have used this method to find reactive partners in epoxides, see https://doi.org/10.1021/acs.macromol.0c02222.


python/lammps.py
```diff
@@ -703,6 +708,21 @@ def get_natoms(self):
...
+
+    Will be precise up to 53-bit signed integer due to the
+    underlying :cpp:func:`lammps_get_nbonds` function returning a double.
+
+    :return: number of bonds
+    :rtype: int
+    """
+    return int(self.lib.lammps_get_nbonds(self.lmp))
+    
+  # -------------------------------------------------------------------------
```
rbberger, line 725: This description doesn't match up with the library implementation, which returns an `int`.

src/library.cpp
```diff
@@ -571,6 +571,24 @@ double lammps_get_natoms(void *handle)
...
+   useful before call to lammps_get_bonds() so can pre-allocate vector
+------------------------------------------------------------------------- */
+
+int lammps_get_nbonds(void *handle)
+{
+  LAMMPS *lmp = (LAMMPS *) handle;
+
+  int nbonds = static_cast<int>(lmp->atom->nbonds);
+  if (nbonds > 9.0e15) return 0; // TODO:XXX why not -1?
+  if (!lmp->force->newton_bond) nbonds *= 2;
```
rbberger, line 585: This causes `extract_global` and `lammps_get_nbonds` to return two different values.

src/library.h
```diff
@@ -112,6 +112,7 @@ void  lammps_commands_string(void *handle, const char *str);
  * ----------------------------------------------------------------------- */
 
 double lammps_get_natoms(void *handle);
+int lammps_get_nbonds(void *handle);
```
rbberger, line 115: Should be `double` to be consistent.

rbberger (changes requested):
@robeme Thanks for your contribution. The reasoning behind using `double` for `get_natoms()` was to simplify the library interface. Since `double` can store a 53bit signed integer, it should be big enough for most systems. In case someone needs the `int64` value, they can still get it via `lammps_extract_global()`. Following the same rationale, you should implement this in the same way. `get_nbonds` should only return the raw value, any special logic for newton on/off should be in the caller.

src/library.cpp
robeme: Changing to `double` in the get_nbonds() method:
```
double lammps_get_nbonds(void *handle)
{
  LAMMPS *lmp = (LAMMPS *) handle;

  double nbonds = static_cast<double>(lmp->atom->nbonds);
  if (nbonds > 9.0e15) return 0; // TODO:XXX why not -1?
  return nbonds;
}
```

yields crazy numbers of nbonds in the example ('bond count:  1188987200') which is a bit weird? Shouldn't it work like this? Or is the variable nbonds in the atom class differently defined as compared to natoms?

rbberger: That's probably because you forgot to declare the return type. By default ctypes uses `int`, so you would truncate the double.

```
    self.lib.lammps_get_nbonds.argtypes = [c_void_p]
    self.lib.lammps_get_nbonds.restype = c_double
```

akohlmey: There should be no "get_nbonds()" functions added to the C-library interface, since the same functionality is available already https://docs.lammps.org/Library_properties.html#extract-system-settings
Adding "get_xxx()" convenience functions for everything will quickly become a maintenance nightmare.
The long term plan should be to get rid of "get_atoms()" as well (although that will be difficult, since that is likely used by many applications of the library interface) and instead break down the "lammps_extract_global()" function to a few more topical subsets, including one that allows to query atom style items (like bonds, angles, dihedrals, ellipsoids, bodies, etc.).
For the python interface it may be acceptable when using some automated way to map "get_nXXX()" to "extract_global('XXX')", but that is for @rbberger to decide as the current caretaker of the Python module. I strongly recommend against making these individual functions.

src/library.cpp
```diff
@@ -2587,6 +2605,94 @@ void lammps_scatter_atoms_subset(void *handle, char *name, int type, int count,
...
+    
+    int *offset;
+    lmp->memory->create(offset,natoms,"lib/gather:copy");
+    for (i = 0; i < natoms; i++) offset[i] = 0;
+    
+    tagint *tag = lmp->atom->tag;
+    int nlocal = lmp->atom->nlocal;
+    int *num_bond = lmp->atom->num_bond;
+  
+    for (i = 0; i < nlocal; i++) offset[tag[i]-1] = num_bond[i];
```
akohlmey, line 2661: The code will segfault here when `lmp->atom->molecular != Atom::MOLECULAR`. There will be no bonds with `Atom::ATOMIC` and there will be no arrays with `Atom::TEMPLATE` as the bond information in that case is provided by molecule templates.

src/library.cpp
```diff
@@ -2587,6 +2605,94 @@ void lammps_scatter_atoms_subset(void *handle, char *name, int type, int count,
...
+      flag = 1;
+    if (lmp->atom->natoms > MAXSMALLINT) flag = 1;
+    if (flag) {
+      if (lmp->comm->me == 0)
+        lmp->error->warning(FLERR,"Library error in lammps_gather_bonds");
+      return;
+    }
+    
+    int natoms = static_cast<int> (lmp->atom->natoms);
+    int nbonds = static_cast<int> (lmp->atom->nbonds);
```
akohlmey, line 2650: You need to check for overflows here. You may have more bonds than atoms and the number of bonds may double when `newton_bond` is off.

akohlmey (changes requested):
There are some checks that need to be added. An addition to the unittest tests would be appreciated, too.

src/library.cpp
robeme: Thank you very much for the many comments. Thanks @rbberger for the hint with the declaration. Actually `ctypes` is a bit new for me and I am glad that I learned something new again. I agree with @akohlmey and I will remove the `get_nbonds()` function because it is not really needed for the `gather_bonds()` function either.

python/lammps.py
robeme: I have removed `get_nbonds()` and used `get_thermo('bonds')` in the example to get the number of bonds which makes this function obsolete.

src/library.h
robeme: I have removed `get_nbonds()` and used `get_thermo('bonds')` in the example to get the number of bonds that make this function obsolete.

src/library.cpp
robeme: I have removed `get_nbonds()` and used `get_thermo('bonds')` in the example to get the number of bonds that make this function obsolete.

src/library.cpp
robeme: Checking now in the beginning of `gather_bonds()` if `Atom::MOLECULAR` is `TRUE` to avoid this issue.

src/library.cpp
robeme: Didn't I took care of this by using `if (!lmp->force->newton_bond) nbonds *= 2;` a line after I get `int nbonds = ...`?

akohlmey: the if statement only accounts for the number of stored bond information items, it does not protect it from overflowing a 32-bit int. In fact the doubling makes it even more likely. Overall, it is probably better to stick to bigint data types inside of library.cpp where possible and only use or convert to int what is accessible to the calling program.

akohlmey:
@robeme please let me know when you have completed any changes on this PR, since there would be some additional changes that the LAMMPS developers need to add for complete integration.

robeme:
Sure! I had to teach yesterday. So I was a little busy, but I'm on it :)

src/library.cpp
robeme: I have now checked if 'nbonds' will overflow when newton is switched off. But I was not sure if I should check if it generally overflows when newton is turned on. Isn't it already checked when LAMMPS initializes the system?

robeme:
I have changed the code and added many more comments to explain my approach a bit better. However, there are some things that are still a bit unclear to me how I can best implement them. The most critical is that I don't know how to return a `bigint` to the python interface. When I do, I get a segfault when I return the contents of a `bigint topo` into data... And yeah, I changed the `restype` in `lammps.py` to `self.c_bigint` accordingly this time :) ...

rbberger (changes requested):

python/lammps.py
```diff
@@ -1439,6 +1444,17 @@ def create_atoms(self,n,id,type,x,v=None,image=None,shrinkexceed=False):
     return self.lib.lammps_create_atoms(self.lmp, n, id_lmp, type_lmp, x_lmp, v_lmp, img_lmp, se_lmp)
 
   # -------------------------------------------------------------------------
+  
+  # return vector of bonds gathered across procs
+  # count should be 3
+
+  def gather_bonds(self):
+    nbonds = int(self.get_thermo('bonds'))
+    data = ((3*nbonds)*c_int)()
```
rbberger, line 1453: this should be `self.c_tagint` instead of `c_int`

robeme: Sorry, I am a bit out of practice. Thanks for all the advice, help and patience with me. I've changed it accordingly.

src/library.cpp
```diff
@@ -2587,6 +2587,119 @@ void lammps_scatter_atoms_subset(void *handle, char *name, int type, int count,
...
+    // move all element to the right and do cumulative sum of elements
+    // in order to get an array which could be used as an offset to 
+    // index all bonds across all procs
+    for(i = natoms-2; i >= 0; i--) offset[i+1] = offset[i]; 
+    for(i = 1; i < natoms; i++) offset[i+1] += offset[i]; 
+    offset[0] = 0; // start with offset 0
+    
+    // topo stores the bond type and the connected atoms of every atom.
+    // Use range from offset to find bonds of specific atoms in topo
+    int *topo;
```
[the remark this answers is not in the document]
robeme: I've changed it accordingly.

robeme:
Hey everyone! Happy New Year, even if it's already a bit late :) Is there anything left to do now or can your requests be considered done?

akohlmey:
> Hey everyone! Happy New Year, even if it's already a bit late :) Is there anything left to do now or can your requests be considered done?

I don't have time to carefully review all the details, so I am answering based on what I recall from memory.

The obvious task that needs to be done is to update this PR with the changes to the LAMMPS python module from upstream as those are currently causing a merge conflict that blocks any attempt of a merge.

As far as I recall, the local overflow handling has not been properly updated. You have to keep in mind that LAMMPS uses a distributed data model, where you can run in parallel systems larger than what can be handled/indexed on a single MPI rank. For example the bond information is stored with one or both of the bond atoms and bond lists are then only compiled locally, while the list for output to a data or restart file is collected on the fly (hence the unsorted output due to atoms migrating between MPI ranks). Since you collect everything into a single array, you have to make certain that the storage and sizes and MPI communication are sufficient to do this correctly.

Beyond that, you probably need to update your examples and run some tests to make certain everything is working as intended.
You are encouraged to add unit test examples for this functionality to the unittest/python folder similar to existing tests, that will both demonstrate and exercise all typical (and incorrect?) uses so they can be confirmed to remain functional also with future changes to the python module.

akohlmey:
@robeme it has been a very long time since I can revisit this pull request again and look into the details of its implementation.
After some thinking and looking at what is done in your pull request, I have decided to implement this feature in an alternate fashion.  We have a method to store bond info into a (local) buffer already implemented in the `AtomVec` class that is used for writing the Bonds section of a data file. All that would be required after that is a call to `MPI_Allgatherv()` to gather the data into a global buffer. And the data in that buffer would be containing each bond only once, regardless of the setting `newton_bond` setting. This will do most of the processing only locally on the individual MPI ranks and thus reduces the memory usage substantially.

akohlmey:
@robeme the re-implementation of your pull request is now complete. It also includes a numpy wrapper that directly returns the data as numpy array.
With the new pull request you can do from python either:
```python
nbonds, bondlist = lmp.gather_bonds()
print("type atom1 atom2:")
for i in range(0,nbonds):
     print(bondlist[3*i], bondlist[3*i+1], bondlist[3*i+2])
```
or 
```python
bondlist = lmp.numpy.gather_bonds()
print("type atom1 atom2:")
for bond in bondlist:
     print(bond[0], bond[1], bond[2])
```

robeme:
Thanks @akohlmey ! This does indeed sound like a much better approach than the one I had originally proposed. I am glad to hear that this is now implemented :+1:

PR #2477, opened 2020-11-11
Misc. KIM updates

yafshar:
**Summary**

This PR contains KIM updates including:

- Clean up the interface
- Update the error messages to be more expressive
- Update the interface and remove the `strtok` to be compliant with LAMMPS
- Update & remove `sstream` where possible and use `fmt::format` 
- Improve the `kim_query` & `kim_interactions` commands
- Remove the unnecessary `fclose`  in `kim_interactions` which causes issue on the c-python interface, when calling `kim_interactions` in a loop
- Extra cases to the unit tests

**Related Issue(s)**

**Author(s)**

Yaser Afshar @yafshar (UMN)

**Licensing**

**Backward Compatibility**

The pull request is backward compatibile

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/src/KIM/kim_init.cpp
+++ b/src/KIM/kim_init.cpp
@@ -92,7 +92,12 @@ void KimInit::command(int narg, char **arg)
   strcpy(user_units,arg[1]);
   if (narg == 3) {
     if (strcmp(arg[2],"unit_conversion_mode")==0) unit_conversion_mode = true;
-    else { error->all(FLERR,"Illegal kim_init command"); }
+    else { 
+      error->all(FLERR,fmt::format("Illegal kim_init command.\nThe argument "
+                                   "followed by unit_style {} is an optional "
+                                   "argument and when is used must "
+                                   "be unit_conversion_mode", user_units));
+    }
   } else unit_conversion_mode = false;
 
   char *model_units;
--- a/src/KIM/kim_init.cpp
+++ b/src/KIM/kim_init.cpp
@@ -187,6 +187,9 @@ void KimInit::determine_model_type_and_units(char * model_name,
   if (kim_error) 
     error->all(FLERR,"Unable to access KIM Collections to find Model");
 
+  auto logID = fmt::format("{}_Collections", comm->me);
+  KIM_Collections_SetLogID(kim_Coll, logID.c_str());
+
   kim_error = KIM_Collections_GetItemType(kim_Coll, model_name, &itemType);
   if (kim_error) error->all(FLERR,"KIM Model name not found");
   KIM_Collections_Destroy(&kim_Coll);
@@ -210,6 +213,8 @@ void KimInit::determine_model_type_and_units(char * model_name,
     model_type = MO;
 
     if (units_accepted) {
+      logID = fmt::format("{}_Model", comm->me);
+      KIM_Model_SetLogID(pkim, logID.c_str());
       *model_units = new char[strlen(user_units)+1];
       strcpy(*model_units,user_units);
       return;
@@ -231,6 +236,8 @@ void KimInit::determine_model_type_and_units(char * model_name,
                                      &units_accepted,
                                      &pkim);
         if (units_accepted) {
+          logID = fmt::format("{}_Model", comm->me);
+          KIM_Model_SetLogID(pkim, logID.c_str());
           *model_units = new char[strlen(systems[i])+1];
           strcpy(*model_units,systems[i]);
           return;
@@ -243,12 +250,15 @@ void KimInit::determine_model_type_and_units(char * model_name,
       error->all(FLERR,"KIM Model does not support the requested unit system");
     }
   } else if (KIM_CollectionItemType_Equal(
-               itemType, KIM_COLLECTION_ITEM_TYPE_simulatorModel)) {
+             itemType, KIM_COLLECTION_ITEM_TYPE_simulatorModel)) {
     KIM_SimulatorModel * kim_SM;
     kim_error = KIM_SimulatorModel_Create(model_name, &kim_SM);
     if (kim_error)
       error->all(FLERR,"Unable to load KIM Simulator Model");
     model_type = SM;
+
+    logID = fmt::format("{}_SimulatorModel", comm->me);
+    KIM_SimulatorModel_SetLogID(kim_SM, logID.c_str());
 
     int sim_fields;
     int sim_lines;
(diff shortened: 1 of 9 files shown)


src/KIM/kim_interactions.cpp
```diff
@@ -312,7 +315,6 @@ void KimInteractions::KIM_SET_TYPE_PARAMETERS(const std::string &input_line) con
           input->one(fmt::format("set type {} charge {}",ia+1,words[1]));
     }
   }
-  fclose(fp);
```
yafshar, line 302: @ellio167 This is the place, which causes issues on the C-Python interface when using `ase` & `lammpslib` and calling the `kim_interactions` multiple times in a loop.

ellio167: Thanks;  Yes, the second call to `fclose` is definitely a bug; both BSD (macOS) and Linux manpages indicate that calling `fclose` a second time gives undefined behavior.

akohlmey: Like with pointers to memory location and malloc()/free(), the usual convention is to set a file pointer to NULL (or nullptr in C++11+) after it has been freed with fclose(). Then you can always add a check if the pointer is non-zero before attempting to close it. Of course that also requires the pointer to be initialized before use.

src/KIM/kim_init.cpp
```diff
@@ -68,6 +68,7 @@
 #include "modify.h"
 #include "universe.h"
 #include "variable.h"
+#include "fmt/format.h"
```
akohlmey, line 71: no need to import `fmt/format.h` here since the `KimInit` class is derived from `Pointers`. Any class derived from `Pointers` can assumed that `lmptype.h`, `mpi.h`, `cstddef`, `cstdio`, `string`, `utils.h` and `fmt/format.h` are already included through `pointers.h`.

src/KIM/kim_interactions.cpp
```diff
@@ -66,9 +67,12 @@
 #include "input.h"
 #include "modify.h"
 #include "update.h"
+#include "fmt/format.h"
```
akohlmey, line 70: see my commend for `kim_init`

src/KIM/kim_param.cpp
```diff
@@ -57,16 +57,21 @@
...
 #include "kim_param.h"
 
+#include "comm.h"
 #include "error.h"
 #include "fix_store_kim.h"
 #include "force.h"
 #include "input.h"
 #include "modify.h"
 #include "pair_kim.h"
+#include "utils.h"
```
akohlmey, line 67: like before. list of include statements can be pruned.

src/KIM/pair_kim.cpp
```diff
@@ -66,9 +66,13 @@
 #include "neigh_request.h"
 #include "neighbor.h"
 #include "update.h"
+#include "utils.h"
```
akohlmey, line 69: here, too.

unittest/commands/test_kim_commands.cpp
```diff
@@ -302,6 +336,133 @@ TEST_F(KimCommandsTest, kim_property)
...
+
+    squery = "kim_query a0 get_lattice_constant_cubic crystal=[fcc] "
+             "species=[Al]";
+    TEST_FAILURE(".*ERROR: OpenKIM query failed:.*", lmp->input->one(squery););
+
+    squery = "kim_query a0 get_lattice_constant_cubic crystal=[fcc] "
+             "units=[\"angstrom\"]";
+    TEST_FAILURE(".*ERROR: OpenKIM query failed:.*", lmp->input->one(squery););
+
+    // if (!verbose) ::testing::internal::CaptureStdout();
```
akohlmey, line 423: what is the deal with the commented out code?

yafshar: @akohlmey the commented ones require a network and special model. Thus they are not general, but I would check them in my update. If you think it is better I can remove them.

src/KIM/kim_init.cpp
yafshar: @akohlmey I updated the include statements

src/KIM/kim_interactions.cpp
yafshar: @akohlmey I updated the include statements

src/KIM/kim_param.cpp
yafshar: @akohlmey I updated the include statements

src/KIM/pair_kim.cpp
yafshar: @akohlmey I updated the include statements

unittest/commands/test_kim_commands.cpp
akohlmey: I would suggest you add an option to `cmake/Modules/Packages/KIM.cmake`, e.g. KIM_EXTRA_UNITTESTS which defaults to OFF and then use that setting to add a corresponding pre-processor define when compiling `test_kim_commands` in `unittest/commands/CMakeLists.txt`. Then you can have the tests included, but only if they are explicitly enabled. The CMake variable should be documented in `doc/src/Build_extras.rst` where it explains what pre-requisites are required.

That way the tests can be easily available to people that have the prerequisites but don't automatically fail with setups that do not have internet access or the required modules installed.

yafshar: @akohlmey sounds great. I would do it. We also have another feature to add which only depends on the new `kim-api` and needs the same pre-processor definition in the `KIM.cmake`. If you agree I will add these two in a new PR  after PR #2488 and the current one are merged into lammps.

yafshar:
@akohlmey My last commit was only to remove unnecessary headers, and the unit tests are done on my systems (both on macOS and Linux/centos) with no issue. So I do not know why the unit_tests build failed here and I can not see the log file.

akohlmey:
> @akohlmey My last commit was only to remove unnecessary headers, and the unit tests are done on my systems (both on macOS and Linux/centos) with no issue. So I do not know why the unit_tests build failed here and I can not see the log file.

It looks like the "kim_commands" test has failed.
If you cannot get what you are looking for, you can switch back to "classic" mode (click on the "[->]" symbol on the top right) and then you can get the logs for the individual runs.

yafshar:
> > @akohlmey My last commit was only to remove unnecessary headers, and the unit tests are done on my systems (both on macOS and Linux/centos) with no issue. So I do not know why the unit_tests build failed here and I can not see the log file.
> 
> It looks like the "kim_commands" test has failed.
> If you cannot get what you are looking for, you can switch back to "classic" mode (click on the "[->]" symbol on the top right) and then you can get the logs for the individual runs.

@akohlmey I commented out the line, which caused an issue. Anyhow this error does not make any sense. It was working correctly before, and the model is part of the `KIM-API` examples distributed by KIM. I do not understand what is going wrong there, but I have tried on several different systems and have not seen this issue.

ellio167 (approved):

unittest/commands/test_kim_commands.cpp
akohlmey: as soon as this is addressed soon, there is not going to be much of a problem. you just need to be aware that when we see large chunks of commented out code without explanations of it begin added for debugging or otherwise, it will eventually be removed.

akohlmey (approved):

PR #2479, opened 2020-11-13
Updated AGNI pair files and added new AGNI force fields

james1123581321:
**Summary**

I have updated the pair_agni files to include a second fingerprinting scheme used in several new papers. I have also included 4 new force field files for Pt, Al, and Cu. Nothing changes from the user's point-of-view, as the toggle for the fingerprinting schemes is contained within the force field files. Therefore, the pair_style and pair_coeff commands remain the same.

**Related Issue(s)**

None that I am aware of. This is just a slight modification of the pair_agni files and the introduction of new force fields.

**Author(s)**

James Chapman, Lawrence Livermore National Lab, work email: chapman37@llnl.gov , personal email: james1123581321@outlook.com

**Licensing**

**Backward Compatibility**

The updated code has been made backwards-compatible with the old Al_jpc.agni potential that is already included as part of LAMMPS. No changes are needed from the user's perspective.

**Implementation Notes**

I just added a toggle for different fingerprinting schemes, that is turned on from the force field file. A small section of code was added regarding the reading of the training data. There are now 2 blocks of code, 1 for the AL_jpc.agni file format, and 1 for the new format. If this is unacceptable, I can update the format in the Al_jpc.agni file. However, I figured as this is only done once during the simulation, that this was easier.

The new force fields have been validated in their respective papers, which can be found in the force field file headers. I also tested a short NVT run for each force field to ensure that there were no obvious bugs in the new code. 

**Post Submission Checklist**


james1123581321:
Hello,
I'm not sure how this gets processed now that I've submitted the merge request, but I know @akohlmey provided insight the last time that the AGNI potential files were updated. Any help you can provide now is greatly appreciated.

rbberger:
@james1123581321 looks like you still have to include `force.h` in order to see `CENTROID_NOTAVAIL`.

james1123581321:
Thank you, I just saw that there were changes made after my initial commits. I've pulled the new changes from lammps-master.

src/USER-MISC/pair_agni.cpp
```diff
@@ -68,7 +68,6 @@ static const char cite_pair_agni[] =
   " year      = {2017},\n"
   "}\n\n";
 
-#define AGNI_VERSION 1
```
akohlmey, line 71: it would be better to use something like
```
enum { AGNI_VERSION_UNKNOWN, AGNI_VERSION_1, AGNI_VERSION_2 };
```
here and then use those constants below instead of "0", "1", and "2" to make it more obvious what is being checked for.

src/USER-MISC/pair_agni.cpp
```diff
@@ -187,7 +188,13 @@ void PairAGNI::compute(int eflag, int vflag)
         const double wZ = cF*delz/r;
 
         for (k = 0; k < iparam.numeta; ++k) {
-          const double e = exp(-(iparam.eta[k]*rsq));
+          	double e = 0.0;
```
akohlmey, line 191: Please retain the LAMMPS code formatting style. Especially:
- use only spaces, no TABs
- use 2 spaces indentation (not 8 or 4)
- no trailing whitespace
- wrap lines of 80 characters and more

src/USER-MISC/pair_agni.cpp
```diff
@@ -187,7 +188,13 @@ void PairAGNI::compute(int eflag, int vflag)
...
+
+		if(ver == 1) //FP1
+			e = exp(-(iparam.eta[k]*rsq));
+		else if(ver == 2) //FP2
```
akohlmey, line 195: please remove those "FP1" "FP2" comments

src/USER-MISC/pair_agni.cpp
```diff
@@ -187,7 +188,13 @@ void PairAGNI::compute(int eflag, int vflag)
...
+			e = (1 / (square(iparam.eta[k]) * iparam.gwidth * sqrt(2 * MathConst::MY_PI))) * exp(-(square(r - iparam.eta[k])) / (2 * square(iparam.gwidth)));
```
akohlmey, line 196: - please use "1.0" instead of "1" for clarity when numbers are meant to be (and promoted to) floating point numbers.
- there is a `MathConst::MY_2PI`

src/USER-MISC/pair_agni.h
```diff
@@ -48,8 +49,8 @@ class PairAGNI : public Pair {
   int *elem2param;              // mapping from element pairs to parameters
   int *map;                     // mapping from atom types to elements
   int nparams;                  // # of stored parameter sets
+  int ver;			// version of fingerprint
```
akohlmey, line 52: please use a more descriptive variable name like "version" or "format_version" or "revision"

akohlmey (changes requested):
Thanks for your submission. I've made some suggestions for cosmetic changes that represent our desire to make the LAMMPS code more consistent and readable.

akohlmey:
@james1123581321 if you feel like taking on additional little programming projects, you may consider refactoring the potential file reader in a similar fashion to what has already been done with several manybody potentials. That should help to simplify the code some and make it check its input better.
Another option to look into would be to add a unit test or two, this can be as simple as adding 1-2 new YAML files to the `unittest/force-styles/tests` folder.

james1123581321:
@akohlmey Thanks for your comments. I might as well add in the changes to the file reader, considering the pair style hasn't been touched in 4+ years. Do you know which potentials have refactored their code, so that I can match AGNI's with their code style? I'm assuming that the changes in the file reader have been standardized among the various potentials?

akohlmey:
> @akohlmey Thanks for your comments. I might as well add in the changes to the file reader, considering the pair style hasn't been touched in 4+ years. Do you know which potentials have refactored their code, so that I can match AGNI's with their code style? I'm assuming that the changes in the file reader have been standardized among the various potentials?

You can look at all styles that use the `PotentialFileReader` class:
```
$ grep -l potential_file_reader src/*.cpp src/*/*.cpp
src/pair_coul_streitz.cpp
src/potential_file_reader.cpp
src/GPU/pair_eam_alloy_gpu.cpp
src/GPU/pair_eam_fs_gpu.cpp
src/KOKKOS/pair_eam_alloy_kokkos.cpp
src/KOKKOS/pair_eam_fs_kokkos.cpp
src/MANYBODY/pair_adp.cpp
src/MANYBODY/pair_airebo.cpp
src/MANYBODY/pair_bop.cpp
src/MANYBODY/pair_comb3.cpp
src/MANYBODY/pair_comb.cpp
src/MANYBODY/pair_eam_alloy.cpp
src/MANYBODY/pair_eam.cpp
src/MANYBODY/pair_eam_fs.cpp
src/MANYBODY/pair_gw.cpp
src/MANYBODY/pair_gw_zbl.cpp
src/MANYBODY/pair_nb3b_harmonic.cpp
src/MANYBODY/pair_polymorphic.cpp
src/MANYBODY/pair_sw.cpp
src/MANYBODY/pair_tersoff.cpp
src/MANYBODY/pair_tersoff_mod_c.cpp
src/MANYBODY/pair_tersoff_mod.cpp
src/MANYBODY/pair_tersoff_zbl.cpp
src/MANYBODY/pair_vashishta.cpp
src/USER-INTEL/pair_eam_alloy_intel.cpp
src/USER-INTEL/pair_eam_fs_intel.cpp
src/USER-MISC/pair_tersoff_table.cpp
src/USER-OMP/pair_eam_alloy_omp.cpp
src/USER-OMP/pair_eam_fs_omp.cpp
src/USER-OMP/pair_tersoff_zbl_omp.cpp
```

akohlmey:
Please also note that your pull request re-introduces the file `potentials/TABTP_10_10.mesont` which we have unbundled from the distribution because of its size. It needs to be removed. The unexpected growth of the LAMMPS source code distribution due to including large potential files is becoming a concern, and the AGNI potentials are on the large side of what we have as well.
We have since added a mechanism to have those unbundled and only downloaded when a corresponding package is installed. 
Will have to discuss with the other LAMMPS core developers where we set a threshold to unbundle potential files as that would require to create a new package and then move the corresponding potential implementation there.

In addition please consider adding a `UNITS: metal` tag to the potential files, so that the selection of units can be checked against (both the `utils::open_optential()` function and the `PotentialFileReader` class will check the current units against the value of this tag and refuse to continue if there is a mismatch. With some additional programming effort is would even be possible to transparently convert the potential file data during the read to real units as is done with several manybody styles like EAM or Tersoff.

akohlmey:
@james1123581321 can you give us a quick update on the status of your pending changes? Also, please let us know if you need help or have questions.

james1123581321:
@akohlmey Sorry, I've been working on another code the last few weeks so I haven't had too much time to look at this. I did plan on adding the changes you'd mentioned and pushing them to my branch by the end of this year. I plan on adding the new file reader, and making changes to the potential files to be more consistent with some of the other manybody files (as well as addressing the cosmetic changes you'd suggested). One concern that I did have was the potential file's size, which you mentioned as well. Is there anything from my end that currently needs to be done regarding this? I can always convert the files to binaries, which could reduce their size, but I don't know if that fits with the format you'd like to keep for LAMMPS. Let me know your thoughts.

akohlmey:
> @akohlmey Sorry, I've been working on another code the last few weeks so I haven't had too much time to look at this.

no problem. sometimes pull requests get abandoned and thus I occasionally ask for a status update.

> One concern that I did have was the potential file's size, which you mentioned as well. Is there anything from my end that currently needs to be done regarding this?

there are essentially two options: 1) shrink the files (applies to some legacy files in the distribution as well, so you are not alone) or 2) move them to an offsite location and have them downloaded only on demand when the pair style is installed.

> I can always convert the files to binaries, which could reduce their size, but I don't know if that fits with the format you'd like to keep for LAMMPS. Let me know your thoughts.

since we make use of 'gzip' and 'zlib' in a few places, perhaps it is time to find a minimal/embeddable implementation and and have gzip support always included in LAMMPS and then this would become a non-issue. that could help with some other larger files like data files and log files as well.

akohlmey:
> I can always convert the files to binaries, which could reduce their size, but I don't know if that fits with the format you'd like to keep for LAMMPS.
Plain binary files are a bit tricky because 1) their header should be human readable and 2) there needs to be a simple way to detect that (e.g. with another tag in the first line) and - most importantly - 3) there needs to be a way to encode/detect endianness so that the files will be portable across platforms (not as big a deal as it used to be, since very few people run on big endian CPUs these days, and platforms like SPARC or IBM Power /w AIX are not very common anymore).

james1123581321:
@akohlmey I've pushed a new build that incorporates your cosmetic requests. I've also started updating the potential reader, but it's not finished yet, so it'll crash upon testing for now. I'm hoping to have it wrapped up by the end of this week.

james1123581321:
@akohlmey You're probably right about the binary potential files, there's just a lot of headache involved, especially since there seems to be more robust options available. I think gzip would be a great idea, and it would certainly solve most of the issues with large files (especially with text files generated by LAMMPS), but to have default support within LAMMPS is probably out of the scope of this pull request.

akohlmey:
> @akohlmey You're probably right about the binary potential files, there's just a lot of headache involved, especially since there seems to be more robust options available.

> I think gzip would be a great idea, and it would certainly solve most of the issues with large files (especially with text files
> generated by LAMMPS), but to have default support within LAMMPS is probably out of the scope of this pull request.

yes.

there are three issues to keep in mind: 1) the storage used by the checkout folder. this is usually dominated by the size of the object files, libraries and executables, especially when debug info is included (which is and should be by default).
2) the size of the repository itself (and thus the time it takes to clone it), which uses compression already. this size is dominated by a few large binary files that crept into the repo by accident a long time ago and cannot be removed without rewriting the entire history. this can be alleviated by making a shallow clone (e.g. for integration testing, we clone with --depth 100, which will skip downloading all the old stuff).
3) the size of files in the folder where you run. that is where binary files and gzip can help. that is already possible for data files and some trajectory files but the savings are moderate. it is usually more effective to make certain to not output trajectory data more often than needed and by writing out only information that is needed.

for now, I don't think there is a need to worry until you want to include several 100s of potential files. ;-)

there are some ideas floating around about implementing alternate approaches to some of the fundamental data files in LAMMPS based on some standardized "envelope" format, but that is premature to consider right now and it will at best be available in parallel with existing support for a while.

james1123581321:
@akohlmey I've added in the new file reader and followed the pair_vashishta code, as our workflow is very similar, but the code still crashes. It runs fine in serial, but on multiple processes it seg-faults. I'm not sure why it's crashing though, since even on multiple processes the forces are predicted correctly. The crash seems to happen after pair::compute has been called, so I'm still investigating why this happens, but right now I'm not sure.

akohlmey:
> @akohlmey I've added in the new file reader and followed the pair_vashishta code, as our workflow is very similar, but the code still crashes. It runs fine in serial, but on multiple processes it seg-faults. I'm not sure why it's crashing though, since even on multiple processes the forces are predicted correctly. The crash seems to happen after pair::compute has been called, so I'm still investigating why this happens, but right now I'm not sure.

I just compiled your branch and it fails for me even in serial. The example in `examples/USER/misc/agni` fails with
```
Reading agni potential file Al_jpc.agni with DATE: 2017-02-24
ERROR: No suitable parameters for requested element found (src/USER-MISC/pair_agni.cpp:396)
Last command: pair_coeff * * Al_jpc.agni Al
```

And if I replace the potential file with Al_Batra_2019.agni I get instead:
```
Reading agni potential file Al_Batra_2019.agni with DATE: 2020-11-12
ERROR on proc 0: Not a valid integer number: '+1.00000000E+00' (src/USER-MISC/pair_agni.cpp:467)
```

So either you have not pushed all your changes or there are variables that are not properly initialized and thus have different initial values for you and me.

james1123581321:
@akohlmey It looks like I forgot to update the potential files within my branch. I'll do that now and push the changes shortly.

akohlmey:
Compiling the code with clang produces this warning showing incorrect use of the `std::string` class:
```
.../lammps/src/USER-MISC/pair_agni.cpp:392:31: warning: object backing the pointer will be destroyed at the end of the full-expression [-Wdangling-gsl]
        const char* element = values.next_string().c_str();
```
It would be better to use `std::string element = values.next_string();` which makes a real copy.
This would explain the first of the two failures I've reported.

akohlmey:
Once you have code that compiles and runs correctly in serial, I can look into why it crashes in parallel. There are ways to do parallel debugging of LAMMPS even without an expensive parallel debugger. ;-)

james1123581321:
@akohlmey You should be able to run tests now using the files in the examples folder.

akohlmey:
> @akohlmey You should be able to run tests now using the files in the examples folder

I don't understand this. This *modifies* the "old" potential file. Didn't you previously state that backward compatibility is preserved?

src/USER-MISC/pair_agni.cpp
```diff
@@ -347,132 +358,123 @@ double PairAGNI::init_one(int i, int j)
...
-    } else if (params && (nwords == nparams+1) && (strcmp(words[0],"element") == 0)) {
       wantdata = -1;
+
+      values = reader.next_values(2);
+      values.next_string(); // ignore
       for (i = 0; i < nparams; ++i) {
+        std::string element = values.next_string();
         for (j = 0; j < nelements; ++j)
-          if (strcmp(words[i+1],elements[j]) == 0) break;
+          if (strcmp(element.c_str(),elements[j]) == 0) break;
```
akohlmey, line 394: Since `element` is a C++ string this line can be simplified to
```
if (element == elements[j]) break;
```

src/USER-MISC/pair_agni.cpp
```diff
@@ -347,132 +358,123 @@ double PairAGNI::init_one(int i, int j)
...
+      values = reader.next_values(2);
+      values.next_string(); // ignore
+      atomic_feature_version = values.next_int();
+      if (atomic_feature_version != AGNI_VERSION_1 && atomic_feature_version != AGNI_VERSION_2)
+        error->all(FLERR,"Incompatible AGNI potential file version");
+
+      values = reader.next_values(2);
+      values.next_string(); // ignore
+      params[curparam].numeta = values.next_int();
+      params[curparam].eta = new double[params[curparam].numeta];
```
akohlmey, line 416: Here is a parallel programming problem. This part of the code only runs on MPI rank 0, so the memory is only allocated on that rank. And even if it was allocated on all MPI ranks, you cannot broadcast the pointer to all MPI ranks since "new" or malloc() will return different addresses on different MPI ranks. Thus this array and the `xU` one in the next line need to be handled special. You can keep the code as is, but after the broadcast of the struct you need to repeat the allocations with "new" on all MPI ranks != 0 and then broadcast the corresponding arrays individually.

src/USER-MISC/pair_agni.cpp
```diff
@@ -347,132 +358,123 @@ double PairAGNI::init_one(int i, int j)
...
+      values = reader.next_values(2);
+      values.next_string(); // ignore
+      params[curparam].cut = values.next_double();
+      
+      values = reader.next_values(2);
+      values.next_string(); // ignore
+      params[curparam].numtrain = values.next_int();
+      params[curparam].alpha = new double[params[curparam].numtrain];
+      for (i = 0; i < params[curparam].numeta; ++i)
+        params[curparam].xU[i] = new double[params[curparam].numtrain];
```
akohlmey, line 437: Here are more allocations that need to be repeated and data communicated for MPI rank != 0.

james1123581321:
@akohlmey It was originally backwards compatible, but as the new file-reader format requires things to be more "hard-coded" I had to make some organizational changes to the old Al_jpc.agni file. I rearranged the parameters above endVar to fit the new format, and I removed the line counter from the lines below endVar. I can keep the backwards compatibility, but I think this is the more streamlined and generalized solution.

src/USER-MISC/pair_agni.cpp
```diff
@@ -347,132 +358,123 @@ double PairAGNI::init_one(int i, int j)
...
-  delete [] words;
+  
+  MPI_Bcast(&nparams, 1, MPI_INT, 0, world);
+  MPI_Bcast(&atomic_feature_version, 1, MPI_INT, 0, world);
+  if(comm->me != 0) {
+    params = memory->create(params,nparams,"pair:params");
+    memset(params,0,nparams*sizeof(Param));
+  }
+  MPI_Bcast(params, nparams*sizeof(Param), MPI_BYTE, 0, world);
 }
```
akohlmey, line 478: A significant difference between the old style file reader and the new reader is that you now do reading *and* parsing only on MPI rank 0, whereas previously you would read a line then broadcast and then parse concurrently on all MPI ranks in parallel.

akohlmey:
> @akohlmey It was originally backwards compatible, but as the new file-reader format requires things to be more "hard-coded" I had to make some organizational changes to the old Al_jpc.agni file. I rearranged the parameters above endVar to fit the new format, and I removed the line counter from the lines below endVar. I can keep the backwards compatibility, but I think this is the more streamlined and generalized solution.

This is a misconception on how you can or should use the file reader code. It looks to me as if you are making matters worse by skipping over strings that could be used to identify what section of the input you have. The new reader code is as flexible as the old version. It is you inflexible implementation that is the problem, what stops you to just read the string and then test against its value and perform the corresponding parsing operation. Ignoring descriptive information in a file format is almost always a bad idea. At the very least there should be a check for compatibility and a suitable warning, but it is better to have a sufficiently flexible reader code if the format is flexible (e.g. does has sections without a fixed order by tagged entries).

james1123581321:
@akohlmey That's fair, and probably the right call in the end. I'll add backwards compatibility into the reader, shouldn't take too long.

james1123581321:
@akohlmey I've updated the file-reader to be backwards compatible. You're right, the new reader is much nicer to work with as an end-user once you know how to use it. I also reverted the Al_jpc.agni file to its previous version, so now nothing has changed with that file from the main branch. I've tried to broadcast things correctly after the file is read in, but it still crashes in parallel so I'm obviously missing something. This version will run in serial though with the example folder.

akohlmey (approved):

akohlmey:
@james1123581321 I fixed the parallel communication issue. You had:
```
    MPI_Bcast(&params[i].alpha, params[i].numtrain, MPI_DOUBLE, 0, world);
    MPI_Bcast(&params[i].eta, params[i].numeta, MPI_DOUBLE, 0, world);
```
But it should have been either:
```
    MPI_Bcast(params[i].alpha, params[i].numtrain, MPI_DOUBLE, 0, world);
    MPI_Bcast(params[i].eta, params[i].numeta, MPI_DOUBLE, 0, world);
```
or:
```
    MPI_Bcast(&params[i].alpha[0], params[i].numtrain, MPI_DOUBLE, 0, world);
    MPI_Bcast(&params[i].eta[0], params[i].numeta, MPI_DOUBLE, 0, world);
```
That would corrupt the params struct data on the MPI ranks > 0 and then fail when broadcasting the "xU" array.
I've also change that array to be allocated with `memory->create()` so the data is allocated in a contiguous chunk and thus the communication can be simplified. Similarly, I've removed redundant communication calls.

In addition, I have updated the source formatting to more closely follow the common style in LAMMPS, fixed a few memory leaks, particularly when using the `pair_coeff` command multiple times, ported the changes to the USER-OMP version of the pair style and created new logfiles for the example inputs 

You may want to add your name to the list of contributing authors in `pair_agni.cpp`.
Otherwise, it looks ready to me, so I have requested reviews from other LAMMPS developers.

james1123581321:
@akohlmey Thanks for your additions/fixes as well as your explanation. I've never broadcasted structs before, so your explanation helps a lot. With my last few additions I have nothing else to add from my end. Thanks again for your help.

src/USER-MISC/pair_agni.cpp
```diff
@@ -66,9 +68,24 @@ static const char cite_pair_agni[] =
   " number    = {1},\n"
   " pages     = {511},\n"
   " year      = {2017},\n"
+  "}\n\n"
+  "@article{huan2019jpc,\n"
```
akohlmey, line 72: There are **far** too many citations here. There should only be citations that describe the method in use, ideally only one citing the most relevant publication. Since you have two variants of the functional form, you could define two citation strings (with the one most relevant publication for each variant) and place the `citeme->add()` call so that only the one variant in use is added. All other related citations can be added to the manual, but should not be included in the code. Thanks.

james1123581321: We actually have a publication that compares the 2 functional forms, so I'll just use that to cover both versions.

sjplimp (approved):

PR #2469, opened 2020-11-05
Fix a bug in pair_kim

yafshar:
**Summary**

This PR includes:
- Bugfix in pair_kim.
- Clean up the code and remove the unnecessary else condition and improve the readability
- Following the latest LAMMPS update #2481, the check is updated as `vflag_atom` for zero vs. non-zero.

This bug results in the wrong computation of stress/atom in LAMMPS when using pair style Kim.
This is created after the PR #1704 , where the `vflag_atom` is updated from `vflag_atom = vflag / 4;` to `vflag_atom = vflag & 4;`. It results in passing the wrong pointer to the KIM-API.

**Author(s)**

Yaser Afshar @yafshar (UMN)

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

This PR is fully backward compatible.

**Post Submission Checklist**

- [x] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [x] The source code follows the LAMMPS formatting guidelines
- [ ] Suitable new documentation files and/or updates to the existing docs are included
- [ ] The added/updated documentation is integrated and tested with the documentation build system
- [x] The feature has been verified to work with the conventional build system
- [x] The feature has been verified to work with the CMake based build system
- [ ] Suitable tests have been added to the unittest tree.
- [ ] A package specific README file has been included or updated
- [ ] One or more example input decks are included

**Further Information, Files, and Links**

PR #2481, opened 2020-11-13
Introduce enums for energy and virial flags

rbberger:
**Summary**

Define energy and virial constants in `force.h` to reduce magic numbers in code.
Also defines constants `CENTROID_SAME`, `CENTROID_AVAIL` and `CENTROID_NOTAVAIL`.

**Related Issue(s)**

#2469 

**Author(s)**

Richard Berger (Temple U) @rbberger 

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/src/CLASS2/pair_lj_class2.cpp
+++ b/src/CLASS2/pair_lj_class2.cpp
@@ -35,7 +35,7 @@ PairLJClass2::PairLJClass2(LAMMPS *lmp) : Pair(lmp)
 {
   respa_enable = 1;
   writedata = 1;
-  centroidstressflag = 1;
+  centroidstressflag = CENTROID_SAME;
 }
 
 /* ---------------------------------------------------------------------- */
--- a/src/compute_centroid_stress_atom.cpp
+++ b/src/compute_centroid_stress_atom.cpp
@@ -126,7 +126,7 @@ void ComputeCentroidStressAtom::init()
 
   // check if pair styles support centroid atom stress
   if (pairflag && force->pair)
-    if (force->pair->centroidstressflag == CENTROID_NOTAVAIL)
+    if (force->pair->centroidstressflag & CENTROID_NOTAVAIL)
       error->all(FLERR, "Pair style does not support compute centroid/stress/atom");
 }
 
@@ -178,7 +178,7 @@ void ComputeCentroidStressAtom::compute_peratom()
   // per-atom virial and per-atom centroid virial are the same for two-body
   // many-body pair styles not yet implemented
   if (pairflag && force->pair && force->pair->compute_flag) {
-    if (force->pair->centroidstressflag == CENTROID_AVAIL) {
+    if (force->pair->centroidstressflag & CENTROID_AVAIL) {
       double **cvatom = force->pair->cvatom;
       for (i = 0; i < npair; i++)
         for (j = 0; j < 9; j++)
--- a/src/compute_centroid_stress_atom.cpp
+++ b/src/compute_centroid_stress_atom.cpp
@@ -257,7 +257,7 @@ void ComputeCentroidStressAtom::compute_peratom()
 
   if (kspaceflag && force->kspace && force->kspace->compute_flag) {
     double **vatom = force->kspace->vatom;
-    for (i = 0; i < nkspace; i++)
+    for (i = 0; i < nkspace; i++) {
       for (j = 0; j < 6; j++)
         stress[i][j] += vatom[i][j];
       for (j = 6; j < 9; j++)
--- a/src/compute_centroid_stress_atom.cpp
+++ b/src/compute_centroid_stress_atom.cpp
@@ -221,8 +221,8 @@ void ComputeCentroidStressAtom::compute_peratom()
 
   // per-atom virial and per-atom centroid virial are the same for bonds
   // bond styles are all CENTROID_SAME
-  // all other styles are CENTROID_AVAIL or CENTROID_NOTAVAIL
-  // add KSpace when it becomes supported
+  // angle, dihedral, improper styles are CENTROID_AVAIL or CENTROID_NOTAVAIL
+  // KSpace styles are all CENTROID_NOTAVAIL, placeholder CENTROID_SAME below
   
   if (bondflag && force->bond) {
     double **vatom = force->bond->vatom;
@@ -255,12 +255,15 @@ void ComputeCentroidStressAtom::compute_peratom()
         stress[i][j] += cvatom[i][j];
   }
 
-  //if (kspaceflag && force->kspace && force->kspace->compute_flag) {
-  //  double **vatom = force->kspace->vatom;
-  //  for (i = 0; i < nkspace; i++)
-  //    for (j = 0; j < 9; j++)
-  //      stress[i][j] += cvatom[i][j];
-  // }
+  if (kspaceflag && force->kspace && force->kspace->compute_flag) {
+    double **vatom = force->kspace->vatom;
+    for (i = 0; i < nkspace; i++)
+      for (j = 0; j < 6; j++)
+        stress[i][j] += vatom[i][j];
+      for (j = 6; j < 9; j++)
+        stress[i][j] += vatom[i][j-3];
+    }
+  }
 
   // add in per-atom contributions from relevant fixes
   // skip if vatom = nullptr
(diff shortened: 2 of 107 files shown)


src/pair.cpp
```diff
@@ -786,14 +786,14 @@ void Pair::ev_setup(int eflag, int vflag, int alloc)
   evflag = 1;
 
   eflag_either = eflag;
-  eflag_global = eflag % 2;
-  eflag_atom = eflag / 2;
+  eflag_global = eflag & ENERGY_GLOBAL;
+  eflag_atom = eflag & ENERGY_PER_ATOM;
```
yafshar, line 790: @rbberger This is creating the wrong value. Similar to what happened at #2469

```cpp
eflag=0 -> 0
eflag=1 -> 0
eflag=2 -> 2
eflag=3 -> 2
```
It should be something like `bool(i & ENERGY_PER_ATOM)`

src/pair.cpp
```diff
@@ -786,14 +786,14 @@ void Pair::ev_setup(int eflag, int vflag, int alloc)
...
 
-  vflag_global = vflag % 4;
-  vflag_atom = vflag & 4;
+  vflag_global = vflag & (VIRIAL_GLOBAL_PW_SUM | VIRIAL_GLOBAL_PW_FDOTR_W_GHOSTS);
+  vflag_atom = vflag & VIRIAL_PER_ATOM;
```
yafshar, line 793: @rbberger the same thing here

src/pair.cpp
akohlmey: > @rbberger This is creating the wrong value. Similar to what happened at #2469

the value doesn't matter. all the tests are for zero vs. non-zero.

[...]

> It should be something like `bool(i & ENERGY_PER_ATOM)`

if you absolutely want 0 vs 1 it should be more like `(eflag & ENERGY_PER_ATOM) ? 1 : 0`
those flags are "int" not "bool".

src/pair.h
```diff
@@ -18,6 +18,20 @@
 
 namespace LAMMPS_NS {
 
+enum {
+  ENERGY_NO_COMPUTATION  = 0x00,
```
akohlmey, line 22: some of those constants are rather long names. not sure that helps with readability.
how about this?
- ENERGY_NONE
- ENERGY_GLOBAL
- ENERGY_PERATOM
- VIRIAL_TALLY
- VIRIAL_FDOTR
- VIRIAL_PERATOM
- VIRIAL_CENTROID

src/pair.cpp
yafshar: @akohlmey I fixed Kim to avoid the requirement for value. Just mentioned the value for the backward compatibility and if there is any other piece of code that checks the value otherwise you are right, the value is not important

src/pair.h
rbberger: @akohlmey sure, that seems ok

sjplimp:
@rbberger yes to the shortened names.  Also these bitmasks are needed more globally than Pair.
Integrate::ev_set uses them to set eflag,vflag.  Then those vars are passed to pair, bond, angle, etc.
So all of them need the masks to decode.  So how about the masks are defined in force.h ?

rbberger:
@sjplimp done. now what is left is updating all other code paths where those flags are used, checked and set. I'll look at this next week.

src/kspace.cpp
```diff
@@ -224,12 +232,12 @@ void KSpace::ev_setup(int eflag, int vflag, int alloc)
...
-  eflag_global = eflag % 2;
-  eflag_atom = eflag / 2;
+  eflag_global = eflag & ENERGY_GLOBAL;
+  eflag_atom = eflag & ENERGY_ATOM;
 
   vflag_either = vflag;
-  vflag_global = vflag % 4;
-  vflag_atom = vflag / 4;
+  vflag_global = vflag & (VIRIAL_PAIR | VIRIAL_FDOTR);
+  vflag_atom = vflag / 4; // TODO
```
rbberger, line 240: @sjplimp if `vflag = VIRIAL_CENTROID` (=8), this will be `2` and therefore true. Not sure if that is intentional or an outdated code path.

akohlmey: This is intentional. It would have been cleaner to have written `vflag_atom = vflag & 12;`
The centroid description of the per-atom stress is equivalent to the virial description for all interactions with spherical symmetry (pair styles like lj/cut or morse). Thus all such pair styles are automatically compatible and that also includes KSpace. Only bonded interactions and many-body potentials need a different derivation of the per-atom stress (for global stress the result is the same again for both representations).

rbberger: @akohlmey thanks for the clarification.

rbberger:
@sjplimp all yours.

src/CLASS2/pair_lj_class2.cpp
```diff
@@ -35,7 +35,7 @@ PairLJClass2::PairLJClass2(LAMMPS *lmp) : Pair(lmp)
 {
   respa_enable = 1;
   writedata = 1;
-  centroidstressflag = 1;
+  centroidstressflag = CENTROID_SAME;
```
sjplimp, line 38: this is the default - the point of changing the default was to remove this lines from all the true pairwise pair styles

sjplimp:
@rbberger back to you.  This will require adding the CENTROID enum to compile. @donatas-surblys I think I added all the restrictions we talked about.  Can you take a look at the bottom of the doc/src/compute_stress_atom.rst doc page to see if it all looks good?  Ditto for src/compute_centroid_stress_atom.cpp

rbberger:
@sjplimp the CENTROID enum was already added in `force.h`. If you want to change the values from 1, 2, 4, you can do it there. The code doesn't compile for various other reasons. Let me know when you are done pushing and change the assignment to me.

src/compute_centroid_stress_atom.cpp
```diff
@@ -228,20 +256,18 @@ void ComputeCentroidStressAtom::compute_peratom()
 
   if (kspaceflag && force->kspace && force->kspace->compute_flag) {
     double **vatom = force->kspace->vatom;
-    for (i = 0; i < nkspace; i++) {
-      for (j = 0; j < 6; j++)
-        stress[i][j] += vatom[i][j];
-      for (j = 6; j < 9; j++)
-        stress[i][j] += vatom[i][j-3];
-    }
+    for (i = 0; i < nkspace; i++)
```
rbberger, line 259: @sjplimp since `kspace->cvatom` doesn't exist, it's unclear what your intentions are with this change.

sjplimp:
@rbberger ok, builds now.  See if you agree with all the logic?

rbberger:
@sjplimp @akohlmey Only found a minor issue. Otherwise, looks good to me. Nothing showed up in the regression testing. Since this a critical change, I'm asking for at least 2 core dev approvals before merging (myself excluded since I created the PR).

src/compute_centroid_stress_atom.cpp
```diff
@@ -226,22 +255,20 @@ void ComputeCentroidStressAtom::compute_peratom()
...
-      for (j = 6; j < 9; j++)
-        stress[i][j] += vatom[i][j-3];
-    }
-  }
+  //if (kspaceflag && force->kspace && force->kspace->compute_flag) {
+  //  double **vatom = force->kspace->vatom;
+  //  for (i = 0; i < nkspace; i++)
+  //    for (j = 0; j < 9; j++)
+  //      stress[i][j] += cvatom[i][j];
+  // }
```
donatas-surblys, line 263: There will probably never be a centroid stress definition for kspace, and even if there was it would probably by a symmetric tensor, so the (currently commented out) code is a bit  misleading. Also, from a pure convenience point of view, if someone wanted to force the use of the original kspace stress definition for centroid/stress/atom (which is OK for total system heat flux), he would need to not only change the centroidstressflag, but also modify this part. My personal preference would be to revert the code or at least leave the original code that uses vatom commented out.

donatas-surblys:
I've looked over the documentation and the code in centroid_stress_atom.cpp. Except for the one comment, everything looks OK.

sjplimp:
@rbberger I will address @donatas-surblys comment in the next few minutes

akohlmey:
> @rbberger I will address @donatas-surblys comment in the next few minutes

@sjplimp this fails to compile now.

rbberger:
@sjplimp I can't merge this without approval. I need either yours or Axel to do it.

sjplimp (approved):

PR #2485, opened 2020-11-16
Kokkos ReaxFF tweaks

stanmoore1:
**Summary**

- Removed unused variable
- Change the default of the `package kokkos neigh/qeq` option to no longer match the `neigh` option if unset, but rather have its own default of `full` on the GPU and `half` on the CPU, which typically gives the best performance.

**Related Issue(s)**

None.

**Author(s)**

Stan Moore (SNL)

**Licensing**

**Backward Compatibility**

Yes


stanmoore1:
@arghdos

rbberger (approved):

PR #2488, opened 2020-11-20
Update KIM.cmake for kim-api-2.2.0

ellio167:
**Summary**

Update to kim-api v2.2.0

**Related Issue(s)**

**Author(s)**
Ryan S. Elliott (UMinn)

**Licensing**

**Backward Compatibility**

None


junghans (approved):

akohlmey (changes requested):
This is insufficient. It is missing the corresponding updates for the conventional build system.

ellio167:
Right!  Thanks for the reminder.

cmake/Modules/Packages/KIM.cmake
```diff
@@ -19,13 +19,10 @@ if(CURL_FOUND)
     target_compile_definitions(lammps PRIVATE -DLMP_NO_SSL_CHECK)
   endif()
 endif()
-find_package(PkgConfig QUIET)
+find_package(KIM-API ${KIM-API_MIN_VERSION} QUIET CONFIG)
 set(DOWNLOAD_KIM_DEFAULT ON)
-if(PKG_CONFIG_FOUND)
-  pkg_check_modules(KIM-API QUIET libkim-api>=${KIM-API_MIN_VERSION})
-  if(KIM-API_FOUND)
+if(KIM-API_FOUND)
```
rbberger, line 24: This fails with an older preinstalled KIM binary package (e.g. [2.1.3 on Ubuntu](https://ci.lammps.org/job/dev/job/pull_requests/job/ubuntu/job/unit_tests/job/cmake_mpi_openmp_smallbig_shared/1370/consoleFull)).
Probably should be
```cmake
if(KIM-API_FOUND AND KIM_API_VERSION VERSION_GREATER_EQUAL KIM-API_MIN_VERSION)
```
if you want to require the newer version.

If I do so, it will download the newer version and attempt to compile it. However, then something gets really messed up. I get all kinds of C++11 errors from unrelated LAMMPS code. Sounds like something resets the compiler standard to C++98. I guess this is caused by:
```
#  kim-api configuration:
#
#    === C++ compile values =================
#    CMAKE_CXX_COMPILER: (GNU, 7.5.0, std=98)
#       /usr/bin/c++
```

rbberger (changes requested):

cmake/Modules/Packages/KIM.cmake
ellio167: OK, I made changes, but the unit_test failed immediately.  Looks like some sort of non-related error.  Is there a way to retrigger the tests?

rbberger: @ellio167 The master got into a bad state because some unit tests weren't failing when they should have. I already have a fix, but I'm waiting on @sjplimp to accept the revert (see #2491). I'll retrigger once the master is back in a good state.

rbberger: @ellio167 I retriggered it, but the behaviour is still different. The problem is that `kim-api-config.cmake` sets the global cmake variables:
```
  set(CMAKE_CXX_STANDARD 98)
  set(CMAKE_CXX_STANDARD_REQUIRED YES)
  set(CMAKE_CXX_EXTENSIONS )
  #
  set(CMAKE_C_STANDARD 90)
  set(CMAKE_C_STANDARD_REQUIRED YES)
  set(CMAKE_C_EXTENSIONS OFF
```
when running `find_package(KIM-API ${KIM-API_MIN_VERSION} QUIET CONFIG)`

ellio167: I can't seem to get to the logs to see the details of what is happening.  How do I find those?

rbberger (approved):

rbberger:
@akohlmey since you asked for changes, can you approve now?

akohlmey (approved):

Issue #1521, opened 2019-06-18
[Feature Request] Clarify the relation between USER-TALLY computes and dynamic groups


donatas-surblys:
**Summary**

Although currently all of USER-TALLY computes accept dynamic groups without complaining, this often produces results that are most likely unintended by the user. I am not sure if this can be classified as a bug, therefore I would like this issue to be at least clarified in the documentation.

**Detailed Description**

Dynamic groups are setup during the post_integrate stage. On the other hand, USER-TALLY computes are implemented as callbacks during the pairwise force computations. Because of this, during initial force computation at step 0 dynamic groups are empty, which is unexpected unless one is familiar with the actual implementation details. This can be solved by executing "run 0" before the main computation to force dynamic group construction, therefore this in itself is not a large issue.

Unfortunately, I've found that when executing USER-TALLY computes with dynamic groups without prepending "run 0", which the user might do if he doesn't care about the values at step 0, also produces incorrect values in the subsequent computations. For example, using pe/tally with dynamic groups in the following script:
```
units real

pair_style lj/cut 12

region box block -15 15 -15 15 -15 15
create_box 1 box
pair_coeff * * 1 3
mass * 100
create_atoms 1 random 10 101 box
velocity all create 100 101

region inf block INF INF INF INF INF INF
group dall dynamic all region inf every 1

fix integ all nve
#run 0

compute   pe  all pe
compute stpe  all pe/tally  all
compute dtpe dall pe/tally dall

thermo 1
thermo_style custom step c_pe c_stpe c_dtpe
thermo_modify format float %.17e

run 5
```

Produces these results:

> Step c_pe c_stpe c_dtpe 
>        0 -4.22768097887963948e-02 -4.22768097887964017e-02 0.00000000000000000e+00 
>        1 -4.22735414333879969e-02 -4.22735414333880247e-02 -1.50152175689973015e-02 
>        2 -4.22702580557454272e-02 -4.22702580557454272e-02 -1.50056628371679785e-02 
>        3 -4.22669596321978633e-02 -4.22669596321978494e-02 -1.49961151144276019e-02 
>        4 -4.22636461391601853e-02 -4.22636461391601784e-02 -1.49865743947735428e-02 
>        5 -4.22603175531332254e-02 -4.22603175531332323e-02 -1.49770406722097276e-02 

[...]

PR #2501, opened 2020-12-01
Set ghost atom groups for simple dynamic regions to enable partial use with tally computes

donatas-surblys:
**Summary**

Enables dynamic groups with only region constraints to set the groups of their ghost atoms via a new "ghost yes" option. This enables the USER-TALLY computes to work correctly with such dynamic groups. There is also an "init yes" option added to force the setting of groups on definition. This PR is basically a proof of concept as I want to know if it would be possible to implement such functions in LAMMPS and if I'm missing any edge cases.

**Related Issue(s)**

A short discussion in #1521.

**Author(s)**

Donatas Surblys (Tohoku University) surblys.donatas at gmail dot com

**Licensing**

**Backward Compatibility**

Full backward compatibility

**Implementation Notes**

As very briefly discussed in #1521, I'd really like to have the USER-TALLY computes work with dynamic groups.  I have an in-house solution, where a custom fix (not in this PR) sets the required dynamic groups in the pre_force phase and does the necessary forward communication, but for dynamic groups with only region constraints no inter-process communication should be needed, as the constraints are known globally. This PR simply enables the setting of atom masks on ghost atoms when the "ghost yes" option is set. The effect of this can be seen with the sample script:
```
units real
pair_style lj/cut 12
region box block -15 15 -15 15 -15 15
create_box 1 box
pair_coeff * * 1 3
mass * 100
create_atoms 1 random 10 101 box
velocity all create 100 101
region inf block INF INF INF INF INF INF

group dall dynamic all region inf every 1
group gall dynamic all region inf every 1 ghost yes init yes

fix integ all nve

compute   pe  all pe
compute stpe  all pe/tally  all
compute dtpe dall pe/tally dall
compute gtpe gall pe/tally gall

thermo 1
thermo_style custom step c_pe c_stpe c_dtpe c_gtpe
thermo_modify format float %.17e

run 2
```
NOTE: `dynamic_group_allow = 1` needs to be set in compute_pe_tally.cpp (not in PR).

This results in:
> Step c_pe c_stpe c_dtpe c_gtpe 
>        0 -7.47091469703964117e-01 -7.47091469703963784e-01 0.00000000000000000e+00 -7.47091469703963784e-01 
>        1 -7.47015326232009835e-01 -7.47015326232009946e-01 -7.31816520122599345e-01 -7.47015326232009946e-01 
>        2 -7.46943164797733816e-01 -7.46943164797734038e-01 -7.31733386758632198e-01 -7.46943164797734038e-01

Notice, how dynamic groups with "ghost yes" give correct potential energies.

**Post Submission Checklist**

### DIFF
--- a/src/fix_group.cpp
+++ b/src/fix_group.cpp
@@ -52,6 +52,8 @@ idregion(nullptr), idvar(nullptr), idprop(nullptr)
   varflag = 0;
   propflag = 0;
   nevery = 1;
+  ghostflag = 0;
+  initflag = 0;
 
   int iarg = 3;
   while (iarg < narg) {
@@ -90,8 +92,23 @@ idregion(nullptr), idvar(nullptr), idprop(nullptr)
       nevery = utils::inumeric(FLERR,arg[iarg+1],false,lmp);
       if (nevery <= 0) error->all(FLERR,"Illegal group command");
       iarg += 2;
+    } else if (strcmp(arg[iarg],"ghost") == 0) {
+      if (iarg+2 > narg) error->all(FLERR,"Illegal group command");
+      if (strcmp(arg[iarg+1],"no") == 0) ghostflag = 0;
+      else if (strcmp(arg[iarg+1],"yes") == 0) ghostflag = 1;
+      else error->all(FLERR,"Illegal group command");
+      iarg += 2;
+    } else if (strcmp(arg[iarg],"init") == 0) {
+      if (iarg+2 > narg) error->all(FLERR,"Illegal group command");
+      if (strcmp(arg[iarg+1],"no") == 0) initflag = 0;
+      else if (strcmp(arg[iarg+1],"yes") == 0) initflag = 1;
+      else error->all(FLERR,"Illegal group command");
+      iarg += 2;
     } else error->all(FLERR,"Illegal group command");
   }
+  if (ghostflag && (!regionflag || varflag || propflag))
+    error->all(FLERR,
+        "Only simple dynamic region groups can be used with 'ghost yes'");
 }
 
 /* ---------------------------------------------------------------------- */
@@ -171,6 +188,8 @@ void FixGroup::init()
   if (warn && comm->me == 0)
     error->warning(FLERR,"One or more dynamic groups may not be "
                    "updated at correct point in timestep");
+
+  if (initflag) set_group();
 }
 
 /* ----------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


akohlmey:
@donatas-surblys step 1 is that @sjplimp is ok with including this change. step 2 would be to add documentation and then step3 to enable this feature in the USER-TALLY computes. Those would also have to check, though if a dynamic group is suitable and that needs to be documented as well.

donatas-surblys:
Provided step 1 is cleared. I can take care of steps 2 and 3. Regarding step 3, since this is probably only going to be used by USER-TALLY computes, I'd opt to simply make the relevant flags in the dynamic group fix public and manually check them in each USER-TALLY compute constructor, erroring out when necessary.

akohlmey:
@sjplimp we are waiting on you to tell us what you think about this change.

donatas-surblys:
@sjplimp It has been a while, but I would still like to get this into LAMMPS. The tally computes would really be much more useful with even partial dynamic group support.
Originally, I didn't want to make too many changes in case it would be rejected, but would doing more preliminary work speed up the review process? Should I rebase it to the newest version?

donatas-surblys:
Sorry for the late reply. As @sjplimp mentioned, in the updated code the groups are updated in the post-force phase, but tally computes are called during the force computations, so the group definition lags by one step. This can be demonstrated by trying to compute potential energy between left and right sides and looking what happens when an atom crosses the boundary:
```
units real
atom_modify map yes
pair_style lj/cut 12
region box block -15 15 -15 15 -15 15
create_box 1 box
pair_coeff * * 1 3.1
mass * 100

create_atoms 1 single -4 0 0
create_atoms 1 single  3 0 0
region left  block INF 0 INF INF INF INF
region right block 0 INF INF INF INF INF
group left  dynamic all region left  every 1
group right dynamic all region right every 1

compute dpe left pe/tally right
variable x2 equal x[2]
group a2 id 2
fix move a2 move linear -2 0 0

thermo 1
thermo_style custom step v_x2 pe c_dpe

run 0
run 3
```
This results in:

>    Step          v_x2          PotEng         c_dpe     
>          0   3             -0.029947      -0.029947     
>          1   1             -0.21429588    -0.21429588   
>          2  -1              1.0588013      1.0588013    
>          3  -3              3147101.1      0       

As we can see, ideally, c_dpe at step 2 should have been 0. In many cases, this inconsistency is probably going to be drowned out by statistical error anyway, so I don’t know if it warrants disabling dynamic groups for tally computes.

PR #2503, opened 2020-12-01
Add warning about missing libcurl at configure time

ellio167:
**Summary**

Add configure-time warning when libcurl is not found

**Author(s)**

Ryan S. Elliott (UMinn)

**Licensing**

**Backward Compatibility**

none

**Implementation Notes**

Updated both cmake and traditional build

**Post Submission Checklist**


cmake/Modules/Packages/KIM.cmake
```diff
@@ -1,4 +1,4 @@
-set(KIM-API_MIN_VERSION 2.1.3)
+set(KIM-API_CUR_VERSION 2.1.3)
```
akohlmey, line 1: @ellio167 can you give a reason for the name change here? the only one seemed more appropriate.
I am thinking to add lists of checksums and versions like we have for the traditional build system, e.g. `KIM-API_SUPPORTED_VERSIONS`) and then have a `KIM-API_VERSION` variable that selects which version is used. it could default to what is found locally, if supported, and otherwise select downloading the latest listed version.
that would allow a bit more control to make certain that LAMMPS is changed when there are API version changes
and correspondingly the list is trimmed, when backward compatibility to older API versions is removed.

PR #2508, opened 2020-12-04
Add support for Python based models to MLIAP package

athomps:
**Summary**

Added a new mliap model style *mliappy *that provides a general interface to Python-based machine-learning models, including PyTorch.

**Related Issue(s)**

n/a

**Author(s)**

Nicholas Lubbers, LANL
Aidan Thompson, SNL

**Licensing**

**Backward Compatibility**

The addition of the *mliappy* model style introduces a dependency on Python that did not exist before.  

**Implementation Notes**

Getting this to work required @lubbersnick adding some code to the LAMMPS Python interface (python/lammps.py,  src/PYTHON/python_impl.cpp, cmake/Modules/Packages/PYTHON.cmake)

**Post Submission Checklist**


athomps:
Building LAMMPS with MLIAP package requires first running these commands:

```
cd src/MLIAP
cythonize mliap_model_python_couple.pyx
```

athomps:
This PR needs work that is beyond my expertise in order to pass the build tests.  Specifically, there needs to be a way to either:
a) run the cythonize command mentioned above before building
or b) make the -DLMP_MLIAPPY flag optional , so that default setting eliminates Python dependence e.g. CMakeLists.txt:if(WITH_MLIAPPY).  Any suggestions appreciated.

akohlmey:
> This PR needs work that is beyond my expertise in order to pass the build tests. Specifically, there needs to be a way to either:
> a) run the cythonize command mentioned above before building
> or b) make the -DLMP_MLIAPPY flag optional , so that default setting eliminates Python dependence e.g. CMakeLists.txt:if(WITH_MLIAPPY). Any suggestions appreciated.

i can give it a try.

athomps:
@akohlmey Thanks!  If you have questions about the Python aspects, @lubbersnick can explain.

cmake/Modules/Packages/MLIAP.cmake
```diff
@@ -0,0 +1,3 @@
+execute_process(COMMAND cythonize mliap_model_python_couple.pyx WORKING_DIRECTORY ../src/MLIAP)
```
rbberger, line 1: We should look for a way that does not produce files in the source directory `../src/MLIAP`

akohlmey: already on it.

akohlmey: just pushed what I currently have to see how it fares with integration testing after it is (mostly) working on my local machine. still need to update documentation.

athomps:
Thanks @akohlmey and @rbberger.  This looks very nice. I noticed with cmake build, if MLIAP_ENABLE_PYTHON on and PKG_PYTHON off, no error is detected during configuration, only when
 compiling mliap_model_python_couple.cpp.

akohlmey:
> Thanks @akohlmey and @rbberger. This looks very nice. I noticed with cmake build, if MLIAP_ENABLE_PYTHON on and PKG_PYTHON off, no error is detected during configuration, only when
> compiling mliap_model_python_couple.cpp.

you probably have a leftover copy in src/MLIAP/ that should be deleted. it will be created on the fly in either src or build/cython during compilation/configuration.

akohlmey:
@athomps please note this was just pushed as far as needed to get the basics working.careful testing is still pending and will happen tomorrow. also, I have not done any checking on the python part of the code. that is something better done by @rbberger. I hope that @junghans can have a look at the CMake code for obvious mistakes and @sjplimp can check if there are any obvious oversights on the conventional build side.

athomps:
@akohlmey Understood. The build instructions in src/MLIAP/README, examples/mliap/README, and doc/src/Packages_details.rst also need to be updated, but I will wait until everything else is resolved.

akohlmey:
> @akohlmey Understood. The build instructions in src/MLIAP/README, examples/mliap/README, and doc/src/Packages_details.rst also need to be updated, but I will wait until everything else is resolved.

Package_details.rst has been updated as needed. The build instructions themselves have been added to Build_extras.rst where they also are for other packages. I didn't update the README files. Feel free to do so. I won't touch it until tomorrow morning.

athomps:
Building LAMMPS with MLIAP package and with/without PYTHON package works as expected.  The cmake dependency checking for `(MLIAP_ENABLE_PYTHON=on && PKG_PYTHON=off)` is not working for me.  It is working for `(PKG_MLIAP=on && PKG_SNAP=off)`

athomps:
> I didn't update the README files. Feel free to do so. I won't touch it until tomorrow morning.

I'll wait until everything is working.

akohlmey:
@rbberger I will tried my best to make it compile and auto adapt to include python support when possible by default.

> Building LAMMPS with `make yes-mliap` and `make yes/no-python` works as expected. The cmake dependency checking for `(MLIAP_ENABLE_PYTHON=on && PKG_PYTHON=off)` is not working for me. It is working for `(PKG_MLIAP=on && PKG_SNAP=off)`

should be fixed now

athomps:
> should be fixed now

Yes, cmake detects the dependence now.

cmake/Modules/Packages/PYTHON.cmake
```diff
@@ -4,6 +4,7 @@ if(CMAKE_VERSION VERSION_LESS 3.12)
   target_link_libraries(lammps PRIVATE ${PYTHON_LIBRARIES})
 else()
   find_package(Python REQUIRED COMPONENTS Development)
+  target_include_directories(lammps PRIVATE ${Python_INCLUDE_DIRS})
```
junghans, line 7: Why is this needed? Should we included in the target below.

cmake/Modules/Packages/MLIAP.cmake
```diff
@@ -0,0 +1,29 @@
...
+    set(MLIAP_ENABLE_PYTHON_DEFAULT ON)
+  endif()
+endif()
+
+option(MLIAP_ENABLE_PYTHON "Build MLIAP package with Python support" ${MLIAP_ENABLE_PYTHON_DEFAULT})
+
+if(MLIAP_ENABLE_PYTHON)
+  find_package(Cythonize REQUIRED)
+  if(NOT PKG_PYTHON)
+    message(FATAL_ERROR "Must install PYTHON package for MLIAP with Python support")
```
junghans, line 15: The message reads a bit strange.

cmake/Modules/Packages/PYTHON.cmake
athomps: I think that might be no longer needed. @lubbersnick can you confirm?

cmake/Modules/Packages/MLIAP.cmake
```diff
@@ -0,0 +1,29 @@
...
+  endif()
+
+  set(MLIAP_CYTHON_DIR ${CMAKE_BINARY_DIR}/cython)
+  file(MAKE_DIRECTORY ${MLIAP_CYTHON_DIR})
+  add_custom_command(OUTPUT  ${MLIAP_CYTHON_DIR}/mliap_model_python_couple.cpp ${MLIAP_CYTHON_DIR}/mliap_model_python_couple.h
+            COMMAND            ${CMAKE_COMMAND} -E copy ${LAMMPS_SOURCE_DIR}/MLIAP/mliap_model_python_couple.pyx ${MLIAP_CYTHON_DIR}/mliap_model_python_couple.pyx
```
junghans, line 21: Do you really have to copy it? Just refer to it using the full path.

cmake/Modules/Packages/MLIAP.cmake
athomps: It does the job. I will review all the install instructions at the end and try to harmonize them.

lubbersnick: Thought: we might change the run-time behavior to put a message in like this right now. As-is if MLIAP+python support is not enabled, it will just give "Invalid Pair Style".

cmake/Modules/Packages/PYTHON.cmake
lubbersnick: My build works without it. It was probably introduced accidentally.

cmake/Modules/Packages/MLIAP.cmake
akohlmey: Yes, this is required. `cythonize` will create the source code at the location of the the .pyx file and we don't want created files mixed with the source.

cmake/Modules/Packages/MLIAP.cmake
junghans: I would maybe say "Python package must be enabled for MLIAP with Python support"

cmake/Modules/Packages/MLIAP.cmake
junghans: Thanks,

cmake/Modules/FindCythonize.cmake
```diff
@@ -0,0 +1,30 @@
...
+#
+# This code sets the following variables:
+#
+#  Cythonize_EXECUTABLE
+#
+# adapted from https://github.com/cmarshall108/cython-cmake-example/blob/master/cmake/FindCython.cmake
+#=============================================================================
+
+if(CMAKE_VERSION VERSION_LESS 3.12)
+    find_package(PythonInterp 3.6 QUIET) # Deprecated since version 3.12
```
junghans, line 11: We are implicitly requiring python-3.6 here, is that wanted?

akohlmey: yes

lubbersnick: Probably we want python >=3.5, although for certain aspects >=3.6 might be preferable. It is definitely not constrained to exactly 3.6. I have mostly been testing on 3.7. Technically cython should work on 3.3+

cmake/Modules/Packages/MLIAP.cmake
junghans: we could use `copy_if_different` instead of `copy` here.

cmake/Modules/Packages/MLIAP.cmake
```diff
@@ -0,0 +1,29 @@
...
+            COMMAND            ${Cythonize_EXECUTABLE} ${MLIAP_CYTHON_DIR}/mliap_model_python_couple.pyx -3
```
junghans, line 22: For readability I would move the `-3` to the front.

cmake/Modules/FindCythonize.cmake
akohlmey: Somewhere in the comments or docs it said Python > 3.5 was required. So that is what I implemented.
For your reference: Our lower boundary for target distributions we want to support is CentOS 7, which comes with Python 3.6. Ubuntu 16.04 has Python 3.5, but we are not testing on it (anymore) and we don't support CMake on it, as the CMake version is too old. So version Python 3.6 is the *de facto* lower boundary for Python 3 (and 2.7 the lower and upper boundary on Python 2)

akohlmey: > Probably we want python >=3.5, although for certain aspects >=3.6 might be preferable. It is definitely not constrained to exactly 3.6. I have mostly been testing on 3.7. Technically cython should work on 3.3+

FYI, the version number specified with `find_package()` is the lower boundary. So we can require Python 3.6 when using CMake to configure (which has its lower bound at version 3.10) and not apply any additional restrictions to what we already have.

cmake/Modules/Packages/MLIAP.cmake
lubbersnick: The -3 shouldn't need to be there, the pyx file has a language_level declaration at the top.

akohlmey: > The -3 shouldn't need to be there, the pyx file has a language_level declaration at the top.

i removed most of those. many didn't seem to have an effect, anyway. i consider it a bad idea to have command line flags embedded in source code unless it cannot be avoided. the flag was added to make things more explicit.

lubbersnick: I had the others around for building this as a standalone module, which is a separate endeavor. I don't have a problem with them removed for lammps:master.

But the cython language_level and distutils : c++ flags are pretty key because they actually give information about how the source file should be read by humans as well, I think it is good to leave them there.

cmake/Modules/FindCythonize.cmake
lubbersnick: Okay, makes sense -- I tried digging through the cmake docs but couldn't figure out if it was an exact or a lower bound.

cmake/Modules/Packages/MLIAP.cmake
akohlmey: ...and those two settings had been retained in the source.

akohlmey:
docs, readme files, and build scripts have been updated to (hopefully) satisfy all concerns raised in the discussion. handing the PR over to @rbberger to review and test the Python component and pytorch intergration.

junghans:
With this package we will have three different ways to interact with python:

- through `ctypes` for `PKG_PYTHON`
- through `cythonize` for `PKG_MLIAP`
- through `swig` for the generated interface

Are they all compatible with each other?

cmake/Modules/FindCythonize.cmake
junghans: Actually we only check if python is >=3.6 not cythonize ifself. CMake could still pick up an earlier version as `_python_path` is only used a hints. Is there a required cython version? I only test this PR with cython 0.29.21

junghans: We could use https://github.com/espressomd/espresso/blob/python/cmake/FindCython.cmake#L25-L30 to check the version.

athomps (commented):
No comment

python/lammps.py
```diff
@@ -506,6 +506,8 @@ def __init__(self,name='',cmdargs=None,ptr=None,comm=None):
     self.lib.lammps_set_fix_external_callback.argtypes = [c_void_p, c_char_p, self.FIX_EXTERNAL_CALLBACK_FUNC, py_object]
     self.lib.lammps_set_fix_external_callback.restype = None
 
+    self.mliappy = MLIAPPY(self)
```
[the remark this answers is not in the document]
athomps: @nicklubbers says he posted a response to above comments earlier today, but it is awaiting moderator approval.  Can he be added to a  pre-approved list?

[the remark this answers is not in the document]
akohlmey: > @lubbersnick says he posted a response to above comments earlier today, but it is awaiting moderator approval. Can he be added to a pre-approved list?

I am not aware of having any moderation feature enabled for lammps on GitHub.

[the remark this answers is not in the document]
lubbersnick: Fixed now. It didn't say anything about "moderators", just listed "pending" next to the comments. I guess it had something to do with my leaving a comment on a part of the code nobody had discussed yet. Sorry about that.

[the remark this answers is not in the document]
akohlmey: > Fixed now. It didn't say anything about "moderators", just listed "pending" next to the comments. I guess it had something to do with my leaving a comment on a part of the code nobody had discussed yet. Sorry about that.

The "pending" status happens when you click on "Start a review" while the behavior you expected would happen with clicking on the "Add single comment" button.

examples/mliap/README
```diff
@@ -0,0 +1,74 @@
...
+by in.mliap.snap.Ta06A.
+
+2: Running a Python script: mliap_pytorch_Ta06A.py
+
+Before testing this, ensure that the first example 
+(using LAMMPS executable) works.
+Also, not all python installations support this mode of operation.
+It requires that the Python interpreter be initialized.
+To check this for your Python library,
+try running the Py_IsInitialized() method.
```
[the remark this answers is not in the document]
athomps: It was called test_pylibs.py.  I removed it because it is outside the mission scope of the examples folder, which is to test whether LAMMPS is running correctly, as opposed to checking the configuration of the user's runtime environment.   LAMMPS should not get into the business of testing peoples computers for them, except in the context of cmake.  Feel free to augment the language in the README to make it easier to understand.

python/lammps.py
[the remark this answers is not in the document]
rbberger: @lubbersnick I agree that `lammps` should be a package. I had that on my agenda for some time now, but it was always pushed back for some reason (e.g., stable release). I'll use this as an excuse to start that process. I'll create a new PR for this change.

examples/mliap/README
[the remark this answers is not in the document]
lubbersnick: Ok, we should put this back into the MLIAPPY object.

python/lammps.py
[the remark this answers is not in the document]
lubbersnick: I am happy to try and help refactoring the file if there is a clear method for me to test my changes, feel free to involve me about that. The part where I'm still pretty new is on the build/configuration side.

For this PR are we ok to add `class MLIAPPY()` to the end of lammps.py? Would you prefer if we moved mliappy_pytorch.py to a new directory /python/mliappy/ for the moment, and update the example so instead of copying, it add this to the sys.path?

examples/mliap/README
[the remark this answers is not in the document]
athomps: Yes, I agree.

python/lammps.py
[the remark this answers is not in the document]
rbberger: @lubbersnick I started with PR #2521. Let me know what you think. For this PR I'm ok having it in lammps.py for now. I only would ask to only create a MLIAPPY instance when accessed, similar to what you see with the optional numpy wrapper. Having that folder and adding it to the sys path sound good for now. It would probably end up in a subpackage, like `python/lammps/mliappy/`

rbberger:
@athomps @lubbersnick I've resolved the merge conflicts after PR #2521 got merged. You can now proceed to making the adjustments we talked about.

rbberger (changes requested):
The change to `setuptools` break out-of-source compilation/installation with CMake. I creates a folder `python/lammps.egg-info` in the source checkout. That can cause problems with multiple parallel builds using the same source folder.

athomps:
@lubbersnick has addressed the Python interface issues. I have updated the example and src README files and verified that all the examples still work.  I did not provide reference output for in.mliap.pytorch.relu1hidden, as the model may get refined later.  Assuming it passes the tests, it is ready to be merged.

rbberger (approved):

Issue #2351, opened 2020-09-08
[Feature Request] Overdamped Langevin dynamics / Brownian Dynamics


XanderDW:
**Summary**

I would like to implement a new fix that simulates Langevin dynamics in the overdamped limit, commonly referred to as Brownian dynamics, so neglecting the inertial second derivative term in the equations of motion.

**Detailed Description**

Langevin dynamics in the overdamped limit, as described here: [Brownian dynamics on Wikipedia](https://en.wikipedia.org/wiki/Brownian_dynamics)

I want to include the translational as well as orientational dynamics (for finite-size particles). I think it can be implemented in a single fix that does the integration of the translational and orientational (quaternion) degrees of freedom. Velocity and angular momentum can be disregarded.

akohlmey:
@XanderDW please clarify: are you announcing that you are starting to program this fix?
or are you looking for a collaborator to implement it?

XanderDW:
@akohlmey Apologies for unclarity: I'm starting development on this fix myself.

akohlmey:
> @akohlmey Apologies for unclarity: I'm starting development on this fix myself.

No problem. I was just asking because some people write *implement* when they really mean *use*.

Thanks for taking the initiative and reporting it here.

XanderDW:
Just finished the implementation and did some sanity checks to confirm that everything seems to work as it should.

I attached the files here. Since this is my first contribution to LAMMPS, perhaps you could check if the code properly adheres to all guidelines and conventions (apart from inclusion of documentation)? 

Specifically I wonder: how should I include 2D support? Do I write a separate fix for the 2D/3D case or should I write my code more generally for d-dimensional spaces or is there another clever trick?

[LAMMPS_fix_brownian.zip](https://github.com/lammps/lammps/files/5208437/LAMMPS_fix_brownian.zip)

akohlmey:
> Just finished the implementation and did some sanity checks to confirm that everything seems to work as it should.
> 
> I attached the files here. Since this is my first contribution to LAMMPS, perhaps someone could check if the code properly adheres to all guidelines and conventions (apart from inclusion of documentation)?

Please create a pull request for your contribution to be integrated. https://lammps.sandia.gov/doc/Howto_github.html
[...]

PR #2511, opened 2020-12-08
Add USER-BROWNIAN package

samueljmcameron:
**Summary**

As far as I know, there is no overdamped brownian dynamics integrator available which allows for both rotational and translational diffusion on lammps. The files fix_bd_sphere.cpp and fix_bd_sphere.h will allow users to do this type of simulation. See the doc/src/fix_bd_sphere.rst for a full description. 

**Related Issue(s)**

#2351

**Author(s)**

Sam Cameron - samuel.j.m.cameron@gmail.com

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

I also made minor change to the random_mars src files, but only to implement a very trivial function (uniform rng between -0.5 and 0.5) so that users of fix_bd_sphere can choose whether to use gaussian rng or uniform rng at runtime.

Correctness of integrator was verified by simulating 2D and 3D systems of non-interacting particles for long enough to measure translational and rotational diffusion. I observed the expected translational MSD = 2*d*D_t*t where D_t is the user-specified translational diffusion coefficient and d is dimension. I observed the expected rotational correlations (e(t) dot e(0)) = e^{-(d-1)D_r t}  where D_r is the user defined rotational diffusion constant.

**Post Submission Checklist**

Good books on overdamped brownian dynamics are "The Theory of Polymer Dynamics" (Edwards and Doi) and "Soft Matter" (Doi).


src/fix_bd_sphere.cpp
```diff
@@ -0,0 +1,303 @@
...
+
+  time_integrate = 1;
+  
+  if (narg != 8 && narg != 10)
+    error->all(FLERR,"Illegal fix bd/sphere command.");
+
+  if (!atom->sphere_flag)
+    error->all(FLERR,"Fix bd/sphere requires atom style sphere");
+  if (!atom->mu_flag)
+    error->all(FLERR,"Fix bd/sphere requires atom attribute mu");
```
akohlmey, line 53: why this? is there nothing intrinsic in brownian dynamics for extended spherical objects that requires a dipole.
please see the source for fix nve/sphere how to keep support for dipoles with such an integrator optional.

src/fix_bd_sphere.cpp
```diff
@@ -0,0 +1,303 @@
...
+	noise_flag = 0;
+      } else {
+	error->all(FLERR,"Illegal fix/bd/sphere command.");
+      }
+    } else {
+      error->all(FLERR,"Illegal fix/bd/sphere command.");
+    }
+    iarg = iarg + 2;
+  }
+
```
akohlmey, line 101: please reduce excess empty lines for better readability of the source code

src/fix_bd_sphere.cpp
```diff
@@ -0,0 +1,303 @@
...
+      
+      omega[i][0] = d3rot * g3* torque[i][0];
+      omega[i][1] = d3rot * g3* torque[i][1];
+      omega[i][2] = g3* torque[i][2];
+      
+      dtheta = sqrt((omega[i][0]*dt)*(omega[i][0]*dt)
+		    +(omega[i][1]*dt)*(omega[i][1]*dt)
+		    +(omega[i][2]*dt)*(omega[i][2]*dt));
+      
+      if (abs(dtheta) < 1e-14) {
```
akohlmey, line 216: please use a global constant or define (usually called SMALL) .
also `abs()` applies to integers, and thus you should be using `fabs()` here.
please also use correct indentation below and note that there should be no tabs, only blanks.

akohlmey (changes requested):
I have added some specific requests to the source code. In general, we try to reduce using initials with new commands, as they become confusing if there are too many. So it may be better to rename the command and files to fix brownian/sphere. This is a bit of a corner case, since BD is usually understood as brownian dynamics, so confusion is less likely.
@sjplimp my chime in with his opinion on this matter.

I didn't see any discussion how this differs from using fix nve/sphere with fix langevin using the omega or angmom option.

Individual single file contributions are usually added to the USER-MISC package.

The documentation files need to be fully integrated into the manual. please see the warnings printed after "make html" or by running "make style_check" directly.

samueljmcameron:
> I have added some specific requests to the source code. In general, we try to reduce using initials with new commands, as they become confusing if there are too many. So it may be better to rename the command and files to fix brownian/sphere. This is a bit of a corner case, since BD is usually understood as brownian dynamics, so confusion is less likely.
> @sjplimp my chime in with his opinion on this matter.

Thank you for your comments here. I will wait to hear back on whether to change the naming from bd to brownian.


> 
> I didn't see any discussion how this differs from using fix nve/sphere with fix langevin using the omega or angmom option.
> 

From what I understand (but very possible I am wrong), fix nve/sphere doesn't allow for overdamped dynamics where both the moment of inertia and mass are set to zero. The difference is instead of integrating m*dv/dt = F - gamma_t*dr/dt and I*domega/dt = T - gamma_r*omega (a system of second order ODEs with I being moment of inertia, omega being angular momentum, F and T being conservative forces and torques respectively) you end up integrating just dr/dt = F/gamma_t and omega = T/gamma_t (system of first order ODEs). Is this already implemented?

> Individual single file contributions are usually added to the USER-MISC package.
> 

I wasn't sure if maybe this was something that would be in high enough demand to have in the main src folder, since it's fairly standard to use overdamped dynamics (at least in the field of e.g. biological or polymer physics). I was somewhat surprised that something like this didn't already exist (but maybe it does from what you said above).

So you are suggesting I add it to the USER-MISC folder instead? 

> The documentation files need to be fully integrated into the manual. please see the warnings printed after "make html" or by running "make style_check" directly.

Okay I will check this, thank you!

src/fix_bd_sphere.cpp
samueljmcameron: Yes good point. I had a specific system in mind here, I will make the changes.

src/fix_bd_sphere.cpp
samueljmcameron: Will do.

src/fix_bd_sphere.cpp
samueljmcameron: Okay thank you!

samueljmcameron:
I've completed the changes (I think) that were requested, although I ran into a (potential) issue when compiling the docs (using make html). I got the following output:



make[1]: Entering directory '/home/mv19141/Documents/samslammps/doc/graphviz'
make[1]: Nothing to be done for 'all'.
make[1]: Leaving directory '/home/mv19141/Documents/samslammps/doc/graphviz'
Running Sphinx v3.3.1
Initializing Spelling Checker 7.1.0
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 889 source files that are out of date
updating environment: [new config] 889 added, 0 changed, 0 removed
reading sources... [100%] reset_timestep .. write_restart                                            
waiting for workers...
/home/mv19141/Documents/samslammps/doc/src/Developer_utils.rst:89: WARNING: Parsing of expression failed. Using fallback parser. Error was:
Error in postfix expression, expected primary expression or type.
If primary expression:
  Invalid C++ declaration: Unexpected end during inside string. [error at 108]
    size_t LAMMPS_NS::utils::trim_and_count_words (const std::string &text, const std::string &separators=" \\")
    ------------------------------------------------------------------------------------------------------------^
If type:
  Invalid C++ declaration: Expected identifier in nested name. [error at 102]
    size_t LAMMPS_NS::utils::trim_and_count_words (const std::string &text, const std::string &separators=" \\")
    ------------------------------------------------------------------------------------------------------^

looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
writing output... [100%] third_order .. write_restart                                                
waiting for workers...
generating indices... genindex py-modindex done
writing additional pages... search done
copying images... [100%] JPG/third_order_force_constant.png                                          
copying static files... done
copying extra files... done
dumping search index in English (code: en)... done
dumping object inventory... done
build succeeded, 1 warning.

The HTML pages are in html.
Copying tabs assets
############################################
Found 910 anchor labels
No anchor label errors.
Found 33 standard and 42 user packages
Parsed style names w/o suffixes from C++ tree in ../src:
   Angle styles:      22    Atom styles:       25
   Body styles:        3    Bond styles:       19
   Command styles:    44    Compute styles:   145
   Dihedral styles:   16    Dump styles:       31
   Fix styles:       230    Improper styles:   13
   Integrate styles:   4    Kspace styles:     15
   Minimize styles:    9    Pair styles:      238
   Reader styles:      4    Region styles:      8
----------------------------------------------------
Total number of styles (including suffixes): 1443
Total number of style index entries: 1066
############################################
Build finished. The HTML pages are in doc/html.



I'm not sure if if that warning is a big deal or not. Please let me know.

akohlmey:
> I've completed the changes (I think) that were requested, although I ran into a (potential) issue when compiling the docs (using make html). I got the following output:

[...]

> Error in postfix expression, expected primary expression or type.
> If primary expression:
> Invalid C++ declaration: Unexpected end during inside string. [error at 108]
> size_t LAMMPS_NS::utils::trim_and_count_words (const std::string &text, const std::string &separators=" \")
> ------------------------------------------------------------------------------------------------------------^
> If type:
> Invalid C++ declaration: Expected identifier in nested name. [error at 102]
> size_t LAMMPS_NS::utils::trim_and_count_words (const std::string &text, const std::string &separators=" \")

> I'm not sure if if that warning is a big deal or not. Please let me know.

this must be due to breathe/doxygen. i have not seen this on my machine with doxygen version 1.8.20
it has no impact on the style checking and is certainly not related to any changes you did.

akohlmey:
> > I have added some specific requests to the source code. In general, we try to reduce using initials with new commands, as they become confusing if there are too many. So it may be better to rename the command and files to fix brownian/sphere. This is a bit of a corner case, since BD is usually understood as brownian dynamics, so confusion is less likely.
> > @sjplimp my chime in with his opinion on this matter.

@sjplimp is currently not receiving these notifications. so it may take a few days until he can comment.

> > I didn't see any discussion how this differs from using fix nve/sphere with fix langevin using the omega or angmom option.
> 
> From what I understand (but very possible I am wrong), fix nve/sphere doesn't allow for overdamped dynamics where both the moment of inertia and mass are set to zero. The difference is instead of integrating m_dv/dt = F - gamma_t_dr/dt and I_domega/dt = T - gamma_r_omega (a system of second order ODEs with I being moment of inertia, omega being angular momentum, F and T being conservative forces and torques respectively) you end up integrating just dr/dt = F/gamma_t and omega = T/gamma_t (system of first order ODEs). Is this already implemented?

@sjplimp is more competent to answer this.

> > Individual single file contributions are usually added to the USER-MISC package.
> 
> I wasn't sure if maybe this was something that would be in high enough demand to have in the main src folder, since it's fairly standard to use overdamped dynamics (at least in the field of e.g. biological or polymer physics). I was somewhat surprised that something like this didn't already exist (but maybe it does from what you said above).
> 
> So you are suggesting I add it to the USER-MISC folder instead?

yes. that would also require a 1-line entry in the corresponding README file and moving the examples to examples/USER/misc.
i noticed when looking over your changes that you keep a cached copy of `dt`. if you do that (and also cache derived values), you should add a `reset_dt()` function, so that your integrator can handle the situation that `Update::dt` is changed during a run.

> > The documentation files need to be fully integrated into the manual. please see the warnings printed after "make html" or by running "make style_check" directly.
> 
> Okay I will check this, thank you!

will take a closer look and possibly push some more cosmetic changes to the PR (mostly whitespace stuff)

samueljmcameron:
> > > Individual single file contributions are usually added to the USER-MISC package.
> > 
> > 
> > I wasn't sure if maybe this was something that would be in high enough demand to have in the main src folder, since it's fairly standard to use overdamped dynamics (at least in the field of e.g. biological or polymer physics). I was somewhat surprised that something like this didn't already exist (but maybe it does from what you said above).
> > So you are suggesting I add it to the USER-MISC folder instead?
> 
> yes. that would also require a 1-line entry in the corresponding README file and moving the examples to examples/USER/misc.

I am wondering now if I should make this part of a small user package. The reason I say this is because my original plan was to create a user package which allowed for simulation of ``active brownian particles'', which would have included this fix_bd_sphere as well as another fix to add an ``active'' force to the system (just adds a self propulsion term, see the fifth slide of this link: https://www.newton.ac.uk/files/seminar/20130618110012001-153602.pdf ). But then I thought that this integrator might be useful outside of those simulations. Do you think it would be better to upload the two fixes (fix_bd_sphere and the active force fix) separately, or add the active force to this pull request and create a USER-ACTIVEBROWNIANPARTICLE folder? Let me know your thoughts.

> i noticed when looking over your changes that you keep a cached copy of `dt`. if you do that (and also cache derived values), you should add a `reset_dt()` function, so that your integrator can handle the situation that `Update::dt` is changed during a run.

Thank you! I will make the changes.

akohlmey:
> I am wondering now if I should make this part of a small user package. The reason I say this is because my original plan was to create a user package which allowed for simulation of ''active brownian particles'', which would have included this fix_bd_sphere as well as another fix to add an ''active'' force to the system (just adds a self propulsion term, see the fifth slide of this link: https://www.newton.ac.uk/files/seminar/20130618110012001-153602.pdf ).

there is already fix propel/self which *is* part of the USER-MISC package.

> But then I thought that the fix_bd_sphere might be useful outside of those simulations. Do you think it would be better to upload the two fixes (fix_bd_sphere and the active force fix) separately, or add the active force to this pull request and create a USER-ACTIVEBROWNIANPARTICLE folder? Let me know your thoughts.

we don't create a user package for just a single (or two) styles unless the package is an interface to an external library (like USER-PLUMED). USER-MISC seems the reasonable choice to me.

samueljmcameron:
> > I am wondering now if I should make this part of a small user package. The reason I say this is because my original plan was to create a user package which allowed for simulation of ''active brownian particles'', which would have included this fix_bd_sphere as well as another fix to add an ''active'' force to the system (just adds a self propulsion term, see the fifth slide of this link: https://www.newton.ac.uk/files/seminar/20130618110012001-153602.pdf ).
> 
> there is already fix propel/self which _is_ part of the USER-MISC package.

Ah yes. So there are two obstacles for me to use the already existing propel/self fix. The first is that it does not add a contributtion to the virial, the second is that it requires ellipsoidal particles. Should I create a pull request for my self-propelling fix to discuss these further? I can also chat about it here but it seems like it would be off topic.

> 
> > But then I thought that the fix_bd_sphere might be useful outside of those simulations. Do you think it would be better to upload the two fixes (fix_bd_sphere and the active force fix) separately, or add the active force to this pull request and create a USER-ACTIVEBROWNIANPARTICLE folder? Let me know your thoughts.
> 
> we don't create a user package for just a single (or two) styles unless the package is an interface to an external library (like USER-PLUMED). USER-MISC seems the reasonable choice to me.

Okay! I have moved things so that it is part of the USER-MISC folder. Thanks for the clarification.

samueljmcameron:
I added another integrator here for ellipsoidal particles.

akohlmey:
FYI. This PR now reverts changes to the main branch that are unrelated. Something must have gone wrong with updating the PR.

samueljmcameron:
> FYI. This PR now reverts changes to the main branch that are unrelated. Something must have gone wrong with updating the PR.

Whoops sorry. I see that now, I have hopefully fixed the issue.

XanderDW:
Hi, I was also working on a new fix to achieve something similar. It also implements Brownian Dynamics in the overdamped limit for both the translational and the rotational degrees of freedom (same EOMs as you), the latter accepting any atom types that define a quaternion, allowing it to be used for active brownian particle simulations in conjunction with the aforementioned fix propel/self. I compared it to your code and noticed a few minor differences, for example in the algorithm for updating the quaternion (my implementation lets you choose between a faster or a more accurate algorithm) and there are a few semantic differences. I included my code alongside with an Active Brownian Particle example and a PDF with some doc. Feel free to compare the code with your implementation and to include parts of it in your implementation if you want (with a brief mention of authorship if it's anything significant :) )
[brownian_LAMMPS.zip](https://github.com/lammps/lammps/files/5720943/brownian_LAMMPS.zip)

samueljmcameron:
> Hi, I was also working on a new fix to achieve something similar. It also implements Brownian Dynamics in the overdamped limit for both the translational and the rotational degrees of freedom (same EOMs as you), the latter accepting any atom types that define a quaternion, allowing it to be used for active brownian particle simulations in conjunction with the aforementioned fix propel/self. I compared it to your code and noticed a few minor differences, for example in the algorithm for updating the quaternion (my implementation lets you choose between a faster or a more accurate algorithm) and there are a few semantic differences. I included my code alongside with an Active Brownian Particle example and a PDF with some doc in a zip file attached to this message. Feel free to compare the code with your implementation and to include parts of it in your code if you want (with a brief mention of authorship if it's anything significant :) )
> [brownian_LAMMPS.zip](https://github.com/lammps/lammps/files/5720943/brownian_LAMMPS.zip)

Hi there! I will definitely look at your code, thank you! I did not realise that there was already a feature request open for brownian dynamics, so I apologise if I stepped on your toes.

samueljmcameron:
@akohlmey I have merged the already existing fix propel/self with my code, and put it all into a USER-BROWNIAN package. I did not edit Packages_user.rst and Packages_details.rst, as I assumed that might be the changes you were alluding to that were better done by you. If not, I can look into changing it myself!

akohlmey:
> @akohlmey I have merged the already existing fix propel/self with my code, and put it all into a USER-BROWNIAN package. I did not edit Packages_user.rst and Packages_details.rst, as I assumed that might be the changes you were alluding to that were better done by you. If not, I can look into changing it myself!

Yes, I will do that, but before doing so, I have a few more requests for changes after looking over the state of the PR right now.

doc/src/fix_brownian.rst
```diff
@@ -0,0 +1,151 @@
...
+   simulations with additive noise. There are two important cases which
+   (conceptually) reduce the number of free parameters in this fix.
+   (a) In equilibrium simulations
+   (where fluctuation dissipation theorems are obeyed), one can define
+   the thermal energy :math:`k_bT=D_t\gamma_t`.
+   
+---------
+
+.. note::
+   Temperature computation using the :doc:`compute temp <compute_temp>`
```
akohlmey, line 63: should we then perhaps have a new temperature compute included in this package that outputs a temperature which is commensurate with the brownian dynamics conditions? most thermostat fixes create a temperature compute internally and offload the computation of the system temperature to that compute and also have an option to modify this or even do some kind of "biasing", where the thermostat can be restricted to what is thermostatted how, e.g. to dimensions, regions or gradients. Note, I am not says that this must be done, but am merely suggesting to think about this. Check out the HowTo document about thermostatting (and the included links) in the manual for some general information. This could also be done in a second step as a separate pull request. This one looks mostly OK, so I would prefer to merge this ASAP and do further improvements later rather than postponing the merge indefinitely until  it is "better than perfect".

Mind you, that for atom style sphere, you would need to compare to compute temp/sphere and for atom style ellipsoid to compute temp/asphere anyway, as compute temp does not account for the rotational kinetic energy.

doc/src/fix_brownian.rst
```diff
@@ -0,0 +1,151 @@
...
+   will not correctly compute temperature of these overdamped dynamics
+   since we are explicitly neglecting inertial effects.
+   See e.g. chapter 6 of :ref:`(Doi) <Doi1>` for more details on this.
+   Temperature is instead defined in terms of the note above (for
+   equilibrium systems).
+
+---------
+
+.. note::
+   The diffusion coefficient :math:`D_t` is measured
```
akohlmey, line 73: the units of keywords are more commonly given where the keywords are listed and not in a "Note" section.

doc/src/fix_brownian.rst
```diff
@@ -0,0 +1,151 @@
...
+complexity is unnecessary, and one should be fine using the *uniform*
+value for reasons argued in :ref:`(Dunweg) <Dunweg6>`.
+
+If the *rng* keyword is used with the *none* value, then the noise
+terms are set to zero.
+
+
+----------
+
+.. include:: accel_styles.rst
```
akohlmey, line 97: There are no accelerated versions of these styles implemented (yet), so this include can be dropped.

doc/src/fix_brownian.rst
```diff
@@ -0,0 +1,151 @@
...
+
+----------
+
+Restart, fix_modify, output, run start/stop, minimize info
+"""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
+
+No information about this fix is written to :doc:`binary restart files <restart>`.
```
akohlmey, line 104: it should be possible to store the state of the pRNGs into the restart file... see fix temp/csvr and fix temp/csld (but note that there is a bug in the MPI_Gather() call that is only fixed in a pending pull request.

doc/src/fix_propel_self.rst
```diff
@@ -8,94 +8,165 @@ Syntax
 
 .. parsed-literal::
 
-   fix ID group-ID propel/self mode magnitude keyword values ...
+   fix ID group-ID propel/self magnitude keyword values
```
akohlmey, line 11: sorry, but this change should be undone. the "mode" keyword should remain at the location where is currently is (for compatibility) since it is required and not a "keyword/value" pair as is intended for the additional arguments.

doc/src/fix_propel_self.rst
```diff
@@ -8,94 +8,165 @@ Syntax
...
+* magnitude = magnitude of self-propulsion force
+* one (and only one) keyword/value pair must be appended to args
+* keyword = *dipole* or *velocity* or *quat*
+
+  .. parsed-literal::
+
+       *dipole* value = none = apply force along dipole direction
+       *velocity* value = none = apply force along velocity direction
+       *quat* values = direction vector *sx* and *sy* and *sz*
+        *sx* = x component of force direction in ellipsoid frame
```
akohlmey, line 24: the direction in the ellipsoid frame should be added as an optional keyword, say, "qvector \<sx\> \<sy\> \<sz\>" or similar.

doc/src/fix_propel_self.rst
```diff
@@ -8,94 +8,165 @@ Syntax
...
+	*sy* = y component of force direction in ellipsoid frame
+	*sz* = z component of force direction in ellipsoid frame
 
-  *types* values = one or more atom types
```
akohlmey, line 20: am I correct to assume that that "types" keyword was dropped because the same effect can be achieved with a fix group?

src/USER-BROWNIAN/fix_brownian_sphere.cpp
```diff
@@ -0,0 +1,321 @@
...
+
+void FixBrownianSphere::init()
+{
+  
+  g1 =  force->ftm2v/gamma_t;
+  g3 = force->ftm2v/gamma_r;
+  if (noise_flag == 0) {
+    g2 = 0;
+    g4 = 0;
+    rng_func = &RanMars::zero_rng;
```
akohlmey, line 149: rather than using function pointers (and adding convenience functions to the pRNG class), I would suggest to just use if statements here and in the post_force() code. To avoid performance penalties, you can follow the example of fix langevin, which introduces a function template for different modes (and combination of modes) which are then selected in post_force() so the actual "if" is done outside the inner loop(s) and the compiler will then happily optimize away the if statements in the templated functions and effectively create multiple instances of the template as needed at compile time without requiring to program them explicitly.

src/USER-BROWNIAN/fix_brownian.cpp
```diff
@@ -0,0 +1,217 @@
...
+   Steve Plimpton, sjplimp@sandia.gov
+
+   Copyright (2003) Sandia Corporation.  Under the terms of Contract
+   DE-AC04-94AL85000 with Sandia Corporation, the U.S. Government retains
+   certain rights in this software.  This software is distributed under
+   the GNU General Public License.
+
+   See the README file in the top-level LAMMPS directory.
+------------------------------------------------------------------------- */
+
```
akohlmey, line 13: Here is another idea to ponder, but not a requirement: since there seems to be a significant amount of shared code, it might be conceivable to have all three brownian dynamics integrators implemented in a shared base class and then the three variants just thin wrappers. See fix_nh.cpp vs. fix_nvt.cpp and fix_npt.cpp for example. Please also see my comment about avoiding performance critical if statements in inner loops through template functions. The choice point vs. sphere vs. ellipsoid could be a second "dimension", i.e. a second template option.

doc/src/fix_propel_self.rst
```diff
@@ -8,94 +8,165 @@ Syntax
...
 
 .. code-block:: LAMMPS
 
-   fix active_group all propel/self velocity 1.0
-   fix constant_velocity all viscous 1.0
-
-   fix active_group all propel/self quat 1.0
-
-   fix active all propel/self quat 1.0 types 1 2 4
+   fix propel/self all 40.0 dipole
```
akohlmey, line 34: the syntax example is incorrect here. for all fixes there must be: fix <fix-ID> <fix-group> <fix-name>

akohlmey (changes requested):
Thanks for all the effort. Here are some requests and some suggestions or ideas. I hope to have clearly indicated which of my comments are optional. Feel free to ask if something is not clear. FWIW, we also have a "LAMMPS developer" slack in case you want a faster turnaround for questions (but keep the time difference in mind). You can email us a slack@lammps.org if you want to get an invite.

XanderDW:
> > Hi, I was also working on a new fix to achieve something similar. It also implements Brownian Dynamics in the overdamped limit for both the translational and the rotational degrees of freedom (same EOMs as you), the latter accepting any atom types that define a quaternion, allowing it to be used for active brownian particle simulations in conjunction with the aforementioned fix propel/self. I compared it to your code and noticed a few minor differences, for example in the algorithm for updating the quaternion (my implementation lets you choose between a faster or a more accurate algorithm) and there are a few semantic differences. I included my code alongside with an Active Brownian Particle example and a PDF with some doc in a zip file attached to this message. Feel free to compare the code with your implementation and to include parts of it in your code if you want (with a brief mention of authorship if it's anything significant :) )
> > [brownian_LAMMPS.zip](https://github.com/lammps/lammps/files/5720943/brownian_LAMMPS.zip)
> 
> Hi there! I will definitely look at your code, thank you! I did not realise that there was already a feature request open for brownian dynamics, so I apologise if I stepped on your toes.

No worries, let me know if you need any clarification. I will be following this thread closely.

doc/src/fix_propel_self.rst
samueljmcameron: That is correct. I am happy to re-introduce it if you think that would be best?

doc/src/fix_propel_self.rst
samueljmcameron: Thanks for catching this!

doc/src/fix_propel_self.rst
samueljmcameron: Okay, I have updated the doc accordingly. I wasn't 100% on syntax so let me know if I need to make further changes.

doc/src/fix_propel_self.rst
samueljmcameron: Okay! I have done as you requested.

samueljmcameron:
I think I have implemented all the fix propel/self changes with the latest commit. I will work on the brownian changes now.

akohlmey:
@samueljmcameron can you please update us on the status of this PR.

samueljmcameron:
> @samueljmcameron can you please update us on the status of this PR.

Yes sorry, I got side-tracked by some other work after I realised that my treatment of the ellipsoidal brownian motion equations was a bit naive. I am revisiting this PR now and should have a new commit with the requested changes and my changes by the end of this week.

doc/src/fix_brownian.rst
```diff
@@ -0,0 +1,151 @@
+.. index:: fix brownian
+
+fix brownian command
```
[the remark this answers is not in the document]
samueljmcameron: I have done this now.

doc/src/fix_brownian.rst
samueljmcameron: Would be interested in doing this in the future, but figured given how long I've dragged my feet on this its better to get this out now and then start a new PR as you've said.

doc/src/fix_brownian.rst
samueljmcameron: Done.

doc/src/fix_brownian.rst
samueljmcameron: Haven't done this yet, would be happy to try to add it once everything else is sorted.

src/USER-BROWNIAN/fix_brownian_sphere.cpp
samueljmcameron: Done (I think), I'm still somewhat new to C++ so I'm not super confident in this, but everything is working as I would expect (e.g. diffusion in non-interacting systems is correct).

src/USER-BROWNIAN/fix_brownian.cpp
samueljmcameron: Sort of did this. The code became a bit more complex when I corrected my treatment of anisotropic rotational diffusion, so I don't think it separated as nicely after all.

samueljmcameron:
@akohlmey new changes are up.

akohlmey:
@samueljmcameron I just updated the pull request branch to the latest upstream code, addressed some minor issues in the code, completed the integration into the build systems, and filled in missing pieces in the manual as well as fixed spelling issues and updated the list of false positives. There may be a few more updates, if our automatic testing detects problems, but after that, it should be formally ready.

akohlmey (approved):

samueljmcameron:
> @samueljmcameron can you please update me on the status of this pull request and whether there is something that you still want to add/update and what I can do to help getting it ready. I am currently looking over some nearly completed pull requests to see if they can be finished and merged before the pending patch release since we are going to make some changes after the patch release that will require additional changes to most pending pull requests.

Hi @akohlmey, I think I'm happy with this on my end.

samueljmcameron:
> @samueljmcameron I just updated the pull request branch to the latest upstream code, addressed some minor issues in the code, completed the integration into the build systems, and filled in missing pieces in the manual as well as fixed spelling issues and updated the list of false positives. There may be a few more updates, if our automatic testing detects problems, but after that, it should be formally ready.

Thanks for all this!

PR #2512, opened 2020-12-08
Add pair/only flag to Kokkos

stanmoore1:
**Summary**

Add a `pair/only` flag to Kokkos. This is a convenient command-line shortcut to run only the `pair_style` and neighbor list on the GPU and everything else on the host CPU, and basically allows the KOKKOS package to easily mimic the GPU package. This mode can be faster for small systems when running on more than 1 MPI rank, or when using diagnostics (fixes/computes) haven't been ported to Kokkos yet. Without this PR, a user must manually edit the input script and change the kokkos package args to get this mode of operation.

**Related Issue(s)**

None.

**Author(s)**

Stan Moore (SNL)

**Licensing**

**Backward Compatibility**

Yes

### DIFF
--- a/src/KOKKOS/fix_qeq_reax_kokkos.h
+++ b/src/KOKKOS/fix_qeq_reax_kokkos.h
@@ -241,7 +241,7 @@ class FixQEqReaxKokkos : public FixQEqReax, public KokkosBase {
 
 template <class DeviceType>
 struct FixQEqReaxKokkosNumNeighFunctor  {
-  typedef DeviceType  device_type ;
+  typedef DeviceType device_type;
   typedef int value_type ;
   FixQEqReaxKokkos<DeviceType> c;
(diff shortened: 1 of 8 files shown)


stanmoore1:
@sjplimp @akohlmey I touched a few core files, can you please take a look? I am open to suggestions if you have a different idea of how to do this.

akohlmey:
@stanmoore1 I like the basic idea of having a flag to go with the -package or -kokkos command line flag to trigger this behavior. That seems to me the more logical place from the user perspective even though it is technically is an option to the suffix command. I am interested to have this available for other accelerator packages as well.
What I am not so comfortable with is the way how it is implemented with adding/removing the suffix flag. I have to think some more about this, but it seems to me a more general solution wold be to take advantage that there can be *two* suffixes with the second one being the fallback for the first. The the pair/only flag would essentially trigger skipping the first suffix for all cases but for pair styles.

stanmoore1:
Closing in favor of #2533.

PR #2514, opened 2020-12-10
A shifted Tersoff Potential

oywg11:
**Summary**

This potential is designed for simulations of closely matched van der Waals heterostructures. For instance, let's
consider the case of a system with few-layers graphene atop a thick hexagonal boron nitride (h-BN) substrate
simulated using periodic boundary conditions. The experimental lattice mismatch of ~1.8% between graphene and h-BN
is not well captured by the equilibrium lattice constants of available potentials, thus a small in-plane strain
will be introduced in the system when building a periodic supercell. To minimize the effect of strain on
simulation results, the *_tersoff/shift_* style is proposed that allows adjusting the equilibrium bond length
of one of the two materials (e.g., h-BN). Validitation, benchmark tests and applications of the *_tersoff/shift_* style
can be found in ACS Nano 13, 7603 (2019) and J. Chem. Theory Comput. 16, 666 (2020).

**Related Issue(s)**

**Author(s)**

Wengen Ouyang, Tel Aviv University, Email: w.g.ouyang@gmail.com
Davide Mandelli, Istituto Italiano di Tecnologia

**Licensing**

**Backward Compatibility**

The backward compatibility is kept

**Implementation Notes**

The *tersoff/shift* style computes the energy E of a system of atoms, whose formula is the same as the original Tersoff potential. The only modification is that the original equilibrium bond length `r_0` of the system is shifted to `r_0-\delta`, where values of `\delta < 0` will result in elongation of the bond length, while values of `\delta > 0` will result in shrinking of the bond length.

**Post Submission Checklist**

Validitation, benchmark tests and applications of the *tersoff/shift* style can be found in https://pubs.acs.org/doi/10.1021/acsnano.9b00645 and https://pubs.acs.org/doi/10.1021/acs.jctc.9b00908.


akohlmey:
I am not sure if this is a contribution we want in LAMMPS because it looks a lot like some quick hack. @sjplimp @athomps any comments on that?

Some more comments on formal issues:
- This class should be implemented as a derived class so that common code is shared.
- This replaces calls to cube() and square() with calls to pow() which are much slower. This should be reverted.
- The rather specific variable name "shiftBN" should be renamed into something more generic as the same shift would be applied to anything and does not pay attention to any specific atom types or elements
- Commented out (original or temporary) code should be deleted
- Any reference to the variable `shiftCC` should be removed as it is unused
- We are about to change our policy about including references for the CiteMe class: There should only be a reference for the most relevant citation describing the **implementation** and not updates to parameters or similar. There may only be multiple references, if they are connected to specific features that can be tested for with an if statement, e.g. when an option is added to an existing style that changes some functions. So most of the references added by this pull request should be dropped.

oywg11:
@akohlmey Hi Axel, thanks very much for your comments. Actually, I agree with you that the contribution of this potential is small, this is the reason why I didn't submit it along with the ILP two years ago. The thing that made me decide to submit it finally is that there are more and more researchers recently, sending emails to request the source code of this potential for similar MD simulations mentioned in this pull request. Thus I thought this potential might still be useful. Please feel free to judge the contribution of this potential from the standard of LAMMPS. If it is passed, I will make the corresponding changes as you suggested.

sjplimp:
@oywg11 Conceptually, I agree this is new potential, useful for certain kinds of models, similar to lj/expand or bond fene/expand.  Perhaps this one should be called tersoff/expand for consistency?  I agree with all of @akohlmey suggestions on the code organizaiton, esp to make this a derived class to avoid code duplication.  I'd add that perhaps it doesn't need a stand-alone doc page?  If it makes sense to add to the pair tersoff page as a new sub-section, similar to pair tersoff/table.

oywg11:
@sjplimp Thanks for the comments. I will rename this potential and modify it as a derived class of the original Tersoff potential and merge the doc as a subsection.

athomps:
This appears to be a fairly simple modification of the original pair tersoff.  I agree that it is best handled as a derived class.   One concern I have is whether the change to the force  and energy calculations are consistent.  If yes, it would be helpful to explain this clearly by presenting modified equations for the energy on the doc page.   If no, a conservative potential corresponding to the forces may not exist. If the forces are non-conservative, the method may still be useful, but it does not meet normal expectations of an interatomic potential, so the source files should not be added to the MANYBODY package, but instead added USER-MISC.  Related to that, it would be nice to see how well energy is conserved with tersoff/shift. The file log.30Nov20.hBN_shift.g++.1 uses shift=0.0, and so does not show this.

oywg11:
@akohlmey I modified this potential as a derived class of Tersoff potential. To minimize the duplication of the codes, a "shift_flag" is introduced. Please let me know whether this modification is approparate or not.
@sjplimp The doc of this potential is merged as a subsection of the Tersoff potential.
@athomps The energy conservation of this potential is checked (see log.30Nov20.hBN_shift0.05_1ns.g++.1). More benchmark tests such as the phonon spectrum can be found in https://pubs.acs.org/doi/10.1021/acsnano.9b00645.

athomps:
@oywg11 Okay, I see the shift0.05 case now.  Energy conservation looks pretty good.  However, the BN model has lambda3=0.0, which turns off one of the radial contributions in \zeta_ij.  It would be good if you could test this case for energy conservation.  Also, I still don't know for sure which radial contributions are shifted and which are not.  Writing down the modified expressions (or modifying the original expressions) for f_R, f_A, f_c, \zeta_ij would solve that.

oywg11:
@athomps Thanks for the suggestions! Now the energy conservation for non-zero lambda3 is checked (see log.30Nov20.Si_shift0.05.g++.1). I will add the expression for the potential later after a careful check.

oywg11:
@athomps I modified the original expresion of Tersoff Potential to include tersoff/shift.

athomps:
@oywg11 Energy conservation and energy equation look good.

akohlmey:
@sjplimp @athomps, if you are OK with having this set of changes so deeply integrated into the standard code path, then there is not really a need to have a separate pair style for it, but it can be become just an optional keyword "shift \<shift value\>" to the tersoff pair style(s) and the change could be propagated to *all* tersoff variants including the accelerated ones (at least the USER-OMP and USER-INTEL versions).

athomps:
Yes, now that I see the modified equations for energy, and the good energy conservation, and think this should be fully integrated with all the Tersoff variants.

sjplimp:
I'm fine with whatever Aidan wants.

athomps (approved):
Nice work @oywg11

athomps (commented):
I was too hasty in my approval.  This feature is sufficiently useful, well-implemented, and tested, that it is worth adding it as a an option to the base Tersoff potential. @oywg11 can you figure out a way to merge this as an option on the pair_style command e.g. 
`pair_style tersoff shift 0.05`

oywg11:
@athomps Thanks for the comments and suggestions! I will try to merge the tersoff/shift as an option of the original Tersoff potential as you suggested.

akohlmey (changes requested):
Waiting for further refactoring and deeper integration into the tersoff styles.

akohlmey:
@oywg11 don't worry about the change request. I only added it to cancel the approval and block merging until the next iteration of this change is complete. Let us know if you need any help propagating the change to the various Tersoff variants. In some cases, we may need to call on developers of the specific package to help.

oywg11:
@athomps I merged the tersoff/shift style as an option of the original Tersoff potential as you suggested, the energy conversation is also checked. Please let me know if you think further modifications are required.
@akohlmey I only modify the pair_tersoff.cpp and didn't touch the accelerate styles such as *intel*, *omp*, *gpu*, and *kk* since I need to learn the codes for these styles. Are the modifications in these styles the same as those in pair_tersoff.cpp?

akohlmey:
> @akohlmey I only modify the pair_tersoff.cpp and didn't touch the accelerate styles such as _intel_, _omp_, _gpu_, and _kk_ since I need to learn the codes for these styles. Are the modifications in these styles the same as those in pair_tersoff.cpp?

please note, that there are several tersoff variants: tersoff/zbl, tersoff/mod, tersoff/mod/c, tersoff/table plus their accelerated versions. except for tersoff/table those variants are all derived classes where several functions are inherited and a few are replaced. thus should be either made consistent or it should be checked that they don't accept the shift argument or don't parse it, but let the base tersoff class parse it. of the accelerated versions, the /omp variants are the most similar and they just replace the most time critical part with alternate versions and keep most of the original functions, if they are thread safe. /intel should be similar, but may be more complex due to its refactoring of data to help with vectorization. /kk and /gpu are the most complex to change. with /kk you may get lucky and a side-by-side comparison could be sufficient to determine what needs to be changed, since your change is rather local. @stanmoore1 is usually available to provide more advice or help with implementing changes. with /gpu it is not advisable for you to make the change, since there is the CPU part and the GPU part of the code and since you add a new parameter, that needs to be communicated as well. in that case it would be better to call on @ndtrung81 to assist with the necessary changes. 

my suggestion is that you work your way through the different variants first, add unit test inputs for all cases with a shift, and then look into adding /omp variants, and possibly /intel versions. all of those would be covered by the unittest tools from the same input with the same reference data, i.e. if you get the regular version to work, you would have to match the accelerated versions to those numbers. i would then suggest to leave /kk and /gpu (and possibly /intel, if you run into problems) for 1-2 separate pull requests. for those you have to compare against your own reference outputs and with /gpu and /intel you need to factor in single vs mixed vs double precision force computations.

the important thing is that all variants where you add the full support for the shift keyword, accept it and let the base class parse and apply it, while the ones not converted have to reject it. that way there are no surprises for people who try some variant or want to use an accelerated version.

oywg11:
@akohlmey  Thanks for the very detailed suggestions!
I made slight changes to the source codes of the unaccelerated tersoff variants (tersoff/zbl, tersoff/mod, tersoff/mod/c, tersoff/table) to make the shift option is rejected by them and is only parsed by the basic tersoff class. I already did unit test inputs for all cases with a shift as you suggested. All of them reported an error message when a shift was added in the input, while the output was the same as the unchanged source files when the shift was removed.
For the accelerated styles, I will first work on the *omp* style and then try the *intel* style, it will take me some time to learn the basic mechanisms. After fully understand these acceleration styles, I will also try to add them for the *ilp/graphene/hbn* and *kolmogorov/crespi/full* pair styles. Therefore, any suggestions on understanding these accelerated styles will be grateful.

athomps:
I am working on refactoring pair_tersoff.cpp to simplify dependence on shift option.

akohlmey:
> I am working on refactoring pair_tersoff.cpp to simplify dependence on shift option.

@athomps please note that your changes are breaking the /omp variants of the tersoff styles. see
https://ci.lammps.org/job/dev/job/pull_requests/job/ubuntu/job/unit_tests/job/cmake_mpi_openmp_smallbig_shared/1560/
for an overview.

athomps:
> @athomps please note that your changes are breaking the /omp variants of the tersoff styles. see
> https://ci.lammps.org/job/dev/job/pull_requests/job/ubuntu/job/unit_tests/job/cmake_mpi_openmp_smallbig_shared/1560/
> for an overview.
All /omp variants should now be correct (I did not test tersoff/zbl) and have support for shift that matches non-omp result.

athomps:
I will take a look at doc page and GPU, KOKKOS, USER-INTEL, and USER-MISC variants on Thursday.

akohlmey:
> I will take a look at doc page and GPU, KOKKOS, USER-INTEL, and USER-MISC variants on Thursday.

thanks. unit tests are passing. I have a request for a cosmetic change: I am currently working on making the code formatting more consistent for conditionals and loop constructs. Could you please change all `if(condition)` statements to `if (condition)`.

athomps:
> change all `if(condition)` statements to `if (condition)`?

Done

athomps:
I have updated documentation for pair tersoff, tersoff/mod, and tersoff/zbl.  I did not add shift keyword to GPU, KOKKOS, USER-INTEL, and USER-MISC variants, instead adding the following to pair_tersoff.rst:

`The shift keyword is not supported by the tersoff/gpu, tersoff/intel, tersoff/kk, or tersoff/table variants.`

Maybe there is an easy way to exit with an error in these cases.

oywg11:
@athomps Many thanks for your work!

ndtrung81:
@oywg11  I can help with propagating the changes to the related tersoff pair styles in the GPU package after this PR is merged.

athomps:
@ndtrung81  That would be very helpful.  Please hold off until the base code has stopped changing.  In the same way we merged tersoff/shift in to the base class, I feel like there are some opportunities to merge other derived classes (mod, mod/c, table, zbl) without losing functionality.  I seem to recall a previous discussion with @akohlmey on this topic.  Related to this, there are some opportunities for serial performance improvement by storing intermediate values, eliminating redundant flops in the j an k loops.  The diversity of threebody cutoff conventions make this a little tricky.  Also, I worry about the impact this might have on the optimized variants: OMP, INTEL, GPU, KOKKOS. @oywg11 are you satisfied that the latest version of `shift` keyword on this PR is correct and performs well relative to your `tersoff/shift`?

akohlmey:
@athomps @oywg11 please let me know when you are satisfied with the state of the changes in this PR. For further substantial changes, I suggest to create a new issue to record the ideas and discuss a strategy and then merge this first and have additional changes later. Before merging I would like to add unit tests for the variants with a non-zero value of the shift flag so we have a reference for future refactoring. I can also take care of adding the checks so that the "shift" keyword is only accepted where supported.

athomps:
I agree with this plan.  I am satisfied with the current state, especially once the checks for `shift` keyword support are added.  @oywg11 you should do your own tests on the current state, see if there are any issues.

akohlmey:
code is updated to current master, checks for shift flag with unsupported accelerated styles are added, and unit tests for tersoff variants with shift keyword were added.

akohlmey (approved):

oywg11:
@athomps I compared the output before and after the change of the codes, there exists a slight difference between the results (see before_change.lmp and after_change.lmp) except the first step. However, I would expect exactly the same output since I turned off the "velocity create" command. I will check the codes more carefully to find out the reason for the difference.

akohlmey:
> @athomps I compared the output before and after the change of the codes, there exists a slight difference between the results (see before_change.lmp and after_change.lmp) except the first step. However, I would expect exactly the same output since I turned off the "velocity create" command. I will check the codes more carefully to find out the reason for the difference.

Your output is not using a sufficient precision of the numbers to make an assessment of the accuracy of initial energy vs. forces.
From the log files it appears to be an issue with the forces, but that can be misleading. You have to look at the output with the full 15-16 digit precision provided by double precision floating point numbers. Please also keep in mind that in this case forces are  more susceptible to floating point math issues due to its non-associative nature, and some of the changes introduced significantly changed the order of operations and precomputed properties used multiple times. cases like "invx = 1.0/x; res=y*invx;" versus "res=x/y;" are notorious for leading to slightly different results.

while the differences seem large, they also seem to remain "conservative", i.e. not exponentially growing over time, which is what one most commonly sees when there are issues with the forces.

to confirm that there is an issue somewhere with the introduction of the shift, I suggest you do a second comparison of the same input decks where you simply set the shift to 0.0 and see whether you get the same magnitude of differences or not. the same also for a test case *with* assigning a temperature (so that you push atoms farther away from their minimums).
in our unit tests (which compare reference forces and energies and stresses to older reference data), the refactored versions are within the acceptable spread for the functional form of the pair style. Since we compare the plain versions against OpenMP variants to have the same forces, energies, and stresses, we have to factor in some variation of the results due to the changing order of operations when accumulating forces, energies, and stresses.

akohlmey:
@athomps we need to find a new suitable location for the example input(s) and perhaps decide on reducing the number of examples to one. Possible locations could be either the `examples/threebody` folder, or a new `examples/tersoff-shift` folder.

oywg11:
@akohlmey Benchmark tests for shift=0 are added, as you can see from the output, the results are almost the same. Therefore, it seems to be an issue with the forces for a nonzero shift and I need to check the codes more carefully to find the reason. Another finding is that the new implementation ran slightly slower than the original one.

akohlmey:
@oywg11 thanks for the additional check. I have changed the status of the pull request back to "Draft". Please click on "Ready for review" when you have been successful in eliminating the discrepancies. We will then need to update the unit test reference results. In fact, those newly added tests should fail with changes that reduce or eliminate the discrepancies you reported.

athomps:
> @athomps we need to find a new suitable location for the example input(s) and perhaps decide on reducing the number of examples to one. Possible locations could be either the `examples/threebody` folder, or a new `examples/tersoff-shift` folder.

I propose we create examples/tersoff and move all the Tersoff variants to a single in.tersoff script, similar to in.threebody.

athomps:
> I turned off the "velocity create" command. 
Starting with zero velocity perfect lattice may create bifurcations that depend on machine precision. Better to use:
```velocity all create 300.0 4928459 loop geom```

oywg11:
@akohlmey @athomps I found the issue in the codes and eliminated the discrepancies (see the new log files).

oywg11:
@akohlmey @athomps I found the new implementation of the Tersoff potential (no matter the shift flag is turned on or off) is ~7% slower than the old implementation, is this behavior be expected?

akohlmey:
@oywg11 adding if statements to (inner) loops can have a negative performance impact. Since the flag does not change over the execution of the compute() method, this can be addressed with making compute() a wrapper around a template function (frequently used in the OPT and USER-OMP package), which will cause the template function to be instantiated multiple times where the if statements are eliminated since the result is already known for each instance.

I have just pushed a change to the MANYBODY and USER-OMP versions of the tersoff pair style that implements/adds this. So please check, if this makes a difference in the performance.

I also updated the yaml files for the unit tests with the corrected reference data so that the unit tests won't produce false positives.

oywg11:
@akohlmey Many thanks for the optimization on the codes, now the issue (negative performance impact of adding the shift option) is solved. In the benchmark tests (see shift0.05_300K_after_change_optimized.lmp and shift0.05_300K_before_change_long.lmp), both the output and speed of the simulation are almost the same.

akohlmey:
I've added some additional changes that correct whitespace issues that have crept in when merging with upstream.

@athomps final TODO item seems to be the relocation/reorganization of the example input(s). Would you mind looking into this?
I'll do a final check for cosmetic issues after you re-assign the PR back to me.

athomps:
I created ./examples/tersoff with a few different Tersoff variants, removed all but one from ./examples/threebody.  @oywg11 We could get rid of ./examples/USER/misc/tersoff_shift, and add in.hBN_shift to ./examples/tersoff. What do you prefer?

oywg11:
@athomps I fully agree.

athomps:
I added /examples/tersoff/in.hBN_shift and got rid of ./examples/USER/misc/tersoff_shift.  I think this work is complete.

akohlmey:
@sjplimp any final comments on this? to me this seems ready for merging as well.

sjplimp (approved):

PR #2515, opened 2020-12-10
Update temp/csvr manual

PabloPiaggi:
**Summary**

These PR updates the manual of temp/csvr with information about the conserved quantity in the dynamics generated by this thermostat. Knowledge of this quantity is important to choose an appropriate timestep in the simulation. I have added the definition of the conserved quantity in terms of quantities computed by Lammps. I have also included a brief discussion of how to use it. Before these changes the only way to understand this was reading the original papers and the Lammps' source code for the thermostat.

I have also corrected the information of the global scalar computed by these fixes to reflect that changes in *kinetic* energy are accumulated.

I would also like to discuss the fact that the manual says that this thermostat is not compatible with fix shake. I have used the two fixes together many times with results very close to those of Gromacs. For instance, the simulation of TIP4P water molecules seems to work properly. If there is no evidence for the incompatibility perhaps that comment should be removed.

**Related Issue(s)**

None that I know of.

**Author(s)**

Pablo Piaggi
ppiaggi at princeton.edu
(Princeton University)

**Licensing**

**Backward Compatibility**

Fully backward compatible.

**Implementation Notes**

The source code was not changed. The manual was updated with the interpretation of the computed quantities.

**Post Submission Checklist**

- [x] The source code follows the LAMMPS formatting guidelines (I think so)
- [x] ~~The feature has been verified to work with the conventional build system~~
- [x] ~~The feature has been verified to work with the CMake based build system~~
- [x] ~~Suitable tests have been added to the unittest tree.~~
- [x] ~~A package specific README file has been included or updated~~
- [x] ~~One or more example input decks are included~~

I used the information in the original articles:
https://journals.aps.org/pre/abstract/10.1103/PhysRevE.75.056707
https://aip.scitation.org/doi/full/10.1063/1.2408420
and also the comments in @GiovanniBussi 's implementation here:
https://sites.google.com/site/giovannibussi/Research/algorithms

### DIFF
--- a/doc/src/fix_temp_csvr.rst
+++ b/doc/src/fix_temp_csvr.rst
@@ -125,7 +125,7 @@ removed from each atom, thermostatting is performed on the remaining
 thermal degrees of freedom, and the bias is added back in.
 
 An important feature of these thermostats is that they have an 
-associated effective energy that is a contant of the motion.
+associated effective energy that is a constant of motion.
 The effective energy is the total energy (kinetic + potential) plus 
 the accumulated kinetic energy changes due to the thermostat. The 
 latter quantity is the global scalar computed by these fixes. This 
--- a/doc/src/fix_temp_csvr.rst
+++ b/doc/src/fix_temp_csvr.rst
@@ -166,7 +166,7 @@ These fixes are not invoked during :doc:`energy minimization <minimize>`.
 
 These fixes compute a global scalar which can be accessed by various
 :doc:`output commands <Howto_output>`.  The scalar is the cumulative
-kinetic energy change due to the fix.  The scalar value calculated by this fix
+energy change due to the fix.  The scalar value calculated by this fix
 is "extensive".
 
 Restrictions
--- a/doc/src/fix_temp_csvr.rst
+++ b/doc/src/fix_temp_csvr.rst
@@ -124,6 +124,19 @@ temperature is calculated taking the bias into account, bias is
 removed from each atom, thermostatting is performed on the remaining
 thermal degrees of freedom, and the bias is added back in.
 
+An important feature of these thermostats is that they have an 
+associated effective energy that is a contant of the motion.
+The effective energy is the total energy (kinetic + potential) plus 
+the accumulated kinetic energy changes due to the thermostat. The 
+latter quantity is the global scalar computed by these fixes. This 
+feature is useful to check the integration of the equations of motion 
+against discretization errors. In other words, the conservation of 
+the effective energy can be used to choose an appropriate integration 
+:doc:`timestep <timestep>`. This is similar to the usual paradigm of 
+checking the conservation of the total energy in the microcanonical 
+ensemble.
+
+
 ----------
 
 Restart, fix_modify, output, run start/stop, minimize info
@@ -153,7 +166,7 @@ These fixes are not invoked during :doc:`energy minimization <minimize>`.
 
 These fixes compute a global scalar which can be accessed by various
 :doc:`output commands <Howto_output>`.  The scalar is the cumulative
-energy change due to the fix.  The scalar value calculated by this fix
+kinetic energy change due to the fix.  The scalar value calculated by this fix
 is "extensive".
 
 Restrictions


akohlmey:
@PabloPiaggi thanks for your submission

Some comments:
- what you write about the conserved quantity applies to multiple thermostats in LAMMPS not just temp/csvr
- the documentation originally said the global scalar is the accumulated change of energy **due** to the fix. since the fix only changes the kinetic energy through velocity rescaling, this is a correct description. Adding that it only accumulates "kinetic" energy makes this more obvious, but it is not required to read the source code to understand this.
- as for fix shake: scaling the velocity with constraints may "lose" some energy due to the rescaling being not 100% compatible with the constraint and thus some of the velocity change being canceled when the constraint is applied. The difference would be usually small.

Hopefully @athomps, who is more of an expert in these matters than me, can clarify some more.

athomps:
I don't understand the point about kinetic energy.  Some fix (e.g. Langevin) manipulate the forces, which could be viewed as a change to potential energy.  I don't see any good reason to distinguish the two. Also, the fix energy is added by LAMMPS to the total potential energy.

sjplimp (approved):

sjplimp:
@PabloPiaggi I approved the doc page changes.  But I agree with @athomp about dropping the added "kinetic" word in the 2nd edit.  As Aidan said, all the thermostat fixes (optionally, see fix_modify energy) add their energy change to the potential energy that LAMMPS outputs, not kinetic.  Although in this case velocities are being rescaled (as with some of the other thermostat fixes), it's effectively keeping the kinetic energy constant, not adding/subtracting kinetic energy.  So I think it is less confusing to just refer to "energy" that the thermostat imparts to the system.

PabloPiaggi:
Thank you @akohlmey, @athomps, and @sjplimp for your feedback! I understand if you reject adding the word "kinetic" in order to make the documentation more homogeneous at the expense of being less precise. The fact that the accumulated changes in kinetic energy are the output of these fixes is clear anyway from the discussion about the conserved quantity.

PabloPiaggi:
> @PabloPiaggi I approved the doc page changes. But I agree with @Athomp about dropping the added "kinetic" word in the 2nd edit. As Aidan said, all the thermostat fixes (optionally, see fix_modify energy) add their energy change to the potential energy that LAMMPS outputs, not kinetic. Although in this case velocities are being rescaled (as with some of the other thermostat fixes), it's effectively keeping the kinetic energy constant, not adding/subtracting kinetic energy. So I think it is less confusing to just refer to "energy" that the thermostat imparts to the system.

@sjplimp the kinetic energy is not constant in these fixes. It has the fluctuations appropriate for the canonical ensemble at variance with a simple rescaling of the velocities. Therefore one is in practice adding and subtracting kinetic energy. I hope I didn't misunderstand your answer and thanks for taking the time to give feedback!

PabloPiaggi:
> @PabloPiaggi thanks for your submission
> 
> Some comments:
> 
> * what you write about the conserved quantity applies to multiple thermostats in LAMMPS not just temp/csvr
> * the documentation originally said the global scalar is the accumulated change of energy **due** to the fix. since the fix only changes the kinetic energy through velocity rescaling, this is a correct description. Adding that it only accumulates "kinetic" energy makes this more obvious, but it is not required to read the source code to understand this.
> * as for fix shake: scaling the velocity with constraints may "lose" some energy due to the rescaling being not 100% compatible with the constraint and thus some of the velocity change being canceled when the constraint is applied. The difference would be usually small.
> 
> Hopefully @athomps, who is more of an expert in these matters than me, can clarify some more.

- It is a pity if other fixes have conserved quantities that are already being calculated by Lammps and there is no information in the manual about how to use them.
- The shake issue seems to be subtle. If your reasoning is correct, the same restrictions apply to other thermostats that do not mention this limitation, such as temp/rescale or temp/berendsen. A possible test would be to analyze the kinetic and potential energy, and control if they have the same variance as those obtained, say, with the Nose-Hoover thermostat (assuming that it is the gold standard).

Thanks @akohlmey for taking the time to review this PR!

sjplimp:
Sorry, I meant constant in a time-averaged sense, not that the KE or temp stays literally constant.  I was thinking of the case where you start with a perfect crystal and velocities set to T = 300K, then equilibrate/thermalize the system with a thermostat to keep the T at 300K.  In that case the thermostat adds a bunch of energy, but the (time-averaged) T or KE is unchanged.  All the added energy effectively changes the PE of the system.  Also we want the KE output by LAMMPS to always exactly match the current T.  So (if the user asks for it) we add the thermostat energy to the PE, not the KE.  I agree this is a bit confusing.  But it's why I think it's best in all the doc pages to simply say the thermostat fixes output the energy (not KE or PE) added/subtracted by the thermostat.

Separate issue: I agree it would be good if the various fix thermostat doc pages mentioned how to output the quantity they conserve, so it can be monitored.  @athomps is better to comment on that.

sjplimp:
Re: the comments about fix shake losing thermostat energy.  I'm not sure if that is still true with the current version of SHAKE or RATTLE.  It would be a good Q for Peter Wirnsberger, who refactored fix shake in the process of adding fix rattle and fix ehex.

athomps:
You mean like this:

```
The fix_modify energy option is supported by these fixes to add the energy change induced by Nose/Hoover thermostatting and barostatting to the system’s potential energy as part of thermodynamic output.
```
On a related note, is it to late to request that this energy be added to total energy, rather than potential energy?

athomps (approved):
Looks good.

PabloPiaggi:
Perhaps before merging this we can look into the issue with fix shake? If we find that there is not a problem anymore we can remove the corresponding line of the documentation. What do you think?

On the other hand, including information about the conserved quantities in the documentation of other fixes can be left for a different PR. Not all fixes will have associated conserved quantities and therefore it has to be done with caution. Furthermore, one would probably have to check for each fix whether the total energy + the energy of the fix is the right quantity.

akohlmey:
@athomps @sjplimp I have to agree with @PabloPiaggi. Also for me it is very confusing to have an approval and at the same time suggestions for changes. The proper way to handle this would be to add a review that asks for changes and then I as the assignee would re-request a review (and possible approval) once I see that some conclusion has been reached and the changes are implemented of good reasons presented for not implementing them.

akohlmey:
> You mean like this:
> 
> ```
> The fix_modify energy option is supported by these fixes to add the energy change induced by Nose/Hoover thermostatting and barostatting to the system’s potential energy as part of thermodynamic output.
> ```

The fix_modify energy option does not apply here. It would need to be added. With `fix temp/csvr` you can access the accumulated energy change only through accessing the scalar compute of the fix.

The fix energy can only be added to the potential energy because it is part of `compute pe` and the total energy is computed on the fly from summing the results from the `compute pe` and recovering the kinetic energy from `compute temp` using the instances held by the Thermo class.

sjplimp:
ok, I have mixed several things together.  For this PR, it is good to go if you remove the 2nd change (no kinetic, just energy), same as other thermostat fixes.  Do that and I will approve it.

@PabloPiaggi I think you should submit another PR to add support for fix_modify energy to your 2 fixes, similar to other thermostat fixes.  This affects both source code and doc page.

If we find out something useful about thermostatting when using fix shake (rattle, etc), we can do another PR that adds text to multiple doc pages.  @PabloPiaggi if you send me an email with your email, I will CC you on the
email I send to Peter W about this.

If @athomps  agrees we should document how to output conserved quantities from the relevant thermostat fixes, we can do another PR for that.

And finally, as Axel said, adding the thermostat energy to etotal instead of pe is not simple, so I think it's fine as is.

sjplimp:
@PabloPiaggi I now see your email at the top of the PR, so no need to send it

PabloPiaggi:
> ok, I have mixed several things together. For this PR, it is good to go if you remove the 2nd change (no kinetic, just energy), same as other thermostat fixes. Do that and I will approve it.
> 

Done!

> @PabloPiaggi I think you should submit another PR to add support for fix_modify energy to your 2 fixes, similar to other thermostat fixes. This affects both source code and doc page.
> 

I am actually satisfied with the current state of affairs. It is easy to output this energy using ``thermo_style`` and then add it to the total energy.

> If we find out something useful about thermostatting when using fix shake (rattle, etc), we can do another PR that adds text to multiple doc pages. @PabloPiaggi if you send me an email with your email, I will CC you on the
> email I send to Peter W about this.
> 

Yes, please keep me in the loop. Thank you. The manual pages for these thermostats now say that they are incompatible with ``fix shake``. If this is true, the limitation probably extends to other thermostats. If it's not, I can open another PR to remove the line that mentions the incompatibility.

> If @athomps agrees we should document how to output conserved quantities from the relevant thermostat fixes, we can do another PR for that.
> 
> And finally, as Axel said, adding the thermostat energy to etotal instead of pe is not simple, so I think it's fine as is.

@sjplimp thank you for reviewing this.

athomps:
> > @PabloPiaggi I think you should submit another PR to add support for fix_modify energy to your 2 fixes, similar to other thermostat fixes. This affects both source code and doc page.
> 
> I am actually satisfied with the current state of affairs. It is easy to output this energy using `thermo_style` and then add it to the total energy.

This is a good point. For fixes that don't apply a conservative force of some kind, silently adding the fix energy to the potential energy is not very useful.  For these, one can explicitly add the fix energy to the total energy, as @PabloPiaggi  suggests.  Of the fixes in base LAMMPS that currently have THERMO_ENERGY set, these ones apply a conservative force:
```
fix_addforce.cpp:
fix_efield.cpp:
fix_external.cpp:
fix_gravity.cpp:
fix_indent.cpp:
fix_orient_bcc.cpp:
fix_orient_fcc.cpp:
fix_restrain.cpp:
fix_spring.cpp:
fix_spring_chunk.cpp:
fix_spring_self.cpp:
fix_wall.cpp:
fix_wall_region.cpp:
```
and these do  not:

```
fix_langevin.cpp:
fix_msst.cpp:
fix_nh.cpp:
fix_temp_berendsen.cpp:
fix_temp_csld.cpp:
fix_temp_csvr.cpp:
fix_temp_rescale.cpp:
``` 

Maybe we should turn off THERMO_ENERGY for the latter.

sjplimp:
I think you are interpreting the code wrong.  No fix has the fix_modify energy flag set by default.
The internal name of the flag is thermo_energy.  The ones you have listed have THERMO_ENERGY set
in the mask, but that is only activated if fix_modify energy is set to yes.  So this means fix temp/csvr and csld
already support adding their energy to the potential energy and thus the total energy of thermo output.
Same as all the other thermostat fixes.

athomps:
Yes, that is correct. I am suggesting that we not set THERMO_ENERGY for the second set of fixes, including  fix temp/csvr and csld.

sjplimp:
Sorry @athomps I'm still confused.  I was responding to "silently adding the fix energy to the potential energy is not very useful", by saying that no fix does that silently.  Only if the user explicitly requests it.

Are you saying that rather than allowing fix_modify energy yes to automate the addition of a thermostat energy
to the total energy (via the potential energy), you would prefer the user create a variable that adds etotal to the fix scalar, and then output that variable with the thermo output.  Leaving the thermo values output by pe and etotal unchanged?  Perhaps b/c that is more accurate conceptually?  Or maybe it is confusing to add a cumulative thermostat energy to the instantaneous potential energy?

Is the conserved quantity (for some thermostats) the instantaneous(ke + pe) + cumulative(thermostat energy)?  If that's the case, what we're doing now is simply a short-cut to monitor that quantity.  But I'm open to changing it.  Regardless I think we should add a line to the thermostat doc pages explaining how to monitor it.

athomps:
Yes, that's it.  The automatic addition of potential energy (and virial) contribution is good for fixes that apply a force, but for things like thermostats it is not useful, since most people don't need it, it can be confusing, and it can be accessed using the scalar value provided by those fixes.

PabloPiaggi:
I agree with the discussion above. In my opinion the most important thing is to explain users which thermostats and barostats have conserved quantities, and how to compute them in terms of things that Lammps can output. There is not much difference between having directly the conserved quantity or having to sum two things, as long as the manual is clear.

akohlmey:
> Yes, that's it. The automatic addition of potential energy (and virial) contribution is good for fixes that apply a force, but for things like thermostats it is not useful, since most people don't need it, it can be confusing, and it can be accessed using the scalar value provided by those fixes.

Just did some research in the commit history. You are asking to change something that has been in LAMMPS since the beginning of the commit history for the C++ version in 2006. Since this behavior must be explicitly requested through `fix_modify` and since it is extremely convenient to use, I honestly don't see the harm in keeping it and I don't see who would be confused that would not be confused just as much to implement the same through the more complex route by using an equal style variable to add this to the total energy manually. If you start questioning these kind of things, there are many more features and functionalities in LAMMPS that need questioning and reconsideration.  As far as I recall the sentiment in other, similar cases has always been: it is described in the manual, so it is ok and the user will have to make an informed choice. If I am not mistaking it would be a first for us to do something that essentially says "we don't trust you will understand it, so we take away the option".

athomps:
@akohlmey yes you are right. It ain’t broke, it’s documented, so we should not fix it.  But I do recommend for thermo/barostats using explicit equal style variable instead of fix_modify energy.

sjplimp:
I can see both sides of this discussion.  I think we should remove the THERMO_ENERGY mask flag from the thermostat fixes (in a new PR, not this one).  And document in each relevant thermostat fix how to calculate the conserved quantity and output it if desired.  My reason is that, while convenient, I think auto-adding a cumulative energy to the instantaneous pe or etotal violates the principle that those thermo keywords (and similar ones) represent instantaneous quantities.  Having the user add an extra line (variable) to the input script is not a big deal.  A middle ground would be add a new mask flag, like CONSERVE_ENERGY, to those fixes, and a new thermo keyword like econserve, which adds etotal to the scalar value produced by CONSERVE_ENERGY fixes.

athomps:
Yes, if we remove THERMO_ENERGY and add instructions on conserved quantity for each fix, a few scripts will break, but it will be transparent how to fix them.

PabloPiaggi:
> A middle ground would be add a new mask flag, like CONSERVE_ENERGY, to those fixes, and a new thermo keyword like econserve, which adds etotal to the scalar value produced by CONSERVE_ENERGY fixes.

I am not pretending to have a voice in this discussion, but this seems a very elegant solution. From the point of view of the user it would be perfectly clear that econserve is the right quantity to print and analyze.

akohlmey:
If the treatment of `THERMO_ENERGY` should be revised then it needs to be revised properly and the changes should be checked for all kinds of side effects and we should try to make things more consistent.

- First off, `THERMO_ENERGY` is unusual as a `Fix::setmask()` property, as it is not enabling a calling a function in the fix. The only other flag of this kind is `FIX_CONST_LAST` which doesn't seem to be used anywhere (anymore?). Thus I think adding a new flag to the list of `Fix::setmask()` properties would just make things more complex and inconsistent from the implementation and maintenance point of view.
- The current expected procedure is the following:
  * The fix sets the `THERMO_ENERGY` flag in `Fix::setmask()` and if the command `thermo_modify <fix-ID> energy yes` is given, it is checked for and then the variable `Fix::thermo_energy` is changed from its default value (0) to 1, with "no" is is changed back to 0, and if the flag is not set buy "yes" is given, an error is triggered.
  * When calling `Thermo::compute_pe()` the Thermo class triggers running `ComputePE::compute_scalar()` and then that will query fixes for their `Fix::compute_scalar()` output to be added to the potential energy.
  * There are some notable exceptions to this procedure. `fix pimd` and `fix rhok` set `thermo_energy = 1` by themselves and `fix msst`, `fix nphug`, and `fix qbmsst` query the value of `Fix::thermo_energy` and depending on that make changes to an internal variable `epot`. and `fix atc` sets `Fix::thermo_energy` depending on the return value of an internal function of the AtC library.
- If making this property (accumulated amount of energy exchanged with the heat bath) accessible as part of the thermo class, the most straightforward way would be to have a new thermo keyword for it, say `ebath` or `exchange` or `ethermo` or `ethermostat` or something like that. Since this is no longer a feature that would be optionally added to another, there is not much of a reason to require a setmask flag unless a new virtual function would be added to the Fix class. Instead it would be sufficient to have a variable in the fix class that is initialized to zero by the `Fix` base class constructor and to which fixes then add the corresponding energy change (or not) and then the evaluation of the thermo keyword would simply loop over all fixes to get the sum of the energy exchanged with the heat bath(s). this approach would keep things more consistent, adds the least amount of code, requires the fewest changes to inputs, and would only cause overhead when the keyword is actually referenced.

As mentioned by others, this should definitely be a separate PR, but also this should then also aim to reduce or remove the inconsistencies in the `Fix::setmask()` flag treatment and review the use cases for the fixes listed above that bypass the expected flow of control and possibly update their implementation. For the valid use cases of `THERMO_ENERGY` I would suggest to consider changing the name of the flag set by `fix_modify <fix-ID> energy yes/no` to `Fix::thermo_energy_flag` and rather than triggering a call to `Fix::compute_scalar()` it would call a new virtual function `Fix::thermo_energy()` which would return 0.0 in its default version. That would make things more consistent and also free up the `Fix::compute_scalar()` function for other uses. Since this change will break scripts, they should be broken for the right reason and we should try to change legacy features (like this seems to be) into something that is more consistent with the rest of the code.

PR #2517, opened 2020-12-11
Fix Kokkos performance regression for small systems

stanmoore1:
**Summary**

90ff2eb6c97638be07f8bbddbbfa6daec081cc40 caused a performance regression in Kokkos for small systems running on the GPU when using `comm no` and `fix nve/kk/host`. This PR fixes the issue.

**Related Issue(s)**

None

**Author(s)**

Stan Moore (SNL), and Evan Weinberg (NVIDIA) @weinbe2 who helped track down the issue

**Licensing**

**Backward Compatibility**

Yes


akohlmey:
@stanmoore1 or @weinbe2 can you please explain a bit more where exactly the performance is lost and why this change fixes it?
Unless I am missing something, this seems to be overriding a function with an identical function.

stanmoore1:
Kokkos uses `DELTA = 10` with an exponential grow(), vanilla LAMMPS uses `DELTA = 16384` with a linear grow(). Using `DELTA = 16384` causes a bunch of unnecessary data movement for Kokkos for small systems.

src/KOKKOS/atom_vec_kokkos.cpp
```diff
@@ -16,8 +16,11 @@
 #include "atom_kokkos.h"
 #include "atom_masks.h"
 #include "comm_kokkos.h"
+#include "error.h"
 #include "domain.h"
 
+#define DELTA 10
```
stanmoore1, line 22: @akohlmey this is the key line.

akohlmey: thanks. can we then please add a comment to the `roundup()` function explaining this?

we should also replace the `#define DELTA` with a `const int DELTA` inside the `roundup()` function, same for `atom_vec.cpp`. The compiler will ultimately create the same code, but we gain more readable code and type checking.

stanmoore1: >we should also replace the #define DELTA with a const int DELTA inside the roundup() function, same for atom_vec.cpp

This type of #define is used everywhere in vanilla LAMMPS too. I suppose the idea for including it up at the top of the file was so you only had to change it in one place?

yafshar: Now that lammps support `c++11`, why don't you replace these macros with `constexpr` e.g. `static constexpr std::size_t kDelta = 10;` at the top?

akohlmey: > > we should also replace the #define DELTA with a const int DELTA inside the roundup() function, same for atom_vec.cpp
> 
> This type of #define is used everywhere in vanilla LAMMPS too. I suppose the idea for including it up at the top of the file was so you only had to change it in one place?

that would make sense if it actually *would* be used in multiple places throughout the entire code and then consistently so by putting the define into a commonly used header.

but in this case it is quite confusing or else I would not have asked the question starting this discussion in the first place.

stanmoore1: Understood, I just like to keep the Kokkos version as similar to vanilla as possible, helps when porting changes  to Kokkos.

akohlmey: at any rate, it would be sufficient for me to merge this pull request to have the comment to the roundup() function(s) added explaining why the KOKKOS version needs to be re-implement.

akohlmey: > Understood, I just like to keep the Kokkos version as similar to vanilla as possible, helps when porting changes to Kokkos.

It was my understanding to change both functions/files. But don't worry about it. I'll deal with that elsewhere.

stanmoore1: OK I'll put in some comments

stanmoore1: @akohlmey I added the comments.

akohlmey:
thanks. I added a change to use constexpr instead of a define for DELTA (and DELTA_BONUS) and move the definition closer to its use. that should reduce the risk of me or others getting confused in the future.

stanmoore1:
Every Kokkos atom_vec style also redefines DELTA, not sure if we should change them all now?

stanmoore1:
Or if there is some way to collapse them all into a single define?

akohlmey:
> Or if there is some way to collapse them all into a single define?

this seems to be because the KOKKOS versions of AtomVec have not been refactored the same way as the core LAMMPS versions. I would leave it "as is" as it should go away once you get to refactoring this part of the code.

if you want to synchronize them, I would rename the define (e.g. KOKKOS_ATOMVEC_DELTA) and put it into a shared header file (by preference an existing one). having a different name for this constant would be an alternate approach to emphasize that this is meant to be a different constant to the one in the AtomVec base class.

stanmoore1:
@akohlmey I synchronized DELTA across Kokkos `atom_vecs`, should be ready to merge.

akohlmey (approved):

PR #2521, opened 2020-12-15
Turn LAMMPS Python module into Python package

### LINKED ISSUE #2529: [BUG] Need to add --embed flag to python-config for python 3.8 and above to allow building with python package
**Summary**

In python 3.8 and above, when `python-config --ldflags` is used without the `--embed` flag, the output doesn't contain `-lpython3.8 `, which makes building LAMMPS fail at the final linking stage. Only tested with classic make.

Latest master version, python 3.8.

**Expected Behavior**

LAMMPS should build correctly regardless of python version.

**Actual Behavior**

LAMMPS fails at the linking stage with a lot of "undefined reference" errors such as:
> /usr/lib/gcc/x86_64-pc-linux-gnu/9.3.0/../../../../x86_64-pc-linux-gnu/bin/ld: ./liblammps_serial.a(pair_python.o): in function `LAMMPS_NS::PairPython::~PairPython()':
> /usr/include/python3.8/object.h:478: undefined reference to `_Py_Dealloc'

**Steps to Reproduce**

1. Have a system, where the default python version is at least 3.8.
2. Enable python package with `make yes-python`.
3. Build LAMMPS: `make serial`.

This change is described in the python release notes[1]:
> To embed Python into an application, a new --embed option must be passed to python3-config --libs --embed to get -lpython3.8 (link the application to libpython). To support both 3.8 and older, try python3-config --libs --embed first and fallback to python3-config --libs (without --embed) if the previous command fails.

[1] https://docs.python.org/3.8/whatsnew/3.8.html

rbberger:
**Summary**

* turns the LAMMPS Python module into a package
* splits up `lammps.py` into multiple files
* Updates legacy `install.py` to take path to package folder via `-p` instead of module path `-m`
* Added a regular `setup.py` for Python-only installation
* Simplify CMake installation by using the setup.py build & install
* Fixes a bug due to changes in `python-config` in newer version
* Updates documentaiton and comments that mentioned `lammps.py`

**Related Issue(s)**

Fixes #2529 
#2508 

**Author(s)**

Richard Berger (Temple U) @rbberger 

**Licensing**

**Backward Compatibility**

yes, all symbols are still visible by only importing `lammps`.

**Implementation Notes**

The `lammps.py` module has grown significantly over time. This PR splits this large file into multiple smaller modules that live inside of a new `lammps` package. By publishing the symbols of all of these submodules in `__init__.py`, the import semantics remain the same.

The CMake and legacy build systems have been updated to install the package instead of the module.  The main difference between the two is that the `install-python` install target copies the shared library into the site-packages folder, instead of a `lib/` folder. To maintain this functionality with the package, the .so file is now copied into the package folder `$SITE_PACKAGE_FOLDER/lammps/liblammps.so`. For CMake, the `install-python` target does the same.

For the CMake `install` target, the LAMMPS Python package will be installed into the installation prefix.
Note that one of the things to watch out for is that this installation maintains a *clean* source directory to avoid collisions during multiple parallel out-of-source builds. This is why distutils was used, together with the `-b` option to set the build directory inside of the CMake binary directory. The "modern" way of doing this seems to be using setuptools and wheels, but I have yet to find out how to do out-of-source builds with this. I also didn't want to introduce a `pip` requirement.

**Post Submission Checklist**


sjplimp (approved):

cmake/CMakeLists.txt
```diff
@@ -691,11 +691,8 @@ if(BUILD_SHARED_LIBS OR PKG_PYTHON)
     find_package(Python COMPONENTS Interpreter)
   endif()
   if (Python_EXECUTABLE)
-    execute_process(COMMAND ${Python_EXECUTABLE}
-      -c "import distutils.sysconfig as cg; print(cg.get_python_lib(1,0,prefix='${CMAKE_INSTALL_PREFIX}'))"
-      OUTPUT_VARIABLE PYTHON_DEFAULT_INSTDIR OUTPUT_STRIP_TRAILING_WHITESPACE)
-    set(PYTHON_INSTDIR ${PYTHON_DEFAULT_INSTDIR} CACHE PATH "Installation folder for LAMMPS Python module")
-    install(FILES ${LAMMPS_PYTHON_DIR}/lammps.py DESTINATION ${PYTHON_INSTDIR})
+    file(MAKE_DIRECTORY ${CMAKE_BINARY_DIR}/python)
+    install(CODE "execute_process(COMMAND ${Python_EXECUTABLE} setup.py build -b ${CMAKE_BINARY_DIR}/python install --prefix=${CMAKE_INSTALL_PREFIX} WORKING_DIRECTORY ${LAMMPS_PYTHON_DIR})")
```
junghans, line 695: Most packagers rely on the injection of `DESTDIR` to change install path (but leaving rpaths and all the other magic the same), so I am wondering if this mechanism will work here.

junghans: Ok, we have a problem here:
```
cmake ../cmake/ -DPKG_PYTHON=ON -DCMAKE_INSTALL_PREFIX=/usr
make -j3
make install DESTDIR=$PWD/test
```
fails with:
```
creating /usr/lib/python3.9/site-packages/lammps
error: could not create '/usr/lib/python3.9/site-packages/lammps': Permission denied
```

So `DESTDIR` is missing.

rbberger: @junghans Can you show me one of those injections?

junghans:
```suggestion
    install(CODE "execute_process(COMMAND ${Python_EXECUTABLE} setup.py build -b ${CMAKE_BINARY_DIR}/python install --prefix=${CMAKE_INSTALL_PREFIX} --root $DESTDIR WORKING_DIRECTORY ${LAMMPS_PYTHON_DIR})")
```
works for make at least.

junghans: Works with `DESTDIR=$PWD/test ninja install` as well.

junghans:
```suggestion
    install(CODE "execute_process(COMMAND ${Python_EXECUTABLE} setup.py build -b ${CMAKE_BINARY_DIR}/python install --prefix=${CMAKE_INSTALL_PREFIX} --root \$ENV{DESTDIR}/ WORKING_DIRECTORY ${LAMMPS_PYTHON_DIR})")
```
seems to work in all case, w/o DESTDIR and make/ninja.

Issue #2002, opened 2020-04-16
[Feature Request] Add support for symbolic atom, bond, angle, dihedral and improper types


akohlmey:
**Summary**

Right now, force field types in LAMMPS are numbers starting from 1. Allowing symbolic (text) based atom types would make a lot of tasks easier like merging data from multiple simulations, supporting reading of existing topology/parameter file formats from CHARMM or Amber, and writing complex inputs. On the other hand, some of the flexibility of LAMMPS stems from the size of data structures being flexible, but then locked in when the simulation cell is created. This causes some conflicts for which some solution has to be implemented that is acceptable for when symbolic atom types are used.

**Detailed Description**

Given the pervasive nature of such a feature, this has to be implemented into the core of LAMMPS. Given the large body of existing work and tools, it also has to be implemented in a way to ensure backward compatibility. And it would need to be implemented incrementally by first adding the facility to manage the symbol to type and reverse mapping and then gradually adapt styles and other code to utilize it.

A key goal would be to have the ability to directly read topology/parameter files from either CHARMM or Amber (parameter and psf files and parmtop files, respectively) or read a file format that is available and well supported for which a well supported converter tool exists for conversion without loss of information (most tools lose some information though). In some way it should be possible to auto-generate types for bonded interactions from atomic types. There also should be a command to populate the force field type database from LAMMPS input script.

The major constraint to consider is that one cannot change the number of types after the box is defined (outside of completely deleting the simulation instance with "clear"). A strategy to deal with that is to reserve more atom types than actually needed *and* have a "default" or "disabled" set of force field parameters (a NULL atom or bond types), so that the checking for unset atom types will not fail due to missing input of parameters.

One strategy would be to allow setting symbolic types *BEFORE* the box is defined. e.g. through a command like `species atom add OW` that would then later allow to either explicitly associate a numerical atom type with a symbolic type (`species atom set type OW 1`) or do this automatically. When creating the box, instead of the number of atom types the keyword AUTO could be used
and then the number would be the number of entries in the table of symbolic types.
We allow have other things associated with those tables, e.g. force field parameters (`species atom coeff OW lj 0.15535 3.166`). In most force fields, the mixed terms are derived from mixing rules, but that could be overridden as well (e.g. to support NBFIX in CHARMM) with a specific command (`species atom mixed OW OH lj 0.0 3.0`)

The equivalent would be done for bonds/angles/etc. that would allow to read a CHARMM parameter file (except for cross terms, but that is a special case anyway).

[...]

PR #2531, opened 2020-12-17
Toward Type Labels

jrgissing:
**Summary**

Add general support for mapping each numeric type to an alphanumeric 'label' for atom types, bond types, etc.

Closes: #2002 

End Goal:

(a) Ability for users to define maps, and optionally use a label type
anywhere a numeric type is used in LAMMPS input (input script, data
file, molecule file) or output (dump files, restart files).  For input, using
a mix of numeric and label types is fine.

(b) Make it simpler for users to define systems using multiple data
and/or molecule files.  B/c use of label types makes it simpler to
avoid and resolve numeric type conflicts across multiple files.

- [x] begin general label map class
- [x] add new 'type label' sections to data file
- [x] add write_data support
- [x] support for multiple data files
  - [x] function to merge label maps
  - [x] function to map one label map to another
  - [x] support for selected atom type sections
  - [x] support for remaining sections and bond, etc. types 
- [x] support for molecule files 
- [x] add labelmap command (see attached examples)
- [x] support for restart files
- [x] add labelmap functions to variable command
- [x] direct support for pair, bond, etc. coeff commands
- [x] support for multiple maps
  - [x] roll back support for multiple maps
- [x] write_data nolabelmap option
- [x] docs draft
- [x] backtest/try to break
- [x] option to directly replace numeric types with labels in data file read/write
- [x] convert internal label maps to C++ map
- [x] error/warning docs (post-review)
- [x] propagate enhancements for Atoms section of data file to all label map types and data file sections.
- [x] implement unit tests for all supported features, molecular systems, reading multiple data files, molecule files
- [x] document type label sections in data files in read_data docs
- [x] document what are allowed characters for type labels
- [x] add convenience function that identifies valid labels
- [x] document how to handle special characters in type labels
- [x] update variable function to look up types for labels
- [ ] explore options to have a variable function returning the label string for the numeric type

**Related Issue(s)**

removed I <= J requirement for pair_coeff, so that type labels can be used without knowing their numeric counterpart

**Author(s)**

Jake Gissinger (NASA),  Yaser Afshar (UMN), Steve Plimpton (SNL), Axel Kohlmeyer (Temple U)

**Licensing**

**Backward Compatibility**

**Implementation Notes**

single-data-file simple test

[simple_test_labels.zip](https://github.com/lammps/lammps/files/5710535/simple_test_labels.zip)

multi-data-file simple test (support for selected atom-type sections):

[labels_test.zip](https://github.com/lammps/lammps/files/5790148/labels_test.zip)

multi-data-file test (includes bond, angles, dihedrals, impropers):

[multidata_example.zip](https://github.com/lammps/lammps/files/5798345/multidata_example.zip)

multi-data-file-and-molecule

[multidata_and_molecule.zip](https://github.com/lammps/lammps/files/5805284/multidata_and_molecule.zip)

multiple data files, molecule template, labelmap command
type labels are added/changed by second data file, used by molecule template, and modified by labelmap command
reflected in final write_data

[add_command_example.zip](https://github.com/lammps/lammps/files/5823735/add_command_example.zip)

in addition to above, add simple example of using 'map/assign' option of 'labelmap' command

[add_map_assign.zip](https://github.com/lammps/lammps/files/7101909/add_map_assign.zip)

in addition to above, *update to new keywords* (map/assign -> mapID) and add 'mass command' example

[update_keywords+mass_command.zip](https://github.com/lammps/lammps/files/7707013/update_keywords%2Bmass_command.zip)


jrgissing:
@sjplimp it is already clear that this is much cleaner approach than previous attempts. Would you like to take a look at these initial changes?

doc/src/labelmap.rst
```diff
@@ -0,0 +1,111 @@
...
+
+.. note::
+
+   Type labels cannot start with a number.
+
+The label map of only one type of interaction (atom, bond, etc.) may
+be modified by a given *labelmap* command. Any number of
+numeric-type/type-label pairs may follow. If a type label already
+exists for a given numeric type, it will be overwritten. Assigning the
+same type label to multiple numeric types is permitted; how this
```
yafshar, line 60: @jrgissing Is this correct? `Assigning the same type label to multiple numeric types is permitted;`
I thought it should be the other way, which is what other codes use. We have C1, C2, C3, ... which are all the same C type.

yafshar: @jrgissing if you assign the same label to multiple numeric types, how do you know which one is which in the current implementation?

jrgissing: assigning the same label to multiple types was permitted as part of our effort to keep type labels as general as possible. currently, the first instance of a label is returned (i.e. corresponding to the lowest numeric type). this behavior should be stated in the docs, for example when using 'labelmap functions' of the variable command.

yafshar:
@jrgissing can you update this PR with the latest master?

jrgissing:
> @jrgissing can you update this PR with the latest master?

done!

doc/src/labelmap.rst
jrgissing: update: type labels now required to be unique

sjplimp:
@jrgissing HI Jake - I think this is nearly ready to approve.  I pushed some doc page edits which you should look over, esp labelmap.rst and Howto_type_labels.rst.  The latter is a new howto section which I think is the right place to expalin the motivation and use of type labels. Esp going forward as command like fix bond/react and pair kim start to use them.  Please grep for JAKE in the doc files to see when I had Qs for you.  I made the other doc pages refer to the new Howto.  In the Howto.rst, do you think it should be under Settings or Force Fields?  Also, I made 2 minor syntax changes to commands in the doc pages.  For labelmap, mapID name for the optional keyword.  For write_data, types numeric/labels for one of the optional keywords you added.  I think this syntax is more LAMMPS-consistent.  I will next give the source code another quick look over.  I think it was basically fine before.

sjplimp:
@jrgissing Also, one other suggestion. I think there should be a way in the data file sections for Type Labels to not define a label for a particular numeric type.  Need to still have Ntype entries, but maybe entries like "3 none" should be allowed?

jrgissing:
@sjplimp those edits look good, I'll add some specific comments/replies soon

> In the Howto.rst, do you think it should be under Settings or Force Fields?

I think putting it under Settings makes sense.

> Also, I made 2 minor syntax changes to commands in the doc pages. For labelmap, mapID name for the optional keyword. For write_data, types numeric/labels for one of the optional keywords you added.

Sounds good, I will update the source code accordingly.

> I think there should be a way in the data file sections for Type Labels to not define a label for a particular numeric type. Need to still have Ntype entries, but maybe entries like "3 none" should be allowed?

Okay. This would break my current check/requirement that the default label map is 'complete and unique' before writing it to data/restart files, and of course involves reserving a label string to signify that there is no label string (potentially confusing?). If you think incomplete maps are necessary, what about just allowing there to be no label, such as:

Atom Type Labels

1 C
2 
3 H

This way might also discourage users from trying to use 'none' in other places that actually need to read a specific type.

sjplimp:
@jrgissing I thought incomplete maps were already allowed with the labelmap command?  So I was just suggesting also allowing the same in a data file.  Your syntax for "2" instead of "2 none" is a good idea.

sjplimp:
@jrgissing Looked over the source code.  Looks generally good. 2 comments:
(1) In the doc pages I edited I thought a labelmap (with an ID) was for a single type, e.g. one labelmap for atom types,
another for bond types, etc.  I see from label_map.h this is not the case.  Not sure which is better.  But if we stick with the code as now, then the doc pages need to explain that differently.
(2) I thought you would store the label map as an actual C++ map, and not a vector of strings.
I'm thinking if someone has labels for 50 atoms types (bio problem) and reads a million-atom data file that
looking up a million labels by scaning a 50-length vec is inefficient?  I.e. a true map would be more better.

yafshar:
> @jrgissing (2) I thought you would store the label map as an actual C++ map, and not a vector of strings. I'm thinking if someone has labels for 50 atoms types (bio problem) and reads a million-atom data file that looking up a million labels by scaning a 50-length vec is inefficient? I.e. a true map would be better?

@sjplimp the vector of strings is implemented for the case where we had a 1 to many mapping (user can have the same label for multiple int types) based on initial discussion with kim. I have implemented and optimized the structure with unordered_map, just checking and testing a bit more and update the PR accordingly.

doc/src/pair_coeff.rst
```diff
@@ -34,20 +34,28 @@ atom types.  The number and meaning of the coefficients depends on the
...
-(inclusive).  A middle asterisk means all types from m to n
-(inclusive).  Note that only type pairs with I <= J are considered; if
-asterisks imply type pairs where J < I, they are ignored.
+I and J can be specified in one of several ways.  Explicit numeric
+values can be used for each, as in the first example above.  Or, one
+or both of the types in the I,J pair can be a type label, which is an
+alphanumeric string defined by the :doc:`labelmap <labelmap>` command
+or in a section of a data file read by the :doc:`read_data
+<read_data>` command, and which converts internally to a numeric type.
+In either case I <= J is required.  Interally, LAMMPS will set
```
jrgissing, line 43: @sjplimp I removed I <= J requirement for pair_coeff, so that type labels can be used without knowing their numeric counterpart

jrgissing:
@sjplimp I am coming across some possible ambiguities while thinking about incomplete maps, especially when reading multiple files. For example, if the first data file has this label map

1 C
2 H

and the second file has this (incomplete) map

1 H
2

then which type would the missing one map to? It might be safest and less confusing to require that all types are assigned a type label, if any are, for a given type-kind (this is current behavior).

jrgissing:
> @jrgissing I thought incomplete maps were already allowed with the labelmap command? So I was just suggesting also allowing the same in a data file. Your syntax for "2" instead of "2 none" is a good idea.

@sjplimp the labelmap command currently allows incomplete maps mostly because I was not sure when/where to check if the map is complete, in that case. however, whenever potential issues could arise (currently only when reading/writing files), an error is thrown if the map is not complete

sjplimp:
@jrgissing In this initial patch release if you want to require complete maps, that is fine.  And think about incomplete as a future enhancement.  In general I do think it is onerous on the user to require complete maps, e.g. if you have 50 atom types in a bio problem.  In your 2 data file example above, I don't understand which label you are applying to type 1 (C or H), much less type 2.
I thought lists of labels were local to the file, but now I see there is a conflict when you try to merge two of them into a label map that persists beyond reading the file.

jrgissing:
> @jrgissing In this initial patch release if you want to require complete maps, that is fine. And think about incomplete as a future enhancement. In general I do think it is onerous on the user to require complete maps, e.g. if you have 50 atom types in a bio problem. 

OK sounds good, I will update some phrasing in the docs accordingly. Then I think we should be about ready for the initial patch release

sjplimp:
One thing to clarify in the docs is if all maps need to be complete (e.g. before a run occurs), or only the ones in data files (not the ones specified by the labelmap command).  If the former, then the code should probably check for completeness in the init() or setup() of a run.

jrgissing:
currently all maps do not need to be complete before a run, perhaps mostly because no commands really use them yet

yafshar:
@jrgissing Can you also update the `mass command` to accept type label as well? I think it will be very convenient to have that, currently, it is not easy to have a label map and assign mass by numeric type. 
I am starting to write unit tests for the PR.

jrgissing:
> @jrgissing Can you also update the `mass command` to accept type label as well? I think it will be very convenient to have that, currently, it is not easy to have a label map and assign mass by numeric type. I am starting to write unit tests for the PR.

OK, done

sjplimp (approved):

sjplimp:
@akohlmey Jake says this is ready to be considered for merging

jrgissing:
@akohlmey @sjplimp the recent updates to data file parsing (#3072) modifies some of the same code as this PR, so a few merge conflicts will have to be resolved for this one now

sjplimp:
@jrgissing can you resolve the conflicts?
@akohlmey this is ready to merge I believe - are we waiting on anything else?

jrgissing:
> @jrgissing can you resolve the conflicts?

@sjplimp yes, I should be able to get to it today

stukowski:
FYI: Upcoming OVITO release 3.7 will likely support reading & writing the extended data file format proposed in this pull request. [OVITO](https://gitlab.com/stuko/ovito) maintains label maps for atom/bond/angle/dihedral/improper types - like LAMMPS is going to. The visualization program uses type names to automatically assign display colors and radii according to chemical element tables or user presets.
Dev builds of OVITO with I/O support for the extended data format are available in case you want to try: 

* [ovito-basic-3.7.0-dev-HEAD-65be4b3-x86_64-cpp17.tar.xz](https://www.ovito.org/download/testing/ovito-basic-3.7.0-dev-HEAD-65be4b3-x86_64-cpp17.tar.xz)
* [ovito-basic-3.7.0-dev-HEAD-65be4b3-win64-cpp17.zip](https://www.ovito.org/download/testing/ovito-basic-3.7.0-dev-HEAD-65be4b3-win64-cpp17.zip)
* [ovito-basic-3.7.0-dev-HEAD-65be4b3-macos-intel-cpp17.dmg](https://www.ovito.org/download/testing/ovito-basic-3.7.0-dev-HEAD-65be4b3-macos-intel-cpp17.dmg)

jrgissing:
@stukowski thanks for the fast adoption of this extended format! Having correctly labeled, colored and sized atoms upon import is very convenient.

doc/src/labelmap.rst
```diff
@@ -40,7 +40,7 @@ As explained on the Howto page, valid type labels can contain any
 alphanumeric character, but cannot start with a number.  They can also
 contain standard characters such as square brackets "[" and "]", dash
 "-", underscore "_", plus "+" and equals "=" signs.  Note that type
-labels cannot contain the comment symbol '#'.
+labels must be put in quotation marks if they contain the comment symbol '#'.
```
jrgissing: relaxing this requirement may require significant changes to read_data and molecule template file parsing

akohlmey: I have not seen any code enforcing this condition. So I am not relaxing anything. So far the only complication is that when processing input commands, "#' starts a comment and thus quotation marks need to be use (and will be removed). That makes it more a problem of the input processing than an issue of the label map processing. There should probably be some check for what is a valid label (restricting it to 7-bit ASCII for portability).

src/variable.cpp
```diff
@@ -4383,6 +4385,67 @@ int Variable::special_function(char *word, char *contents, Tree **tree, Tree **t
...
+     dlabel(typelabel),ilabel(typelabel)
+------------------------------------------------------------------------- */
+
+int Variable::labelmap_function(char *word, char *contents, Tree **tree,
+                                Tree **treestack, int &ntreestack,
+                                double *argstack, int &nargstack, int ivar)
+{
+  // word not a match to any label map function
+
+  if (strcmp(word,"label") && strcmp(word,"blabel") &&
```
akohlmey: This is a unexpected choice for the function name. For a function "label()" I would expect that it returns the label to a numeric type, not the numeric type of a label. So it would be better to call it, for example, "label2type()".

doc/src/labelmap.rst
akohlmey: FWIW, the following input seems to be working as expected:
```
region box block -2 2 -2 2 -2 2
create_box 4 box

labelmap atom 1 C1 2 C2 3 C3 4 "C#"

create_atoms 1 single 0.0 0.0 0.0
create_atoms 2 single 0.0 0.0 0.0
create_atoms 3 single 0.0 0.0 0.0
create_atoms 4 single 0.0 0.0 0.0

mass C1 1.0
mass C2 2.0
mass C3 3.0
mass "C#" 4.0

pair_style zero 2.0
pair_coeff * *
pair_coeff C1 C2
pair_coeff C3 "C#"

write_data label.data

clear

pair_style zero 3.0
read_data label.data

print "Atom labels: $(label(C1)) $(label(C2)) $(label(C3)) $(label(C#))"
```

I'll clarify in the docs that the quotation marks are only needed in a context with "#" would be interpreted as starting a comment, e.g. in the input file.

jrgissing: type labels can also be used in the Atoms section of data files, directly replacing the numeric type (same with molecule templates). it looks like this is broken by # symbols, at the moment

src/variable.cpp
jrgissing: ok, makes sense. i think Steve suggested these names, so he may have a stronger opinion

doc/src/labelmap.rst
akohlmey: @jrgissing It seems that this breakage can be addressed with some rather small changes and without having to use quotes. See commit https://github.com/lammps/lammps/pull/2531/commits/d43051d07cde598c9289d7eab9643b24f367601e.
Do you recall to which other sections or parsers this would need to be propagated?
Adding support for quotes would be possible with using `utils::split_words()` instead of the Tokenizer class, but that is significantly slower and would slow down data file parsing, so I am hesitant to go this far.

In general, we should probably define a list of allowed characters (ideally limited to 7-bit ASCII, disallowing control characters and whitespace) and then set up a corresponding check against a regular expression with `utils::strmatch()`. So far, number, '#', and '*' are disallowed as first character, but by using quotes, the current code would also accept whitespace, which would be very difficult to support in data files.

jrgissing: nice! other sections include the Bonds, Angles, Dihedrals and Impropers sections of data files, and the Types, Bonds, Angles, Dihedrals and Impropers sections of molecule templates

also, while we are trying to break things, atom types in several different force fields contain the single or double-quote character, such as the type c3"

akohlmey: Supporting single/double quotes should be easier for anything using the Tokenizer class to parse lines of text into words.
In the input file those should just require using single/double quotes around the label.

However, this is something where we need some kind of well defined policy of what we want to allow or not, and suggestions for how to "encode" or "translate" unsupported characters etc. I have summarized my perspective on that in an email copied to all stakeholders. For something as fundamental as this feature, it is generally better to have a statement of the "we support X" kind rather than a "we do not support Y".

akohlmey: Single/double quotes seem to be working with proper quoting in the input file and without changes to the read_data parser. The only thing that doesn't seem to work are "immediate" variables within an already quoted string (as with print and fix print), but that can be handled by defining explicit variables.
```
region box block -2 2 -2 2 -2 2
create_box 4 box

labelmap atom 1 C1 2 "C2'" 3 'C3"' 4 "C#"

create_atoms 1 single 0.0 0.0 0.0
create_atoms 2 single 0.0 0.0 0.0
create_atoms 3 single 0.0 0.0 0.0
create_atoms 4 single 0.0 0.0 0.0

mass C1 1.0
mass "C2'" 2.0
mass 'C3"' 3.0
mass "C#" 4.0

pair_style zero 2.0
pair_coeff * *
pair_coeff C1 "C2'"
pair_coeff 'C3"' "C#"

write_data label.data.new

clear

pair_style zero 3.0
read_data label.data.new

variable t1 equal label(C1)
variable t2 equal "label(C2')"
variable t3 equal 'label(C3")'
variable t4 equal "label(C#)"

print "Atom labels: ${t1} ${t2} ${t3} ${t4}"
```

jrgissing: i wonder if this solution will run into problems for a bond type like C2'-C3"

support for escaping symbols might solve these issues and be easier to explain in the docs. but i don't have a sense of how reasonable it would be to implement

akohlmey: > i wonder if this solution will run into problems for a bond type like C2'-C3"

You could use triple quotes `labelmap atom 1 C1 2 """C2'-C2" """ 3 'C3"' 4 "C#"`, but a blank after 'C2"' is currently required, so that in the label definition that trailing blank must be removed.

akohlmey: Escaping would not be easy. Encoding might be an alternative. Like in HTML where a literal '<' becomes `&lt;`

akohlmey: This now works, too:
```
region box block -2 2 -2 2 -2 2
create_box 4 box

labelmap atom 1 C1 2 """C2'-C2" """ 3 'C3"' 4 """ C1"-C2'"""
create_atoms 1 single 0.0 0.0 0.0
create_atoms 2 single 0.0 0.0 0.0
create_atoms 3 single 0.0 0.0 0.0
create_atoms 4 single 0.0 0.0 0.0

mass C1 1.0
mass """C2'-C2" """ 2.0
mass 'C3"' 3.0
mass """ C1"-C2'""" 4.0

pair_style zero 2.0
pair_coeff * *
pair_coeff C1 """C2'-C2" """
pair_coeff 'C3"' """ C1"-C2'"""

write_data label.data.new

clear

pair_style zero 3.0
read_data label.data.new

variable t1 equal label(C1)
variable t2 equal """label(C2'-C2")"""
variable t3 equal 'label(C3")'
variable t4 equal """label(C1"-C2')"""

print "Atom labels: ${t1} ${t2} ${t3} ${t4}"
```

jrgissing: Looks good to me

src/variable.cpp
akohlmey: @sjplimp how about implementing only one function, label2type(), and then using atom,bond,angle,dihedral,improper as the first argument and the type label as the second, e.g.:
```
variable a1 equal label2type(atom,C1)
variable b1 equal label2type(bond,C1-C1)
```

doc/src/labelmap.rst
akohlmey: Using triple quotes with a blank before and after the label seems to be the most general way of quoting for all special cases so far. Just added a "labelmap write" command that should work similar to "write_coeffs" but for the label maps and used that there, too.

akohlmey:
@jrgissing I have now completed my review and refactoring of this pull request. I've discussed most open issues with Steve 1-on-1 and have implemented his suggestions, too. There are only a few minor items that still need testing like restart functionality and reading multiple data files. I don't expect that there will be a change in behavior, if changes to the code would be needed at all.

So please have a careful look at the code and add notes or questions, where you see problems or need explanations from me.
I hope to be able to merge this PR before Friday.

doc/src/variable.rst
```diff
@@ -962,6 +962,12 @@ types, bond types and so on. For the full list of available keywords
 *name* and their meaning, see the documentation for extract_setting()
 via the link in this paragraph.
 
+The label2type() function converts type labels into numeric types, using label
+maps created by the :doc:`labelmap <labelmap>` or :doc:`read_data <read_data>`
+commands.  The first argument must be label map kind (atom, bond, angle,
+dihedral, or improper) and the second argument the label.  Iwill return
```
jrgissing: looks like a typo on this line. also, for the previous sentence, I would suggest "The first argument is the label map kind (atom, bond, angle, dihedral, or improper), and the second argument is the type label."

src/atom.cpp
```diff
@@ -1884,33 +1884,63 @@ void Atom::allocate_type_arrays()
    called from reading of data file
    type_offset may be used when reading multiple data files
 ------------------------------------------------------------------------- */
+// clang-format on
 
-void Atom::set_mass(const char *file, int line, const char *str, int type_offset,
-                    int labelflag, int *ilabel)
+void Atom::set_mass(const char *file, int line, const char *str, int type_offset, int labelflag,
```
jrgissing, line 1861: support for the Masses data file section is nice, though it might open the door for people to ask why other per-type sections aren't also supported (and not sure if that would be worth the effort)

jrgissing: also, I don't think type label support for Masses section is documented

jrgissing: also, a more general point, some tools put the trailing '#' without a space after the number, for example:

Masses

   1  12.011150# cp

While I would consider this 'incorrect', previously lammps would process this fine, whereas now it throws an error

jrgissing: on a related note, statements like "Lines can have a trailing comment starting with ‘#’ that is ignored" in the read_data and molecule file docs are not strictly true anymore

jrgissing:
> So please have a careful look at the code and add notes or questions, where you see problems or need explanations from me. I hope to be able to merge this PR before Friday.

in general, looks great! I started taking a closer look and will post comments as I go

src/atom.cpp
jrgissing: > also, a more general point, some tools put the trailing '#' without a space after the number, for example:

okay, i remember where i've seen this. it is only in the Dihedral Coeffs of the files produced by msi2lmp. so, really a bug in that code, and currently a nonissue

doc/src/variable.rst
akohlmey: fixed

src/atom.cpp
akohlmey: > support for the Masses data file section is nice, though it might open the door for people to ask why other per-type sections aren't also supported (and not sure if that would be worth the effort)

That door was opened by submitting the pull request.  People *will* ask. Even for cases where it won't make sense or is close to impossible to implement. We have to be ready to have a substitution eventually for any place where a (numeric) type is used.

akohlmey: > > also, a more general point, some tools put the trailing '#' without a space after the number, for example:
> 
> okay, i remember where i've seen this. it is only in the Dihedral Coeffs of the files produced by msi2lmp. so, really a bug in that code, and currently a nonissue

confirmed and fixed

akohlmey (approved):

src/variable.cpp
```diff
@@ -4374,6 +4375,43 @@ int Variable::special_function(char *word, char *contents, Tree **tree, Tree **t
...
+
+  } else if (strcmp(word,"label2type") == 0) {
+    if (narg != 2) print_var_error(FLERR,"Invalid label2type() function syntax in variable formula",ivar);
+
+    if (!atom->labelmapflag)
+      print_var_error(FLERR,"Cannot use label2type() function without a labelmap",ivar);
+
+    int value;
+    std::string kind = args[0];
+    std::string typestr = utils::trim(utils::utf8_subst(args[1]));
```
jrgissing: one last comment/clarification. the refactored label2type() variable function technically changes what a valid type label is. previously, type labels could contain commas, but this is no longer supported

sjplimp (approved):

src/variable.cpp
akohlmey: Technically, the limitation only applies to the variable function. There are other limitations there, too, when using immediate variable evaluation with `$(label2type(kind,label))`. The case of '.' can be addressed, but there are also the characters '(' and ')' that cannot be used with the variable function. Well, you *can* use something like '(A)' or '(A-B)' as label with the label2type() function since that has a pair of parenthesis.

akohlmey: In all cases the label would still be a valid label, you just cannot use them with the variable function.

jrgissing:
ok, this PR looks good to me!

akohlmey:
> ok, this PR looks good to me!

Do you want me to wait until I have feedback from Ellad and Ryan?

jrgissing:
> Do you want me to wait until I have feedback from Ellad and Ryan?

we resolved all outstanding issues that I know of, but it wouldn't hurt to wait for a confirmation from them

yafshar:
This PR looks good to me. 
@jrgissing @akohlmey, is there any plan to add the provided examples to LAMMPS? or should we add them as unit tests?

yafshar (approved):

akohlmey:
> This PR looks good to me. @jrgissing @akohlmey, is there any plan to add the provided examples to LAMMPS? or should we add them as unit tests?

There are already significant number of unit tests included in this PR. In fact, that is how I noticed a bunch of issues and oversights or inconsistencies. This already covers a lot of the added functionality. There is always room for improvement though.

I have very deliberately not (yet) integrated any of the examples because a) there are still parts of LAMMPS that need to be adjusted to accept labels instead of numeric types and b) we should wait and see a bit how this is used to have representative examples in the examples tree.  As with the implementation, this can have a significant long-term impact and provide a blueprint for others, so they need to be particularly well designed and checked.

akohlmey:
@stukowski you may need to review type label support in Ovito after we have reworked this pull request some more and have converged to formats and interfaces that have a good chance to become permanent. Since this is a significant change and write support for symbolic types in data files in LAMMPS is still somewhat lacking (reading is more complete), there may be minor changes. The general intent is however at this point to stick to the format and strategies implemented here if possible.
This will also be needed for other features that are built on top of it.

I have started with adapting VMD for it today. Writing data files already works (which is easier since VMD already has all types as symbols internally). Ping me personally, if you want me to ship you some example data files to make sure those are mutually compatible. Similarly, I'd love to have a couple of test cases for data files with symbolic types written by OVITO to make sure that VMD stays compatible, if possible. I've changed the CGCMM marker in the first line to LABELMAP as indicator since some of the comments are different since adding the symbolic types as comments would be redundant.

PR #2532, opened 2020-12-17
Add more granular control of Kokkos pair and fix comm

stanmoore1:
**Summary**

Currently there is no way to run Kokkos pair and fix forward comm on the host. This PR adds more granular control of Kokkos pair and fix comm through Kokkos package options. 

**Related Issue(s)**

None.

**Author(s)**

Stan Moore (SNL)

**Licensing**

**Backward Compatibility**

This PR changes the Kokkos package `cuda/aware` keyword to `gpu/aware`, since Kokkos supports both CUDA and HIP.


src/KOKKOS/out.txt
```diff
@@ -0,0 +1,44 @@
+pair_buck_coul_cut_kokkos.cpp
```
rbberger, line 1: probably these two files shouldn't be here

stanmoore1: Yes good catch, I removed them.

stanmoore1:
I tested both `fix qeq` and `pair eam`, everything seems to be working.

stanmoore1:
This PR will almost certainly collide with #2533. Once #2533 is merged, I will fix the conflicts then it should be ready.

stanmoore1:
A few ReaxFF regression tests are failing...

stanmoore1:
Kokkos regression tests passed, should be ready for merge.

akohlmey:
@stanmoore1 how about accepting both, "cuda/aware" and "gpu/aware", but document only the latter as a way to preserve backward compatibility for a transition period? Some time in the future we could then remove the older keyword.

stanmoore1:
Yes that sounds great to me.

akohlmey:
> Yes that sounds great to me.

done

stanmoore1:
Thank you, this looks good to me.

akohlmey (approved):

PR #2533, opened 2020-12-18
Add pair/only option for accelerator styles

akohlmey:
**Summary**

This is an alternate, more minimal approach to implement a pair/only option for suffixes that makes the (primary) suffix applicable only to pair styles, which can be useful for GPU acceleration to better overlap with host computations.

**Related Issue(s)**

This tries to implement the same behavior as #2512 but in a way that is not limited to KOKKOS.
GPU package should already be supported, but it could be also applied to USER-INTEL and USER-OMP.

**Author(s)**

Stan Moore, SNL and Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issue.

**Post Submission Checklist**

### DIFF
--- a/doc/src/package.rst
+++ b/doc/src/package.rst
@@ -548,7 +548,7 @@ available (currently only possible with OpenMPI v2.0.0 or later), then
 the *cuda/aware* keyword is automatically set to *off* by default. When
 the *cuda/aware* keyword is set to *off* while any of the *comm*
 keywords are set to *device*\ , the value for these *comm* keywords will
-be automatically changed to *host*\ . This setting has no effect if not
+be automatically changed to *no*\ . This setting has no effect if not
 running on GPUs or if using only one MPI rank. CUDA-aware MPI is available
 for OpenMPI 1.8 (or later versions), Mvapich2 1.9 (or later) when the
 "MV2_USE_CUDA" environment variable is set to "1", CrayMPI, and IBM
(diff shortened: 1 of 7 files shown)


akohlmey:
@stanmoore1 here is a variant of the "pair/only" switch that can also be used with the GPU package and is a little less "disruptive". Please have a look, perhaps give it a try, and let me know what you think.

stanmoore1:
@akohlmey nice, this looks good to me.

akohlmey:
> @akohlmey nice, this looks good to me.

thanks. should we allow to revert the setting? like in https://github.com/lammps/lammps/pull/2533/commits/959f67962dcd1e3961afc0ece4236b3bc4292649 which still misses the reverting of the communication settings (which I currently don't know how to do). or should we just add something like this?

```
} else {
  if ((lmp->suffixp) && (me == 0))
    error->warning(FLERR,"Cannot turn 'pair/only' setting off after turning it on");
```

stanmoore1:
Ah, there is logic in kokkos.cpp for reverting the communication setting for `CUDA/Aware`, I think that can just be reused for `pair/only`. I can add this.

akohlmey:
> Ah, there is logic in kokkos.cpp for reverting the communication setting for `CUDA/Aware`, I think that can just be reused for `pair/only`. I can add this.

ok. assigning the PR to you to complete and test the KOKKOS part until you are happy with it. Will then look into the rest.

stanmoore1:
@akohlmey I am testing this PR now

akohlmey:
> @akohlmey I am testing this PR now

ok. i checked this morning and it works as expected for the GPU package and I also added documentation.

sjplimp (approved):

ndtrung81:
@akohlmey Thanks for adding this option!

akohlmey:
> @akohlmey Thanks for adding this option!

It was the idea of @stanmoore1 
I only re-implemented it slightly different so it can be applied to other packages as well.

ndtrung81 (approved):

stanmoore1:
@akohlmey I've tested this for Kokkos, everything is working as far as I can tell. I think this is ready to merge from my POV.

akohlmey:
> @akohlmey I've tested this for Kokkos, everything is working as far as I can tell. I think this is ready to merge from my POV.

excellent! I will have to wait until the CI testing clears and then merge it.

PR #2536, opened 2020-12-20
Add faster method for multi neighboring

jtclemm:
**Summary**

This pull request contains a slightly modified implementation of a new polydisperse neighboring method developed by Tom Shire, Kevin Stratford, and Kevin Hanley (references below). In the current multi implementation, there is a single binlist with a bin size set by the smallest interaction length. A separate stencil is then constructed for each atom type. For each itype, the stencil extends to the maximum interaction distance with other particles. In the new implementation, a separate binlist is used for each atom type and ~ntype^2 stencils are then constructed for itype-jtype pairs. Stencils extend up to the itype-jtype interaction distance. 

This approach greatly speeds up the construction of neighbor lists and may also allow fewer ghost atoms to be communicated. If Newton is on and there are only half neighborlists, one only needs to loop through small particles to look for large neighbors to find cross-interactions. Therefore, small ghost atoms only need to be communicated a distance set by the small-small interaction. In all cases tested, the new method improves performance over the old multi method by a factor of 3+. Therefore, we renamed the old method multi/old and replaced multi with this new method.

In addition to upgrading multi, this pull request also cleans up redundancies in the nstencil classes.  Previously, most stencil classes had a full, half+newton, and half+newtoff variety.  However, the full and half+newtoff classes were redundant as both created what one might consider a fully "spherical" stencil. This contrasts to half+newton where stencils only included bins to the upper right creating a "hemi-spherical" stencil. To remove this redundancy, all half+newtoff stencils were removed and half+newton stencils were renamed half stencils. The logic in neighbor->choose_stencil() was accordingly updated.
 
**Related Issue(s)**

NA

**Author(s)**

Tom Shire (U. Glasgow)
Kevin Stratford (U. Edinburgh)
Kevin Hanley (U. Edinburgh)
Joel Clemmer (SNL - jtclemm@sandia.gov)

**Licensing**

**Backward Compatibility**

No foreseen issues, input scripts using the old multi style will automatically use the new multi style

**Implementation Notes**

This PR is still in progress. Remaining tasks include:

- [x] Allow users to create custom categories by grouping different types to avoid ntype^2 overhead
- [x] Add check in comm_init() to confirm a full neighborlist isn't used with hierachical communication (small-to-large lookups)
- [x] Resize bins and atom2bin arrays (# atoms in group vs nall)
- [ ] Check compatibility of new stencils with Kokkos
- [x] Add examples
- [x] Update documentation when arguments are finalized (reference new paper)
- [x] Add size_one methods to granular pairstyles
- [x] Add option to specify collections by cutoff intervals
- [x] Add option to specify non-sequential collections of types
- [x] Revamp multi neighbor and comm to use a new, separate collection array of size nlocal+nghost

Extra items (probably a different PR):

1) Update fix pour to work with >1 atom types
2) Update fix pour/adapt to work with multi if atom radii can change

**Post Submission Checklist**

Relevant links:
[Journal article for old method](https://www.sciencedirect.com/science/article/pii/S001046550800115X)
[Journal article for new method](https://link.springer.com/article/10.1007/s40571-020-00361-2)
[Technical document for new method](http://www.archer.ac.uk/community/eCSE/eCSE12-09/eCSE12-09.php)
[Original implementation of new method](https://github.com/EPCCed/lammps)


jtclemm:
The primary features of this pull request are finished and works well on my system. However, I see the Jenkins bot is reporting failed builds. Is there a way to get an error log?

I had one specific concern I wanted feedback on. If users have many atom types, it is more efficient to combine types with similar cutoffs and create fewer binlists and stencils. I referred to these as groups of types. However when discussing with others, my use of the word group in the code/documentation has already produced confusion since it has nothing to do with regular groups in LAMMPS. Maybe there's a better way to express it?

Thank you for any advice. I look forward to suggestions for improvement.

akohlmey:
> The primary features of this pull request are finished and works well on my system. However, I see the Jenkins bot is reporting failed builds. Is there a way to get an error log?

You click on the "Details" link and then you should be able to click on the line following "Triggered Builds", since the compilation tests are test "bundles" of multiple tests, you need to do this twice. However, these links do not always show up immediately after a test failure in the new "Blue Ocean" visualization of the CI status. For that you can switch to the old style view by clicking on the '[->]' symbol on the top right and then click on the links of the "unstable" builds and on "Console Output" for the innermost of those. Yet another option to access those links is to join the '#testing' channel on the LAMMPS slack which allows to monitor the testing in real time and each individual failed test will create a message with an URL that will take you directly to the corresponding website on the CI server.

> I had one specific concern I wanted feedback on. If users have many atom types, it is more efficient to combine types with similar cutoffs and create fewer binlists and stencils. I referred to these as groups of types. However when discussing with others, my use of the word group in the code/documentation has already produced confusion since it has nothing to do with regular groups in LAMMPS. Maybe there's a better way to express it?

I think the term "collection" is not used yet.


> 
> Thank you for any advice. I look forward to suggestions for improvement.

jtclemm:
Ah now I see the build log. I didn't get that I had to click twice. I need to fix issue with a user package. 

Collection would work well so I'll also swap group->collection.

sjplimp:
@jtclemm For type groups (collections), is the wildcard syntax used by other commands (e.g. 3*5) sufficiently general for this command?  Say 3 different size particles are read from a data file, with types 1,2,3.
Then the user resets the types of boundary atoms for the different sizes to 4,5,6, so they can be frozen or viz differently.  Then the desired grouping for this command would be 3 groups = 1,4 and 2,5 and 3,6, correct?  So maybe the collection syntax should be one collection = a,b,c,... where each of a,b,c can be a single integer or a wildcard?  E.g 3,5 or 6*9,10,11*15.

sjplimp:
@jtclemm Also does the code have logic to insure that all types are included in exactly one collection?

jtclemm:
That is correct, there currently is no way for users to group/collect non-sequential atom types. Users would have to plan atom types before hand. It would be a nice feature but I didn't see any existing syntax in LAMMPS to define a list of arbitrary length.

I discussed this with Gary a bit and one idea was to have users specify a list (maybe like 1,2,3,5 or maybe even 1\*3,5). Another possibility would be to have users specify a series of tuples. Defining a group/collection #s followed by types, e.g. to define group 1 as types 1-3, and 5 and group 2 as type 4:
_neigh_modify multi/custom 1 1*3 1 5 2 4 _
Although this is hard to read and may be prone to error. Open to other ideas as well.

In terms of missing types, currently neigh_modify will error if a user assigns a type to >1 group/collection.

sjplimp:
yes, your first suggestion 1,3*5 is what I was suggesting.  Re: error, does it also give an error if any type is not assigned to any collection?

jtclemm:
Oh I completely missed reading the 2nd half of your 1st post, my bad! Then I can certainly work on swapping the syntax. Are there any issues if I use the C++ string class to parse a list? 

Re: Re: error, if a type is not specified it will be assigned it to its own group/collection. However, maybe it's better if the code doesn't make assumptions.

akohlmey:
hmm... is it really necessary to explicitly specify the atom types? wouldn't it be safer and more effective to have LAMMPS do the grouping into collections instead? e.g. based on some "cutoff binning" parameter? Then there would be the grouping into the collections through explicit lists of pairs of atom types which would have to be created based on whatever input syntax is chosen with explicit atom types and wildcards anyway.

akohlmey:
> Oh I completely missed reading the 2nd half of your 1st post, my bad! Then I can certainly work on swapping the syntax. Are there any issues if I use the C++ string class to parse a list?

yes. we are slowly converting from using C-style string handling to using the more readable C++ string processing where meaningful and convenient. Please also note the availability of the Tokenizer classes and utility functions in the `LAMMPS_NS::utils` namespace. We usually require that functions accepting `std::string` will have these as `const std::string &` arguments and return, if needed a copy, so that those functions would be thread safe. That way, also passing a `const char *` or `char *` as argument is supported and those are automatically converted.

jtclemm:
> 
> 
> hmm... is it really necessary to explicitly specify the atom types? wouldn't it be safer and more effective to have LAMMPS do the grouping into collections instead? e.g. based on some "cutoff binning" parameter? Then there would be the grouping into the collections through explicit lists of pairs of atom types which would have to be created based on whatever input syntax is chosen with explicit atom types and wildcards anyway.

I'm not quite sure I follow. Are you suggesting that the user would alternatively specify a binning distance which LAMMPS would use to create groups/collections by sorting types into bins based on their cutoffs?

>
>
> yes. we are slowly converting from using C-style string handling to using the more readable C++ string processing where meaningful and convenient. Please also note the availability of the Tokenizer classes and utility functions in the LAMMPS_NS::utils namespace. We usually require that functions accepting std::string will have these as const std::string & arguments and return, if needed a copy, so that those functions would be thread safe. That way, also passing a const char * or char * as argument is supported and those are automatically converted.

Thank you for the advice, I'll be sure to check out the tokenizer class. I have embarrassingly little experience with C-style strings so I'm glad some things are moving to c++ strings.

akohlmey:
> > hmm... is it really necessary to explicitly specify the atom types? wouldn't it be safer and more effective to have LAMMPS do the grouping into collections instead? e.g. based on some "cutoff binning" parameter? Then there would be the grouping into the collections through explicit lists of pairs of atom types which would have to be created based on whatever input syntax is chosen with explicit atom types and wildcards anyway.
 
> I'm not quite sure I follow. Are you suggesting that the user would alternatively specify a binning distance which LAMMPS would use to create groups/collections by sorting types into bins based on their cutoffs?

Something like that. If I read the description/discussion correctly then you initially have `Atom::ntypes^2` stencils to build the multi-cutoff neighbor lists and each has its associated cutoff (or cutoff^2) and now you want to speed up the process by grouping pairs of atom types with sufficiently similar cutoffs into "collections" that can use the same stencil. If this would have to be done manually, it is quite error prone (to avoid duplication) and may be inefficient (when types are chosen badly or cutoffs are changed but not the grouping into collections). Thus if there would be a single parameter stating when cutoffs are similar enough to be grouped and when not, this would be simpler to specify, more likely to get correct and less tedious to manage and update. All the extra work would be done internally and that is something that computers are good at. ;-)

sjplimp:
I think there are cases when defining the grouping by types could be tedious or error-prone.  However I also think there are cases (e.g. fine-grained polydispersity) when doing it by types would be easy and doing it by cutoff distance(s) would be trickier.  Maybe a compromise for now would be to do it by types, but also have a warning or error check if the particle sizes (or cutoffs) for one type group overlaps with another type group.

akohlmey:
> However I also think there are cases (e.g. fine-grained polydispersity) when doing it by types would be easy and doing it by cutoff distance(s) would be trickier.

I don't understand this statement. Can you give an example?

sjplimp:
Say you have 10 types for particles of different size (1,2,3, ..., 10).  And you want 3 groups, 1*3, 4*6, 7*10.
How do you define this?  "single parameter stating when cutoffs are similar enough to be grouped and when not"
since all consecutive sizes differ by the same amount.

jtclemm:
> 
> 
> I think there are cases when defining the grouping by types could be tedious or error-prone. However I also think there are cases (e.g. fine-grained polydispersity) when doing it by types would be easy and doing it by cutoff distance(s) would be trickier. Maybe a compromise for now would be to do it by types, but also have a warning or error check if the particle sizes (or cutoffs) for one type group overlaps with another type group.

There are definitely advantages/disadvantages to both options. I do actually like the simplicity of giving LAMMPS a single cutoff distance. However, it wouldn't give the user as much flexibility and it may not produce the most efficient grouping/collection. This brings up a fundamental issue in this code: how to define the size of an atom type. 

The easiest metric is to say the size of a particle is equal to it's same-type interaction, cut[itype][itype]. This is currently what the code does when it constructs a hierarchy of small to large particles to sort look ups. However, this may not make sense if cut[itype][jtype] is larger than both cut[itype][itype] and cut[jtype][jtype]. The code still works, but it may not make the most efficient choice. This ambiguity would also present issues if LAMMPS automatically binned atom types by their cutoffs.

akohlmey:
> This brings up a fundamental issue in this code: how to define the size of an atom type.

This is not a code issue, but an issue of the potential used or how it is implemented. In LAMMPS the "size" of a finite size (spherical) particle is given by the "radius" property (which is independent of the atom type, BTW). Otherwise you only have an "effective size" that depends on the potential function of *both* interacting atoms of a given pair. And then again, that effective size is usually not dependent on the interaction cutoff but rather on selected potential parameters (e.g. sigma in the case of Lennard-Jones).  This becomes a bit more complicated when looking at the colloid pair style in which case the "radius" *is* part of the potential parameters and thus tied to the atom type (and the interaction cutoff has to be set accordingly). But that case is quite the exception rather the the normal case in LAMMPS, and I suspect it was implemented this way because there was not yet a per-atom radius property available in LAMMPS at that time.

So the grouping in the way I see it would not have to be done based on the individual atom type (or its cutoff or effective size or radius or whatever), but rather on the corresponding parameter for the *pair* of atom types.

The case of `cutoff[itype][jtype]` being larger than either `cutoff[itype][itype]` or `cutoff[jtype][jtype]` seems to me a rather unusual model. Typically it would be somewhere in between.

akohlmey:
> Say you have 10 types for particles of different size (1,2,3, ..., 10). And you want 3 groups, 1*3, 4*6, 7*10.
> How do you define this? "single parameter stating when cutoffs are similar enough to be grouped and when not"
> since all consecutive sizes differ by the same amount.

that is straightforward. you don't compare each pair and say you start a new group when the difference between the pair is larger than a given threshold, but you start with the smallest and then start a new group when the difference between the smallest and the next candidate is larger than the threshold parameter you choose (which may be linear or more likely you need some non-linear function of the property you look at, which can be optimized empirically, if needed).

sjplimp:
I'm not thinking of any common use cases besides granular and colloidal systems.
Both are conceptually the same in these ways:
a) two or more discrete particle sizes (except see d)
b) cut_ii < cut_ij < cut_jj
However they differ in these ways:
c) colloidal types define the cutoffs (pair coeff),
   granular radii define cutoff (not necessarily types as Axel said)
d) granular can have continuous polydispersity,
   e.g. particle size is a random # between 1 and 10
So @jtclemm, how should the new multi method group the particles
for (1) few vs (2) many discrete sizes vs (3) continuous?  If we know the
answer to that, we can decide out how to let the user guide it.

jtclemm:
> 
> 
> > This brings up a fundamental issue in this code: how to define the size of an atom type.
> 
> This is not a code issue, but an issue of the potential used or how it is implemented. In LAMMPS the "size" of a finite size (spherical) particle is given by the "radius" property (which is independent of the atom type, BTW). Otherwise you only have an "effective size" that depends on the potential function of _both_ interacting atoms of a given pair. And then again, that effective size is usually not dependent on the interaction cutoff but rather on selected potential parameters (e.g. sigma in the case of Lennard-Jones). This becomes a bit more complicated when looking at the colloid pair style in which case the "radius" _is_ part of the potential parameters and thus tied to the atom type (and the interaction cutoff has to be set accordingly). But that case is quite the exception rather the the normal case in LAMMPS, and I suspect it was implemented this way because there was not yet a per-atom radius property available in LAMMPS at that time.
> 
> So the grouping in the way I see it would not have to be done based on the individual atom type (or its cutoff or effective size or radius or whatever), but rather on the corresponding parameter for the _pair_ of atom types.
> 
> The case of `cutoff[itype][jtype]` being larger than either `cutoff[itype][itype]` or `cutoff[jtype][jtype]` seems to me a rather unusual model. Typically it would be somewhere in between.

You're right, I'm being imprecise with my language. I'm saying size to refer to some metric of their interaction distance, whether that's `cut[itype][itype]` (as I currently used in the code) or maybe something like `neighbor->cuttype[itype]` (the max interaction distance for itype). Sorting or binning atom types by these two metrics is not guaranteed to give the same result. As you pointed out, maybe it's better to think of pairs of types. 

In terms of `cut[itype][jtype]` > `cut[itype][itype]` and `cut[jype][jtype]`, I agree it's unusual. I don't know what models have a larger mixed interactions but I received an input script from a collaborator to test which had such a setup. 



> 
> 
> I'm not thinking of any common use cases besides granular and colloidal systems.
> Both are conceptually the same in these ways:
> a) two or more discrete particle sizes (except see d)
> b) cut_ii < cut_ij < cut_jj
> However they differ in these ways:
> c) colloidal types define the cutoffs (pair coeff),
> granular radii define cutoff (not necessarily types as Axel said)
> d) granular can have continuous polydispersity,
> e.g. particle size is a random # between 1 and 10
> So @jtclemm, how should the new multi method group the particles
> for (1) few vs (2) many discrete sizes vs (3) continuous? If we know the
> answer to that, we can decide out how to let the user guide it.

If I remember correctly, LAMMPS still defines `cut[itype][jtype]` for granular particles by calculating the maximum radius of each atom type. So users need to separate large/small particles into different types. If particles are continuously distributed over a larger range, users will want to use more types. Of course, users may define extra atom types if there are grains of >1 material type. So grouping/collecting types is still useful.

akohlmey:
> If I remember correctly, LAMMPS still defines `cut[itype][jtype]` for granular particles by calculating the maximum radius of each atom type. So users need to separate large/small particles into different types. If particles are continuously distributed over a larger range, users will want to use more types. Of course, users may define extra atom types if there are grains of >1 material type. So grouping/collecting types is still useful.

This seems to be more a consequence of adapting the granular code to the framework of cutoff based pair styles (which is the common case in LAMMPS). LAMMPS some how needs to collect the largest interaction cutoff for defining the neighbor list cutoff (and from it the default for the communication cutoff). If for some reason there are atom types defined in a granular simulation (there is no intrinsic need for it), then of course it needs to collect this using the same framework by looking for the largest cutoff across all pairs of types (and then collect the largest radii for each type to do that). The existing multi-neighbor list style was constructed with the colloid pair style in mind, which *has* its radius *and* the corresponding cutoff tied to atom types, but usually also has a natural BIG-BIG, BIG-SMALL, and SMALL-SMALL grouping of interactions. But for atom style sphere, you cannot make that assumption. I would say it looks convenient to you here because your current implementation makes it convenient.

You also need to keep in mind that the "granular" pair style is some kind of super-hybrid pair style that allows to mix different granular models in a single style. So in that case the grouping by atom types would naturally be motivated by the interaction model in use and not the particles sizes (unless there is a correlation due to the material simulated).

I apologize is my insistence is irritating. My many years of following the lammps-users mailing list have made me very paranoid about user doing things in LAMMPS in ways that can be completely unexpected when somebody implements a feature. They often do make sense in hindsight and then sometimes it is painful to realize that one has to work around arbitrary limitations that could be avoided.  I know from personal experience how off-putting this can be, however I have also learned that it pays off in the long run to be thorough at the beginning and make sure there is a well defined behavior even for corner cases and configurations that may not look as people would want to use them. That you get comments from Steve and me makes this even more difficult, since we have a tendency to approach such issues from different directions (which in the end has often led us to superior solutions when finding a suitable compromise or being able to define clearly which cases should be covered and what not, even if it turned out to be a bit more laborious for one of us). In the end, how beneficial a new feature is to users depends as much on the quality of the implementation as it depends on the quality and clarity of its documentation and description.

jtclemm:
> 
> 
> > If I remember correctly, LAMMPS still defines `cut[itype][jtype]` for granular particles by calculating the maximum radius of each atom type. So users need to separate large/small particles into different types. If particles are continuously distributed over a larger range, users will want to use more types. Of course, users may define extra atom types if there are grains of >1 material type. So grouping/collecting types is still useful.
> 
> This seems to be more a consequence of adapting the granular code to the framework of cutoff based pair styles (which is the common case in LAMMPS). LAMMPS some how needs to collect the largest interaction cutoff for defining the neighbor list cutoff (and from it the default for the communication cutoff). If for some reason there are atom types defined in a granular simulation (there is no intrinsic need for it), then of course it needs to collect this using the same framework by looking for the largest cutoff across all pairs of types (and then collect the largest radii for each type to do that). The existing multi-neighbor list style was constructed with the colloid pair style in mind, which _has_ its radius _and_ the corresponding cutoff tied to atom types, but usually also has a natural BIG-BIG, BIG-SMALL, and SMALL-SMALL grouping of interactions. But for atom style sphere, you cannot make that assumption. I would say it looks convenient to you here because your current implementation makes it convenient.
> 
> You also need to keep in mind that the "granular" pair style is some kind of super-hybrid pair style that allows to mix different granular models in a single style. So in that case the grouping by atom types would naturally be motivated by the interaction model in use and not the particles sizes (unless there is a correlation due to the material simulated).
> 
> I apologize is my insistence is irritating. My many years of following the lammps-users mailing list have made me very paranoid about user doing things in LAMMPS in ways that can be completely unexpected when somebody implements a feature. They often do make sense in hindsight and then sometimes it is painful to realize that one has to work around arbitrary limitations that could be avoided. I know from personal experience how off-putting this can be, however I have also learned that it pays off in the long run to be thorough at the beginning and make sure there is a well defined behavior even for corner cases and configurations that may not look as people would want to use them. That you get comments from Steve and me makes this even more difficult, since we have a tendency to approach such issues from different directions (which in the end has often led us to superior solutions when finding a suitable compromise or being able to define clearly which cases should be covered and what not, even if it turned out to be a bit more laborious for one of us). In the end, how beneficial a new feature is to users depends as much on the quality of the implementation as it depends on the quality and clarity of its documentation and description.

No worries at all about insistence, I completely understand it's intended to be helpful not irritating. I also would prefer this PR be done correctly so I don't have to return to it in the future. As you say, most of my choices were made to fit into the existing structure of LAMMPS and the granular package. I'm hesitant to break away from this pre-existing framework since I do not know the development history of LAMMPS but if everyone agrees that the best solution is to move away, I am happy to do so.

Steve and I did discuss the possibility of divorcing neighboring from atom types back in the fall. I think we both agreed that ideally we'd use something other than type but didn't have a good alternative (correct me if I'm wrong Steve!). To help myself think about it, the main idea behind this PR is that using >1 binlist greatly accelerates polydisperse simulations (up to a factor of 10 over the old multi). Maybe the main questions are:
1) How many binlists do we build 
2) What's the bin size for each list
3) How do we sort atoms into those binlists
4) What stencils are required
5) Which binlists do atoms need to search to find their neighbors

Currently the answers are:
1) One binlist per group/collection of types - a better way of expressing it is that a collection is basically a binlist ID
2) Each collection/group calculates an intra-collection "size" which corresponds to the max `cut[itype][jtype]` for all itype, jtype in the collection. Bin sizes are half of this collection "size."
3) Use an array to map each atom type to a collection, each atom is then placed into a single binlist
4) A stencil is then required for every pair of collections, the radii of each stencil is equal to the maximum interaction distance between any pair of atoms in the two collections.
4) Typically, atoms search for neighbors within their own collection's binlist using a hemispherical/half stencil. Atoms also search binlists of other collections which have a larger "size" using a fully-spherical/full stencil. Exceptions include full neighbor lists or newton-off where all binlists need to be searched.

The main advantage of this setup is that users can create as many atom types they need and have the freedom to group them in any way into collections to maximize efficiency. The disadvantages are that atom types may not be the correct way of grouping of polydisperse particles (as in the granular case), syntax may be unnecessarily complex to create custom collections of types, the use of `cut[itype][jtype]` can be ambiguous and/or ad hoc (as in the granular case).

Throwing out an alternative, maybe we could keep the above structure but completely divorce collections from types. Maybe the neighbor class could trigger creation of a new per-atom quantity (using a public method in atom?). This per-atom vector could be used to map atoms to collections using any criteria (such as binning interaction lengths or radii). Collections would then be used to map atoms to bin lists (a collection id is just a binlist id). This would, of course, require more memory/communication. Is this maybe a better/worse possibility?

I think what Axel said about basing neighboring around "pairs of types" does have a lot of truth as well. Trying pursue that path, we could also completely break away from the above structure and create one binlist per interaction length (or set of similar interaction lengths). This could be even more efficient since an optimal binsize could be chosen for each interaction, but mapping atoms to binlists could get complicated. E.g. in a binary system of small/large particles, large particles would need to be placed in both the large-large binlist and the small-large binlist. I'll have to think about it more and answer those 5 questions I wrote above. 

I'm open to any of these possibilities (or others). They would just require iterating the code in this PR. However, don't want to go down a random path unless it's agreed upon.

sjplimp:
Good summary Joel.  Here is a slightly different idea that would divorce the multi-style neighbor list build from types.  Say that it only makes sense to use neigh multi with a couple of current pair styles, e.g. colloid and granular.  If so, then those pair styles could implment one or two functions (if both are needed), say size_one(i) and size_two(i,j).  Which would return the interaction length (effective size) for particle I or an I,J pair of particles.  The granular styles could return the particle diameter(s).  The colloid style could return the pair coeff size value associated with that particle.  

Is it low overhead for the the neighbor binning methods to just query those methods each time a list is built?  No need to store them with the atoms.  This would also allow for use of fix adapt to vary granular particles sizes over time.  The user of neigh multi would just need to define how the collections of sizes are defined.

If there are uses for multi with other pair styles in special cases, e.g. LJ with a range of cutoffs, then the 2 methods could be default-defined in the Pair parent class as returning cut[itype] or cut[itype][jtype] the same as now.

sjplimp:
Also, it is fine by me if we release this PR as-is with a "collection" = 1 or more consecutive types,
since (a) it works well, and (b) is compatible with wildcarding of types used in other settings, like
the pair_coeff command.  Then finalize later an enhancement to categorize particles for mutli neigh purposes in a more general way, using the ideas we're discussing.  In the interim we may get feedback from other users as well.

jtclemm:
> 
> 
> Good summary Joel. Here is a slightly different idea that would divorce the multi-style neighbor list build from types. Say that it only makes sense to use neigh multi with a couple of current pair styles, e.g. colloid and granular. If so, then those pair styles could implment one or two functions (if both are needed), say size_one(i) and size_two(i,j). Which would return the interaction length (effective size) for particle I or an I,J pair of particles. The granular styles could return the particle diameter(s). The colloid style could return the pair coeff size value associated with that particle.
> 
> Is it low overhead for the the neighbor binning methods to just query those methods each time a list is built? No need to store them with the atoms. This would also allow for use of fix adapt to vary granular particles sizes over time. The user of neigh multi would just need to define how the collections of sizes are defined.
> 
> If there are uses for multi with other pair styles in special cases, e.g. LJ with a range of cutoffs, then the 2 methods could be default-defined in the Pair parent class as returning cut[itype] or cut[itype][jtype] the same as now.

This could be an option. If size_one/two() returns cut[itype][i/jtype] for all non-granular pairstyles, it'd be the same as our current implementation. So I'd just need to figure out granular pair styles.

However, I just thought of an additional complication for granular systems. Currently the multi mode of comm is based on atom types. So if we restructure neighbor/multi to avoid using types, we'll need to also rewrite the multi communication mode. Otherwise, users will still have to bin granular atoms into different types in order to get the communication speedups. 


> 
> 
> Also, it is fine by me if we release this PR as-is with a "collection" = 1 or more consecutive types,
> since (a) it works well, and (b) is compatible with wildcarding of types used in other settings, like
> the pair_coeff command. Then finalize later an enhancement to categorize particles for mutli neigh purposes in a more general way, using the ideas we're discussing. In the interim we may get feedback from other users as well.

If we decide to keep this PR as is, I can still add in an option to specify a non-sequential list (e.g. 1,4*7).  If users don't specify a collection for an atom type, would it be better to error or automatically put that type in its own collection (w/ or w/o a warning)?

sjplimp:
I think size_one(i) could just return radius[i] or 2*radius[i], and size_two(i,j) = radius[i] + radius[j].  So the functions are one-liners (in-lined)?  For comm multi you just need size_one() for each particle - is that right?  For the neigh building do you have to call size_two() for every pair of particles? I'm not clear on how this logic is needed for binning and the stencil operations.  For N collections there are N sets of bins and N^2/2 sets of stencils?  I just want to avoid doing something expensive.

jtclemm:
> 
> 
> I think size_one(i) could just return radius[i] or 2*radius[i], and size_two(i,j) = radius[i] + radius[j]. So the functions are one-liners (in-lined)? For comm multi you just need size_one() for each particle - is that right? For the neigh building do you have to call size_two() for every pair of particles? I'm not clear on how this logic is needed for binning and the stencil operations. For N collections there are N sets of bins and N^2/2 sets of stencils? I just want to avoid doing something expensive.

In theory I think comm would need to know max[size_two(i,j) such that size_one(j) < size_one(i)]. Generally it's probably okay to approximate this as size_one(i) = 2*radius_i, but I think the granular pair style allows users to have attractive long-range JKR interactions. So size_two(i,j) is not always equal to radius_i + radius_j. I'll have to ask Dan. 

Currently yes, for N collections there are N bins and ~N^2 stencils (I think N^2/2 for newton on and N^2 for newton off).

jtclemm:
Reverting this PR back to a draft to add in new ideas. New plan is to create a separate local array in the neighbor class which specifies which collection each atom belongs to. This will allow users to specify collections using either types or intervals of particle radii.

jtclemm:
I'm mostly finished the 2nd version of this PR. I still need to run some more test cases and update the examples but am holding off until we confirming this is the direction we want to go. The main aspect that changed is neighboring and communication no longer depend on particle types. This addresses Axel's observation that particle type is not indicative of interaction distances for finite particles. 

The neighbor class now creates a new local per-atom array which maps atoms to a collection. A new method was added, `neighbor->build_collection()`, to construct this array. Since this information is not communicated, this method must be called each time a neighbor list is constructed. The multi comm mode will also use this collection array. So `build_collection()` must also be called in `comm->setup()`, before `comm->borders()`, and during `comm->border()` when new ghost atoms are added. This may be a bit conservative at the moment.

Typically, this array is constructed using atom types. By default, each atom type is mapped to its own collection but users can specify custom mappings using `comm_modify collection/type`. I updated the syntax for this command so users can now specify lists of ranges (e.g. `1*4,6,10*`). 

However, there is a new comm_modify command, `collection/interval`, which allows users to specify the upper cutoff for N intervals. Each interval then defines a collection. For most pair styles, this command simply looks up `cutsq[itype][jtype]` and automatically maps atom types to collections. The exception is granular pairstyles (and potentially others in the future) where the cutoff depends on atomic quantities. For such pairstyles, I created a new method `pair->atom2cut(i)` which returns the interaction distance between a particle `i` and a copy of itself. Typically this returns `radius[i]*2` but there are some exceptions such as the JKR granular pairstyle which has an extended cutoff. The `atom2cut(i)` method is then used to sort atoms into collections without using `cutsq[itype][jtype]`. Another method, `pair->radii2cut(r1,r2)`, was also defined for these pairstyles to calculate the hypothetical maximum interaction distance between particles with radii r1 and r2. This is needed in `neighbor->init()` to calculate the maximum possible interaction distance between different collections. 

Other miscellaneous notes:
1) The `comm_modify cutoff/multi X Y` command was tricky since the # of collections is not necessarily known when the command is called. So I simply saved all the arguments to cutoff/multi commands and waited until `comm->setup()` to parse them, maybe less than ideal
2) The old comm multi mode was renamed multi/old. This mode is still compatible with neighbor style multi or multi/old. 
3) All npair, nbin, and nstencil classes were updated to use this new collection array. 

Hopefully this write up helps. Thanks again for any feedback.

jtclemm:
I finished updating the examples, adding a system with a continuously range of particle sizes (power-law distributed, with a disparity ratio of ~1000) to demonstrate the collection/interval command. I attached a video.

https://user-images.githubusercontent.com/63308469/107128534-a9903b00-687b-11eb-9e4d-d4dfbd57171f.mp4

akohlmey:
@jtclemm FYI I just looked through the failing unit test cases and it seems they all fail trying to run a style from the USER-INTEL package.

jtclemm:
> 
> 
> @jtclemm FYI I just looked through the failing unit test cases and it seems they all fail trying to run a style from the USER-INTEL package.

Thank you for the info Axel, I poked around but couldn't figure out which unit tests were failing. I fixed some of the bitmasks for nbin classes in different packages (including USER-INTEL) which appears to have fixed issues with using the intel accelerator on my system. But it looks like there are still some issues with unit tests I'll have to figure out.

akohlmey:
> > @jtclemm FYI I just looked through the failing unit test cases and it seems they all fail trying to run a style from the USER-INTEL package.
> 
> Thank you for the info Axel, I poked around but couldn't figure out which unit tests were failing. I fixed some of the bitmasks for nbin classes in different packages (including USER-INTEL) which appears to have fixed issues with using the intel accelerator on my system. But it looks like there are still some issues with unit tests I'll have to figure out.

There are just two failures left for two different compilation settings:
https://ci.lammps.org/blue/organizations/jenkins/dev%2Fpull_requests%2Fubuntu%2Funit_tests%2Fcmake_mpi_openmp_smallbig_shared/detail/cmake_mpi_openmp_smallbig_shared/1758/tests
https://ci.lammps.org/blue/organizations/jenkins/dev%2Fpull_requests%2Fubuntu%2Funit_tests%2Fcmake_mpi_openmp_smallsmall_shared/detail/cmake_mpi_openmp_smallsmall_shared/1758/tests

These look somewhat random, so I would speculate that they are not issues with code directly, but rather something where you have access to not properly initialized memory. The unit tests often run rests where you have multiple "clear" commands within a single input and those can sometimes expose uninitialized memory access issues. The first time memory dynamically allocated on Linux is set to all zeroes by default, but then after multiple malloc/free calls the malloc library may "recycle" previously allocated memory which has random data in it.

A good way to expose such situations would be to run a variety of example input with valgrind.

jtclemm:
Thanks for the valgrind tip! I sorted out a bunch of problems with access of uninitialized memory which seems to have resolved the unit test failure. I also spotted nd fixed some cases where users might change the # of collections which could cause LAMMPS to fail. I see there may be some new issues to sort out which I'll try to take a look at in the morning.

akohlmey:
@jtclemm after running more extended tests (not run by default and most of them are shortened versions of the example inputs), you should also look at the following example input decks:
- examples/granregion/in.granregion.mixer
- examples/granregion/in.granregion.funnel
- examples/srd/in.srd.mixture

And this input here is failing (not sure if your changes are supposed to be fully backward compatible or the test may need updating):
https://github.com/lammps/lammps-testing/blob/master/tests/commands/comm_modify/in.comm_modify-002

Once your feature is complete, we should see, if there are ways to write tests either like this as full inputs or with direct introspection into the C++ code through a new unittest tester program.

jtclemm:
> 
> 
> @jtclemm after running more extended tests (not run by default and most of them are shortened versions of the example inputs), you should also look at the following example input decks:
> 
>     * examples/granregion/in.granregion.mixer
> 
>     * examples/granregion/in.granregion.funnel
> 
>     * examples/srd/in.srd.mixture
> 
> 
> And this input here is failing (not sure if your changes are supposed to be fully backward compatible or the test may need updating):
> https://github.com/lammps/lammps-testing/blob/master/tests/commands/comm_modify/in.comm_modify-002
> 
> Once your feature is complete, we should see, if there are ways to write tests either like this as full inputs or with direct introspection into the C++ code through a new unittest tester program.

I updated the info class called in in.comm_modify so it now works with the new and old multi.  However, there are two backwards compatibility issues that emerged:
1) Using `comm_modify mode multi` now requires the multi neighbor style since it uses the collection arrays. I updated the LAMMPS examples to also use the  multi neighbor style so they now run. Alternatively, one could also use `comm_modify mode multi/old` which is compatible with all neighbor styles.
2) I used zero-indexing for collections so the `comm_modify cutoff/multi` calls in in.comm_modify no longer work (they use types 1 and 2 as arguments but now need to use collections 0 and 1 as arguments). The entire collection code uses zero based numbering. What would be preferable: updating all of the code to one-based numbering, simply adding +1 to the `cutoff/multi` arguments so collections "look" like they use one-based numbering to users, or leave everything zero-indexed?


Also found a few more misc memory issues which are now patched.

doc/src/neighbor.rst
```diff
@@ -55,14 +55,24 @@ For the *bin* style, the bin size is set to 1/2 of
...
+cutoffs, one can reduce the extra overhead by defining custom collections 
+using the :doc:`neigh_modify <neigh_modify>` command. See the 
+:doc:`comm_modify mode bytype <comm_modify>` command for compatible 
+communication options that may be beneficial for simulations of this kind.
+
+An alternate style, *multi/old*, sets the bin size to 1/2 of the shortest 
+cutoff distance and multiple sets of bins are defined to search over for 
+different atom types. This algorithm used to be the default *multi* 
+algorithm in LAMMPS but was found to be significantly slower than the new
+approach. Although, there may be instances where the *multi/old* style
```
sjplimp, line 74: For now we are keeping the old option in case there are use cases where multi/old outperforms the new multi style.

doc/src/neighbor.rst
```diff
@@ -55,14 +55,24 @@ For the *bin* style, the bin size is set to 1/2 of
 the largest cutoff distance between any pair of atom types and a
 single set of bins is defined to search over for all atom types.  This
 can be inefficient if one pair of types has a very long cutoff, but
-other type pairs have a much shorter cutoff.  For style *multi* the
-bin size is set to 1/2 of the shortest cutoff distance and multiple
-sets of bins are defined to search over for different atom types.
+other type pairs have a much shorter cutoff. The *multi* style uses
+different sized bins for collections of different sized particles. Different
```
sjplimp, line 59: ... sized particles, where "size" may mean the physical size of the particle or its cutoff distance for interacting with other particles.

doc/src/neighbor.rst
```diff
@@ -55,14 +55,24 @@ For the *bin* style, the bin size is set to 1/2 of
...
+sets of bins are then used to construct the neighbor lists as as further 
+described by Shire, Hanley, and Stratford :ref:`(Shire) <bytype-Shire>`. 
 This imposes some extra setup overhead, but the searches themselves
-may be much faster for the short-cutoff cases.
-See the :doc:`comm_modify mode multi <comm_modify>` command for a
-communication option that may also be beneficial for simulations of
-this kind.
+may be much faster. By default, separate collections of particles are defined
```
sjplimp, line 63: By default, each atom type defines a separate collection of particles.  For systems where two or more atom types have the same size (either physical size or cutoff distance), the definition of collections can be customized, which can result in less overhead and faster performance.  See the neigh_modify XYZ command for how to define custom collections.  Whether the collection definition is customized or not, also see the comm_modify mode bytype command for communication options that further improve performance in a manner consistent with neighbor style multi.

doc/src/comm_modify.rst
```diff
@@ -28,9 +32,9 @@ Examples
 
 .. code-block:: LAMMPS
 
-   comm_modify mode multi
+   comm_modify mode multi reduce/multi
    comm_modify mode multi group solvent
-   comm_modift mode multi cutoff/multi 1 10.0 cutoff/multi 2*4 15.0
+   comm_modift mode multi cutoff/multi 0 10.0 cutoff/multi 2*4 15.0
```
sjplimp, line 37: modift -> modify

doc/src/neigh_modify.rst
```diff
@@ -188,15 +196,39 @@ atom can have.
...
+Typically these are good values for minimizing the time for
 neighbor list construction.  This setting overrides the default.
 If you make it too big, there is little overhead due to
 looping over bins, but more atoms are checked.  If you make it too
 small, the optimal number of atoms is checked, but bin overhead goes
 up.  If you set the binsize to 0.0, LAMMPS will use the default
 binsize of 1/2 the cutoff.
 
+The *collection/type* option allows you to define custom collections of atom 
+types for the *multi* neighbor mode. By grouping atom types with
```
sjplimp, line 210: types, used by the multi ...

doc/src/neigh_modify.rst
```diff
@@ -188,15 +196,39 @@ atom can have.
...
 neighbor list construction.  This setting overrides the default.
 If you make it too big, there is little overhead due to
 looping over bins, but more atoms are checked.  If you make it too
 small, the optimal number of atoms is checked, but bin overhead goes
 up.  If you set the binsize to 0.0, LAMMPS will use the default
 binsize of 1/2 the cutoff.
 
+The *collection/type* option allows you to define custom collections of atom 
+types for the *multi* neighbor mode. By grouping atom types with 
+similar cutoffs, one may be able to improve performance by reducing
```
sjplimp, line 211: with similar physical size or interaction cutoff length, one may ...

doc/src/neigh_modify.rst
```diff
@@ -188,15 +196,39 @@ atom can have.
...
+overhead. You must first specify the number of custom collections N to be
```
sjplimp, line 212: drop "custom"

doc/src/neigh_modify.rst
```diff
@@ -188,15 +196,39 @@ atom can have.
...
+defined followed by N lists of types. Each list consists of a series of type
+ranges separated by commas. The range can be specified as a 
+single numeric value, or a wildcard asterisk can be used to specify a range
+of values.  This takes the form "\*" or "\*n" or "n\*" or "m\*n".  For
+example, if M = the number of atom types, then an asterisk with no numeric 
+values means all types from 1 to M.  A leading asterisk means all types 
+from 1 to n (inclusive).  A trailing asterisk means all types from n to M 
+(inclusive).  A middle asterisk means all types from m to n (inclusive). 
+Note that all atom types must be included in a custom collection.
```
sjplimp, line 221: ... included in exactly one of the N collections.

doc/src/neigh_modify.rst
```diff
@@ -188,15 +196,39 @@ atom can have.
...
+
+The *collection/interval* option provides a similar capability. 
+This command allows a user to define custom collections by specifying a 
+series of cutoff intervals. LAMMPS will automatically sort atoms into these intervals
+based on their type-dependent cutoffs or their finite size.
+You must first specify the number of custom collections N to be
```
sjplimp, line 227: drop "custom"

doc/src/neigh_modify.rst
```diff
@@ -188,15 +196,39 @@ atom can have.
...
+defined followed by N values representing the upper cutoff of each interval.
+This command is particularly useful for granular pairstyles where the interaction distance
+of particles depends on their radius and may not depend on their atom type.
+
```
sjplimp, line 231: Not clear to me why the user doesn't specify N-1 values to produce N collections.  What happens if some atom is larger than the Nth value and is thus not in any collection?

doc/src/comm_modify.rst
```diff
@@ -87,12 +97,26 @@ warning is printed, if this bond based estimate is larger than the
...
-cutoffs are determined per atom type, a type specifier is needed and
-cutoff for one or multiple types can be extended. Also ranges of types
-using the usual asterisk notation can be given. For granular pair styles,
-the default cutoff is set to the sum of the current maximum atomic radii
-for each type.
+communication mode *multi* instead. Since the communication
+cutoffs are determined per atom collections, a collection specifier is needed and
+cutoff for one or multiple collections can be extended. Also ranges of collections
+using the usual asterisk notation can be given. 
+Collections are indexed from 0 to N-1 where N is the total number of collections.
```
sjplimp, line 104: yes, the collections should be indexed from 1 to N by the user in the input script.  Internally, they can be indexed however you like.

jtclemm:
Hi Axel, all of Steve's requested changes are now implemented so the pull request is ready for review again. Would you be able to run this PR back through the wringer of extended tests? Last time you ran the extra tests it helped identify some errors (all of which I believe should now fixed).

akohlmey:
@jtclemm testing has been mostly successful. I have updated the branch to the latest upstream version and did some (mostly) cosmetic changes. When running the [in.comm_modify-002.txt](https://github.com/lammps/lammps/files/6466286/in.comm_modify-002.txt)
input from our extended test suite, I get the following (unexpected) segmentation fault.
```
Program received signal SIGSEGV, Segmentation fault.
0x00007fffec668806 in LAMMPS_NS::Comm::modify_params (this=0x289030, narg=8, arg=0x2bb5a0)
    at /home/akohlmey/compile/lammps/src/comm.cpp:361
361	        cutusermulti[i-1] = cut;
```
This looks like either something is not properly initialized when processing the input or some check requiring additional input or a different order of inputs needs to be added.

akohlmey:
FYI, all other failing tests have been updated to be compatible with the new code or are due to a bug in upstream that is fixed in a different pending PR. So addressing the one problem from above should be sufficient for now. In a future pull request we will have to look into creating some proper unit tests for this new feature (and other neighbor list features).

jtclemm:
> FYI, all other failing tests have been updated to be compatible with the new code or are due to a bug in upstream that is fixed in a different pending PR. So addressing the one problem from above should be sufficient for now. In a future pull request we will have to look into creating some proper unit tests for this new feature (and other neighbor list features).

Thank you for running the tests, I spent all my time verifying user-specified collections worked and forgot to check default cases. The problematic input script should now work with no memory errors.

sjplimp (approved):

PR #2540, opened 2020-12-22
Update Kokkos library in LAMMPS to v3.3.1

stanmoore1:
**Summary**

Update Kokkos library in LAMMPS to v3.3.1

**Related Issue(s)**

None.

**Author(s)**

Stan Moore (SNL)

**Licensing**

**Backward Compatibility**

Yes


stanmoore1:
@akohlmey Kokkos now requires `c++14`. I can update the docs, but I'm not sure if other changes (i.e. to CMake) are needed.

akohlmey:
Is this the version of Kokkos that requires C++14 instead of C++11?
Then we need to add checks for that and also put a statement into the CMake config to enable the C++14 setting instead of C++11 which is enabled by the toplevel CMakeLists.txt file.

akohlmey:
> @akohlmey Kokkos now requires `c++14`. I can update the docs, but I'm not sure if other changes (i.e. to CMake) are needed.

our messages crossed. I can look into this, if you want me to. But that would have to wait until tomorrow. I have to tend to a few other things today.

akohlmey:
enabling C++14 with CMake is straightforward. but this now fails to compile with our intel compiler CI environment:
https://ci.lammps.org/job/dev/job/pull_requests/job/intel/job/cmake_kokkos_mpi_openmp_smallbig_icc_shared/1601/console

stanmoore1:
>but this now fails to compile with our intel compiler CI environment

OK how hard is it to use a newer Intel compiler version in the CI? I will try running this through Kokkos regression testing.

akohlmey:
> > but this now fails to compile with our intel compiler CI environment
> 
> OK how hard is it to use a newer Intel compiler version in the CI? I will try running this through Kokkos regression testing.

@rbberger is already working on it.

akohlmey:
> > but this now fails to compile with our intel compiler CI environment
> 
> OK how hard is it to use a newer Intel compiler version in the CI? I will try running this through Kokkos regression testing.

intel compiler version is icpc (ICC) 19.1.1.217 20200306
that is newer then what the Kokkos README.md file says they use for testing.
there may be some other reason for it to fail compiling.

stanmoore1:
OK, I'm trying with GNU 7.3.1 now.

stanmoore1:
GNU 7.3.1 gave an ICE, sigh...

stanmoore1:
GCC 8 + CUDA works fine though. I'll try with Intel.

akohlmey:
> GCC 8 + CUDA works fine though. I'll try with Intel.

just got info from the Kokkos slack. the problem is not the intel compiler but using CentOS 7 with default GCC. Intel morphs into using the C++ headers of the host and GCC 4.8 is too old for C++14.

rbberger:
Updated our Intel compiler to 19.1.2.254 (2020 update 2), that's the latest version I can get with my current license.

rbberger:
CentOS 7 Intel container now has GCC 8.4 as default GCC

stanmoore1:
Awesome, I worked around all the GCC 7 bugs, running Kokkos regression tests now.

stanmoore1:
I'm seeing some performance regressions with Kokkos 3.3.0, will have to resolve before this can be merged.

stanmoore1:
We tracked down the source of the performance regressions in Kokkos. Once the issue is fixed in upstream Kokkos, I'll cherry-pick the commits here. Until then, this PR is on hold.

stanmoore1:
Found a new performance regression (besides the one that was fixed), Kokkos developers are working on it.

stanmoore1:
The Kokkos 3.3.1 release is coming soon, so we are going to skip 3.3.0 and go straight to 3.3.1 in this PR.

stanmoore1:
Re-running Kokkos regression tests now.

stanmoore1:
Kokkos regression tests passed. @akohlmey do we need to put a NOTE in the docs about the new C++14 requirement?

cmake/Modules/Packages/KOKKOS.cmake
```diff
@@ -50,7 +53,7 @@ if(DOWNLOAD_KOKKOS)
   target_link_libraries(lammps PRIVATE LAMMPS::KOKKOS)
   add_dependencies(LAMMPS::KOKKOS kokkos_build)
 elseif(EXTERNAL_KOKKOS)
-  find_package(Kokkos 3.2.01 REQUIRED CONFIG)
+  find_package(Kokkos 3.3.00 REQUIRED CONFIG)
```
junghans, line 56:
```suggestion
  find_package(Kokkos 3.3.01 REQUIRED CONFIG)
```

akohlmey:
> Kokkos regression tests passed. @akohlmey do we need to put a NOTE in the docs about the new C++14 requirement?

Yes, I think that would be required so we have something to point to in case people start complaining. Are you aware of any regularly updated list of tested compilers for Kokkos? The list in the lib/kokkos/README.md file seems to be lagging behind.
It would be nice to insert a pointer to that info, so we won't have to maintain our own list.

akohlmey:
> > Kokkos regression tests passed. @akohlmey do we need to put a NOTE in the docs about the new C++14 requirement?
> 
> Yes, I think that would be required so we have something to point to in case people start complaining. Are you aware of any regularly updated list of tested compilers for Kokkos? The list in the lib/kokkos/README.md file seems to be lagging behind.
> It would be nice to insert a pointer to that info, so we won't have to maintain our own list.

Another item to keep in mind would be to check the list of supported architectures and devices.

cmake/Modules/Packages/KOKKOS.cmake
stanmoore1: @junghans good catch, fixed.

stanmoore1:
>Yes, I think that would be required so we have something to point to in case people start complaining. Are you aware of any regularly updated list of tested compilers for Kokkos? The list in the lib/kokkos/README.md file seems to be lagging behind.
It would be nice to insert a pointer to that info, so we won't have to maintain our own list.

I put a NOTE in the docs, can you take a look?

>Another item to keep in mind would be to check the list of supported architectures and devices.

Done.

akohlmey (approved):

akohlmey:
> > Yes, I think that would be required so we have something to point to in case people start complaining. Are you aware of any regularly updated list of tested compilers for Kokkos? The list in the lib/kokkos/README.md file seems to be lagging behind.
> > It would be nice to insert a pointer to that info, so we won't have to maintain our own list.
> 
> I put a NOTE in the docs, can you take a look?
> 
> > Another item to keep in mind would be to check the list of supported architectures and devices.
> 
> Done.

Looking good. thanks.

junghans (approved):

junghans:
diff stats said -26k lines, is that right?

akohlmey:
> diff stats said -26k lines, is that right?

wouldn't this be a good thing?

PR #2544, opened 2020-12-25
Added USER-PINN package with new pair style pinn

gangapun:
**Summary**

Physically Informed Neural Network (PINN) model has been addedd. Examples of PINN potentials for Al and Ta are included.
**Related Issue(s)**

**Author(s)**
Ganga P Purja Pun, George Mason University, Fairfax, Virginia.
Email: gangapurjapun@gmail.com

**Licensing**

**Backward Compatibility**

**Implementation Notes**

This addition will not affect other features in LAMMPS. Use 'pairstyle gmu/pinn' for proper pair style. Intel MKL library is required to compile the source code.

**Post Submission Checklist**


potentials/Al_2020.pinn
```diff
@@ -0,0 +1,1076 @@
+# Development of a general-purpose machine-learning interatomic
```
akohlmey, line 1: The first line of bundled potential files should follow the conventions of other bundled potential files and have a DATE:,  CONTRIBUTOR:, CITATION:, and UNITS: tag.

src/USER-PINN/pair_gmu_pinn.cpp
```diff
@@ -0,0 +1,2251 @@
...
+  tmp_fc_d = ann_fc_d (r, param);
+  tmp_exp = bop_fr (itype, jtype, r);
+
+  return tmp_exp * (tmp_fc_d - tmp_fc * alpha[itype][jtype]) / r;
+}
+
+
+/* ---------------------------------------------------------------------- */
+
+void PairGMUPINN::read_file(char *file)
```
akohlmey, line 1276: You may consider updating the file reading part using the `PotentialReader` class for more compact, thread safe, and buffer overflow safe reading/parsing for potential files, like it is done in several of the manybody potentials already.

src/USER-PINN/pair_gmu_pinn.cpp
```diff
@@ -0,0 +1,2251 @@
...
+    }
+  }
+
+  // sanity check !
+  if (nwords != nparams) error->all(FLERR, "Insufficient weights and biases in PINN potential file");
+
+  if (NNparams) delete [] NNparams;
+
+
+  // only for debugging
```
akohlmey, line 1626: more debugging code that should be removed before the code gets merged.

akohlmey (changes requested):
Thanks for your submission.

I have added some specific comments to specific sections of the code
that discuss issues I noticed while reading through the submitted changes.
I am adding here some more generic suggestions and comments. Since
there are quite a few items discussed, we will likely need more than one iteration to reach a state where your code can be added to the distribution.

- Explicitly requiring the MKL library is going to massively reduce
the accessibility of the package to users and will prohibit it from
being included in most packages that provide pre-compiled LAMMPS
executables due to the MKL license. I suggest to seriously consider
refactoring the code to make this an option, but to allow use of
a regular BLAS/LAPACK library, if necessary through alternate code
paths with preprocessor defines. The performance benefits of calling
BLAS functions like DAXPY and DCOPY is minimal, and that of DGEMV
is limited when compared to DGEMM, and a large part of the DGEMV
benefit also applies to non-MKL BLAS libraries, which will mean
that the impact on the overall performance will be even more limited.
The user USER-ATC and USER-AWPMD packages incorporate BLAS/LAPACK
calls through their Fortran interface, BTW, so they can be used with
either MKL or OpenBLAS or the bundled "linalg" library (which just
copies the required subset of BLAS/LAPACK functions from netlib).
Another option would be to interface with GNU gsl, which is used
by MSCG, USER-PLUMED, and USER-SCAFACOS.

- There is an inconsistency in the naming of the style. In some parts
it is referred to as "pinn" and others as "gmu/pinn". Our preference
would be "pinn" for simplicity (since the "gmu" part has not relevance
to the science itself). At any rate, all references to the style in
source code, readme, and documentation must be using the same name
and the associated files and classes should be named correspondingly.

- There are many function calls in the style "name (args)".  That is
rather confusing. Please use "name(args)" to be consistent with the
coding style used elsewhere in LAMMPS.  Correspondingly, please replace
any use of "*/* ... */" with "* /* ... */" to avoid confusing text
editors with syntax highlighting support.

- No file should have executable permissions except for shell
or python scripts.

- There is no documentation and integration of the contributed package
  into the documentation.
  + the package should be described in Packages_user.rst and
    Packages_details.rst. See USER-AWPMD for comparison.
  + the MKL requirement and associated compilation flags should be
    documented in Build_package.rst and Build_extras.rst. See USER-AWPMD for
    comparison
  + the pair style needs to be documented in a pair_pinn.rst (or
    pair_gmu_pinn.rst) file in the doc src folder and referenced in the
    Commands_pair.rst and pair_style.rst files.

- There should be at least one small(!) example in the examples/USER/pinn
  folder demonstrating the use of the potential and its features. Just a
  simple demo, no(!) research quality calculation. See, e.g. examples/mliap

- The contributed package needs to be properly added to the conventional
  make and CMake based build systems.

- Not sure how feasible this would be, but there is a more generic interface for machine learning potentials in the MLIAP package. @athomps can comment on that.

PR #2548, opened 2020-12-30
Add USER-PINN package with new pinn pair style

gangapun:
**Summary**

Included new pair style as ``pinn``.
**Related Issue(s)**

**Author(s)**
Ganga P Purja Pun, gangapurjapun@gmail.com

**Licensing**

**Backward Compatibility**

Not tested for backward compatibility.

**Implementation Notes**

It will not affect other features in LAMMPS; requires Intel MKL library.

**Post Submission Checklist**

For the PINN model refer Phys Rev Mat 4, 113807 (2020).


akohlmey (changes requested):
Many issues mentioned in the review comments on PR #2544 are still unresolved. In addition the build instructions provided are specific for the Intel compilers and not compatible with other compilers. The integration into the conventional build system is incomplete (and CMake non-existing).
The LAMMPS code is meant to be highly portable and thus we expect all packages to be able to be compiled without specific compilers or tools that meet the minimum requirements. As a point of reference even the USER-INTEL package can be compiled and run without the Intel compilers using alternate code paths for compiler specific features and - of course - with some performance degradation for non-Intel compilers.

Please add a comment when you think your submission is complete and all pending issues addressed as far as you are willing or able to do. Don't hesitate to ask questions if you need help with some of the requests.

gangapun:
> Many issues mentioned in the review comments on PR #2544 are still unresolved. In addition the build instructions provided are specific for the Intel compilers and not compatible with other compilers. The integration into the conventional build system is incomplete (and CMake non-existing).
> The LAMMPS code is meant to be highly portable and thus we expect all packages to be able to be compiled without specific compilers or tools that meet the minimum requirements. As a point of reference even the USER-INTEL package can be compiled and run without the Intel compilers using alternate code paths for compiler specific features and - of course - with some performance degradation for non-Intel compilers.
> 
> Please add a comment when you think your submission is complete and all pending issues addressed as far as you are willing or able to do. Don't hesitate to ask questions if you need help with some of the requests

In the interest of time I may not be able to replace MKL function calls with appropriate BLAS calls.

akohlmey:
> In the interest of time I may not be able to replace MKL function calls with appropriate BLAS calls.

Please explain "interest of time".

Have you looked at GNU GSL, which has a similar CBLAS style interface ( https://www.gnu.org/software/gsl/doc/html/blas.html ) or just plain CBLAS ( http://www.netlib.org/blas/#_cblas ) which is included in the OpenBLAS package available in most Linux distribution? The MKL code may remain in place, but it should be something that would allow to be replaced through preprocessor directives with alternate calls that allow building a version that is compatible with redistribution of binaries without restrictions. As already explained, in the current form the utility of this package is rather limited due to being unconditionally dependent on MKL (and for functionality that is not really MKL specific).

gangapun:
I will look into it.

Ganga

gangapun:
Updated pair_pinn.cpp

akohlmey:
> Updated pair_pinn.cpp

more requested updates are still open.

akohlmey:
@gangapun I did some tests with the code in this branch and made a few observations:
- the execution is agonizingly slow compared to other machine learning based pair styles in LAMMPS and unbelievably slow when not using compiler optimization. the latter indicates some problem in the code structure.
- from profiling with perf it looks like the "hotspot", i.e. where most time is spent seems to be the loop over the Legendre polynomial section. 
- I get very different energies and pressure with "newton off" compared to "newton on". that must not be. if the code is only intended to be used with "newton on", then a test for refusing "newton off" must be added.
- I see no reference to tools or the procedure for how to derive new potentials outside of the reference to the publication. the documentation should also describe the file format, so that others could use the method to build their own potentials.
- the setup_params() function has a hardcoded constant of 1.5 to multiply params->cut with to give the actual cutoff. where does this come from? i don't see it used anywhere else. it seems to be the same as params->cut_range? is that a coincidence?
- there is a memory leak in the code that causes a loss of a few hundred bytes for every step.

gangapun:
> @gangapun I did some tests with the code in this branch and made a few observations:
> 
>     * the execution is agonizingly slow compared to other machine learning based pair styles in LAMMPS and unbelievably slow when not using compiler optimization. the latter indicates some problem in the code structure.
>  It is slow and I have not optimized for speed. 

>     * from profiling with perf it looks like the "hotspot", i.e. where most time is spent seems to be the loop over the Legendre polynomial section.
>  There are code segments that require optimization.

>     * I get very different energies and pressure with "newton off" compared to "newton on". that must not be. if the code is only intended to be used with "newton on", then a test for refusing "newton off" must be added.
>  I will add a test to refuse "newton off".

>     * I see no reference to tools or the procedure for how to derive new potentials outside of the reference to the publication. the documentation should also describe the file format, so that others could use the method to build their own potentials.
>  I will add a section to describe the potential file format in the documentation.

>     * the setup_params() function has a hardcoded constant of 1.5 to multiply params->cut with to give the actual cutoff. where does this come from? i don't see it used anywhere else. it seems to be the same as params->cut_range? is that a coincidence?
>  It is used to computed feature vectors.

>     * there is a memory leak in the code that causes a loss of a few hundred bytes for every step.
 I will check if there are pointers not deallocated properly.

akohlmey:
@gangapun as a point of reference, I did set up a plane-wave DFT calculation for the same system geometry (32 Al atoms in a cubic box) and using a norm-conserving pseudo potential and a 30Ry cutoff and a 3x3x3 kspace grid and that calculation for a single point energy calculation from scratch is still about 10x faster than a single step of the pinn potential (any MD or optimization calculation will be faster, because it can re-use the wavefunction(s) from the previous step(s) as starting point.

gangapun:
Inserted check always "newton on". Included 'pinn' potential file format.

akohlmey:
> Inserted check always "newton on". Included 'pinn' potential file format.

Sorry, but I don't see such a check in this branch. Also, adding a PDF file especially without any reference to it is in practice the same as not documenting the file format. Except for some legacy cases, we have now all documentation included in the .rst files (and plan to convert and integrate those as well, as we have the time to do it). There are plenty of pair styles that can serve as an example, particularly in the MANYBODY package. The need for typesetting mathematical expressions is no restriction anymore, as those can be handled by the math extension we have enabled.

I also see that requested cosmetic changes are still missing. We want to establish a common and consistent programming style in all of LAMMPS (not just the core files) so that people have it easier to read and compare code.  As with other items, there are exceptions for legacy code that has not yet been transformed yet, but for all newly added code conforming to coding style is easier to realize and thus a requirement.

A more serious open issue is the disappointing performance. The code is currently multiple orders of magnitude slower than what would make it a useful addition. I cannot (and do not want to) judge the physics and qualities of the model itself, and we don't expect its performance to match code that has been optimized by experts for years, yet the gap is far larger than what we feel is acceptable when comparing it to other "expensive" calculations.

We would appreciate it if you could give us some update on how and when you plan to address these items.

athomps:
The paper indicates that the simulation rate for aluminum on typical CPU nodes is about 0.8 katom-step/core-sec and about 170x slower than EAM, which seems reasonable. If the example is running much slower than that, something may be wrong with the implementation.

gangapun:
I left the research group and the institution (George Mason University) since January 1, 20201. I am doing this only as a volunteer. Therefore I am not in a position to address the issues and the time line.

akohlmey:
> I left the research group and the institution (George Mason University) since January 1, 20201.

so do I get this right? you are contacting us from 18180 years in the future? ;-)

> I am doing this only as a volunteer. Therefore I am not in a position to address the issues and the time line.

The situation is that in its current state this pull request does not meet our requirements to have it merged.
Not having the time to complete something is not a valid reason to have changes included. I have multiple feature branches of my own in that sad state. But unless somebody else volunteers to take over that work from me or I find the time to restart the project(s) they will remain in that state, regardless how useful the changes in them would be.

So I think you have the choice between four options:
1. withdraw the current pull request, work on the feature branch in your own time and resubmit it when you feel all requested changes have been implemented and all concerns are addressed or provide convincing proof that those are not relevant or due to mistakes
2. withdraw the current pull request, contact your old research group, and figure out if somebody there is interested, willing, and able take care of the pending issues and then have that person submit a new pull request with the updated code
3. withdraw the current pull request and distribute and maintain it independently as a separate package or similar, you can even have github host a small website about it for you. there are a number of LAMMPS add-ons that are distributed like that where people for a variety of reasons did not submit their code for inclusion into the LAMMPS distribution.
4. withdraw the current pull request and abandon it. of course we don't want to see work go to waste. I mention this merely for completeness and to point out that you would not be the first person to have a lot of effort invested into something that never sees any application beyond your own use.

gangapun:
Yes. You are right; the current pull request does not meet the requirements. It is better to close the current pull request and let someone from the group resubmit later when all the issues are addressed. So, I will close the current pull request.

Thanks for you reviews.

Ganga

PR #2172, opened 2020-06-19
SNAP optimizations - AoSoA data layouts, chem snap indexing conventions, misc perf improvements

weinbe2:
**Summary**

This PR provides various performance optimizations to the evaluation of the SNAP potential within the Kokkos package when using the Cuda backend (though it should generalize well to any GPU), focusing on routines for group theory contractions. It also modifies the form of data structures for chem snap with the intent of making future data layout transformations easier (described in greater detail below). This does further bifurcate the CPU/GPU code paths, which I discuss under Implementation Notes.

There is a perf regression on the CPU, appears to be because of the View rank change, which could be rolled back.

Performance notes below. For robustness, I note that `master` is specifically commit `7e2f29bb6576f68bdab06e8ffc8cc80cc6f85f19` at time of filing.

Performance improvements on V100-16GB:

|Input file  |  Timesteps/sec master | Timesteps/sec PR | Perf improvement | Comment |
| ------------- | ------------- | -------------| ----------- | -------- |
| 2J8 benchmark  | 239 | 278 | 1.16x | Benchmark problem size (2000 atoms) |
| 2J14 benchmark | 39.4 | 50.4 | 1.28x | ECP KPP problem size (2000 atoms) |
| in.snap.W.2940 | 1534 | 1580 | 1.03x | `/examples/snap/` single atom (128 atoms) |
| in.snap.WBe.PRB2019 | 1510 | 1540 | 1.02x | `/examples/snap/` multi-atom, no chem (128 atoms) |
| in.snap.InP.JCPA2020 | 422 | 476 | 1.13x | `/examples/snap/` multi-atom, yes chem (512 atoms) |

Performance regressions on 20 core Broadwell (used KOKKOS_ARCH=BDW appropriately), source appears to be changing rank of Views as CPU code is otherwise untouched. 

|Input file  |  Timesteps/sec master | Timesteps/sec PR | Perf change | Comment |
| ------------- | ------------- | -------------| ----------- | -------- |
| 2J8 benchmark  | 11.2 | 10.3 | 0.92x | Benchmark problem size (2000 atoms) |
| 2J14 benchmark | 1.34 | 1.25 | 0.93x | ECP KPP problem size (2000 atoms) |
| in.snap.W.2940 | 185 | 161 | 0.87x | `/examples/snap/` single atom (128 atoms) |
| in.snap.WBe.PRB2019 | 192 | 163 | 0.85x | `/examples/snap/` multi-atom, no chem (128 atoms) |
[...]

PR #2547, opened 2020-12-30
Collected small fixes and code formatting updates

akohlmey:
**Summary**

This pull request includes multiple small changes to fix bugs, silence compiler warnings, and make the source code formatting of conditionals and loops more consistent across LAMMPS.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

No known issues.

**Implementation Notes**

The following small fixes are included:
- remove unused variables in various places
- avoid a buffer overflow from the terminating NULL in fix bond/react from using sprintf() to write into a `char` variable.
- fix off-by-one bug for arrhenius style constraints in fix bond/react
- correct several minor formatting and spelling issues in the manual
- whitespace fixes

In addition, the following source code formatting convention `keyword (condition) {` with a blank after the keyword and no blank with the parenthesis and a blank between the closing parenthesis and the opening brace was applied where it could be found and fixed through simple scripting with grep and sed.

Also, some changes to improve formatting consistency in CMake scripting were included. The changes there follow a common CMake script style, which differs from the LAMMPS C++ style. Imported files with different but consistent formatting like the NetCDF modules were left untouched.

**Post Submission Checklist**

- [x] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [x] The source code follows the LAMMPS formatting guidelines
- [x] The feature has been verified to work with the conventional build system
- [x] The feature has been verified to work with the CMake based build system

PR #2549, opened 2020-12-30
Kokkos SNAP optimizations --- Recursive polynomial evaluation optimizations, improved parallelism, various C++11-isms

weinbe2:
**Summary**

This PR provides various performance optimizations to the evaluation of the SNAP potential within the Kokkos package when using the Cuda backend (though it should generalize well to any GPU), primarily featuring a large refactor of the algorithm used to compute the Wigner matrices (and derivatives thereof). This approach utilizes the AoSoA data layout introduced for some data structures in #2172, improving the consistency of the approach taken on the GPU. There are other small optimizations that improve performance of the Kokkos/SNAP-internal compressed neighbor table (which could have value elsewhere) and expose additional parallelism for small systems.  This PR does not change build options or the external interface as it stands, but there are open questions for @stanmoore1 below that may introduce changes. Further details are given under Implementation details.

This PR effectively "does no harm" to the CPU Kokkos backend (1% regressions here and there, which seems to be a theme with each GPU-specific PR). This has no effect on the CPU implementation in the SNAP package, which has been historically faster than the Kokkos/CPU backend.

Performance notes below. For robustness, I note that master is specifically commit 427d0377476a1a52570734145539b88a41cc41c9 at time of filing.

Performance improvements on Quadro V100, CUDA 11.1 for various benchmarks. Inputs are averages over 15 runs, sig figs truncated to reflect fluctuations between runs. I included measurements for different problem sizes to show benefits at smaller vs larger atom counts. Some performance equalities between `in.snap.W.2940` and `in.snap.WBe.PRB2019` are _not_ typos.

| Input file | 2Jmax | Problem Size (atoms) | Timesteps/sec master | Timesteps/sec PR | Perf improvement | Comment |
|---|---|---|---|---|---|---|
| 2J8 benchmark | 8 | 4000 | 151.911 | 271.101 | 1.78x | Benchmark problem |
| 2J14 benchmark | 14 | 4000 | 24.1521 | 26.6446 | 1.10x | ECP KPP problem |
| in.snap.W.2940 | 8 | 128 | 1580.82 | 1769.9 | 1.12x | `/examples/snap/` single atom |
| " | " | 1024 | 554.811 | 816.739 | 1.47x | increased problem size |
| in.snap.WBe.PRB2019 | 8 | 128 | 1573.58 | 1765.4 | 1.12x | `/examples/snap/` multi-atom, no chem |
| " | " | 1024 | 554.217 | 803.303 | 1.45x | increased problem size |
| in.snap.InP.JCPA2020 | 6 | 512 | 512.125 | 915.31 | 1.79x |  `/examples/snap/` multi-atom, yes chem |
| " | " | 4096 | 82.6831 | 168.69 | 2.04x |  increased problem size, batchsize 2048 |

Larger boosts for smaller 2Jmax and larger atom counts are reasonable: former due to these optimizations being targetted towards kernels scaling as `Jmax^3` as opposed to `Jmax^5`, latter due to small-system latency overheads.

**Related Issue(s)**

Further improvements on #2172. This PR (and others preceding it) establish precedence for bifurcating the CPU and GPU codepaths of the Kokkos/SNAP implementation.

**Author(s)**

**Licensing**

**Backward Compatibility**

No changes.

**Implementation Notes**

* Modifies `ComputeUi` and `ComputeFusedDeidrj` to exploit a new depth-first algorithm for computing the Wigner matrices and derivatives thereof
* Converts the data layout of `ulisttot` to AoSoA, reflecting the behavior of the new algorithm
* Eliminates `TransformYi` as it is not needed for the new data layouts used in `ComputeFusedDeidrj`
* Increases the parallelism exposed in `TransformUi`, requiring a new look-up table `idxu_full_half` which maps between symmetrized indices and full indices in the context of the symmetries of the Wigner matrices, improving performance for small systems
* Splits the Kokkos/SNAP-specific compressed neighbor table routine into separate GPU (ComputeNeigh) and CPU (ComputeNeighCPU) routines, matching previous naming conventions
* Exposes additional parallelism in the Kokkos/SNAP-specific neighbor table routine
* Splits pre-computing the Cayley-Klein parameters into a separate routine `ComputeCayleyKlein`, improving load balancing. As part of this I removed the `CayleyKleinPack` structure, replacing it with individual AoSoA structures for `a`, `b`, `da`, etc
* Updated `grow_rij` and `memory_usage` appropriately
* Replaces all internal instances of `double` with a new template parameter `real` attached to the `PairSNAPKokkos` and `SNAKokkos`
* Replaces all explicit "magic" factors of 32 (inner-most A in the AoSoA layouts, corresponding to the number of threads in a warp on an NVIDIA GPU) with a new, arguably more intuitive, template parameter `vector_length`, set to 1 (and otherwise unused) for the CPU backend

Correctness has been verified over several configurations (listed in timings above) by comparing thermodynamic quantities between Kokkos/Cuda and Kokkos/Serial runs with this PR against Kokkos/Cuda runs of the current `master` branch.

Some open points of discussion:
* I had to make a few arguably imperfect implementation choices to get the new template parameters to play nicely with the LAMMPS pair style factory and `PairStyle` declarations at the top of `src/KOKKOS/pair_snap_kokkos.h` --- for example, explicitly using `PairSNAPKokkos<LMPDeviceType, double, 32>` threw compiler errors, so I created a thin wrapper `PairSNAPKokkosDevice` class that sidesteps these issues. I'm not married to this approach, we can certainly work out something clearer or reason out the issue necessitating the wrapper.
* I didn't modify the default batch size away from `2000` as I've previously discussed with @stanmoore1 offline. I still think it'd be a good idea (at least for the GPU backend), but I wanted to revisit the details before making a blanket decision since it has implications for both the CPU and the GPU. If this gets changed it necessitates changes to the documentation, which is why I left those checkboxes below. (If anything we should change it to something that divides into a sufficiently large power of 2 --- 32 at least.)

**Post Submission Checklist**

None.

### DIFF
--- a/src/KOKKOS/kokkos_type.h
+++ b/src/KOKKOS/kokkos_type.h
@@ -1076,15 +1076,29 @@ struct params_lj_coul {
 
 // Pair SNAP
 
+#define SNAP_KOKKOS_REAL double
+#define SNAP_KOKKOS_HOST_VECLEN 1
+
+#ifdef LMP_KOKKOS_GPU
+#define SNAP_KOKKOS_DEVICE_VECLEN 32
+#else
+#define SNAP_KOKKOS_DEVICE_VECLEN 1
+#endif
+
+
+// intentional: SNAreal/complex gets reused beyond SNAP
 typedef double SNAreal;
 
 //typedef struct { SNAreal re, im; } SNAcomplex;
-template <typename real>
-struct alignas(2*sizeof(real)) SNAComplex
+template <typename real_>
+struct alignas(2*sizeof(real_)) SNAComplex
 {
+  using real = real_;
+  using complex = SNAComplex<real>;
   real re,im;
 
-  SNAComplex() = default;
+  KOKKOS_FORCEINLINE_FUNCTION SNAComplex()
+   : re(static_cast<real>(0.)), im(static_cast<real>(0.)) { ; }
 
   KOKKOS_FORCEINLINE_FUNCTION SNAComplex(real re)
    : re(re), im(static_cast<real>(0.)) { ; }
@@ -1117,6 +1131,15 @@ struct alignas(2*sizeof(real)) SNAComplex
     return *this;
   }
 
+  KOKKOS_INLINE_FUNCTION
+  static constexpr complex zero() { return complex(static_cast<real>(0.), static_cast<real>(0.)); }
+
+  KOKKOS_INLINE_FUNCTION
+  static constexpr complex one() { return complex(static_cast<real>(1.), static_cast<real>(0.)); }
+
+  KOKKOS_INLINE_FUNCTION
+  const complex conj() { return complex(re, -im); }
+
 };
 
 template <typename real>
@@ -1125,18 +1148,6 @@ KOKKOS_FORCEINLINE_FUNCTION SNAComplex<real> operator*(const real& r, const SNAC
 }
 
 typedef SNAComplex<SNAreal> SNAcomplex;
-
-// Cayley-Klein pack
-// Can guarantee it's aligned to 2 complex
-struct alignas(32) CayleyKleinPack {
-
-  SNAcomplex a, b;
-  SNAcomplex da[3], db[3];
-  SNAreal sfac;
-  SNAreal dsfacu[3];
-
-};
-
 
 #if defined(KOKKOS_ENABLE_CXX11)
 #undef ISFINITE
--- a/src/KOKKOS/pair_snap_kokkos.h
+++ b/src/KOKKOS/pair_snap_kokkos.h
@@ -13,9 +13,15 @@
 
 #ifdef PAIR_CLASS
 
-PairStyle(snap/kk,PairSNAPKokkos<LMPDeviceType>)
-PairStyle(snap/kk/device,PairSNAPKokkos<LMPDeviceType>)
-PairStyle(snap/kk/host,PairSNAPKokkos<LMPHostType>)
+#include "kokkos_type.h"
+
+PairStyle(snap/kk,PairSNAPKokkosDevice<LMPDeviceType>)
+PairStyle(snap/kk/device,PairSNAPKokkosDevice<LMPDeviceType>)
+#ifdef LMP_KOKKOS_GPU
+PairStyle(snap/kk/host,PairSNAPKokkosHost<LMPHostType>)
+#else
+PairStyle(snap/kk/host,PairSNAPKokkosDevice<LMPHostType>)
+#endif
 
 #else
 
@@ -33,9 +39,11 @@ namespace LAMMPS_NS {
 // Routines for both the CPU and GPU backend
 template<int NEIGHFLAG, int EVFLAG>
 struct TagPairSNAPComputeForce{};
-struct TagPairSNAPComputeNeigh{};
+
 
 // GPU backend only
+struct TagPairSNAPComputeNeigh{};
+struct TagPairSNAPComputeCayleyKlein{};
 struct TagPairSNAPPreUi{};
 struct TagPairSNAPComputeUi{};
 struct TagPairSNAPTransformUi{}; // re-order ulisttot from SoA to AoSoA, zero ylist
@@ -44,10 +52,10 @@ struct TagPairSNAPBeta{};
 struct TagPairSNAPComputeBi{};
 struct TagPairSNAPTransformBi{}; // re-order blist from AoSoA to AoS
 struct TagPairSNAPComputeYi{};
-struct TagPairSNAPTransformYi{}; // re-order ylist from AoSoA to AoS
 struct TagPairSNAPComputeFusedDeidrj{};
 
 // CPU backend only
+struct TagPairSNAPComputeNeighCPU{};
 struct TagPairSNAPPreUiCPU{};
 struct TagPairSNAPComputeUiCPU{};
 struct TagPairSNAPTransformUiCPU{};
@@ -59,7 +67,7 @@ struct TagPairSNAPComputeYiCPU{};
 struct TagPairSNAPComputeDuidrjCPU{};
 struct TagPairSNAPComputeDeidrjCPU{};
 
-template<class DeviceType>
+template<class DeviceType, typename real_, int vector_length_>
 class PairSNAPKokkos : public PairSNAP {
 public:
   enum {EnabledNeighFlags=FULL|HALF|HALFTHREAD};
@@ -67,6 +75,10 @@ public:
   typedef DeviceType device_type;
   typedef ArrayTypes<DeviceType> AT;
   typedef EV_FLOAT value_type;
+
+  static constexpr int vector_length = vector_length_;
+  using real = real_;
+  using complex = SNAComplex<real>;
 
   PairSNAPKokkos(class LAMMPS *);
   ~PairSNAPKokkos();
@@ -78,10 +90,10 @@ public:
   double memory_usage();
 
   template<class TagStyle>
-  void check_team_size_for(int, int&, int);
+  void check_team_size_for(int, int&);
 
   template<class TagStyle>
-  void check_team_size_reduce(int, int&, int);
+  void check_team_size_reduce(int, int&);
 
   template<int NEIGHFLAG, int EVFLAG>
   KOKKOS_INLINE_FUNCTION
@@ -92,14 +104,17 @@ public:
   void operator() (TagPairSNAPComputeForce<NEIGHFLAG,EVFLAG>,const typename Kokkos::TeamPolicy<DeviceType, TagPairSNAPComputeForce<NEIGHFLAG,EVFLAG> >::member_type& team, EV_FLOAT&) const;
 
   KOKKOS_INLINE_FUNCTION
-  void operator() (TagPairSNAPComputeNeigh,const typename Kokkos::TeamPolicy<DeviceType, TagPairSNAPComputeNeigh>::member_type& team) const;
-
-  KOKKOS_INLINE_FUNCTION
   void operator() (TagPairSNAPBetaCPU,const int& ii) const;
 
   // GPU backend only
   KOKKOS_INLINE_FUNCTION
-  void operator() (TagPairSNAPPreUi,const typename Kokkos::TeamPolicy<DeviceType, TagPairSNAPPreUi>::member_type& team) const;
+  void operator() (TagPairSNAPComputeNeigh,const typename Kokkos::TeamPolicy<DeviceType, TagPairSNAPComputeNeigh>::member_type& team) const;
+
+  KOKKOS_INLINE_FUNCTION
+  void operator() (TagPairSNAPComputeCayleyKlein, const int iatom_mod, const int jnbor, const int iatom_div) const;
+
+  KOKKOS_INLINE_FUNCTION
+  void operator() (TagPairSNAPPreUi,const int iatom_mod, const int j, const int iatom_div) const;
 
   KOKKOS_INLINE_FUNCTION
   void operator() (TagPairSNAPComputeUi,const typename Kokkos::TeamPolicy<DeviceType, TagPairSNAPComputeUi>::member_type& team) const;
@@ -123,12 +138,12 @@ public:
   void operator() (TagPairSNAPComputeYi,const int iatom_mod, const int idxz, const int iatom_div) const;
 
   KOKKOS_INLINE_FUNCTION
-  void operator() (TagPairSNAPTransformYi,const int iatom_mod, const int idxu, const int iatom_div) const;
-
-  KOKKOS_INLINE_FUNCTION
   void operator() (TagPairSNAPComputeFusedDeidrj,const typename Kokkos::TeamPolicy<DeviceType, TagPairSNAPComputeFusedDeidrj>::member_type& team) const;
 
   // CPU backend only
+  KOKKOS_INLINE_FUNCTION
+  void operator() (TagPairSNAPComputeNeighCPU,const typename Kokkos::TeamPolicy<DeviceType, TagPairSNAPComputeNeighCPU>::member_type& team) const;
+
   KOKKOS_INLINE_FUNCTION
   void operator() (TagPairSNAPPreUiCPU,const typename Kokkos::TeamPolicy<DeviceType, TagPairSNAPPreUiCPU>::member_type& team) const;
 
@@ -173,7 +188,7 @@ protected:
   t_bvec bvec;
   typedef Kokkos::View<F_FLOAT***> t_dbvec;
   t_dbvec dbvec;
-  SNAKokkos<DeviceType> snaKK;
+  SNAKokkos<DeviceType, real, vector_length> snaKK;
 
   int inum,max_neighs,chunk_size,chunk_offset;
   int host_flag;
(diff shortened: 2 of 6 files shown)


weinbe2:
FYI @athomps @charlessievers @akkamesh

src/KOKKOS/pair_snap_kokkos.h
```diff
@@ -13,9 +13,15 @@
 
 #ifdef PAIR_CLASS
 
-PairStyle(snap/kk,PairSNAPKokkos<LMPDeviceType>)
-PairStyle(snap/kk/device,PairSNAPKokkos<LMPDeviceType>)
-PairStyle(snap/kk/host,PairSNAPKokkos<LMPHostType>)
+#include "kokkos_type.h"
```
akohlmey, line 16: This include statement breaks compilation of LAMMPS when KOKKOS is **not** enabled.
All `FeatureStyle(style/name,ClassName)` macros must be compatible to compile without any packages installed and without any references to package specific code, since there is a list of them compiled into the LAMMPS class in order to provide error messages for which package a style is provided by in case it is used but the package is not included into the LAMMPS executable in use.

weinbe2: Thank you for explaining this @akohlmey , I was a bit confused. I'll figure out a workaround momentarily.

akohlmey: The code failing is the `LAMMPS::init_pkg_lists()` function. As you may be able to see from the code, the ClassName macros argument is ignored in this function, so a simple `#ifdef LMP_KOKKOS` wrapper around the `#include` may be a sufficient solution.

weinbe2: Ah, thank you for a solution suggestion---it's something I'll keep in my "back pocket".

In this case here I can remove the `#include kokkos_type.h` outright---it's a sloppy artifact of some intermediate attempts to use (for example) `PairSNAPKokkos<LMPDeviceType, SNAP_KOKKOS_REAL, SNAP_KOKKOS_DEVICE_VECLEN>`, which depended on macros in `kokkos_type.h`, before I switched to using a thin wrapper instead: https://github.com/weinbe2/lammps/blob/fea-snap-ui-aosoa/src/KOKKOS/pair_snap_kokkos.h#L261. `SNAP_KOKKOS_REAL` and `SNAP_KOKKOS_DEVICE_VECLEN`, and the associated template parameters to `PairSNAPKokkos`, are new with this PR, which is why these types of issues hadn't come up in the past.

I'll push the fix in a bit, just doing some due-diligence tests on my end first. This is an example of what happens when I automate my own testing a little too much!

akohlmey: No worries, that is why we try to have a wide matrix of compilation tests for different platforms and different compilation setting on all pull requests. Too many compilation issues went past us in the past and even though it would not be feasible to go over all possible permutations of compilation settings with both supported build systems, we cover quite a lot of ground with the existing choices and those have been extremely helpful to identify potential portability issues and compilation failures.
https://github.com/lammps/lammps-testing/blob/master/scripts/builds/test_coverage.rst

weinbe2:
A quick question @akohlmey --- I can see that the `checkstyle` test is detecting "Style issues". Is there a way that I can access the list of issues? I couldn't find them by clicking on the "Details" list (which I imagine is broadly reasonable for permissions reasons). If there's a way I can see them that I missed, or if you could please send me a copy of the list of style issues, I'd be happy to put the fixes relevant for this PR in. (It'd arguably be easier than hunting for trailing whitespaces, etc, by hand.)

akohlmey:
> A quick question @akohlmey --- I can see that the `checkstyle` test is detecting "Style issues". Is there a way that I can access the list of issues? I couldn't find them by clicking on the "Details" list (which I imagine is broadly reasonable for permissions reasons). If there's a way I can see them that I missed, or if you could please send me a copy of the list of style issues, I'd be happy to put the fixes relevant for this PR in. (It'd arguably be easier than hunting for trailing whitespaces, etc, by hand.)

Please see: https://docs.lammps.org/Build_development.html#coding-style-utilities
The whitespace fixup currently does not detect/eliminate tabs. Those are most easily removed using the built in tools from emacs or vim.

weinbe2:
> > A quick question @akohlmey --- I can see that the `checkstyle` test is detecting "Style issues". Is there a way that I can access the list of issues? I couldn't find them by clicking on the "Details" list (which I imagine is broadly reasonable for permissions reasons). If there's a way I can see them that I missed, or if you could please send me a copy of the list of style issues, I'd be happy to put the fixes relevant for this PR in. (It'd arguably be easier than hunting for trailing whitespaces, etc, by hand.)
> 
> Please see: https://docs.lammps.org/Build_development.html#coding-style-utilities
> The whitespace fixup currently does not detect/eliminate tabs. Those are most easily removed using the built in tools from emacs or vim.

Thank you!

akohlmey:
@weinbe2 if you have difficulties accessing CI test output, you can switch from the fancy "Blue Ocean" visualization to the old style output by clicking on the `[->]` icon next to the "Login" button. That will show links with detailed outputs and - for grouped runs - links to the individual runs and from there you can access the console output.

akohlmey:
@weinbe2 FYI. I just looked at the CI page myself after clicking on "Details" in the PR page. You can see the output from the whitespace checker after clicking on the yellow circle labeled "Whitespace" and then on the ">" line starting with "singularity", which has the output of the check. Some of the issues are already addressed in PR #2547, so you really only need to fix issues that would be introduced by your changes (that is why style issues are mentioned but not blocking merges (currently)).

weinbe2:
@akohlmey thanks for the tip on switching back to the old style output, but also on where I could find the issues in the Blue Ocean vis style. You also headed off my questions about the MILAP formatting issues.

athomps:
@stanmoore1 Can you review this?  We would like to start using it for production runs.

stanmoore1:
Sure, will do.

src/KOKKOS/kokkos_type.h
```diff
@@ -1076,15 +1076,29 @@ struct params_lj_coul {
...
+#endif
+
+
+// intentional: SNAreal/complex gets reused beyond SNAP
 typedef double SNAreal;
 
 //typedef struct { SNAreal re, im; } SNAcomplex;
-template <typename real>
-struct alignas(2*sizeof(real)) SNAComplex
+template <typename real_>
```
stanmoore1, line 1093: Can we change `real` to `real_type`? Also why the underscore?

src/KOKKOS/kokkos_type.h
```diff
@@ -1076,15 +1076,29 @@ struct params_lj_coul {
...
+struct alignas(2*sizeof(real_)) SNAComplex
 {
+  using real = real_;
```
stanmoore1, line 1096: Not sure why this is needed, i.e. can we just use the template parameter?

src/KOKKOS/pair_snap_kokkos_impl.h
```diff
@@ -632,201 +698,145 @@ void PairSNAPKokkos<DeviceType>::operator() (TagPairSNAPComputeNeigh,const typen
...
+  int flattened_idx = team.team_rank() + team.league_rank() * team_size;
+
+  // extract neighbor index, iatom_div
+  const int iatom_div = flattened_idx / (max_neighs * (twojmax + 1)); // plug in int_fastdiv
+  const int jj_jbend = flattened_idx - iatom_div * (max_neighs * (twojmax + 1));
+  const int jbend = jj_jbend / max_neighs;
+  const int jj = jj_jbend - jbend * max_neighs;
+
+  // check base atom numbers, hopefully this doesn't lead to blocking conditions?
+  // yep, we can't have a return in here
```
stanmoore1, line 747: Please clean up comments.

src/KOKKOS/pair_snap_kokkos_impl.h
```diff
@@ -632,201 +698,145 @@ void PairSNAPKokkos<DeviceType>::operator() (TagPairSNAPComputeNeigh,const typen
...
-void PairSNAPKokkos<DeviceType>::operator() (TagPairSNAPComputeUi,const typename Kokkos::TeamPolicy<DeviceType,TagPairSNAPComputeUi>::member_type& team) const {
-  SNAKokkos<DeviceType> my_sna = snaKK;
+void PairSNAPKokkos<DeviceType, real, vector_length>::operator() (TagPairSNAPComputeUi,const typename Kokkos::TeamPolicy<DeviceType,TagPairSNAPComputeUi>::member_type& team) const {
+  SNAKokkos<DeviceType, real, vector_length> my_sna = snaKK;
 
-  // Extract the atom number
-  int ii = team.team_rank() + team.team_size() * (team.league_rank() % ((chunk_size+team.team_size()-1)/team.team_size()));
-  if (ii >= chunk_size) return;
+  constexpr int team_size = sizeof(real) == 4 ? 8 : 4;
+  if (team.team_size() != team_size) return; // error
```
stanmoore1, line 735: If this check is true, does that mean there is a bug, and it should never get here? Or should error out?

stanmoore1 (changes requested):

stanmoore1 (changes requested):
Looks good overall, I had some minor change requests in the comments.

src/KOKKOS/kokkos_type.h
weinbe2: Will do!

src/KOKKOS/kokkos_type.h
weinbe2: You can use it inside the definition of the `SNAComplex` class, but not outside the definition. Sort of a contrived example, but you can do something like:

```
typedef SNAComplex<double> my_complex;
[...]
using real = typename my_complex::real; // double
```
But not
```
typedef SNAComplex<double> my_complex;
[...]
using real = typename my_complex::real_; // iirc throws a compile time error
```

All this said---in the changes I've made it isn't strictly necessary, that is, I could have just used `real` everywhere and it would have been fine, but that's only because everything uses a uniform precision---`PairSNAPKokkos`, `SNAKokkos`, all reductions within it---as opposed to mixed precision. This is more of a forward-thinking change.

src/KOKKOS/pair_snap_kokkos_impl.h
weinbe2: Can do, sorry, I thought I'd cleaned these up.

src/KOKKOS/pair_snap_kokkos_impl.h
weinbe2: Ah, this is somewhat sloppy, it was meant to enforce consistency between that line and this line: https://github.com/weinbe2/lammps/blob/fea-snap-ui-aosoa/src/KOKKOS/pair_snap_kokkos_impl.h#L327

I can replace this with a `static constexpr int` that gets explicitly reused both places.

weinbe2:
@stanmoore1 what are your thoughts on increasing the default chunk size?

src/KOKKOS/pair_snap_kokkos_impl.h
weinbe2: Done, and amusingly removing that check gave a small speed-up.

src/KOKKOS/pair_snap_kokkos_impl.h
weinbe2: Done, did an additional sweep for sloppy comments too.

src/KOKKOS/kokkos_type.h
weinbe2: Done here and elsewhere. I use `int vector_length_`  a few places, I didn't touch those, couldn't think of a good clarifying name. Those instances are pretty closely followed by a `static constexpr int vector_length = vector_length_;` anywho.

weinbe2:
> Looks good overall, I had some minor change requests in the comments.

Thank you @stanmoore1 ! I've addressed them up to a few open questions. I also did a quick compile-and-correctness check.

src/KOKKOS/kokkos_type.h
weinbe2: Oh, I may have misunderstood your request. Do you mean replacing `real` with `real_type` everywhere, and not my specific change of `real_` -> `real_type`?

rbberger:
Ignore the intel and windows compilation failures on this PR. Having issues with ccaches not properly working with intel compilers. I'll look at it tomorrow.

src/KOKKOS/kokkos_type.h
stanmoore1: >Do you mean replacing real with real_type everywhere

Yes, that was my thought. I'm open to other suggestions if you have a preference.

weinbe2: No preference, `real_type` is growing on me. I'll take care of it later today.

weinbe2:
> > what are your thoughts on increasing the default chunk size?
> 
> I think we should increase it. Is 4000 a good value?

Let's use 4096, the motivation being problem sizes with larger powers of two. (For example---with a block size of 4000, a problem size of 4096 will have one full chunk then one latency-driven chunk with 96; 8192 will have two full chunks then one latency-driven of size 192, etc.)

If you're okay with that, I can make the change across the `SNAP` package, documentation, I have the PR you introduced the chunk size in for reference.

stanmoore1:
4096 sounds good. I can make the changes, I'm working on this now.

src/KOKKOS/pair_snap_kokkos.h
stanmoore1: Another option (for the future) is to use the `accelerator_kokkos.h` wrapper, which has dummy implementations of Kokkos for when Kokkos isn't enabled.

src/KOKKOS/kokkos_type.h
akohlmey: @weinbe2 @stanmoore1 please note that using "real" as a type or variable name can lead to conflicts with some libraries and legacy codes. If I remember correctly, we used to have a definition of "real" in some header of some LAMMPS package in the past and that created some nasty and hard to debug compilation issues when certain combination of packages were included. This is one of the reasons why I have been consistently pushing for keeping our style headers with the minimum amount of include files and PIMPLify cases where this cannot easily be done.

Even if there is no conflict the use of a common name like "real" can be confusing, especially for people with a significant exposure to Fortran. ;-) Thus making the name more specific, e.g. "real_kokkos" or "real_kk_t" or more descriptive of the use case could help with future maintenance.

stanmoore1: I just switched `real` to `real_type`, is that acceptable?

stanmoore1:
I'm double-checking performance and correctness on Summit now.

weinbe2:
> I'm double-checking performance and correctness on Summit now.

Thanks for making the changes @stanmoore1 , I'll do a quick local check too.

src/KOKKOS/kokkos_type.h
akohlmey: I don't recall any library/software using "real_type", so yes, that'll do. Ultimately, the real mistake is in those libraries or packages that define "real" in their (global) headers instead of using something more unique. But since those kind of conflicts as so nasty to track down, it is a good idea to be proactive.
In LAMMPS also there is a hierarchy of how much of a problem this can be. Local definitions inside a function in a .cpp file are the least concern (if at all) while a (pair) style header file, which is included together with many other such files at the same time, is the biggest issue. So a package specific header file that is only used by .cpp files in the same package is a lesser risk anyway.

weinbe2:
All looks good from my cursory checks.

stanmoore1:
On Summit I see a 1.6x speedup for 2J8 and 1.2x speedup for 2J14. This is for 2000 atoms instead of 4000 atoms though.

weinbe2:
> On Summit I see a 1.6x speedup for 2J8 and 1.2x speedup for 2J14. This is for 2000 atoms instead of 4000 atoms though.

That all seems reasonable---slightly different GPUs and CUDA versions, different arch w/increased latencies, different atom count, and replacing the `ComputeUi` and `ComputeFusedDeidrj` team size with a `constexpr` gave a slight speed-up not reflected in my numbers on top.

stanmoore1:
Yeah, I'm including the ~5% speedup from CUDA version too.

stanmoore1:
I'm seeing slight differences in the forces/trajectories between GPU and vanilla CPU. It shows up in latest master too, so this PR isn't the source, but we aren't getting the agreement I would expect for only 1 timestep. This seems like a small numerical regression, any ideas?

stanmoore1:
I turned off fix Langevin for the correctness check, so the pRNG isn't the source.

athomps:
There were issues with velocity assignment earlier.  Are you using 
```velocity all create 300.0 4928459 loop geom```
?

athomps:
`loop geom` eliminates effect of MPI ranks.

stanmoore1:
I verified correctness. I have one or two more checks, then should be ready to merge.

weinbe2:
> Yeah, I'm including the ~5% speedup from CUDA version too.

Using CUDA 11.2 and 2000 atoms, I'm seeing (in units timestep/sec)

|Problem size | master | PR | speed-up |
|---|---|---|---|
|2J8|350|548|1.56x|
|2J14|52|60|1.15x|

And for 4000 atoms

|Problem size | master | PR | speed-up |
|---|---|---|---|
|2J8|154|300|1.95x|
|2J14|26|30|1.15x|

* 2000 atoms just isn't saturating the GPU for the 2J8 problem
* I know the 11.2 release lead to some non-trivial speed-ups for the AoSoA-style kernels, so I'm not surprised re-writing `ComputeUi` and `ComputeFusedDeidrj` in that form gave an impressive boost relative to 11.1.

stanmoore1:
I get 525 and 66 timesteps/s on Summit. Seems weird that 2J14 is better on Power9?

weinbe2:
> I get 525 and 66 timesteps/s on Summit. Seems weird that 2J14 is better on Power9?

I'm using a Quadro instead of a Tesla, so it's reasonable.

stanmoore1:
Ah gotcha!

stanmoore1 (approved):

stanmoore1:
I believe this is ready for merge. Thanks @weinbe2 for the impressive speedups!

akohlmey (approved):

Issue #2551, opened 2021-01-06
[Feature Request] More configuration introspection for accelerator packages


akohlmey:
**Summary**

There are already several introspection APIs that allow to determine which features/packages/settings are available in LAMMPS, but we are currently missing the option to detect compile time settings for accelerator styles, particularly GPU, KOKKOS, and USER-INTEL. This would be extremely useful for unit testing (so they can better decide which tests to run and how) and running LAMMPS from script or library interfaces.

**Detailed Description**

For GPU we would need to know which API is used (CUDA, OPENCL, or HIP) and what precision setting is used (SINGLE, MIXED or DOUBLE). For KOKKOS we would need to know what features are available (device types and host types). for USER-INTEL we would need to know whether OFFLOAD is active or not. This is what it currently looks like.

PR #2552, opened 2021-01-07
Added functions for detecting the compile time configurations of the GPU package

ndtrung81:
**Summary**

This PR addresses the feature request #2551, adding functions that return the compile time settings of the GPU package. Two functions are added to return the API (CUDA, OpenCL or HIP) and the precision mode (single, mixed or double) used to build the GPU library.

**Related Issue(s)**

N/A

**Author(s)**

Trung Nguyen (Northwestern)

**Licensing**

**Backward Compatibility**

Yes

**Implementation Notes**

This implemetation may need further changes/re-arrangement to have a consistent interface across accelerator packages when detection is used in unit testing, or when using LAMMPS as a library from external codes. I am open for suggestions.

**Post Submission Checklist**

### DIFF
--- a/lib/gpu/Makefile.hip
+++ b/lib/gpu/Makefile.hip
@@ -144,6 +144,9 @@ $(OBJ_DIR)/lal_%.o: lal_%.cpp $(CUHS) $(ALL_H)
 $(LIB_DIR)/libgpu.a: $(OBJS)
 	$(AR) -crs $@ $(OBJS)
 	printf "export HIP_PLATFORM := %s\n%s\n" "$(HIP_PLATFORM)" "$(HIP_LIBS_TARGET)" > Makefile.lammps
+	@echo "api = HIP_API" > configure.options
+	@echo "precision = $(HIP_PRECISION)" >> configure.options
+	@echo "arch = $(HIP_ARCH)" >> configure.options
 
 # test app building
 
--- a/lib/gpu/Nvidia.makefile
+++ b/lib/gpu/Nvidia.makefile
@@ -902,6 +902,9 @@ $(BIN_DIR)/nvc_get_devices: ./geryon/ucl_get_devices.cpp $(NVD_H)
 $(GPU_LIB): $(OBJS) $(CUDPP)
 	$(AR) -crusv $(GPU_LIB) $(OBJS) $(CUDPP)
 	@cp $(EXTRAMAKE) Makefile.lammps
+	@echo "api = CUDA_API" > configure.options
+	@echo "precision = $(CUDA_PRECISION)" >> configure.options
+	@echo "arch = $(CUDA_ARCH)" >> configure.options
 
 clean:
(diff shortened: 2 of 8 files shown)


akohlmey:
@ndtrung81 thanks for your submission. There is a conceptual problem with this approach. It is not possible to query what is available without first using the package. While for the precision setting, that may be acceptable, it is not for the accelerator API in use. One of the use cases is the unittest functionality. While there we have compiled executables, they are using the library interface and effectively those are just "adaptive input files" and most groups of tests are independent LAMMPS executions where a new command line is issued (including suffix and package options).
The workflow, particularly for testing force computing styles, would roughly like this:
- check if a package is compiled in
- check if all required (optional) styles are available
- skip over special cases (e.g. styles where some tests cannot be done for technical reasons)
- if all checks pass run groups of tests and compare results to reference from plain CPU run
thus for adding GPU package support it would be needed to know which API is used (and then tested if that is available on the test machine) so the command line can be adapted, and whether the library is compiled with double precision, so that results can be matched.

ndtrung81:
@akohlmey thank you for the explanation. For the use case you mentioned, which involves checks outside of FixGPU and the LAMMPS executables, maybe after building the GPU library we generate/update a text file, called configure.options (after VMD), that lists all the configuration options, like

api = CUDA_API
precision = SINGLE_DOUBLE

Then in for the check if a package is complied in (step 1 in the example workflow), if the GPU package is installed, we can next parse the configure.options file in lib/gpu to get the API and precision mode. This approach can be extended to other accelerator packages I suppose.

akohlmey:
> @akohlmey thank you for the explanation. For the use case you mentioned, which involves checks outside of FixGPU and the LAMMPS executables, maybe after building the GPU library we generate/update a text file, called configure.options (after VMD), that lists all the configuration options, like
> 
> api = CUDA_API
> precision = SINGLE_DOUBLE
> 
> Then in for the check if a package is complied in (step 1 in the example workflow), if the GPU package is installed, we can next parse the configure.options file in lib/gpu to get the API and precision mode. This approach can be extended to other accelerator packages I suppose.

that might work, but it would be a deviation from what we already have. after some thinking and looking at what is done in other parts of LAMMPS, it would be most straightforward to have a define -DLMP_GPU added (either in the CMake script or in the package install script), similar as it is done with -DLMP_USER_INTEL and -DLMP_KOKKOS.
then it would be straightforward to add something like the introspection functions in the library interface or in the LAMMPS or Info class and just protect the call to the GPU library with an `#if defined(LMP_GPU)` `#endif` bracket, so it won't produce a link error in case the GPU package is not installed. this could be made a more general API to query all targeted accelerator packages.

ndtrung81:
@akohlmey I agree, it would be much more straightforward that way.

akohlmey:
@ndtrung81 I have just pushed some simple proof of concept code to a feature branch in my repo:
https://github.com/lammps/lammps/compare/master...akohlmey:accelerator-introspection
Please have a look and let us know what you think.

ndtrung81:
@akohlmey I like the proposed implmentation, it's very clear and simple.

PR #2556, opened 2021-01-14
Add eam/he pair style for modeling He bubbles in metals

akohlmey:
**Summary**

This adds a new pair style eam/he which is derived from eam/fs and mostly replaces its compute() method.
This also requires minor modifications to the potential file and the corresponding reader, but those are so minor, that they are implemented into eam/fs and protected with a flag. For simplicity and new global variable `rhomin` has been added to the eam pair style and together with `rhomax` they are thus always initialized and thus avoiding uninitialized data access in pair styles not using, but communicating them.

**Author(s)**

Xiaowang Zhou (SNL) with some minor modifications by Axel Kohlmeyer (Temple U)

**Licensing**

**Backward Compatibility**

No known issues.

**Post Submission Checklist**

**Corresponding Publication **

https://doi.org/10.1103/PhysRevB.103.014108


doc/src/pair_eam.rst
```diff
@@ -417,6 +423,45 @@ eV-Angstroms) as in EAM *setfl* files.  Note that in Finnis/Sinclair,
...
+
+The units of these quantities in line 1 are the same as for *setfl*
+files.  Note that the rho(r) arrays in Finnis/Sinclair can be
+asymmetric (:math:`\rho_{\alpha\beta} (r) \neq \rho_{\beta\alpha} (r)` )
+so there are Nelements\^2 of them listed in the file.
+
+Following the Nelements sections, Nr values for each pair potential
+phi(r) array are listed in the same manner (r\*phi, units of
+eV-Angstroms) as in EAM *setfl* files.  Note that in Finnis/Sinclair,
+the phi(r) arrays are still symmetric, so only phi arrays for i >= j
```
athomps, line 462: This block from 444-462 is identical to previous description for eam/fs and perhaps should be ommitted.

src/MANYBODY/pair_eam_he.cpp
```diff
@@ -0,0 +1,238 @@
...
+    if (p < -1.0) {
+      p = -1.0;
+    } else if (p > 1.0) {
+      p = 1.0;
+    }
+    coeff = frho_spline[type2frho[type[i]]][m];
+    fp[i] = (coeff[0]*p + coeff[1])*p + coeff[2];
+    if (eflag) {
+      phi = ((coeff[3]*p + coeff[4])*p + coeff[5])*p + coeff[6];
+      if (rho[i] < rhomin) {
```
athomps, line 146: It seems that the only difference in compute() relative to pair_eam.cpp is right here.  Could this be handled in another way, without duplicating code?

akohlmey: I already removed some duplicated code from the original version sent to me (which was a modified version of pair_eam_fs.cpp with the modified compute() method added), but didn't see a simple way to do this without impacting performance and complicating the entire "tree" of derived classes. This seems to be a bit of a niche application, so I stopped at this and did not consider to generalize the change like we just did in tersoff (which would be a similar if not larger effort with the added complication that the contributing author is not as comfortable with git and github and thus asked me to help getting this contribution included). A major difference to the tersoff case is that this modification also requires a modified potential file (and file reader). The entire difference between the two compute methods is given below. If you feel that could/should be merged, you are probably best off contacting the contributing author (at SNL) directly via e-mail.
```
@@ -85,21 +84,33 @@
 
   // fp = derivative of embedding energy at each atom
   // phi = embedding energy at each atom
-  // if rho > rhomax (e.g. due to close approach of two atoms),
-  //   will exceed table, so add linear term to conserve energy
+  // if rho > rhomax or rho < rhomin (i.e., table is exceeded),
+  //   add linear term to conserve energy
 
   for (ii = 0; ii < inum; ii++) {
     i = ilist[ii];
-    p = rho[i]*rdrho + 1.0;
+    p = (rho[i]-rhomin)*rdrho + 1.0;
     m = static_cast<int> (p);
-    m = MAX(1,MIN(m,nrho-1));
+    if (m < 2) {
+      m = 2;
+    } else if (m > nrho-1) {
+      m = nrho-1;
+    }
     p -= m;
-    p = MIN(p,1.0);
+    if (p < -1.0) {
+      p = -1.0;
+    } else if (p > 1.0) {
+      p = 1.0;
+    }
     coeff = frho_spline[type2frho[type[i]]][m];
     fp[i] = (coeff[0]*p + coeff[1])*p + coeff[2];
     if (eflag) {
       phi = ((coeff[3]*p + coeff[4])*p + coeff[5])*p + coeff[6];
-      if (rho[i] > rhomax) phi += fp[i] * (rho[i]-rhomax);
+      if (rho[i] < rhomin) {
+        phi += fp[i] * (rho[i]-rhomin);
+      } else if (rho[i] > rhomax) {
+        phi += fp[i] * (rho[i]-rhomax);
+      }
       phi *= scale[type[i]][type[i]];
       if (eflag_global) eng_vdwl += phi;
       if (eflag_atom) eatom[i] += phi;
```

athomps:
Thanks for the diff. Yes, this is a little more than I realized. It could be made to work using the template trick, but propagating that change through the derived classes would create a lot of extra code churn.  I am happy with this version.

athomps (approved):
I approve

doc/src/pair_eam.rst
akohlmey: done

PR #2558, opened 2021-01-15
Replace redundant 3-vector function definitions in pair styles with MathExtra functions

### LINKED ISSUE #2550: [Feature Request] Consolidate redundant "vec3_XXX()" functions from manybody styles to use MathExtra:: equivalents
**Summary**

Several manybody potentials have definitions of simple 3-vector operations coded redundantly in their header files
while the same functions are also available in the MathExtra namespace (also as inline functions). We should get rid of this redundancy.

akohlmey:
**Summary**

This replaces redundant definitions of 3-vector convenience functions from class headers and the "math_vector.h" include file with their equivalents in the MathExtra namespace. This improves consistency and any future performance improvements in the MathExtra routines will translate to all code using them and avoids having to propagate such optimizations to all implementations.

With this change I see a small (1-2%) performance improvement, presumably because of the more liberal use of const qualifiers and having `scale3(const double s, double *v)` in addition to `scale3(const double s, const double *in, double *out)`
of the case of `in == out`.

**Related Issue(s)**

This implements and thus closes #2550 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

Yes. All modified styles except smtbq (and that only used the trivial dot3()) have unittests which pass.

**Implementation Notes**

A couple of overloads were added to MathExtra to accommodate the different calling sequences. Also a few const qualifiers were added in order to help the compiler to optimize the code.

Also "math_vector.h" and "math_complex.h" (only a subset of those was used anyway) are removed and replace with functionality in MathExtra. A few definitions and functions from those headers, that were used in kspace style ewald/disp only, were copied locally into the corresponding .cpp file and the header file cleaned up to use forward declarations.

**Post Submission Checklist**


akohlmey:
@stanmoore1 this is not touching any KOKKOS code since everything seems to be defined separately for KOKKOS and thus the changes to the non-KOKKOS styles did not cause any problems with KOKKOS as far as I can tell.

athomps (approved):
This is a good change.  I looked at trying to simplify some of the Tersoff code by adding a function `inline double len_normalize3(const double *v, double *ans);` but it would not help much.

PR #2562, opened 2021-01-22
Fix issues in Kokkos package with pair/only on

stanmoore1:
**Summary**

Fix issues in Kokkos package with `pair/only on`. In some places, views were being sync'd to Device instead of the template parameter, and some lambdas were running with the default execution space instead of the template parameter.

**Related Issue(s)**

None

**Author(s)**

Stan Moore (SNL), reported by Evan Weinberg (NVIDIA) @weinbe2

**Licensing**

**Backward Compatibility**

Yes

### DIFF
--- a/src/KOKKOS/fix_dpd_energy_kokkos.cpp
+++ b/src/KOKKOS/fix_dpd_energy_kokkos.cpp
@@ -58,7 +58,8 @@ void FixDPDenergyKokkos<DeviceType>::take_half_step()
 
   auto dt = update->dt;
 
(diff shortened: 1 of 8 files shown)


weinbe2:
Not that I expected any surprises since we were looking at similar benchmarks, but I confirmed everything works and gives consistent results between:
* Fully running on the device
* Host offload via `pair/only on`
* Targeted offload via replacing `shake` with `shake/kk/host` in the input script.

Thanks for the quick fix @stanmoore1

akohlmey (approved):

PR #2564, opened 2021-01-26
Detect UTF-8 multibyte characters input and replace with ASCII equivalents where known

### LINKED ISSUE #2563: [Feature Request] Handling UTF-8 characters
**Summary**

The use of non-ASCII UTF-8 characters is becoming common and we should make a decision for a policy to deal with them.

**Detailed Description**

Particularly when people are using cut-n-paste from PDF file and websites it is possible they create input data that LAMMPS cannot process because PDF viewers and web browsers keep replacing ASCII characters with equivalent UTF-8 encoded characters for better visual appeal. Examples are spaces with different width or using the (wider) math symbol '-' sign instead of the ASCII '-' character.

Currently LAMMPS will error out with some rather confusing error message (especially when echoing the failing string, which looks correct).

I see three possible steps in order to mitigate this.
1. We could test all lines/words of input before they are parsed whether they contain non-ASCII characters and print a warning
2. We could replace known UTF-8 characters with their ASCII counterpart
3. We could force using the "C" locale when LAMMPS is running and that should make non-compliant characters more visible.

We could be doing any combination of those. I have experimented with the first two and it would be simple to also add a call to
```
#include <clocale>

setlocale(LC_ALL,"C");
```
to either `main()` or `LAMMPS::LAMMPS()` and perhaps in the case of a latter providing an option to control any locale/UTF-8 related handling for or even during a run.

Thoughts? Alternative suggestions?

- https://en.wikipedia.org/wiki/UTF-8
- https://www.fileformat.info/info/charset/UTF-8/list.htm
- [Feature branch with a functions to detect and replace non-ASCII UTF-8 characters](https://github.com/lammps/lammps/compare/master...akohlmey:detect-utf8-multibytes)

akohlmey:
**Summary**

This adds two utility functions `utils::has_utf8()` and `utils::utf8_subst()` that can detect if non-ASCII characters are present in a string and try to replace cases where ASCII equivalents are known. This modifies the `utils::*numeric()` functions, the `ValueTokenizer` class and `Input::parse()` to make use of those functions and thus transparently capture and correct most of those cases. A warning is printed when the input contains non-ASCII, non-comment characters.

**Related Issue(s)**

Closes #2563 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

Yes, all unit tests pass.

**Implementation Notes**

The `utils::has_utf8()` function is implemented to be inline and fast, but may trigger on non-UTF-8 non-ASCII characters.
This is reasonable, since those would be very rare and should trigger non-ASCII warnings anyway.

**Post Submission Checklist**

### DIFF
--- a/doc/src/Commands_parse.rst
+++ b/doc/src/Commands_parse.rst
@@ -162,3 +162,26 @@ LAMMPS:
    triple quotes can be nested in the usual manner.  See the doc pages
    for those commands for examples.  Only one of level of nesting is
    allowed, but that should be sufficient for most use cases.
+
+.. admonition:: ASCII versus UTF-8
+        :class: note
+
+   LAMMPS expects and processes 7-bit ASCII format text internally.
+   Many modern environments use UTF-8 encoding, which is a superset
+   of the 7-bit ASCII character table and thus mostly compatible.
+   However, there are several non-ASCII characters that can look
+   very similar to their ASCII equivalents or are invisible (so they
+   look like a blank), but are encoded differently.  Web browsers,
+   PDF viewers, document editors are known to sometimes replace one
+   with the other for a better looking output.  However, that can
+   lead to problems, for instance, when using cut-n-paste of input
+   file examples from web pages, or when using a document editor
+   (not a dedicated plain text editor) for writing LAMMPS inputs.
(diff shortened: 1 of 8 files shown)


athomps (approved):
I approve

PR #2565, opened 2021-01-26
Explicit netcdf dumper error message when trying to append to non-existant file

prs513rosewood:
**Summary**

Previous behavior would print "at keyword requires use of 'append yes'" when trying to append to non-existant file, which was confusing. It now prints "cannot append to a non-existant file `filename`".

**Author(s)**

Lucas Frérot, Department of Physics and Astronomy, Johns Hopkins University

**Licensing**

**Backward Compatibility**

Should be ok.

**Implementation Notes**

I make use of `utils::file_is_readable` instead of `access(2)` from `unistd.h`, and use of `fmt::format` to print the filename.

**Post Submission Checklist**


akohlmey:
@prs513rosewood thanks for the change. This looks nice. Getting rid of "unistd.h" is a good thing as will improve portability.

sjplimp (approved):

PR #2570, opened 2021-01-29
Add new Package USER-RANN with pair style rann for using a neural network to compute energies and forces

kipbarrett:
**Summary**

Pair rann is a pair style which uses a feed forward neural network to compute atom energies and forces. We are requesting that it be included as a new USER package in LAMMPS. 

**Related Issue(s)**

N/A

**Author(s)**

Lead developer:             Christopher Barrett <barrett@me.msstate.edu
with contributions by      Doyl Dickel <doyl@me.msstate.edu>

**Licensing**

**Backward Compatibility**

Should be fully compatible.

**Implementation Notes**

1) Fingerprints and Activations are designed with a modular framework based on the way pairs, atom styles, fixes, computes, etc. are used in LAMMPS. This is so that more types of fingerprints and activations can easily be added later. 

2) We designed some fingerprints to interact with the SPIN package. The non-spin fingerprints and the rest of the code are fully functional without the SPIN package enabled. If spin fingerprints are used, the code will ensure that atom style spin is being used. If the SPIN package is not installed, the existing code already prevents atom style spin from being enabled.

3) We provided four fitted potentials in the potentials folder for examples, and detailed documentation.

**Post Submission Checklist**


akohlmey:
@kipbarrett thanks for your submission.
I have not looked at this too closely, but from what I saw, there will be some requests for moving files and some changes in the organization of the implementation coming. Some trivial, some not so trivial, and some where we need to have some internal discussions among the core LAMMPS developers first. Since we are currently busy with preparing the next patch release, this will have to wait a while before it we can do all of this. I will try to get a first round of (mostly cosmetic) suggestions out to you as quickly as possible, since it will probably take some time to get all of those sorted out.

src/fingerprint_bond.cpp
```diff
@@ -0,0 +1,917 @@
...
+	delete [] coeff;
+	delete [] coeffx;
+	delete [] coeffy;
+	delete [] coeffz;
+	delete [] Mf;
+	delete [] rinvsqrttable;
+}
+
+bool Fingerprint_bond::parse_values(char * constant, char * line1){
+	char **words=new char *[MAXLINE];
```
akohlmey, line 74: since we moved to C++11 as minimum C++ standard requirement, we are also switching away from some of the C-string string functions and use the C++ std::string class instead. Comparisons become more readable, search and substitution requires less code and functions that are not thread-safe like strtok() can be avoided. strtok() in particular is something that we want to get rid of and not add new uses to the code.
For parsing/processing text files there are utility functions in the utils:: namespace and particular there are the Tokenizer and ValueTokenizer classes that are already applied to a bunch of parsers for reading potential files.

src/fingerprint_bond.cpp
```diff
@@ -0,0 +1,917 @@
...
+	else if (strcmp(constant,"dr")==0){
+		dr = strtod(words[0],NULL);
+	}
+	else if (strcmp(constant,"k")==0){
+		k = strtol(words[0],NULL,10);
+	}
+	else if (strcmp(constant,"m")==0){
+		m = strtol(words[0],NULL,10);
+	}
+	else pair->errorf("Undefined value for bond power");
```
akohlmey, line 102: This kind of error condition should either throw an exception or directly call `Error::one()` or `Error::all()`, you should be able to access those via `pair->lmp->error->`.

src/fingerprint_bond.cpp
```diff
@@ -0,0 +1,917 @@
...
+void Fingerprint_bond::allocate(){
+	generate_exp_cut_table();
+	generate_coefficients();
+	generate_rinvssqrttable();
+}
+
+//Generate table of complex functions for quick reference during compute. Used by do3bodyfeatureset_singleneighborloop and do3bodyfeatureset_doubleneighborloop.
+void Fingerprint_bond::generate_exp_cut_table(){
+  int m,n;
+	double r1;
```
akohlmey, line 183: there is inconsistent indentation here. as a reminder, in LAMMPS indentations are 2 spaces per level, no TABs, and no whitespace at the end of a line.

src/fingerprint_bond.h
```diff
@@ -0,0 +1,80 @@
...
+	double *expcuttable;
+    double *dfctable;
+    double dr;
+    double *alpha_k;
+    double re;
+    int **coeff;
+    int **coeffx;
+    int **coeffy;
+    int **coeffz;
+    int k;
```
akohlmey, line 69: please do not use class members with such trivial names and particularly not characters that are common to be used for loop indices (i,j,k,l,m,n)

src/fingerprint_bond.cpp
```diff
@@ -0,0 +1,917 @@
...
+				dfeaturesx[jnum*f+count]-=dfeaturesx[jj*f+count];
+				dfeaturesy[jnum*f+count]-=dfeaturesy[jj*f+count];
+				dfeaturesz[jnum*f+count]-=dfeaturesz[jj*f+count];
+				count++;
+			}
+		}
+	}
+}
+
+int Fingerprint_bond::factorial(int n) {
```
akohlmey, line 912: we have an optimized `factorial()` function in the MathSpecial namespace. Please use that one.

src/fingerprint_bond.cpp
```diff
@@ -0,0 +1,917 @@
...
+						dfeaturesz[kk*f+count]+=dot1*(m*c12+c62);
+						dot1*=dot;
+						features[count++]+=2*dot1;
+					}
+				}
+			}
+			kk=jj;
+			if (ktypes == nelements || ktypes == jtype){
+				count = startingneuron;
+				double dot = (y[jj][0]*y[kk][0]+y[jj][1]*y[kk][1]+y[jj][2]*y[kk][2]);
```
akohlmey, line 799: Please note that there are a whole set of functions for operations on 3-vectors in the MathExtra namespace.

src/pair_rann.h
```diff
@@ -0,0 +1,171 @@
...
+#ifdef PAIR_CLASS
+
+PairStyle(rann,PairRANN)
+
+#else
+
+#ifndef LMP_PAIR_RANN
+#define LMP_PAIR_RANN
+
+#define MAXLINE 1024
```
akohlmey, line 40: the define and the following include files should be moved from the header to the .cpp file. only the essential include statements like for pair.h should remain. Please see https://github.com/lammps/lammps/blob/master/doc/include-file-conventions.md for some explanations.

src/pair_rann.h
```diff
@@ -0,0 +1,171 @@
...
+  ~PairRANN();
+  void compute(int, int);
+  void settings(int, char **);
+  void coeff(int, char **);
+  void init_style();
+  double init_one(int, int);
+  void init_list(int , NeighList *);
+  void errorf(const char*);
+  int count_words(char *);
+  //black magic for modular fingerprints and activations
```
akohlmey, line 77: this should be done differently. please see the prototype code provided elsewhere.

src/pair_rann.h
```diff
@@ -0,0 +1,171 @@
...
+  bool allscreen;//all fingerprints use screening so screened neighbors can be completely ignored
+  bool dospin;
+  int res;//Resolution of function tables for cubic interpolation.
+  int memguess;
+	double *screening_min;
+	double *screening_max;
+	bool **weightdefined;
+	bool **biasdefined;
+
+	struct Simulation{
```
akohlmey, line 110: this and the following struct do not need to be define here. a forward declaration should be sufficient. the full struct can be defined in the implementation.

src/pair_rann.cpp
```diff
@@ -0,0 +1,1247 @@
...
+   init for one type pair i,j and corresponding j,i
+------------------------------------------------------------------------- */
+
+double PairRANN::init_one(int i, int j)
+{
+  return cutmax;
+}
+
+void PairRANN::errorf(const char * message){
+	this->error->all(FLERR,message);
```
akohlmey, line 1209: please note that `Error::all()` may **only** be used for cases where **all** MPI ranks call it at the same time. Otherwise, `Error::one()` must be used or else LAMMPS will "hang".

src/pair_rann.cpp
```diff
@@ -0,0 +1,1247 @@
...
+				dsz[jj][j] = dssumz[jj][j];
+			}
+		}
+	}
+}
+
+
+
+//treats # as starting a comment to be ignored.
+int PairRANN::count_words(char *line){
```
akohlmey, line 1170: there is no need for a custom function here. there are several convenience functions for parsing and custom string processing in the utils namespace and you may also switch to use the Tokenizer/ValueTokenizer class for even more efficient and compact processing of parameter files.

src/pair_rann.cpp
```diff
@@ -0,0 +1,1247 @@
...
+	int *type = atom->type;
+	i1=ilist[ii];
+	itype = map[type[i1]];
+	NNarchitecture net1 = net[itype];
+//	numneigh = listfull->numneigh;
+//	jnum = numneigh[ilist[ii]]+1;//extra value on the end of the array is the self term.
+//	firstneigh = listfull->firstneigh;
+//	jlist = firstneigh[i1];
+	int L = net1.layers-1;
+	double layer[net1.maxlayer];
```
akohlmey, line 1060: these "fortran style" variable length array definitions on the stack are not acceptable and not compliant with C++ standards. They are accepted by GNU C++ compilers only in standard mode with accepts GNU extensions.
Since `jnum` can be large, they must not be allocated on the stack in the first place. Most Linux installations have limits on the stack anyway and thus would cause crashes. Please see the Memory class with its create(), grow(), and destroy() functions. You should also set things up so you avoid frequent re-allocations by storing a `jnummax` variable and then call "memory->grow()" only when jnum is larger than jnummax. there are plenty of example uses of this code construct elsewhere in LAMMPS.

src/fingerprint_bond.cpp
```diff
@@ -0,0 +1,917 @@
...
+//						char str[128];
+//						sprintf(str,"%f %f %f\n",Bb[a2],yprod,y4[M[a2]]);
+//						std::cout<<str;
+//					}
+					yprod *= y4[M[a2+1]];
+				}
+			}
+			if (atomtypes[1]!=atomtypes[2]){//Bb!=Bg
+				double Bg[mb];
+//				double dBgx;
```
akohlmey, line 458: please remove all commented out code. if there is debug code that should remain, please place it into
```
#if DEBUG_RANN
// debug code for XXX
[...]
#endif
```
and then place a `#define DEBUG_RANN 0` in a prominent position at the beginning of the file.

akohlmey (changes requested):
As promised, here is a first round of suggestions, mostly of the cosmetic kind, but also several related to programming style and code layout and proper integration into LAMMPS.

This should become a package USER-RANN and thus all files need to be
moved to src/USER-RANN, the package must be added to the conventional
make and the CMake based build system, the package must be documented
in the relevant places in the manual, the docs for pair style rann should
mention the need to install the package under Restrictions, and there
should be 1-2 small example inputs/logs in the examples/USER/rann folder.

There should be a new namespace RANN inside of LAMMPS_NS for all
functions and classes inside the USER-RANN package that are not the
pair style. Correspondingly the "fingerprint*.*" and "activation*.*"
files should be renamed to "rann_fingerprint*.*" and "rann_activation*.*,
respectively.

Please only use C-style stdio for informational output and only use
the predefined "screen" or "logfile" FILE pointers. Unless desired
otherwise, should should protect this from producing multiple output
when running in parallel by bracketing this with `if (comm->me == 0)`
For formatted output using sprintf() to a buffer should be avoided
and instead either fmt::format() or fmt::print() can be used.

Please note that for and if statements should be formatted as follows:
```
for (init; condition; increment) {
  ...
}

if (condition) {
  ...
}
```
i.e. with `) {` and not `){`. If the body is a single line,
the curly braces may be omitted (better not if there also is an else)
and if the line is very short (e.g. a break or continue statement,
or a single assignment) the body may be on the same line.

The use of this-> should be avoided where possible, and if this is
required for disambiguation, the variable or parameter names should
be renamed. You also should avoid having local variable names
"shadow" class member names.

The re-use of the implementation strategy of the factory pattern for
pair styles and so on is a very nice idea, but a) it would only make sense
to do it in this style if people would want to include/exclude those
style all the time and b) if there was the corresponding support in the
two build systems. As it is, this pattern collides with some of the
assumptions made in the build process, and thus we better find a way to
do the same thing without the preprocessor magic. At the link below you will find a minimal demo implementation that showcases how this could look like.

[factory-demo.tar.gz](https://github.com/lammps/lammps/files/5901108/factory-demo.tar.gz)

You will find a whole bunch of individual suggestions attached to individual
locations in the added source files.

kipbarrett:
Hi Axel,

Thanks for the quick response. I'll try to get these changes tended to over the next few days.

Best regards,

Christopher D Barrett
Assistant Professor
Department of Mechanical Engineering
Phone: 662-325-7325
Email: barrett@me.msstate.edu<mailto:barrett@me.msstate.edu>
Website: http://www.me.msstate.edu/people/christopher-barrett/<http://www.me.msstate.edu/people/christopher-barrett/>
Office: 213 Carpenter Hall
Mississippi State University
________________________________

src/activation.h
```diff
@@ -0,0 +1,52 @@
...
+the Engineer Research and Development Center (ERDC).  The work described in
+this document was conducted at CAVS, MSU.  Permission was granted by ERDC
+to publish this information. Any opinions, findings and conclusions or
+recommendations expressed in this material are those of the author(s) and
+do not necessarily reflect the views of the United States Army.​”
+
+DISTRIBUTION A. Approved for public release; distribution unlimited. OPSEC#4918
+ */
+
+#ifndef ACTIVATION_H_
```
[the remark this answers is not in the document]
kipbarrett: Done

src/activation.h
```diff
@@ -0,0 +1,52 @@
...
+#define ACTIVATION_H_
+
+#include "pair_rann.h"
```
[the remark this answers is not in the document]
kipbarrett: Done

src/activation.h
```diff
@@ -0,0 +1,52 @@
...
+
+namespace LAMMPS_NS {
+
+	class Activation {
```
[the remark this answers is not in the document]
kipbarrett: I believe I have this resolved everywhere now.

src/activation_linear.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+
+using namespace LAMMPS_NS;
+
+Activation_linear::Activation_linear(PairRANN *pair) : Activation(pair){
+	empty = false;
+	style = "linear";
+}
+
+double Activation_linear::activation_function(double A)
```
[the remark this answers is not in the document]
kipbarrett: I put all the activations as headers only. This should be generally appropriate for activations.

src/activation_sigI.cpp
```diff
@@ -0,0 +1,52 @@
+/* -*- c++ -*- ----------------------------------------------------------
```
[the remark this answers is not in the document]
kipbarrett: Done

src/fingerprint_bond.cpp
kipbarrett: I have tried to make all the fingerprints and activations rely only on things defined by pair_rann. This is because I want to use the same code with minimal changes in my calibration routines outside of LAMMPS. For that reason, I put a wrapper around the error function, created a struct with the NeighborList information, etc. I've modified the error function now to pass the correct line and file to LAMMPS Error::one().

src/fingerprint_bond.cpp
kipbarrett: Done. I created a wrapper for it in pair_rann.

src/fingerprint_bond.cpp
kipbarrett: This is the most computationally intensive part of the code. I think adding function calls for clarity is not worthwhile here.

src/pair_rann.h
kipbarrett: I've made some changes to this which resolve the conflicts with the two build systems. I haven't done it exactly the way you suggested though. The reason why is that I would prefer that adding fingerprints/activations require as little modification of existing code as possibly. Reusing the LAMMPS style approach was a good way to do that. 

I agree that there is no reason to have LAMMPS build the style files on each compile. Instead, I suggest that developers can simply add new fingerprints/activations manually to the list of headers in rann_list_*. Changing these names to not use style_* fixes the build conflicts. Hopefully this approach is ok.

src/pair_rann.h
kipbarrett: I tried to get this to work without success. I was able to compile pair_rann.cpp with it defined as you suggest, but the fingerprint files would not compile since they also reference the structs. Maybe I'm not doing it correctly.

src/pair_rann.cpp
kipbarrett: I've changed all calls to Error::one().

src/pair_rann.cpp
kipbarrett: I removed this function.

src/pair_rann.cpp
kipbarrett: I've changed all of these to use the LAMMPS memory class. Unfortunately this seems to slow computation by more than 10% compared to the non-compliant method.

src/pair_rann.h
kipbarrett: Done.

src/fingerprint_bond.h
kipbarrett: Done.

src/fingerprint_bond.cpp
kipbarrett: I've removed all uses of strtok and used the Tokenizer instead. I did not use the text file reader because it has the line limit of 1024 hard coded. This conflicts with my potential file convention which has formatted matrices. The weight matrices can have more than 60 columns for alloys which with 16 digit precision will exceed that line length.

src/fingerprint.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+DISTRIBUTION A. Approved for public release; distribution unlimited. OPSEC#4918
+ */
+
+#include "fingerprint.h"
+
+
+using namespace LAMMPS_NS;
+
+Fingerprint::Fingerprint(PairRANN *pair)
```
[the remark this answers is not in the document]
kipbarrett: Done.

akohlmey:
Thanks for your changes. Unfortunately, they address some problems, but also contain new problems. I am tied up with teaching for the next couple of weeks, so it will likely take a while before I can look at this again properly and provide feedback and more suggestions.

akohlmey:
@kipbarrett after some thinking we will insist on removing that convoluted factory code with complex pre-processing and a function pointer map and replace it with something simpler. We have just done the same thing for the KIM package and in the process figured out a cleaner way to do this, and I am actively thinking about ways to do a similar thing to the code in the core of LAMMPS. It is in its current form mostly as an intermediate step for a better solution from a previous implementation that would break with some compilers.  Please find attached a revised demonstration of how this could be done instead. Please note that I removed the "empty" boolean as it is redundant. Its function could be replaced by checking whether the pointer to the instance is NULL or not. Please also note that the activation function should be a pure function in the base class as the base class should not be used.

[factory-demo.tar.gz](https://github.com/lammps/lammps/files/6037247/factory-demo.tar.gz)

src/USER-RANN/pair_rann.h
```diff
@@ -0,0 +1,174 @@
...
+    void compute(int, int);
+    void settings(int, char **);
+    void coeff(int, char **);
+    void init_style();
+    double init_one(int, int);
+    void init_list(int , NeighList *);
+    void errorf(const char*,int,const char*);
+    int factorial(int);
+    //black magic for modular fingerprints and activations
+    class RANN::Activation ***activation;
```
akohlmey, line 69: no "class" needed here since there is already a forward declaration above.

src/USER-RANN/pair_rann.h
```diff
@@ -0,0 +1,174 @@
...
+    double **Biases;
+    int *activations;//unused
+    int maxlayer;//longest layer (for memory allocation)
+    };
+
+    Simulation *sims;
+    NNarchitecture *net;//array of networks, 1 for each element.
+
+   private:
+    template <typename T> static RANN::Fingerprint *fingerprint_creator(PairRANN *);
```
akohlmey, line 143: these two lines should also be removed

src/USER-RANN/pair_rann.h
```diff
@@ -0,0 +1,174 @@
...
+    template <typename T> static RANN::Activation *activation_creator(PairRANN *);
+    //new functions
+    void allocate(std::vector<std::string>);//called after reading element list, but before reading the rest of the potential
```
akohlmey, line 146: passing containers by values are a *very* bad idea. This forces needless copies. Those should be const references.

src/USER-RANN/pair_rann.h
```diff
@@ -0,0 +1,174 @@
...
+    void read_file(char *);//read potential file
+    void strip_comments(char *,FILE *fp);
+    void read_atom_types(std::vector<std::string>,std::vector<std::string>);
```
akohlmey, line 149: are all of these functions really needed as class members? couldn't those just be static functions in the class implementation that take a pointer to the data they should update?

src/USER-RANN/pair_rann.cpp
```diff
@@ -0,0 +1,1261 @@
...
+        break;
+      }
+    }
+    strip_comments(linetemp,fp);
+    line1=linetemp;
+    Tokenizer values = Tokenizer(line,": ,\t_\n");
+    Tokenizer values1 = Tokenizer(line1,": ,\t_\n");
+    linev = values.as_vector();
+    line1v = values1.as_vector();
+    if (linev[0].compare("atomtypes")==0)read_atom_types(linev,line1v);
```
akohlmey, line 325: why use std::string::compare()? it would be much more readable to use `if (linev[0] == "atomtypes") { `
also the lack of a blank between the conditional and the function calls makes this part of the file difficult to read.

src/USER-RANN/pair_rann.cpp
```diff
@@ -0,0 +1,1261 @@
...
+    ptr = fgets(line,4096,fp);
+    if (ptr==NULL) error->one(FLERR,"Unexpected end of RANN file");
+    utils::trim_comment(line);
+    if (strlen(line) == 0) continue;
+    comment = false;
+    }
+  }
+}
+
+void PairRANN::read_atom_types(std::vector<std::string> line,std::vector<std::string> line1) {
```
akohlmey, line 369: why make this a function? this would be simpler (and better to follow) to just do in place. BTW: appending an item to a vector is simply done with `vector::push_back()`.

src/USER-RANN/pair_rann.cpp
```diff
@@ -0,0 +1,1261 @@
...
+  line1[nwords] = "all";
+  allocate(line1);
+}
+
+void PairRANN::read_mass(std::vector<std::string> line,std::vector<std::string> line1) {
+  if (nelements == -1)error->one(FLERR,"atom types must be defined before mass in potential file.");
+  int nwords = 0,i;
+  for (i=0;i<nelements;i++) {
+    if (line[1].compare(elements[i])==0) {
+      mass[i]=strtod(line1[0].c_str(),NULL);
```
akohlmey, line 383: please use `utils::numeric()` for conversions of strings to numbers.

kipbarrett:
The factory code has now been replaced with the suggested hardcoded form. I've also dealt with most of the other comments.

src/USER-RANN/pair_rann.h
kipbarrett: Technically, no, they're not necessary as class members. However, static functions would require a lot of rewriting and I would argue it makes the code more convoluted and harder to read, especially since many of them call each other. I can't see that there is a strong reason to change this.

src/USER-RANN/pair_rann.cpp
kipbarrett: It is easier to read the potential file parser if all the pieces of the file are read in a consistent way. Thus, each section of the file has its own function and the read_file function can have a nice compact block of conditionals to check where to direct each pair of lines for parsing. This one is pretty short, but it is better to leave it alone. vector::push_back() has now been implemented.

kipbarrett:
@akohlmey and @sjplimp,

Thanks for looking over this and the pertinent suggestions. Hopefully this version is closer to being ready add to LAMMPS. @akohlmey, I'm not sure what you mean by a "Fortran feel". The code is written primarily to be as fast and flexible as possible for a neural network. I'm hopeful we can have it included soon since the new potentials we've developed for this are proving to work really well and we'd like to be able to share them more easily.

akohlmey:
@kipbarrett please let us know of your plans for this pull request. It currently fails to build for some configurations and has a merge conflict with cmake. If you need help or explanations or how to find the details of the build failures, please let us know.

Please note that we currently have multiple pull requests with neural network based pair styles pending and ideally we want to do the final review and processing of all of them for the same patch release so responses may be slow at times while we focus on ongoing refactoring work in the code base.

kipbarrett:
Hi Axel,

Some guidance on where the build failures occur would be helpful. I haven't found detailed logs for where it is failing on github, and on my machine it builds fine with both make and cmake. Also, I don't understand the merge conflict with cmake. The only thing I changed in cmake is adding the package name in the appropriate file.

akohlmey:
> Hi @akohlmey ,
> 
> Some guidance on where the build failures occur would be helpful. I haven't found detailed logs for where it is failing on github, and on my machine it builds fine with both make and cmake. 

the build failures are consistently with cmake only. most likely there is are/files that you have removed locally but they are not removed in the pull request branch that cause the conflict. the message is more-or-less the same. please check out the console log for one run here:
https://ci.lammps.org/blue/organizations/jenkins/dev%2Fpull_requests%2Fcentos7%2Fcmake_mpi_openmp_bigbig_static/detail/cmake_mpi_openmp_bigbig_static/2257/pipeline

> Also, I don't understand the merge conflict with cmake. The only thing I changed in cmake is adding the package name in the appropriate file.

There likely have been changes to upstream that produce the conflict. You could try doing a `git pull git://github.com/lammps/lammps.git master` and then resolve the resulting merge conflict and push that updated version back to github.

kipbarrett:
Thanks @akohlmey, 

That resolved all the build failures.

kipbarrett:
Hi @akohlmey ,

Could we get a status update?

akohlmey:
> Hi @akohlmey ,
> 
> Could we get a status update?

The pull request is still pending and I have not had the time to look into it. this is mostly because it will take a significant amount of time to get things properly checked and to look into better alternatives to elements that I believe can be improved.
Unfortunately, there is currently nobody else that has the time necessary to look into pull requests at that level and also we have a bunch of rather urgent projects that need to be completed before the stable release, the publication of a new LAMMPS paper, and the LAMMPS workshop this summer, so those are taking precedence.

Your pull request is not alone in that state. There are a few more on the pending list that have been sitting there even longer.

I understand that it is frustrating to see other pull requests going in while you are waiting on yours. Yet, there is little that can currently be done about it unless we have more people joining the core development team. In fact, if it had not been for the pandemic, we would be behind much more.

akohlmey:
@kipbarrett i've made some formal changes to the pull request to update it to the latest upstream version, resolve pull request, have it pass through integration testing and fix up some format and style issues. I will be looking into some code refactoring next, but I wanted to touch base with you before doing that.

There are two things that I would like your feedback on:

1) We have now an agreed upon "clang-format" configuration for source code in the "src" folder. this is not mandatory, but it helps keeping things consistent (even though not everything remains readable, so different parts need to be protected with '// clang-format off/on' comments. The default setting is to enable clang-format processing for .h files and only selectively for .cpp files. If you have clang-format (version 8 or later) installed on your machine, then please update your local branch of the pull request and try running 'clang-format -i <some .cpp file>' (after removing the '// clang-format off' comment at the top and see if you are comfortable with those changes or would prefer to keep the files as is. I have processed one pair for files for demonstration (`rann_fingerprint.cpp/.h`).

2) we have the convention of labeling potential files with useful information (units, date, authors, citation) as a comment in the first line and also check the UNITS: tag against the units setting in LAMMPS. I thing that would be a good thing for rann pair style potentials to follow. Related to that, I would also prefer to rename the potential to end in .rann instead or .nn to stick with the convention for most potential files.

Please let me know what your take on these issues are.

kipbarrett:
Hi @akohlmey 

I'm fine with both of these suggestions, as long as the clang formatting doesn't break anything. I'll update the potential files as soon as I can. Do you want me to do the clang formatting, or just do it yourself since you're already working in the code?

akohlmey:
> Hi @akohlmey
> 
> I'm fine with both of these suggestions, as long as the clang formatting doesn't break anything. I'll update the potential files as soon as I can. Do you want me to do the clang formatting, or just do it yourself since you're already working in the code?

It is trivial for me to do the conversion while editing files, since I have my editor configured to run it on a keyboard shortcut.
That also allows to quickly undo the change, if the result is not desirable.

I will generate reference output for the provided example and confirm that any changes (be it from clang-format, be it from my refactoring attempts) will not change the results. I will also add a tag to the current state to allow to roll back any changes to this state, since the current state fulfills the minimum formal requirements to be merged. I still need to review the code for functional issues, but that will be effectively included in the planned refactoring to take better advantage of the convenience functions and classes included in LAMMPS and avoid function calls and code constructs we want to get rid of where possible.

akohlmey (approved):

akohlmey:
@kipbarrett change of plans. After a discussion with @sjplimp and considering some "logistical issues" related to planned changes, we will be merging this PR as is today and then do a patch release later today. The refactoring, reformatting and other changes will then be proposed in a later pull request, *after* the reorganization of the LAMMPS source tree.

sjplimp (approved):

sjplimp:
@kipbarrett Thanks for contributing this to LAMMPS and working with us so it could be refactored to be more LAMMPS compatible.

PR #2573, opened 2021-02-01
Ls potential

Kurufinve:
**Summary**

**Related Issue(s)**

**Author(s)**

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**


akohlmey:
@Kurufinve thank you for your submission. I can already see from a cursory look, that we will need multiple review passes and suggestions for changes to adapt your contribution so that it fits in well with the rest of LAMMPS. I will start by making some general suggestions and ask a few questions:

- what does "LS" stand for? is this some established name? if yes, where and how? is there something that is longer and thus more recognizable? we have LJ, EAM, and MEAM are well established, but the rest is mostly described with a full or shortened word (morse, gauss, coul, tersoff, etc.). There are a few exceptions, but in general we prefer more descriptive names for styles and particularly for packages.
- there doesn't seem to be a documentation for the pair style. nothing gets added to LAMMPS without documentation. for a package you need to have two pieces of documentation: a) the documentation for the styles inside the package and b) the instructions for including the package and specific requirements, authorship etc. This all needs to be cleanly integrated so that there are no spelling, or link or other issues. When you type: "make html" in the doc folder, you will see some automated check of what is missing.
- is there a way to drop the dependency on LAPACK? if I read the code correctly the only use are calls to DGESV to compute spline coefficients, and those are usually calls that would be done during initialization of a calculation, so would it be possible to replace the call to this function with a standalone local copy of a simple, non-optimized code (as some static functions in the .cpp file)?

In the next round (probably tomorrow), I will go over the source files and add comments there.

akohlmey:
Also please complete the pull request description test at the top of the pull request with a summary, a list of author(s) and affiliation(s) and (permanent) email contact, and references to relevant publications, if any.

Kurufinve:
Dear Axel,

First of all, I would like to apologize for making this incomplete pull
request. It was an accidental request because I simply wanted to merge two
branches in my forked version of LAMMPS.
However, I planned to make a pull request with this contribution to the
main branch of LAMMPS in the (probably, near) future.
Maybe it would be better to decline this pull request and review the second
one, which I will make in the future when all requirements will be met,
especially given the fact that this pair style wasn't tested in the
parallel mode yet.

Meanwhile, I'll try to answer your questions:
1. what does "LS" stand for? is this some established name? if yes, where
and how?
   LS stands for the first letters of the second names of authors
(Lipnitskii and Savelyev) of the ansatz and method for construction of the
interatomic potentials (the original article describing this method is in
pinned message).
   The full name Lipnitskii-Savelyev is too long, so I decided to shorten
it to the LS. If you suggest a more suitable and not long name for this new
pair style or suggest retaining full author second names in the name of the
pair style, I'll gladly change it.
2. The documentation will be added, of course.
3. Initially I tried to drop the dependency on LAPACK by writing dgesv and
its dependencies in c++ as local functions of PairLS class, but it was not
too easy, and I decided to retain this dependency because installation of
lapack on the local machine is very simple for the end-user (only one
command is needed: sudo apt install liblapack-dev, for example), while on
the most of HPCs the lapack library is often present.

Sincerely yours,
Daniil Poletaev


вт, 2 февр. 2021 г. в 02:07, Axel Kohlmeyer <notifications@github.com>:

akohlmey:
> Meanwhile, I'll try to answer your questions: 1. what does "LS" stand for? is this some established name? if yes, where and how? LS stands for the first letters of the second names of authors (Lipnitskii and Savelyev) of the ansatz and method for construction of the interatomic potentials (the original article describing this method is in pinned message). The full name Lipnitskii-Savelyev is too long, so I decided to shorten it to the LS. If you suggest a more suitable and not long name for this new pair style or suggest retaining full author second names in the name of the pair style, I'll gladly change it.

you can leave it, if you want. just keep in mind that a package or pair style with the name "ls" is not very recognizable and thus will reduce its visibility. perhaps you can find something else that is more descriptive, perhaps something based on the methodology or concept used or its application. my interest is that users that would benefit from features in LAMMPS can find it and that with a name like "ls" your contribution will be hiding in plain sight.

> 3. Initially I tried to drop the dependency on LAPACK by writing dgesv and its dependencies in c++ as local functions of PairLS class, but it was not too easy, and I decided to retain this dependency because installation of lapack on the local machine is very simple for the end-user (only one command is needed: sudo apt install liblapack-dev, for example), while on the most of HPCs the lapack library is often present.

It is not about availability. there are other packages using it and a (suitable) LAPACK subset is even included in the lib/linalg folder of the LAMMPS distribution. The concern is that without the LAPACK dependency the new style would not need to be in a package of its own, but could just be added to USER-MISC (like many other, similar, single style contributions). This also would make it more likely that people would have the pair style included in their binaries, BTW.

As requested, I will close this PR now.

PR #2581, opened 2021-02-04
Charge regulation

tc387:
**Summary**

We implemented a fix that performs Monte Carlo sampling for charge regulation. For example, the fix can sample acid/base reactions within coarse-grained simulations. 
The implementation has been used to study charge-regulation in self-assembly (https://arxiv.org/abs/2009.08544) just accepted in PRL. We would now like to make the implementation available to the community. Moreover, we also wrote a manuscript that details the MC sampling method and its implementation in LAMMPS.

We believe the fix should be of broad interest to the LAMMPS community. Research into charge regulation is picking up.  For example, a similar method (grand reaction-ensemble method) is implemented in EspressoMD package.  However, LAMMPS currently does not have the ability to simulate simple acid/base ionization processes. 

**Related Issue(s)**

**Author(s)**

Tine Curk, Northwestern University
Email: curk@northwestern.edu, tcurk5@gmail.com

Jiaxing Yuan, Shanghai Jiao Tong University

**Licensing**

**Backward Compatibility**

The fix is only compatible with the latest LAMMPS version on GitHub. Older LAMMPS versions use "force->numeric" for parsing input parameters, which seems to have been recently changed to "utills::numeric". 

**Implementation Notes**

No changes to existing LAMMPS files are necessary. Only the file pair (fix_charge_regulation.cpp, fix_charge_regulation.h) has been added.
The fix performs charge transfer reactions. The Monte Carlo sampling alters the charge value of specified atom types and performs grand-canonical counter-ion insertion/deletion. The ion exchange part is loosely based on fix_gcmc from the MC package. 

We would like this fix to be initially included in the USER-MISC package. However, if possible, we hope the fix can be made part of the MC package since it performs MC sampling of basic reactions. 

**Post Submission Checklist**


akohlmey:
This pull request is incorrect and merging it would delete LAMMPS. Please check out https://docs.lammps.org/Howto_github.html and try again.

PR #2582, opened 2021-02-04
Pair kim refactor

ellio167:
**Summary**

Fix up some behavior of pair_kim to better support changing parameters of a KIM model and to correctly support the use of multiple calls to `pair_style kim`
**Related Issue(s)**

n/a

**Author(s)**

Ryan S. Elliott (UMinn)
Yaser Afshar (UMinn)

**Licensing**

**Backward Compatibility**

no issues

**Implementation Notes**

use increments of `instance_me` to ensure neighbor list requests are updated appropriately.

### DIFF
--- a/src/KIM/pair_kim.cpp
+++ b/src/KIM/pair_kim.cpp
@@ -877,7 +877,7 @@ void PairKIM::kim_init()
     &modelWillNotRequestNeighborsOfNoncontributingParticles);
   if (neighborLists) {
     delete [] neighborLists;
-    neighborLists = nulptr;
+    neighborLists = nullptr;
   }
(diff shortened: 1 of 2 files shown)


src/KIM/pair_kim.cpp
```diff
@@ -859,7 +877,7 @@ void PairKIM::kim_init()
     &modelWillNotRequestNeighborsOfNoncontributingParticles);
   if (neighborLists) {
     delete [] neighborLists;
-    neighborLists = 0;
+    neighborLists = nulptr;
```
akohlmey, line 880: There is a typo here.

sjplimp (approved):

Issue #2525, opened 2020-12-16
[Feature Request] Reorganize package names and redistribute some styles, especially from USER-MISC


akohlmey:
**Summary**

There is a need to reorganize the location of some styles for consistency and better maintainability.
Especially the USER-MISC package is getting too big and we should consider creating, e.g. a USER-MANYBODY and/or a USER-MOLECULE and/or a USER-KSPACE package. Similarly, some of the less commonly used styles in src/ could be moved to MISC (or USER-MISC?).

When discussing with contributors it is getting sometimes difficult to justify why they are asked to do things differently than it was for other contributions.

When shuffling things around (and thus confusing some users), we should also consider renaming a few more packages with respect to their status as "user" package (to have the confusion only once and not twice or more often). USER-COLVARS, USER-REAXC, USER-MEAMC, and possibly also USER-OMP could lose the USER prefix and LATTE, MSCG, POEMS, and MPIIO look like it would be justified to add a USER prefix when looking at the maintenance status and type of contribution. I have already multiple times voiced my concern that the designation of user packages as given in the relevant documentation and web pages does not match well with the current reality.

----

After offline discussions with @sjplimp the proposal is now the following:
- we rename some packages that have their names for historical reasons (e.g. USER-MEAMC to MEAM, USER-REAXC to REAX or REAXFF), also file names, class names and style name may be renamed (with backward compatibility, as already available in meam/c)
- we review the contents of USER-MISC and move styles that comply with strict coding standards of the core LAMMPS code to existing packages (e.g. MOLECULE, KSPACE, MANYBODY, MISC) where applicable. 
- less commonly used styles in src are moved to appropriate packages (e.g. MISC)
- we create a few new packages to group thematically styles that don't fit well elsewhere and otherwise would stay in USER or go to MISC. that could be thermostat/time-integrator fixes, pair styles, computes/fixes for analysis.
- new contributions from users that are just a single style are initially added to the USER-MISC package in the same way as is done with USER-MISC right now. depending on the details of the implementation and the application another package may be suggested.
- the conditions for creating a new package are either 1) 3 or more styles that are thematically closely related where the purpose does not match well with existing packages or 2) code that depends on an external or bundled library or use a different programming language (e.g. Fortran, CUDA, etc.)

----

#### Renamed packages
- MLIAP -> **ML-IAP**
- SNAP -> **ML-SNAP**
- USER-ADIOS -> ADIOS
- USER-ATC -> ATC 
[...]

PR #2585, opened 2021-02-05
Charge regulation2

tc387:
**Summary**

We implemented a fix that performs Monte Carlo sampling for charge regulation. For example, the fix can sample acid/base equilibria within coarse-grained simulations.
The implementation has been used to study charge-regulation in self-assembly (https://arxiv.org/abs/2009.08544) just accepted in PRL. We would now like to make the implementation available to the community. Moreover, we also wrote a manuscript that details the MC sampling method and its implementation in LAMMPS.

We believe the fix should be of broad interest to the LAMMPS community. Research into charge regulation is picking up. For example, a similar method (grand reaction-ensemble method) is implemented in EspressoMD package. However, LAMMPS currently does not have the ability to simulate simple ionization processes.

**Related Issue(s)**

**Author(s)**

Tine Curk, Northwestern University
Email: curk@northwestern.edu, tcurk5@gmail.com

Jiaxing Yuan, Shanghai Jiao Tong University

**Licensing**

**Backward Compatibility**

**Implementation Notes**

No changes to existing LAMMPS files are necessary. Only the file pair (fix_charge_regulation.cpp, fix_charge_regulation.h) has been added.
The fix performs charge transfer reactions. The Monte Carlo sampling alters the charge value of specified atom types and performs grand-canonical counter-ion insertion/deletion. The ion exchange part is loosely based on fix_gcmc from the MC package.

We would like this fix to be initially included in the USER-MISC package. However, if possible, we hope the fix can be made part of the MC package since it performs MC sampling of basic reactions.

**Post Submission Checklist**


src/USER-MISC/fix_charge_regulation.cpp
```diff
@@ -0,0 +1,1335 @@
...
+   the GNU General Public License.
+
+   See the README file in the top-level LAMMPS directory.
+------------------------------------------------------------------------- */
+
+/* ----------------------------------------------------------------------
+   Contributing author: Tine Curk (tcurk5@gmail.com) and Jiaxing Yuan (yuanjiaxing123@hotmail.com)
+------------------------------------------------------------------------- */
+#include "fix_charge_regulation.h"
+#include <mpi.h>
```
akohlmey, line 18: Please rearrange the include files to match our conventions (header for implementation, then LAMMPS headers sorted alphabetically, then system headers w/o the headers that are included by default) as outlined in https://github.com/lammps/lammps/blob/master/doc/include-file-conventions.md

See `src/MC/fix_widom.cpp` for example for a similar file that is fully compliant with those conventions.

doc/src/fix_charge_regulation.rst
```diff
@@ -0,0 +1,184 @@
+
+.. Yuan documentation master file, created by
+   sphinx-quickstart on Sat Jan 30 14:06:22 2021.
+   You can adapt this file completely to your liking, but it should at least
+   contain the root `toctree` directive.
+   tc387: Multiple text additions/changes, Feb 2 2021
+.. index:: fix fix_charge_regulation
```
akohlmey, line 7: This is giving the fix name as fix_charge_regulation which is incorrect and thus produces an incorrect index.

doc/src/fix_charge_regulation.rst
```diff
@@ -0,0 +1,184 @@
+
+.. Yuan documentation master file, created by
+   sphinx-quickstart on Sat Jan 30 14:06:22 2021.
```
akohlmey, line 3: Best to remove this personal comment.

src/USER-MISC/fix_charge_regulation.h
```diff
@@ -0,0 +1,117 @@
...
+   Contributing author: Tine Curk (tcurk5@gmail.com) and Jiaxing Yuan (yuanjiaxing123@hotmail.com)
+------------------------------------------------------------------------- */
+
+#ifdef FIX_CLASS
+
+FixStyle(charge_regulation,Fix_charge_regulation)
+
+#else
+
+#ifndef LMP_FIX_charge_regulation_H
```
akohlmey, line 24: This should be all caps: LMP_FIX_CHARGE_REGULATION_H

src/USER-MISC/fix_charge_regulation.h
```diff
@@ -0,0 +1,117 @@
...
+   See the README file in the top-level LAMMPS directory.
+------------------------------------------------------------------------- */
+
+/* ----------------------------------------------------------------------
+   Contributing author: Tine Curk (tcurk5@gmail.com) and Jiaxing Yuan (yuanjiaxing123@hotmail.com)
+------------------------------------------------------------------------- */
+
+#ifdef FIX_CLASS
+
+FixStyle(charge_regulation,Fix_charge_regulation)
```
akohlmey, line 20: The name of this fix style should be `charge/regulation` and the name of the class should be `FixChargeRegulation` to be consistent with the conventions applied elsewhere in LAMMPS. This change needs - of course - be propagated to all relevant files in this pull request.

src/USER-MISC/fix_charge_regulation.cpp
```diff
@@ -0,0 +1,1335 @@
...
+#include "neighbor.h"
+
+using namespace LAMMPS_NS;
+using namespace FixConst;
+using namespace MathConst;
+
+// large energy value used to signal overlap
+#define MAXENERGYSIGNAL 1.0e100
+#define MAXENERGYTEST 1.0e50
+#define small 0.0000001
```
akohlmey, line 53: Please make all defines uppercase (i.e. use SMALL instead of small).

src/USER-MISC/fix_charge_regulation.cpp
```diff
@@ -0,0 +1,1335 @@
...
+#define PI 3.1415926
```
akohlmey, line 54: Please do not define your own (and rather imprecise) version of PI and use MY_PI from the MathConst namespace instead.

src/USER-MISC/fix_charge_regulation.cpp
```diff
@@ -0,0 +1,1335 @@
...
+  int mask = 0;
+  mask |= PRE_EXCHANGE;
+  return mask;
+}
+
+void Fix_charge_regulation::init() {
+
+  triclinic = domain->triclinic;
+
+  char *id_pe = (char *) "thermo_pe";
```
akohlmey, line 143: This line is not needed since `Modify::find_compute()` will (now) accept constant strings, so you can use the constant directly in the call in the next line.

src/USER-MISC/fix_charge_regulation.cpp
```diff
@@ -0,0 +1,1335 @@
...
+
+  if (domain->dimension == 2)
+    error->all(FLERR, "Cannot use fix charge regulation in a 2d simulation");
+
+  // create a new group for interaction exclusions
+  // used for attempted atom deletions
+  // skip if already exists from previous init()
+
+  if (!exclusion_group_bit) {
+    char **group_arg = new char *[4];
```
akohlmey, line 168: This segment can be simplified due to recent improvements of the core LAMMPS code. Please see fix widom.

src/USER-MISC/fix_charge_regulation.cpp
```diff
@@ -0,0 +1,1335 @@
...
+    error->warning(FLERR,
+                   "reaction distance (rxd) is larger than half the box dimension, resetting default: xrd = 0.");
+    reaction_distance = 0;
+  }
+  // volume in units of (N_A * mol / liter)
+  volume_rx = (xhi - xlo) * (yhi - ylo) * (zhi - zlo) * pow(llength_unit_in_nm, 3) * 0.602214;
+  if (reaction_distance < small) {
+    vlocal_xrd = volume_rx;
+  } else {
+    vlocal_xrd = 4.0 * PI * pow(reaction_distance, 3) / 3.0 * pow(llength_unit_in_nm, 3) * 0.602214;
```
akohlmey, line 270: Please do not use pow() on expressions with integer exponents. There is a more accurate (and faster) function called `powint()` in the MathSpecial namespace. For powers of 2 or 3 there also are `square()` and `cube()`, respectively.

src/USER-MISC/fix_charge_regulation.cpp
```diff
@@ -0,0 +1,1335 @@
...
+
+  if ((reaction_distance > fabs(domain->boxhi[0] - domain->boxlo[0]) / 2) ||
+      (reaction_distance > fabs(domain->boxhi[1] - domain->boxlo[1]) / 2) ||
+      (reaction_distance > fabs(domain->boxhi[2] - domain->boxlo[2]) / 2)) {
+    error->warning(FLERR,
+                   "reaction distance (rxd) is larger than half the box dimension, resetting default: xrd = 0.");
+    reaction_distance = 0;
+  }
+  // volume in units of (N_A * mol / liter)
+  volume_rx = (xhi - xlo) * (yhi - ylo) * (zhi - zlo) * pow(llength_unit_in_nm, 3) * 0.602214;
```
akohlmey, line 266: What does this constant 0.602214 represent? I have an educated guess, but this should be using a symbolic name defined at the top of the file so that it is explicit and obvious to the reader.

src/USER-MISC/fix_charge_regulation.cpp
```diff
@@ -0,0 +1,1335 @@
...
+  pKb = 100;
+  pKs = 14.0;
+  nevery = 100;
+  nmc = 100;
+  pmcmoves[0] = pmcmoves[1] = pmcmoves[2] = 0.33;
+  llength_unit_in_nm= 0.72;
+
+  reservoir_temperature = 1.0;
+  reaction_distance = 0;
+  seed = 12345;
```
akohlmey, line 1190: please use a default value for the seed that is not valid in the following check, e.g. 0.

src/USER-MISC/fix_charge_regulation.cpp
```diff
@@ -0,0 +1,1335 @@
...
+  pI_minus = 100;
+  acid_type = -1;
+  base_type = -1;
+  pKa = 100;
+  pKb = 100;
+  pKs = 14.0;
+  nevery = 100;
+  nmc = 100;
+  pmcmoves[0] = pmcmoves[1] = pmcmoves[2] = 0.33;
+  llength_unit_in_nm= 0.72;
```
akohlmey, line 1186: what does this number, 0.72, represent?

src/USER-MISC/fix_charge_regulation.cpp
```diff
@@ -0,0 +1,1335 @@
...
+  pI_plus = 100;
+  pI_minus = 100;
+  acid_type = -1;
+  base_type = -1;
+  pKa = 100;
+  pKb = 100;
+  pKs = 14.0;
+  nevery = 100;
+  nmc = 100;
+  pmcmoves[0] = pmcmoves[1] = pmcmoves[2] = 0.33;
```
akohlmey, line 1185: if this is supposed to be 1/3rd please use THIRD from the MathConst namespace.

src/USER-MISC/fix_charge_regulation.cpp
```diff
@@ -0,0 +1,1335 @@
...
+/* ----------------------------------------------------------------------
+   parse input options
+------------------------------------------------------------------------- */
+
+void Fix_charge_regulation::options(int narg, char **arg) {
+  if (narg < 0) error->all(FLERR, "Illegal fix charge regulation command");
+
+  // defaults
+
+  pH = 7.0;
```
akohlmey, line 1175: I see a lot of expressions like `pow(10,-pXXX)` or `pow(10,-pXXX+pYYY)` throughout the code. Since the `pow()` function is among the most expensive functions in the standard math library, one can sometimes achieve a significant performance boost by pre-computing and storing those values in stead of the "natural" ones and thus avoid having to use `pow()` a lot during the evaluation of the body of this fix but either use the pre-computed value directly or product/fraction in case exponents are added or subtracted.

akohlmey (changes requested):
Thanks for resubmitting. This looks much more like it should be. ;-)
Please find some comments, mostly about cosmetic and style issues, scattered through individual parts of the code. Please have a look and apply the suggested changes or provide an explanation why this would be a problem or ask questions if you need more explanations from me.

I have not yet compiled the code in the PR or ran any of the examples. So there will be a second iteration of review comments once the current set has been resolved.

akohlmey:
@athomps this seems to be another fix style that uses fix gcmc as a template, similar to fix widom and thus also a candidate to have it moved to the MC package. From my POV it is a less obvious choice because of the more complex implementation, but I have reviewed only formal issues so far. Can you have a look at the realization of the algorithm and the user interface and portability (e.g. across units and other settings like newton on/off etc.) or running in parallel. I would expect that most of the concerns in fix gcmc would apply here as well because of its heritage.

doc/src/fix_charge_regulation.rst
```diff
@@ -0,0 +1,184 @@
...
+
+**(Curk1)** T. Curk, J. Yuan, and E. Luijten, "Coarse-grained simulation of charge regulation using LAMMPS", preprint (2021).
+
+.. _Curk2:
+
+**(Curk2)** T. Curk and E. Luijten, "Charge-regulation effects in nanoparticle self-assembly", PRL (2021)
+
+.. _Landsgesell:
+
+**(Landsgesell)** J. Landsgesell, P. Hebbeker, O. Rud, R. Lunkad, P. Kosovan, and C. Holm, “Grand-reaction method for simulations of ionization equilibria coupled to ion partitioning,” Macromolecules 53, 3007–3020 (2020).
```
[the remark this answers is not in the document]
tc387: done.

src/USER-MISC/fix_charge_regulation.cpp
tc387: done.

doc/src/fix_charge_regulation.rst
tc387: done.

akohlmey:
@tc387 thanks for the updates and improvements. I've made a few minor corrections, simplifications, changes to reduce compiler warnings, and changes to be closer aligned with the usual LAMMPS programming style, reformatted and spellchecked the docs and updated the branch to include the latest changes from upstream. I've also reformatted and shortened the example runs to be more similar to what is usually used in LAMMPS. So from a formal perspective the code is now set up to be included.

I am now handing off this PR to @athomps and @sjplimp for further review and to decide whether this should be moved to the MC package or left in USER-MISC.

tc387:
@akohlmey Thank you for your speedy corrections and improvements!

akohlmey:
@athomps @sjplimp can you please have a look at this pull request? thanks.
Since this borrows a lot from fix gcmc, we should consider moving it to the MC package.

sjplimp:
MC package would be fine with me.

athomps (changes requested):
This is a nice capability.  It has a lot in common with more conventional GCMC simulations.  I think it should be included in the MC package.  Perhaps the authors can make that change.  Following on Axel's comments, it would be a good idea to translate one of the examples to a different unit style, such as real, to verify that the various unit conversion factors (boltz, mvv2e, etc.) are being applied corectly, as many of these are unity for unit style lj.

tc387:
An example using real units has been added as suggested by @athomps. The conversion factors seem to be applied correctly; using lj or real units leads to the same charge distribution.

akohlmey (approved):

akohlmey:
@tc387 I've updated the PR to the current upstream, moved it to the MC package, added a few changes to the documentation that are required for a full integration and fixed a few issues flagged by compiler warnings and running with valgrind. I have not moved the examples because they will be moved anyway when we implement issue #2525, which will be rather soon.

athomps (approved):
Looks good to me.

PR #2586, opened 2021-02-06
Final small changes and bugfixes for the patch release.

akohlmey:
**Summary**

This PR collects bugfixes for recent refactoring, changes indicated from static code analysis that were missed by previous "bulk" replacements, and some small documentation updates and corrections.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

- fix for an indexing bug for fix ave/histo introduced in #2571
- document the `-DCUDA_PROXY` define for compiling the GPU library with conventional make, add Makefile.cuda_mps to lib/gpu to demonstrate
- fixes for a few issues where intermediate results for computing timestep series and ranges could overflow before cast to bigint
- rearrange a few expressions computing address offsets to avoid 32-bit integer overflows
- silence CodeQL reports about 32-bit integer overflows of intermediate results
- update CMake and GNU make configuration files for the GPU package to include the consumer level amber architecture
- fix some memory leaks and initialization issues reported by valgrind
- fix issue with `install-python` custom target not using ${LAMMPS_MACHINE} to name the shared library to install

**Post Submission Checklist**


sjplimp (approved):

akohlmey:
> Sorry, my bad. I didn't think that in a_b_c, a*b would be computed as int before the second multiplication happens.

Don't sweat it. I expect that the compiler will get it right with optimization enabled anyway. It is more a matter of silencing CodeQL to have important warnings not drown in a flood of unimportant ones. Same as we do with compiler warnings. Most of the CodeQL related changes in this PR are of a similar kind, and it took me multiple iterations to figure out all necessary changes (granted the first two rounds were done with sed, so those were bound to miss a few cases, but that still got rid of over 200 warnings).

Issue #2588, opened 2021-02-08
[BUG] _Python Package_


finkeljos:
**Summary**

The LAMMPS Python package does not appear to update atoms positions or velocities when invoking the python/move fix when running the in.fix_python_move_nve_melt example provided in the ~/lammps/examples/python folder

**LAMMPS Version and Platform**

I am running in my local cluster environment which uses CentOS 7 and using the current master branch for LAMMPS.

**Expected Behavior**

I expect to see a typical MD simulation thermo output.

**Actual Behavior**

No reported errors from LAMMPS but the kinetic energy remains fixed for length of the example simulation and atoms do not appear to move.

**Steps to Reproduce**

I executed the following shell commands in this order

1. git clone
2. cd ~/lammps/src
3. make yes-python
4. make mode=shared serial -j
5. make install-python
6. export PYTHONPATH=${PYTHONPATH}:${HOME}/lammps/python
7. export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:${HOME}/lammps/src
8. cd ~/lammps/examples/python/
9. ~/lammps/src/lmp_serial < in.fix_python_move_nve_melt

giacomofiorin:
@rbberger To complete what Josh has reported so far, I built LAMMPS (both `master` and `stable_29Oct2020`) on a different machine also CentOS 7, with Python 3.6 detected from CMake and `-DPYTHON=on`.

I was able to build LAMMPS fine, but with the same example I got:
```
AttributeError: .../build/lammps/stable_29Oct2020/bin/lmp: undefined symbol: lammps_extract_setting
```

rbberger:
@finkeljos please attach a log output

rbberger:
@giacomofiorin you sure you're loading the correct .so file?

finkeljos:
@rbberger Here is the log output I get
[fix_py_move.log](https://github.com/lammps/lammps/files/5947372/fix_py_move.log)

rbberger:
@finkeljos Which version of python are you running? Also, doing both `make install-python` and setting `PYTHONPATH` and `LD_LIBRARY_PATH` at the same time is a bit weird. The first copies the python package in your user installation (`$HOME/.local/lib/pythonX.Y/site-packages/` by default), which should be automatically visible in your python installation and the latter points back to the compilation of the source folder.

finkeljos:
@rbberger I have the anaconda module loaded on my node, it's using Python 3.7.4.

[...]

PR #2589, opened 2021-02-09
Add version check for liblammps vs. python module

akohlmey:
**Summary**

This PR adds a `__version__` global variable to the LAMMPS python module. The version is reset to the (numerical representation of the) current LAMMPS version. Upon creating a LAMMPS instance those two versions are compared
and if they do not match, and AttributeError exception is raised.

**Related Issue(s)**

This is to prevent library/module version mismatches as discussed in #2588

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Post Submission Checklist**


akohlmey:
@rbberger can you think of a way to have `__version__` in `lammps/__init.py__` updated **after** it is copied into its final location?
But then again, perhaps we should just update it with each patch release, and not automatically update it.

rbberger (changes requested):
I'd rather not make changes to python files by some mechanism. But instead use the package metadata:

```python
# python >= 3.8
from importlib.metadata import version
version('lammps')

# python < 3.8
import pkg_resources
pkg_resources.get_distribution('lammps').version
```

rbberger (approved):
better, thanks!

PR #2592, opened 2021-02-10
Fix a few python module/interface issues with python2.x

### LINKED ISSUE #2588: [BUG] _Python Package_

### LINKED ISSUE #2593: [Feature Request] Capture exceptions when calling Python from LAMMPS and report back
**Summary**

Currently all calls to evaluate python code from LAMMPS in the PYTHON package ignore the return value.
That means errors are simply ignored and the evaluation will be aborted. That makes it difficult to debug the python code called from LAMMPS. For an example see Issue #2588 

**Detailed Description**

Somehow we need to capture any exceptions **inside** the python code (perhaps with wrapper function?) and could store its information in a global python variable, that we then read in case the call to evaluate python returns a NULL.
Then we can call `Errorr::all()` and report that since with the additional information that the Python evaluation failed.

akohlmey:
**Summary**

This addresses an ambiguity issue when importing "numpy", where python 2.7 imports a different module than python 3.x.
This is done by renaming file in the lammps package and thus removing this particular ambiguity.
It didn't happen earlier since all components were in a single file, lammps.py.

It also avoids a byte-compilation error when installing the package under python 2.7 by simply not including mliap support with python 2.7. python 3.x will be required for using that. The docs say the python 3.6 is the required minimum, anyway, so the CMake script for MLIAP was updated to check for that and refuse to try compiling python support into MLIAP for a python version that does not match that minimum requirement.

It also adds error checking and printing of the python error message and abort for fix python/move and fix python/invoke.

**Related Issue(s)**

This closes #2588 
This also closes #2593

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Post Submission Checklist**


sjplimp (approved):

giacomofiorin:
I'm doing onsite maintenance work today, so I can't test the code but otherwise looks good.

PR #2595, opened 2021-02-10
cmake: update lib soversion

junghans:
**Summary**

Use the numerical date string as so version for the library.

**Related Issue(s)**

https://github.com/lammps/lammps/issues/2588#issuecomment-775558250

**Author(s)**

@junghans

**Licensing**

**Backward Compatibility**

Nope

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -10,7 +10,6 @@ endif()
 ########################################
 
 project(lammps CXX)
-set(SOVERSION 0)
 
 get_filename_component(LAMMPS_DIR ${CMAKE_CURRENT_SOURCE_DIR}/.. ABSOLUTE)
 get_filename_component(LAMMPS_LIB_BINARY_DIR ${CMAKE_BINARY_DIR}/lib ABSOLUTE)
@@ -41,6 +40,7 @@ endif()
 include(LAMMPSUtils)
 
 get_lammps_version(${LAMMPS_SOURCE_DIR}/version.h PROJECT_VERSION)
+get_lammps_version(${LAMMPS_SOURCE_DIR}/version.h SOVERSION)
 
 include(PreventInSourceBuilds)
 


akohlmey:
This is a CMake-only solution. We don't want different behavior for something like this between CMake and GNU make. That will make the documentation a big mess. I have spent a considerable effort (programming and arguing with people) to have the GNU make process more aligned with CMake, so that behavior is more predictable and consistent and it is easier to explain in the documentation. I don't want to lose that again.

junghans:
Well, CMake always had an soversion on the library and make never did, so this change doesn't make it anymore different.

akohlmey:
> Well, CMake always had an soversion on the library and make never did, so this change doesn't make it anymore different.

Then we should rather get rid of it in CMake for as long the traditional make is a fully supported feature.

junghans:
> > Well, CMake always had an soversion on the library and make never did, so this change doesn't make it anymore different.
> 
> Then we should rather get rid of it in CMake for as long the traditional make is a fully supported feature.

Well, we shouldn’t do that either, because having versioned libraries is a requirement for packages on the rpm-based distros. (Not sure about Ubuntu, but as we are only in a ppa there, it is the wild west anyway.)

We could add it to make (http://www.microhowto.info/howto/build_a_shared_library_using_gcc.html), but the so version is only useful then LAMMPS is installed, which cannot be done with make currently.

akohlmey:
> > > Well, CMake always had an soversion on the library and make never did, so this change doesn't make it anymore different.
> > 
> > 
> > Then we should rather get rid of it in CMake for as long the traditional make is a fully supported feature.
> 
> Well, we shouldn’t do that either, because having versioned libraries is a requirement for packages on the rpm-based distros. (Not sure about Ubuntu, but as we are only in a ppa there, it is the wild west anyway.)

The "laws of packaging" are not relevant here. Package maintainers are **not** our target audience. I don't mind if this improves the situation for a "normal" LAMMPS user, but we're not doing things for the sake of whatever some committee has decided how the world of software has to look like. For a package you can always include a patch in the package build scripts to comply with whatever rules and guidelines the distributions come up. Package maintainers should know how to do that.

We've been through this before, I don't want to do the whole discussion all over again.

junghans:
This patch isn't about packaging, we are good on that front.

The change is about making two installs of different versions distinguishable at the linker level, and yes this PR only improves the current status on the CMake side, but we can do the same on Make side if you like.

giacomofiorin:
Throwing my two cents here: putting the issue of packaging completely aside, there is still the problem of how to distinguish shared objects representing the same LAMMPS versions, but different build configurations (compilers, flags, package selections, etc).

It's just simpler to keep all LAMMPS builds separate, in distinct installation paths.

akohlmey:
> This patch isn't about packaging, we are good on that front.
> 
> The change is about making two installs of different versions distinguishable at the linker level, and yes this PR only improves the current status on the CMake side, but we can do the same on Make side if you like.

It does not improve anything where it matters. 1) the issue you are referring to was not at all about the shared library but due to the recent python module refactoring. 2) the LAMMPS python module does not know anything about an SOVERSION but will look for a ".so" or ".dylib" or ".dll" file depending on the platform, so setting/changing the SONAME does not do anything there. When you do "make install-python" it will copy a file to a .so filename into the site-packages sub-folder, and if not found there will look it up through LD_LIBRARY_PATH or DYLD_LIBRARY_PATH or %PATH%. So having a liblammps.so.something with the "something" changing all the time is making things more complicated (people who do "make install" will just accumulate copies with different LAMMPS versions they install). They need the symbolic link to the .so file anyway. The python module will open whatever the link is pointing to, and that is why checking the **internal** version inside the library makes sense.

junghans:
Yeah, you are right, we hard-coded `liblammps.so` in `lammps.py`.

The binary loads the versioned library though:
```
$ ldd lmp | grep lammps
	liblammps.so.0 => /home/junghans/lammps/build/liblammps.so.0 (0x00007fbfeff49000)
```
but as the lib is always called `liblammps.so.0` loading the wrong library has hit be a couple of times before.

Maybe it would just be easier to have `lammps.sh` set the right LD_LIBRARY_PATH.

akohlmey:
> Yeah, you are right, we hard-coded `liblammps.so` in `lammps.py`.
> 
> The binary loads the versioned library though:
> 
> ```
> $ ldd lmp | grep lammps
> 	liblammps.so.0 => /home/junghans/lammps/build/liblammps.so.0 (0x00007fbfeff49000)
> ```

yes, but if you install a new binary, it will overwrite *both*. It still doesn't solve the case of having the same version but different compilation settings etc. with two liblammps.so files in different folder but both added to LD_LIBRARY_PATH. So if you want something that is really unique you would need to generate an soname from a hash computed from CMakeCache.txt *and* the version. But we also already have a mechanism to create unique executables and libraries in CMake: you can set LAMMPS_MACHINE. That is *much* more reliable, since you get to choose what is different and where. 

And you don't even need that right away, you can also modify this after the fact (i.e. shared library dependencies or soname) with a tool like "patchelf".

However I consider it a bad situation when the default behavior is to accumulate many copies of different liblammps.so.<version> files because you overwrite `lmp` but not `liblammps.so.<version>` for every new version you compile and install into the same location (to update your local LAMMPS binary).

> but as the lib is always called `liblammps.so.0` loading the wrong library has hit be a couple of times before.
> 
> Maybe it would just be easier to have `lammps.sh` set the right LD_LIBRARY_PATH.

That doesn't really solve any problem either. Please note that the hack with checking the version number from inside python module is not to guarantee that you get the identical version of liblammps.so that you copied when installing the module, but that you refuse to load a version that may be incompatible due to changes in the library interface and then cause memory corruption or otherwise bad behavior. It cannot protect you from accidentally using a version compiled with different features. For confirming those there are by now a whole lot of introspection features available, though.

In short, this is not going to happen unless you convince @sjplimp to change my mind. We have already spent too much time on this.

PR #2555, opened 2021-01-14
Improvements to the GPU library Makefiles

ndtrung81:
**Summary**

This PR added two Makefiles for CUDA and OpenCL APIs similar to what has been done in Makefile.hip for the HIP API. The underlying Nvidia.makefile and Opencl.makefile are also updated to be less tedious, and more effective for adding new pair styles.

**Related Issue(s)**

N/A

**Author(s)**

Trung Nguyen (Northwestern)

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

Should be maintained.

**Implementation Notes**

Nvidia.makefile is now the same as Nvidia.makefile_multi.

**Post Submission Checklist**

- [x] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [ ] The source code follows the LAMMPS formatting guidelines
- [ ] Suitable new documentation files and/or updates to the existing docs are included
- [ ] The added/updated documentation is integrated and tested with the documentation build system
- [x] The feature has been verified to work with the conventional build system
- [ ] The feature has been verified to work with the CMake based build system
- [ ] Suitable tests have been added to the unittest tree.
- [ ] A package specific README file has been included or updated
- [ ] One or more example input decks are included

**Further Information, Files, and Links**

PR #2603, opened 2021-02-16
GPU Package Update feb2021

wmbrownIntel:
**Summary**

This update includes significant performance improvements for Intel GPUs, significant improvements for 3-body styles on NVIDIA, OpenCL 2 support, usability enhancements and bug fixes. 3-commits separating documentation update, GPU package update, and changes to LAMMPS core files. For the latter, lammps.cpp is modified to issue "package gpu 0" instead of "package gpu 1" with the suffix command, atom.cpp is modified to change the sort bin setup (only when GPU package is installed and active - similar to user-intel package), and std=c++11 is added to a few Makefiles.

- Significant performance improvements for 3-body styles (neighbor and force)
- Optimizations for shared main memory subsystem are now enabled by default for integrated GPUs
- Significantly reduced memory overhead
- New neighbor list build algorithm with automatic bin size heuristic
- Limited OpenMP and SIMD support for host code has been added along with some gpu styles for time integration that run only on the host.
- Reduced host<->device communication for NPT.
- Device ID(s) selection and GPU count are now automatic
-   Package default is now (package gpu 0) with 0 indicating automatic count
- Support for OpenCL versions >= 2 (earlier versions still supported)
- Most all out-of-memory issues should error through the standard LAMMPS error reporting as opposed to cryptic device runtime errors.
- Improved accuracy for device overhead estimation
- Simplified device configuration and updated OpenCL custom configuration parameters
- Subgroup support has been added when OpenCL >= 2 along with horizontal vector operations (shuffle) with Intel OpenCL extensions
- CUDA warp intrinsics now meet requirements for independent thread scheduling (only known manifestation was with Tersoff)
- Explicit CUDA textures are disabled for versions 11 and higher where it is deprecated (no performance impact)
- The 'blocksize' package argument now (correctly) changes the parameters for 3-body and ellipsoid styles. DPD will use the threads_per_atom setting instead of threads_per_charge. Some modifications to tersoff for consistent behavior with CPU when different cutoffs are used for i,j,k triplets.
- Fixed bug for 3-body styles in the case where inum<nlocal and atom sorting is used.
- OpenCL runtime compilation errors are now reported in non-debug builds and a preprocessor option is available to dump all compiled programs, build flags, and log output.
- Fixed memory leak in P3M.
- Fixes for recent changes for tiled P3M and OpenCL makefiles that broke some functionality.
- Added lj/charmm/coul/charmm/gpu.
- Added nve/gpu, npt/gpu, nvt/gpu, and nve/asphere/gpu that still run on the host in full double precision but have some OpenMP and SIMD support for acceleration.
- The 'short' neighbor list optimizations for 3-body are now consistent with what is used for ellipsoid styles.
- Any GPUs/devices with main memory NUMA are automatically partitioned to separate GPU devices.
- The 'device' keyword for the package command is removed; 'platform' and 'device_type' should be used instead.
- Some progress in cleaning up indentation / whitespace issues 

**Related Issue(s)**

**Author(s)**

Mike Brown and Nitin Dhamankar (Intel)

**Licensing**

**Backward Compatibility**

The Ngpu parameter to the package now accepts 0 (for automatic determination of number of GPUs) and this is the default for the '-sf gpu' issue.
The gpuID keyword now only takes 1 argument. Using two was redundant with the Ngpu parameter and awkward with the new auto-selection for devices.
The device keyword is removed (to better facilitate control of auto-device selection). It is replaced with 'platform' and 'device_type' keywords.
The automatic OpenCL platform detection will choose the platform with the best GPU (in terms of compute cores) and understands additional restrictions from Ngpu, gpuID, and device_type keywords.
For OpenCL, device configuration parameters for NVIDIA, AMD, INTEL are automatically determined at runtime without requirement for a default configuration at build time (old OCL_TUNE preprocessor define).
The custom OpenCL configuration string is different.

These changes are updated in the package documentation.

**Implementation Notes**

I put effort into not breaking HIP, but I have no way to test this.

**Post Submission Checklist**


akohlmey:
@wmbrownIntel thanks for your efforts. This looks like a lot of work. @rbberger and me are currently tied up with teaching obligations, so it will take a little while before we can take a closer look and try it out on our hardware. It is probably time to update our GPU singularity containers as well. GPU testing has been taking a backseat recently in integration testing, since we needed to overcome problems for testing on the CPU as well and are struggling with some complications from using jenkins.

It seems there are changes to building with CMake needed, since those runs are currently failing in integration testing. 

@Vsevak would you mind taking a look at how this PR is working with HIP on AMD and whether additional changes are needed. It seems to me, that what is included should provide a performance boost to running on AMD GPUs as well.

sjplimp (approved):
Thanks Mike!

rbberger:
@wmbrownIntel Fixed two minor issues that broke the legacy CI builds and patched up the CMake build. We still need to review the changes to the defines so we can propagate/remove those options in CMake.

rbberger:
@wmbrownIntel just need a clarification about `-fopenmp-simd` and `-qopenmp-simd`. As far as I understand it, `-fopenmp` implies `-fopenmp-simd`. The latter allows you to compile with SIMD without having to link to omp runtime ([link](https://openmpcon.org/wp-content/uploads/openmpcon2015-simd-tutorial.pdf)). So it's not really needed then, correct? Are the semantics for the Intel compiler different?

wmbrownIntel:
> @wmbrownIntel just need a clarification about `-fopenmp-simd` and `-qopenmp-simd`. As far as I understand it, `-fopenmp` implies `-fopenmp-simd` (if OpenMP v4). The latter allows you to compile with SIMD without having to link to omp runtime ([link](https://openmpcon.org/wp-content/uploads/openmpcon2015-simd-tutorial.pdf)). So it's not really needed then, correct? Are the semantics for the Intel compiler different?

@rbberger Thanks for fixing up cmake. Sorry - I haven't started using this yet. My understanding is the same - enabling openmp implicitly enables openmp-simd.

rbberger:
@wmbrownIntel Ok. That means I can keep the options in CMake the same and don't have to probe for `opemp-simd` support. The only other thing I see is some outdated makefiles such as `Makefile.opencl`. I'm now always adding `GERYON_NUMA_FISSION` for OpenCL, not sure if we should make it an option in CMake.

rbberger (approved):

wmbrownIntel:
@rbberger I think it is OK to have the NUMA fission everywhere for now. Sorry, I missed Makefile.opencl since it didn't include Opencl.makefile. You can remove or I can do it. Thanks again.

akohlmey:
> @rbberger I think it is OK to have the NUMA fission everywhere for now. Sorry, I missed Makefile.opencl since it didn't include Opencl.makefile. You can remove or I can do it. Thanks again.

Makefile.opencl is **not** outdated. It was added by @ndtrung81 only very recently during some refactoring of the makefiles in lib/gpu as PR #2555. It is more the other way around that using Opencl.makefile (or Nvidia.makefile) is the outdated way of doing things.

wmbrownIntel:
> > @rbberger I think it is OK to have the NUMA fission everywhere for now. Sorry, I missed Makefile.opencl since it didn't include Opencl.makefile. You can remove or I can do it. Thanks again.
> 
> Makefile.opencl is **not** outdated. It was added by @ndtrung81 only very recently during some refactoring of the makefiles in lib/gpu as PR #2555. It is more the other way around that using Opencl.makefile (or Nvidia.makefile) is the outdated way of doing things.

Makefile.opencl is broken. It doesn't pick up the extra includes needed in the OpenCL program strings for 3-body and ellipsoid styles and it does not handle dependencies correctly forcing developers to clean and rebuild for every change. I modified Opencl.makefile to fix these issues while still using wildcards for reduced maintenance. If there were other enhancements in Makefile.opencl that were missed we can do something to fix but it seems simpler to me to use an Opencl.makefile include rather than replicate everything from that file in Makefile.opencl?

akohlmey:
> > > @rbberger I think it is OK to have the NUMA fission everywhere for now. Sorry, I missed Makefile.opencl since it didn't include Opencl.makefile. You can remove or I can do it. Thanks again.
> > 
> > 
> > Makefile.opencl is **not** outdated. It was added by @ndtrung81 only very recently during some refactoring of the makefiles in lib/gpu as PR #2555. It is more the other way around that using Opencl.makefile (or Nvidia.makefile) is the outdated way of doing things.
> 
> Makefile.opencl is broken. It doesn't pick up the extra includes needed in the OpenCL program strings for 3-body and ellipsoid styles and it does not handle dependencies correctly forcing developers to clean and rebuild for every change. I modified Opencl.makefile to fix these issues while still using wildcards for reduced maintenance. If there were other enhancements in Makefile.opencl that were missed we can do something to fix but it seems simpler to me to use an Opencl.makefile include rather than replicate everything from that file in Makefile.opencl?

That is for @ndtrung81 to answer.

ndtrung81:
I confirm that Makefile.opencl breaks the tersoff* pair styles, which is my overlook when running the tests on my side. I will go with the current corrections to Opencl.makefile and removal of Makefile.opencl. Thanks Mike for pointing this out and fix it.

akohlmey:
@ndtrung81 @wmbrownIntel When compiling this branch on my desktop I see a whole lot of warnings from the OpenCL headers about missing a setting for `CL_TARGET_OPENCL_VERSION` and then announcing it would be set to 220 (i.e. OpenCL v2.2) by default. However, that is inconsistent with the behavior in that it can run with the Nvidia OpenCL, even though that only supports OpenCL up to v2.0. In fact, if I set the define explicitly, the Nvidia device will not even load, complaining about a lack of OPENCL_2.2.

So should there be a configuration setting that allows to choose the OpenCL target version? Or should we just set it to 200? or leave the default and live with those (rather annoying) warnings?

wmbrownIntel:
@akohlmey Yes, this is annoying and a problem I don't currently have an elegant solution for. Is it OK to do the latter (let the compiler decide and possibly generate warnings) and I will talk with an OpenCL standard expert here about the best way to address this in general? I hate to specialize the build for this... but maybe we can disable the warning message somehow.

akohlmey:
> @akohlmey Yes, this is annoying and a problem I don't currently have an elegant solution for. Is it OK to do the latter (let the compiler decide and possibly generate warnings) and I will talk with an OpenCL standard expert here about the best way to address this in general? I hate to specialize the build for this... but maybe we can disable the warning message somehow.

I checked the source code of the files in the /usr/include/CL folder and found there is no other way to bypass this warning than explicitly defining CL_TARGET_OPENCL_VERSION

On the other hand I also found out, that there is an inconsistency on my system. While it finds the libOpenCL.so file in /usr/lib64/ for linking (which is just the OpenCL v2.2 compatible reference "loader" for vendor modules and matches the header), it uses the libOpenCL.so.1 file bundled with the Nvidia driver at runtime (which then fails to load for OpenCL > 2.0). But after resetting the shared library loading order, it will work. So the kernels seem to be smart enough to adapt.

So I am thinking to pick the lowest common denominator (which seems to be OpenCL v2.0) and set this as default like so:
```
  diff --git a/lib/gpu/geryon/ocl_device.h b/lib/gpu/geryon/ocl_device.h
  index b0a3e3d583..5901908e3a 100644
  --- a/lib/gpu/geryon/ocl_device.h
  +++ b/lib/gpu/geryon/ocl_device.h
  @@ -28,6 +28,10 @@
   #include <vector>
   #include <iostream>
   
  +#ifndef CL_TARGET_OPENCL_VERSION
  +#define CL_TARGET_OPENCL_VERSION 200
  +#endif
  +
   #ifdef __APPLE__
   #include <OpenCL/cl.h>
   #include <OpenCL/cl_platform.h>
  diff --git a/lib/gpu/geryon/ocl_macros.h b/lib/gpu/geryon/ocl_macros.h
  index 5fb7665817..c7ca59c702 100644
  --- a/lib/gpu/geryon/ocl_macros.h
  +++ b/lib/gpu/geryon/ocl_macros.h
  @@ -4,6 +4,10 @@
   #include <cstdio>
   #include <cassert>
   
  +#ifndef CL_TARGET_OPENCL_VERSION
  +#define CL_TARGET_OPENCL_VERSION 200
  +#endif
  +
   #ifdef __APPLE__

```
and if that causes problems, the user can just set `CL_TARGET_OPENCL_VERSION` to something else and it will be honored.
If needed we can make it a documented configuration option later, similar to picking Nvidia architecture settings.

wmbrownIntel:
I would rather make the default 210 since we have a lot of checks for that version in the library and the preferred build for NVIDIA should be CUDA since MPS is not supported for OpenCL among other things.

akohlmey:
> I would rather make the default 210 since we have a lot of checks for that version in the library and the preferred build for NVIDIA should be CUDA since MPS is not supported for OpenCL among other things.

Ok, I can do that as well. I know now how to deal with it and that it is ultimately an issue with how my Linux distribution handles Nvidia driver compatibility and not a principal problem. I mostly care about getting rid of the warnings and that happens with either setting. ;-)
Just be mentally prepared that it might be more likely to get complaints about "LAMMPS not working anymore with GPU" with this setting.

Vsevak:
This update is impressive!
@akohlmey The first testing revealed some problems with the HIP. I continue here https://github.com/lammps/lammps/pull/2622#issuecomment-784587805

PR #2610, opened 2021-02-19
Fixing issue from recent GPU package update with OMP_NUM_THREADS env …

wmbrownIntel:
…being overridden in GPU library.

Fixing race condition with OpenMP for GPU styles using torque (missed in regression tests due to the first fix)

Documenting GPU package option for setting the number of threads (consistent with USER-INTEL and USER-OMP).

**Summary**

Bux fix for using GPU package update with multiple OMP threads.

**Related Issue(s)**

**Author(s)**

Mike Brown (Intel)

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/doc/src/package.rst
+++ b/doc/src/package.rst
@@ -32,10 +32,12 @@ Syntax
              size = bin size for neighbor list construction (distance units)
            *split* = fraction
              fraction = fraction of atoms assigned to GPU (default = 1.0)
-           *tpa* value = Nthreads
-             Nthreads = # of GPU vector lanes used per atom
+           *tpa* value = Nlanes
+             Nlanes = # of GPU vector lanes (CUDA threads) used per atom
            *blocksize* value = size
              size = thread block size for pair force computation
+           *omp* value = Nthreads
+             Nthreads = number of OpenMP threads to use on CPU (default = 0)
            *platform* value = id
              id = For OpenCL, platform ID for the GPU or accelerator
            *gpuID* values = id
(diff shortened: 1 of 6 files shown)


doc/src/package.rst
```diff
@@ -619,6 +598,18 @@ input.  Not all features of LAMMPS support OpenMP threading via the
 USER-OMP package and the parallel efficiency can be very different,
 too.
 
+.. note::
+
+   If you build LAMMPS with the GPU, USER-INTEL, and / or USER-OMP
```
akohlmey, line 603: What about KOKKOS? with OpenMP?

wmbrownIntel: Can KOKKOS be used in combination with these packages?

akohlmey: Theoretically it should be possible. Whether it is practical, I do not know since KOKKOS requires many kokkos-ized styles.

wmbrownIntel: @stanmoore1 Hi Stan, on the package doc page we have a warning about setting the number of OMP threads when using multiple packages. KOKKOS hasn't been included in this, but maybe should be. Can you comment on how this should be modified to include KOKKOS:

+   If you build LAMMPS with the GPU, USER-INTEL, and / or USER-OMP
+   packages, be aware these packages all allow setting of the *Nthreads*
+   value via their package commands, but there is only a single global
+   *Nthreads* value used by OpenMP.  Thus if multiple package commands are
+   invoked, you should insure the values are consistent.  If they are
+   not, the last one invoked will take precedence, for all packages.
+   Also note that if the :doc:`-sf hybrid intel omp command-line switch <Run_options>` is used, it invokes a "package intel" command, followed by a
+   "package omp" command, both with a setting of *Nthreads* = 0. Likewise
+   for a hybrid suffix for gpu and omp.

sjplimp (approved):

akohlmey (approved):

doc/src/package.rst
stanmoore1: @wmbrownIntel @akohlmey Sorry for the delay. I've never tried using KOKKOS with other packages, but the same should be true. It is also possible to invoke more than 1 package command for the same package, in which only the last setting will be used. We should also make sure that the user knows the package command will overwrite the `OMP_NUM_THREADS` env variable.

PR #2613, opened 2021-02-19
Fix bug in dump image SSAO depth shading

stanmoore1:
**Summary**

Fix bug in dump image ssao depth shading. If the image width wasn't divisible by the number of MPI ranks, then some rows at the top of the image were excluded from SSAO depth shading. This typically isn't noticed for small proc counts if there is empty space at the top of the image.

**Related Issue(s)**

None.

**Author(s)**

Stan Moore, Steve Plimpton (SNL)

**Licensing**

**Backward Compatibility**

Yes.

**Implementation Notes**

Changes were copied from SPARTA: https://github.com/sparta/sparta/blob/master/src/image.cpp

I visually verified this PR fixes the issue.

### DIFF
--- a/src/image.cpp
+++ b/src/image.cpp
@@ -113,6 +113,11 @@ Image::Image(LAMMPS *lmp, int nmap_caller) : Pointers(lmp)
   backLightColor[2] = 0.9;
 
   random = nullptr;
+
+  // MPI_Gatherv vectors
+
+  recvcounts = nullptr;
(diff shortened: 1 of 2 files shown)


akohlmey:
Looks like this is missing changes to `image.h`

stanmoore1:
Yeah I thought it got added.

sjplimp (approved):

PR #2620, opened 2021-02-23
Refactor kim commands

yafshar:
**Summary**

This PR includes,

- Refactor `kim commands` to a single command with sub-commands following it for different options
   - Update the `kim init` command. Now it sets the skin and timestep default values as 2.0 Angstroms and 1.0
femtosecond for every unit system for kim models
   - Update the `kim query` command. Now, the `kim query` works with the `model` argument and can also be used with no `kim init` call requirement.
   - Update the `kim param` command and make it independent of `kim init` call at the start
   - Update the `kim command` doc and references to it
   - Update the `kim command` examples and log files with new command interface
   - Update the `kim command` unit tests and add extra test cases for better coverage
- Extend the `kim query` command, to get all available models meeting certain requirements. To query for KIM models the query function is `get_available_models`
- Fix some mistakes in the code and clean up the interface

**Related Issue(s)**

**Author(s)**

Yaser Afshar (UMN)
Ryan S. Elliott (UMN)
Ellad Tadmor (UMN)
Axel Kohlmeyer, (Temple U)

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

This PR is **not backward compatible**. It includes refactoring the `kim commands` to a single command with sub-commands following it for different options, which will fail if used with old input files.

**Implementation Notes**

**Post Submission Checklist**

- [ ] The feature or features in this pull request is complete
- [ ] Licensing information is complete
- [ ] Corresponding author information is complete
- [ ] The source code follows the LAMMPS formatting guidelines
- [ ] Suitable new documentation files and/or updates to the existing docs are included
- [ ] The added/updated documentation is integrated and tested with the documentation build system
- [ ] The feature has been verified to work with the conventional build system
- [ ] The feature has been verified to work with the CMake based build system
- [ ] Suitable tests have been added to the unittest tree.
- [ ] A package specific README file has been included or updated
- [ ] One or more example input decks are included

**Further Information, Files, and Links**

PR #2625, opened 2021-02-25
Updating the kim citation report to lammps conventions

yafshar:
**Summary**

This PR includes updating the kim citation report to lammps conventions. 

**Related Issue(s)**

This is reported in https://github.com/lammps/lammps/pull/2620#issuecomment-784271608

**Author(s)**

Yaser Afshar (UMN)

**Licensing**

**Backward Compatibility**

Backward compatible

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/src/KIM/kim_init.cpp
+++ b/src/KIM/kim_init.cpp
@@ -70,7 +70,6 @@
 #include "variable.h"
 
 #include <cstring>
-#include <regex>
 
 extern "C" {
 #include "KIM_SimulatorHeaders.h"
@@ -494,10 +493,10 @@ void KimInit::write_log_cite(char *model_name)
   if (!lmp->citeme) return;
 
   std::string model_name_str(model_name);
-  std::regex re("[MOS]{2}_[0-9]{12}_[0-9]{3}");
-  std::smatch sm;
+  std::string re = "[MS][OM]_\\d\\d\\d\\d\\d\\d\\d\\d\\d\\d\\d\\d_\\d\\d\\d";
+  std::string kim_id = utils::strfind(model_name_str, re);
 
-  if (!std::regex_search(model_name_str, sm, re)) return;
+  if (kim_id.empty()) return;
 
   KIM_Collections *collections;
   int err = KIM_Collections_Create(&collections);
@@ -524,10 +523,8 @@ void KimInit::write_log_cite(char *model_name)
     return;
   }
 
-  auto pos = sm.prefix().length();
-  auto cite_openkim_potential = fmt::format("OpenKIM potential: "
-    "https://openkim.org/cite/{}#item-citation\n\n",
-    model_name_str.substr(pos, 19));
+  auto cite_id = fmt::format("OpenKIM potential: https://openkim.org/cite/"
+                             "{}#item-citation\n\n",kim_id);
 
   for (int i = 0; i < extent; ++i) {
     char const *fileName;
@@ -539,10 +536,10 @@ void KimInit::write_log_cite(char *model_name)
     if (err) continue;
 
     if (utils::strmatch(fileName, "^kimcite") && availableAsString)
-      cite_openkim_potential += fileString;
+      cite_id += fileString;
   }
 
-  lmp->citeme->add(cite_openkim_potential.c_str());
+  lmp->citeme->add(cite_id);
 
   KIM_Collections_Destroy(&collections);
 }
--- a/src/KIM/kim_init.cpp
+++ b/src/KIM/kim_init.cpp
@@ -70,6 +70,7 @@
 #include "variable.h"
 
 #include <cstring>
+#include <regex>
 
 extern "C" {
 #include "KIM_SimulatorHeaders.h"
@@ -489,7 +490,15 @@ void KimInit::do_variables(const std::string &from, const std::string &to)
 
 void KimInit::write_log_cite(char *model_name)
 {
-  KIM_Collections * collections;
+  if (!lmp->citeme) return;
+
+  std::string model_name_str(model_name);
+  std::regex re("[MOS]{2}_[0-9]{12}_[0-9]{3}");
+  std::smatch sm;
+
+  if (!std::regex_search(model_name_str, sm, re)) return;
+
+  KIM_Collections *collections;
   int err = KIM_Collections_Create(&collections);
   if (err) return;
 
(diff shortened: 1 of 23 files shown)


src/KIM/kim_init.cpp
```diff
@@ -514,19 +523,25 @@ void KimInit::write_log_cite(char *model_name)
...
-        &availableAsString, &fileString);
+      collections, i, &fileName, nullptr, nullptr,
+      &availableAsString, &fileString);
     if (err) continue;
 
-    if (0 == strncmp("kimcite", fileName, 7)) {
-      if ((lmp->citeme) && (availableAsString)) lmp->citeme->add(fileString);
-    }
+    if ((strncmp("kimcite", fileName, 7) == 0) and availableAsString)
+      cite_openkim_potential += fileString;
```
yafshar, line 541: @akohlmey is the concatenation safe here from your view or should I change it to `cite_openkim_potential += std::string(fileString);` or `cite_openkim_potential += fmt::format("{}", fileString);`

rbberger:
@yafshar as much as I would like to use `std::regex`, note that with GCC  < 4.9.0 it is an empty shell. It compiles (!!!) but does random things since it wasn't actually implemented until 4.9.0 (see [this](https://stackoverflow.com/questions/12530406/is-gcc-4-8-or-earlier-buggy-about-regular-expressions) for more details)

yafshar:
> @yafshar as much as I would like to use `std::regex`, note that with GCC < 4.9.0 it is an empty shell. It compiles (!!!) but does random things since it wasn't actually implemented until 4.9.0 (see [this](https://stackoverflow.com/questions/12530406/is-gcc-4-8-or-earlier-buggy-about-regular-expressions) for more details)

@rbberger I agree, that is why I asked about the LAMMPS interface in `utils regex-module`. The use of regex in this PR can simply be replaced with LAMMPS regex-module if you put it in the interface.

src/KIM/kim_init.cpp
akohlmey: yes, that is fine. I am more concerned about the use of `strncmp()`. That is error prone and difficult to debug. Better would be using `utils::strmatch(fileName,"^kimcite")` which will return true/false

akohlmey:
> > @yafshar as much as I would like to use `std::regex`, note that with GCC < 4.9.0 it is an empty shell. It compiles (!!!) but does random things since it wasn't actually implemented until 4.9.0 (see [this](https://stackoverflow.com/questions/12530406/is-gcc-4-8-or-earlier-buggy-about-regular-expressions) for more details)
> 
> @rbberger I agree, that is why I asked about the LAMMPS interface in `utils regex-module`. The use of regex in this PR can simply be replaced with LAMMPS regex-module if you put it in the interface.

For matching strings to regular expressions, you can use `utils::strmatch()`. Please note, that the embedded regex library is very minimal and has limitations. It is not intended to expose the entire library, but only utility functions using it. It is rather a stopgap measure until we can use a proper system library that is guaranteed to be available with the same interface for all supported platforms.

yafshar:
> > > @yafshar as much as I would like to use `std::regex`, note that with GCC < 4.9.0 it is an empty shell. It compiles (!!!) but does random things since it wasn't actually implemented until 4.9.0 (see [this](https://stackoverflow.com/questions/12530406/is-gcc-4-8-or-earlier-buggy-about-regular-expressions) for more details)
> > 
> > 
> > @rbberger I agree, that is why I asked about the LAMMPS interface in `utils regex-module`. The use of regex in this PR can simply be replaced with LAMMPS regex-module if you put it in the interface.
> 
> For matching strings to regular expressions, you can use `utils::strmatch()`. Please note, that the embedded regex library is very minimal and has limitations. It is not intended to expose the entire library, but only utility functions using it. It is rather a stopgap measure until we can use a proper system library that is guaranteed to be available with the same interface for all supported platforms.

@akohlmey  here I need to extract a simple pattern `"[MOS]{2}_[0-9]{12}_[0-9]{3}"` from the model name and thus I need more than checking

akohlmey:
> @akohlmey here I need to extract a simple pattern `"[MOS]{2}_[0-9]{12}_[0-9]{3}"` from the model name and thus I need more than checking. This is very simple and totally possible with utils regex-module

Let me have a closer look and think about an interface.

akohlmey:
@yafshar since we are cleaning up legacy C-style string processing for simpler/safer/readable alternatives, here is another opportunity:
```
 if (!(strcmp(arg[0], "create") == 0) &&
      !(strcmp(arg[0], "destroy") == 0) &&
      !(strcmp(arg[0], "modify") == 0) &&
      !(strcmp(arg[0], "remove") == 0) &&
      !(strcmp(arg[0], "dump") == 0)) {
```
could become
```
const std::string subcmd(arg[0]);
 if ((subcmd != "create") &&  (subcmd != "destroy") && (subcmd != "modify")
     && (subcmd != "remove") && (subcmd != "dump")) {
```
There are similar uses of `strcmp()` elsewhere in the KIM package. Not all of them make sense to refactor, but particularly cases where the same word is compared to strings multiple times would become more readable.

src/KIM/kim_init.cpp
akohlmey: On second thought, there is a problem with this code. Not in how the string is constructed, but the fact that the string is going out of scope. The citeme class currently stores the address of the string pointer in a set to identify strings that have already been submitted. But that will break with this use. I will have to change that part of `CiteMe::add()` to use a hash instead. Then I can also make it use a `const std::string &` as argument and having temporaries added would not be a problem.

yafshar:
> There are similar uses of `strcmp()` elsewhere in the KIM package. Not all of them make sense to refactor, but particularly cases where the same word is compared to strings multiple times would become more readable.

I am cleaning all the possible cases as suggested and will commit them after testing.

src/KIM/kim_init.cpp
akohlmey: PR #2629 should address my concerns about temporaries.

akohlmey:
@yafshar I have updated the PR to use the new `utils::strfind()` function I just added. Mind you, the mini-regex code does not support `{#}`, so I had to adjust the regex string with explicit matches (but it has "\d" for digits). It seems to be working as expected for me and passes the available unit tests. Please check it out and make (final?) adjustments, if needed. Then this could be merged. Also the update to the CiteMe class is included, so it should work reliably with the generated citation strings.

yafshar:
@akohlmey thanks for the update. I check the code and do all my testing.

yafshar:
@akohlmey I did my tests and everything seems good to me. Thanks to your modification the citations are correct and the string is not going out of scope.

sjplimp (approved):

akohlmey (approved):

PR #2629, opened 2021-02-25
Use a 64-bit hash in the CiteMe class to identify unique citations

akohlmey:
**Summary**

The CiteMe class was using a set from the used char pointer addresses to determine whether a citation had been added before and thus not output it a second time. That however will break if we have citation strings that are not compiled in constants, but are generated on the fly and then may be using recycled pointers from reused allocated memory.

This pull request changes the procedure to now compute a 64-bit hash via `std::hash` and store that hash instead.

As a bonus we can use a `const std::string &` argument to `CiteMe::add()` now.

This also adds a simple unit test to confirm the intended behavior.

**Related Issue(s)**

Needed for PR #2625 to work properly.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

This is internal to the CiteMe class and thus no compatibility issues expected.

**Post Submission Checklist**

### DIFF
--- a/src/citeme.cpp
+++ b/src/citeme.cpp
@@ -15,6 +15,8 @@
 #include "comm.h"
 #include "universe.h"
 
+#include <cstdint>
+
 using namespace LAMMPS_NS;
 
 static const char cite_separator[] =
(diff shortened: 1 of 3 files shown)


rbberger:
You could also use `unordered_set<std::string>`, that uses `std::hash` internally.

akohlmey:
> You could also use `unordered_set<std::string>`, that uses `std::hash` internally. (is supported by GCC 4.8.5)

`std::hash` fits the bill. Thanks. Just 2 extra lines added. And no need to keep copies of the strings around. The set is "permanent", the buffers for the logfile are flushed at every "run" or "minimize".

I had been looking for something like it, but somehow didn't see anything appealing. The documentation reads in parts like constructing an atomic bomb(?) so I didn't even make it to the examples. I was just looking for a simple function not a function object. Perhaps I should not have added "simple" to the google search ;-)
So I picked the smallest public domain C code from the examples I found (no point to use something more "expensive" like MD5 or SHA256).

rbberger:
@akohlmey std::hash returns size_t

akohlmey:
> @akohlmey std::hash returns size_t

Well spotted. This was a leftover from using the 32-bit CRC code. Should be fixed now (and the title of the PR ;-) ).

rbberger (approved):

PR #2637, opened 2021-03-01

### LINKED ISSUE #2623: [BUG] RPATH not set correctly for KIM-API with *installed* version of liblammps
**Summary**

When building lammps with the kim-api installed in a non-system location and doing a `make install`, the resulting lmp executable does not have the needed RPATH setting to find the libkim-api library.  However, the build-dir version of the lmp executable is correct.

Behavior is observed on macOS with current master branch.

**Expected Behavior**

executing `lmp` with valid input is successful 

**Actual Behavior**

dyld: Library not loaded: @rpath/liblammps.0.dylib
  Referenced from: /Users/relliott/unison-sync/KIM/git/lammps/build/test-install/./bin/lmp
  Reason: image not found
Abort trap: 6

**Steps to Reproduce**

* download kim-api and lammps and put in a directory:
```
$ mkdir try-it
$ tar Jxvf kim-api-2.2.1.txz
$ cd kim-api-2.2.1
$ mkdir build
$ cd build
$ CC=gcc CXX=g++ FC=gfortran cmake -DCMAKE_INSTALL_PREFIX=${PWD}/test-install -DCMAKE_BUILD_TYPE=Debug ..
$ make -j4 install
$ cd ../../
$ source ./kim-api-2.2.1/build/test-install/bin/kim-api-activate
$ git clone lammps ....
$ cd lammps
$ mkdir build
$ cd build
$ cmake -DCMAKE_INSTALL_PREFIX=${PWD}/test-install -DPKG_KIM=ON ../cmake
$ make -j4
$ otool -L lmp
$ otool -l lmp | grep LC_RPATH -A2  % has setting
$ make install
$ otool -L test-install/bin/lmp
$ otool -l test-install/bin/lmp | grep LC_RPATH -A2  % does not have setting
```

akohlmey:
**Summary**

**Related Issue(s)**

Closes #2639 
Closes #2623 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

- have a better error message when an invalid string was handed to `utils::bounds()`
- LAMMPS python module: explicitly check for existence of the shared library file when using the full path instead of relying on an exception that may not always be thrown as expected. Fixes failing unittests on some platforms
- Add `LAMMPS_INSTALL_RPATH` option to CMake configuration that would select whether to embed runtime paths into installed binaries
- initialize `next_reneighbor` in `setup_pre_exchange()` for fix deposit and fix pour so they are not confused by running a minimizations and other operations that advance the time step counter and won't be run at all.
- nfirst in fix pour and fix deposit should be a bigint since it stores a timestep number
- avoid instabilities with timer based load balancing due to limited timer resolution or precision by adding a minimum time of 0.1 seconds to the computed time.

**Post Submission Checklist**


ellio167 (commented):
An additional note: The traditional make system does not provide this sort of control.  Since it uses pkg-config for the kim-api (which include flags to add a RPATH value as part of the link flags), it will add the RPATH value to the lmp executable.  One of my aims with the original PR was to make the lammps cmake build consistent with the traditional build.  Is this inconsistency something you are worried about?

I think the approach you suggest here is OK.

akohlmey:
> An additional note: The traditional make system does not provide this sort of control. Since it uses pkg-config for the kim-api (which include flags to add a RPATH value as part of the link flags), it will add the RPATH value to the lmp executable. One of my aims with the original PR was to make the lammps cmake build consistent with the traditional build. Is this inconsistency something you are worried about?

No. The traditional make procedure is at the mercy of the individual package maintainers and how people edit the corresponding files or set the required symbolic links. Unlike with CMake there is no common interface and integrated process that makes managing some form of internal consistency possible at all. CMake is particularly attractive to maintainers of pre-compiled binary packages and people at HPC centers, so to have a behavior that is consistent in itself matters mostly there. That is why I favor the default behavior.

> I think the approach you suggest here is OK.

Thanks. I am not exactly eager to have more compilation settings, but I didn't see a viable alternative to serve both conflicting but also justified interests.

ellio167:
Sounds good.  Thanks for working/thinking on this!

junghans:
> Sounds good. Thanks for working/thinking on this!

I agree! This will be useful.

It is also virtually identical with what we use in VOTCA: 
https://github.com/votca/votca/blob/master/CMakeLists.txt#L46-L51

sjplimp (approved):

PR #2639, opened 2021-03-02
Add "INSTALL_RPATH_USE_LINK_PATH True" for lmp and lammps targets

### LINKED ISSUE #2623: [BUG] RPATH not set correctly for KIM-API with *installed* version of liblammps

ellio167:
Also add find_packate_message() indicating kim-api found

Closes #2623

**Summary**

Fix rpath cmake settings for kim-api.

**Related Issue(s)**

closes #2623 

**Author(s)**

Ryan S. Elliott (UMN)

**Licensing**

**Backward Compatibility**

None

**Post Submission Checklist**

### DIFF
--- a/cmake/Modules/Packages/KIM.cmake
+++ b/cmake/Modules/Packages/KIM.cmake
@@ -71,7 +71,11 @@ if(DOWNLOAD_KIM)
 else()
   if(KIM-API_FOUND AND KIM-API_VERSION VERSION_GREATER_EQUAL 2.2.0)
     # For kim-api >= 2.2.0
+    set_target_properties(lammps PROPERTIES INSTALL_RPATH_USE_LINK_PATH True)
+    set_target_properties(lmp    PROPERTIES INSTALL_RPATH_USE_LINK_PATH True)
     find_package(KIM-API 2.2.0 CONFIG REQUIRED)
+    get_target_property(_kim_loc KIM-API::kim-api IMPORT_LOCATION)        # Determine found message
+    find_package_message(KIM-API "Found KIM-API: ${_kim_loc}" "KIM-API")  # since package does not for <=2.2.1; can be removed once 2.2.2 or above is available
     target_link_libraries(lammps PRIVATE KIM-API::kim-api)
   else()
     # For kim-api 2.1.3 (consistent with previous version of this file)


junghans (approved):
This feature might be useful for other packages that link other external shared libraries as well.

cmake/Modules/Packages/KIM.cmake
```diff
@@ -71,7 +71,11 @@ if(DOWNLOAD_KIM)
 else()
   if(KIM-API_FOUND AND KIM-API_VERSION VERSION_GREATER_EQUAL 2.2.0)
     # For kim-api >= 2.2.0
+    set_target_properties(lammps PROPERTIES INSTALL_RPATH_USE_LINK_PATH True)
+    set_target_properties(lmp    PROPERTIES INSTALL_RPATH_USE_LINK_PATH True)
     find_package(KIM-API 2.2.0 CONFIG REQUIRED)
+    get_target_property(_kim_loc KIM-API::kim-api IMPORT_LOCATION)        # Determine found message
```
junghans, line 77: Isn't that stored in `KIM-API_DIR` as well?

ellio167: `KIM-API_DIR` will give the path to where the cmake config file is found, but IMPORT_LOCATION is the full path name of the actual library.  For example, I have a setup where `KIM-API_DIR` has the value `/Users/relliott/unison-sync/KIM/git/kim-api/build/test-install/share/cmake/kim-api` and `IMPORT_LOCATION` has the value `/Users/relliott/unison-sync/KIM/git/kim-api/build/test-install/lib/libkim-api.2.2.2.dylib`

I think using the import location is more consistent with the behavior of `find_package()` and `pkg_check_modules()`

ellio167:
> This feature might be useful for other packages that link other external shared libraries as well.

Agreed.  Seems like it wouldn't hurt to set `CMAKE_INSTALL_RPATH_USE_LINK_PATH` to `True` early in the lammps `CMakeLists.txt`.

akohlmey:
> > This feature might be useful for other packages that link other external shared libraries as well.
> 
> Agreed. Seems like it wouldn't hurt to set `CMAKE_INSTALL_RPATH_USE_LINK_PATH` to `True` early in the lammps `CMakeLists.txt`.

For the sake of consistency, this should be either set globally or not at all. Having this kind of setting enabled only when a specific package is included is creating inconsistent behavior.

However, there are multiple pitfalls and unexpected behavior that can be caused by preserving the rpath setting for **external** libraries. The biggest concern is that it is not as obvious as setting LD_LIBRARY_PATH or DYLD_LIBRARY_PATH and requires technical expertise to detect. Also what happens, if the specific installation folder goes away and is replaced with a new version of the same library in a different folder, that is ABI backward compatible? Suddenly the executable will stop working and one would have to set LD_LIBRARY_PATH to find the updated one. What happens if the build is referring to an automounted folder where the current path would be /auto/some/path, but the canonical path should be /some/path since there are symbolic links directing there even if the /auto path would be different on different machines (can happen on HPC clusters).

My personal preference is - as indicated in the issue - the "public" libraries should be found with LD_LIBRARY_PATH or via /dev/ld.so.conf.d or similar means, but not rpath. rpath instead is a good choice for finding "internal" libraries, i.e. the ones that the public libraries are linked to and that are specific for that version of that library (and not trying to be backward compatible).

akohlmey:
@ellio167 can you please let me know why setting LD_LIBRARY_PATH (or equivalent) from kim-api-activate is not an option?

ellio167:
> @ellio167 can you please let me know why setting LD_LIBRARY_PATH (or equivalent) from kim-api-activate is not an option?

My preference is to not use LD_LIBRARY_PATH if reasonably possible.  (I prefer to avoid--as much as possible--program dependence on the user's environment.)  If you are doing things like using auto mount or moving directories, then I suppose you will have to use LD_LIBRARY_PATH.   I just prefer to not force all users to use environment variables.

If you prefer to not use the INSTALL_RPATH_USE_LINK_PATH setting, we could instead explicitly add the kim-api library path to the lmp and lammps targets' INSTALL_RPATH property.  That way, this would not have any impact on other libraries that lammps links to.

akohlmey:
> > @ellio167 can you please let me know why setting LD_LIBRARY_PATH (or equivalent) from kim-api-activate is not an option?
> 
> My preference is to not use LD_LIBRARY_PATH if reasonably possible. (I prefer to avoid--as much as possible--program dependence on the user's environment.) If you are doing things like using auto mount or moving directories, then I suppose you will have to use LD_LIBRARY_PATH. I just prefer to not force all users to use environment variables.

I need to have a closer look and experiment a bit with different options on a test project to have a fully formed opinion.
However, what you do goes against typical conventions and thus may confuse particularly non-technical people. Not sure how much you get exposed to it, but in my other job this is quite common hence my concerns.

To me it seems inconsistent to not have LD_LIBRARY_PATH as part of kim-api-activate, as you are setting other environment variables there as well. I would expect that if I had compiled two variants/versions of the KIM API in two different folders, that this would switch between them, but with setting rpath, activating the second package would still use the library from the first. But then again, CMake itself is acting against that, as it records full paths and to reset them one would have to either manually reset them or delete them from the cache file. So depending on the perspective, a certain degree of inconsistency seems unavoidable.

If I would want to avoid setting LD_LIBRARY_PATH, I would not build the library as a shared library, but rather as static library but with relocatable objects. But that is a different kettle of fish and has issues at link time (you need to know all required dependent libraries and know how to link to them). There is a discussion about this in the manual about using LAMMPS as a library, where the conclusion is to build the LAMMPS library as a shared library and link with that. In that case using LD_LIBRARY_PATH is the price to pay for the convenience (where the dependencies of LAMMPS can be quite messy with many packages installed).

How would you feel if this feature was a (global) cmake option that is off by default?

> If you prefer to not use the INSTALL_RPATH_USE_LINK_PATH setting, we could instead explicitly add the kim-api library path to the lmp and lammps targets' INSTALL_RPATH property. That way, this would not have any impact on other libraries that lammps links to.

I prefer not to have different behavior for different packages or libraries.

ellio167:
The `kim-api-activate` script is only for build-time settings; except it also add the bin dir to the path so that `kim-api-collections-management` is available, but otherwise no activation is needed to use a program already linked to the kim-api.  If the kim-api is installed in a standard system location no rpath is set.  However, if the kim-api is installed into a non-system directory, then the libkim-api soname is set to `@rpath/libkim-api...` and it is expected that the linking library/executable will have an rpath entry pointing to the correct location.

Maybe `activate` is a poor name choice.

I don't want to require users to have to execute `kim-api-activate` in every shell (or in their .bashrc) before their lammps executable will work.

Alternatively, I guess the linking application should inform its users how to set LD_LIBRARY_PATH appropriately so that the library is found.  I guess that is up to you.  If you don't want to use an rpath setting as expected by the kim-api, then we could print out a message instructing the users what path to add to their LD_LIBRARY_PATH to make things work.  (This seems inconvenient to me.)

akohlmey:
@ellio167 I have read up on the CMake docs and made some experiments and looked into what would be a meaningful compromise. Please have a look at the following commit in PR #2637 which makes this a configurable (global!) option, but then also includes the library installation folder into the list of embedded runtime paths.

https://github.com/lammps/lammps/pull/2637/commits/26086e04a5ee7563b89062bbb9079a871254e641

Please let me know what you think.

Issue #2621, opened 2021-02-23
[BUG] Using pair style hybrid with GPU package and neighbor lists on GPU


akohlmey:
**Summary**

There appears to be a problem using neighbor lists on the GPU with pair style hybrid inputs. The following input does not work reliably with just "-sf gpu" added, but using "-pk gpu 0 neigh no" is avoiding the crashes or corrupted data.
This is supposed to work since PR #1430. Below is an abbreviated version of an input posted on lammps-users.

**LAMMPS Version and Platform**

Current LAMMPS development head revision ce4dc4e2c6978af3a5684139e480e4b631b1a5dd

**Expected Behavior**

Input will run with "-sf gpu" using 1 or multiple MPI processes without having to disable neighbor lists on the GPU.

**Actual Behavior**

LAMMPS reports "nan" or incorrect results or crashes with different errors or warnings. (CUDA has device failures, OpenCL asks to boost neigh_one when run on 1 CPU but ran work with setting binsize, but fails with more MPI ranks).

**Steps to Reproduce**

Use the following input:

```
units        metal
dimension    3
boundary    p p p
atom_style    atomic
variable Ar_mass    equal 39.95                        # [aem]
variable Au_mass    equal 196.97                        # [aem]
variable Ar_eps        equal 0.0103                        # epsilon Ar-Ar, [eV]
variable Ar_sigma    equal 3.4                        # sigma Ar-Ar, [A]
variable Ar_crit    equal 3.5*${Ar_sigma}                    # cut-off distance, [A]
variable rho        equal 1.41                        # density [g/cm^3]

# simulated box
region     simbox     block     0 50     0 50     0 50     side in units box
create_box    2     simbox

########################
# Version #1  w/o hybrid
########################
#pair_style    lj/cut ${Ar_crit}
#pair_coeff    * * ${Ar_eps} ${Ar_sigma}  ${Ar_crit}

############################################
# Version #2 fails with neighbor list on GPU
############################################
pair_style     hybrid lj/cut ${Ar_crit} eam
pair_coeff    1 1 eam Au_u3.eam
pair_coeff    1 2 lj/cut ${Ar_eps} ${Ar_sigma}      # Au-Ar
pair_coeff    2 2 lj/cut ${Ar_eps} ${Ar_sigma}        # Ar-Ar

mass    1    ${Au_mass}
mass    2    ${Ar_mass}

variable Coef        equal 66.34     # to gr/cm^3
[...]

PR #2640, opened 2021-03-03
Fixing bugs on eam*/gpu for pair hybrid with neigh yes, where the gpu…

ndtrung81:
… pair style eam is used for only a subset of the pair types.

**Summary**

For each atom, the eam/gpu force kernels need to accumulate forces/energies for atoms of types that are not skipped (i.e. those with cutsq[itype][jtype] > 0). The changes in the PR passes my tests with the original input script in #2621, and the unitest runs:

unittest/force-styles/test_pair_style ../unittest/force-styles/tests/atomic-pair-hybrid-eam.yaml --gtest_filter=PairStyle.gpu

**Related Issue(s)**

Issue #2621 

**Author(s)**

Trung Nguyen (Northwestern)

**Licensing**

**Backward Compatibility**

Should be maintained

**Implementation Notes**

CUDA and OpenCL builds give different outcomes.

**Post Submission Checklist**


rbberger:
@ndtrung81 Something is also up with EAM which shows with `GPU_PREC=double`. I see memory corruption in the `k_energy_fast` kernel. Something as simple as an added print statement can make the error go away.
```
[14790.491862] NVRM: Xid (PCI:0000:06:00): 13, pid=0, Graphics SM Warp Exception on (GPC 0, TPC 0, SM 0): Out Of Range Address
[14790.491870] NVRM: Xid (PCI:0000:06:00): 13, pid=0, Graphics SM Global Exception on (GPC 0, TPC 0, SM 0): Multiple Warp Errors
[14790.491874] NVRM: Xid (PCI:0000:06:00): 13, pid=0, Graphics Exception: ESR 0x504730=0xc05000e 0x504734=0x24 0x504728=0x4c1eb72 0x50472c=0x174
[14790.492303] NVRM: Xid (PCI:0000:06:00): 43, pid=70041, Ch 00000030
```
I've printed out the values of `type2rhor_z2r_in` and I see that random locations are filled with garbage. I've also tried adding extra init code to `EAMT::init` (`lal_eam.cpp`). That removes some of the uninitialized values, but I see that some that are initialized with 0 on the host are changed to garbage on the GPU.

ndtrung81:
@rbberger changing the call stack making the code run or crash that way is puzzling and needs to be addressed. The array in question (type2rhor_z2r) is of type int2, not related to the precision mode. It seems to me the issue is side effects of memory access errors (out-of-bound) to other arrays. Can you try the last commit to see if there's any difference?

Also, it is likely that this PR may become irrelevant, as I found that neigh yes for pair hybrid currently does not work with OpenCL builds where atom sorting is not available.

rbberger:
@ndtrung81 I tried your latest commit, but I still get the `Out of Range Address` Warp exceptions. FYI, this is what I do:

```bash
mkdir build
cd build
cmake -C ../cmake/presets/minimal.cmake \
            -D PKG_GPU=on \
            -D GPU_API=cuda \
            -D GPU_PREC=double \
            -D ENABLE_TESTING=on \
            -D LAMMPS_EXCEPTIONS=on ../cmake
make -j 12
unittest/force-styles/test_pair_style ../unittest/force-styles/tests/atomic-pair-eam.yaml --gtest_filter=PairStyle.gpu
```

You'll have to apply this diff right now to see the CUDA error:

```diff
diff --git a/unittest/force-styles/test_pair_style.cpp b/unittest/force-styles/test_pair_style.cpp
index 057f2b5352..23ee9c816d 100644
--- a/unittest/force-styles/test_pair_style.cpp
+++ b/unittest/force-styles/test_pair_style.cpp
@@ -792,11 +792,11 @@ TEST(PairStyle, gpu)
         argc    = sizeof(args_noneigh) / sizeof(char *);
     }
 
-    ::testing::internal::CaptureStdout();
+    //::testing::internal::CaptureStdout();
     LAMMPS *lmp = init_lammps(argc, argv, test_config, false);
 
-    std::string output = ::testing::internal::GetCapturedStdout();
-    if (verbose) std::cout << output;
+    //std::string output = ::testing::internal::GetCapturedStdout();
+    //if (verbose) std::cout << output;
 
     if (!lmp) {
         std::cerr << "One or more prerequisite styles with /gpu suffix\n"
@@ -807,8 +807,8 @@ TEST(PairStyle, gpu)
         GTEST_SKIP();
     }
 
-    EXPECT_THAT(output, StartsWith("LAMMPS ("));
-    EXPECT_THAT(output, HasSubstr("Loop time"));
+    //EXPECT_THAT(output, StartsWith("LAMMPS ("));
+    //EXPECT_THAT(output, HasSubstr("Loop time"));
 
     // abort if running in parallel and not all atoms are local
     const int nlocal = lmp->atom->nlocal;
```

```
Cuda driver error 1 in call at file '/home/rberger/lammps/lib/gpu/geryon/nvd_kernel.h' in line 382.
```

ndtrung81:
@rbberger thanks for the info, and I can reproduce the crash. The stdout shows that the potential eam files are missing, so I copied the potential files Cu_u3.eam and Al_jnp.eam into the build folder, and the test now passes. Can you try it again? maybe the path to the potential files needs to be updated?

rbberger:
@ndtrung81 no, I just forgot that detail in the instructions. The crash happens with LAMMPS_POTENTIALS set to the correct directory.

ndtrung81:
@rbberger when I set the environment variable LAMMPS_POTENTIALS to the correct path I couldn't reproduce the crash on my machine (CUDA 11.2, driver 460.67, Fedora 33, g++ 10.3.1). For debugging purposes, I attach here a simple input script that is based on in.metal in unittest that runs correctly. Can you check if that the crash persists with this one (with Cu_u3.eam and Al_jnp.eam in the same folder)? Even better if you can share an input script that produces the crash on your side. Thanks!

[in.metal.txt](https://github.com/lammps/lammps/files/6438895/in.metal.txt)

rbberger:
@ndtrung81 Ok, looks like the latest master merge 7c3a7cb didn't get all the updates. Merging the current master fixes it for me on both my test systems. And yes, it seems efd6f7b actually fixes it. If I revert that one, it happens again.

akohlmey:
> @ndtrung81 Ok, looks like the latest master merge [7c3a7cb](https://github.com/lammps/lammps/commit/7c3a7cb2e4ff389877a685f5fe4c1c38b20ee5a6) didn't get all the updates. Merging the current master fixes it for me on both my test systems. And yes, it seems [efd6f7b](https://github.com/lammps/lammps/commit/efd6f7b2d06abd6b97312e18c5b6f3d99d489ff4) actually fixes it. If I revert that one, it happens again.

FYI, I am cherry-picking that one commit into PR #2756 now so that we can recover unit testing for GPUs on master and possibly also enable for pull requests soon.

ndtrung81:
@akohlmey @rbberger I think the PR is ready for review and more thorough regression tests at this point.

akohlmey:
@ndtrung81 ran this with OpenCL on an Intel GPU with double precision and hybrid styles run ok with neighbor lists on the CPU but neighbor lists on the GPU cause significant differences in forces and stress. In short, it doesn't do any harm, but it does not fully address the issue described in the description.

ndtrung81:
@akohlmey I confirm that something is still wrong with with neigh yes for eam/gpu as a sub-style in pair hybrid, as shown for the unittest with atomic-pair-hybrid-eam. I convert it back to draft for now.

akohlmey:
@ndtrung81 can we merge this and then do the remaining changes in a new PR? even though not all issues are resolved, it is working correctly and has improvements that we should include.

ndtrung81:
@akohlmey yes, please go ahead with review and merge this PR. Let's leave other changes for a separate PR.

ndtrung81:
FWIW, I have been trying to see why the current unittest fails with pair hybrid and pair hybrid overlay. What I found so far is that gpu pair styles, as sub-styles of pair hybrid, do not distinguish pair types (i,j) that are not explicitly set with those that are already set by other sub-styles. Consequently, the gpu substyle calls init_one() and get cutoff value as if they are mixed values, instead of setting them to be zero. There may be more than one  ways to make this work correctly.

Also, pair hybrid support substyle none for pair coeff, which effectively exclude the specified type pairs. These are not supported by the current neighbor list build on the GPU. So if we can find a way make pair_coeff none works with gpu substyles, we can gradually allow neigh exclude to work with GPU neighbor builds.

So at this point, GPU neighbor list builds would work with pair hybrid if the type pair coeffs of all the sub-styles are explicitly specified (cutoff = 0 for those that are to be excluded for a sub-style).

sjplimp (approved):

PR #2645, opened 2021-03-08
relax "fix deform erate" error if run may be halted

jrgissing:
**Summary**

relax "fix deform erate" error to a warning if there is also a "fix halt", which may stop the run before problematic conditions occur

**Related Issue(s)**

none

**Author(s)**

JG

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/src/fix_deform.cpp
+++ b/src/fix_deform.cpp
@@ -467,7 +467,10 @@ void FixDeform::init()
       set[i].hi_stop = set[i].hi_start +
         0.5*delt*set[i].rate * (set[i].hi_start-set[i].lo_start);
       if (set[i].hi_stop <= set[i].lo_stop)
-        error->all(FLERR,"Final box dimension due to fix deform is < 0.0");
+        if (modify->find_fix_by_style("^halt") == -1)
+          error->all(FLERR,"Final box dimension due to fix deform is < 0.0");
+        else
+          if (comm->me == 0) error->warning(FLERR,"Final box dimension due to fix deform is < 0.0");
     } else if (set[i].style == TRATE) {
       set[i].lo_stop = 0.5*(set[i].lo_start+set[i].hi_start) -
         0.5*((set[i].hi_start-set[i].lo_start) * exp(set[i].rate*delt));
--- a/src/fix_deform.cpp
+++ b/src/fix_deform.cpp
@@ -928,6 +928,10 @@ void FixDeform::end_of_step()
     if (set[5].style) domain->xy = set[5].tilt_target;
   }
 
+  for (i = 0; i < 3; i++)
+    if (domain->boxhi[i] <= domain->boxlo[i])
+      error->all(FLERR,"Final box dimension due to fix deform is < 0.0");
+
   domain->set_global_box();
   domain->set_local_box();
 


akohlmey:
@jrgissing this is not a good change for public distribution, since there is no guarantee that fix halt will check for the invalid condition and we cannot have the fix apply invalid conditions or LAMMPS crash in unspecified ways.
We could perhaps have fix deform output in the error message the projected step number when the invalid condition would happen, so that it would be easier to adapt the input.

jrgissing:
that sounds fine to me. And/or, could we just stop fix deform before it tries to make a negative box length?

akohlmey:
> that sounds fine to me. And/or, could we just stop fix deform before it tries to make a negative box length?

Couldn't that be implemented without changes to the code by using variables?

jrgissing:
it seems reasonable to have lammps check this as well. if so, please find that updated. this will now also catch a similar issue which can arise with the 'variable' keyword of fix deform

akohlmey:
> it seems reasonable to have lammps check this as well. if so, please find that updated. this will now also catch a similar issue which can arise with the 'variable' keyword of fix deform

Sorry, I must be missing something here. When LAMMPS can compute when there are too many timesteps with a given deformation rate, so can a user beforehand and just reduce the number of steps.
What is the use scenario where this cannot be done and thus we would need to let fix deform continue to run for a projected number of steps that we know up front it cannot reach and have a physically meaningful trajectory at the same time?

And - as I also mentioned - if you want the more complex behavior, e.g. a deformation rate that changes over time (e.g. stays fixed until a given timestep is reached or a given box dimension), you can realize this though using the variable keyword and equal style variables. 

There are *so* many ways that people can enter bad parameters or settings and we don't work around them either.

As mentioned before, the check for fix halt is something that should not be put into a general release in this too generic form. It may work well for your specific workflows, but we cannot assume that that applies to everybody. Nothing stops you from putting this kind of change into your personal version of LAMMPS for your personal convenience. But in that case, I would not put a check for an external fix halt into the code, but rather have it either issue the fix halt command itself (that is the only way to be certain it checks for suitable conditions) or just also add a new keyword and then adapt the code at the end of `FixHalt::end_of_step()` into your hacked version of fix deform and issue the timeout status directly to stop the current run.

jrgissing:
> Sorry, I must be missing something here. When LAMMPS can compute when there are too many timesteps with a given deformation rate, so can a user beforehand and just reduce the number of steps.
> What is the use scenario where this cannot be done and thus we would need to let fix deform continue to run for a projected number of steps that we know up front it cannot reach and have a physically meaningful trajectory at the same time?

my use case was quickly using 'fix halt' to stop the current run, without worrying about preemptive errors far from any realistic deformation

so, are you okay with just converting the preemptive error to a warning, and relying on the on-the-fly error? if not, we can close this

akohlmey:
> > Sorry, I must be missing something here. When LAMMPS can compute when there are too many timesteps with a given deformation rate, so can a user beforehand and just reduce the number of steps.
> > What is the use scenario where this cannot be done and thus we would need to let fix deform continue to run for a projected number of steps that we know up front it cannot reach and have a physically meaningful trajectory at the same time?
> 
> my use case was quickly using 'fix halt' to stop the current run, without worrying about preemptive errors far from any realistic deformation

Sorry, but that does not explain why you cannot just reduce the number of steps right away. If need be you can compute those on-the-fly.
 
> so, are you okay with just converting the preemptive error to a warning, and relying on the on-the-fly error? if not, we can close this

I don't like to change established behavior without a use case that is impossible or very difficult to realize without that change. So far the only motivation I see is convenience and the limitation imposed by fix deform can be rather easily worked around without changes to the code itself.

jrgissing:
this established behavior predates 'fix halt' and perhaps should be updated (for convenience)

akohlmey:
> this established behavior predates 'fix halt' and perhaps should be updated (for convenience)

My objection has nothing to do with fix halt.

You have now multiple times avoided to answer my question why you cannot reduce the number of steps from the beginning or use an equal style variable to compute the number of maximally possible steps from the deformation rate or use a pair of equal style variables to define the deformation rate and box volume in such a way that fix deform will not stop or crash.
All of this - as far as I understand - will address the issue of fix deform stopping right away without having to change the code and without imposing a significant inconvenience and you may still keep using fix halt the way you currently do.

akohlmey:
there have been no comments from other LAMMPS developers or updates that omit the objectionable check for a fix halt instance for 3 weeks now so I assume this is abandoned.

jrgissing:
note that this also contains an on-the-fly check for invalid box dimensions, to gracefully catch this error with other keywords in the same way as the 'erate' keyword

PR #2650, opened 2021-03-09
New PLUGIN package with a LAMMPS plugin loader command

akohlmey:
**Summary**

This pull request provides a new package with a "plugin" command that allows LAMMPS to load features (i.e. pair, fix, compute styles) from separate .so (or .dylib) files at runtime. Those .so files can be built using only a few LAMMPS headers and without recompling and relinking LAMMPS itself.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

This is new functionality.

**Implementation Notes**

Currently only supports pair, bond, angle, dihedral, improper, compute, fix, region, and command styles are supported. The list may be expanded on demand.

The folder examples/plugins contains examples for a "morse2" and "morse2/omp" pair style (copies of morse, but with different class names) as well as for an "nve2" fix style (copy of fix nve but with different class name) or a "hello" command style.

Current capabilities:
- will ignore plugin commands when LAMMPS was not compiled as a shared library
- can load/unload plugins for pair styles
- can load/unload plugins for fix styles
- can load/unload plugins for commands
- can list the loaded plugins
- can register multiple style plugins from a single shared object
- may only unload/disable styles that have been previously loaded as plugins
- may not load the same plugin a second time without unloading it first
- introspection functions exported to Python module

**NOTE:** some small adjustments for more consistent APIs to manage fixes, computes, and regions were added.

**Post Submission Checklist**


src/lammpsplugin.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+#ifdef _WIN32
+#include <windows.h>
+#else
+#include <dlfcn.h>
+#endif
+
+namespace LAMMPS_NS
+{
+  static std::vector<lammpsplugin_entry_t> pluginlist;
```
rbberger, line 31: @akohlmey shouldn't this be an instance variable of LAMMPS or of a child object? E.g. `LAMMPS::Plugins`? These methods also look like they should become part of the library interface and follow the same naming conventions: `lammps_plugin_load`, `lammps_plugin_register`.

akohlmey: nope. I have thought about that.  Whether those are loaded or not is a global property of the executable/shared library. I was actually thinking the other way around, that the maps for the individual styles should also be global lists and not part of the objects that create them. Adding the plugin system is making things a bit more complicated and adds new problems like "should we allow to "unload" statically loaded classes?" or "how do we avoid leaking dso handles?", but it can also be useful for development since you can load new functionality without having to rebuild the full executable/library and you can replace existing styles with new implementations without rebuilding.

akohlmey: > Then it needs some additional protections.

Certainly. This is a bit of a "you know it when you see it" situation, where you first need to figure out a way to do something, then study its utility and usability and then refine or redo.  Which is why I make this public now while I am still exploring. This must be the 3rd time I am starting to looking into this seriously, and now is the first time I am getting somewhere that looks like it can work the way I expect it to work. One thing I am trying to avoid now is to over-engineer the initial version. Previously, I would always start with something too complex; now I am trying to wait.

If you want, we can discuss some more about this and how to manage the "factory maps" person-to-person.

akohlmey:
@sjplimp @rbberger this PR is approaching now a state where I am running out of ideas for what to check for and what needs to be implemented or how things could go wrong. I have also added some docs (for users and programmers). The major missing pieces are now to expand the support for more different styles (compute, bond, angle, dihedral, improper, dump(?), region(?), reader(?), integrate(??), minimize(??)) and to come up with some way of (unit) testing.

One known limitation is currently that there is no dependency tracking. If you have a class derived from a class that is loaded as a plugin, you need to put both into the same shared object file. Otherwise there will be a crash is the base plugin is unloaded and the corresponding DSO handle closed. I don't think that dependency tracking can be reliable unless it is done automatically, but I see now way to do it. For the DSO handle I have implemented reference counting, so the handle will only be closed after the last plugin from a file is deleted.

In the mean time, I will look into checking this on platforms other than Linux. Not sure if it is worth adding Windows support. It should be possible (it works for molfile plugins, but those are self-contained and use a C function-only interface), but then building on Windows is more complex and requires compatible tools, but it could be a way for, say, the OpenKIM folks to have their package rolled into a plugin and loaded after the fact into the existing Windows packages.

akohlmey:
@sjplimp @rbberger @stanmoore1 @athomps this is ready for discussion now. The only thing that I would add from here on is support for more "kinds of styles", but I have no other features on my mind. Talking points are the following items.
- do we want (and support) this kind of feature in LAMMPS?
- should it be unconditionally included or be made into a package and thus a compile time choice?
- it can only function if LAMMPS is compiled in "shared" mode, thus it could (and probably should) be only compiled in if that compilation mode is chosen. should we add a define for that?
- I've tested it successfully on Linux (Fedora, Ubuntu, RHEL/CentOS), MacOS, and FreeBSD. Support for Windows (with cross compiler) turned out to be beyond my skills (although it might be particularly useful there). Any other platform worth checking?
- while it works well for its intended functionality, this can be used to "corrupt" a LAMMPS binary so it can make debugging and supporting people difficult (particularly those with limited levels of experience). Do we want that? Mind you, since people can compile from source, they can do the same by altering the source.

akohlmey:
@sjplimp this PR is not "feature complete" for a first set of plugin styles that make the most sense to support. Additional ones are possible, but I would only want to look into those with a given use case.  This is by choice rather minimal in what is being checked for.  Additional checks and improvements can be made as people use this feature and report problems.

sjplimp (approved):

PR #2655, opened 2021-03-11
Collected small code refactoring changes

akohlmey:
**Summary**

This pull request combines multiple small refactoring changes that adapt the LAMMPS code to avoid certain operations (e.g. calling `strtok()`) and make better use of convenience functions in the utils namespace or the tokenizer classes.
With this pull request merged, there are no more uses of `strtok()` in the core LAMMPS code in `src`. Only in packages and libraries.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

- avoid calling `strtok()` when parsing the "partition" command; simplify the code
- avoid calling `strtok()` when reading data files with `read_data`
- avoid calling `strtok()` in `group` command
- avoid calling `strtok()` when reading *file* and *atomfile* style variables and use ValueTokenizer for the latter
- avoid calling `strtok()` from various dump styles. simplify string processing using `std::string` and `utils::strdup()`
- refactor `utils::expand_args()` to use c++ string tools and fmtlib to recombine the expanded string.
- let ValueTokenizer throw an exception, if the next value is requested but there is no token left.
- add unit tests for out of token exceptions thrown by ValueTokenizer
- add unit test for partition command

**Post Submission Checklist**

### DIFF
--- a/src/group.cpp
+++ b/src/group.cpp
@@ -311,11 +311,11 @@ void Group::assign(int narg, char **arg)
         try {
           ValueTokenizer values(arg[iarg],":");
           start = values.next_tagint();
-          if (utils::strmatch(arg[iarg],"^\\d+$")) {
+          if (utils::strmatch(arg[iarg],"^-?\\d+$")) {
             stop = start;
-          } else if (utils::strmatch(arg[iarg],"^\\d+:\\d+$")) {
+          } else if (utils::strmatch(arg[iarg],"^-?\\d+:-?\\d+$")) {
             stop = values.next_tagint();
-          } else if (utils::strmatch(arg[iarg],"^\\d+:\\d+:\\d+$")) {
+          } else if (utils::strmatch(arg[iarg],"^-?\\d+:-?\\d+:\\d+$")) {
             stop = values.next_tagint();
             delta = values.next_tagint();
           } else throw TokenizerException("Syntax error","");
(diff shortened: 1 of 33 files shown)


src/group.cpp
```diff
@@ -303,20 +304,24 @@ void Group::assign(int narg, char **arg)
...
-          ptr = strtok(nullptr,":");
-          stop = utils::tnumeric(FLERR,ptr,false,lmp);
-          ptr = strtok(nullptr,":");
-          if (ptr) delta = utils::tnumeric(FLERR,ptr,false,lmp);
-        } else {
-          start = stop = utils::tnumeric(FLERR,arg[iarg],false,lmp);
+        try {
+          ValueTokenizer values(arg[iarg],":");
+          start = values.next_tagint();
+          if (utils::strmatch(arg[iarg],"^\\d+$")) {
```
rbberger, line 314: @akohlmey Instead of double-parsing with `strmatch` here you could instead store `values.count()` as `ncount` and then do:
```
if (ncount == 1) {
  ...
} else if (ncount == 2) {
  ...
} else if (ncount == 3) {
  ...
}
```
or use a `switch` statement. The `next_tagint()` functions will already throw a `TokenizerException` if the value isn't an integer.

akohlmey: I had something similar at some point but it was not enforcing the syntax properly. Incorrect input like "1:" or "1::4" or ":2" are happily accepted while they are not allowed. I added the `utils::strmatch()` calls to enforce the syntax more strictly. The counting of tokens is then a welcome side effect as I can omit `if (values.has_next())`.

rbberger (approved):

sjplimp (approved):

PR #2659, opened 2021-03-14
Updating displace.mod

evoyiatzis:
**Summary**

The pressure tensor was not computed as time-averaged quantity but as instantaneous value (this is the case for the elastic script at zero temperature)

**Related Issue(s)**

No related issue

**Licensing**

**Backward Compatibility**


athomps (approved):
Thanks for fixing this.  One clarifying comment.  The original script was correct for the first (negative) deformation, but incorrect for the second (positive) deformation.

PR #2661, opened 2021-03-16
[BUGFIX] Scaling of cdof for fix ave/chunk norm all

ssande7:
**Summary**

When calculating temperature, fix ave/chunk uses (cdof + Natoms*adof) as the degrees of freedom. However, under norm all, Natoms is the number of atoms summed over all samples, while cdof isn't scaled accordingly. The difference is apparent in the attached input script, which uses a single system-wide chunk to compute temperature, setting cdof to -2 to match the temperature compute used by fix nvt. Using norm sample produces the correct temperature measurement, whereas norm all reports a temperature that is slightly too low (most obvious in smaller systems with long averaging periods). This bugfix adds a scaling factor to cdof of Nrepeats in the case that norm all is used. Running the same input script with this bugfix produces identical temperature output with norm all and norm sample, as would be expected.

**Related Issue(s)**

None.

**Author(s)**

Stephen Sanderson - University of Queensland

**Licensing**

**Post Submission Checklist**

Example script with output from old and new versions
[cdof_bug.tar.gz](https://github.com/lammps/lammps/files/6146739/cdof_bug.tar.gz)

### DIFF
--- a/doc/src/fix_ave_chunk.rst
+++ b/doc/src/fix_ave_chunk.rst
@@ -307,7 +307,9 @@ atoms in the chunk.  The averaged output value for the chunk on the
 average over atoms across the entire *Nfreq* timescale.  For the
 *density/number* and *density/mass* values, the volume (bin volume or
 system volume) used in the final normalization will be the volume at
-the final *Nfreq* timestep.
+the final *Nfreq* timestep. For the *temp* values, degrees of freedom and
+kinetic energy are summed separately across the entire *Nfreq* timescale, and
+the output value is calculated by dividing those two sums.
 
 If the *norm* setting is *sample*\ , the chunk value is summed over
 atoms for each sample, as is the count, and an "average sample value"
--- a/src/fix_ave_chunk.cpp
+++ b/src/fix_ave_chunk.cpp
@@ -881,7 +881,7 @@ void FixAveChunk::end_of_step()
       if (count_sum[m] > 0.0)
         for (j = 0; j < nvalues; j++) {
           if (which[j] == ArgInfo::TEMPERATURE) {
-            values_sum[m][j] *= mvv2e / ((cdof + adof*count_sum[m]) * boltz);
+            values_sum[m][j] *= mvv2e/((repeat*cdof + adof*count_sum[m])*boltz);
           } else if (which[j] == ArgInfo::DENSITY_NUMBER) {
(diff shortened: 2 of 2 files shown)


sjplimp:
@ssande7 Good catch on this.  I think this is the case but want to verify.  This change will work if the # of atoms in a particular chunk varies from one sampling interval to the next?

ssande7:
Thanks @sjplimp. I think this change should make things consistent between norm all and norm sample. Norm sample calculates the temperature as:
T = Sum_samples(Sum_atoms(m*v*v) / (k_B * (cdof + N_atoms * adof))) / N_samples
and norm all now calculates it as:
T = Sum_samples(Sum_atoms(m*v*v)) / (k_B * Sum_samples(cdof + N_atoms*adof))

There is the interesting question though of whether a degrees of freedom per chunk is a definable quantity when the number of atoms per chunk isn't consistent between chunks, since two chunks with a different number of atoms but the same cdof would report a slightly different temperature even if the KE per atom was fixed across the system. In the extreme case, a chunk with a very low number of (or no) atoms could end up with <= 0 DoF, and report a nan or negative temperature. This is consistent between norm all and norm sample though, and can be inferred from the documentation of fix ave/chunk. In such cases, I think DoF corrections are more appropriately applied via adof than cdof.

sjplimp:
Thanks for the clarification. I think the only thing this lacks is some added text on the fix ave/chunk doc page. I think the way it is now, the description of nsample all and the formulas for DOF(adof,cdof) and the description of the cdof,adof keywords would make a user think it was computed the old way. (Which makes sense since I probably wrote the code and the doc info to match).  So if we are changing that to your new formula, then the doc page should explain that is how it is done when nsample all is used.

ssande7:
My first interpretation of the documentation was that this was already the intended behaviour, but upon re-reading I see your point. I've updated the norm all doc text to hopefully clarify.
Another thing I noticed though is that there may be some interesting interactions with the 'ave running' option, which I'll look into quickly now.

ssande7:
> Another thing I noticed though is that there may be some interesting interactions with the 'ave running' option, which I'll look into quickly now.

I couldn't see anything obvious in the code that would cause issues, and a quick test with the included script from above (modified to use ave running) gave identical results for norm sample and norm all so I think it's ok, but someone more familiar than I am with that bit of code should probably double check.

sjplimp (approved):

PR #2662, opened 2021-03-16
Addition of a PyLammps script to compute elastic constants using OpenKIM

evoyiatzis:
**Summary**

A serial python script based on PyLammps interface to compute the elastic constants of a system at zero temperature with the force field being defined via OpenKIM 

**Related Issue(s)**

No related issue

**Author(s)**

Evangelos Voyiatzis (Royal DSM) evoyiatzis@gmail.com

**Licensing**

**Backward Compatibility**

No backward compatibility issues

**Post Submission Checklist**


python/examples/pylammps/elastic/elastic.py
```diff
@@ -0,0 +1,309 @@
...
...03, 'Ne': 20.179, 'Na': 22.98977, 'Mg': 24.305, 'Al': 26.98154, 'Si': 28.0855, 'P': 30.97376, 'S': 32.06, 'Cl': 35.453, 'K': 39.0983, 'Ar': 39.948, 'Ca': 40.08, 'Sc': 44.9559, 'Ti': 47.90, 'V': 50.9415, 'Cr': 51.996, 'Mn': 54.9380, 'Fe': 55.847, 'Ni': 58.70, 'Co': 58.9332, 'Cu': 63.546, 'Zn': 65.38, 'Ga': 69.72, 'Ge': 72.59, 'As': 74.9216, 'Se': 78.96, 'Br': 79.904, 'Kr': 83.80, 'Rb': 85.4678, 'Sr': 87.62, 'Y': 88.9059, 'Zr': 91.22, 'Nb': 92.9064, 'Mo': 95.94, 'Tc': (98), 'Ru': 101.07, 'Rh': 102.9055, 'Pd': 106.4, 'Ag': 107.868, 'Cd': 112.41, 'In': 114.82, 'Sn': 118.69, 'Sb': 121.75, 'I': 126.9045, 'Te': 127.60, 'Xe': 131.30, 'Cs': 132.9054, 'Ba': 137.33, 'La': 138.9055, 'Ce': 140.12, 'Pr': 140.9077, 'Nd': 144.24, 'Pm': (145), 'Sm': 150.4, 'Eu': 151.96, 'Gd': 157.25, 'Tb': 158.9254, 'Dy': 162.50, 'Ho': 164.9304, 'Er': 167.26, 'Tm': 168.9342, 'Yb': 173.04, 'Lu': 174.967, 'Hf': 178.49, 'Ta': 180.9479, 'W': 183.85, 'Re': 186.207, 'Os': 190.2, 'Ir': 192.22, 'Pt': 195.09, 'Au': 196.9665, 'Hg': 200.59, 'Tl': 204.37, 'Pb': 207.2, 'Bi': 208.9804, 'Po': (209), 'At': (210), 'Rn': (222), 'Fr': (223), 'Ra': 226.0254, 'Ac': 227.0278, 'Pa': 231.0359, 'Th': 232.0381, 'Np': 237.0482, 'U': 238.029}
```
rbberger, line 189: any particular reason why some values are in parentheses? E.g. `'Pm': (145)`

python/examples/pylammps/elastic/elastic.py
```diff
@@ -0,0 +1,309 @@
...
+
+    # Compute elasic constant from pressure tensor
+    c1pos = lmp.variables["d1"].value
+    c2pos = lmp.variables["d2"].value
+    c3pos = lmp.variables["d3"].value
+    c4pos = lmp.variables["d4"].value
+    c5pos = lmp.variables["d5"].value
+    c6pos = lmp.variables["d6"].value
+
+    # Combine posiive and negative
```
rbberger, line 130:
```suggestion
    # Combine positive and negative
```

python/examples/pylammps/elastic/elastic.py
akohlmey: If I am not mistaken (it has been 30 years since I last dealt with this), that is just a Chemistry/Physics convention. The masses in the periodic table usually are the mass due to the natural mix of isotopes for a given element. Some elements are rather short lived so that there is no "natural mix" (or no natural occurrence at all). Then the convention is to provide the mass of the most long lived isotope in parenthesis.

python/examples/pylammps/elastic/elastic.py
```diff
@@ -0,0 +1,309 @@
...
+    cfac = 1.0e-4
+
+    # Define minimization parameters
+    L.variable("dmax equal 1.0e-2")
+
+    L.boundary("p", "p", "p") # periodic boundary conditions in all three directions
+    L.box("tilt large") # to avoid termination if the final simulation box has a high tilt factor
+
+    # use the OpenKIM model to set the energy interactions
+    L.kim_init(args.kim_model, "metal", "unit_conversion_mode")
```
rbberger, line 182: This fails with `ERROR: This command is no longer available (src/deprecated.cpp:43)`. This should now be
```suggestion
    L.kim("init", args.kim_model, "metal", "unit_conversion_mode")
```

evoyiatzis: thanks for fixing that. it seems i am using an old version of the interface

python/examples/pylammps/elastic/elastic.py
evoyiatzis: Actually I should remove the parentheses. In the unlikely event someone finds a universal FF in OpenKIM the parentheses will lead to an error

giacomofiorin: This is quite interesting... Since we're going back, I do remember that nuclear binding forces will measurably offset the total mass of the nucleus based on how much it is stable. E = mc^2, and although most of it is mass of protons and neutrons, the contribution due to nuclear forces is not negligible. The latter results in a higher/lower total mass for the nucleus than you would expect. Further, protons and neutrons have unequal masses and when increase in mass their optimal mix deviates from the exact 1:1 ratio that you have in lighter elements.

So how come those masses in parenthesis happen to be all exact multiples of the Dalton? Some of those elements are long lived enough for more precise mass measurements, I would think.

athomps:
These masses have no effect on elastic constants.  Maybe they should all be set to a dummy value?

python/examples/pylammps/elastic/elastic.py
rbberger: @evoyiatzis that's just one of many, basically all `kim_xyz` commands are now `kim xyz`, so the rest still needs to be updated.

evoyiatzis:
> These masses have no effect on elastic constants. Maybe they should all be set to a dummy value?

I thought they were needed for damped dynamics methods such as fire & quickmin. If this is not the case, I will remove the dictionary

python/examples/pylammps/elastic/elastic.py
evoyiatzis: @rbberger I will not update my local installation in the next weeks. I can make some changes following the manual but without testing it does not make a lot of sense. should i close the request or leave it hanging for some time?

athomps:
As long as you set all masses to some non-zero value, the default min_style CG will be unaffected. Some min_styles may require a mass in the right physical ballpark e.g. 10.0, but again the precise value should have no effect of converged elastic constant.

athomps:
@evoyiatzis In the original ./examples/ELASTIC script, I used

mass * 1.0e-20

rbberger (approved):

PR #2666, opened 2021-03-17

### LINKED ISSUE #2672: [BUG] fix gcmc doesn't work when USER-INTEL is used in MD simulation
**Summary**
As documented, *fix gcmc* can be used with *fix nvt* to run grand canonical MD simulation. However, if USER-INTEL package is used in such simulation, *fix gcmc* seems stop working.

Both tested on a cluster and my own computer.
cluster: LAMMPS patch_10Mar2021, Red Hat Enterprise Linux Server release 7.3 (Linux version 3.10.0), Intel Xeon E5-2690A V4
my computer: pre-built binary from http://packages.lammps.org/windows.html (29Oct2020, 64bit non-MPI), Win10 20H2, AMD Ryzen 5 2600X

**Expected Behavior**
The atom count should increase or decrease (increase in this case) due to GCMC exchanges, like this in thermo output
Step Atoms
       0        2
     100        3
     200        3
     300        6
     400        9

The behavior is tested correct when using no acceleration packages, or using GPU (CUDA), or KOKKOS (CUDA), or USER-OMP.

**Actual Behavior**
The atom count doesn't change over time (like thermo output below). It can be further comfirmed by looking at dump file.
Step Atoms 
       0        2 
     100        2 
     200        2 
     300        2 
     400        2 

This behavior occurs when using USER-INTEL.

**Steps to Reproduce**
Here is a demo input: 
[demo.zip](https://github.com/lammps/lammps/files/6170101/demo.zip)
run with USER-INTEL: *lammps_executable_path -sf intel -i in.in*
run without USER-INTEL: *lammps_executable_path -i in.in*

### LINKED ISSUE #2673: [BUG] Format error when read_data nocoeff
**Summary**

When running a Lammps simulation with an hybrid pair_style and a read_data put after the pair_style line, Lammps crashes with the error:

`fmt::format_error: string pointer is null`

Last dev version from github : **LAMMPS (10 Mar 2021)**

**Expected Behavior**

Not crashing

**Actual Behavior**

It crashes

**Steps to Reproduce**

Here's a simple example with an input and data file:

[bug_format.zip](https://github.com/lammps/lammps/files/6171357/bug_format.zip)

The exception comes from `neighbor.cpp` code, in the `print_pairwise_info` method.

The line `char *pname = force->pair_match_ptr((Pair *) rq->requestor);` returns `null` and so the format raises an exception.

The `null` comes because the `force->pairstyle` has been set to `zero` in the `read_data.cpp` code, and has not been reput to the old values.
In `read_data.cpp`, when `nocoeff` the pair is saved to `saved_pair` and restore, but `force->pairstyle` remains at `zero`.

With the `03 Mar 2020`, the bug does not arise, but the output contains `(null)`:
` 
 (1) pair (null), perpetual, skip from (3)
      attributes: half, newton on
      pair build: skip
      stencil: none
      bin: none
  (2) pair (null), perpetual, skip from (3)
      attributes: half, newton on
      pair build: skip
      stencil: none
      bin: none
...
`
This part of the code is a little too tricky for me, it's why I didn't provide you a fix.

Thanks !

akohlmey:
**Summary**

This PR collects multiple small changes and fixes that don't warrant a pull request on their own.

**Related Issue(s)**

Fixes #2672 
Fixes #2673 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues

**Implementation Notes**

- avoid a segfault on empty strings in the library interface when converting a list of commands to a block of commands
- simplify code by using `utils::strdup()` to duplicate strings
- fix bug in `utils::expand_args()`
- fix a missing thread library issue when compiling with Intel compilers and using C++11 style threads
- fix for fix gcmc and fix widom that allows them to be used with USER-INTEL
- bugfix for read_data that was not preserving style names and thus could trigger an exception on neighbor list output

**Post Submission Checklist**


src/library.cpp
```diff
@@ -444,12 +444,11 @@ is passed to :cpp:func:`lammps_commands_string` for processing.
...
 void lammps_commands_list(void *handle, int ncmd, const char **cmds)
 {
-  LAMMPS *lmp = (LAMMPS *) handle;
   std::string allcmds;
 
   for (int i = 0; i < ncmd; i++) {
     allcmds.append(cmds[i]);
-    if (allcmds.back() != '\n') allcmds.append(1,'\n');
+    if (allcmds.empty() || (allcmds.back() != '\n')) allcmds.append(1,'\n');
   }
```
akohlmey, line 452: @rbberger somehow this bug went unnoticed by the automated unit tests while it created an exception/segmentation fault on my local machine in the PythonCommands test (you must not use `std::string.back()` on an empty string). Was this just by chance or do some tests fall through the grid of the various configurations?

sjplimp (approved):

PR #2667, opened 2021-03-17
[BUGFIX] Collection of DoF-related fixes for compute temp/profile

ssande7:
**Summary**

There are a few things going on here, some of which may need further discussion. At the same time as I discovered #2661 I also found some inconsistencies/issues in the treatment of degrees of freedom by compute temp/profile, but I thought they should be in a separate PR since they're not directly related to fix ave/chunk.

Firstly, while the documentation states that DoF is calculated as (N\*dim - nbins\*dim), the default behaviour actually calculates DoF as (N\*dim - (nbins+1)\*dim), since extra_dof defaults to 2, rather than 0. Fixing this required that Compute::reset_extra_dof() be made virtual, but there were no locally overridden versions of reset_extra_dof in any other computes that I could find with a fuzzy search, so I don't think making it virtual in compute.h will break anything. Defaulting extra_dof to 0 does have some implications for compute_array() though, which I'll discuss below.

There were a couple of issues with compute_array(). To start with (and perhaps this part should have been in a separate PR), it was overwriting dof and tfactor with the bin-specific values, but in the case of non-dynamic groups those values weren't restored, and so after a call to compute_array, compute_scalar was returning incorrect values that could influence thermostatting. This can be seen, for example, by using `thermo_style custom c_profile c_profile[1][2]`, where c_profile is a temp/profile compute, and is also used by the thermostat.

The other compute_array() issue is possibly more philosophical, and I expect some debate here - the current calculation for DoF within a bin uses dim\*N - extra_dof, neglecting fix_dof, and assuming that extra_dof includes the subtraction required for the streaming velocity adjustment if needed. However, because extra_dof is applied on a per-bin basis here, this is inconsistent with compute_scalar(). The inconsistency is as-documented, but makes it difficult to choose a value for extra_dof when both outputs are required (for example, setting extra_dof to 0 would still include constraints due to streaming velocity adjustment in compute_scalar, but would not include them in compute_array). I think the more appropriate behaviour here (not yet considering some further points that I will make below) would be for a bin's DoF to be calculated as dim\*N - dim - extra_dof, and therefore the streaming velocity constraints would be consistently considered in both compute_scalar and compute_array. However, in addition, by subtracting a non-zero extra_dof on a per-bin basis the degrees of freedom summed across all bins is not consistent with the total degrees of freedom used by compute_scalar(), and therefore compute_array() reports temperatures that are on average slightly lower than that of compute_scalar() when extra_dof is positive (which could be problematic if the same compute is used for thermostatting as well as per-bin temperature output, for example). Hence my proposed change is that extra_dof (and fix_dof) be applied fractionally to each bin based on the fraction of atoms in that bin, (DoF = dim\*N - dim - (extra_dof + fix_dof)\*N/N_tot), so that the DoF summed over all bins is consistent with the total DoF of the group as calculated by compute_scalar(). The only edge case I can think of where DoF wouldn't be consistent under this method is if some bins didn't contain any atoms, but I think that's easily addressed with some minor changes to compute_scalar which I can make and add to this PR if we decide that's the correct approach.

Finally, in the referenced Evans and Morriss paper, they subtract dim\*nbins degrees of freedom from the system as they are subtracting a dim-dimensional streaming velocity. However, LAMMPS allows streaming velocity to be subtracted in only some dimensions, in which case dim\*nbins would not be the correct adjustment. The obvious example of this is if `compute temp/profile 0 0 0 y 10` were used - despite the 10 bins, no streaming velocity is being subtracted, yet dim*10 would be subtracted from DoF instead of just dim (assuming the total center-of-mass velocity is fixed). Another example is the same command, but with streaming velocity only subtracted in one direction - center-of-mass velocity in the dimensions without streaming velocity subtraction is only constant across the whole system, not within a bin, but DoF is being adjusted as if it were separately constant in each bin. Hence, the change I've included here counts the directions in which a streaming velocity is subtracted and multiplies that by nbins, then (by default) subtracts a single additional degree of freedom for each direction in which the streaming velocity is *not* subtracted. Those additional DoFs in non-streaming directions are subtracted via extra_dof for consistency with the extra_dof overriding behaviour of compute temp, and so that they are applied fractionally per-atom in compute_array as discussed above. If the decision is made to not include the fractional behaviour discussed in the previous paragraph, then a discussion is probably needed about how (or whether) system-wide constraints should be factored into the DoF of a given bin.

The included updates to the documentation should reflect all of the implemented changes.

**Related Issue(s)**

None.

**Author(s)**

Stephen Sanderson - University of Queensland

**Licensing**

**Backward Compatibility**

This changes the treatment of extra_dof for compute temp/profile, and implications of that should be considered. While I think the proposed changes make the behaviour more consistent (and should fix the minor bug described in the first paragraph), it does mean that existing scripts which rely on the current behaviour will produce slightly different results that are most obvious in small systems. However, anecdotally, the only times I've needed to set extra_dof with compute temp/profile thus far has been to correct the total DoF when streaming velocity is only  subtracted in one direction, and in that case the inconsistency with compute_array means that either fix ave/chunk or a separate compute temp/profile is needed for correct per-bin temperature output.

**Implementation Notes**

See above.

**Post Submission Checklist**


ssande7:
One further thing to note after thinking about this some more - the current implementation allows the user to define via extra_dof a per-bin DoF adjustment when it is applied in compute_array (albeit at the cost of inconsistent behaviour in compute_scalar). These proposed changes would remove that possibility, so perhaps it would be a good idea to rename nstreaming to something like extra_dof_bin, and expose it to the user in a similar manner to extra_dof. I'm unsure if there's any actual use case for that though, so it may not be necessary.

athomps:
@ssande7 Thanks for this carefully described modification.  It took me quite a while to understand the subtleties, but I think I agree with all of your modifications. The most important change might be one you did not explicitly mention.  In the old code, `ndof` for a bin is reduced by `ndim`, even if `nstreaming < ndim`. It would be quite natural for a user to only apply bias in one direction, in which case the bin temperatures would be too high be a factor of `1+(ndim-1)/ndim/(N-1)`, which is substantial when the number of atoms in a bin happens to be quite small.

I added some extra assertions to the note on the treatment of subtracted degrees of freedom.  This will help future users and developers understand the intended behavior. I think these assertions should be true, but I did not test them.  @ssande7, please convince yourself that these are true in most cases, further modifying the code if necessary.  Once that is done, I will approve the PR.

athomps (changes requested):
See final part of my comment in discussion.

ssande7 (commented):
Thanks @athomps . I agree with those assertions with the added qualifying statement making note of the fact that an empty bin that would count its DoF as negative will instead truncate to 0, in which case those sums will not be equal.

I've also added a warning note about the tricky (although likely rare) case of inhomogeneous blends of rigid molecules, since I realised that by distributing DoF constraints in `fix_dof` equally between bins, we are inherently assuming the system is homogeneous. This could possibly be addressed more accurately by leveraging `compute chunk/atom molecule` to store a "DoF per atom" value that is aware of the constraints per molecule, but doing so would require explicit knowledge of all rigid-body fixes, and so perhaps would be better left to be handled by the user. I can try to put together an example input script for this case if you think it's worthwhile.

athomps (approved):
I approve. I made some further changes to the language in the caveats to the assertions.

ssande7:
Thanks, looks good to me.

PR #2677, opened 2021-03-22
Optimizations to Kokkos SNAP

stanmoore1:
**Summary**

Evan Weinberg (NVIDIA) @weinbe2 pointed out a couple optimizations in the Kokkos SNAP kernel:
- Use existing `zlist` to reduce the cost of `compute_yi` when using quadratic option or computing energy
- Improve memory coalescing in Kokkos snap force computation, which speeds up large numbers of atoms/GPU (technically this fixes a regression from a previous refactor)

**Related Issue(s)**

**Author(s)**

Stan Moore (SNL), pointed out by Evan Weinberg (NVIDIA)

**Licensing**

**Backward Compatibility**

Yes


stanmoore1:
I tested this for tungsten 2J14 with 2000 atoms, no performance regression and numerics are correct. I also tested numerics and performance on a quadratic case, everything looks good.

sjplimp (approved):

PR #2698, opened 2021-04-06
New option for granular pairstyles and updates to granular walls

jtclemm:
**Summary**

This pull request adds a new keyword, "limit_damping", to all the granular pairstyles and fixes. This allows users to toggle on/off potential attractive forces at high velocities due to damping. By default, this option is turned off such that there is no change in behavior.

The granular wall fixes are also updated to include the mindlin/force and mindlin_rescale/force models which were previously added to pair granular.

**Related Issue(s)**

The issue of attractive forces has been previously brought up in the [LAMMPS mailing list](https://lammps.sandia.gov/threads/msg69973.html) and was originally addressed in [this commit](https://github.com/lammps/lammps/pull/510/commits/066123007cd9277d4c03669707cf34ab6c05bc06). However, this commit was reversed to allow time for discussion.

**Author(s)**

Joel Clemmer (SNL jtclemm@sandia.gov)
Dan Bolintineanu (SNL dsbolin@sandia.gov)

**Licensing**

**Backward Compatibility**

Pair cofficients will no longer be correctly read from old restart files.

**Post Submission Checklist**


doc/src/fix_wall_gran.rst
```diff
@@ -175,8 +177,6 @@ the clockwise direction for *vshear* > 0 or counter-clockwise for
 *vshear* < 0.  In this case, *vshear* is the tangential velocity of
 the wall at whatever *radius* has been defined.
 
-
```
akohlmey, line 178: why delete these lines?

src/GRANULAR/fix_wall_gran.cpp
```diff
@@ -118,6 +119,12 @@ FixWallGran::FixWallGran(LAMMPS *lmp, int narg, char **arg) :
       kt /= force->nktv2p;
     }
     iarg = 10;
+    
+    if (strcmp(arg[iarg],"limit_damping") == 0) {
+      limit_damping = 1;
```
akohlmey, line 124: I don't see any place where limit_damping is initialized to a default value (of 0). Is this code "valgrind clean"?

akohlmey (changes requested):
I've pointed out two issues that seem to need fixing.
I am also wondering whether fix wall/gran/region would need changes as well to be consistent with the changes in fix wall/gran.

doc/src/fix_wall_gran.rst
jtclemm: Oops, that was a mistake. I'll revert the deletion.

src/GRANULAR/fix_wall_gran.cpp
jtclemm: I'll also fix this. Checking with Valgrind, this fix has a few extra, unrelated issues with uninitialized values. I'll patch them all and submit an update later. 

In terms of fix wall/gran/region, no changes were necessary as it calculates forces using the methods defined in the parent wall/gran class.

akohlmey:
@jtclemm just noticed a minor coding style issue. We generally prefer using '&&' over 'and' and '||' over 'or' and so on.
And for multiple chained logical expressions, we try to insert parenthesis to define precedence as people are not always fully aware of some of all of the precedence rules.

jtclemm:
> 
> 
> @jtclemm just noticed a minor coding style issue. We generally prefer using '&&' over 'and' and '||' over 'or' and so on.
> And for multiple chained logical expressions, we try to insert parenthesis to define precedence as people are not always fully aware of some of all of the precedence rules.

I will make the swap and keep that in mind for the future (and for the multi PR I previously submitted).

I also added changes from upstream and fixed the remaining errors Valgrind.

akohlmey:
@jtclemm @dsbolin have you checked whether your changes need to be ported to the accelerated versions of the pair styles you are modifying in KOKKOS and USER-OMP? Since you have changes in the compute() functions, this seems quite likely.

jtclemm:
> 
> 
> @jtclemm @dsbolin have you checked whether your changes need to be ported to the accelerated versions of the pair styles you are modifying in KOKKOS and USER-OMP? Since you have changes in the compute() functions, this seems quite likely.

The granular pairstyle in KOKKOS should be updated. I did not realize USER-OMP had granular pairstyles, but you are right those need to be updated too. I'll work on that.

akohlmey (approved):

sjplimp (approved):

PR #2702, opened 2021-04-09
New pair style hybrid/scaled

akohlmey:
**Summary**

This adds a new pair style which is similar to hybrid/overlay, but in addition allows to specify factors by which forces and energies from each sub-style are multiplied before added to the total.  The scale factors may be (equal style) variables.

In addition there is a wrapper around pair style coul/cut called coul/cut/global, that only allows using a global cutoff and has a changed extract() method, so that it is compatible with long-range coulomb styles in hybrid pair styles. In combination with 'hybrid/scaled' this can now be used to *subtract* unwanted (local) coulomb interactions from a system without the damping due to the long-range coulomb, similar to how *inside* the long-range pair styles scaled interactions for "special" neighbors are treated.

While testing this, bugs with restarting and initializing data when using inferred cutoffs from mixing were found and fixed.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Implementation Notes**

This adds a force accumulator array to the pair style. When looping over the sub-styles, the forces are first cleared and then added to the accumulator array with the scaling factor applied. Similar for energies and virial. At the end of the force computation, the accumulated forces are copied to the regular per-atom force array.

**Post Submission Checklist**


akohlmey:
@sjplimp how about calling this style "hybrid/mixed" (or hybrid/cocktail? ;-))

sjplimp:
sorry, I think scaled is better than mixed, or even cocktail ...

akohlmey:
> sorry, I think scaled is better than mixed, or even cocktail ...

too bad, I was already preparing some "shaken, not stirred" jokes. ;-)

akohlmey:
@sjplimp documentation is now fully updated and all relevant items and limitations of `hybrid/scaled` discussed

doc/src/pair_hybrid.rst
```diff
@@ -13,15 +14,19 @@ pair_style hybrid/overlay command
 
 Accelerator Variants: *hybrid/overlay/kk*
 
+pair_style hybrid/scale command
```
sjplimp, line 17: scaled

doc/src/pair_hybrid.rst
```diff
@@ -37,36 +42,57 @@ Examples
    pair_coeff * * lj/cut 1.0 1.0
    pair_coeff * * coul/long
 
+   variable one equal ramp(1.0,0.0)
```
sjplimp, line 45: also include an example line with constants, not variables

doc/src/pair_hybrid.rst
```diff
@@ -37,36 +42,57 @@ Examples
...
-*hybrid/overlay* style, one or more pair styles can be assigned to
-each pair of atom types.  The assignment of pair styles to type pairs
-is made via the :doc:`pair_coeff <pair_coeff>` command.
+The *hybrid*, *hybrid/overlay*, and *hybrid/scaled* styles enable the
+use of multiple pair styles in one simulation.  With the *hybrid* style,
+exactly one pair style is assigned to each pair of atom types.  With the
+*hybrid/overlay* and *hybrid/scaled* styles, one or more pair styles can
+be assigned to each pair of atom types.  The assignment of pair styles
+to type pairs is made via the :doc:`pair_coeff <pair_coeff>` command.
+The *hybrid/scaled* style differs from the *hybrid/overlay* style by
```
sjplimp, line 60: The only difference between overlay and scaled is that scaled adds a scale factor for each sub-style's contribution ...

doc/src/pair_hybrid.rst
```diff
@@ -37,36 +42,57 @@ Examples
...
+requiring a factor for each pair style that is used to scale all
+forces, energies and stresses computed by each sub-style.  Because of
+the additional complexity, the *hybrid/scaled* style will have more
```
sjplimp, line 63: ... the scaled style has more overhead and thus may be slower ...

doc/src/pair_hybrid.rst
```diff
@@ -37,36 +42,57 @@ Examples
...
+*hybrid/overlay* style could be used as in the second example above,
+where multiple potentials are superposed in an additive fashion to
+compute the interaction between atoms.  In this example, using *lj/cut*
+and *coul/long* together gives the same result as if the
+*lj/cut/coul/long* potential were used by itself.  In this case, it
+would be more efficient to use the single combined potential, but in
+general any combination of pair potentials can be used together in to
+produce an interaction that is not encoded in any single pair_style
+file, e.g. adding Coulombic forces between granular particles.  The
+*hybrid/scaled* style enables more complex combinations of pair styles
```
sjplimp, line 80: Make this sentence start a new paragraph.

If h/scaled were used in place of h/overlay, then each sub-style contribution could be weighted, e.g. fractional ... or negative ...
Furthermore ... during a simulation, this enables smoothly switching ...

doc/src/pair_hybrid.rst
```diff
@@ -260,7 +288,10 @@ the specific syntax, requirements and restrictions.
 ----------
 
 The potential energy contribution to the overall system due to an
-individual sub-style can be accessed and output via the :doc:`compute pair <compute_pair>` command.
+individual sub-style can be accessed and output via the :doc:`compute
```
sjplimp, line 291: Confused by this unscaled comment.  Is that somehow different (for some pair styles?) than what the user would expect?

sjplimp (changes requested):
see my comments/Qs on PH doc page

doc/src/pair_hybrid.rst
akohlmey: I put this here to emphasize that the scaling only happens when when contributions are summed. People may get confused when the (plain) sum of the contributions from `compute pair` is different than the value of `epair`.

akohlmey:
@sjplimp docs are updated

sjplimp:
@akohlmey good to merge not, but I'm not seeing a Review button any more

sjplimp:
@akohlmey sorry, meant good to merge now

akohlmey:
> @akohlmey sorry, meant good to merge now

re-requested the review. still need your approval to merge.

sjplimp (approved):

PR #2703, opened 2021-04-09

akohlmey:
**Summary**

This pull request collects multiple sets of small changes and bugfixes

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

- correct mention of ghost atoms in the description of how pressure is computed from F dot r.
- avoid temporary fixed size char buffers on the stack when piping files to/from gzip
- correct precedence in the equilibrium angle computation for angle style cosine/periodic
- simplify suffix detection for binary or compressed files
- avoid uninitialized data access reported by static code analysis
- update include file ordering to match current conventions in several files
- avoid redundant check in hybrid pair style coeff parsing and thus simplify that code
- fix bug in processing `pair_modify nofdotr`
- update "melt" example to use velocity with "loop geom" so that serial and parallel runs have the same thermo output

**Post Submission Checklist**


src/GRANULAR/fix_wall_gran.cpp
```diff
@@ -1261,6 +1261,7 @@ void FixWallGran::granular(double rsq, double dx, double dy, double dz,
 
   k_tangential = tangential_coeffs[0];
   damp_tangential = tangential_coeffs[1]*damp_normal_prefactor;
+  Fscrit = tangential_coeffs[2] * Fncrit;
```
akohlmey, line 1264: @dsbolin @jtclemm please note the changes in this commit. Our weekly static code analysis has reported that `Fscrit` was used before it was initialized.  You may want to check any simulations that used the affected features.

dsbolin: Thanks for catching this. The bug was introduced by me very recently in PR #2698 , before that `Fscrit` was not used until after line 1347. Older simulations should be fine.

sjplimp (approved):

stanmoore1:
@akohlmey can you add this change to this PR:

```patch
diff --git a/src/finish.cpp b/src/finish.cpp
index 810155f..5e5cd72 100644
--- a/src/finish.cpp
+++ b/src/finish.cpp
@@ -401,7 +401,7 @@ void Finish::end(int flag)
   }
 #endif
 
-  if (lmp->kokkos && lmp->kokkos->ngpus > 0)
+  if (comm->me == 0 && lmp->kokkos && lmp->kokkos->ngpus > 0)
     if (const char* env_clb = getenv("CUDA_LAUNCH_BLOCKING"))
       if (!(strcmp(env_clb,"1") == 0)) {
         error->warning(FLERR,"Timing breakdown may not be accurate "
```

PR #2705, opened 2021-04-12

akohlmey:
**Summary**

This pull request combines multiple small changes and bugfixes

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

The versioning of the LAMMPS module has been changed to comply with Python packaging requirements.

**Implementation Notes**

- fixes for version number and package metadata handling for the LAMMPS python module. Previously unit tests could fail when a different version of LAMMPS would be installed as a local package
- eliminate use of temporary char buffers and s(n)printf() in QEQ package
- make sure that `comb3` pair style is not matched when looking for `comb`
- update USER-PLUMED to support new versions 2.5.7, 2.6.3, 2.7.1
- small tweaks in the documentation
- accumulated whitespace fixes
- include file updates from IWYU report (mostly deletions)
- silence compiler warnings when compiling with Python 2.7
- use safe fread in pair style lj/relres
- add a single string convenience version of `Neighbor::modify_params()` and apply it
- handle a bunch of possible NULL pointer dereferences in fix charge/regulation reported by static code analysis
- search for faster linkers to link unit test tools only on verified linux distributions

**Post Submission Checklist**


src/PYTHON/fix_python_invoke.cpp
```diff
@@ -92,8 +92,9 @@ int FixPythonInvoke::setmask()
 void FixPythonInvoke::end_of_step()
 {
   PyUtils::GIL lock;
+  char fmt[] = "O";
 
-  PyObject * result = PyObject_CallFunction((PyObject*)pFunc, "O", (PyObject*)lmpPtr);
+  PyObject * result = PyObject_CallFunction((PyObject*)pFunc, fmt, (PyObject*)lmpPtr);
```
rbberger, line 97: @akohlmey Just make it a cast like the others. I dislike it either way, but that's the price we pay for Python 2 support.

rbberger (approved):

src/PYTHON/fix_python_invoke.cpp
akohlmey: thanks for reminding me. this was my intention, but over all the other changes, I must have forgotten to change this back.

sjplimp (approved):

Issue #2643, opened 2021-03-05
Provide clear and detailed guidelines and requirements for potential contributions to LAMMPS


akohlmey:
**Summary**

We need to provide more specific and explicit guidelines and lists of requirements for contributions to LAMMPS.
This information is currently scattered across multiple files with different level of detail and different target audiences.
This should be condensed to a single location with all information and that should also indicate clearly which are
strict requirements (i.e. code will not be included unless those are met) and what is stating a preference (i.e. which are
optional or where the degree of compliance can be negotiated).

**Details**

https://github.com/lammps/lammps/blob/master/.github/CONTRIBUTING.md
https://github.com/lammps/lammps/blob/master/doc/include-file-conventions.md
https://github.com/lammps/lammps/blob/master/doc/documentation_conventions.md
https://github.com/lammps/lammps/blob/master/doc/github-development-workflow.md
https://docs.lammps.org/Modify_contribute.html

A draft document that may be edited and will be expanded follows below.

------

# Requirements for contributing code to LAMMPS

The following is a summary of the current requirements for including
contributions into the LAMMPS software distribution.  The LAMMPS
developers are committed to providing a software package that is
versatile, high-quality, portable, and easy to maintain or modify.
The requirements and preferences listed in this document are aimed
to maintain or improve on that status.

It is indicated which requirements are strict, and which represent a
preference but are negotiable or optional

Please feel free to contact the LAMMPS core developers in case you need
additional explanations or clarifications or in case you need assistance
in realizing the (strict) requirements for your contributions.

## Licensing requirements (strict)

Contributing authors agree that their contributions can be
distributed under the LAMMPS license conditions. This is the GNU public
license in version 2 (not 3 or later) for the publicly distributed
versions, e.g. on the LAMMPS homepage or on GitHub.  On request we also
make a version of LAMMPS available under LGPL 2.1 terms; this will
usually be the latest available or a previous stable version.

Contributions may be under a different license so long as that
license does not conflict with the aforementioned terms. Contributions
that use code with a conflicting license can be split into two parts:

1. the core parts (i.e. parts that must be in the `src` tree) that
   are licensed under compatible terms and bundled with the LAMMPS
   sources
2. an external library that must be downloaded and compiled
   (either separately or as part of the LAMMPS compilation)

Please note, that this split licensed mode may complicate including the
contribution in binary packages.

## Using Pull Requests on GitHub (preferred)

All contributions to LAMMPS are processed as pull requests on GitHub
[...]

PR #2708, opened 2021-04-15
Refactoring and improving REAXFF and QEQ packages

akohlmey:
**Summary**

This PR updates the REAXFF and QEQ package code to make it more compliant to LAMMPS coding conventions and remove unused functionality.

The following sets of changes are implemented
- correct support for per-atom stress for REAXFF version and OPENMP version
- rename of styles and files to have reaxff instead of reax/c or reax with full backward compatibility.
- style headers no longer "leak" internal data structure definitions, pair styles have been PIMPL-ified
- the support code has been placed in a ReaxFF namespace
- the number of header files has been greatly reduced
- unused structs and data types are removed
- unused function arguments are removed
- local functions have been declared static
- declaration of structs/unions are converted to C++
- refactoring of control file and force field file reading/parsing using TextFileReader class
- removal of "native" trajectory output
- inactive/ignored control file parameters will cause a warning
- fix qeq/reaxff now computes a global scalar which is the number of QEq iterations
- made handling the maxiter limit in fix qeq/reax and counting the iterations consistent across all variants including the "dual" setting for fix qeq/reaxff/omp
- added a "nowarn" keyword to fix qeq/reax variants, so that with a very small tolerance and setting maxiter to a sufficient but small er number than the default always a fixed number of QEq iterations is used, which in turn will produce consistent QEq results in serial and parallel and thus minimizes divergence of trajectories in those cases, which is useful for regression testing.

Also the reading of the parameter file in the QEQ package has been updated to use the TextFileReader class, to support a "warn no" option, the example input has been updated to use all qeq fixes and unit tests YAML files were added for some tests.

About 2,800 lines are removed by this PR.

**Related Issue(s)**

#2643 

**Author(s)**

Axel Kohlmeyer, Temple U
Aidan Thompson, SNL

**Licensing**

**Backward Compatibility**

- While lots of files and styles have been renamed, it is intended to retain full backward compatibility
  by defining aliases for all styles, keywords, and checks for style names. In many cases the check has been replaced with a 
  regexp check (using utils::strmatch) for "^reax.." which will match both the old "reax/c" as well as the new "reaxff" containing  
  names.
- Support for writing "native" trajectories in the reaxff pair style has been removed.
 If this feature is reguired, LAMMPS version 2 July 2021 or older must be used.

**Implementation Notes**

A new constructor has been added to the `TextFileReader` class that takes an already opened file descriptor instead of a file name. Also, the `next_values()` method has been updated to throw an `EOFException` if the next line is a null pointer.

**Post Submission Checklist**


akohlmey:
@athomps after the extended changes and removal of the "native" trajectory format, I have small differences in the computed pressure and after longer runs also in energies. There also are differences between serial and parallel runs, but they are smaller in magnitude. Now I am a bit at a loss and looking for ideas how I can determine the origin and whether they are significant or whether I (by accident) fixed a bug due to some (small) data corruption caused by the removed code. The energies look like they are now better conserved, but then again, that can also be by chance. Do you have any suggestions?

athomps:
@akohlmey ReaxFF trajectories are particularly fragile because very small differences in any part of the ReaxFF force calculation can result in addition or removal of an extra iteration in the charge equilibration solver, causing substantial differences in subsequent trajectory.

athomps:
I ran my own tests of `examples/reax/in.reaxc.rdx` and `examples/reax/in.reaxc.tatb` comparing master and refactor-reaxff-parser, yielding no differences.

athomps:
@akohlmey What exactly are the differences you are seeing?

akohlmey:
> @akohlmey What exactly are the differences you are seeing?

For serial and parallel, I also don't see any differences between `master` and `HEAD` of this PR with the rdx and tatb examples. But they differ form the reference log files in the folder and also there is a difference between serial and parallel.
I am attaching some outputs from the latest code version.

[log.tatb.parallel-new.txt](https://github.com/lammps/lammps/files/6333038/log.tatb.parallel-new.txt)
[log.tatb.serial-new.txt](https://github.com/lammps/lammps/files/6333039/log.tatb.serial-new.txt)
[log.tatb.serial-old.txt](https://github.com/lammps/lammps/files/6333040/log.tatb.serial-old.txt)
[log.rdx.serial-old.txt](https://github.com/lammps/lammps/files/6333041/log.rdx.serial-old.txt)
[log.rdx.serial-new.txt](https://github.com/lammps/lammps/files/6333042/log.rdx.serial-new.txt)

akohlmey:
> @akohlmey ReaxFF trajectories are particularly fragile because very small differences in any part of the ReaxFF force calculation can result in addition or removal of an extra iteration in the charge equilibration solver, causing substantial differences in subsequent trajectory.

For the sake of getting more consistent results for regression testing, would it make sense to set the QeQ convergence to be very tight? so tight that always the (same) maximum number of QeQ iterations is reached?

akohlmey:
> > @akohlmey ReaxFF trajectories are particularly fragile because very small differences in any part of the ReaxFF force calculation can result in addition or removal of an extra iteration in the charge equilibration solver, causing substantial differences in subsequent trajectory.
> 
> For the sake of getting more consistent results for regression testing, would it make sense to set the QeQ convergence to be very tight? so tight that always the (same) maximum number of QeQ iterations is reached?

To answer my own question, yes it does. With setting maxiter to 100 and convergence to 1.0e-10 I get very consistent results between parallel and serial runs. Will have to look into adding an option to the qeq fixes so they can be better tested (and don't produce a gazillion of warnings).

akohlmey:
@athomps quick update. I've added the option to return the number of iterations to fix qeq/reax and its USER-OMP and KOKKOS variants as well as made the counting consistent.  With the addition of a "nowarn" keyword it is now possible to use a small tolerance and then have a fixed number of QEq iterations without getting flooded with warnings, and those produce very consistent results across all versions, running in serial, MPI parallel or OpenMP parallel.

So it seems to me that I you agree that the changes in this PR are consistent and correct as they currently are, we will need to update the input decks for regression testing to require a fixed number of QEq iterations and thus achieve better reproducible and consistent results in the future.

athomps:
I think I have resolved the problem with nofdotr, which just required removing dead code and twiddling the v_tally settings.  I am having trouble pushing to this branch, but I will try again later.

akohlmey:
> I think I have resolved the problem with nofdotr, which just required removing dead code and twiddling the v_tally settings. I am having trouble pushing to this branch, but I will try again later.

@athomps this is excellent news. Thanks for taking the time. You could also just email me the `git diff` output or push the modified branch into your personal repository and tell me. I can grab what is missing and then run the checks again.

athomps:
nofdotr fix is now pushed to this branch.  It gives the same pressure with/without nofdotr. I also verified that per-atom energy and stress sum to the global counterparts, which was already the case.  I have not attempted to propagate these changes to OMP or KOKKOS.  Removing the dead code and streamlining the treatments of various vflags should make that easier.  I may not have gotten everything right, so it would be good to have another pair of eyeballs on this.

athomps:
I uncovered an additional problem due to n-fold undercounting of contributions to the global virial tensor in the manybody Pair::v_tally() functions.  In addition to ReaxFF, it also affected the global virials for Tersoff and AIREBO, but only with non-default nofdotr flag, and the scalar pressure was unaffected, which is probably how it escaped notice until now..  I think these modifications are more consistent and less arbitrary. Because it was getting so complicated, I made a script to check all 4 paths to the global stress tensor for AIREBO, Tersoff, and ReaxFF. See report in next comment.

athomps:
```
PXY  Pair=airebo          pxy        pxyatom       deltapxyatom
PXY Regular      -3648.64173046569 -3648.64173046575 -5.9...e-11
PXY With nofdotr -3648.64173046573 -3648.64173046575 -1.8...e-11
PXY Delta        -3.9...e-11          0               4.0...e-11
PXY  Pair=tersoff          pxy        pxyatom       deltapxyatom
PXY Regular      -8461.66861756007 -8461.66861756012 -4.9...e-11
PXY With nofdotr -8461.66861756011 -8461.66861756012 -9.0...e-12
PXY Delta        -4.0...e-11          0               4.0...e-11
PXY  Pair=reax          pxy        pxyatom       deltapxyatom
PXY Regular      -12739.3756897676 -12739.3756897669 6.9...e-10
PXY With nofdotr -12739.3756898521 -12739.3756898506 1.5...e-09
PXY Delta        -8.4...e-08        -8.3...e-08      8.2...e-10
```

athomps:
I will see if I can port these changes to OMP now.

athomps:
I beat on the reaxc/omp pretty hard and fixed a large number of minor things.  It passes the tatb test (although the tatb reference file got replaced with rdx output, which I did not fix).  It also gets a lot of other things right, for example  per-atom energy and virial matches global counterparts under a variety of conditions.  But global stress is still not right with `nofdotr` and `-sf omp.`  The biggest limitation is that I tried to add some new `v_tally*thr_proxy()` functions to thr_omp.h, but could not get them to work.  They are commented out in `reaxc_torsion_angles_omp.cpp` and `reaxc_valence_angles_omp.cpp`

athomps:
In fact, running with -sf omp gives non-reproducible results.

akohlmey:
> In fact, running with -sf omp gives non-reproducible results.

@athomps I will have a look at this after lunch. To minimize merge conflicts, please wait until I have assigned the PR to you before making any changes and then start with pulling whatever I have added. You can then signal to me that you want me to do more changes by assigning me again. Thanks.

akohlmey:
@athomps quick update. yesterday has been busy with a lot of things, but I did get started on this. However, this PR is becoming difficult to bring to a conclusion since there are now multiple issues addressed in it that are somewhat connected. Thus it seems easier to first resolve the non-ReaxFF issues and do that separately. I have already "extracted" the file reader/parser changes to a different branch (which is now merged). Next will be the nofdotr changes. For that I have augmented our pair style testing tool to insert a check where forces/energies/stress from a "run 0" are compared between the reference data and a call with and without "pair_modify nofdotr" for both plain and /omp styles. It looks like there is an issue with the corresponding tally functions or something else in USER-OMP that will need to be addressed.

akohlmey:
@athomps don't be surprised to see your changes (temporarily) vanish from this branch. I have reset it to the state before you started correcting the nofdotr situation. I will now first merge with upstream to get the general changes from PR #2736, resolve any conflicts and then update it with those parts of your changes that still apply and also try to fix any remaining issues also in the OPENMP version.

If you still have a local checkout, you should abandon it, since after the forced update your commit history will have diverged from mine (no worries, I kept a copy of the original branch on my local machine).

akohlmey:
@athomps and @sjplimp this has now the accumulation of per-atom and global virial consistent between the regular and OpenMP version and produces the same energies and pressure with and without `pair_modify nofdotr` and between regular and OpenMP. Turned out less effort than expected in the end. Implementing PR #2736 first really helped and made things here more straightforward.

Now starting with the last step of renaming the styles and files.

akohlmey:
@sjplimp @athomps @stanmoore1 this has now all the discussed changes, it compiles and produces consistent results with the REAXFF, OPENMP, KOKKOS (w/ OpenMP) versions. Stress/atom is consistent (and doesn't crash) with OPENMP as well and the fdotr based pressure is consistent with the one computed without as well as the one from compute stress/atom and compute reduce. Also I have tested it with a selection of old and updated input scripts and all checks and keywords have been adapted so that they now support using "reaxff" instead of "reax/c" while remaining backward compatible. The documentation has been updated to (only) feature the new syntax.

Thus unless you can find anything is still missing, this is now ready for merging.
The next step for refactoring this code would be to change the memory management to abandon the old style scratch array approach and instead use the Memory class with its create/destroy/grow methods. But this could be a project for someone else...

src/KOKKOS/pair_reaxff_kokkos.h
```diff
@@ -81,25 +85,25 @@ template<int NEIGHFLAG>
...
-struct PairReaxComputeBond2{};
+struct PairReaxFFomputeBond2{};
 
 template<int NEIGHFLAG, int EVFLAG>
-struct PairReaxComputeMulti1{};
+struct PairReaxFFomputeMulti1{};
 
 template<int NEIGHFLAG, int EVFLAG>
-struct PairReaxComputeMulti2{};
+struct PairReaxFFomputeMulti2{};
```
stanmoore1, line 97: Find/replace mistake here and some other places, should be `PairReaxFFCompute*`

src/KOKKOS/pair_reaxff_kokkos.h
```diff
@@ -15,19 +15,21 @@
 
 #ifdef PAIR_CLASS
 // clang-format off
-PairStyle(reax/c/kk,PairReaxCKokkos<LMPDeviceType>);
-PairStyle(reax/c/kk/device,PairReaxCKokkos<LMPDeviceType>);
-PairStyle(reax/c/kk/host,PairReaxCKokkos<LMPHostType>);
+PairStyle(reaxff/kk,PairReaxFFKokkos<LMPDeviceType>);
+PairStyle(reaxff/kk/device,PairReaxFFKokkos<LMPDeviceType>);
+PairStyle(reaxff/kk/host,PairReaxFFKokkos<LMPHostType>);
+PairStyle(reax/c/kk,PairReaxFFKokkos<LMPDeviceType>);
```
stanmoore1, line 21: This code seems duplicated?

stanmoore1 (changes requested):
I put a few change requests in the code. Kokkos changes look good otherwise. One comment: do we have to use `ReaxFF` everywhere (instead of `Reax`), like in `fix qeq/reaxff`? Seems cumbersome.

src/KOKKOS/pair_reaxff_kokkos.h
akohlmey: This is intentional. This way we stay backward compatible.
Input files that have 'reax/c' instead of 'reaxff' will continue to work.
Please see, e.g., src/REAXFF/pair_reaxff.h that has the same kind of duplicated line.

src/KOKKOS/pair_reaxff_kokkos.h
akohlmey: well spotted. thanks.

src/KOKKOS/pair_reaxff_kokkos.h
stanmoore1: Ah I missed that, thanks.

stanmoore1 (approved):
See my comment about `ReaxFF` vs `Reax`.

akohlmey:
> I put a few change requests in the code. Kokkos changes look good otherwise. One comment: do we have to use `ReaxFF` everywhere (instead of `Reax`), like in `fix qeq/reaxff`? Seems cumbersome.

I had asked about how to handle those cases and @sjplimp and @athomps both wanted qeq/reaxff and similar.
I didn't change all internal variables and types, though (there is still reaxflag and similar).

athomps:
> > do we have to use `ReaxFF` everywhere (instead of `Reax`), like in `fix qeq/reaxff`? Seems cumbersome.
> 
> I had asked about how to handle those cases and @sjplimp and @athomps both wanted qeq/reaxff and similar.
> I didn't change all internal variables and types, though (there is still reaxflag and similar).

The official name is ReaxFF, so it makes sense to use that consistently in the user-facing names.

stanmoore1:
>The official name is ReaxFF, so it makes sense to use that consistently in the user-facing names.

Alright, I can live with that.

PR #2714, opened 2021-04-19
Improvement of the SPIN package by removing a limitation

julient31:
make html**Summary**

This PR:
- corrected a limitation of the SPIN package: only one fix precession/spin or langevin/spin could be declared for each simulation. This is no longer the case.
- implemented a spin-transfer torque in fix precession/spin
- improved the test_problem setup

**Related Issue(s)**

None.

**Author(s)**

Julien Tranchida (Sandia National Labs), julien.tranchida1@gmail.com
Theophile Chirac (CEA, helped with the implementation of the spin-transfer torque), theophile.chirac@cea.fr

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

Yes: all former input scripts will still be fully valid.

**Implementation Notes**

All validation tests of the SPIN package were successfully passed. This is important as changes were made in fix nve/spin, langevin/spin and precession/spin. 
The documentation was modified to account for the new possible interaction (SST in fox precession/spin)and successfully recompiled (make html). 

**Post Submission Checklist**

- [x] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [x] The source code follows the LAMMPS formatting guidelines
- [x] Suitable new documentation files and/or updates to the existing docs are included
- [x] The added/updated documentation is integrated and tested with the documentation build system
- [x] The feature has been verified to work with the conventional build system
- [x] The feature has been verified to work with the CMake based build system
- [ ] Suitable tests have been added to the unittest tree.
- [ ] A package specific README file has been included or updated
- [ ] One or more example input decks are included

PR #2693, opened 2021-04-01
Adding a spin/kokkos atom style

julient31:
**Summary**

This PR contains three main features:

- adding a spin/kokkos atom style 
- correcting a small bug in pair_spin_exchange_biquadratic.cpp
- adding the possibility of running fix nve/spin on non-triclinic boxes

**Related Issue(s)**

None

**Author(s)**

Julien Tranchida (Sandia)
Stan Moore (Sandia)
Svetoslav Nikolov (Sandia)

Contact: julien.tranchida1@gmail.com

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

All changes should be completely transparent to the associated LAMMPS input files.

**Implementation Notes**

A change was made in verlet_kokkos.cpp: adding the fm and fm_long spin vectors to the associated force_clear. 

All spin validation tests (in examples/SPIN/test_problems) were adapted and  successfully ran to check the new atom style. 

**Post Submission Checklist**

- [x] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [x] The source code follows the LAMMPS formatting guidelines
- [x] Suitable new documentation files and/or updates to the existing docs are included
- [x] The added/updated documentation is integrated and tested with the documentation build system
- [x] The feature has been verified to work with the conventional build system
- [x] The feature has been verified to work with the CMake based build system
- [ ] Suitable tests have been added to the unittest tree.
- [ ] A package specific README file has been included or updated
- [x] One or more example input decks are included

**Further Information, Files, and Links**

PR #2717, opened 2021-04-22

akohlmey:
**Summary**

This pull request combines multiple small changes and bugfixes

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues

**Implementation Notes**

The following individual changes
- Fix whitespace issues introduced in recent merges
- Add ASCII substitution support for Unicode general punctuation left/right single/double quotation marks in LAMMPS input script
- Fix a memory leak in fix atom/swap
- Fix bug in spin/kk atom style border communication introduced in PR #2693
- Fix cut-n-paste bug and incorrect detection of number of active pair style instance introduced in PR #2714. Also improve accuracy of style name matching.
- Use library name consistent with other libraries build during LAMMPS build for PACE lib
- Update package lists in CMake presets and src/Makefile variables to include recent additions
- Silence some compiler warnings
- Output compiled in accelerator configuration with -help
- Download a specific MathJax version for building the html/epub docs. Update the version to the latest bugfix release, 3.1.4
- Tweak a few epsilon values for pair style unit tests so they pass with GPU acceleration in mixed precision mode

**Post Submission Checklist**


src/KOKKOS/atom_vec_spin_kokkos.cpp
```diff
@@ -637,11 +637,12 @@ int AtomVecSpinKokkos::unpack_border_hybrid(int n, int first, double *buf)
 
   m = 0;
   last = first + n;
-  for (i = first; i < last; i++)
+  for (i = first; i < last; i++) {
```
akohlmey, line 640: @julient31 FYI this was flagged by compiling with clang and using full compiler warnings (-Wall -Wextra)

julient31: Hey @akohlmey !

Ouch, yes, that was quite horrendous... How did I miss that? ^^

Thanks for finding this Axel!

julient31 (approved):

src/KOKKOS/atom_vec_spin_kokkos.cpp
akohlmey: I didn't find it, the Clang compiler did. When compiling with GCC I use `-Wall -Wextra -Wno-maybe-uninitialized` and when using Clang I use `-Wall -Wextra` and that produces warnings about unused variables, set but unused variables, unused local functions, local variables shadowing class members, and so on. When there are warnings about "inconsistent indentation" it is often (but not always) a sign of missing braces like in this case. We try to reduce such "easily fixable" compiler warnings to the minimum (sometimes they cannot be avoided or the fix would only help hiding but not fixing problems) so that new warnings stand out more. This is essentially a first round of static code analysis to help maintaining code quality. As second round it running specific static analysis tools like ["Coverity Scan"](https://scan.coverity.com/projects/lammps) or ["CodeQL"](https://github.com/lammps/lammps/security/code-scanning) that are run either after every merge to the master branch (CodeQL, faster but less capable) or once a week (Coverity Scan, slower but more powerful) and review the reported issues (there are plenty of false positives because of the coding style and flow of control employed in LAMMPS), but they can point out common programming errors that are sometimes hard to find.

julient31: I took notes and saved your message. Actually, I remember you already mentioned the -Wall -Wextra -Wno-maybe-uninitialized  flags for gcc. I need to more systematically check this before submitting PRs.

julient31: Thanks a lot for the explanation Axel!

akohlmey: > I took notes and saved your message. Actually, I remember you already mentioned the -Wall -Wextra -Wno-maybe-uninitialized flags for gcc. I need to more systematically check this before submitting PRs.

Then you should also make a note to do "make check-whitespace" before submitting and possibly "make fix-whitespace" (and then check and commit the changes for "your" files) to remove whitespace-only changes that produce extra differences when comparing different commits.

julient31: Done! ;-)

src/SPIN/fix_nve_spin.cpp
```diff
@@ -221,74 +227,74 @@ void FixNVESpin::init()
...
     }
   }
-  
-  // init length of vector of ptrs to precession/spin styles
+
+  // init length of vector of ptrs to langevin/spin styles
 
   if (nlangspin > 0) {
-    locklangevinspin = new FixLangevinSpin*[nprecspin];
+    locklangevinspin = new FixLangevinSpin*[nlangspin];
```
akohlmey, line 275: @julient31 ... and here we have a cut-n-paste bug that was detected by using static code analysis with coverity scan.

src/SPIN/fix_nve_spin.cpp
```diff
@@ -197,18 +203,18 @@ void FixNVESpin::init()
   int count1 = 0;
   if (npairspin == 1) {
     count1 = 1;
-    spin_pairs[0] = (PairSpin *) force->pair_match("spin",0,0);
+    spin_pairs[0] = (PairSpin *) force->pair_match("^spin",0,0);
```
akohlmey, line 206: @julient31 please note that not using the caret (^) can lead to false matches, same as for comparisons with `strstr()`. There may be styles that are not in the SPIN package and just by accident contain the string "spin" somewhere.
That is why we introduced the `utils::strmatch()` function, which has been applied here and is used internally by `Force::pair_match()` for inexact matches. Searching for styles that **start** with the string spin is much more specific.

src/SPIN/fix_nve_spin.cpp
```diff
@@ -172,15 +175,18 @@ void FixNVESpin::init()
 
   // loop 1: obtain # of Pairs, and # of Pair/Spin styles
 
-  if (force->pair_match("spin",0,0)) {        // only one Pair/Spin style
-    pair = force->pair_match("spin",0,0);
-    npairs = pair->instance_total;
```
akohlmey, line 177: @julient31 I noticed this while improving the inexact matches. `Pair::instance_total` is **not** the total number of pair styles but the total number of pair styles **ever created**. This is to give a unique number to identify pair styles that are otherwise the same, e.g. when using the same pair style as a hybrid sub-style. This is used for example by the neighbor list code to by able to distinguish individual neighbor list requests. So if you replace an initial pair style (say lj/cut with morse) its number will become 2 even though there is only one pair style instance active.

src/velocity.cpp
```diff
@@ -578,13 +578,9 @@ void Velocity::scale(int /*narg*/, char **arg)
 
   int tflag = 0;
   if (temperature == nullptr) {
-    char **arg = new char*[3];
-    arg[0] = (char *) "velocity_temp";
-    arg[1] = group->names[igroup];
-    arg[2] = (char *) "temp";
-    temperature = new ComputeTemp(lmp,3,arg);
+    modify->add_compute(fmt::format("velocity_temp {} temp"));
```
rbberger, line 581: @akohlmey that was too simple ;-)
```suggestion
    modify->add_compute(fmt::format("velocity_temp {} temp", group->names[igroup]));
```

PR #2719, opened 2021-04-22
Overload utils::logmesg(), Error::all(), and Error::one() so they can take a format string and a variable number of arguments

akohlmey:
**Summary**

This follows the approach from fmtlib to implement the "fmt::print()" convenience function.
With this change we can replace `utils::logmesg(lmp,fmt::format(...))` with `utils::logmesg(lmp,...)` and do the equivalent changes for calls to `Error::all()` and `Error::one()`. Some other functions could get the same treatment (e.g. `Error::warning()`, `Error::message()`, `Error::universe_all()`, and `Error::universe_one()`).

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

Yes. we retain the original version since the variable args version requires at least one argument to the format string.

**Implementation Notes**

This follows the example given in the fmtlib docs: https://fmt.dev/latest/api.html#argument-lists

**Post Submission Checklist**


rbberger (approved):

sjplimp (approved):
this looks nice and clean

PR #2724, opened 2021-04-23
Consistent LAMMPS header comments and URL fixes

akohlmey:
**Summary**

This PR adds missing LAMMPS comment header to files in the src/ tree where missing, updates the homepage URL, flags them as C++ and applies a few more similar fixes.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

Yes. only comments are changed.

**Post Submission Checklist**

### DIFF
--- a/src/ASPHERE/compute_erotate_asphere.h
+++ b/src/ASPHERE/compute_erotate_asphere.h
@@ -1,6 +1,6 @@
 /* -*- c++ -*- ----------------------------------------------------------
    LAMMPS - Large-scale Atomic/Molecular Massively Parallel Simulator
-   http://lammps.sandia.gov, Sandia National Laboratories
+   https://lammps.sandia.gov/, Sandia National Laboratories
    Steve Plimpton, sjplimp@sandia.gov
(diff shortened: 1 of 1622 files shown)


athomps (approved):
I accept, although it is not clear that the change from http to https is necessary, since both work, even from Sandia network.

akohlmey:
> I accept, although it is not clear that the change from http to https is necessary, since both work, even from Sandia network.

The main reason is to have everything (more) consistent, so that when we have the need to make changes (e.g. to the LAMMPS home page URL), they can be automated. Also some files were missing the LAMMPS comment entirely (which is required to apply the  GPL).

Mind you, many current web browsers will switch automatically from http:// to https:// if the latter is available, some web browsers are encouraging using an https:// only mode and will print increasingly serious warnings about "unsafe" connections if you don't have/use https://. So it is just a matter of time until this change will be required.

rbberger (approved):

sjplimp (commented):
alll good

PR #2733, opened 2021-04-26
Update Kokkos library in LAMMPS to v3.4.0

stanmoore1:
**Summary**

Update Kokkos library in LAMMPS to v3.4.0

Highlights:

- SYCL Backend Almost Feature Complete
- OpenMPTarget Backend Almost Feature Complete
- Performance Improvements for HIP backend

**Related Issue(s)**

None.

**Author(s)**

Stan Moore (SNL)

**Licensing**

**Backward Compatibility**

Yes


stanmoore1:
Starting Kokkos regression tests.

stanmoore1:
Kokkos regression tests passed for GPU and CPU. Still need to check performance.

akohlmey:
@stanmoore1 the requirement of CMake version 3.16 is a serious blow against portability to a widely used Linux system. E.g. on Ubuntu 18.04LTS you have CMake 3.10 (that is *why* we chose 3.10 as minimum even though 3.12 would have avoided having to do a bunch of hacks to be compatible). Compiling/installing a newer version from source is not an option for a lot of LAMMPS users. Unless we change out CI testing accordingly, this PR will not become eligible for merging because it will keep failing the compilation tests.

stanmoore1:
> @stanmoore1 the requirement of CMake version 3.16 is a serious blow against portability to a widely used Linux system. E.g. on Ubuntu 18.04LTS you have CMake 3.10 (that is _why_ we chose 3.10 as minimum even though 3.12 would have avoided having to do a bunch of hacks to be compatible). Compiling/installing a newer version from source is not an option for a lot of LAMMPS users. Unless we change out CI testing accordingly, this PR will not become eligible for merging because it will keep failing the compilation tests.

That is unfortunate. @crtrott @junghans @masterleinad Can you please comment on why Kokkos needed to update to CMake version 3.16? This seems to be a hardship for LAMMPS.

crtrott:
3.16 is the default in the 20LTS Ubuntu line. Basically there was a whole lot of stuff we were able to drop by going to 3.16 and a number of things we couldn't support pre at least 3.14.  Generally ASC / ECP went to 3.17 (which for example is the minimum for Trilinos now) and we didn't hear any pushback when we announced this. The main problem for us is that our CMake system is already ridiculously complicated due to requirements for supporting all kinds of stuff CMake doesn't really support (including a whole bunch of compilers older cmakes don't know anything about). 

Installing cmake goes like this btw. on linux x86:
```
wget https://github.com/Kitware/CMake/releases/download/v3.20.1/cmake-3.20.1-linux-x86_64.sh
sh cmake-3.20.1-linux-x86_64.sh --prefix=INSTALL_PATH --skip-license
```

That said I get that you may not want to ask that from everyone. But we made the decision that compared to "you need a new C++14 compiler" this isn't a large burden either.  Truth be told my main problem was that I don't have enough CMake people to maintain CMake 3.10 compatibility after Jeremiah left.

crtrott:
The Makefile system still works without CMake btw. :-) That one turns out to be easier to maintain in general (though it doesn't come with as many safe guards and auto detections etc.).

jrmadsen:
FWIW, there are many ways to get pre-compiled cmake binaries. One of the easiest is via python+pip (which nearly everyone has installed):

```console
python -m pip install cmake>=3.16
python -m pip install --user cmake>=3.16
```

akohlmey:
> FWIW, there are many ways to get pre-compiled cmake binaries. One of the easiest is via python+pip (which nearly everyone has installed):
> 
> ```
> python -m pip install cmake>=3.16
> python -m pip install --user cmake>=3.16
> ```

You are completely missing the point.  **I** don't need to learn 20 easy ways of installing CMake. I am acting as an advocate of the LAMMPS user community and pointing out that this kind of change will create problems for them (as did the change to C++14 instead of C++11, but that just had less impact to the range Linux platforms we have set as our goal to support). In fact, we only raised the minimum of C++11 in LAMMPS after a long discussion weighing the benefits versus the disadvantages and we **still** have users that are confused with the implications and need help to resolve them.

We will just exclude testing KOKKOS with platforms that are not compatible **by default**, since we want to be certain of the correct behavior of compiling/installing/running LAMMPS with a **default** configuration of those environment. That will clear the CI status here and the update can be included. The consequences of that however is something that **you** will have going to live with. This kind of change makes it less likely to have KOKKOS included in prebuilt packages and will create additional confusion and frustration of people that are not software developers and would rather focus on doing science than having to deal with technical issues like having to compile/install **another** piece of software (i.e. people that are already confused and frustrated).

I can appreciate the difficulties of maintaining compatibility to older versions of tools or across platforms (LAMMPS is covering a much wider range than KOKKOS does) and perhaps you should see the need of requiring a newer version as a sign that you are trying to solve a problem with CMake scripting that perhaps needs a different solution. I have the same doubts about some of the things we do in LAMMPS as well.

Anyway, you have made your point clear and there is (obviously) not going to be a change and I have pointed out what this could mean for the LAMMPS user community (I may be wrong there, I don't have a crystal ball) and that should conclude this topic.

jrmadsen:
@akohlmey are you aware what "FWIW" stands for? Your response is needlessly hostile. I was simply offering an easy one line solution that should be very familiar to people that "just want to do science" bc in my experience these people overwhelming use Python. Please don't bother responding to this, your needless aggression had made me regret simply trying to help.

akohlmey:
@stanmoore1 FYI we have updated the integration tests so that this Kokkos update is not failing anymore.

akohlmey:
@stanmoore1 can you please update me on the status of this PR. If there are no remaining problems, I would prefer to merge it ASAP.

stanmoore1:
@akohlmey I'm seeing a small slowdown for small LJ on V100, looking into this now.

stanmoore1:
@akohlmey The performance regression is small (few percent) so let's merge this. Kokkos developers may be able to improve performance for the next release.

akohlmey (approved):

PR #2736, opened 2021-04-27
Fix issues exposed by running pair style unit tests with `pair_modify nofdotr`

akohlmey:
**Summary**

Pair styles frequently use `F dot r` to determine the (global) virial and thus compute pressure. However, for computing per-atom stress this is accumulated for each pair. In addition there is the command `pair_modify nofdotr` to enforce that code path, even when no per-atom stress is requested. Of course those two code paths are supposed to produce the same results within the level possible by floating-point math. This pull request attempts to fix or flag situations where there is currently a discrepancy.
Also a corresponding check was added to the unit test tool for pair styles.

**Related Issue(s)**

This was detected while working on PR #2708 

**Author(s)**

Aidan Thompson, SNL and Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

Some setups that would previously run and produce erroneous results may abort now.

**Implementation Notes**

The basic fix to correct the behavior is that tallying of the virial needs to be done for either `vflag_global` or `vflag_atom`, i.e. the test should be for `vflag_either` and corresponding the the `v_tally*()` or `ev_tally*()` functions need to compute the virial contribution for either case and then tally it to either the global or per-atom accumulators. For the OPENMP versions this requires changing the API so that the pointer to the "pair" class is passed along so that the various flags set by `Pair::ev_init()` are accessible.

Beyond that, some pair styles had typos, were missing a call to a matching tally function or had a sign error on the arguments to the tally function. For the MEAM package the tallying had to be done manually and thus API for `meam_force()` changed accordingly.

Pair styles affected (checked items have a fix in this PR):
- [x] buck/intel
- [x] buck/coul/long/intel
- [x] lj/charmm/coul/chamm/intel
- [x] lj/charmm/coul/long/intel
- [x] eam/intel (and variants)
- [x] edip/omp
- [x] meam
- [x] polymorphic
- [ ] reax/c          (will be corrected in PR #2708)
- [ ] reax/c/omp  (will be corrected in PR #2708)
- [x] airebo (and variants)
- [x] airebo/omp (and variants)
- [x] bop
- [x] comb
- [x] comb/omp
- [x] comb3
- [x] extep
- [x] gw (and variants)
- [x] lcbop
- [x] tersoff (and variants)
- [x] tersoff/omp (and variants)
- [x] tersoff/table/omp
- [x] rann

This also include several peripheral fixes:
- USER-INTEL pair styles did not use `nighbor->find_request(this)` to look up existing neighbor list requests.
- detect the new LLVM based Intel compilers alongside the old Intel compilers in CMake scripts
- an uninitialized data access fix in a unit test tool

**Post Submission Checklist**


sjplimp:
@akohlmey @athomps These are all INTEL package pair styles?  Should Mike Brown be a reviewer on this?

akohlmey:
> @akohlmey @athomps These are all INTEL package pair styles? Should Mike Brown be a reviewer on this?

Not "all", only the first few. But it is pointless to try and fix them in USER-INTEL when the core code doesn't address the problem consistently for all non-accelerated pair styles. Also, for USER-INTEL the fix is often to check and error out, since many of them do not tally per-atom stress at all (for performance reasons).

akohlmey:
@athomps this is now almost complete. The only pending issue are the INTEL package specific failures, which are likely due to a limitation caused by the optimizations therein. Those pair styles don't support per-atom stress, for example. So I may have to address those in the tester tool rather than in the pair styles themselves.

athomps (approved):
Great work.  I approve.

PR #2738, opened 2021-04-27
Make read_data compatible with files that have lines longer than 256 chars

### LINKED ISSUE #2735: Max line length in read_data
There appears to be a maximum line length permitted in a read_data file of 255 characters. The most recent master branch LAMMPS (8 Apr 2021) will produce the following misleading error message if the line length exceeds this limit: "ERROR: Unexpected empty line in BondCoeffs section (src/read_data.cpp:1829)" The most recent unstable branch LAMMPS (8 Apr 2021) produces a segmentation fault.

The error can be reproduced by adding characters to the dreiding example:
```
diff  examples/dreiding/data.dreiding examples/dreiding/data.dreiding_error
29c29
<     1             350            1.42 # O_3 C_3
---
>     1             350            1.42 # O_3 C_3 aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaeaaaaaaaaaaaaaaaaa
```

I dug around the manual and didn't see a reference to a maximum line length in a read_data file, perhaps it could be added to the manual. It could also be helpful to users if the error message indicated the cause was a line that was too long.

I'm currently converting a data file to LAMMPS format and trying to retain as much of the parameter provenance as  comments in the file, which is causing the lines to exceed the limit in some cases.

akohlmey:
**Summary**

The read data command currently has a compile time line length limit of 256 characters. While that is plenty, it can cause problems when lengthy comments are added. Those will lead to  parser errors, since only the first part of that long line is read and then reading continues. So the comment becomes the first part of the next line and that will lead to parser errors and too many lines per section.

This PR adds a new utility function `utils::fgets_trunc_nl()` which will read data up to the given buffer size and make certain it is NULL terminated and has a newline at the end. But it will keep reading long lines until the actual end of the line is reached.
That data is ignored. This function replaces `fgets()` in `utils::read_lines_from_file()` (which will become simpler since newline is guaranteed by the new function). There are possible other applications in text and potential reader classes.

**Related Issue(s)**

Closes #2735 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

All unit tests pass.

**Post Submission Checklist**

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -168,7 +168,7 @@ const char *utils::guesspath(char *buf, int len, FILE *fp)
 // read line into buffer. if line is too long keep reading until EOL or EOF
 // but return only the first part with a newline at the end.
 
-char *utils::fgets_trunc_nl(char *buf, int size, FILE *fp)
+char *utils::fgets_trunc(char *buf, int size, FILE *fp)
 {
   constexpr int MAXDUMMY = 256;
   char dummy[MAXDUMMY];
@@ -271,7 +271,7 @@ int utils::read_lines_from_file(FILE *fp, int nlines, int nmax,
   if (me == 0) {
     if (fp) {
       for (int i = 0; i < nlines; i++) {
-        ptr = fgets_trunc_nl(ptr,nmax,fp);
+        ptr = fgets_trunc(ptr,nmax,fp);
         if (!ptr) break; // EOF?
         // advance ptr to end of string
         ptr += strlen(ptr);
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -165,6 +165,42 @@ const char *utils::guesspath(char *buf, int len, FILE *fp)
   return buf;
 }
 
+// read line into buffer. if line is too long keep reading until EOL or EOF
+// but return only the first part with a newline at the end.
+
+char *utils::fgets_trunc_nl(char *buf, int size, FILE *fp)
+{
+  constexpr int MAXDUMMY = 256;
+  char dummy[MAXDUMMY];
+  char *ptr = fgets(buf,size,fp);
+
+  // EOF
+  if (!ptr) return nullptr;
+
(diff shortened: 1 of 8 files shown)


src/utils.cpp
```diff
@@ -165,6 +165,42 @@ const char *utils::guesspath(char *buf, int len, FILE *fp)
   return buf;
 }
 
+// read line into buffer. if line is too long keep reading until EOL or EOF
+// but return only the first part with a newline at the end.
+
+char *utils::fgets_trunc_nl(char *buf, int size, FILE *fp)
```
rbberger, line 171: I would just call this `fgets_trunc` without the `nl`. First of all it reads like `n1` and it doesn't add new information compared to the default behavior of `fgets`.

akohlmey: > I would just call this `fgets_trunc` without the `nl`. First of all it reads like `n1` and it doesn't add new information compared to the default behavior of `fgets`.

I agree on the n1/nl thing, but it **does** indicate an additional difference in behavior. `fgets()` will add the terminating NULL, but this function **also** guarantees a newline character at the end. (which is what was initially injected by `utils::read_lines_from_file()` into the block of lines that is then broadcast so it can be split and parsed concurrently on all MPI ranks)

sjplimp (approved):

src/utils.cpp
akohlmey: made the rename

rbberger (approved):

PR #2744, opened 2021-04-30
Update OpenCL ICD loader library to OpenCL specifications version 3.0.7

akohlmey:
**Summary**

This updates the OpenCL loader library used by LAMMPS with the latest updates from upstream.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Post Submission Checklist**


sjplimp (approved):

rbberger (approved):
Seems to be fine on the hardware I have. However, all of those support OpenCL 3.0

akohlmey:
> Seems to be fine on the hardware I have. However, all of those support OpenCL 3.0

No problem here with an OpenCL library providing OpenCL 1.2. As far as I understand, the loader defines the **maximum** OpenCL level possible and the kernels then will have to adapt to what is provided by the actual driver library. So there should be no benefit to artificially limiting this to OpenCL 2.1 like it previously was.

Issue #2740, opened 2021-04-27
[BUG] GPU package compiled with CUDA fails


charlessievers:
**Summary**

Compiling and trying to run with the GPU package on NVIDIA drivers v440.100 either with CUDA 10.2 fails.

**LAMMPS Version and Platform**

current master, ubuntu 18.04, Intel(R) Core(TM) i9-9920X CPU @ 3.50GHz, GeForce 2080 Ti

**Expected Behavior**

```
 $ cd examples/melt 
 $ ../../build/lmp -in in.melt -sf gpu 
```

This should run and complete without error

**Actual Behavior**

```
$ ../../build/lmp -in in.melt -sf gpu 
LAMMPS (8 Apr 2021)
OMP_NUM_THREADS environment is not set. Defaulting to 1 thread. (src/comm.cpp:94)
  using 1 OpenMP thread(s) per MPI task
Lattice spacing in x,y,z = 1.6795962 1.6795962 1.6795962
Created orthogonal box = (0.0000000 0.0000000 0.0000000) to (16.795962 16.795962 16.795962)
  1 by 1 by 1 MPI processor grid
Created 4000 atoms
  create_atoms CPU = 0.000 seconds

CITE-CITE-CITE-CITE-CITE-CITE-CITE-CITE-CITE-CITE-CITE-CITE-CITE

Your simulation uses code contributions which should be cited:
- GPU package (short-range, long-range and three-body potentials):
The log file lists these citations in BibTeX format.

CITE-CITE-CITE-CITE-CITE-CITE-CITE-CITE-CITE-CITE-CITE-CITE-CITE


--------------------------------------------------------------------------
- Using acceleration for lj/cut:
-  with 1 proc(s) per device.
-  Horizontal vector operations: ENABLED
-  Shared memory system: No
--------------------------------------------------------------------------
Device 0: GeForce RTX 2080 Ti, 68 CUs, 11/11 GB, 1.5 GHZ (Mixed Precision)
--------------------------------------------------------------------------

Initializing Device and compiling on process 0...Done.
Initializing Device 0 on core 0...Done.

Setting up Verlet run ...
  Unit style    : lj
  Current step  : 0
  Time step     : 0.005
Cuda driver error 1 in call at file '/xxxx/xxxxxxxx/lammps/lib/gpu/geryon/nvd_kernel.h' in line 379.
--------------------------------------------------------------------------
MPI_ABORT was invoked on rank 0 in communicator MPI_COMM_WORLD
[...]

PR #2745, opened 2021-04-30
Updates to the GPU package

rbberger:
**Summary**

* Fixes a bug in the GPU neighbor list kernels when compiling with `LAMMPS_BIGBIG` and OpenCL
* Reverts change that causes crash with older CUDA (10.2)
* Changes how OpenCL kernel headers are generated to detect changes and recreate them when necessary

**Related Issue(s)**

#2740 

**Author(s)**

@rbberger (Temple U)

**Licensing**

**Backward Compatibility**

CUDA 10.2  (can use new neighbor code)
CUDA 11.0 (can use new neighbor code)
CUDA 11.1 (can use new neighbor code)
CUDA 11.2 (uses old neighbor code)
CUDA 11.3 (uses old neighbor code)

**Implementation Notes**

`tagint` is a 64bit integer when LAMMPS_BIGBIG is set. The OpenCL equivalent is `long`.

**Post Submission Checklist**

https://www.khronos.org/registry/OpenCL/sdk/1.0/docs/man/xhtml/scalarDataTypes.html

### DIFF
--- a/lib/gpu/lal_neighbor_gpu.cu
+++ b/lib/gpu/lal_neighbor_gpu.cu
@@ -115,7 +115,7 @@ __kernel void kernel_calc_cell_counts(const unsigned *restrict cell_id,
 #define tagint int
 #endif
 #ifdef LAMMPS_BIGBIG
-#define tagint long long int
+#define tagint long
 #endif
 #ifdef LAMMPS_SMALLSMALL
 #define tagint int
--- a/cmake/Modules/GenerateOpenCLHeader.cmake
+++ b/cmake/Modules/GenerateOpenCLHeader.cmake
@@ -1,3 +1,3 @@
-# utility script to call GenerateOpenCLHeader function
+# utility script to call WriteOpenCLHeader function
 include(${SOURCE_DIR}/Modules/OpenCLUtils.cmake)
 WriteOpenCLHeader(${VARNAME} ${HEADER_FILE} ${SOURCE_FILES})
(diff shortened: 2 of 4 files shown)


ndtrung81 (approved):
@rbberger thanks for this PR!

rbberger:
@junghans I changed how the OpenCL kernels headers are generated in CMake. Touching any of the OpenCL kernel source files will now trigger the header to be rebuild. Previously you had to run CMake again. Can you please take a look over it?

junghans (approved):

lib/gpu/lal_neighbor_gpu.cu
```diff
@@ -115,7 +115,7 @@ __kernel void kernel_calc_cell_counts(const unsigned *restrict cell_id,
 #define tagint int
 #endif
 #ifdef LAMMPS_BIGBIG
-#define tagint long long int
+#define tagint long
```
akohlmey, line 118: Does CUDA/OpenCL/HIP support `int64_t`?

That may be a safer bet in the long run (pun intended). We are currently not building on Windows with BIGBIG and I am not aware of any other platform that have 32-bit long ints, so this is likely ok (for now).

rbberger: CUDA and HIP use int64_t, this line is the OpenCL code.

rbberger: Unlike CUDA and HIP, the OpenCL code is compiled at runtime by the driver and should follow the spec. The spec says `long` it's 64bit.

akohlmey: excellent. that resolves this then.

rbberger: @akohlmey just for future reference: if the kernels would be compiled to OpenCL C++ instead of OpenCL C, one could include `<opencl_def>`, which defines `int64_t`. See https://www.khronos.org/registry/OpenCL/specs/2.2/html/OpenCL_Cxx.html. Trying to use `int64_t` in OpenCL C will result in a compilation error.
```
<kernel>:784:34: error: unknown type name 'int64_t
```

PR #2749, opened 2021-05-03
Make warning and info messages vararg compatible and implement "thermo_modify warn"

akohlmey:
**Summary**

This pull request implements a feature to control the amount of warning messages printed. Warnings are counted and when their total exceeds a given limit a final warning is printed and then all future warnings are suppressed. With the command "thermo_modify warn" the behavior of this feature can be controlled. It is possible to tell LAMMPS to ignore all warnings or print all warnings or change the maximum number of warnings or reset the warning counter or reset everything to default.

With this change some internal APIs are changed that allow using a format string and a variable number of arguments and thus apply the same change that was introduced in PR #2719 also to the `Error::message()` and `Error::warning()` functions.

In the process of updating the parts of the code using the `Error::warning()` function, also the replicated warning message code for dihedral and improper problems was consolidated into a convenience function `problem()` in the respective base classes.

Also multi-line warning messages were reformatted so that each line starts with "WARNING:" and thus making removing them from log or output with grep easier.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No functional changes, but the specific output of logfiles/outputs can change depending on which and how many warnings are printed during a run.

**Post Submission Checklist**

### DIFF
--- a/doc/src/thermo_modify.rst
+++ b/doc/src/thermo_modify.rst
@@ -76,22 +76,33 @@ are drifting out of the box through a fixed boundary condition (see
 the :doc:`boundary <boundary>` command).  In this case one atom may be
 deleted before the rest of the molecule is, on a later timestep.
 
-The *warn* keyword determines whether LAMMPS will print warning
-messages and how many of them.  Certain warning messages can be
-quite verbose and thus quickly blow up the size of the log file
-and screen output.  Thus a limit of 100 warnings messages is applied
-by default.  If there are more warnings, LAMMPS will print one final
-warning that it will not print any additional warning messages.
+The *warn* keyword allows you to control whether LAMMPS will print
+warning messages and how many of them.  Most warning messages are only
+printed by MPI rank 0.  They are usually pointing out important issues
+that should be investigated, but LAMMPS cannot determine for
+certain whether they are an indication of an error.
+
+Some warning messages are printed during a run (or immediately before)
+each time a specific MPI rank encounters the issue, e.g. bonds that are
+stretched too far or dihedrals in extreme configurations. These can
+become verbose for and thus quickly blow up the size of the log file
+and screen output.  Thus a limit of 100 warning messages is applied
+by default.  The warning count is applied to the entire input unless
+reset with a ``thermo_modify warn reset`` command.  If there are more
+warnings than the limit, LAMMPS will print one final warning that it
+will not print any additional warning messages.
+
 Any number after the keyword *warn* will change the warning limit
 accordingly.  With the value *ignore* all warnings will be suppressed,
 with the value *always* no limit will be applied and warnings will
-always be printed, with the value *reset* the internal warning
-counter will be reset to zero,  and with the value *default*, all
-settings will be reset to their defaults and the counter reset.
-An example usage of this keyword would be that during equilibration
-warnings would be ignored while the system is still adjusting, but
-then one would reset everything to the defaults for the production
-run.
+always be printed, with the value *reset* the internal warning counter
+will be reset to zero, and with the value *default*, the counter is
+reset and the limit set to 100.  An example usage of either *reset* or
+*default* would be to reenable warnings that were disabled or have
+reached the limit during equilibration, where the warnings would be
+accepteable while the system is still adjusting, but then one would want
+to all warnings again for the production run, where they would indicate
+problems that would require a closer look at what is causing them.
 
 The *norm* keyword determines whether various thermodynamic output
 values are normalized by the number of atoms or not, depending on
--- a/doc/src/thermo_modify.rst
+++ b/doc/src/thermo_modify.rst
@@ -201,9 +201,9 @@ Related commands
 Default
 """""""
 
-The option defaults are lost = error, norm = yes for unit style of
-*lj*\ , norm = no for unit style of *real* and *metal*\ , flush = no,
-and temp/press = compute IDs defined by thermo_style.
+The option defaults are lost = error, warn = 100, norm = yes for unit
+style of *lj*\ , norm = no for unit style of *real* and *metal*\ ,
+flush = no, and temp/press = compute IDs defined by thermo_style.
 
 The defaults for the line and format options depend on the thermo
 style.  For styles "one" and "custom", the line and format defaults
--- a/doc/src/thermo_modify.rst
+++ b/doc/src/thermo_modify.rst
@@ -98,9 +98,9 @@ with the value *always* no limit will be applied and warnings will
 always be printed, with the value *reset* the internal warning counter
 will be reset to zero, and with the value *default*, the counter is
 reset and the limit set to 100.  An example usage of either *reset* or
-*default* would be to reenable warnings that were disabled or have
+*default* would be to re-enable warnings that were disabled or have
 reached the limit during equilibration, where the warnings would be
-accepteable while the system is still adjusting, but then one would want
+acceptable while the system is still adjusting, but then one would want
 to all warnings again for the production run, where they would indicate
 problems that would require a closer look at what is causing them.
 
--- a/doc/src/thermo_modify.rst
+++ b/doc/src/thermo_modify.rst
@@ -11,12 +11,13 @@ Syntax
    thermo_modify keyword value ...
 
 * one or more keyword/value pairs may be listed
-* keyword = *lost* or *lost/bond* or *norm* or *flush* or *line* or *format* or *temp* or *press*
+* keyword = *lost* or *lost/bond* or *warn* or *norm* or *flush* or *line* or *format* or *temp* or *press*
 
   .. parsed-literal::
 
        *lost* value = *error* or *warn* or *ignore*
        *lost/bond* value = *error* or *warn* or *ignore*
+       *warn* value = *ignore* or *reset* or *default* or a number
        *norm* value = *yes* or *no*
        *flush* value = *yes* or *no*
        *line* value = *one* or *multi*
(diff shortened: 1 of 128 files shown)


akohlmey:
An example log output:
```
[...]
WARNING: Dihedral problem: 0 0 214 211 209 219
WARNING:   1st atom: 0 -3.8798100 -1.7553400 -35.133200
WARNING:   2nd atom: 0 -4.1667100 -2.9647800 -34.293300
WARNING:   3rd atom: 0 -5.7400300 -3.2600200 -34.419400
WARNING:   4th atom: 0 -6.1769000 -3.7682500 -35.730500 (src/MOLECULE/dihedral_charmm.cpp:147)
WARNING: Dihedral problem: 0 0 211 214 216 217
WARNING:   1st atom: 0 -4.1667100 -2.9647800 -34.293300
WARNING:   2nd atom: 0 -3.8798100 -1.7553400 -35.133200
WARNING:   3rd atom: 0 -3.5808000 -2.0478600 -36.456800
WARNING:   4th atom: 0 -3.1201100 -1.3423700 -36.995400 (src/MOLECULE/dihedral_charmm.cpp:147)
Per MPI rank memory allocation (min/avg/max) = 140.0 | 140.0 | 140.0 Mbytes
WARNING: Too many warnings: 56829 vs 100. All future warnings will be suppressed (src/thermo.cpp:422)
---------------- Step        0 ----- CPU =      0.0000 (sec) ----------------
TotEng   =    -25356.2064 KinEng   =     21444.8313 Temp     =       299.0397 
PotEng   =    -46801.0377 E_bond   =      2537.9940 E_angle  =     10921.3742 
[...]
```

doc/src/thermo_modify.rst
```diff
@@ -75,6 +76,23 @@ are drifting out of the box through a fixed boundary condition (see
 the :doc:`boundary <boundary>` command).  In this case one atom may be
 deleted before the rest of the molecule is, on a later timestep.
 
+The *warn* keyword determines whether LAMMPS will print warning
+messages and how many of them.  Certain warning messages can be
```
sjplimp, line 80: I think this should say something like this:
Most warning messages are only printed by processor 0.  These are often important to read.  Some warning messages are printed during a run (or immediately before) by every processor that encouters the issue, e.g. bonds that are stretched too far or dihedrals that in extreme configurations.  These can blow up ...

doc/src/thermo_modify.rst
```diff
@@ -75,6 +76,23 @@ are drifting out of the box through a fixed boundary condition (see
 the :doc:`boundary <boundary>` command).  In this case one atom may be
 deleted before the rest of the molecule is, on a later timestep.
 
+The *warn* keyword determines whether LAMMPS will print warning
```
sjplimp, line 79: The warn keyword allows you to control whether LAMMPS ...

doc/src/thermo_modify.rst
```diff
@@ -75,6 +76,23 @@ are drifting out of the box through a fixed boundary condition (see
...
+messages and how many of them.  Certain warning messages can be
+quite verbose and thus quickly blow up the size of the log file
+and screen output.  Thus a limit of 100 warnings messages is applied
```
sjplimp, line 82: warnings -> warning.  Also explain whether N is cumulative over the lifetime of the input script or something else.

doc/src/thermo_modify.rst
```diff
@@ -75,6 +76,23 @@ are drifting out of the box through a fixed boundary condition (see
...
+by default.  If there are more warnings, LAMMPS will print one final
+warning that it will not print any additional warning messages.
+Any number after the keyword *warn* will change the warning limit
+accordingly.  With the value *ignore* all warnings will be suppressed,
+with the value *always* no limit will be applied and warnings will
+always be printed, with the value *reset* the internal warning
+counter will be reset to zero,  and with the value *default*, all
+settings will be reset to their defaults and the counter reset.
```
sjplimp, line 90: with the value default, the counter is reset and the limit is set to 100.

src/MOLECULE/bond_fene.cpp
```diff
@@ -95,7 +89,7 @@ void BondFENE::compute(int eflag, int vflag)
 
     // force from LJ term
 
-    if (rsq < TWO_1_3*sigma[type]*sigma[type]) {
+    if (rsq < MY_CBRT2*sigma[type]*sigma[type]) {
```
sjplimp, line 92: Please change this to MY_CUBEROOT2 - it went from being obvious (at least to me) to no idea what it is

src/thermo.cpp
```diff
@@ -400,42 +400,56 @@ void Thermo::call_vfunc(int ifield_in)
...
+  // ntotal = current # of atoms, and Error class warnings
 
-  bigint ntotal;
-  bigint nblocal = atom->nlocal;
-  MPI_Allreduce(&nblocal,&ntotal,1,MPI_LMP_BIGINT,MPI_SUM,world);
-  if (ntotal < 0)
+  bigint nlocal[2], ntotal[2] = {0,0};
+  nlocal[0] = atom->nlocal;
+  nlocal[1] = error->get_numwarn();
+  MPI_Allreduce(nlocal,ntotal,2,MPI_LMP_BIGINT,MPI_SUM,world);
```
sjplimp, line 413: Does this mean the count of warnings issued is only updated every thermo output step?  If so, that is probably fine, but it is not what the thermo_mod warn doc page says, right?  If so, I think the doc page should be more clear.  Since for thermo 10000 you could get a zillion WARNINGS before the next thermo output, is this a conceptual problem?

sjplimp (changes requested):

src/thermo.cpp
akohlmey: No. There are two checks. Each MPI rank counts the warnings (numwarn) and it is periodically (when LAMMPS also checks for lost atoms) summed into a total sum (allwarn). The printing of warning messages stops when either counter reaches the limit.

akohlmey:
@sjplimp documentation is updated. please check again.

src/thermo.cpp
sjplimp: Do you mean each proc stops when its local count reaches (e.g.) 100?  In which case (if thermo output is rare) and you are running on 1000 MPI tasks, you will get 100,000 warnings to the screen and logfile?  If so, that's not what the thermo_modify warn doc page would make you think.

akohlmey: yes and no. Each MPI rank will stop when either the local count is over the limit **or** the global count. and the global count is updated whenever the lost atom check is done. so, you could indeed construct a case where there are warnings on *all* MPI ranks and no thermo updates where there will be more than the given limit. but that is a rare exception. a more likely  case is that there are either warnings only on a few MPI ranks or regular updates of the global count so the number of warnings happening will be very close to the given limit. and at any rate, you won't be worse off than you currently are with no option at all to control the number. in the pathological scenario, you have now at least the option to (temporarily) turn off warnings altogether.

I wanted to avoid adding an additional and more frequent call to MPI_Allreduce() at all cost.

sjplimp: I agree no change to the code is fine, due to the MPI_Allreduce.  But I think the doc page should add a sentence to say what the limit actually means.  Currently it sounds like 100 is the most warnings you will see.

doc/src/thermo_modify.rst
```diff
@@ -75,6 +76,34 @@ are drifting out of the box through a fixed boundary condition (see
...
+
+Any number after the keyword *warn* will change the warning limit
+accordingly.  With the value *ignore* all warnings will be suppressed,
+with the value *always* no limit will be applied and warnings will
+always be printed, with the value *reset* the internal warning counter
+will be reset to zero, and with the value *default*, the counter is
+reset and the limit set to 100.  An example usage of either *reset* or
+*default* would be to re-enable warnings that were disabled or have
+reached the limit during equilibration, where the warnings would be
+acceptable while the system is still adjusting, but then one would want
```
sjplimp, line 103: but then change to all warnings for the production ...

src/thermo.cpp
akohlmey: @sjplimp I am open to suggestions for improving the docs. But as you can see from the discussion here, it is quite complicated to describe all details very specifically and in most cases it won't make a (big) difference. My thinking was that it will be more effective to not be 100% precise with the description (and then have to explain if somebody has one of those exceptional cases) so that it doesn't confuse people too much that have "normal" situations.

doc/src/thermo_modify.rst
```diff
@@ -75,6 +76,34 @@ are drifting out of the box through a fixed boundary condition (see
...
 
+The *warn* keyword allows you to control whether LAMMPS will print
+warning messages and how many of them.  Most warning messages are only
+printed by MPI rank 0.  They are usually pointing out important issues
+that should be investigated, but LAMMPS cannot determine for
+certain whether they are an indication of an error.
+
+Some warning messages are printed during a run (or immediately before)
+each time a specific MPI rank encounters the issue, e.g. bonds that are
+stretched too far or dihedrals in extreme configurations. These can
```
sjplimp, line 87: The number of these can quickly blow up ...

doc/src/thermo_modify.rst
```diff
@@ -75,6 +76,34 @@ are drifting out of the box through a fixed boundary condition (see
...
+become verbose for and thus quickly blow up the size of the log file
+and screen output.  Thus a limit of 100 warning messages is applied
+by default.  The warning count is applied to the entire input unless
+reset with a ``thermo_modify warn reset`` command.  If there are more
+warnings than the limit, LAMMPS will print one final warning that it
+will not print any additional warning messages.
+
```
sjplimp, line 94: NOTE (in rst lingo) The limit is enforced on either a per-processor basis or summed count across processors which is only updated at thermodyanmic output steps.   Thus if running on 1000s of processors with rare thermodynamic output, many more warnings than the limit could potentially be output

sjplimp:
@akohlmey  made my suggested changes

doc/src/thermo_modify.rst
akohlmey: Done. I like this. This keeps the complex explanation optically separate from the simplified description.

rbberger (approved):

sjplimp (approved):

PR #2750, opened 2021-05-04
Updated pair style bop

akohlmey:
**Summary**

This pull request contains and updated version of pair style bop with performance improvements and bug fixes.
Also included are corrected potential files and corresponding changes to the potential file parser.

**Author(s)**

Xiaowang Zhou, SNL

**Licensing**

**Backward Compatibility**

The potential file reader will fail on incorrectly formatted older potentials. So the updated versions from this PR must be used.

**Implementation Notes**

The "TabularFunction" class was unbundled from the pair style and made available as a general use class.

**Post Submission Checklist**


sjplimp (approved):

ilia-nikiforov-umn:
Hi Axel,

These changes introduced a requirement that the minimum size of the simulation box is six times bigger than the largest cutoff in pair_style bop. Is this intentional and unavoidable? It seems like a very stringent requirement. We discovered this because none of the Verification Checks in OpenKIM run on pair_style bop anymore since this change.

Cheers, ilia

akohlmey:
> These changes introduced a requirement that the minimum size of the simulation box is six times bigger than the largest cutoff in pair_style bop. Is this intentional and unavoidable?

Yes. These were required by the developer of the pair style (I only adapted and modernized the code to better align with refactoring changes in the core of LAMMPS).

sjplimp:
I agree that seems like an odd requirement.  I'll ask Xiaowang about it.

sjplimp:
@ilia-nikiforov-umn We removed that error check on the box size in pair_bop.cpp.  in another check we did find that BOP gives a different answer (energy, forces, virial) for a tiny periodic system (1,2,or 4 atoms) versus a 2x2x2 replicated version of the same system.  This is likely a bug, so we are looking into it.  Do your KIM checks flag something like that?

ilia-nikiforov-umn:
@sjplimp Thanks for looking into this! We do indeed check for that (energy and forces only) in our [PeriodicitySupport Verification Check](https://openkim.org/id/PeriodicitySupport__VC_895061507745_003). Right now the grade is N/A for BOP potentials because they fail to run because of the check.

PR #2756, opened 2021-05-07
Collected small changes and bugfixes for the next patch release

### LINKED ISSUE #2754: [BUG] Unexpected pair coeff mix behavior with hybrid/overlay
**Summary**

I'm interested in combining `lj/mdf` and `coul/long` using `hybrid/overlay` since a `lj/mdf/coul/long` pair style does not exist. To demonstrate the unexpected behavior of hybrid/overlay the example uses `lj/cut` and `coul/cut` for simplicity. 

My expectation from reading the manual was that these inputs would all behave similarly:
```
pair_style lj/cut 9
pair_style lj/cut/coul/cut 9 9
pair_style hybrid/overlay lj/cut 9
pair_style hybrid/overlay lj/cut 9 coul/cut 9
```

Any of the first three lines will run as expected, but if the fourth is used it will produce the error:
> ERROR on proc 0: All pair coeffs are not set (src/pair_hybrid.cpp:659)

It's surprising to me, as a user, that LAMMPS can produce the mixed coeffs using standard mixing rules like geometric or arithmetic for `hybrid/overlay lj/cut` but cannot if `coul/cut` is added.

I'm not sure if this is a bug or intended behavior but it is surprising. I understand hybrid mixing can become complicated quickly but I was hoping that for this simple case there would be some option to allow standard mixing rules be applied to lj terms.

The current implementation also easily leads to incorrect/unintended results by missing mixed lj terms. Because coul/cut has no parameters if a user does: `pair_coeff * * coul/cut` the script will run with no errors but will produce incorrect lj values because mix terms will only be assigned coul/cut interactions. It is quite surprising that `hybrid/overlay lj/cut 9` and` hybrid/overlay lj/cut 9 coul/cut 9` can produce different van der Waals energies.

LAMMPS (8 Apr 2021)

**Expected Behavior**

See above

**Actual Behavior**

See above

**Steps to Reproduce**

A simple script and datafile is attached, based off the the tiny_epoxy example in USER/reaction/tiny_epoxy, to reproduce the unexpected behavior.

[in.script.hybrid-overlay.txt](https://github.com/lammps/lammps/files/6435054/in.script.hybrid-overlay.txt)
[in.script.txt](https://github.com/lammps/lammps/files/6435056/in.script.txt)
[tiny_epoxy.data.txt](https://github.com/lammps/lammps/files/6435057/tiny_epoxy.data.txt)

### LINKED ISSUE #2755: [BUG] Infinite loop with failed insertion from fix deposit
**Summary**

When using a bonded molecular template, LAMMPS can hang indefinitely if a deposition fails. This bug can be easily reproduced by increasing the deposition rate in the example script in.deposit.molecule and running on 2 processors.

The bug manifests when a new bond topology list is built. The NTopoBond classes calls the `domain->closest_image()` method. This method searches the `atom->sametag` array. When a deposition fails, this array can sometimes contain a closed loop (e.g. index 235 points to 245 points to 255 points to 235) which LAMMPS cycles through indefinitely.

In fix deposit,  nghost is set to zero before adding atoms and clearing the map. When `atom->map_set()` is later called, it only clears data up to nlocal+nghost, leaving junk in the atom array maps which can create these loops (might not happen for hash maps). This doesn't matter if the deposition is successful since fix deposit explicitly calls `atom->map_init()` which also clears data up to `atom->map_tag_max`.

This bug can be easily fixed by clearing the map in fix deposit before setting nghost to zero. I think adding the line
`if (atom->map_style != Atom::MAP_NONE) atom->map_clear();`
should fix the bug (I can no longer reproduce it). Alternatively, the lines:
`    
   if (atom->map_style != Atom::MAP_NONE) { 
      atom->map_init();  
      atom->map_set();  
    }
`
could be moved out of the `if (success)` block so the map is always reinitialized.

As an aside, would it be appropriate to add a small error check to the `domain->closest_image()` method to avoid loops in the `atom->sametag` array? One could record the initial atom index and error if the index is seen again. Ideally LAMMPS should never create these loops to begin with, but it might help debugging potential errors in the future?

If either of these solutions sound good, I'm happy to submit a PR with the changes (unless it's easier to add it to one of the collected small fixes PRs). 

Thanks!

Current LAMMPS master

**Expected Behavior**

LAMMPS should not hang after failed depositions with bonds.

**Actual Behavior**

LAMMPS can hang after failed depositions with bonds.

**Steps to Reproduce**

Change line 37 in the molecule deposit example to:
`fix		4 addatoms deposit 100 0 1 12345 region slab near 1.0 &`
and run on 2 processors, it should hang somewhere after 600 timesteps.

NA

akohlmey:
**Summary**

This pull request  combines multiple small fixes and changes that don't warrant a pull request of their own.

**Author(s)**

Axel Kohlmeyer, Temple U
Richard Berger, Temple U
Nguyen Trung, Northwestern U

**Licensing**

**Backward Compatibility**

No known issues

**Implementation Notes**

- add a clarification to pair hybrid/overlay mixing rules. Closes #2754 
- use system provided instead of self compiled OpenCL loader library by default on macOS
- fix copy-n-paste bug in fix ttm/mod reported on the MatSci forum
- update Swig interface file (lammps.i) for recent additions to the LAMMPS library interface
- use simpler shape of 1d numpy arrays when extracting neighbor lists via Python interface
- bugfix in eam kernels leading to crashes with GPU_PREC=double
- fixes possible deadlocks in fix deposit and fix pour as reported by @jtclemm. Closes #2755 

**Post Submission Checklist**


sjplimp (approved):

src/MISC/fix_deposit.cpp
```diff
@@ -621,10 +615,13 @@ void FixDeposit::pre_exchange()
...
-    if (atom->map_style != Atom::MAP_NONE) {
-      atom->map_init();
-      atom->map_set();
-    }
+  }
+
+  // rebuild atom map
+
+  if (atom->map_style != Atom::MAP_NONE) {
+    atom->map_init();
```
sjplimp, line 623: if (success) atom->map_init();

akohlmey: done

PR #2762, opened 2021-05-12
Prepare LAMMPS to support updating the coding style with clang-format

akohlmey:
**Summary**

This pull request includes changes required to use clang-format on the LAMMPS sources in the 'src' tree.

**Related Issue(s)**

#2643 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

Yes. No functionality should be changed.

**Implementation Notes**

- This assumes clang-format version 8 or later is used.
- A suitable `.clang-format` file is provided in the "src" folder
- The automatic format is currently only applied to `.h` files and a few selected .cpp files. The remaining .cpp files have a `// clang-format off` as first line added.
- XXXStyle() macros in the header files must be protected by `// clang-format off` / `// clang-format on` and must have a trailing semicolon (;)
- a few code changes to enable the header formatting are included as well

**Post Submission Checklist**


sjplimp (approved):

ellio167 (commented):
This looks quite good so far.

Some thoughts/questions:
* What is the reason to not format `.cpp` files at this time?
* Maybe `examples/plugins/.clang-format` could be a symlink, so as to avoid the need to remember to update it when `src/.clang-format` is changed?
* What about the `unittest` directory?
* How is "automatic formatting" to be accomplished? Should there be a mechanism for easy setup of pre-commit hooks (etc.) to enforce formatting? [like this](https://github.com/openkim/kim-api/blob/cd2fe213feb5dacfb84167c1b328f37b5914f739/scripts/configure-git-repo-hooks-etc)
* Probably some updates to `.github/CONTRIBUTING.md` should be included?

akohlmey:
> This looks quite good so far.
> 
> Some thoughts/questions:
> 
> * What is the reason to not format `.cpp` files at this time?

Because many sections of those files were not written with clang-format in mind and running the automatic format makes a lot of them very difficult to understand compared to the current state. Also there is likely some work needed to protect cases (e.g. certain use of macros) from clang-format messing them up so much that it breaks functionality (example: the XXXXStyle() macros in the header files). The .cpp file are not as uniform as the .h files (and even those required a substantial amount of work to just get them to compile and to pass the unit tests).

> * Maybe `examples/plugins/.clang-format` could be a symlink, so as to avoid the need to remember to update it when `src/.clang-format` is changed?

At this moment it is not clear whether they will remain the same.

> * What about the `unittest` directory?

There is a clang-format configuration in place already (and has been for a while) with different settings (because the nature of the code is different; e.g. there are not many deeply nested loops with if statments etc.

> * How is "automatic formatting" to be accomplished? Should there be a mechanism for easy setup of pre-commit hooks (etc.) to enforce formatting? [like this](https://github.com/openkim/kim-api/blob/cd2fe213feb5dacfb84167c1b328f37b5914f739/scripts/configure-git-repo-hooks-etc)

Currently not at all and it will likely never be. The purpose is to make it easy to produce code that follows a style defined in the config file.  Not to force people to use clang-format. So the most likely scenario is that during the preparation of pull requests, either the submitter or the person handling the PR will apply clang-format and then either adjustments are made or the change will be undone and other ways to approximate the formatting requirements may be used and the file either wholly or partially excluded from clang-format processing. I expect this to be a gradual process, as people get more familiar with using clang-format, they may consider its effect already when writing code, so that code will comply automatically.

> * Probably some updates to `.github/CONTRIBUTING.md` should be included?

This file is not very visible. Please see #2643 . the expectation is that we will have one canonical version of this document somewhere (as a separate URL or inside the manual) and then have all other suitable places point to it.

ellio167:
All very reasonable responses.

ellio167 (approved):

PR #2765, opened 2021-05-13
Fix compiling the SNAP module with Kokkos and SYCL

masterleinad:
**Summary**

SYCL requires a different namespace for math functions invoked on the device. Since `sincos` is non-standard we don't provide it in `Kokkos` yet. This is needed for compiling the `SNAP` module with `Kokkos` and `SYCL`.

**Author(s)**

Daniel Arndt, ORNL.

**Licensing**

**Backward Compatibility**

The changes should be backward-compatible. They only affect compiling with an `SYCL` compiler.

**Post Submission Checklist**

### DIFF
--- a/src/KOKKOS/sna_kokkos.h
+++ b/src/KOKKOS/sna_kokkos.h
@@ -23,6 +23,10 @@
 #include <Kokkos_Core.hpp>
 #include "kokkos_type.h"
 
+#ifdef __SYCL_DEVICE_ONLY__
+#include <CL/sycl.hpp>
+#endif
+
 namespace LAMMPS_NS {
 
 template<typename real_type_, int vector_length_>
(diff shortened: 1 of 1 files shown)


weinbe2:
@masterleinad to be honest, it might be better to just replace the `sincos_wrapper` function with back-to-back `sin` and `cos` calls in the three places it appears in the code. I'm the one who stuck it in in the first place, and I'll admit I never systematically investigated if it was worth it---I just did a quick test in CUDA, and it seems to make no perf difference (and if it does it's on the fifth digit, well below the noise).

weinbe2:
If you're seeing it makes a difference in SYCL, though, leave it as is :)

sjplimp (approved):
this is really up to Stan

PR #2775, opened 2021-05-19
Print warning when a tip4p pair style may cause incorrect results

akohlmey:
**Summary**

When using a tip4p pair style the point "M" that carries the charge is only "visible" inside the pair style and that charge is assigned to the corresponding oxygen atom otherwise, which can lead on incorrect results when charges are accessed, e.g. in hybrid pair styles or in computations (e.g. compute dipole/chunk).

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

Only warnings are added, no change in functionality.

**Post Submission Checklist**

### DIFF
--- a/doc/src/pair_lj_cut_tip4p.rst
+++ b/doc/src/pair_lj_cut_tip4p.rst
@@ -148,7 +148,7 @@ the pair_style command.
 .. warning::
 
    Because of how these pair styles implement the coulomb interactions
-   by implicitly defining a fourth coordinate for the negative charge
+   by implicitly defining a fourth site for the negative charge
    of the TIP4P and similar water models, special care must be taken
    when using these pair styles with other computations that also use
    charges.  Unless they are specially set up to also handle the implicit
--- a/doc/src/pair_lj_cut_tip4p.rst
+++ b/doc/src/pair_lj_cut_tip4p.rst
@@ -145,6 +145,22 @@ specified since a Coulombic cutoff cannot be specified for an individual I,J
 type pair. All type pairs use the same global Coulombic cutoff specified in
 the pair_style command.
 
+.. warning::
+
+   Because of how these pair styles implement the coulomb interactions
+   by implicitly defining a fourth coordinate for the negative charge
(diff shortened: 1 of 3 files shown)


akohlmey:
@sjplimp @athomps any suggestions where else this kind of warning should be added as well? in the code or the docs?
fix qeq? dump?

athomps:
I think anyone who is using LAMMPS to construct a TIP4P model will not benefit from this warning. Either they know what they are doing or they do not.  Hence there is no value adding it to `fix qeq` etc. Conversely, people who are running a pre-existing correct TIP4P model do need to be warned that LAMMPS reads/writes the charge with a position that is incorrect. I think it is sufficient to put the warning in the `pair_lj_cut_tip4p` pair style.  It is not necessary to put it everywhere that charge and positions are mentioned.

akohlmey:
@athomps I think the warning in the pair style hybrid code is required, since I have now repeatedly come across people thinking that they **have** to use a hybrid of 'coul/long' and 'tip4p/long' when they have a system which has not only TIP4P water. It was even noticed in an example shipped with LAMMPS(!) (for a "user" package).  Also the situation with computing dipoles has come up before. We have to recognize that as our user base is growing we can no longer assume that they know what they are doing to the degree it was possible 10 years ago.  What is worse is that people more an more use external tools to generate the inputs and thus do not read the manual as much as they did before (since the tool does the dirty work).

doc/src/pair_lj_cut_tip4p.rst
```diff
@@ -145,6 +145,22 @@ specified since a Coulombic cutoff cannot be specified for an individual I,J
 type pair. All type pairs use the same global Coulombic cutoff specified in
 the pair_style command.
 
+.. warning::
+
+   Because of how these pair styles implement the coulomb interactions
+   by implicitly defining a fourth coordinate for the negative charge
```
athomps, line 151: change `coordinate` to `site`?

athomps (approved):
I approve.  I suggested one word change in TIP4P doc page

PR #2784, opened 2021-05-27
Internally set PYTHONHOME for LAMMPS on Windows

akohlmey:
**Summary**

When packaging LAMMPS with the PYTHON package included on Windows we need to bundle a matching Python runtime.
This in turn needs the PYTHONHOME environment variable set to reliably find it. However that would be causing all kinds of problems in case another, standalone python installation is present. So will look for a LAMMPSHOME environment variable instead and set PYTHONHOME only internally from inside LAMMPS when creating the initial LAMMPS instance, i.e. before the Python runtime is loaded. This is only enabled on Windows and when the PYTHON package is installed.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Post Submission Checklist**


sjplimp (approved):

stanmoore1:
@akohlmey Seems like the GPL commits should be in a separate pull request since they have nothing to do with `Internally set PYTHONHOME for LAMMPS on Windows`? I understand the convenience of tacking on some commits to an existing PR, but it is nice to have a paper trail if we ever need to search for changes...

akohlmey:
> @akohlmey Seems like the GPL commits should be in a separate pull request since they have nothing to do with `Internally set PYTHONHOME for LAMMPS on Windows`? I understand the convenience of tacking on some commits to an existing PR, but it is nice to have a paper trail if we ever need to search for changes...

This was "a shot from the hip" to get all changes in quickly and in sync with what Richard is doing on www.lammps.org. That way I can tag the patch release and we have the final confirmation that everything is working as expected.
Under normal circumstances, this PR would have been the usual "mixed bag of changes" PR with multiple individually listed items. Just didn't expect it was needed when I submitted this. :-(

PR #2786, opened 2021-06-01
Collected small changes

### LINKED ISSUE #2791: Documentation error for Kolmogorov-Crespi (KC) potential
**Summary**

I believe the documentation for the Kolmogorov-Crespi potential is incorrect on the LAMMPS documentation [website](https://docs.lammps.org/pair_kolmogorov_crespi_full.html#kolmogorov1).

(I'm not sure of the best place to bring up documentation issues, so feel free to move or remove this issue.)

Not applicable, documentation issue only.

**Details**

Here is the specification of the potential from the paper:

paper

And here is the potential in the documentation (note the difference in parentheses/brackets in V):

docs

The documentation only describes an existing implementation of the potential, so this issue is unlikely to affect any calculations, but based on my understanding of the potential I believe the first image from the paper gives the correct specification of the potential.

akohlmey:
**Summary**

Various small changes.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues

**Implementation Notes**

- fix bug in fix wall/* when accessing the fix energy vector (reported on [MatSci]( https://matsci.org/t/force-on-wall-lj93-does-not-change-with-time/36793))
- smarter detection of whether stdout is connected to a terminal
- properly handle style dependencies for QEQ package
- do not test molecule templates with bonds without MOLECULE package installed
- silence compiler warnings
- silence some LGTM warnings
- check more code on LGTM.com
- fix portability issue when compiling N2P2 library on FreeBSD (needs CompPhysVienna/n2p2#116)
- fix typos in Kolmolgorov-Crespi potential formula, Closes #2791 
- update fedora singularity definition to Fedora 34 and recent changes in building windows packages
- work around a compilation issue for Plumed2 with newer GCC and Clang compilers (fix is included upstream)
- documentation updates for recent changes to the windows binary packages
- silence compiler warnings and fix communication bug in USER-MDI

**Post Submission Checklist**


src/USER-MDI/fix_mdi_engine.cpp
```diff
@@ -228,9 +228,11 @@ int FixMDIEngine::execute_command(const char *command, MDI_Comm mdicomm)
   // respond to any driver command
 
   if (strcmp(command, ">NATOMS") == 0) {
-    ierr = MDI_Recv((char *) &atom->natoms, 1, MDI_INT, mdicomm);
+    int mdi_natoms = 0;
+    ierr = MDI_Recv((char *) &mdi_natoms, 1, MDI_INT, mdicomm);
     if (ierr != 0) error->all(FLERR, "MDI: Unable to receive number of atoms from driver");
-    MPI_Bcast(&atom->natoms, 1, MPI_INT, 0, world);
+    atom->natoms = mdi_natoms;
+    MPI_Bcast(&atom->natoms, 1, MPI_LMP_BIGINT, 0, world);
```
akohlmey, line 235: @taylor-a-barnes @sjplimp this change is simply to correct the MPI_Bcast().
But should we simply overwrite this property? There is more to adding atoms than just increasing the variable.
Also please note the inconsistency with the "<NATOMS" command in the MDI library calls.

Furthermore, there are likely indexing issues for large systems since all expressions must be 64-bit save and also the MPI calls may need some adjustments. For the time being, It may be better to restrict the use of this functionality to cases where `natoms << 2billion`.

taylor-a-barnes: Yes, the lines in question are definitely insufficient to correctly change the number of atoms in the system.  These lines are really just placeholders for possible eventual support for the ">NATOMS" command.  Currently, LAMMPS does not register the ">NATOMS" command with MDI, and will raise an error if a driver attempts to use it.  The lines can be safely removed, if desired.

MDI now supports communication of 64-bit ints.  For example:

`ierr = MDI_Recv((char *) &atom->natoms, 1, MDI_INT64_T, mdicomm);`

If the driver is using differently sized integers, MDI automatically makes the conversion.  As long as both the engine and driver use types that are sufficient for the full scope of their possible use cases, there shouldn't be any risk of integer overflows.

akohlmey: Ok, thanks. I've updated the code to print an error, if it every encounters this statement and added a comment explaining why.

sjplimp (approved):

PR #2788, opened 2021-06-02
Track properties of pair interactions

jtclemm:
**Summary**

This pull request includes a method to track all pairwise interactions and output data pertaining to the history of the interaction when the two particles stop interacting (e.g. the initial time of the interaction, the average distance of the interaction over its lifetime, etc). This is helpful for generating distributions of interaction lifetimes or efficiently dumping spatial locations of lost interactions to identify regions of rapid particle rearrangement. 

Changes include the addition of a new pair style tracker, a new fix pair/tracker (other name suggestions?), and minor changes to the dump and fix neigh/history classes. Users can overlay the new pair style using pair hybrid/overlay as it calculates no forces. It only monitors interactions using an instance of fix neigh/history. The new pair style is compatible with both atomic and finite particles.

I also deleted a duplicate paragraph in the pair granular documentation page.

**Related Issue(s)**

NA

**Author(s)**

Joel Clemmer (SNL)

**Licensing**

**Backward Compatibility**

No issues.

**Implementation Notes**

The pair style must be used in conjunction with the new fix. Whenever two particles separate beyond the cutoff of the interaction, the pair style sends data to the new fix which stores and processes results. Several properties can be calculated including the ids of participating atoms, the interaction lifetime, the center of mass position when the atoms separate, and the average/minimum distance between atoms. Data can be filtered by additional arguments to exclude short lived interactions or interactions between particular combinations of atom types. Data is stored in local arrays and can be output by dump local or processed by other fixes such as ave/histo.

A new keyword was added to dump_modify, `header yes` or `header no`, which flags whether or not the header is written since the information is not relevant for this method. An additional flag check was added to fix neighbor/history to enable compatibility with non-granular pairstyles.

One complication that Steve and I discussed a bit was the usage of the two special-bonds bits on neighbor lists. Fix neigh/history assumes granular pair styles don't use special bonds (maybe a bad assumption that should be addressed in the future) so it recycles those two bits to temporarily flag if particles are in contact. This isn't necessary (it's turned off for attractive granular pairstyles like JKR) but does complicate the usage of fix neigh/history with non-granular pair styles that use these bits for special bonds and the usage of a single neighbor list by two instances of fix neigh/history. As a temporary patch, I added a flag which can be externally set by the pair style which will cause fix neigh/history to skip these bits.

Examples and documentation are included.

Possible issues include:
1) At long times, it’s possible the summation of interaction distances could overflow leading to errors in the average distance calculated by the fix.
2) This method is currently incompatible with granular pair interactions that extend beyond contact (JKR and DMT).
3) The above issue with fix neigh/history.

**Post Submission Checklist**

NA


akohlmey:
@jtclemm I had to make multiple updates. Most importantly, you had added the new styles twice (in the MISC package and in src) which prevented compiling with CMake. But there were also issues related to integrating the styles into the documentation and coding style related updates.

jtclemm:
> 
> 
> @jtclemm I had to make multiple updates. Most importantly, you had added the new styles twice (in the MISC package and in src) which prevented compiling with CMake. But there were also issues related to integrating the styles into the documentation and coding style related updates.

Thank you for making those changes. I went through your commits to figure out all the documentation links. I'll try to integrate correctly in the future.

akohlmey:
@sjplimp this has passed our extended regression tests. So if there is any impact to the regular code paths, it would have to be subtle and/or rare.

sjplimp (approved):

sjplimp:
Thanks Joel for adding this!  @akohlmey this is now ready to merge from my perspective

PR #2795, opened 2021-06-10
Add fabric tensor compute

jtclemm:
**Summary**

This pull request adds a new compute that calculates the fabric tensor which quantifies the orientation of pair interactions. Four types of the fabric tensor are included: the contact tensor, the branch tensor, the normal force tensor, and the tangential force tensor (granular pair styles only). These tensors are commonly calculated in discrete element models of granular systems.  Hence the compute was added to the GRANULAR package. However, I can move it to MISC if preferred since the compute is generally compatible with most pair styles. 

**Related Issue(s)**

NA

**Author(s)**

Joel Clemmer (SNL)
Ishan Srivastava (LBL)

**Licensing**

**Backward Compatibility**

NA

**Implementation Notes**

Each tensor is calculated by summing various contributions from pairwise interactions using an occasional neighbor list. Interaction distances are either determined using the `force->pair->cutsq` array or by the sum of particle radii. The 6 components of each tensor are then stored in a local vector (xx, yy, zz, xy, xz, yz). If N tensors are requested, a vector of length N*6+1 is produced where the components of each tensor are stored in the order requested by the user and the final value is the number of interactions used to calculate the tensors.

Contributions from pairs are only included if both atoms are in the compute group. Using optional commands, one can also specify which pairs contribute to the tensors based on atom type.

**Post Submission Checklist**

NA


doc/src/compute_fabric.rst
```diff
@@ -0,0 +1,178 @@
...
+
+   compute 1 all fabric type contact force/normal type/include 1,2 3*4
+   compute 1 all fabric radius force/normal force/tangential
+
+Description
+"""""""""""
+
+Define a compute that calculates various fabric tensors for pairwise
+interactions :ref:`(Ouadfel) <Ouadfel>`. The *type* and *radius* settings
+are used to select whether interactions cutoffs are determined by atom types
```
sjplimp, line 52: interaction

sjplimp:
Thanks for adding this @jtclemm - Three additional comments.

a) Can you say something in the first paragraph of the doc page about how it is useful in particular for granular models but also non-granular, which you mention in the paragraph at the top of the PR

b) In the doc page, can you add some info on the cost of this compute, e.g. roughly the cost of a pair style invocation.  So as to caution users about using it too often.  B/c it loops over the neigh list and some options invoke pair->single()

c) What do you think of removing the pair count from the vector output and make that the scalar output?  I think that would simply use of this compute's output in other LAMMPS commands, e.g. fix ave/time or variable formulas where you only want to process the tensor values.

jtclemm:
I can make those changes. Moving the # of contacts to a scalar output makes sense. The only downside I see is that there's no way to ensure the user always outputs the vector with the scalar (as far as I know?). Therefore compute_vector() may have to call compute_scalar() (repeating some calculation if the user requests both) to ensure the fabric calculation always has an updated # of contacts.

sjplimp:
What I suggest is make the calcuation for counting the contacts a new internal function, that be invoked by both compute_scalar() and compute_vector().  And have a flag that indicates if it has already been invoked on this timestep.  Both the scalar and vector methods know the current timestep and can use that flag to decide whether to re-call the counting function.

sjplimp:
Probably counting and storing the contact list is what the new function should do?

jtclemm:
I see, that makes sense. What if compute_vector() just always called compute_scalar() and compute_scalar() first checks the invoked_scalar variable and compare to the current timestep to determine whether to reuse the old value? Is this the intended use of the invoked_X variables?

sjplimp:
that's fine too

jtclemm:
Great, I implemented the suggested changes and confirmed the compute still works.

sjplimp (approved):

PR #2796, opened 2021-06-10
Add support for SO3 descriptor to MLIAP package

macstein:
**Summary**
SO3 descriptor feature is implemented in MLIAP module

**Related Issue(s)**

**Author(s)**
@athomps @macstein @qzhu2017

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**


src/MLIAP/mliap_so3.cpp
```diff
@@ -0,0 +1,1422 @@
+/* ----------------------------------------------------------------------
+ LAMMPS - Large-scale Atomic/Molecular Massively Parallel Simulator
+ https://lammps.sandia.gov/, Sandia National Laboratories
```
akohlmey, line 3: The LAMMPS home page is now `https://www.lammps.org`

src/MLIAP/mliap_so3.cpp
```diff
@@ -0,0 +1,1422 @@
...
+  memory->create(m_ellpl1, totali, "MLIAP_SO3:m_ellpl1");
+  memory->destroy(m_ellm1);
+  memory->create(m_ellm1, totali, "MLIAP_SO3:m_ellm1");
+
+  for (int l = 1; l < m_lmax + 1; l++) {
+    m_ellpl1[l] = get_sum(0, l + 2, 1, 2);
+    m_ellm1[l] = get_sum(0, l, 1, 2);
+  }
+
+  double pfac1 = 1.0 / 4.0 / M_PI;
```
akohlmey, line 178: the `M_PI` constant is a commonly used but non-standard extension to the C/C++ standard. Using it unconditionally will cause compilation failure on some platforms. Please use `MY_PI` from the `MathConst` namespace. That will also promote a platform independent value for PI.

src/MLIAP/mliap_descriptor_so3.cpp
```diff
@@ -0,0 +1,298 @@
...
+  rfac0 = 0.99363;
+  rmin0 = 0.0;
+
+  // open SO3 parameter file on proc 0
+
+  FILE *fpparam;
+  if (comm->me == 0) {
+    fpparam = utils::open_potential(paramfilename,lmp,nullptr);
+    if (fpparam == nullptr)
+      error->one(FLERR,fmt::format
```
akohlmey, line 75: `fmt::format()` is no longer needed for printing error messages with formatted strings. That functionality has been included into `Error::one()` and `Error::all()` through overloading.

src/MLIAP/mliap_descriptor_so3.cpp
```diff
@@ -0,0 +1,298 @@
...
+        ("Cannot open SO3 parameter file {}: {}",
+          paramfilename, utils::getsyserror()));
+  }
+
+  char line[MAXLINE],*ptr;
+  int eof = 0;
+  int n,nwords;
+
+  while (1) {
```
akohlmey, line 84: We are trying to get rid of this kind of file parser code construct and reduce the use of C-style string functions and particularly get rid of `strtok()` (which is not thread safe). There are multiple convenience classes and functions available. Please see https://docs.lammps.org/Developer_utils.html and the examples given or how those are used throughout other places in LAMMPS. This is a work in progress, but that is why we are particularly interested in having newly added code adopt these tools.

src/MLIAP/mliap_descriptor_so3.cpp
```diff
@@ -0,0 +1,298 @@
...
+    if (nwords == 0) continue;
+
+    // words = ptrs to all words in line
+    // strip single and double quotes from words
+
+    char* keywd = strtok(line,"' \t\n\r\f");
+    char* keyval = strtok(nullptr,"' \t\n\r\f");
+
+    if (comm->me == 0) {
+      utils::logmesg(lmp, fmt::format("SO3 keyword {} {} \n",
```
akohlmey, line 111: Also `utils::logmesg()` does accept [a format string and a variable number of arguments now](https://docs.lammps.org/Developer_utils.html#_CPPv4I0DpEN9LAMMPS_NS5utils7logmesgEvP6LAMMPSRK1SDpRR4Args)

src/MLIAP/mliap_descriptor_so3.cpp
```diff
@@ -0,0 +1,298 @@
+/* ----------------------------------------------------------------------
```
akohlmey, line 1: If you want that the LAMMPS clang-format formatting is applied, then please reformat your code with it, otherwise add a `// clang-format off` as the first line.

src/MLIAP/mliap_descriptor_so3.cpp
```diff
@@ -0,0 +1,298 @@
...
+   Copyright (2003) Sandia Corporation.  Under the terms of Contract
+   DE-AC04-94AL85000 with Sandia Corporation, the U.S. Government retains
+   certain rights in this software.  This software is distributed under
+   the GNU General Public License.
+
+   See the README file in the top-level LAMMPS directory.
+------------------------------------------------------------------------- */
+
+#include "mliap_descriptor_so3.h"
+#include "atom.h"
```
akohlmey, line 15: there should be an empty line before `#include "atom.h" ` so that clang-format's sorting of include statements does not move the class header, which should be the very first include file.

src/MLIAP/mliap_descriptor_so3.cpp
```diff
@@ -0,0 +1,298 @@
...
+        elementsflag = 1;
+      } else if (strcmp(keywd,"radelems") == 0) {
+        for (int ielem = 0; ielem < nelements; ielem++) {
+          radelem[ielem] = utils::numeric(FLERR,keyval,false,lmp);
+          keyval = strtok(nullptr,"' \t\n\r\f");
+        }
+        radelemflag = 1;
+      } else if (strcmp(keywd,"welems") == 0) {
+        for (int ielem = 0; ielem < nelements; ielem++) {
+          wjelem[ielem] = atof(keyval);
```
akohlmey, line 138: Using `atof()` is dangerous since it does not check for a valid argument. At the very least this should use `utils::numeric()`, but even better would be to adapt the parser to use the aforementioned convenience classes, which do check for correctness.

src/MLIAP/mliap_descriptor_so3.cpp
```diff
@@ -0,0 +1,298 @@
...
+  read_paramfile(paramfilename);
+
+  so3ptr = new MLIAP_SO3(lmp,rcutfac,lmax,nmax,alpha);
+
+  ndescriptors = so3ptr->ncoeff;
+}
+
+MLIAPDescriptorSO3::~MLIAPDescriptorSO3()
+{
+  delete so3ptr;
```
akohlmey, line 50: This looks as if there would be memory leakage due to missing `delete` statements. Have you checked whether your code is "valgrind clean"?

src/MLIAP/mliap_descriptor_so3.cpp
```diff
@@ -0,0 +1,298 @@
...
+      ij++;
+    }
+
+  }
+}
+
+void MLIAPDescriptorSO3::compute_force_gradients
+  (class MLIAPData* data)
+{
+
```
akohlmey, line 274: If those functions are intentionally empty, they can be defined in the header. If not, they probably should contain some call to the Error class here.

src/MLIAP/mliap_so3.cpp
```diff
@@ -0,0 +1,1422 @@
...
+
+#include "mliap_so3.h"
+#include <cmath>
+#include "math_const.h"
+#include "math_special.h"
+#include "memory.h"
+#include "error.h"
+#include "comm.h"
+
+#ifdef MLIAP_SO3_MKL
```
akohlmey, line 26: Is there a significant performance advantage from using MKL here?
If yes, then searching for MKL and a more generic define (e.g. `-DLMP_MLIAP_USE_MKL`) should be used otherwise we should just stick with the simpler code path.

src/MLIAP/mliap_so3.cpp
```diff
@@ -0,0 +1,1422 @@
...
+  double pfac1 = 1.0 / 4.0 / M_PI;
+  m_pfac_l1 = m_lmax + 2;
+  m_pfac_l2 = (m_lmax + 2) * (m_lmax + 2) + 1;
+  totali = m_pfac_l1 * m_pfac_l2;
+  memory->destroy(m_pfac);
+  memory->create(m_pfac, totali, "MLIAP_SO3:m_pfac");
+  for (int l = 0; l < m_lmax + 2; l++)
+    for (int m = -l; m < l + 1; m++)
+      m_pfac[l * m_pfac_l2 + m] = sqrt((2.0 * l + 1.0) * pfac1)
+        * pow(-1, m);
```
akohlmey, line 187: You probably want to use `MathSpecial::powsign()` instead of `pow(-1,m)`.

src/MLIAP/mliap_so3.cpp
```diff
@@ -0,0 +1,1422 @@
...
+}
+
+void MLIAP_SO3::swap(double *a, double *b)
+{
+  double t = *a;
+  *a = *b;
+  *b = t;
+}
+
+int MLIAP_SO3::partition(double arr[], int low, int high,
```
akohlmey, line 353: The Jacobi class included in LAMMPS can return sorted Eigenvalues directly, IIRC. So this would be superfluous.

akohlmey (dismissed):
This is a first pass with comments on your submission purely from visual inspection. There are likely more suggestions coming when I get around to compiling and testing the code.

src/MLIAP/mliap_so3_math.h
```diff
@@ -0,0 +1,684 @@
+
+#ifndef SRC_SOT_MLIAP_SO3_MATH_H_
+#define SRC_SOT_MLIAP_SO3_MATH_H_
+
+#include<cassert>
+
+#ifndef _MATRIX_ALLOC_JPD_H
```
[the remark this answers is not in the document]
macstein: @akohlmey I appreciate for all important comments which I am working on now. 
Regarding mliap_so3_math.h, yes I copied code from "https://github.com/jewettaij/jacobi_pd". It is free of copyright under CC0 1.o Universal license. The author said that "This repository was written from scratch" and "You can do anything you like with this code". I think that the author did not copy other source, hence the author has the copy right and allow us use it freely. Honestly, I don't understand about the copy right perfectly, I need deep studying these license. 
Do you allow that I put the proper acknowledge and just modify code formatting according to LAMMPS internal format rule?
If not, I need to use  "math_eigen_impl.h". I need to get eigen value with dynamic matrix(n,n), where n is not fixed. Do you think  I can easily implement the function from "mat_eigen_impl.h"? 
I also need matrix(n,n) inverse function. Does LAMMPS include a function which support the operation?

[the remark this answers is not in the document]
akohlmey: Please have a closer look at the code in math_eigen_impl.h. You may be surprised.

src/MLIAP/mliap_so3_math.h
```diff
@@ -1,684 +1,181 @@
...
-  void MaxEntry(Scalar const *const *M, int& i_max, int& j_max) const;
-
-  void SortRows(Vector v, Matrix M, int n,
-                SortCriteria s=SORT_DECREASING_EVALS ) const;
-
-  void Alloc(int N);
-  void Init();
-  void Dealloc();
+  int i,j;
+  double mat_cpy[n][n];
```
akohlmey, line 28: This statement is not legal C++ it is only acceptable in C98 (and allowing it is considered a mistake). GCC will warn about it with -Wvla. You have to either explicitly allocate and delete it or use std::unique_ptr.

src/MLIAP/mliap_descriptor_so3.cpp
macstein: I have checked memory leak with "valgrind". I found no leak adding those lines.

akohlmey:
@macstein I make some cosmetic updates and silenced compiler warnings. There were a few minor issues also detected by compiler warnings (mismatched array/non-array delete statements and some additional parenthesis) that I corrected.

There are several functions that have unused parameters (you can recognize them from the commented out parameter names).
Those can probably be removed since those don't look like mandatory parameters set by a base function.

akohlmey:
@athomps @macstein @qzhu2017

This PR is nearing completion. I did a little refactoring of the MLIAP base code to resolve some memory management issues, reduce code redundancy, and also avoid memory leaks. There is now also a unit test entry (which I used for the memory leak testing, as that covers many more usage scenarios with restarts and specifying/changing pair styles etc.).

The major missing piece seems to be the documentation. I've added the minimal parts to the `doc/src/pair_mliap.rst` file, but they need to be extended/completed where you can see the FIXME tags.

It looks like the new descriptor does some very large memory allocations, but I cannot easily tell where and if or how they could be avoided.

akohlmey:
Oh, and there should be a correct "contributing author(s)" attribution in the added source files.

macstein:
> @athomps @macstein @qzhu2017
> 
> This PR is nearing completion. I did a little refactoring of the MLIAP base code to resolve some memory management issues, reduce code redundancy, and also avoid memory leaks. There is now also a unit test entry (which I used for the memory leak testing, as that covers many more usage scenarios with restarts and specifying/changing pair styles etc.).
> 
> The major missing piece seems to be the documentation. I've added the minimal parts to the `doc/src/pair_mliap.rst` file, but they need to be extended/completed where you can see the FIXME tags.
> 
> It looks like the new descriptor does some very large memory allocations, but I cannot easily tell where and if or how they could be avoided.

@akohlmey With respect to very large memory allocations, we actually intended to use large memory to speed up the SO3 descriptor calculation. To avoid repeating operations, we pre-computed some factors saving in the memory, and reused it during the calculations. By this approach, we could improve the speed about 80 times faster. For 8000 Si atoms system, the maximum usage of the memory reached to 1.2GB (as far as I remember). Within my capablilities, this is the best solution so far.

akohlmey:
> With respect to very large memory allocations, we actually intended to use large memory to speed up the SO3 descriptor calculation. To avoid repeating operations, we pre-computed some factors saving in the memory, and reused it during the calculations. By this approach, we could improve the speed about 80 times faster. For 8000 Si atoms system, the maximum usage of the memory reached to 1.2GB (as far as I remember). Within my capablilities, this is the best solution so far.

That is ok, but I think that should be mentioned in the documentation and particularly which factors determine the amount of memory used for these purposes, e.g. as a `.. note::` paragraph. When you are running on Linux, you can query the maximum memory size in use during a run on MPI rank 0 with `info memory`.

qzhu2017:
@akohlmey @macstein @athomps  I just edited a bit on `pair_mliap.rst` to improve the documentation on both SO3 descriptor and NN model. Please take a look. I also fix a tiny typo for the reference in `pair_snap.rst`.

akohlmey:
> For 8000 Si atoms system, the maximum usage of the memory reached to 1.2GB (as far as I remember). Within my capablilities, this is the best solution so far.

@macstein I've looked through the code with this in mind and made a few additional changes.
- I have added two accumulator variables to keep a more accurate tally of the allocated memory and use those to report memory usage
- I have promoted some 32-bit integers to "bigint", i.e. 64-bit integer under normal circumstances to reduce the risk of them overflowing for larger systems. At the moment these systems would probably be too slow for most people to run, but that can change.
- I have simplified the explanation of the large memory allocation. To state that they depend on the total number of neighbors per MPI rank should be a sufficient information as that implies that the memory usage can be reduced by using more MPI ranks.

I would also suggest to change the name of the "natoms" parameter. Within the context of the LAMMPS code the (local) variables "natoms", "nlocal", "nghost", and "nall" have a particular meaning by convention (or repetition). "natoms" is the total number of atoms in the system, "nlocal" the number of 'owned' atoms of an MPI rank, "nghost" the corresponding number of 'ghost' atoms and "nall" the sum of "nlocal" and "nghost". When processing neighbor lists, the size of the neighbor list is often stored in a local variable "inum", but it is not necessary to use that specific variable. It is just that using "natoms" can be confusing to people because they will assume that it stores the total number of atoms in the system (which it does not).

akohlmey:
@athomps do you have any comments on this PR?

athomps (approved):
Everything looks good to me.

PR #2798, opened 2021-06-11
Collected small changes

akohlmey:
**Summary**

This pull request fixes multiple smaller issues and collects small updates.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

- multiple small updates and bugfixes in the USER-DIELECTRIC package
- additional valgrind suppressions for newer GCC compilers
- update the version of googletest used to v1.11
- update clang CMake preset to use `flang` as Fortran compiler, if available
- recognize a few more compiler ids when adapting settings according to compilers

**Post Submission Checklist**


ndtrung81 (approved):
@akohlmey thanks for cleaning up and for the bugfixes. In fact, I haven't had a chance to test pair style lj/long/coul/long/dielectric, so it could be a mistake to include it in the package at this point.

akohlmey:
> @akohlmey thanks for cleaning up and for the bugfixes. In fact, I haven't had a chance to test pair style lj/long/coul/long/dielectric, so it could be a mistake to include it in the package at this point.

It is ok for as long as we have not yet create a patch release. We can remove (or hide) the style in that case before the release.
We should also look into having a template to run unit tests on those styles. The unit tests have been extremely valuable in identifying memory and programming issues for a lot of pair styles.

sjplimp (approved):

Issue #2800, opened 2021-06-14
[Feature Request] Stillinger criterion in fix gcmc


evenmn:
**Summary**

The `fix gcmc` command takes an argument `overlap_cutoff`, which is the smallest accepted separation distance between two atoms. Similarly, it would be natural that the command could take the opposite argument, `max_sep` (?), which is the maximum distance between an added atom and all other atoms in the box/region. This is what is known as the Stillinger cluster criterion [1].

**Detailed Description**

I am currently working on evaluating water MD potentials, and to measure properties such as the surface tension. For this, it would be very useful to define a cluster using the Stillinger cluster criterion (SCC) when performing GCMC simulations. SCC is just a way to define which atoms that belong to a cluster by assigning atoms separated by a distance smaller than r_1 to the same cluster.

First, I wrote a code that defined the region that is part of a cluster according to SCC. Then, I realized that the ´fix gcmc´ command does not support dynamic regions. It does not seem like it is possible to achieve this with the current version of LAMMPS, but I could be wrong.

Anyway, it should not be very much work to implement this as an argument to ´fix gcmc´ as it would be just the opposite of the ´overlap_cutoff´ argument. The difference is that the distance to all particles has to be larger than r_1 to set the interaction energy to max, but for ´oveerlap_cutoff´ only one separation distance has to be smaller than r_0 for the interaction energy to be set to max.

**Further Information, Files, and Links**

[1] [Rigorous Basis of the Frenkel‐Band Theory of Association Equilibrium](https://aip.scitation.org/doi/10.1063/1.1776907)

akohlmey:
@evenmn Is this still something you are working on? Please let us know.

evenmn:
I am currently not working on this. It turned out to be more complicated than what I first thought, and I fear that I will need other Monte Carlo samling techniques (like aggregate-volume bias Monte Carlo) to achieve what I want. To implement other samling techniques would require significant changes of the MC package, which I neither have time or knowledge to do at this point. If other people have the need of this feature and would like to implement it, that would be great, but meanwhile I turn my attention to other (less user-friendly) Monte Carlo packages.

akohlmey:
@evenmn thanks for responding. Making LAMMPS more MC-friendly is not a priority, since many MC operations are conflicting with the domain decomposition parallelization that is at the core of the code design. That is part of the reason why fix gcmc has become so complex and convoluted. So in general we will stick to MC operations that need to be called infrequently for its common use cases.

PR #2801, opened 2021-06-14
Added 'max_sep' argument to fix gcmc command

evenmn:
**Summary**

The ´fix gcmc´ command now takes max separation distance, ´max_sep´, as a proposal argument. It makes it possible to force all atoms to be in a cluster according to the Stillinger cluster criterion. If the initial system consists of one cluster, there will never be more than one cluster. 

**Related Issue(s)**

This PR is related to issue #2800. 

**Author(s)**

Even Marius Nordhagen, University of Oslo

**Licensing**

**Backward Compatibility**

There is no reason the added code should not be backward compatible with older LAMMPS versions.

**Implementation Notes**

Implementation of the ´max_sep´ argument was more or less copy pasting the implementation of the ´overlap_cutoff´ argument. LAMMPS was compiled with SMALLBIG and BIGBIG. If the initial system is empty, the system will remain empty during the GCMC simulation when this argument is used. This should probably raise an error message.

**Post Submission Checklist**

### DIFF
--- a/src/MC/fix_gcmc.cpp
+++ b/src/MC/fix_gcmc.cpp
@@ -269,7 +269,9 @@ void FixGCMC::options(int narg, char **arg)
   energy_intra = 0.0;
   tfac_insert = 1.0;
   overlap_cutoffsq = 0.0;
+  max_sepsq = 1.0e50;
   overlap_flag = 0;
+  max_sep_flag = 0;
   min_ngas = -1;
   max_ngas = INT_MAX;
 
@@ -380,6 +382,12 @@ void FixGCMC::options(int narg, char **arg)
       double rtmp = utils::numeric(FLERR,arg[iarg+1],false,lmp);
       overlap_cutoffsq = rtmp*rtmp;
       overlap_flag = 1;
+      iarg += 2;
+    } else if (strcmp(arg[iarg],"max_sep") == 0) {
+      if (iarg+2 > narg) error->all(FLERR,"Illegal fix gcmc command");
+      double rtmp = utils::numeric(FLERR,arg[iarg+1],false,lmp);
+      max_sepsq = rtmp*rtmp;
+      max_sep_flag = 1;
       iarg += 2;
     } else if (strcmp(arg[iarg],"min") == 0) {
       if (iarg+2 > narg) error->all(FLERR,"Illegal fix gcmc command");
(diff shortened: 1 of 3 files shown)


akohlmey:
@evenmn thanks for the pull request. To be considered for inclusion, you would also have to update the documentation in `doc/src/fix_gcmc.rst`.

evenmn:
Updated documentation is added

src/MC/fix_gcmc.cpp
```diff
@@ -2281,16 +2314,24 @@ double FixGCMC::energy_full()
...
 
-        if (rsq < overlap_cutoffsq) {
+        if (overlap_flag && rsq < overlap_cutoffsq) {
           overlaptest = 1;
           break;
         }
+        if (max_sep_flag && rsq > max_sepsq) {
+          max_septest = 1;
+          break;
+        }
```
athomps, line 2324: This logic seems wrong.  A connected cluster will naturally have many occurrences of `rsq > max_sepsq`.

athomps (commented):
Thanks for this submission.  It is a nice idea.  However, I think there are some issues with the logic in energy_full(), maybe else where too. See my comment near line 2317.  In general, it is simpler to detect overlap than isolation.  Also, in the code and in the doc page, you refer to this feature as ensuring a single cluster.  Formally, this would not be correct if the system initially contains more than one cluster.  A better way to think about it is prevention of an isolated atom or molecule.  Do you intend to make this work for polyatomic molecules? If not, document that add an error check.  Also, in the code, I suggest changing `belongs_to_cluster` to the logical inverse `isolated`. Finally, how do you handle cases where deletion of an atom breaks the cluster in to two parts.  If this is not handled, that should be explained in the documentation e.g. ``this method prevents insertion of an isolated molecule, but does not prevent formation of isolated molecules by deletion''

src/MC/fix_gcmc.cpp
evenmn: Thank you for a thorough review. I very much agree with you that it is better to look at this function as a way to prevent isolated atoms, rather than enforce atoms to be in a cluster. To answer your questions, we need to consider what would be the typical use case of this function. Usually, one wants to prevent isolated atoms when only the cluster(s) are of interest, which is also my case. Then it makes sense to reject moves that split up a cluster in addition to rejecting moves where isolated atoms are added. I didn't really think about this issue before you mentioned it. One possible way of doing this would be to count the number of clusters and check whether or not the number of clusters is conserved when performing a move. However, I don't know if this would be the most efficient way.

Another major issue with this approach, is that most of the swap moves will be rejected. It is very unlikely that a particle will be inserted inside a cluster, and the zone around a cluster where a particle is allowed to be inserted is typically very narrow. To overcome this issue, modified Monte Carlo sampling techniques, like aggregation-volume-bias Monte Carlo (AVBMC) [1] and even energy-biased aggregation-volume-bias Monte Carlo (EB-AVBMC) [2] have been suggested. Do you think it would be possible to implement these kinds of techniques in LAMMPS, or is it out of the scope of the package? 

Anyway, I realize that there are several issues with this implementation that I were not aware of. I may come back with a more thought-through PR later, but for now on I think it's better to not merge this PR.

Thank you for your help!

[1] https://pubs.acs.org/doi/10.1021/jp709693g
[2] https://pubs.acs.org/doi/10.1021/acs.jctc.5b00466

PR #2809, opened 2021-06-21
Upgrades fmtlib to v8.0.1

rbberger:
**Summary**

This PR upgrades the included fmtlib files to v8.0.1

**Author(s)**

Richard Berger and Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

We have a few custom patches in fmtlib that had to be reapplied afterward. Does not include the `src/fmt.cc` file, which is a C++20 module.

**Post Submission Checklist**

https://github.com/fmtlib/fmt/releases/tag/8.0.1


akohlmey:
FYI, looks like version 8.0.0 changes the behavior for formatting floating point numbers with `{:.8}`. With version 7.x it would always print 8 digits output and thus have more "even" output. Now it is skipping over trailing zeroes, which is what caused the mismatch in the unit tests.

akohlmey:
@rbberger should we merge this now? It doesn't look like there will be a a version 8.0.2 and all compilation issues plus the unit test failures seem to be resolved.

rbberger:
@akohlmey LGTM

sjplimp (approved):

PR #2810, opened 2021-06-21
Collected small changes

akohlmey:
**Summary**

This pull request combines multiple smaller sets of changes

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

- new `compute dipole` command to compute global scalar or vector similar to `compute dipole/chunk`
- fix cut-n-past bug in AtomVec class
- replace non standard compliant variable length arrays in MLIAP SO3 support with normal new/delete
- downgrade googletest from version 1.11 to 1.10, since GCC 4.8 fails to compile version 1.11 due to incomplete C++11 support
- turn off extra warnings in gcc.cmake preset and introduce a pedantic.cmake preset with those warnings and triggering an error for variable length arrays
- update USER-RANN example to be more consistent with other examples. Create log files.
- skip tests for the Fortran library interface, if CMake cannot identify the Fortran compiler (e.g. the new flang aka f18)
- fix minor documentation format and spelling issues

**Post Submission Checklist**


src/MLIAP/mliap_so3.cpp
```diff
@@ -374,23 +374,20 @@ void MLIAP_SO3::compute_W(int nmax, double *arr)
   double *sqrtD = new double[n * n];
   double *tempM = new double[n * n];
 
-  double **temparr;
-  double tempout[n];
```
akohlmey, line 378: @macstein FYI, this and the two cases of variable length arrays below slipped through our automated tests. These "variable length" arrays are not allowed by C++ (only in C98). In C++ "n" would have to be a compile time constant.

athomps (approved):
I approve

PR #2821, opened 2021-07-01
USER-CGDNA: New Class AtomVecOxdna

ohenrich:
**Summary**

This pull request implements the new AtomVecOxdna class and fixes a problem with occasionally and temporarily frustrated stacking interactions. It stores the polarity of the nucleotide strand through per-atom information of the 5'-partner of each atom in a bond. Note that the directionality was previously worked out 'on-the-fly', which proved to be not robust enough in very flexible single-stranded overhangs.

The pull request contains as well updated benchmarks and examples, which consist now exclusively of equilibrated systems with thermal translational and rotational velocities.

**Author**

Oliver Henrich
Department of Physics
University of Strathclyde
Glasgow, UK

**Licensing**

**Backward Compatibility**

Backward compatibility is not retained in the input files, but can be easily restored by adding the keyword "oxdna" to the atom style like "atom_style hybrid bond ellipsoid oxdna".  

**Implementation Notes**

The atom ID of the 5'-partner is set from the "Bonds" section in the data file. If a bond is defined as "bond_ID bond_type atom1 atom2", then atom2 is the 5' partner of atom1 and this information is stored with atom1 in a pointer **id5p defined in the Atom class. 

A method data_bonds_post() has been defined in AtomVecOxdna that reprocesses the bond section of the data file and sets the id5p values accordingly. The polarity information is used in the pair styles oxdna/stk and oxrna2/stk to decide which atom lies in 3' and 5' direction.

The new implementation has been verified against the existing benchmark cases in the examples section. Note that these have been updated and all configurations in the data files are now equilibrated. 

**Post Submission Checklist**


sjplimp:
@ohenrich - looking better.  There are 3 things I still don't understand:
a) why is iid5p a peratom one-column array and not a peratom vector?
b) where is id5p initialized to 0 ?  this would be the data_atom_post code in (5) of my earlier email
c) is there some reason I'm not seeing why you need to call this from read_data, and not from Atom::data_bonds() as I suggested?  The latter is the way the other style-specific hook functions in AtomVec are designed to work.

ohenrich:
@sjplimp

> @ohenrich - looking better. There are 3 things I still don't understand:
> a) why is id5p a peratom one-column array and not a peratom vector?

I tried this and it didn't work at first when I had  add_peratom("id5p",&id5p,tagintsize,0); Something went wrong in a grow->pointers() method. Now I tried again and it works (see latest push). My repository might have been unclean.

> b) where is id5p initialized to 0 ? this would be the data_atom_post code in (5) of my earlier email

The initialisation to 0 comes from the allocation. The buffer holds the bond information a la 'bond_id bond_type atom1 atom2' The id5p of atoms that are only an 'atom2' in the data file are left untouched by AtomVecOxdna:data_bonds_post(). Unsafe?

> c) is there some reason I'm not seeing why you need to call this from read_data, and not from Atom::data_bonds() as I suggested? The latter is the way the other style-specific hook functions in AtomVec are designed to work.

I tried this by a call to avec->data_bonds() from within Atom:data_bonds() as you suggested, but it never called the child class in AtomVecOxdna, only the empty parent class in AtomVec. There might be something I don't understand. I will give it another try on Monday.

sjplimp:
@ohenrich @akohlmey 
Re: init to 0, I don't think allocation does that.  Axel, correct me if wrong.  So you must init values to 0 that won't be set by the bond list.  My eariler email gave details on what func AVOx needs to provide to do this.
Re: calling into AVOx from Atom::data_bonds. This should definitely work.  What I forgot is that you are using atom style hybrid.
So there needs to be a data_bond_post() method added to AVHybrid.  Similar to the one already there for date_atom_post(),
with the args you are passing from Atom::data_bonds().  That new AVH method willl pass the args along to all the hybrid sub-styles. From your current code, it looks you will also need id_offset as an arg to these data_bond_post() methods in the various Avec styles.

akohlmey:
> Re: init to 0, I don't think allocation does that. Axel, correct me if wrong. So you must init values to 0 that won't be set by the bond list. My eariler email gave details on what func AVOx needs to provide to do this.

Neither `malloc()` not `new` will initialize the allocated data. However, on Linux system they often appear to be initialized to zero since the glibc malloc library reserves memory through doing `mmap()` on `/dev/zero` with copy-on-write semantics. That will speed up large memory allocations and postpone having to allocate physical RAM or swap until it is actually used (on a per memory page basis). However, when using the `clear` command and doing other memory operations or using the library interface to run multiple simulations, the memory will be freed and then "recycled" and then those pages are no longer filled with zeroes. This is one of the reasons why it is important to implement those unit tests, since those will do those operations (and worse). Also running with valgrind's memory checker often spots such "implicit initialization" cases.

sjplimp:
@ohenrich Also to be clear, the call from Atom::data_bonds() into AVOx should just pass the data for a single line read from the Bonds sections of the data file, not N line strings.  And it should be called for both the newton on/off code paths, and have 5 args I think: m, num_bond[m], atom1, atom2, id_offset.  The data_bond_post() method you write for AVOx can use that to set (or not) id5p.  Other future atom styles could implement that method for a different purpose and they might need to know which bond for which atom that data is now stored with, i.e. m and num_bond[m].

akohlmey:
@ohenrich your PR has become a victim of our "big package reorganization" that is dropping the "USER" designation from bundled packages (except USER-MISC, but a lot of its content will be redistributed to other packages).
Since merging with upstream can be complicated in this case, you probably want to make a copy of the changed files, create a new feature branch from the current master and then re-add those files and any additional changes and then submit this as a new pull request.

akohlmey:
Superseded by #2827

PR #2825, opened 2021-07-04
Improve performance of fourier/simple angle style [DRAFT]

### LINKED ISSUE #2819: [Feature Request] I Would Like To Improve the Performance of the Fourier/Simple Angle Style
**Summary**

The [fourier/simple](https://docs.lammps.org/angle_fourier_simple.html) angle style does not have an optimized version. Namely, it is using the cmath cos() and acos() functions which appear to be a bottleneck in my use case. 
I would like to implement the corresponding improvements and propose them, but would request some guidance on the accepted accuracy of approximations before opening a pull request.

**Detailed Description**

I am currently involved in the simulation of polymer networks, namely PDMS. According to a recent paper [1], an angle term of the form of $\kappa * k_B * T * (1 - cos(x))$ should be used with $\kappa = 0.013$. 
This can be modeled in LAMMPS with a fourier/simple angle style with K = $\kappa * T * k_B$ and c = n = 1.
We noticed that adding this angle term leads to a slow-down of ca. 40-50%. 
I identified the bottleneck to be the cmath cosine and arcus cosine as used currently. I also noticed that while the documentation mentions optimized versions with differing accurracy ("Styles with a gpu, intel, kk, omp, or opt suffix are functionally the same as the corresponding style without the suffix. They have been optimized to run faster, depending on your available hardware..."), those do not seem to actually exist for this angle style, except for the OMP version. 
I thereafter implemented a version of the angle style with approximations for [cos](https://stackoverflow.com/questions/18662261/fastest-implementation-of-sine-cosine-and-square-root-in-c-doesnt-need-to-b/28050328#28050328) and [acos](https://stackoverflow.com/questions/3380628/fast-arc-cos-algorithm/26030435#26030435), which reduced the slow-down of adding the angle term to ca. 5-10%.

I would like to contribute to LAMMPS and offer such a performance-improved version, but before I open a corresponding pull-request, I would like to know what "acceptable accuracy policy" you have for the optimized versions, meaning, what floating-point accuracy would you expect from the trigonometric approximations?
Also, two software design questions: I assume this would be a new style (suffix opt), correct? And the trigonometric approximation fuctions, should they live in the namespace of this angle style or in e.g. the MathExtra namespace?

Performance observations on [Euler](https://scicomp.ethz.ch/wiki/Euler) on 6 nodes, average of 15 subsequent observations:

Network of 200 chains with 500 beads per chain and 100 cross-linkers with a functionality of 3.89, FENE bonds, LJ-potential and Langevin thermostat with parameters as described in [1]:

Without angle term: 137(3) ktau/day
With angle term: 101(3) ktau/day
With angle term, fast cos: 130(5) ktau/day

Network of 5000 chains with 20 beads per chain and 2500 cross-linkers with a functionality of 3.97, FENE bonds, LJ-potential and Langevin thermostat with parameters as described in [1]:

Without angle term: 128.9(4) ktau/day
With angle term: 93(5) ktau/day
With angle term, fast cos: 118(4) ktau/day

**References**

[1] [Macromolecules 2020, 53, 6, 1901–1916](https://pubs.acs.org/doi/abs/10.1021/acs.macromol.9b02428)

GenieTim:
**Summary**

Adds a fourier/simple/approx angle style that is very much faster than fourier/simple by using approximations of cosine and related trigonometric functions.

**Related Issue(s)**

Fixes #2819.

**Author(s)**

Tim Bernhard, ETHZ, tim@bernhard.dev

**Licensing**

**Backward Compatibility**

No backwards incompatibilities.

**Implementation Notes**

- uses (exact) same test as fourier/simple with exception to the error epsilon
- performance resp. accuracy could be further tuned by using different approximations
- the approximation still requires calculations, meaning, depending on the target device, a lookup table might have been even faster and possibly more accurate

**Post Submission Checklist**


src/USER-MISC/angle_fourier_simple_approx.cpp
```diff
@@ -0,0 +1,218 @@
...
+}
+
+/* ---------------------------------------------------------------------- */
+
+/**
+ * Approximation of cos
+ * maximum error is about 0.00109 for the range -pi to pi
+ * Source: https://stackoverflow.com/a/28050328/3909202
+ */
+double AngleFourierSimpleApprox::fastCos(double x) {
```
akohlmey, line 52: This function does not need to be a class member. It could as well just be a static function, which will improve its chances to be made inline.

src/USER-MISC/angle_fourier_simple_approx.cpp
```diff
@@ -0,0 +1,218 @@
...
+  constexpr double tp = 1./(TWO_PI);
```
akohlmey, line 53: For improved readability, please use `invtwopi` pr `inv2pi` instead of `tp` and `1.0` / `0.25` instead of `1.` / `.25` and equivalent.

src/USER-MISC/angle_fourier_simple_approx.cpp
```diff
@@ -0,0 +1,218 @@
...
+/**
+ * Approximation of acos 
+ * Returns the arccosine of x in the range [0,pi], expecting x to be in the range [-1,+1]. 
+ * Absolute error <= 6.7e-5
+ * Range mapping not necessary as C++ standard library would throw domain error
+ * Source: https://developer.download.nvidia.com/cg/acos.html
+ */
+double AngleFourierSimpleApprox::fastAcos(double x) {
+  double negate = double(x < 0);
+  x = abs(x);
```
akohlmey, line 74: Is this supposed to be `abs()` (operates on and returns an integer) or `fabs()` (operates on and returns a double precision floating point number)?

src/USER-MISC/angle_fourier_simple_approx.cpp
```diff
@@ -0,0 +1,218 @@
...
+}
+
+/* ---------------------------------------------------------------------- */
+
+/**
+ * Approximation of acos 
+ * Returns the arccosine of x in the range [0,pi], expecting x to be in the range [-1,+1]. 
+ * Absolute error <= 6.7e-5
+ * Range mapping not necessary as C++ standard library would throw domain error
+ * Source: https://developer.download.nvidia.com/cg/acos.html
```
akohlmey, line 70: Better to quote the original source (Abramowitz / Stegun) and look up and include the meaning/derivation of the constants. Best also promote them to double precision.

src/USER-MISC/angle_fourier_simple_approx.cpp
```diff
@@ -0,0 +1,218 @@
...
+
+    if (fabs(c)-1.0 > 0.0001) {
+      a = k[type]*C[type]*N[type]*sin(nth)/sin(th);
+    } else {
+      if (c >= 0.0) {
+        term = 1.0 - c;
+        sgn = 1.0;
+      } else {
+        term = 1.0 + c;
+        sgn = ( fmodf((float)(N[type]),2.0) == 0.0f )?-1.0:1.0;
```
akohlmey, line 178: This should be double precision. I'll include the corresponding change in the upstream non-approximated version
(cf. https://github.com/lammps/lammps/commit/455b7b93929040e977ecf35ee39d7350e25b7a66 )

akohlmey (commented):
Please see my individual comments for (formal) improvements and the pending commit for some formatting/programming style improvements to the upstream versions (to apply the latest include file conventions and use the recently added clang-format settings).

Nice idea to include the non-approximated unit test data.

Please note that this style needs to be added to the documentation, but you can just add it to the non-approximated documentation and the respective command lists and just mention it in the command summary and add one short paragraph stating that this is an approximated version with an estimated accuracy of XXX (seems to be around that of single precision floating point math).

GenieTim:
Thank you very much for all the insights and suggestions, dear @akohlmey, this early on in this draft. Very helpful indeed!

src/USER-MISC/angle_fourier_simple_approx.h
```diff
@@ -0,0 +1,93 @@
...
+ * Source: https://stackoverflow.com/a/28050328/3909202
+ */
+static double fastCos(double x)
+{
+  constexpr double inv2pi = 1. / (MathConst::MY_2PI);
+  // TODO: check if range map is necessary
+  double x_wrapped = x - MathConst::MY_2PI * floor(x * inv2pi);
+  x_wrapped *= inv2pi;
+  x_wrapped -= 0.25 + floor(x_wrapped + 0.25);
+  x_wrapped *= 16.0 * (abs(x_wrapped) - 0.5);
```
akohlmey, line 46: Shouldn't you be using fabs() here in both cases?

GenieTim:
While the acos() approximation seems quite sensible, the cosine one does not really match my expectations in terms of speed and accuracy. Without it (but with acos), the tests pass and the speedup compared to the fourier_simple without approx seems to be in the range of 10-20% slow-down (instead of 40-50% without optimizations or 5-10% with a cosine approximation). I will try to look for other approximations and improvements (while still trying not to use a lookup-table to reduce the memory footprint).

src/USER-MISC/angle_fourier_simple_approx.h
GenieTim: You are right, thanks!

akohlmey:
> While the acos() approximation seems quite sensible, the cosine one does not really match my expectations in terms of speed and accuracy. Without it (but with acos approximation), the tests pass and the speedup compared to the fourier_simple without approx seems to be in the range of 10-20% (instead of 40-50% omitting the whole angle term or 30-40% with a cosine approximation). I will try to look for other approximations and improvements (while still trying not to use a lookup-table to reduce the memory footprint).

I did a little researching and benchmarking myself: a simple way to get some speedup in exchange for accuracy on sin()/cos() seems to be using single precision. With GNU libc there also is a sincos() or sincosf() function where you can compute sine and cosine of the same angle at the same time (and with lower cost than both individually).
Finally, the ultimate speedup on sine/cosine is with using some Taylor/MacLaurin expansion, at the expense of further reducing accuracy. Some fast example would be (only works on little endian machines):
```
double fast_sin(double x) {
int k;
double y;
double z;
union 
{
    int i;
    double d;
} di;
    

z = x;
z *= 0.3183098861837907;
di.d = z + 6755399441055744.0;
z = k = di.i;
z *= 3.1415926535897932;
x -= z;
y = x;
y *= x;
z = 0.0073524681968701;
z *= y;
z -= 0.1652891139701474;
z *= y;
z += 0.9996919862959676;
x *= z;
k &= 1;
k += k;
z = k;
z *= x;
x -= z;

return x;
}


double fast_cos(double x) 
{
    return fast_sin(x+0.5*M_PI);
}
```

with single precision I get a relative(!) error across 10000 random values of about 5.e-6. 
with the polynomial expansion a relative error of about 2.e-4
the speedup for sinf() over sin() is about 1.8 and for the polynomial it is about 5.0

src/USER-MISC/angle_fourier_simple_approx.h
```diff
@@ -0,0 +1,157 @@
...
+
+  // 4 pieces of hills: wrap x within [0, pi/2]
+  if (x < MathConst::MY_PI2)
+    return sinePolynomial(x);
+  else if (x < MathConst::MY_PI)
+    return sinePolynomial(MathConst::MY_PI - x);
+  else if (x < 3.0f * MathConst::MY_PI2)
+    return -sinePolynomial(x - MathConst::MY_PI);
+  else
+    return -sinePolynomial(MathConst::MY_2PI - x);
```
GenieTim, line 77: The branching (and wrapping above) here is what makes this method slower than the Taylor series. Without it, it is faster (in a benchmark over 10'000'000 random, uniformly distributed angles).

akohlmey: @GenieTim which compiler and which compiler flags do you use to compile LAMMPS?
Have you experimented with using '-O3 -ffast-math' (with GCC)?
or using the Intel compiler with '?-O3 -xHOST -ffast-math'?

I've experimented with your polynomial approximation for a quadrant and merged it with the other code I had found which does an approximation for `[-pi/2;pi/2]` and used some tricks to avoid if statements, yet made the code work on a wider range of arguments. I am attaching a small standalone test code that first fills arrays with the random inputs and results and later compares them so that the compiler cannot optimize the loops away. To get better statistics on timing, and avoid that memory bandwidth becomes a significant fact, the code allows to set the size of the array (neval) and then does repetitions (nrepeat) and you can set the interval for the arguments. The test code demonstrates that at full optimization single precision slows you down and gives reduced accuracy for larger arguments. But with full optimization, the Intel 2021.2 and GCC 11.1 compilers beat all approximation with their inline functions.
[fast-sin-tester.cpp.txt](https://github.com/lammps/lammps/files/6796078/fast-sin-tester.cpp.txt)

GenieTim: For reference, I just uploaded the benchmark I was using to the following repository: https://github.com/GenieTim/cosine-benchmark
For compilation, I used the following: `clang++ main.cpp -std=c++20 -O3 -march=native -ffast-math -ftree-vectorize -Rpass-missed=loop-vectorize -Rpass-analysis=loop-vectorize`.

I highly appreciate your assistance, comments and the optimizations you found to wrap the ranges. Sorry that I cannot invest as much time as I would like to currently, I am confident though that my objective is going to be reached in not too long.

GenieTim: Running your (@akohlmey) benchmark on Euler, compiled with `icpc -O3 -xHOST -ffast-math -std=c++11 -qopt-report=2 -qopt-report-phase=vec fast-sin-tester.cpp`, I get the following results when called with `./a.out 12 10000000 0 100` (though different parameters lead to comparable results): 
```
sin():  120000000 evaluations @ 2.16156e-05 us/call
sinf(): 120000000 evaluations @ 2.045e-08 us/call
poly_sinf(): 120000000 evaluations @ 0.0212622 us/call
poly_sin(): 120000000 evaluations @ 0.0049543 us/call
fast_sin(): 120000000 evaluations @ 0.00235397 us/all
```
Very notable is the fact that the c-standard `sin()` and `sinf()` seem to be faster than any approximations. I can replicate this actually with my cosine benchmark, but not with an equivalent acos benchmark. Looking at the assembly, icpc generates calls to `__svml_cos4_l9`, the Intel short vector math library. This is not the case with clang or gcc (naturally), where "native" cosine calls are not vectorized I believe. While these observations further support the use of  an approximation for acos, a cosine approximation would need to be used on non-intel compilers only. It is not clear to me yet why in LAMMPS itself my initial implementation was faster, but there are a multitude of explanations of course (daily load, statistical fluctuation, flags not applied/CMake misunderstood, Intel compiler did not do this optimization, etc.).
Question for the last possibility: while I know how to get my vectorization reports and assembly for one file only when I compile this one file only, how can you get those for one file/component only if you compile it into the whole (LAMMPS) project (is there some CMake magic I do not know of?)?

akohlmey: @GenieTim Unlike the simple test code with simple and independent loops, most of the LAMMPS code will not easily vectorize, so it is rather meaningless to compare with benchmark numbers from compilations that have vectorization enabled. You have to look at the code in the INTEL package (and its rather convoluted code with preprocessor magic, vectorization directives and copying data to more vector friendly data structure) to see what kind of effort is required to get some meaningful vectorization out of the code.
GCC doesn't do quite as good a job as Intel in vectorization but can also dispatch to SVML.
But even without vectorization, the "native" GCC inline functions are about twice as fast as the best approximation (and with full accuracy) in my tests when using `-ffast-math`. That said, it may still be worth using the Intel compilers, since they also have their own math library, which is usually faster than with glibc offers. I've even has success with using the shared library version of it after the fact via LD_PRELOAD to speed up math heavy code (with lots of log() and exp() which are  more expensive than sin() or cos()) without recompiling.

I think the only additional thing to try out would be using the sincos() function since you need a sine and cosine of the same angle, IIRC. Toy codes have limited utility for modern compilers unless you explicitly turn off many optimizations and thus dumb them down enough so you can see numbers that are applicable to the real application at hand.

You can pass custom compiler flags to CMake for LAMMPS via `-DCMAKE_TUNE_FLAGS`

akohlmey: FWIW, I rarely look at assembly code. Most of the time after the fact, e.g. via `perf report` after using `perf record` to get profiling data. Perf can "zoom in" and then uses objdump to disassemble the object code. When compiling with -g this is often useful enough for me to find out what I want to see (despite the debug info becoming rather unreliable with high optimization), since I mostly want to see whether serial or parallel SSE/AVX instructions are used and where the "hot spots" are.

GenieTim:
As an experiment, I did a naïve AVX2 vectorization manually (see attached file). _Without any approximations_ (but with using Intel's SVML trigonometric functions), this already reduces the overhead of using the angle term to 10% e.g. for the case of the system with N = 500 This is lower than what I observed with any of the approximations. But, of course, this code is not portable. I noticed that LAMMPS provides an abstraction, namely `intel_intrinsics.h`, which I could use to make my vectorized version portable (probably turns it into a angle_style fourier/simple/intel or use some preprocessor magic). @akohlmey, do you happen to know if there is any documentation for this abstraction which I did not find yet?

[angle_fourier_simple_avx2.zip](https://github.com/lammps/lammps/files/6817731/angle_fourier_simple_avx2.zip)

akohlmey:
> @akohlmey, do you happen to know if there is any documentation for this abstraction which I did not find yet?

The only "documentation" I am aware of is the source code. For questions beyond that, you would have to contact @wmbrownIntel

akohlmey:
@GenieTim could you send a small and ready to use input deck that has the original angle style to my personal email, so that I can look at doing some real world profiling and look into what can be done as portable optimizations to the default code paths?

GenieTim:
As @akohlmey kindly hinted me in a private E-Mail, the already existing angle_style fourier/simple/omp already runs much faster than fourier/simple alone, even with OMP_NUM_THREADS=1 (only 10% overhead). While I will do some further investigations to try to understand why this is the case after all (given that I would expect 1 thread to be equivalent to my implementation without approximations), I consider the use of approximations not as urgent anymore and will close this pull request and corresponding issue. 

For anyone coming here through Google due to slow fourier/simple angle style:
Make sure to compile LAMMPS with OMP support as well as -ffast-math (i.e. with CMAKE flags `-DPKG_OPENMP -DCMAKE_TUNE_FLAGS="-O3 -ffast-math"`) and then use either the LAMMPS suffix flag (i.e. `-sf omp`) or explicitly the angle style fourier/simple/omp.

PR #2826, opened 2021-07-04
Collected small changes

### LINKED ISSUE #2820: [BUG] LAMMPS with zstd support does not compile on some Linux distributions with default libraries and traditional make
**Summary**

Compiling LAMMPS with zstd compression support fails, because the library bundled with some Linux distributions is too old.

Current master, Linux Mint 19.3 (Tricia)

**Expected Behavior**

The minimal version of the zstd library should be mentioned in the manual or source files.

**Actual Behavior**

Compilation fails, because functions in `src/COMPRESS/zstd_file_writer.cpp` use features not available in the library bundled with Linux.

**Steps to Reproduce**

1. Install zstd library: `sudo apt install libzstd-dev`
2. Modify a makefile by including `-DLAMMPS_ZSTD`
3. Try to compile LAMMPS.

EDIT: I tried compiling with cmake instead of make and there is a version requirement, so this issue only exists for traditional make.

akohlmey:
**Summary**

This pull request contains multiple small changes that each do not warrant a pull request of their own.

**Author(s)**

Axel Kohlmeyer (Temple U)

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

- fixes for the documentation to reflect the recent package rename for cases with typos or where the update was missed
- some `NULL` to `nullptr` changes that were previously missed
- avoid integer overflow issues reported by CodeQL static code analysis
- avoid single precision floating point math in angle style fourier/simple
- document the requirement of needing a zstd library version 1.4 or later. add check in zstd_writer class. closes #2820 

**Post Submission Checklist**


src/ML-IAP/mliap_so3.cpp
```diff
@@ -733,28 +733,27 @@ void MLIAP_SO3::get_sbes_array(int nlocal, int *numneighs, double **rij, int lma
...
-               (j + 1) * m_sbes_array[gindex + (i - 1) * mindex + j + 1]) /
+          m_sbes_darray[gindex + i1mindex + j] = xi *
+              (j * m_sbes_array[gindex + i1mindex + j - 1] +
+               (j + 1) * m_sbes_array[gindex + i1mindex + j + 1]) /
               (2 * j + 1);
 
-        m_sbes_darray[gindex + (i - 1) * mindex + j] = xi *
-            (j * m_sbes_array[gindex + (i - 1) * mindex + j - 1] + (j + 1) * exts) / (2 * j + 1);
-        m_sbes_darray[gindex + (i - 1) * mindex + 0] = xi * sb;
+        m_sbes_darray[gindex + (i - 1) * mindex + j] =
```
athomps, line 754: Could replace (i-1)*mindex with m1index here too?

athomps (approved):
I approve, although I noted one place where mliap_so3.cpp could be cleaned up even more.

src/ML-IAP/mliap_so3.cpp
akohlmey: Thanks. I found this and another case that must have slipped through the previous find+replace search.

PR #2827, opened 2021-07-05
CG-DNA: New Class AtomVecOxdna

ohenrich:
**Summary**

This pull request implements the new AtomVecOxdna class and fixes a problem with occasionally and temporarily frustrated stacking interactions. It stores the polarity of the nucleotide strand through per-atom information of the 5'-partner of each atom in a bond. Note that the directionality was previously worked out 'on-the-fly', which proved to be not robust enough in very flexible single-stranded overhangs.

The pull request contains as well updated benchmarks and examples, which consist now exclusively of equilibrated systems with thermal translational and rotational velocities.

**Author**

Oliver Henrich
Department of Physics
University of Strathclyde
Glasgow, UK

**Licensing**

**Backward Compatibility**

Backward compatibility is not retained in the input files, but can be easily restored by adding the keyword "oxdna" to the atom style like "atom_style hybrid bond ellipsoid oxdna".

**Implementation Notes**

The atom ID of the 5'-partner is set from the "Bonds" section in the data file. If a bond is defined as "bond_ID bond_type atom1 atom2", then atom2 is the 5' partner of atom1 and this information is stored with atom1 in a pointer **id5p defined in the Atom class.

A method data_bonds_post() has been defined in AtomVecOxdna that reprocesses the bond section of the data file and sets the id5p values accordingly. The polarity information is used in the pair styles oxdna/stk and oxrna2/stk to decide which atom lies in 3' and 5' direction.

The new implementation has been verified against the existing benchmark cases in the examples section. Note that these have been updated and all configurations in the data files are now equilibrated.

**Post Submission Checklist**

 The feature has been verified to work with the CMake based build system


ohenrich:
@akohlmey @sjplimp I created the new pull request and will add the changes you requested in PR #2821

ohenrich:
@sjplimp
> Also to be clear, the call from Atom::data_bonds() into AVOx should just pass the data for a single line read from the Bonds sections of the data file, not N line strings. And it should be called for both the newton on/off code paths, and have 5 args I think: m, num_bond[m], atom1, atom2, id_offset. The data_bond_post() method you write for AVOx can use that to set (or not) id5p. Other future atom styles could implement that method for a different purpose and they might need to know which bond for which atom that data is now stored with, i.e. m and num_bond[m].

There is a problem as my routine can only be active in the first part and needs to be silent in the second part called if newton_bond==0 even if called. Passing in the newton_bond flag is obviously not an option. 

How can I organise this with calls in both parts? A firstpass, second pass criterion reset at every processed line?

ohenrich:
@sjplimp  There is another problem: The call from Atom::data_bonds() to avec->data_bonds_post() never calls the child method in AtomVecOxdna, only the method in AtomVecHybrid. This is the problem I had all the time with calling from Atom::data_bonds(). 

Note: I introduced a new method data_bonds_post2() for demonstration purposes. It will eventually replace the other one.

ohenrich:
@sjplimp Hi Steve, I included all the requested changes. The arguments of data_bonds_post() are now also populated to the hybrid sub-styles.

src/CG-DNA/atom_vec_oxdna.h
```diff
@@ -0,0 +1,48 @@
...
+
+namespace LAMMPS_NS {
+
+class AtomVecOxdna : public AtomVec {
+ public:
+  AtomVecOxdna(class LAMMPS *);
+  ~AtomVecOxdna();
+
+  void grow_pointers();
+  virtual void data_bonds_post(int, int, tagint, tagint, tagint);
```
sjplimp, line 33: this should not be virtual

src/atom_vec_hybrid.h
```diff
@@ -55,6 +55,9 @@ class AtomVecHybrid : public AtomVec {
   void unpack_restart_init(int);
   void create_atom_post(int);
   void data_atom_post(int);
+
+  virtual void data_bonds_post(int, int, tagint, tagint, tagint);
```
sjplimp, line 59: this should not be virtual

src/CG-DNA/atom_vec_oxdna.cpp
```diff
@@ -0,0 +1,96 @@
...
+                                       tagint atom2, tagint id_offset)
+{
+  tagint *id5p = atom->id5p;
+
+  if (id_offset) {
+    atom1 += id_offset;
+    atom2 += id_offset;
+  }
+
+  if ((m = atom->map(atom1)) >= 0) {
```
sjplimp, line 93: I'm not 100% clear the logic in this if statement is correct, based on what you explained earlier.
(1)you are overwriting the passed input arg "m", which means you are ignoring it.
Even if that's OK, you should use a different variable name.
(2) Can you verity that your code still works and gives the same answers if newton bond is set to off?  I think the code here in that case is simply setting id5p[] twice for the same atom, but it would be good to verify that.  I thought you were planning to skip doing this setting on the 2nd call when newton bond off, but maybe this is simpler.

sjplimp (changes requested):

sjplimp:
Hi Oliver - I think this is nearly done.  If you can look at my couple of code annotations and verify it is working with newton bond off,
it should be good to merge.

src/CG-DNA/atom_vec_oxdna.cpp
ohenrich: @sjplimp 
Ad (1): I use now a different variable name, which is internally defined. I agree this is better as indeed the argument 'm' was ignored.

Ad(2): The code still works with newton off. During the second call to avec->data_bonds_post() in Atom::data_bonds() with newton off it simply assigns the same values id5p[n]=atom2 with n = atom->map(atom1), so indeed what you say it sets it twice. This is far easier than having a first/second pass criterion, a possibility I didn't see in the first instance.

ohenrich:
Thanks for your help! In the end it is really so easy. I just went down zillions of wrong rabbit holes. But I learnt a lot, which will come in handy.

sjplimp (approved):

sjplimp:
@ohenrich great - code all looks good now - ready to merge @akohlmey Oliver - thanks for persisting on this!  If you have new images or movies for what the extended models enable, we can post them on the website.

PR #2832, opened 2021-07-14
Revert Kokkos CPU SNAP to original code

stanmoore1:
**Summary**

After years of optimizing Kokkos SNAP for GPUs, the Kokkos CPU performance has suffered and is significantly slower than the original (non-Kokkos) CPU implementation, despite the CPU and GPU code-paths being fully bifurcated. There are future plans to refactor the Kokkos SNAP code to improve CPU performance and re-integrate CPU/GPU code to reduce bifurcation. However, as a quick, temporary measure, this PR reverts the CPU version of Kokkos SNAP to use the better performing original code so that users aren't hurt if they use the Kokkos CPU version of SNAP.

**Related Issue(s)**

None.

**Author(s)**

Stan Moore (SNL)

**Licensing**

**Backward Compatibility**

Yes


stanmoore1:

stanmoore1:
@akohlmey I don't think 54cf985280b18b535cff8633528d4a9db2100a99 should be made a warning. I could see this causing users to turn off threading to get rid of a warning when that would hurt say pair zbl/kk performance.

sjplimp (approved):

athomps (approved):
I approve.

PR #2835, opened 2021-07-16
Documentation: CG-DNA package

ohenrich:
**Summary**

A warning has been added to the documentation of the bond style oxdna/fene, oxdna2/fene and oxrna2/fene. The write_data command does not necessarily write the bond partners in the same as they are defined in an initial data or restart file. This pull request contains as well a new test case of a double-stranded DNA ring.

**Author(s)**

Oliver Henrich
Department of Physics
University of Strathclyde
Glasgow, UK

**Licensing**

**Backward Compatibility**

No backward compatibility is broken.

**Post Submission Checklist**


akohlmey:
@ohenrich How about adding an optional(!) OxDNA section to the data file?
Upon write_data, this section will be written and has for each bond the stored value.
Upon read_data, the code checks if the section is present and if yes, reads the data from there, otherwise infers it from the Bonds section as currently done.

ohenrich:
This sounds like the right idea. But I have no time at the moment to develop this for this upcoming stable release. How about aiming for this for the winter/spring release?

sjplimp:
Unclear about a couple things here.
(a) Why is the write_data Bonds section not the same as the original read_data Bonds section?  I'm looking at AtomVec::pack_bonds() and don't see the issue.  Is this something to do with newton bond on vs off?
(b) If AV::pack_bond() is not doing the right thing, its possible you could override it with a AVOxdna::pack_bonds() to do the right thing.
(c) If you go Axel's suggested route and add a section to the data file, then note you do not need per-bond info.  Your new id5p is a per-atom vector, not per-bond.  So you could actually discard your new atom class, and define id5p with fix property/atom, which has an option to read it from the data file.  It will then also then be written by write_data.  But that requires any user pre-process their original data file to include those values.  I thought you were trying to avoid that with your new atom style.

sjplimp:
One more thought on (c) in the previous comment.  If you want to use fix property/atom instead of your new atom sytel oxdna
to create id5p, then you could add a simple python script to your package (or better in lammps/tools).  Which would use the python data.py module which can read/write LAMMPS files.  The script could simply read the user's original data file, scan the Bonds section to create the per-atom id5p values, and add that as a new section to a new data file it outputs.  The logic for creating id5p values would be what you encoded in the data_bonds_post() method in your oxdna atom style.   So running this python script would be a pre-processing step for the user.  Would run quickly for anything but humongous data files.

Also I agree this is future work, whatever needs to be added now for a warning/error is fine.

ohenrich:
> Unclear about a couple things here. 
> (a) Why is the write_data Bonds section not the same as the original read_data Bonds section? I'm looking at AtomVec::pack_bonds() and don't see the issue. Is this something to do with newton bond on vs off?

No, this is because LAMMPS does not keep a strict record of how the bond was defined in the data file (atomID1 atomID2 or atomID2 atomID1). This is the very reason why the new atom style was necessary.

For instance a bond that closes a ring defined as '666 1 1 74' is the same as '666 1 74 1'. write_data simply writes all IDs in sequence, so for example 
...
664 1 72 73
665 1 73 74
**666 1 74 1**
667 1 1 2
668 1 2 3
...
in the initial data file becomes
...
664 1 72 73
665 1 73 74
666 1 1 2
**667 1 1 74**
668 1 2 3
...
with write_data, which means not as they may have been set in the initial data file.

> (b) If AV::pack_bond() is not doing the right thing, its possible you could override it with a AVOxdna::pack_bonds() to do the right thing.

Not sure this is possible because of the above mentioned limitation. There is simply no concept of bond directionality.

> (c) If you go Axel's suggested route and add a section to the data file, then note you do not need per-bond info. Your new id5p is a per-atom vector, not per-bond. So you could actually discard your new atom class, and define id5p with fix property/atom, which has an option to read it from the data file. It will then also then be written by write_data. But that requires any user pre-process their original data file to include those values. I thought you were trying to avoid that with your new atom style.

I forgot this means breaking backward compatibility! We don't want to do this, at least not at the moment. We have quite a few scripts and routines that would all require a rewrite on top of the inconvenience it causes for the users. We may go down that route for oxDNA3 as it would be the right occasion to iron out all suboptimal choices I made when I started working with LAMMPS.

> One more thought on (c) in the previous comment. If you want to use fix property/atom instead of your new atom sytel oxdna to create id5p, then you could add a simple python script to your package (or better in lammps/tools). Which would use the python data.py module which can read/write LAMMPS files. The script could simply read the user's original data file, scan the Bonds section to create the per-atom id5p values, and add that as a new section to a new data file it outputs. The logic for creating id5p values would be what you encoded in the data_bonds_post() method in your oxdna atom style. So running this python script would be a pre-processing step for the user. Would run quickly for anything but humongous data files.

I had the idea of a script as well that checks for differences between the original data file and what write_data produces. The thing is this problem really only occurs if you write a data file. Normally, you would write restart files as they contain more information and are more accurate. A data file is only handy if you want to change the topology during two consecutive runs.

I will try to produce such a script if I have the time and add it somewhere as I see you extended the DL for the next stable release to 5th Aug.

> Also I agree this is future work, whatever needs to be added now for a warning/error is fine.

Sigh ... And thanks.

sjplimp:
@ohenrich All makes sense - thanks for the details.  Just want to belabor my first Q a bit more.  For the data file change you illustrated: 1 74 1 --> 1 1 74.  I see in the code why that happens when newton bond off at line 2065 of AtomVec::pack_bond().
The bond is stored twice and the code has to choose only one of them to write to the data file.  But I don't why that ordering
flip will happen for newton bond on.  Am I right about that?  If so, does your package require newton bond off for some other reason?  If not, you could just output your warning or error when newton bond is off ?

ohenrich:
@sjplimp I think it's a good idea to throw out a warning message for the newton off setting. No, newton off is not required. I just want to give users the freedom to minimise communication overheads.

All of my tests have newton off as it maximises numerical roundoff errors and tests different parts of the code, so is more complex. I will rerun the tests with newton on, resubmit them with this pull request and include the changed code that implements the warning message.

I will also edit the warning in the docs to reflect this change and try to provide a script that corrects data files written with newton off, if possible for the next stable release.

ohenrich:
@sjplimp I updated the input files and tests to have now the setting 'newton on', added code to throw a warning message and updated the documentation. I added also a pointer name in to cpp files, which was obviously missing following Axel's last reformatting and caused seg faults.

sjplimp (approved):

sjplimp:
@ohenrich Looks good - thanks.  Axel, this is ready to merge from my perspective.

PR #2837, opened 2021-07-16
Update and document support for fix external in the library interface and python module

### LINKED ISSUE #2834: [Feature Request] lammps_fix_external_set_virial_global python interface
**Summary**

LAMMPS coupling with other code is possible by "fix external" mechanism.
Using C library API it is possible to  virial (stress tensor) which is needed for NPT simulations.
But I couldn't find similar functionality using python API. 

**Detailed Description**

The possibility for setting virial in addition to external force is requested.
See the pseudo code below.

```python
from lammps import lammps

def callback(caller, ntimestep, nlocal, tag, pos, fext):
      # set fext
      fext[:] = ...

      # requested functions:
      caller.set_fix_external_energy_global(...)   #  <- requested feature
      caller.set_fix_external_virial_global(...)      # <- requested feature

lmp = lammps() 
# ...
lmp.command("fix 1 all external pf/callback 1 1")
lmp.set_fix_external_callback("1", callback)
# ...
```

file:
lammps/python/lammps/core.py

lines:
299    self.lib.lammps_fix_external_set_energy_global = ...
300    self.lib.lammps_fix_external_set_virial_global = ...

akohlmey:
**Summary**

This pull request updates the support for using fix external from the C-library interface and the Python module.
This now supports all functionality described in the fix external documentation. Also unit tests for both the C-library
and the Python module are added.

**Related Issue(s)**

Fixes #2834 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

New functionality:
- `extract()` method added to `FixExternal` for simple access to its `fexternal` array
- `lammps_fix_external_get_force()` function added to support the *fp/array* mode of fix external in addition to *pf/callback*.
- New library interface functions for per-atom energy and virial
- New library interface functions for `FixExternal::set_vector()` and  `FixExternal::set_vector_length()`
- Python module interface for all `lammps_fix_external_get/set*()` functions
- Unit tests for all updated and new functionality

**Post Submission Checklist**


akohlmey:
@sjplimp I noticed a few things when writing the tests and documenting the functions.
1) there is an inconsistency for handling the "global" values: only the global energy is copied into variable in the fix and thus retained for future timesteps. thus only this function is compatible with the "pf/array" mode; the global virial and the per-atom values can only be set from the callback function since they are copied directly into the corresponding accumulator storage (which is cleared in every time step). It would be more consistent, when setting the global virial would be persistent, same as the global energy.
2) when setting the per-atom values, no check is made, whether they are consistent with the global values. moreover, if both should be set, both functions must be called. how about the per-atom functions would do a reduction and thus also set the global values? that would make the numbers consistent and avoid additional function calls.

akohlmey:
@rbberger do you have an idea for how to best do the Python interface for the per-atom data?
Otherwise, I an pondering to leave it unimplemented (just add a FIXME comment) and wait for an application.

amirhajibabaei:
I ran into some issues while testing.
Consider the following:
```python
# ...
vir = np.array([vxx, vyy, vzz, vxy, vxz, vyz])
lmp.fix_external_set_virial_global('1', vir)
lmp.command("thermo_style custom pxx pyy pzz pxy pxz pyz")
# ...
```
Shouldn't the thermo output `press = [pxx, pyy, pzz, pxy, pxz, pyz]` satisfy `press = vir + kinetic-term`?
I couldn't verify this relation.
I suspect there might be an issue with unit conversion, 1/3 denominator, etc.
I tried passing `vir` both in pressure units and force/distance^2 units, but none of them worked.
I was testing with `units = metal` setup.
Let me know if my testing code was needed for checking.

akohlmey:
> I ran into some issues while testing.
> Consider the following:
> 
> ```python
> # ...
> vir = np.array([vxx, vyy, vzz, vxy, vxz, vyz])
> lmp.fix_external_set_virial_global('1', vir)
> lmp.command("thermo_style custom pxx pyy pzz pxy pxz pyz")
> # ...
> ```
> 
> Shouldn't the thermo output `press = [pxx, pyy, pzz, pxy, pxz, pyz]` satisfy `press = vir + kinetic-term`?
> I couldn't verify this relation.
> I suspect there might be an issue with unit conversion, 1/3 denominator, etc.
> I tried passing `vir` both in pressure units and force/distance^2 units, but none of them worked.
> I was testing with `units = metal` setup.
> Let me know if my testing code was needed for checking.

Where does it say that the second argument to `lmp.fix_external_set_virial_global('1', vir)` is a numpy array?

amirhajibabaei:
> > I ran into some issues while testing.
> > Consider the following:
> > ```python
> > # ...
> > vir = np.array([vxx, vyy, vzz, vxy, vxz, vyz])
> > lmp.fix_external_set_virial_global('1', vir)
> > lmp.command("thermo_style custom pxx pyy pzz pxy pxz pyz")
> > # ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Shouldn't the thermo output `press = [pxx, pyy, pzz, pxy, pxz, pyz]` satisfy `press = vir + kinetic-term`?
> > I couldn't verify this relation.
> > I suspect there might be an issue with unit conversion, 1/3 denominator, etc.
> > I tried passing `vir` both in pressure units and force/distance^2 units, but none of them worked.
> > I was testing with `units = metal` setup.
> > Let me know if my testing code was needed for checking.
> 
> Where does it say that the second argument to `lmp.fix_external_set_virial_global('1', vir)` is a numpy array?

I tried `list` instead of `np.array` and it didn't make any difference.

akohlmey:
Well, it works in the test example: https://github.com/lammps/lammps/blob/1c4e8aba122e5d792788464bf5f88988aa91f348/unittest/python/python-fix-external.py#L7

amirhajibabaei:
> Well, it works in the test example:
> 
> https://github.com/lammps/lammps/blob/1c4e8aba122e5d792788464bf5f88988aa91f348/unittest/python/python-fix-external.py#L7

From the unittests it is clear that external virial is internally divided by volume which isn't clear from the docs.
Aside from that, in units other than `lj`, there is another conversion which is not clear to me.
See an example which I have prepared for `metal` units [here](https://github.com/amirhajibabaei/my_lammps_tests/tree/master/python/fix_external_metal).
Therefore I think a comment about the units of virial is needed in docs.
For energy, it is mentioned that the input should be in energy units, but not for virial.

akohlmey:
@amirhajibabaei Pressure is: `-stress/vol`. This is the way how the virial stress is handled consistently throughout LAMMPS.  Check out https://docs.lammps.org/compute_pressure.html and https://docs.lammps.org/compute_stress_atom.html
So what is implemented is not at all unexpected.

The documentation is still work in progress after all this is still a draft pull request, so thanks for the suggestion.

rbberger (approved):
Looks good

akohlmey:
> Looks good

Just added some documentation fixes and managed to get the numpy wrapper for per-atom energy. Could not figure out how to do the same with the virial, though.

akohlmey:
@amirhajibabaei this has now the desired consistent behavior and updated documentation. To avoid inconsistencies most of the details for how these functions are supposed to work are kept with the C-library function calls and only differences between C and Python are explained.

akohlmey:
@rbberger some more digging through stackoverflow and the mystery of passing a 2d-array via ctypes has been resolved (although it looks *very* hackish). Added some dimension checks, too.

PR #2839, opened 2021-07-18
Fixed bug in fix np[ht] with runstyle respa

craabreu:
**Summary**

Fixes a bug in r-RESPA integration involving volume changes and kspace calculations.

The bug shows up in the following situation:
* Pair stye involves kspace calculations
* MD simulation is isobaric via any `FixNH` child class with `pstat_flag=1`
* r-RESPA integration with at least 2 time scales

In this specific situation, update of kspace parameters is being done before volume changes actually occur.

At the outermost time scale, `initial_integrate_respa` is called from inside the `recurse` function in `respa.cpp`:

https://github.com/lammps/lammps/blob/95792ac9283347d9d5ebca875eb2203a4bac43dd/src/respa.cpp#L609

Then, after all due calculations, kspace parameters are updated:

https://github.com/lammps/lammps/blob/95792ac9283347d9d5ebca875eb2203a4bac43dd/src/fix_nh.cpp#L1013-L1014

Upon return to `recurse`, and before the slowest forces (including the kspace contribution) are computed, a recursive call is made:

https://github.com/lammps/lammps/blob/95792ac9283347d9d5ebca875eb2203a4bac43dd/src/respa.cpp#L669

Because volume will change at the innermost time scale, kspace parameters will be outdated when the slowest forces are actually evaluated.

**Related Issue(s)**

None.

**Author(s)**

Charlles Abreu, Federal University of Rio de Janeiro

**Licensing**

**Backward Compatibility**

Not backward compatible.

**Implementation Notes**

The bug was solved by moving the lines below from `initial_integrate_respa` to a newly implemented `pre_force_respa` function.

https://github.com/lammps/lammps/blob/95792ac9283347d9d5ebca875eb2203a4bac43dd/src/fix_nh.cpp#L1013-L1014

**Post Submission Checklist**

None.

### DIFF
--- a/src/BOCS/fix_bocs.cpp
+++ b/src/BOCS/fix_bocs.cpp
@@ -1100,7 +1100,12 @@ void FixBocs::initial_integrate_respa(int /*vflag*/, int ilevel, int /*iloop*/)
     nve_x();
     if (pstat_flag) remap();
   }
+}
 
+/* ---------------------------------------------------------------------- */
+
+void FixBocs::pre_force_respa(int /*vflag*/, int ilevel, int /*iloop*/)
+{
   // if barostat, redo KSpace coeffs at outermost level,
   // since volume has changed
 
--- a/src/BOCS/fix_bocs.h
+++ b/src/BOCS/fix_bocs.h
@@ -37,6 +37,7 @@ class FixBocs : public Fix {
   virtual void initial_integrate(int);
   virtual void final_integrate();
   void initial_integrate_respa(int, int, int);
+  void pre_force_respa(int, int, int);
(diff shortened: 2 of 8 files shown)


akohlmey:
@craabreu Thanks for your analysis and fix. There are multiple variants of fix_nh.cpp in the distribution. From a quick check it looks like `src/UEF/fix_nh_uef.cpp` may need the same change.

craabreu:
@akohlmey It looks like `FixNHUef` will correctly inherit the modifications from `FixNH`.

I was able to find three other classes that required changes, namely `FixBocs`, `FixTGNHDrude`, and `FixNPTCauchy`.

PR #2853, opened 2021-07-27
Fix bug in verlet_kokkos.cpp with GRANULAR package

stanmoore1:
**Summary**

The Kokkos version of Verlet was missing the `modify->post_neighbor()` call, which creates a bug for the GRANULAR package. This PR syncs `verlet_kokkos.cpp` with `verlet.cpp`.

**Related Issue(s)**

None.

**Author(s)**

Stan Moore (SNL), reported by Dan Bolintineanu (SNL)

**Licensing**

**Backward Compatibility**

Yes


stanmoore1:
@akohlmey @sjplimp Just a reminder, can we make sure when there are changes to `verlet.cpp` that they don't get merged until Kokkos verlet is updated as well?

src/KOKKOS/verlet_kokkos.cpp
```diff
@@ -140,11 +141,10 @@ void VerletKokkos::setup(int flag)
     atomKK->sync(force->pair->execution_space,force->pair->datamask_read);
     force->pair->compute(eflag,vflag);
     atomKK->modified(force->pair->execution_space,force->pair->datamask_modify);
-    timer->stamp(Timer::PAIR);
   }
   else if (force->pair) force->pair->compute_dummy(eflag,vflag);
 
-  if (atomKK->molecular) {
+  if (atomKK->molecular != Atomic::ATOMIC) {
```
rbberger, line 147: @stanmoore1 Probably should be:
```suggestion
if (atomKK->molecular != Atom::ATOMIC) {
```

stanmoore1: @rbberger True, thanks!

sjplimp (approved):

PR #2867, opened 2021-08-06
Enabled newton pair on for gpu pair styles

ndtrung81:
**Summary**

Currently newton pair is required to be off for all the gpu styles. This PR attempts to allow newton pair on for most use cases. The requirement of newton pair off is only applicable to _particle_split < 1, that is, when some local atom forces are left on the CPU side. 

Because the changes in this PR are involving (although in a simple way -- commenting out the newton pair off requirement in all the gpu pair styles in src/GPU/, and removing a single line in FixGPU), regression tests with gpu pair styles with newton pair on/off will be needed. For double precision my tests passed with lj, eam and hybrid lj/cut coul/cut. Will need help from the core developers.

**Related Issue(s)**

N/A

**Author(s)**

Trung Nguyen (Northwestern)

**Licensing**

**Backward Compatibility**

Supposed to maintain.

**Implementation Notes**

By commenting out a single line in fix_gpu.cpp in FixGPU::post_force():

if (force->pair->vflag_fdotr) force->pair->virial_fdotr_compute();

we remove the double count in the virial caused by force->pair->virial_fdotr_compute() when force->pair->vflag_fdotr is set (when newton pair is on). In the lines above this line, the virial copied back from the GPU are already tallied to the force->pair->virial. More explanation for the logic is given in the comment in fix_gpu.cpp (will be removed if all the regression tests passed).

This fix may work because FixGPU::post_force() is called after comm->reverse_comm(), and because the forces from reverse_comm() are not conflicting with those by cpu_compute() in the GPU pair styles by enforcing _particle_split == 1 (to ensure no call to cpu_compute()). 

Then in all the pair_*_gpu.cpp, we remove all the requirement of newton pair to be off. FixGPU will error out if newton pair is on and _particle_split < 1.

**Post Submission Checklist**


sjplimp (approved):

akohlmey:
@ndtrung81 I will look into updating the unit tests.

wmbrownIntel:
@ndtrung81 - So is the benefit here for some simulations where newton must be on for CPU calculations? Otherwise, we are just adding in an unnecessary comm step, right? I'm also a little confused about the fdotr restrictions working here.

ndtrung81:
@wmbrownIntel Yes, you are correct: The benefit is for the cases where 1) using pair hybrid with some sub-styles (not yet supported by the GPU package) need newton pair on, and where 2) some parts in the compute() function of a pair style reside on the CPU need to have newton pair on. 

For the other cases (all the pair styles in a simulation are fully supported by the GPU package), we just use newton pair off as usual to avoid the unnecessary reverse comm.

For _particle_split < 1, the gpu pair style invokes cpu_compute() which uses ev_tally_full() to sum the virial contribution due to the local atom forces residing on the CPU. The function ev_tally_full() is compatible with full neighbor lists when newton pair is off. The restriction of _particle_split = 1 when newton pair is on is a lazy workaround from me. Please advise if there are better solutions.

ndtrung81:
@wmbrownIntel I think a possible solution to enable newton_pair on for particle_split < 1 is to follow the approach described in your CPC 2015 paper with the INTEL package, where the cpu_compute() function loops through all the local atoms, but for those in the GPU only the forces on the ghost neighbors are computed, while for those in the GPU the atom forces are accumulated from local neighbors. This approach is of course involving, and needs deserving effort. Besides, the benefit of using particle_slit < 1 and dynamic host-device load balancing is not as clear these days as it was before.

With the commit 904a2ef, newton_pair is off by default to ensure backward compatibility, so that the current input scripts and tests work the way they do before the PR. If users intentionally want to turn on newton_pair, they can specify "newton on" in the package gpu command. So far, I have tested with lj, beck, rhodo, gayberne, dipole, eam, sw, and tersoff, all give the same answers as without the changes in this PR.

If accepted, this PR will benefit a host of practical simulations where newton_pair on is required for non-GPU sub-styles. Please let me know if you have questions or suggestions to improve the PR. Thanks!

wmbrownIntel:
@ndtrung81 Thanks Trung for the contribution. Rather than working on particle_split < 1, I think a more interesting direction would be to specialize with templates/JIT to use atomics and eliminate redundant computation.

For this PR, we can probably just remove the newton checks from the pair files rather than commenting, right? And keep the Makefile mixed precision?

ndtrung81:
@wmbrownIntel Thanks Mike! I agree that we can set aside particle_split for now. Your suggestions are interesting indeed, and worth exploring.

For this PR, you’re correct that we just need to remove the checks in the gpu pair styles (not just commenting out). The Makefile cuda needs to revert to mixed precision. I will do these changes before switching to the ready for review mode.

wmbrownIntel (approved):

ndtrung81:
@akohlmey @sjplimp and @rbberger the PR is now ready for review.

sjplimp (approved):

PR #2869, opened 2021-08-07
Have add_fix(), replace_fix() and add_compute() in Modify return class instance and refactor/simplify code

akohlmey:
**Summary**

This replaces some code construct that seems to be inherited from the Fortran version of LAMMPS with a more C-style equivalent.  There are several cases where fixes or computes are created and the creating code needs to keep a reference
to the created class instance. This is currently done by assuming new instances are always added at the end and by accessing the internal storage in the Modify class. Instead, it would be simpler to have the add/replace functions directly return a pointer to the new class instance.

This pull request implements this change and also updates the existing code to take advantage of it.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

Yes. This changes functions that previously returned void to returning a pointer, so for all old usages the return value will be ignored like before.

**Implementation Notes**

This also simplifies/modernizes a few related code segments to use improved or more convenient APIs.

**Post Submission Checklist**


src/KOKKOS/pair_gran_hooke_history_kokkos.cpp
```diff
@@ -67,20 +68,13 @@ void PairGranHookeHistoryKokkos<DeviceType>::init_style()
...
   if (history && fix_history == nullptr) {
-    char dnumstr[16];
-    sprintf(dnumstr,"%d",3);
-    char **fixarg = new char*[4];
-    fixarg[0] = (char *) "NEIGH_HISTORY_HH";
-    fixarg[1] = (char *) "all";
+    auto cmd = std::string("NEIGH_HISTORY_HH") + std::to_string(instance_me) + " all ";
     if (execution_space == Device)
-      fixarg[2] = (char *) "NEIGH_HISTORY/KK/DEVICE";
+      cmd += "NEIGH_HISTORY/KK/DEVICE 3";
```
stanmoore1, line 73: What does the "3" mean?

sjplimp (approved):

stanmoore1:
@akohlmey it would be great to merge this soon since it includes bugfixes: fixes a compile issue with the GPU package as well as a bug that was introduced into the Kokkos package. Thanks

PR #2872, opened 2021-08-08
Update pair_table.cpp

ydsumith:
The table cutoff error was not well documented and user have to refer code. The updated error message will help them to understand the rcut requirements.

**Summary**

**Related Issue(s)**

**Author(s)**

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/src/pair_table.cpp
+++ b/src/pair_table.cpp
@@ -292,8 +292,8 @@ void PairTable::coeff(int narg, char **arg)
     rhi = tb->rhi;
   }
   if (tb->cut <= rlo || tb->cut > rhi)
-    error->all(FLERR,"Invalid pair table cutoff: check if rlo < rcut <= rhi");
-  if (rlo <= 0.0) error->all(FLERR,"Invalid pair table cutoff: rlo cannot be zero or negative");
+    error->all(FLERR,"Invalid pair table cutoff");
+  if (tb->cut <= 0.0) error->all(FLERR,"Invalid pair table cutoff");
 
   // match = 1 if don't need to spline read-in tables
   // this is only the case if r values needed by final tables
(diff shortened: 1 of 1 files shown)


akohlmey:
@ydsumith I don't see any changes updating the documentation here. I agree the documentation is lacking some detail.
I don't agree though, that the changes to the error messages are resolving the issue.

The first error message does not need to be updated since "Invalid pair table cutoff" is exactly what is wrong for that case.
The second error message, however, should be updated to something like "Invalid pair table range" or "Invalid pair table data". The detailed explanation of what the requirements for both cutoff and rlo/rhi etc. should be given in doc/src/pair_table.rst (and they *are* given for the most part). An additional short explanation/recommendation for the error messages could be placed in https://docs.lammps.org/Errors_messages.html (which used to be created from post-processing .h files, e.g. src/pair_table.h in this case).

ydsumith:
@akohlmey I agree with your comments. I have updated the pull request with rlo replaced with rcut. I don't see a point to enforce rlo <0, that must be a typo and should be rcut < 0. I will update the documentation shortly.

akohlmey:
> @akohlmey I agree with your comments. I have updated the pull request with rlo replaced with rcut. I don't see a point to enforce rlo > 0, that must be a typo and should be rcut > 0. I will update the documentation shortly.

No, that is definitely **not** a typo. `rlo` **must** be larger than zero or else you can get a division by zero. Please note that the documentation recommends that `rlo` is chosen as large as possible to avoid instabilities when computing splines. That also implies that rlo > 0. Most functional forms for pair potentials have a divergence at `r=0` anyway. Checking for a cutoff larger than 0 is pointless as using 0 would merely turn off all interactions, which is not an error the code needs to check for.  r cannot be < 0.

Please note that the code for pair style table has been in LAMMPS a very, very long time and most of the functionality has been unchanged or only seen minor/cosmetic changes since its original inception. So it is not likely to have obvious typos in there.

akohlmey:
@ydsumith please also fill out the pull request template form properly using the commented out text as a guide. We cannot accept a pull request otherwise for formal reasons.

akohlmey:
superseded by PR #2896

PR #2875, opened 2021-08-09
Change library interface to allow compiling DLL version with MSVC

akohlmey:
**Summary**

We can build shared libraries (aka DLL files) for Windows currently with CMake and MinGW, but that is not following the official protocol defined by Microsoft for exporting a library interface for a shared library. This pull request aims to provide such an interface and should be a(nother) step for making LAMMPS fully compatible with the native visual studio compilers on Windows.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues. Should have no effect on non-Windows platforms. Uses a more consistent export of C-style APIs.

**Post Submission Checklist**

### DIFF
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -16,6 +16,7 @@
 // See the manual for detailed documentation.
 
 #define LAMMPS_LIB_MPI 1
+#define _LAMMPS_BUILD_LIB 1
 #include "library.h"
 #include <mpi.h>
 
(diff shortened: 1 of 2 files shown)


akohlmey:
unfortunately, this change breaks our use of doxygen :-(

junghans:
Would the `PREDEFINED` option in doxygen help?

Something like:
```
PREDEFINED           = "LAMMPS_EXPORT_API=_LAMMPS_DLL_FUNC"
```

PR #2881, opened 2021-08-17
Fix issue when building Kokkos with CMake

stanmoore1:
**Summary**

CMake wasn't exporting the `LMP_KOKKOS` define or `/src/KOKKOS` headers to `main.cpp`, so the "dummy" version of Kokkos in `accelerator_kokkos.h` was always being instantiated in `main.cpp`. Therefore, the real `Kokkos::finalize` was never called here: https://github.com/lammps/lammps/blob/master/src/main.cpp#L106, leading to an error from Kokkos and other issues with Kokkos tools.

@junghans @anjohan

**Related Issue(s)**

None

**Author(s)**

Christoph Junghans (LANL), Stan Moore (SNL), reported by Anders Johansson (Harvard)

**Licensing**

**Backward Compatibility**

Yes

### DIFF
--- a/cmake/Modules/Packages/KOKKOS.cmake
+++ b/cmake/Modules/Packages/KOKKOS.cmake
@@ -74,7 +74,7 @@ else()
   target_link_libraries(lammps PRIVATE kokkos)
   target_link_libraries(lmp PRIVATE kokkos)
 endif()
-target_compile_definitions(lammps PRIVATE -DLMP_KOKKOS)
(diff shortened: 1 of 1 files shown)


junghans:
`LMP_KOKKOS` is not used in library headers, so let's only use it in the build_interface.

junghans (approved):

stanmoore1:
@junghans thanks, I confirm this still works.

akohlmey (approved):

PR #2882, opened 2021-08-18
Add Kokkos accelerated styles for shear flow simulations (SLLOD)

emilyviolet:
**Summary**
This set of patches introduces two Kokkos accelerated styles, `fix nvt/sllod/kk` and `compute temp/deform/kk`, and modifies Kokkos accelerated PPPM to support triclinic geometries. These modifications allow for simulations of constant-temperature shear-flow, via the SLLOD equations of motion and tilting periodic boundary conditions, to be conducted almost entirely on
the device for simple systems. The new accelerated styles are enabled in the usual fashion by adding via the `/kk` suffix, e.g.

```
...
fix   1 all nvt/sllod/kk temp 1.0 1.0 1.0 tchain 1
fix   2 all deform/kk  1 xy erate 0.01 remap v
compute mytemp all temp/deform/kk
...
```

(I have not made any changes to the already existing `fix deform/kk`).

**Author(s)**
Emily Kahl (The University of Queensland)

**Licensing**

**Backward Compatibility**

Yes.

**Implementation Notes**
- `PPPMKokkos` - Implementation of triclinic calculations hews closely to the non-Kokkos accelerated triclinic routines. I have tried to keep kernel structure and range policies consistent with the rest of the class.
- `FixNVTSllodKokkos` - Implementation is almost identical to the non-accelerated class, with some slight loop refactoring to better work as a Kokkos kernel.
- `ComputeTempDeformKokkos` - I have only implemented overloaded versions of `compute_vector()`, `compute_scalar()` and `remove/restore_bias_all()`. Also made some tweaks to the base `ComputeTempDeform` destructor and `init()` functions to properly handle the derived Kokkos style.
- `ComputeTempKokkos` - moved `s_CTEMP` out of the toplevel LAMMPS namespace and into the class namespace to avoid collisions with `ComputeTempDeform`. This was the easiest way to get it to compile cleanly with the new classes, but I'm not wedded to this solution and am happy to change if there's a better way.

I have tested the new code by manually comparing the results with serial calculations for a simple LJ-fluid, crystalline NaCl, and SPCE water and found it to agree to within numerical uncertainty for both CPU multithreading and GPU offloading. I am happy to provide the input files for these tests, and to run any others if required. I have not written unit tests as there does not yet appear to be full support for Kokkos styles in the test suite.

**Post Submission Checklist**

### DIFF
--- a/doc/src/Commands_compute.rst
+++ b/doc/src/Commands_compute.rst
@@ -152,7 +152,7 @@ KOKKOS, o = OPENMP, t = OPT.
    * :doc:`temp/chunk <compute_temp_chunk>`
    * :doc:`temp/com <compute_temp_com>`
    * :doc:`temp/cs <compute_temp_cs>`
-   * :doc:`temp/deform <compute_temp_deform>`
+   * :doc:`temp/deform (k) <compute_temp_deform>`
    * :doc:`temp/deform/eff <compute_temp_deform_eff>`
    * :doc:`temp/drude <compute_temp_drude>`
    * :doc:`temp/eff <compute_temp_eff>`
--- a/doc/src/Commands_fix.rst
+++ b/doc/src/Commands_fix.rst
@@ -148,7 +148,7 @@ OPT.
    * :doc:`nvt/body <fix_nvt_body>`
    * :doc:`nvt/eff <fix_nh_eff>`
    * :doc:`nvt/manifold/rattle <fix_nvt_manifold_rattle>`
-   * :doc:`nvt/sllod (io) <fix_nvt_sllod>`
+   * :doc:`nvt/sllod (iko) <fix_nvt_sllod>`
    * :doc:`nvt/sllod/eff <fix_nvt_sllod_eff>`
    * :doc:`nvt/sphere (o) <fix_nvt_sphere>`
    * :doc:`nvt/uef <fix_nh_uef>`
--- a/doc/src/compute_temp_deform.rst
+++ b/doc/src/compute_temp_deform.rst
@@ -1,7 +1,10 @@
 .. index:: compute temp/deform
+.. index:: compute temp/deform/kk
 
 compute temp/deform command
 ===========================
+
+Accelerator Variants: *temp/deform/kk*
 
 Syntax
 """"""
--- a/doc/src/fix_nvt_sllod.rst
+++ b/doc/src/fix_nvt_sllod.rst
@@ -1,11 +1,12 @@
 .. index:: fix nvt/sllod
 .. index:: fix nvt/sllod/intel
 .. index:: fix nvt/sllod/omp
+.. index:: fix nvt/sllod/kk
 
 fix nvt/sllod command
 =====================
 
-Accelerator Variants: *nvt/sllod/intel*, *nvt/sllod/omp*
+Accelerator Variants: *nvt/sllod/intel*, *nvt/sllod/omp*, *nvt/sllod/kk*
 
 Syntax
 """"""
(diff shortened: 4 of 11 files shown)


akohlmey:
@emilyviolet thank you very much for this nice looking contribution. I have a few minor modifications that I would like to have included before @stanmoore1 will review and test your code. Since your institutional GitHub account does not allow pushing into its pull request branches directly, I have thus created a [pull request](https://github.com/CTCMS-UQ/lammps/pull/1) that - when merged by you - should update *this* pull request accordingly.

stanmoore1:
@emilyviolet I agree, thank you for this significant contribution!

emilyviolet:
@akohlmey @stanmoore1 Thank you for that! I've merged the pull request and it looks like it's updated this one appropriately. Sorry about the runaround, I'll see if I can get the permissions fixed when I get to work (if only so any future contributions go more smoothly).

stanmoore1:
I ran this PR through Kokkos regression testing and all the existing tests passed. I will do a bit more testing of the new features.

stanmoore1:
While testing this PR I found what looks like a small bug in `remap v` not related to the changes here, looking more into that.

stanmoore1:
@akohlmey we need to add the new styles to `/src/KOKKOS/Install.sh`. I also need to test Kokkos triclinic PPPM real quick. The Kokkos SLLOD works as expected.

emilyviolet:
@stanmoore1 My bad, I should've checked the Make-based build system more thoroughly when I first tested this PR; I've made a change to (hopefully) address this. I wasn't sure if I should do this myself or leave it to you, but I figured it's one less step for me to do it since I have push-access to the fork.

stanmoore1:
@emilyviolet no worries, looks great.

@akohlmey do we have any unit tests for triclinic PPPM? If not, we should add one: the trivial test is to skew a multiple of box length, which should give the same energy as unskewed.

akohlmey:
> @akohlmey do we have any unit tests for triclinic PPPM? If not, we should add one: the trivial test is to skew a multiple of box length, which should give the same energy as unskewed.

We do have a test at https://github.com/lammps/lammps/blob/master/unittest/force-styles/tests/kspace-pppm_tri.yaml
But those tests are - despite the denomination unit test - more like regression test as they store the reference result and then have to reproduce it with all supported suffix variants and with multiple settings (newton on/off, pair_modify nofdotr etc.) (see the "skip_tests" keyword). The comparison that you describe would currently by a manual one, i. e. one would create two different inputs and then compare the reference files. I have done that for the existing pppm-tri kspace test, since it uses an untilted box. however, the grid is slightly different, so the numbers do not match to the high degree the force-style testers operate.

akohlmey:
@emilyviolet it look like I just created a new issue for you to resolve: We just added some whitespace corrections and finalized the integration of clang-format processing. In that process, a merge conflict has show up, so you should merge with the upstream master version and resolve the conflict. it seems to be just one file though.
Please see:
https://docs.lammps.org/latest/Build_development.html#coding-style-utilities
and:
https://docs.lammps.org/latest/Build_development.html#clang-format-support
for the documentation related to those changes.

akohlmey:
@emilyviolet for your convenience, I've submitted another pull request that will update this branch

emilyviolet:
@akohlmey That was indeed convenient, thank you!

stanmoore1 (approved):

stanmoore1:
@emilyviolet thanks again for this nice contribution. @akohlmey this is ready to merge from my point of view.

PR #2885, opened 2021-08-18
Grid adjust

sjplimp:
**Summary**

Alters the way grid points are assigned to processors when a global grid is partitioned across procs.  A new Comm::partition_grid() method was added so other classes which define global grids can do this consistently.  For now, this is PPPM, PPPMDisp, MSM.  The previous method could assign grid points slightly outside a proc sub-domain to the proc which is slightly inefficient.  When global grids are coarser than the proc grid (in a dimension), the previous method was even more inefficient.

**Related Issue(s)**

N/A

**Author(s)**

Steve

**Licensing**

**Backward Compatibility**

This change may lead to slight round-off changes in how grid-based quantities are computed.  E.g. long-range forces in PPPM.

**Implementation Notes**

**Post Submission Checklist**


sjplimp:
@akohlmey @stanmoore1 I don't think this has any effect on accelerator packages that I could see.  It may break some regression tests, as indicated about re: round-off changes.

sjplimp:
@stanmoore1 you should definitely test MSM and PPPM/disp with this change

akohlmey:
@sjplimp I am seeing issues with the regression tests. For example, the attached input produces consistent results with 1, 2, or 4 MPI ranks, but different with 8.
[in.tpc.msm.txt](https://github.com/lammps/lammps/files/7017764/in.tpc.msm.txt)

sjplimp:
@akohlmey I think Stan and I fixed this issue with MSM non-periodic.  Glad your tests caught it.

akohlmey:
> @akohlmey I think Stan and I fixed this issue with MSM non-periodic. Glad your tests caught it.

@sjplimp did you already push this change? I cannot see it and check it out.

sjplimp:
@akohlmey I just resynced, I think the latest changes to msm.cpp are there

stanmoore1 (approved):

PR #2890, opened 2021-08-20
Fix bug with v remap option in Kokkos fix deform

stanmoore1:
**Summary**

When `domain.cpp` was checking for `fix deform`, the Kokkos suffix was excluded, and the `v remap` command was therefore ignored when using Kokkos.

**Related Issue(s)**

#2882 

**Author(s)**

Stan Moore (SNL)

**Licensing**

**Backward Compatibility**

Yes

### DIFF
--- a/src/KOKKOS/domain_kokkos.cpp
+++ b/src/KOKKOS/domain_kokkos.cpp
@@ -27,14 +27,8 @@ using namespace LAMMPS_NS;
 
 /* ---------------------------------------------------------------------- */
 
-DomainKokkos::DomainKokkos(LAMMPS *lmp) : Domain(lmp) {}
-
-/* ---------------------------------------------------------------------- */
-
-void DomainKokkos::init()
-{
+DomainKokkos::DomainKokkos(LAMMPS *lmp) : Domain(lmp) {
   atomKK = (AtomKokkos *) atom;
-  Domain::init();
 }
 
 /* ----------------------------------------------------------------------
--- a/src/KOKKOS/domain_kokkos.h
+++ b/src/KOKKOS/domain_kokkos.h
@@ -30,7 +30,6 @@ class DomainKokkos : public Domain {
  public:
   DomainKokkos(class LAMMPS *);
   ~DomainKokkos() {}
-  void init();
   void reset_box();
   void pbc();
   void remap_all();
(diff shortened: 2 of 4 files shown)


stanmoore1:
@akohlmey would it make sense to go through the code systematically and make it more "friendly" to suffix styles, i.e. using the `^` option with `utils::strmatch`?

akohlmey:
> @akohlmey would it make sense to go through the code systematically and make it more "friendly" to suffix styles, i.e. using the `^` option with `utils::strmatch`?

Absolutely. I have done this to a significant degree already in the more common code paths, searching for uses of `strncmp()` and `strstr()` especially, but also `strcmp()`, but those cases are more frequently correct and don't need to be changed unless we also want to replace a `char *arg` function argument into `const std::string &arg` and then use `(arg == "string")` instead of `(strcmp(arg,"string") == 0) which can (but need not) make the code more readable. I usually don't change it unless there are additional string operations like getting a substring or embedding it into a command and so on.

Most of the time, when I see a case that could be updated, I search for other uses of the same string to identify more cases in need of an update.

sjplimp (approved):

akohlmey:
As a point of reference:
```bash
$ grep utils::strmatch src/{,*/}*.cpp | wc -l
    576
$ grep strcmp src/{,*/}*.cpp | wc -l
   5957
$ grep strncmp src/{,*/}*.cpp | wc -l
     82
$ grep strstr src/{,*/}*.cpp | wc -l
     78
```

stanmoore1:
@akohlmey thanks, instead of using `utils::strmatch` would it be better to make a `utils::strmatch_suffix` command that only matched the active suffix style, instead of anything that starts with the string?

akohlmey:
> @akohlmey thanks, instead of using `utils::strmatch` would it be better to make a `utils::strmatch_suffix` command that only matched the active suffix style, instead of anything that starts with the string?

Then you can end up right where you are now, since you could be using a non-suffixed style or a differently suffixed style (when using the hybrid suffix option) and often you want to match not just the one style but all of it variants, too (see the fix rigid styles for example). So far, I have taken great care that the matched string applies to all styles that have the functionality that is being tested for, i.e. make the regex generic but not too generic.

The challenge from the maintenance perspective is to not allow any new fixes to be added with style names that incorrectly would match the existing patterns. Steve and I have been requesting changes to contributed styles for that reason *and* for making it easier for users to identify suitable fixes in the list of commands tables for several years now. This is becoming increasingly important. However, this is still way better than checking for substring matches with `strstr()` (those are the most problematic of the existing checks, with strncmp() the risk is mostly that one may forget to update the length to check for with any modification)

PR #2897, opened 2021-08-25
Correcting non-fdotr Virial tensor computation for interlayer potentials

dxhisboy:
**Summary**

I found that `pair/ilp/graphene/hbn` cannot produce similar pressure result with `vflag_fdotr` on/off.

I think the problem is due to incorrect method of calling `ev_tally_xyz` in the for-k loop in the `calc_FRep` function.

A simple proof is described in Implementation Notes.

**Related Issue(s)**

None

**Author(s)**

Xiaohui Duan  (dxhisboy@126.com), National Supercomputing Center in Wuxi
**Licensing**

**Backward Compatibility**

**Implementation Notes**

Changes calling `ev_tally` with `fk` and `delkj` to `fk` and `delki`. The proof is as follows:

image

The problem can be addressed as follows:

Giving following input script (which disables intra-layer potential because it produces a large pressure that hides the difference of pressure in inter-layer potential):

```
variable        Kz equal 0.1688
variable        nodotr index 0
units      metal
boundary   p p p
atom_style full

lattice custom 2.4595 a1 1  0  0  a2 0 1.73203 0  a3 0 0 4 &
               basis 0 0 0  basis 0.5 0.16666666666666666 0 &
               basis 0.5 0.5 0 basis 0 0.6666666666666666 0
region van block -10 10 -5 5 0 20 units lattice
region grn0 block -10 10 -5 5 7.5 8.5 units lattice
region grn1 block -10 10 -5 5 8.5 9.5 units lattice
region grn union 2 grn0 grn1
region all union 2 grn van
create_box 1 all
create_atoms 1 region grn0
create_atoms 1 region grn1
set region grn0 mol 1
set region grn1 mol 2
group top region grn1
displace_atoms top random 0.025 0.025 0.5  4123
mass 1 12.0107

pair_style      hybrid/overlay tersoff ilp/graphene/hbn 16.0
pair_coeff              * * tersoff SiC.tersoff C
pair_coeff              * * ilp/graphene/hbn BNCH.ILP C
if "${nofdotr}==1" then "pair_modify     pair ilp/graphene/hbn nofdotr compute yes"
pair_modify     pair tersoff compute no
neighbor                2.0 bin
neigh_modify    one 20000 page 1000000

write_data grn.data
dump            1 all atom 100000 sys.lammpstrj
thermo          1 #1000
thermo_style    custom step temp etotal pxx pyy pzz pxy pxz pyz
thermo_modify   flush yes
min_style cg
minimize  0.0 1.0e-6 1000 1
```

For the six directional pressure tensor `Pxx Pyy Pzz Pxy Pxz Pyz` in the first timestep which is related to Virial stress of the system.

When using `virial_fdotr_compute`, the pressure tensor is:
```
-5.9064665   -5.9056294   -31.284293 -0.0003513579 0.0011400409 -0.00089151243
```

When adding `nofdotr` to `pair_modify`, it is:
```
-5.904713   -5.9001243   -31.291564 0.0029140813 0.0052877203  0.013248622
```

In my modified code, it is:
```
-5.9064665   -5.9056294   -31.284293 -0.0003513579 0.0011400409 -0.00089151243
```

**Post Submission Checklist**

### DIFF
--- a/src/INTERLAYER/pair_ilp_graphene_hbn.cpp
+++ b/src/INTERLAYER/pair_ilp_graphene_hbn.cpp
@@ -482,7 +482,7 @@ void PairILPGrapheneHBN::calc_FRep(int eflag, int /* vflag */)
   double dprodnorm1[3] = {0.0, 0.0, 0.0};
   double fp1[3] = {0.0, 0.0, 0.0};
   double fprod1[3] = {0.0, 0.0, 0.0};
-  double delkj[3] = {0.0, 0.0, 0.0};
+  double delki[3] = {0.0, 0.0, 0.0};
   double fk[3] = {0.0, 0.0, 0.0};
 
   inum = list->inum;
@@ -588,12 +588,12 @@ void PairILPGrapheneHBN::calc_FRep(int eflag, int /* vflag */)
           f[k][0] += fk[0];
           f[k][1] += fk[1];
           f[k][2] += fk[2];
-          delkj[0] = x[k][0] - x[j][0];
-          delkj[1] = x[k][1] - x[j][1];
-          delkj[2] = x[k][2] - x[j][2];
+          delki[0] = x[k][0] - x[i][0];
+          delki[1] = x[k][1] - x[i][1];
+          delki[2] = x[k][2] - x[i][2];
           if (evflag)
-            ev_tally_xyz(k, j, nlocal, newton_pair, 0.0, 0.0, fk[0], fk[1], fk[2], delkj[0],
-                         delkj[1], delkj[2]);
+            ev_tally_xyz(k, i, nlocal, newton_pair, 0.0, 0.0, fk[0], fk[1], fk[2], delki[0],
+                         delki[1], delki[2]);
         }
 
         if (eflag) pvector[1] += evdwl = Tap * Vilp;
--- a/src/INTERLAYER/pair_kolmogorov_crespi_full.cpp
+++ b/src/INTERLAYER/pair_kolmogorov_crespi_full.cpp
@@ -478,7 +478,7 @@ void PairKolmogorovCrespiFull::calc_FRep(int eflag, int /* vflag */)
   double dprodnorm1[3] = {0.0, 0.0, 0.0};
   double fp1[3] = {0.0, 0.0, 0.0};
   double fprod1[3] = {0.0, 0.0, 0.0};
-  double delkj[3] = {0.0, 0.0, 0.0};
+  double delki[3] = {0.0, 0.0, 0.0};
   double fk[3] = {0.0, 0.0, 0.0};
 
   inum = list->inum;
@@ -585,12 +585,12 @@ void PairKolmogorovCrespiFull::calc_FRep(int eflag, int /* vflag */)
           f[k][0] += fk[0];
           f[k][1] += fk[1];
           f[k][2] += fk[2];
-          delkj[0] = x[k][0] - x[j][0];
-          delkj[1] = x[k][1] - x[j][1];
-          delkj[2] = x[k][2] - x[j][2];
+          delki[0] = x[k][0] - x[i][0];
+          delki[1] = x[k][1] - x[i][1];
+          delki[2] = x[k][2] - x[i][2];
           if (evflag)
-            ev_tally_xyz(k, j, nlocal, newton_pair, 0.0, 0.0, fk[0], fk[1], fk[2], delkj[0],
-                         delkj[1], delkj[2]);
+            ev_tally_xyz(k, j, nlocal, newton_pair, 0.0, 0.0, fk[0], fk[1], fk[2], delki[0],
+                         delki[1], delki[2]);
         }
 
         if (eflag) {


dxhisboy:
Could @oywg11  have a review on this PR? Thanks.

oywg11:
@dxhisboy Thanks for reading the codes carefully and propose your question.
I checked your derivation carefully and I think you didn't notice that there is no interaction between atom i and atom k. Thus the newton's third law should be applied to atom j instead of atom i (BTW, the dimension of your Eq. 3 is incorrect), attached please find my derivation.
image

akohlmey:
@oywg11 do you have an explanation why the provided example input then produces inconsistent stresses between using fdotr and `pair_modify nofdotr`?

dxhisboy:
> @dxhisboy Thanks for reading the codes carefully and propose your question.
> I checked your derivation carefully and I think you didn't notice that there is no interaction between atom i and atom k. Thus the newton's third law should be applied to atom j instead of atom i (BTW, the dimension of your Eq. 3 is incorrect), attached please find my derivation.
> image

Hi @oywg11 . Thank you for your explanation.
image
Thus, to grant a correct total pressure from `ev_tally` calls, the calling of `ev_tally_xyz` should be:
```C
for (i...){
  for (j...){
    ...
    for (k...){
      ...
     //just same as original code
      ev_tally_xyz(k, j, nlocal, newton_pair, 0.0, 0.0, fk[0], fk[1], fk[2], delkj[0],
                         delkj[1], delkj[2]);
    }
    ...
    //the force term needs to be corrected
    ev_tally_xyz(i, j, nlocal, newton_pair, evdwl, 0.0, fkcx - fprod1[0] * Tap, fkcy - fprod1[1] * Tap, fkcz- fprod1[2] * Tap, delx, dely, delz);
  }
}
```
It is equivalent to my code in total press but the distribution of per-atom stresses differs:
image

Since I did not find the strict definition of per-atom Virial stress, so I cannot tell which way is better.

I used to read the code of `reax/c`, and can almost tell while valance angle energies (just regard it as `F(i,j,k,bo(i,j), bo(i,k))`) have a gradient to to the bond order between i and j (modifies `bo_ij->Cdbo`), the force corresponds to gradient of bond order does not distribute Virial stresses to atom `k`.

dxhisboy:
In the Tersoff potential, there is a attractive term like: formula). Where formula is something like formula). It calls `v_tally3` to handle such forces. That is, the stress brought by Fk is distributed to i, j and k evenly.

In the ReaxFF potential, the valance angle term like: formula), and formula). We can see most forces of the valance angle is processed by `v_tally3`, but the gradient respecting formula only distributes Virial stress to atom i and j.

I think in ILP, the gradient respect to formula can be regarded as something like bond order, that is, distribute Virial stress brought by Fk to i and k just like my solution.

oywg11:
@dxhisboy Thanks for the detailed explanation. Now I understand your point. To answer Axel's question, I rechecked the codes and also noticed the issue you mentioned, i.e., fkc should be replaced by fkc-fprod*Tap. With fixing this issue, I confirmed that the total virial stress becomes consistent, as you already proved. Now the problem is which way is better, I also don't know how to judge. Maye Axel can comment on it? @akohlmey

akohlmey:
There is yet another way to tally the stresses if you look at pair style drip. I have been checking out the other INTERLAYER pair styles recently and was trying to correct the code for tallying stress in there in PR #2896. What is confusing me is that my own model system used for testing the INTERLAYER potentials does not show the erratic behavior demonstrated in this PR.

@dxhisboy please note that we recently refactored the ReaxFF (with the help of @athomps who is the real expert among the LAMMPS developers on these matters) and corrected nofdotr processing for this and a few other many-body pair styles (PR #2736. When you refer to tallying stress contributions from a "bond order"-like interaction, that is now tallied in reaxff with v_tally2_newton() and requires processing the neighbor list with newton on semantics. I confirmed that requirement by checking with newton pair set to off for the two /z interlayer styles and obtained inconsistent stresses again, so I also made `newton on` a requirement for the two pair potentials in question (it is already for the two pair styles discussed here).

akohlmey:
> Now the problem is which way is better, I also don't know how to judge. Maye Axel can comment on it? @akohlmey

My gut feeling is that you need to use a different tally method for the different types of contributions. I.e. regular ev_tally() for the "spherical symmetric" force term and one of the other v_tally() methods for the "directional" force term.

oywg11:
> > Now the problem is which way is better, I also don't know how to judge. Maye Axel can comment on it? @akohlmey
> 
> My gut feeling is that you need to use a different tally method for the different types of contributions. I.e. regular ev_tally() for the "spherical symmetric" force term and one of the other v_tally() methods for the "directional" force term.

Thanks for the suggestion. I will first read the way to tally the stresses in pair style drip and then reconsider how to process.

dxhisboy:
> There is yet another way to tally the stresses if you look at pair style drip. I have been checking out the other INTERLAYER pair styles recently and was trying to correct the code for tallying stress in there in PR #2896. What is confusing me is that my own model system used for testing the INTERLAYER potentials does not show the erratic behavior demonstrated in this PR.
> 
> @dxhisboy please note that we recently refactored the ReaxFF (with the help of @athomps who is the real expert among the LAMMPS developers on these matters) and corrected nofdotr processing for this and a few other many-body pair styles (PR #2736. When you refer to tallying stress contributions from a "bond order"-like interaction, that is now tallied in reaxff with v_tally2_newton() and requires processing the neighbor list with newton on semantics. I confirmed that requirement by checking with newton pair set to off for the two /z interlayer styles and obtained inconsistent stresses again, so I also made `newton on` a requirement for the two pair potentials in question (it is already for the two pair styles discussed here).

I'm know `newton off` will cause inconsistent between tally and fdotr, because the actual interaction site and virial accumulation site differ for boundary atoms in fdotr method, especially for periodical boundary conditions.

The problem illustrated in this PR is not related to newton pairs. Also, in hybrid Tersoff+ILP simulation, Tersoff produces much larger pressure, making the virial error in ILP hidden. That may be the reason your system cannot find it.

I had a look of the new ReaxFF. Luckily it doesn't change the per-atom virial distribution: most bond-order related forces are still distributed to i and j evenly, except conjugate pi-bonds.

akohlmey:
> The problem illustrated in this PR is not related to newton pairs. Also, in hybrid Tersoff+ILP simulation, Tersoff produces much larger pressure, making the virial error in ILP hidden. That may be the reason your system cannot find it.

My unit tests do not use multiple pair styles check out the pair_style and pair_coeff sections in the test input.
https://github.com/lammps/lammps/blob/master/unittest/force-styles/tests/manybody-pair-ilp-graphene-hbn.yaml

To explain, those tests are not using regular LAMMPS inputs, but the test program will assemble those inputs on-the-fly in multiple variants and permutations for different suffixes and settings and all compare them to the reference data obtained from the plain input. More explanations about how these tests are done are in the manual: https://docs.lammps.org/latest/Build_development.html#code-coverage-and-unit-testing-cmake-only

We are currently not testing for consistency between the sum of per-atom stress and global stress and that per-atom stress is reproduced since this is unpractical due to the huge amount of numerical noise on per-atom stress components that are close to zero. The purpose of these tests is twofold: 1) they require consistency between different ways to compute the same thing (e.g. Pair::compute() and Pair::single(), or using the multi-cutoff r-RESPA computes and the regular compute() within the same class = source file and different accelerator variants, 2) they provide a reference that must be met when modifying/refactoring code so that the old behavior is reproduced (unless a re-blessing is justified due to fixing a bug or changes that change the numerical precision without changing the algorithm).

PR #2898, opened 2021-08-25
AMOEBA and HIPPO force fields

sjplimp:
**Summary**

Adds support for the Tinker AMOEBA and HIPPO force fields

**Related Issue(s)**

N/A

**Author(s)**

Josh Rackers, Trung Nguyen, Steve Plimpton

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

These pair styles were created from the Tinker MD code Fortran files given to us by Jay Ponder's group at Washington University at St Louis

**Post Submission Checklist**

### DIFF
--- a/src/AMOEBA/amoeba_induce.cpp
+++ b/src/AMOEBA/amoeba_induce.cpp
@@ -388,7 +388,7 @@ void PairAmoeba::induce()
     // NOTE: could make this an error
 
     if (iter >= maxiter || eps > epsold)
-      if (comm->me == 0) 
+      if (comm->me == 0)
         error->warning(FLERR,"AMOEBA induced dipoles did not converge");
   }
 


akohlmey:
@sjplimp FYI, this still fails integration testing when compiling with -DLAMMPS_BIGBIG. It seems there is another assignment of a tagint pointer to an int pointer that I had overlooked in my previous changes.
I also noticed that in several other places, code constructs could be updated, e.g. for output to screen/logfile, creating/accessing fixes and more.

Do you mind, if I do another round of updates (since I have some time now) or would you rather prefer me to wait until later?

sjplimp:
I thnk I just fixed the tagint issue.  We're in the middle of another round of debugging for forces and virial.  So the
updates would be appreciated, but if you can wait until we finish the debugging that would be ideal.

akohlmey:
> I thnk I just fixed the tagint issue. We're in the middle of another round of debugging for forces and virial. So the updates would be appreciated, but if you can wait until we finish the debugging that would be ideal.

Sure.

I think you may need one more change:
```
  diff --git a/src/AMOEBA/atom_vec_amoeba.h b/src/AMOEBA/atom_vec_amoeba.h
  index 8953023550..3a50d449ec 100644
  --- a/src/AMOEBA/atom_vec_amoeba.h
  +++ b/src/AMOEBA/atom_vec_amoeba.h
  @@ -38,8 +38,8 @@ class AtomVecAmoeba : public AtomVec {
    private:
     int *num_bond,*num_angle,*num_dihedral,*num_improper;
     int **bond_type,**angle_type,**dihedral_type,**improper_type;
  -  int **nspecial;
  -  int *nspecial15,**special15;
  +  int **nspecial, *nspecial15;
  +  tagint **special15;
   
     int any_bond_negative,any_angle_negative,
       any_dihedral_negative,any_improper_negative;

```

sjplimp:
@akohlmey This is now ready to merge from mine and Josh's perspective.  I tested the new version of tools/tinker/tinker2lmp.py.  Using py2 it produced the same files as before. With py3 it added some precision to a few fields which I don't quite understand, possibly a difference in how 2 vs 3 does arithmentic or rounding?  Also it reordered some fields of the output, which is harmless, but I also don't understand.

akohlmey:
> @akohlmey This is now ready to merge from mine and Josh's perspective. I tested the new version of tools/tinker/tinker2lmp.py. Using py2 it produced the same files as before. 
> With py3 it added some precision to a few fields which I don't quite understand, possibly a difference in how 2 vs 3 does arithmentic or rounding?

Yes. In python2 the str() of a floating point number was rounded to 12 significant digits, but internally floating point represents more (as can be seen from using repr()). In python3 those two functions were made the same with str() printing all significant digits. https://stackoverflow.com/questions/25898733/why-does-strfloat-return-more-digits-in-python-3-than-python-2
We can make both return the same (to produce the python3 output), if needed.

> Also it reordered some fields of the output, which is harmless, but I also don't understand.

You are storing sections in a dictionary, which means that when looping over the keys of the dictionary, the order of the keys is arbitrary and dependent on the hashing algorithm used in the dictionary implementation unless you explicitly sort them.  Apparently, that algorithm is different in python2 compared to python3. In the data.py code, all sections and header with known keywords are printed in the order in which those keywords are stored in a list of known keywords, but all remaining keywords in native order.
We can sort those alphabetically to make them consistent across versions, if needed.

sjplimp:
@akohlmey thanks for the explanation - no need for sorting the data file fields

PR #2722, opened 2021-04-22
Avoid double free in some Kokkos pair styles destruction

rbberger:
**Summary**

During the destruction of some Kokkos Pair styles the Pair destructor tries to free non-null pointers. I got some of them, but I suspect there is more like these.

This is what happened with Kokkos and OpenMP during tests, which create LAMMPS instances one by one.
```
#0  0x00007ffff715fef5 in raise () from /usr/lib64/libc.so.6
#1  0x00007ffff7149862 in abort () from /usr/lib64/libc.so.6
#2  0x00007ffff71a1f38 in __libc_message () from /usr/lib64/libc.so.6
#3  0x00007ffff71a9bea in malloc_printerr () from /usr/lib64/libc.so.6
#4  0x00007ffff71aafbc in _int_free () from /usr/lib64/libc.so.6
#5  0x00007ffff71aeca8 in free () from /usr/lib64/libc.so.6
#6  0x000055555637be6b in LAMMPS_NS::Memory::sfree (this=0x55555a4481f0, ptr=0x55555a531b00) at /home/richard/GitHub/lammps/lammps/src/memory.cpp:118
#7  0x0000555556315ce1 in LAMMPS_NS::Memory::destroy<double> (this=0x55555a4481f0, array=@0x55555af139a8: 0x55555a531b00) at /home/richard/GitHub/lammps/lammps/src/memory.h:79
#8  0x000055555640c1e6 in LAMMPS_NS::Pair::~Pair (this=0x55555af138b0, __in_chrg=<optimized out>) at /home/richard/GitHub/lammps/lammps/src/pair.cpp:144
#9  0x0000555556e7e712 in LAMMPS_NS::PairCoulCut::~PairCoulCut (this=0x55555af138b0, __in_chrg=<optimized out>) at /home/richard/GitHub/lammps/lammps/src/pair_coul_cut.cpp:37
#10 0x0000555557bf5391 in LAMMPS_NS::PairCoulCutKokkos<Kokkos::OpenMP>::~PairCoulCutKokkos (this=0x55555af138b0, __in_chrg=<optimized out>) at /home/richard/GitHub/lammps/lammps/src/KOKKOS/pair_coul_cut_kokkos.cpp:51
[...]

PR #2900, opened 2021-08-26
Fix issues in KOKKOS package

stanmoore1:
**Summary**

Fix issues in KOKKOS package:

- Fix deallocation issue in `pair_lj_charmm_coul_charmm_kokkos`, missed in #2722
- Makefile.kokkos: fix (standard_in) 1: syntax error (https://github.com/kokkos/kokkos/pull/4173)
- Fix issue in `fix_nvt_sllod_kokkos` on multiple processors when running `nemd` example that slipped through #2882
- Fix issue with Kokkos granular and `pair/only on`
- Port changes from #2903

**Related Issue(s)**

#2722, https://github.com/kokkos/kokkos/pull/4173, #2882 

**Author(s)**

Stan Moore (SNL)

**Licensing**

**Backward Compatibility**

Yes


akohlmey:
> I'm still tracking down one more pesky issue. Once that is in, I'll run this through Kokkos regression testing and it will be ready to merge.

@stanmoore1 any updates on this?
Would you mind if I otherwise merge this later today or early tomorrow morning?
I have pushed back the patch release, but if we wait any longer we will delay the stable version even more than it already is. There will still be the opportunity to merge bugfixes and add (minor) changes to non-core parts of LAMMPS.

stanmoore1:
I'm swamped today, let's merge as-is and then submit a new small bugfix PR later for that other issue.

stanmoore1:
All Kokkos regression tests now pass with this PR.

akohlmey (approved):

PR #2342, opened 2020-09-04
Enable fix property/atom to create custom arrays

sjplimp:
**Summary**

Enhance fix property/atom to allow creation custom per-atom arrays, not just vectors,
with an i2_name, d2_name syntax.  In addition to custom vectors with an i_name, d_name syntax.
This includes additional data structures in atom.cpp/h.  And enabling other commands to work with 
either custom vectors or arrays: dump custom, compute property/atom, set, fix store/state, etc.
Also to allow the arrays to be extracted through the lib interface, via atom->extract().

**Related Issues**

**Author(s)**

Steve

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

Should be N/A.  Except that some error checking will be added to fix property/atom and fix store/state when restarting, which will change the restart file format if those commands are used.

**Implementation Notes**

**Post Submission Checklist**

- [ ] The feature or features in this pull request is complete
- [ ] Licensing information is complete
- [ ] Corresponding author information is complete
- [ ] The source code follows the LAMMPS formatting guidelines
- [ ] Suitable new documentation files and/or updates to the existing docs are included
- [ ] The added/updated documentation is integrated and tested with the documentation build system
- [ ] The feature has been verified to work with the conventional build system
- [ ] The feature has been verified to work with the CMake based build system
- [ ] Suitable tests have been added to the unittest tree.
- [ ] A package specific README file has been included or updated
- [ ] One or more example input decks are included

**Further Information, Files, and Links**

PR #2906, opened 2021-08-27
Adding package for bonded particle models

jtclemm:
**Summary**

The pull request includes a new package for bonded particle models (BPMs) – a set of mesoscale models of solid elasticity and fracture based on bonds that store their reference state and can break. The package is still a work in progress as I still need to upload examples, add other bond and angle styles, and run more tests,

**Author(s)**

Joel Clemmer (SNL)

**Licensing**

**Backward Compatibility**

No issues expected

**Implementation Notes**

To store per-bond data, a new internal fix bond/history was added. It uses the [new array](https://github.com/lammps/lammps/pull/2342) structures in fix property/atom to allocate and maintain extra per-bond data. This fix is initialized and used by BPM bond styles but in theory could be used by any other future bond style. A similar fix angle/history also exists but still needs to be uploaded.

These bond styles function similar to the quartic style as they can break. This is done by setting the bond type to zero and by clearing bond data stored in peratom arrays. Since bonds break during a simulation, the default special bonds list will become out of date and cannot be used to censor pair forces. In the quartic bond style, this was overcome by calling `pair->single()` and subtracting the forces. However, this is not ideal for BPMs since they often use granular pair styles which exert non-central-body forces and torques. Each BPM bond style would need to maintain pair-style-dependent logic to appropriately remove pair forces. Alternatively, a new fix (update/special/bonds) is included which updates the neighbor list and 1-2 special bond list as bonds break. I also have an older, local version of the bond styles which subtract pair forces which I still need to benchmark against.

Another challenge with special bond lists is that they are fundamentally incompatible with the gran* pair styles as fix neigh/history uses the special bond bits to store its own information. This conflict also complicates the usage of fix neigh/history with non-granular pair styles. I tried to fix this conflict by moving fix neigh/history to a 3rd bit by editing bit masking operations in various classes and updating lmptypes.

New comm forward and reverse methods were also added for bonds (copying the template for pair styles). 

Fix property/atom now also zeros out memory after copying arrays or packing data to exchange to clear out old data  to mirror the behavior in `grow_arrays()`. 

A new fix store/local was added to generalize and replace fix pair/tracker. This fix can now be used to generally accumulate and output local data. It can be used with the BPM bond styles to accumulate data on broken bonds.

**Post Submission Checklist**


sjplimp:
I have a conceptual Q for @jtclemm  and @akohlmey  as to whether the new fix
store/local in this PR should be a user command (as it is now), or a
hidden fix (like fix store) that other commands instantiate to archive
the local data they generate.  Currently I think this is just the pair
tracker and bond bpm/spring and bpm/rotational commands.

The advantage of making it internal is that the user doesn't have to
specify it in their script.  The only args it takes are Nevery and
Nvalues.  The latter is known to the instantiating command.  The
former could be another param the instantiating command provides.

The disadvantage of making it internal is that the fix needs an ID
that the user can reference in a command like dump local.  We could
document what ID the instantiating command uses (we do that for things
like the temperature compute created by fix nvt).  But it's a bit more
complicated to come up with a unique fix ID from a pair or bond style,
since it has no unique ID itself.  Or we could make the new fix
store/local ID an arg to the pair or bond instantiating command.

Maybe there are other reasons to have fix store/local be a user
command?  E.g. future options that will be added to it?  If that's the
case, I think it should remain a user command.

Thoughts?

jtclemm:
> The disadvantage of making it internal is that the fix needs an ID that the user can reference in a command like dump local. We could document what ID the instantiating command uses (we do that for things like the temperature compute created by fix nvt). But it's a bit more complicated to come up with a unique fix ID from a pair or bond style, since it has no unique ID itself. Or we could make the new fix store/local ID an arg to the pair or bond instantiating command.

Could we have the ID be `"pair_tracker_" + std::to_string(instance_me)`? And describe that multiple pair trackers will have IDs f_pair_tracker_0, f_pair_tracker_1, ...? Or maybe the first instance is just f_pair_tracker and subsequent instances have + _2, _3, etc.

sjplimp:
So how would the instantiating class (pair or bond style) know the fix instance count?  Also, the user has
to know the ID given to the fix, so that the input script can access it, e.g. for dump output.  And the user
won't know the instance counts.  @akohlmey any suggestions?

jtclemm:
As we discussed, I was thinking of the pair or bond instance count. However, now I'm thinking it might be better for the user to just provide a fix ID to pair tracker or bond bpm/*, ensures they know exactly what ID stores their data and avoids all issues with multiple bond/pair styles. 

I'll make these changes, but can always swap to another setup if we decide there's a better naming convention.

sjplimp:
That seems fine.  Not sure we do that anywhere else, but can't think of a better solution at the moment.

sjplimp (approved):

sjplimp:
@akohlmey Joel and I think this is ready to be merged.  Aside from a few merge conflicts Joel can fix.  And he may have an additional Q for you.

jtclemm:
Yes, this pull request is finished and ready for final review.

I mostly answered my question myself, which was how to restart bond styles. I added some code to input.cpp to accomplish this. However, I do not know whether all the suffixes are relevant to bond styles. Is suffixp suffix package or suffix pair?

sjplimp:
@jtclemm I see that you've copied code in input.cpp from the pair_style() method to the bond_style() method, presumably
to just change settings of an existing bond style, rather than re-instantiate it.

@akohlmey - I can't recall why we do this for pair styles.  I don't see anything on the pair_style doc page descriging this functionality.  Is a pair_style (or now a bond_style) storing some state that needs to be preserved if its global params are changed in the input script?

sjplimp:
@jtclemm @akohlmey Also, I don't see that pair hybrid follows this logic from input.cpp.  If you specify pair hybrid twice,
only settings() will be called the 2nd time.  But it wipes out all the sub-styles and re-instantiates them.  Unlike what input.cpp is doing when it just calls settings().

I'm not saying hybrid should be changed.  Just that it's inconsistent if there is some reason not to just always re-instantiate
the entire pair style.  That's what we do for every other style isn't it?

akohlmey:
> @akohlmey - I can't recall why we do this for pair styles. I don't see anything on the pair_style doc page descriging this functionality. Is a pair_style (or now a bond_style) storing some state that needs to be preserved if its global params are changed in the input script?

`LAMMPS::suffixp` is a cached copy of the first suffix (i.e `LAMMPS::suffix`) in case either the GPU or the KOKKOS package are used in combination with the with the "package pair/only on" setting. In that case `LAMMPS::suffix` is set to NULL. The logic is then: use `LAMMPS::suffixp` if non-NULL for pair styles only, then check for `LAMMPS::suffix` and - if non-NULL - for `LAMMPS::suffix2`. All other styles will see that `LAMMPS::suffix` is NULL and not apply any suffixes.

sjplimp:
I'm not asking about the details of the matching.  I'm asking about why we follow this comment above Input::pair_style()

// if old pair style exists and new style is same, just change settings
// else create new pair class

That's what @jtclemm is now also doing in bond_style().

I'm asking why do we bother to do that and not just always re-instantiate the pair/bond style ?

akohlmey:
> I'm not asking about the details of the matching. I'm asking about why we follow this comment above Input::pair_style()

I was responding to the first comment, not the second. Bottom line: do not pay attention to `LAMMPS::suffixp` for bond styles.

jtclemm:
For context, I added the ability to restart bond styles so the internal fixes could correctly restart.

I noticed recent changes to the granular pair styles where the internal fixes were given unique ids (e.g. "NEIGH_HISTORY_HH" + instance_me) and used a similar style for the ids of internal fixes in the bond styles. Since the fix ID depends on the bond instance, the bond style had to be restored from the restart file to preserve the value of instance_me.

Alternatively I could revert back to a static internal fix id.

doc/src/atom_style.rst
```diff
@@ -10,7 +10,7 @@ Syntax
 
    atom_style style args
 
-* style = *angle* or *atomic* or *body* or *bond* or *charge* or *dipole* or  *dpd* or *edpd* or *electron* or *ellipsoid* or *full* or *line* or *mdpd* or *molecular* or *oxdna* or *peri* or *smd* or *sph* or *sphere* or *spin* or *tdpd* or *tri* or *template* or *hybrid*
+* style = *angle* or *atomic* or *body* or *bond* or *charge* or *dipole* or  *dpd* or *edpd* or *electron* or *ellipsoid* or *full* or *line* or *mdpd* or *molecular* or *oxdna* or *peri* or *smd* or *sph* or *sphere* or *sphere/bpm* or *spin* or *tdpd* or *tri* or *template* or *hybrid*
```
akohlmey, line 13: Why atom style `sphere/bpm`? Why not just atom style `bpm`?
We don't have any other pair styles with a "slash" (except for suffixed versions for kokkos) and there are others that are supersets (e.g. atomic -> bond -> angle -> molecular -> full, or atomic -> charge -> full)

jtclemm: There are many BPMs in the literature and they combine many different types of bond and atom styles including point particles, finite particles, aspherical particles, etc. Since there isn't such a thing as a universal bpm atom style, I avoided using that label. I tried to make the name derivative of atom style sphere since particles are rotational spheres with a fixed density/diameter. Is there an appropriate way to use a two word name? 

I'll also try and think of one word names.

akohlmey: > Is there an appropriate way to use a two word name?

The "natural" approach to that from within LAMMPS would be to use something like  `atom_style hybrid sphere bpm` and then have the bpm atom style only implement what is missing from sphere.  After the most recent refactoring of the AtomVec classes by @sjplimp the difference in the realization of, say, `atom_style full` versus `atom_style hybrid charge molecular` has become very small.

Mind you I have not really looked at the code but just at the documentation, so I cannot make any statements about possible complications in the implementation.

Mind you, my time is currently rather limited, so I won't be able to do a proper code review and my own tests for a while.
There are a few windows of opportunity. But then again, I am currently waiting for the PR to clear the integration and unit testing before even trying to figure out what needs looking at. Since this touches a lot of core functionality, it will be required to look very carefully to minimize the risk that existing functionality is broken.

sjplimp:
My guess is the logic in Input::pair_style() for suffixes was added later.  Separate from suffixes, if you issue a second pair_style lj/cut Rc command, it is only going to invoke settings() and change the Rc value, whereas the first time it creates the pair style and then invokes settings().  I'm trying to figure out why we don't simply always do the latter.  As you say, that seems inconsistent with how we treat all other style commands.

I think this is hidden from the user, so I don't expect it to break any inputs.  If it does for some reason, that's the use case movitvating why we do this.  I'd like to know if such a case exists.  Or maybe long ago we just thought if was more efficient not to re-instantiate the pair style.

akohlmey:
> I think this is hidden from the user, so I don't expect it to break any inputs. If it does for some reason, that's the use case movitvating why we do this. I'd like to know if such a case exists. Or maybe long ago we just thought if was more efficient not to re-instantiate the pair style.

Don't you keep the pair_coeff settings when you just call `Pair::settings()`?

At any rate, this choice was made long before I started using LAMMPS and it was something that confused me for a while when I started.

sjplimp:
ah, this is the reason.  That should be documented on the pair_style doc page, then we wouldn't have had this discussion :)
@jtclemm is something similar the case for your new bond style?  I.e. there are complex settings made by bond_coeff commands and also some global settings made in the bond_style command.  And you want to be able to change the latter w/out respecifying all the bond_coeff commands?  If not, I'm unclear why you are adding this logic in Input::bond_style().

jtclemm:
@sjplimp No, I only added that logic such that the `instance_me` variable wouldn't be incremented every time a restart file was loaded. This allowed me to mirror the internal fix naming convention in the gran pair styles (seen [here](https://github.com/lammps/lammps/blob/develop/src/GRANULAR/pair_gran_hooke_history.cpp#L447)) which used `instance_me` to make unique fix IDs. Otherwise if the bond style was recreated, the `instance_me` would increment twice and the bond style wouldn't be able to find and load its internal fixes from the restart file.

I could also revert back to having a set fix ID since I don't know how many applications would actually use multiple copies of the same bond style. Maybe someone would want some settings to apply to some atoms but not others? I have no strong opinions.

doc/src/atom_style.rst
jtclemm: I see, in this case it'd be `atom_style hybrid sphere bond X` where `X` would be a new atom style that just stores what is missing (namely a quaternion). I'll think about possible alternate solutions.

I understand completely about time. I know this PR has some hairy changes - like the reworking of special bond/history bits in neighbor lists.

I fixed an overlooked merge conflict I introduced earlier today and that resolved the build issue. I can also look at the issues with the unit tests but I admittedly do not know where to find details on what failed.

jtclemm:
I updated this branch to address the above discussion by
a) renaming atom/sphere/bpm to atom/bpm/sphere 
b) replacing internal fix IDs with static strings without the `instance_me` tag so there is no longer a need to match bond styles in the Input class
I also fixed merge conflicts and cleaned up a few minor portions of the code/documentation.

Thanks again for the help!

akohlmey:
@jtclemm I will take a look after the next patch release.

akohlmey:
@jtclemm I've updated the code in the pull request to the latest development version from upstream and resolved some issues. Also modernized some obsolescent code constructs. But this was just what I saw during the resolution of merge conflicts.

The documentation still needs a *lot* of work. Running "make html spelling" in the doc folder yields a long list of errors and warnings that need to be addressed.

jtclemm:
> @jtclemm I've updated the code in the pull request to the latest development version from upstream and resolved some issues. Also modernized some obsolescent code constructs. But this was just what I saw during the resolution of merge conflicts.
> 
> The documentation still needs a _lot_ of work. Running "make html spelling" in the doc folder yields a long list of errors and warnings that need to be addressed.

Thank you Axel. I went through and read your edits, making particular note of the modernization. I'll keep these code/style changes in mind when writing future code.

Unless you have already started editing the documentation, I will run the html make command and start combing through the errors/warning later this afternoon.

akohlmey:
> Unless you have already started editing the documentation, I will run the html make command and start combing through the errors/warning later this afternoon. 

No, I have not started.

jtclemm:
@akohlmey I fixed the majority of errors/warnings from the html build. The few remaining issues do not appear to be related to the changes in the PR (e.g. in Developer_utils.rst).

However, I have not yet run a spell check since Sphinx reports an error that I need to look into (Builder name spelling not registered or available through entry point).

akohlmey:
> @akohlmey I fixed the majority of errors/warnings from the html build. The few remaining issues do not appear to be related to the changes in the PR (e.g. in Developer_utils.rst).
> 
> However, I have not yet run a spell check since Sphinx reports an error that I need to look into (Builder name spelling not registered or available through entry point).

Don't worry about this. I'll take over from here and will let you know in case I need a helping hand again.
Minor issues are often much easier and faster for me to resolve since I have so much practice. :wink:

akohlmey:
> Builder name spelling not registered or available through entry point

my guess would be that you don't have a compatible version of the "enchant" library installed that is used by the spell checker, so it is not loaded by sphinx

akohlmey:
@jtclemm spelling errors are already resolved...

src/fix_property_atom.cpp
```diff
@@ -26,21 +25,21 @@
...
-  Fix(lmp, narg, arg),
-  nvalue(0), styles(nullptr), index(nullptr), astyle(nullptr)
+    Fix(lmp, narg, arg), nvalue(0), styles(nullptr), index(nullptr), astyle(nullptr)
 {
-  if (narg < 4) error->all(FLERR,"Illegal fix property/atom command");
+  if (narg < 4) error->all(FLERR, "Illegal fix property/atom command");
 
   restart_peratom = 1;
   wd_section = 1;
+  create_attribute = 1;
```
akohlmey, line 39: @jtclemm @sjplimp this setting here and the set_arrays() function below create a problem: reading custom properties via fix property/atom from either data or restart files is broken. Instead they are reset to 0 *after* they have been read in. This breaks unit and regression tests.

I could work around this for reading data files with the following hack, but I could not figure out the equivalent change for restarts and I feel that this is not the right approach, anyway. @jtclemm can you provide some insight why this is needed for the BPM package?

I am also attaching so a small input to test this by writing out a data and restart file and then reading them back and writing out another data file. The second and third data file written has all custom properties set to 0.
[in.property.txt](https://github.com/lammps/lammps/files/8480773/in.property.txt)

```
diff --git a/src/fix_property_atom.cpp b/src/fix_property_atom.cpp
index 7be3d8f343..b4b87932b3 100644
--- a/src/fix_property_atom.cpp
+++ b/src/fix_property_atom.cpp
@@ -51,6 +51,8 @@ FixPropertyAtom::FixPropertyAtom(LAMMPS *lmp, int narg, char **arg) :
   nvalue = 0;
   values_peratom = 0;
 
+  noinit_flag = 0;
+
   while (iarg < narg) {
     if (strcmp(arg[iarg], "mol") == 0) {
       if (atom->molecule_flag)
@@ -244,6 +246,8 @@ void FixPropertyAtom::init()
 
   if (strcmp(astyle, atom->atom_style) != 0)
     error->all(FLERR, "Atom style was redefined after using fix property/atom");
+
+  noinit_flag = 0;
 }
 
@@ -264,6 +268,8 @@ void FixPropertyAtom::read_data_section(char *keyword, int n, char *buf, tagint
     atom->map_set();
   }
 
+  noinit_flag = 1;
+
   // loop over lines of atom info
   // tokenize the line into values
   // if I own atom tag, assign its values
@@ -579,6 +585,7 @@ void FixPropertyAtom::copy_arrays(int i, int j, int /*delflag*/)
 void FixPropertyAtom::set_arrays(int i)
 {
   int k, ncol;
+  if (noinit_flag) return;
 
   for (int nv = 0; nv < nvalue; nv++) {
     if (styles[nv] == IVEC)
diff --git a/src/fix_property_atom.h b/src/fix_property_atom.h
index 5e57ef9f91..92eb970ff5 100644
--- a/src/fix_property_atom.h
+++ b/src/fix_property_atom.h
@@ -62,6 +62,7 @@ class FixPropertyAtom : public Fix {
 
   int values_peratom;    // # of values per atom, including multiple for arrays
   int nmax_old;          // length of peratom arrays the last time they grew
+  int noinit_flag;       // do not initialize added data
 };
 
 }    // namespace LAMMPS_NS


```

jtclemm: > I could work around this for reading data files with the following hack, but I could not figure out the equivalent change for restarts and I feel that this is not the right approach, anyway. @jtclemm can you provide some insight why this is needed for the BPM package?

This change allowed bonded molecules to be created mid way through a simulation via fix deposit, fix pour, etc. By setting arrays to a predictable value, namely zero, fix bond/history can identify when histories need to be initialized. An older solution was to alternatively clean up data left behind when atoms moved.

akohlmey: > By setting arrays to a predictable value, namely zero, fix bond/history can identify when histories need to be initialized.

can you provide or point me to a sufficiently small representative example demonstrating the flow of control and what property is provided through fix property/atom? I have some thoughts about how this issue could be addressed differently and the broken functionality restored.

akohlmey: Please see commit 0f1097d7e60a49328cf582c6334cbd67b20f87e0 for an alternate approach to achieve initialization of custom properties when creating atoms from molecule files. This seems a much more logical place to do this. Beyond setting those to zero, it may also be possible to have custom properties stored in molecule files and then look them up and initialize them not just to zero.

jtclemm: Here's an [example](https://github.com/jtclemm/lammps/blob/BPM/examples/bpm/pour/in.bpm.pour) of pouring molecules where bonds store their initial length and the unit vector orienting the bond. As falling atoms exit a processor domain and leave old data in fix property/atom, newly created bonds would have the wrong reference state, calculate large forces/torques, and explode.

If it would be helpful, I can reduce this script and try to find a processor layout that quickly triggers this error if the set_array() method is removed.

sjplimp: Not seeing the logic of this.  Fix property/atom is where the memory is owned.  So having it zero it
out when an atom is added make sense to me.  There are other fixes which own per-atom data which do the same thing
with the same function.
Also, there are other uses for fix property/atom besides data for molecule atoms.  So putting the zeroing in atom->add_molecule_atom() doesn't seem sufficient.

sjplimp: also, not sure what this means:
>> reading custom properties via fix property/atom from either data or restart files is broken
I'm using this in the AMOEBA PR, it works fine, at least from data files (haven't tried it from restarts)

akohlmey: > also, not sure what this means:
 
> > > reading custom properties via fix property/atom from either data or restart files is broken

it means that the changes in this pull request which *introduce* the FixPropertyAtom::set_arrays() function and enable it break both read_restart and read_data and multiple tests depending on that functionality.

> > > I'm using this in the AMOEBA PR, it works fine, at least from data files (haven't tried it from restarts)

because *your* branch does not have the change that breaks it.

akohlmey: > If it would be helpful, I can reduce this script and try to find a processor layout that quickly triggers this error if the set_array() method is removed.

Yes, that would be needed, because right now I am not seeing any issues when running with 1 or 4 processors on my (tiny) desktop.

sjplimp: But if I understand, your commit removes set_arrays() from fix property/atom which means you will be breaking the zeroing for all the other use cases of fix property/atom ?

akohlmey: > But if I understand, your commit removes set_arrays() from fix property/atom which means you will be breaking the zeroing for all the other use cases of fix property/atom ?

Have a look at the file in *your* checkout. Does it have a set_arrays() function?

akohlmey: > Not seeing the logic of this. Fix property/atom is where the memory is owned.

I disagree. The custom properties are "owned" by the Atom class. Fix property/atom only "manages" them.
Please see the `Atom::add_custom()` function.

akohlmey: @jtclemm I have enabled hyper-threading and adjusted the processor distribution and I don't see any of the problems you mention when running the bpm.pour example. There is plenty of transitions/handovers between subdomains happening (indicated by the yellow lines). https://youtube.com/shorts/vzj9ZUVUknw

akohlmey (approved):
As far as I am concerned this is now ready for merging.

src/fix_property_atom.cpp
jtclemm: > @jtclemm I have enabled hyper-threading and adjusted the processor distribution and I don't see any of the problems you mention when running the bpm.pour example. There is plenty of transitions/handovers between subdomains happening (indicated by the yellow lines). https://youtube.com/shorts/vzj9ZUVUknw

Thank you again for the thorough testing and for cleaning up all the issues in this PR. I also ran a bunch of tests this morning/afternoon for different pours with the new commits and did not see any issues. I have a few more jobs to test, but even if there is an issue I think it's going to be a rare edge case.

akohlmey:
@jtclemm please let me know if you are done testing. I would prefer to merge this *before* #3217 since it'll be easier to resolve the expected merge conflict (and it would be my job, too).

jtclemm:
> @jtclemm please let me know if you are done testing. I would prefer to merge this _before_ #3217 since it'll be easier to resolve the expected merge conflict (and it would be my job, too).

Just finished testing. I don't see any issues with the change from zeroing values from set_array() to add_molecule(). Thanks again!

PR #2922, opened 2021-09-04
Work around issues from calling Py_Finalize() when the Python class is reset using `clear`

akohlmey:
**Summary**

LAMMPS may crash when using the python command multiple times when LAMMPS has been reset by the `clear` command. This happens for example when importing the numpy module in each instance. This pull request implements a workaround that avoids triggering the crash.

**Related Issue(s)**

numpy/numpy#8097, numpy/numpy#11925

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

As stated in the linked issues, the problem arises when initializing Python as second time after calling `Py_Finalize()`. In LAMMPS this function is called in the destructor which is triggered by the LAMMPS::destroy() function, that deletes the base classes and is called by either the `clear` command or the `LAMMPS::~LAMMPS()` destructor. By adding a flag to the class, a `finalize()` member function to change it, and then calling that `finalize()` function only from the LAMMPS class destructor, calls to `Py_Finalize()` are no longer possible during the `clear` command.

**Post Submission Checklist**

### DIFF
--- a/src/PYTHON/python_impl.cpp
+++ b/src/PYTHON/python_impl.cpp
@@ -69,6 +69,10 @@ PythonImpl::PythonImpl(LAMMPS *lmp) : Pointers(lmp)
   // pyMain stores pointer to main module
   external_interpreter = Py_IsInitialized();
 
+  // only call Py_Finalize() if fully destroying the LAMMPS instance
+  // which will call finalize() before deleting the PythonImpl instance.
+  need_finalize = false;
+
 #ifdef MLIAP_PYTHON
   // Inform python intialization scheme of the mliappy module.
   // This -must- happen before python is initialized.
@@ -113,7 +117,7 @@ PythonImpl::~PythonImpl()
   // shutdown Python interpreter
   if (!external_interpreter) {
     PyGILState_Ensure();
-    Py_Finalize();
+    if (need_finalize) Py_Finalize();
   }
 
   memory->sfree(pfuncs);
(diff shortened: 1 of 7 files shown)


rbberger (approved):
It should be noted that this prevents the Python interpreter to be fully reset after clear. Given that we don't reset the interpreter if it is external in the first place, this actually makes it more consistent. Meaning, the Python state is not linked to the LAMMPS state anymore. Should we add a `lammps_python_finalize` function, similar to what we do for KOKKOS? Or just document it somewhere that Python has to be finalized explicitly when using LAMMPS from the library?

akohlmey:
> It should be noted that this prevents the Python interpreter to be fully reset after clear. Given that we don't reset the interpreter if it is external in the first place, this actually makes it more consistent. Meaning, the Python state is not linked to the LAMMPS state anymore. Should we add a `lammps_python_finalize` function, similar to what we do for KOKKOS? Or just document it somewhere that Python has to be finalized explicitly when using LAMMPS from the library?

If we want to go this route, then the implementation needs to be changed. The boolean can be removed and instead a static function added that just finalizes Python and then this function is called from the LAMMPS destructor *after* deleting the Python class instance. That function could be exposed in the library interface, but it  is not quite clear to me why that would be needed. Why would you need to "finalize" Python when you are still *in* Python?

rbberger:
> Why would you need to "finalize" Python when you are still _in_ Python?

Because Python isn't the only thing that can create and delete LAMMPS instances via the library interface. The moment you delete the first instance, the destructor is called and `Py_Finalize` is called, even in this PR.

akohlmey:
superseded by PR #2923

Issue #2751, opened 2021-05-04
fmt headers not installed alongside lammps


snikolov3:
**Summary**

The fmt headers are currently not installed alongside with lammps. This can create issues when using lammps as an external library in other projects (missing fmt headers).

**LAMMPS Version and Platform**

LAMMPS version: 8 Apr 2021
Tested on Power8 node with openmpi/3.1.0/gcc/7.2.0

akohlmey:
The "install" target is currently not officially sanctioned by the LAMMPS developers and was an ad hoc addition done at some point by a contributor that went in mostly unnoticed and unreviewed.

The only sanctioned library interface currently is the `library.h` file and that should work without fmtlib and - in fact - without C++ or MPI even. For the rest the assumption (still) is that one will build applications using LAMMPS as a (C++) library based on the source code distribution. There are some plans to have a suitable and reasonably portable/abstracted "exported" C++ library interface in the future, but there is no timeline on it.

akohlmey:
> I'm having this issue when trying to create a plugin for lammps, to be loaded with the PLUGIN package. Since my plugin defines a new pair style, it needs to include `pair.h`, which includes `pointers.h`, which includes `fmt/format.h` and fails the compilation.
> 
> I can find a workaround by writing more cmake code & linking to a non `make install`ed version of LAMMPS.

as already discussed in this (closed!) issue. The "make install" target is not officially sanctioned. What is there was added to comply with some packaging requirements for some Linux distributions, but it was not reviewed. There are plans to have an "official" LAMMPS API, but for that some significant changes first need to be made to the organization of the LAMMPS header files. those changes are made rather complex by the need to support two different build systems and it is not a high priority item currently.

> As a related question, what is the currently recommended way to write a new pair style for future inclusion in LAMMPS? The possibilities I see are

> * write the code outside of LAMMPS & compile as a plugin

this is likely the easiest approach. You can use the content of the examples/plugins folder as a template for how to set it up.
there are some comments in the CMakeLists.txt file for how you can adapt it. CMake should be given a
 -DLAMMPS_HEADER_DIR=<...> with the path to the LAMMPS source folder.

> * write the code outside of LAMMPS & copy/paste the files inside `lammps/src`. Unfortunately, this does not integrate well with cmake, which I would really prefer to use (my pair style depends on the C++ torch library, which would be a nightmare to integrate without cmake)

since you will need to compile/link an external library, there are some complications.
[...]

PR #2929, opened 2021-09-08
Recover Kokkos compilation

stanmoore1:
**Summary**

The Kokkos compile is currently broken due to changes to the parent `GridComm` class. The `gridcomm_kokkos.h` file depends on `fftdata_kokkos.h` which is only installed with the `KSPACE` package. This PR adds a requirement that the `KSPACE` package be installed to use `gridcomm_kokkos`. The long-term solution is to extract raw pointers from the views and pass those around as `void*` to support other use cases of the gridcomm class (outside the scope of this PR).

**Related Issue(s)**

None

**Author(s)**

Stan Moore (SNL)

**Licensing**

**Backward Compatibility**

Yes

### DIFF
--- a/src/KOKKOS/Install.sh
+++ b/src/KOKKOS/Install.sh
@@ -163,8 +163,8 @@ action fix_dpd_energy_kokkos.cpp fix_dpd_energy.cpp
 action fix_dpd_energy_kokkos.h fix_dpd_energy.h
 action fix_rx_kokkos.cpp fix_rx.cpp
 action fix_rx_kokkos.h fix_rx.h
-action gridcomm_kokkos.cpp gridcomm.cpp
-action gridcomm_kokkos.h gridcomm.h
+action gridcomm_kokkos.cpp fft3d.h
+action gridcomm_kokkos.h fft3d.h
 action improper_class2_kokkos.cpp improper_class2.cpp
 action improper_class2_kokkos.h improper_class2.h
 action improper_harmonic_kokkos.cpp improper_harmonic.cpp


stanmoore1:
@akohlmey does this need a change for CMake?

junghans:
@snikolov3 had some problems with the install in #2751, but I am not sure if he used kokkos there.

akohlmey:
> @akohlmey does this need a change for CMake?

I'll check it out and will update the PR, if needed.

akohlmey:
> @snikolov3 had some problems with the install in #2751, but I am not sure if he used kokkos there.

That was a completely different issue. Having a "sanitized" and "installable" set of LAMMPS headers (that will live in a "include/lammps" folder and be included with "lammps/some_name.h") is a goal for the next development cycle. How to deal with fmtlib in that case has yet to be figured out. We'll probably need a define that indicates that we are compiling LAMMPS itself and not using the headers for using the C++ interface in a different code. For now, the only clean way to do this is to not "install" LAMMPS but use it in place.

snikolov3:
Yea Axel is correct. In my case, I did use Kokkos, but I was using Lammps
in another code which had trouble finding the fmtlib header files.

akohlmey:
> > @akohlmey does this need a change for CMake?
> 
> I'll check it out and will update the PR, if needed.

That was already pre-empted. As it stands, `fft3d_kokkos.cpp`, `gridcomm_kokkos.cpp` and `remap_kokkos.cpp` are only compiled if the KSPACE package is enabled. The logic in CMake is a bit different.
Fortunately in this case.

stanmoore1:
>As it stands, fft3d_kokkos.cpp, gridcomm_kokkos.cpp and remap_kokkos.cpp are only compiled if the KSPACE package is enabled. The logic in CMake is a bit different.

Great thank you for checking.

akohlmey (approved):

Issue #551, opened 2017-06-26
Kokkos Reax Package Development Plan


jimkress65:
Is there any working being done on (or planned for) the Kokko Reax package to incorporate ACKS2 and eReaxFF?

akohlmey:
Please post questions about LAMMPS to the lammps-users mailing list. We would like to keep this forum reserved for bug reports and discussions between developers and contributors. Thanks for your understanding.

stanmoore1:
Or it is also fair to rephrase this as an enhancement request and post it here on GitHub, like "Incorporate ACKS2 and eReaxFF into Kokkos Reax package".

jimkress65:
> Or it is also fair to rephrase this as an enhancement request and post it here on GitHub, like "Incorporate ACKS2 and eReaxFF into Kokkos Reax package".

That would be fine with me.  I figured I’d get yelled at for posting it in the LAMMPS forum, since it is KOKKOS specific.

akohlmey:
It is a bit of a borderline case with no straightforward response. 

i don't mind seeing a well formulated feature request here with a suitable subject line and a good motivation. but unless it is a feature, that is obviously useful and easy to implement, or you are offering to participate in implementing it, chances are small to receive the kind of response you are looking for. at least, that is what it looks like when i am looking at the issues that were filed so far. also, please keep in mind, that not all LAMMPS developers currently pay as much attention to the github issue tracker as they do to the lammps-users mailing list. i don't think being kokkos-specific makes it off-topic for the mailing list. after all, you are not asking to have your research explained to you or similar. the most useful posts here so far have been either bug reports or discussions among developers/contributors.

feature requests are labeled "enhancement", so you can see some examples at  https://github.com/lammps/lammps/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Aenhancement%20

what tipped the scales toward closing the issue is, that i am very worried about setting a precedent that may encourage others to ask **definitely** off-topic questions on the issue tracker (like "how do i do XXX?" or "what does this error mean?" and so on).

i hope that explains it a bit better, why i think posting to the mailing list would be the more suitable choice in this particular case.

jimkress65:
Axel,

 

Thank you for your detailed response.  I submitted the question to the LAMMPS list as you suggested.

 

Jim

PR #2931, opened 2021-09-08
Add ACKS2 charge equilibration method to REAXFF and support for elect…

stanmoore1:
…ric fields in QEq

**Summary**

- Add the `ACKS2` charge equilibration method to REAXFF, see Verstraelen et al., J. Chem. Phys. 138, 074108 (2013), https://doi.org/10.1063/1.4791569

- Add support for electric fields (`fix efield`) in `QEq` and `ACKS2` based on the formulation in Assowe et al., J. Phys. Chem. A, 116, 48, 11796–11805 (2012), https://doi.org/10.1021/jp306932a

**Related Issue(s)**

#551

**Author(s)**

Stan Moore (SNL). Code was ported from a serial version of ACKS2 in PuReMD, with help from Kurt Ohearn (MSU) and Metin Aktulga (MSU). Thanks to Jason Koski (SNL) and Matt Lane (SNL) for testing the new code and reporting bugs.

**Licensing**

**Backward Compatibility**

Yes

**Implementation Notes**

Implementation was verified by comparing to the serial version of ACKS2 in PuReMD

**Post Submission Checklist**

### DIFF
--- a/src/KOKKOS/fix_acks2_reaxff_kokkos.cpp
+++ b/src/KOKKOS/fix_acks2_reaxff_kokkos.cpp
@@ -1,0 +1,1930 @@
+/* ----------------------------------------------------------------------
+   LAMMPS - Large-scale Atomic/Molecular Massively Parallel Simulator
+   http://lammps.sandia.gov, Sandia National Laboratories
+   Steve Plimpton, sjplimp@sandia.gov
+
+   Copyright (2003) Sandia Corporation.  Under the terms of Contract
+   DE-AC04-94AL85000 with Sandia Corporation, the U.S. Government retains
+   certain rights in this software.  This software is distributed under
+   the GNU General Public License.
+
+   See the README file in the top-level LAMMPS directory.
+------------------------------------------------------------------------- */
+
+/* ----------------------------------------------------------------------
+   Contributing authors: Stan Moore (SNL)
+------------------------------------------------------------------------- */
+
+#include "fix_acks2_reaxff_kokkos.h"
+#include <cmath>
+#include "kokkos.h"
+#include "atom.h"
+#include "atom_masks.h"
+#include "atom_kokkos.h"
+#include "comm.h"
+#include "force.h"
+#include "neighbor.h"
+#include "neigh_list_kokkos.h"
+#include "neigh_request.h"
+#include "update.h"
+#include "integrate.h"
+#include "memory_kokkos.h"
+#include "error.h"
+#include "pair_reaxff_kokkos.h"
+
+using namespace LAMMPS_NS;
+using namespace FixConst;
+
+#define SMALL 0.0001
+#define EV_TO_KCAL_PER_MOL 14.4
+
+/* ---------------------------------------------------------------------- */
+
+template<class DeviceType>
+FixACKS2ReaxFFKokkos<DeviceType>::
+FixACKS2ReaxFFKokkos(LAMMPS *lmp, int narg, char **arg) :
+  FixACKS2ReaxFF(lmp, narg, arg)
+{
+  kokkosable = 1;
+  atomKK = (AtomKokkos *) atom;
+  execution_space = ExecutionSpaceFromDevice<DeviceType>::space;
+
+  datamask_read = X_MASK | V_MASK | F_MASK | MASK_MASK | Q_MASK | TYPE_MASK | TAG_MASK;
+  datamask_modify = Q_MASK | X_MASK;
+
+  nmax = m_cap = 0;
+  allocated_flag = 0;
+  nprev = 4;
+
+  memory->destroy(s_hist);
+  memory->destroy(s_hist_X);
+  memory->destroy(s_hist_last);
+  grow_arrays(atom->nmax);
+  memoryKK->create_kokkos(k_s_hist_last,s_hist_last,2,nprev,"acks2/reax:s_hist_last");
+  d_s_hist_last = k_s_hist_last.template view<DeviceType>();
+
+  d_mfill_offset = typename AT::t_int_scalar("acks2/kk:mfill_offset");
+
+  comm_me_0_flag = (comm->me == 0);
+}
+
+/* ---------------------------------------------------------------------- */
+
+template<class DeviceType>
+FixACKS2ReaxFFKokkos<DeviceType>::~FixACKS2ReaxFFKokkos()
+{
+  if (copymode) return;
+
+  memoryKK->destroy_kokkos(k_s_hist,s_hist);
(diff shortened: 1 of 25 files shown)


stanmoore1:
I'm debugging Kokkos ACKS2, everything else is working as far as I can tell.

src/KOKKOS/fix_acks2_reaxff_kokkos.cpp
```diff
@@ -16,23 +17,25 @@
...
+#include "neigh_list_kokkos.h"
+#include "neigh_request.h"
+#include "neighbor.h"
 #include "pair_reaxff_kokkos.h"
+#include "update.h"
+
+#include <cmath>
 
-using namespace LAMMPS_NS;
+sing namespace LAMMPS_NS;
```
stanmoore1, line 38: Typo?

akohlmey: yup!

stanmoore1:
I see one more small issue on GPUs, tracking it down.

akohlmey:
@stanmoore1 it would be nice if you could replace the "dummy" example that I added with something more meaningful.

stanmoore1:
OK will do.

stanmoore1:
The issue I was debugging looks like roundoff error instead--caused 1 more iteration in the solver and hence a difference set of charges within the tolerance.

stanmoore1:
@akohlmey this now passes my tests, including for GPUs.

akohlmey (approved):

hasanmetin (approved):

src/fix_efield.cpp
```diff
@@ -449,3 +452,70 @@ double FixEfield::compute_vector(int n)
...
+          domain->unmap(x[i],image[i],unwrap);
+          energy[i] -= fx*unwrap[0] + fy*unwrap[1] + fz*unwrap[2];
+          if (fabs(fx*(x[i][0]-unwrap[0])) + fabs(fy*(x[i][1]-unwrap[1])) +
+              fabs(fz*(x[i][2]-unwrap[2])) > 0.0)
+            warn_flag_local = 1;
+        }
+      }
+    }
+  } else {
+    error->all(FLERR,"Cannot yet use fix qeq/reaxff with variable efield");
```
stanmoore1, line 512: @akohlmey @sjplimp is is OK to have this QEq specific code in `fix_efield.cpp`? This is the only sticking point I can think of in this PR... I could move this into `fix_qeq` but it wouldn't be very convenient.

akohlmey: @stanmoore1 I am not so much concerned about the fact that this function is called (there should be some more detailed comments about what it does and when it is called by what), but I am more worried about using unwrapped coordinates. 
Having an electric field across PBC is a bit of a problem. An atom that will pass through the PBC will see a jump in the potential, but if you use unwrapped coordinates the atoms will disperse and atoms that are on one level neighbors are not on the other. With planewave DFT codes the "solution" usually is to assume a sawtooth shaped potential and require a vacuum gap in where the potential "snaps" back and no atoms or electrons may go. But even assuming the approach used here is documented and its inconsistency accepted, I don't understand why there then have to be warnings for non-zero image flags (and why that is tested for in such a complex way)? For PBC this is practically a given to have non-zero image flags and for non-PBC it is not a problem (we already warn for that and zero them), and unlike planewave DFT we don't need to force PBC for a slab system with ReaxFF because it doesn't use kspace. So I am wondering, am I misssing something?

stanmoore1: This electric field code should not be used with ReaxFF when there is a periodic boundary in the field direction, hence the warning. Are you saying you zero image flags whenever a periodic BC is converted into a non-periodic BC?

stanmoore1: If LAMMPS always zeros out image flags when we can just convert the check into a check for PBC in the field direction...

akohlmey: I think that is what happens when you use read_data and change_box. This was changed not so long ago, though. But if you want to disallow the use of the field in the way you describe, having a warning here is not a good solution. Starting with the fact that it is difficult to interpret the warning and that it may accumulate a lot of output. How about adding an "extract" function and being able to query the fix for the field vector components and whether it is constant or not and then make a check for PBC (via `domain->periodic` or similar) during Pair::init_style() in case the field fix is active? Then the calculation could simply terminate with a more descriptive error message ("must not have periodicity in ? direction with field vector <x,y,z>").

stanmoore1: Yes, I think this is a better solution.

akohlmey: @stanmoore1 do you want me to try and implement this or would prefer to do this yourself?

stanmoore1: If you have cycles to work on this, that would be great, otherwise I can do it. I think adding an "extract" method to `fix efield` as you suggested and then moving all the field code into fix `qeq/reaxff` is the cleanest. It shouldn't require any unwrapped coordinates if we can confirm that image flags cannot be non-zero if a PBC is change to a non-PBC, then add an error check for PBC in the same direction as the field.

akohlmey: I am not subject to the kind of constraints that people in national labs are. That is why I asked in the first place. In my case this just means something else has to wait a little longer. My list of things that I could work on is always very long... Just look at the "feature request" issues that I created recently. and that is only the tip of the iceberg. :wink:

akohlmey: Thus I will have a look at this. Perhaps over the weekend. I don't want this PR to linger around much longer because of some of the more sweeping changes planned (see e.g. issue #2946) that will then create extra work merging pull requests, but already merged changes will be processed "for free". Converting the PR back to a draft for the time being.

stanmoore1: Thanks!

stanmoore1:
Thanks @akohlmey! I will do a final check tomorrow morning, then it should be ready to merge.

stanmoore1:
Thanks. This is ready to merge from my point of view.

sjplimp:
Just to clarify, you are talking about a restriction for ReaxFF QEq to use a single fix efield.  Not a restriction on fix efield to only be used once (in the absence of ReaxFF) ??

stanmoore1:
@sjplimp correct.

akohlmey:
@stanmoore1 @sjplimp 
- force style unit tests are corrected
- a check for a single instance of fix efield has been added
- the restrictions for using fix efield have been documented in the manual (and the fact that it is supported)
- a check and a warning for the presence of fix efield has been added to the qeq fixes from the QEQ package. In that process also some shared code was moved to the base class and the efield restriction documented.

PR #2940, opened 2021-09-11
Support multi-config builds with CMake

akohlmey:
**Summary**

This PR adds support for multi-config build tools like Ninja and the MS Visual Studio builder.
In the process this PR also changes how external libraries that support CMake are integrated in to LAMMPS:
Rather than using `ExternalProject_add()`, which is not compatible with multi-configuration builds, we provide our own function that includes the external package into the LAMMPS build. 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues this far.

**Implementation Notes**

Rather than building google test as an external project (for which there is no way to propagate the build configuration information) it is now downloaded and unpacked during the cmake configuration and then the
existing CMake configuration of googletest is included with add_subdirectory(). Based on suggestions from: https://google.github.io/googletest/quickstart-cmake.html#set-up-a-project
Same for MSCG (which also failed to build with `ExternalProject_add()` for cases where there were multiple libraries used for LAPACK).

**Post Submission Checklist**


rbberger (approved):

junghans (approved):

cmake/Modules/Packages/MSCG.cmake
```diff
@@ -12,34 +12,48 @@ if(DOWNLOAD_MSCG)
...
-  add_library(LAMMPS::MSCG UNKNOWN IMPORTED)
-  set_target_properties(LAMMPS::MSCG PROPERTIES
-    IMPORTED_LOCATION "${BINARY_DIR}/libmscg.a"
-    INTERFACE_INCLUDE_DIRECTORIES "${SOURCE_DIR}/src"
-    INTERFACE_LINK_LIBRARIES "${LAPACK_LIBRARIES}")
-  target_link_libraries(lammps PRIVATE LAMMPS::MSCG)
-  add_dependencies(LAMMPS::MSCG mscg_build)
+  # always compile a static lib but with position independent code
+  # make a copy of current settings for later use
+  set(OLD_SHARED_LIBS ${BUILD_SHARED_LIBS})
```
junghans, line 17: I think the more elegant solution would be to define a CMake function and overwrite these 4 variables in there. Variable values are scoped inside the function once the function ends variables are restored.

Also the download & unpack part is used 3 times already, so it might be worth a function anyhow.

akohlmey: Thanks for the suggestion. Was trying to work around some limitations for "ExternalProject_add()":
- it doesn't seem to propagate the current configuration with multi-configuration build systems (e.g. ninja multi-config)
- it doesn't seem to be possible to pass BLAS or LAPACK library definitions unless it is a single library file. With "flexiblas" on Fedora, building any such external project breaks with an error like "unknown command '-ldl'". So it seems in the long run, the only clean way to have external CMake based libraries included is by using "add_subdirectory()", but then the issue is that we want the library code to be included in liblammps.so.0, so that there is less of a mess of n+1 shared libraries that need to be kept alongside liblammps.so.0. hence the hackery with turning shared/static on/off. not a very satisfactory solution...

Not sure if this is the right way to move forward at all. It just came out of some experiments I did on a different branch and I didn't want to "lose" the changes for multi-config and passing flexiblas style LAPACK libs.

PR #2951, opened 2021-09-20
Add utility function to parse boolean parameters

akohlmey:
**Summary**

Add a convenience function `utils::logical()` similar to `utils::numeric()` that will parse "yes/no" values but also supports equivalent variants like "on/off", "true/false" and "1/0". This will both simplify the code parsing commands and at the same time resolve some inconsistencies where some keywords would accept "yes/no" when others required "on/off", in some cases even within the same command (e.g. `package gpu`).

This pull request also applies the new function to suitable commands.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues. Valid input should remain valid. Error messages for invalid input will change from "illegal XXX command" to "expected boolean parameter instead of 'yyy' in input or data file".

**Implementation Notes**

The added unit tests also add explicit tests for the "numeric" functions.
Despite the new function, documentation, and added unit tests, this PR removes more lines than it adds. :smile: 

**Post Submission Checklist**


sjplimp (approved):

akohlmey:
@sjplimp this currently disregards capitalization, but after going over some the commands to use the new function and also comparing to the constants used in the variable command expressions, it may be more consistent to only accept lowercase. Any thoughts?

athomps (approved):
I like this
'''
      mtk_flag = utils::logical(FLERR,arg[iarg+1],false,lmp);
'''

stanmoore1 (approved):

PR #2952, opened 2021-09-22
Final changes and bugfixes for the stable release

### LINKED ISSUE #2954: [BUG] mismatch in path for pace style
**Summary**

directions (for Makefile style) compilation of pace pair_style lead to the actual code in `lib/pace/src/USER-PACE`, but the compilation lines look for it in `lib/pace/src/ML-PACE`. 

git master branch 8d8c710982, Linux

**Expected Behavior**

successfully compile lammps

**Actual Behavior**

`make mode=shlib mymachine` ends with the following output:
```
g++ -g -O0  -fPIC -DLAMMPS_GZIP -DLAMMPS_MEMALIGN=64 -DLAMMPS_EXCEPTIONS  -I../STUBS    -I../../lib/pace/src/ML-PACE  -c ../pair_lj_mdf.cpp
../pair_pace.cpp:46:10: fatal error: ace_evaluator.h: No such file or directory
 #include "ace_evaluator.h"
          ^~~~~~~~~~~~~~~~~
compilation terminated.
```

Checking the content of the `lib/pace` subdirectory shows
```
> ls -l ../lib/pace/src/
total 44
-rw------- 1 bernstei bernstei 35149 Apr  9 07:08 LICENSE
-rw------- 1 bernstei bernstei  1342 Apr  9 07:08 README.md
drwx------ 2 bernstei bernstei  4096 Apr  9 07:08 USER-PACE/
```
**Steps to Reproduce**

```
cd lammps
git checkout master
git pull
cd src
make pu
make clean-mymachine
rm liblammps*
rm lmp_mymachine
make lib-pace
make lib-pace "args=-b"
make yes-ml-pace
make mode=shlib mymachine
```

akohlmey:
**Summary**

This pull request collects changes to take the 20 September 2021 patch release to the next stable release.

**Related Issue(s)**

Fixes #2954

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

The following individual change sets are included:
- Fixes build script bugs in traditional make procedure that kept the ML-PACE package and libs from being built correctly
- Works around an issue in the GPU package where the GPU neighborlist build would incorrectly report neighbor list overflows for sparse systems
- Fix issues detected by running the example inputs with an instrumented LAMMPS binary that creates floating point exceptions and warnings about undefined behavior:
   - do not call memset() in read_data on null pointers, even if the length is 0 (required by the standard, glibc's memset() does not crash)
   - fix an underflow issue in fix dt/reset where a division by zero could happen when a>0 and b>0 but a*b == 0
- Fixes to files in the GPU package library as well as the DPD-REACT, DRUDE, and KOKKOS packages to make them compatible with C++20 compilers
- Add check to the GPU package code to bail out early with a meaningful error message when compiled for mixed or double precision but the GPU does not support it.
- Fix some memory leaks in fix mscg and modernize argument parsing
- The KOKKOS package shall not lower the required C++ standard if something later than C++14 is required elsewhere (CMake)

**Post Submission Checklist**


cmake/Modules/Packages/KOKKOS.cmake
```diff
@@ -1,6 +1,8 @@
 ########################################################################
 # As of version 3.3.0 Kokkos requires C++14
-set(CMAKE_CXX_STANDARD 14)
+if(CMAKE_CXX_STANDARD LESS 14)
```
akohlmey, line 3: @junghans @rbberger this is a "quick hack workaround" for the stable release, but not a satisfactory solution. It seems that when globally the standard is set to c++11 some files are compiled with -std=c++11 and then the remainder with -std=c++14. This only is to avoid having this to lower the standard required globally. So probably this statement would have to come where the standard is set globally and also bracketed by "if(PKG_KOKKOS)". Is there some other way in CMake to change a "global" setting retroactively from within some some included script code?

src/DPD-REACT/random_external_state.h
```diff
@@ -78,8 +78,8 @@
 namespace random_external_state {
 typedef uint64_t es_RNG_t;
 
-enum { MAX_URAND = 0xffffffffU };
-enum { MAX_URAND64 = 0xffffffffffffffffULL - 1 };
+constexpr uint32_t MAX_URAND = 0xffffffffU;
+constexpr uint64_t MAX_URAND64 = 0xffffffffffffffffULL - 1;
```
akohlmey, line 82: @stanmoore1 this may impact the KOKKOS version of the DPD-REACT package, so I am a little uncomfortable making this change now. However, I want to make certain that somebody can compile LAMMPS when writing new code that requires the latest C++ standard and C++20 is much more strict with respect to types returned by enums (which are usually left to the discretion of the compiler). There are some similar, but not "deadly" issues within Kokkos source code (some .hpp files) as well. Will have to contact the Kokkos folks about that.

src/KOKKOS/kokkos_type.h
```diff
@@ -25,7 +25,9 @@
 #include <Kokkos_ScatterView.hpp>
 #include <Kokkos_UnorderedMap.hpp>
 
-enum{FULL=1u,HALFTHREAD=2u,HALF=4u};
+constexpr unsigned int FULL = 1u;
```
akohlmey, line 28: @stanmoore1 this is even more critical than the DPD-REACT change. Should this be `unsigned int` or `unsigned char`? or something else. By using constexpr it becomes a compile time constant, so the compiler may adjust the type if needed where this is used.

stanmoore1: @akohlmey it should be `unsigned int` since we compare to `int neighflag`.

src/DPD-REACT/random_external_state.h
stanmoore1: @akohlmey I don't see those enums used anywhere in the KOKKOS version. KOKKOS uses its own pRNG so I think we are OK. I also have decent internal regression tests for DPD-REACT including for KOKKOS.

stanmoore1: It looks like that file was adapted from an old version of Kokkos, but it isn't used by the LAMMPS KOKKOS package. The file in the Kokkos repo has been updated, see https://github.com/kokkos/kokkos/blob/master/algorithms/src/Kokkos_Random.hpp#L725.

akohlmey: @stanmoore1 thanks a lot for the feedback. After some consideration, both the changes to the enumerators feel too risky for the stable release. I'd rather have not all of the stable version of LAMMPS compatible with a C+20 or later mode compiler than some unexpected side effects for which we don't have enough time to see if they show up. So I will revert those now and reapply later. The other C++20 related changes are corrections for code that is not 100% correct even at C++11 level, so those will remain. It was only the more strict rules in C++20 that required them instead of tolerating the small deviation.

src/KOKKOS/kokkos_type.h
akohlmey: @stanmoore1 Thanks, again. Any particular reason why it is unsigned? It doesn't need the extra bit and CPUs and compilers "like" signed integers more than unsigned and you can even see performance penalties when mixing them.
Thus I've been trained to "use signed unless you required unsigned".

cmake/Modules/Packages/KOKKOS.cmake
junghans: Is it a problem to mix C++11 lammps with C++14 kokkos?

src/KOKKOS/kokkos_type.h
stanmoore1: Not sure, I believe this was added by Christian Trott before I started working on Kokkos. So if signed int is better that is fine with me.

cmake/Modules/Packages/KOKKOS.cmake
akohlmey: I don't expect that there is a problem, but it is an inconsistent behavior because C++11 is applied to all sources "processed" before the standard is switched by KOKKOS.cmake. 

On top of that, all C++ code linked together (i.e. using the same headers) is supposed to be compiled with consistent settings with respect to C++ (e.g. exceptions and rtti and alike), so by switching midway this "rule" is violated (which also triggers concerns about including external packages like an external Kokkos installation. Those likely will have to be compiled and use with consistent settings). Now I don't think that is a problem in practice (yet), but more sweeping changes, like those introduced by C++20 may be more sensitive to that. So effectively it looks to me as if adding the if statement here helps by not lowering the standard for files where the later standard is expected, but if LAMMPS requires "only" C++11 it may cause undefined behavior (probably not for C++11 vs C++14 but perhaps if Kokkos or some other future package would require a much later standard).

In short, for the stable release this is acceptable, but in general, some thinking may be needed. The same concern about consistent settings would also apply to the case of using and external project versus fetch content and adding the subdirectory like in the multi-config PR. When I read up on multi-config build systems, there was significant discussions about having the same config setting (Debug, Release, etc.) for all "modules" of a compilation in the same sense as I mentioned above.

junghans: I see, that is because we `include(KOKKOS.cmake)` instead of using a local CMake scope like a subdirectory, function or target property, I guess we could apply C++14 only on the kokkos files.

Your change here looks ok, but I guess setting the standard in the main CMake file might be more obvious even if  it is inside a `if(PKG_KOKKOS)` block.

akohlmey: That is exactly my thinking and the change I am preparing for later. I think overall we'll be better off to have the same C++ standard setting for all files, even if it puts some extra effort on testing and how to write code.

sjplimp (approved):

PR #2958, opened 2021-09-27
nm split styles

jddietz:
**Summary**

Style “nm/cut/split” is a derived class of style “nm/cut/“.  Both styles compute monomer-monomer interactions based on the generalized Lennard-Jones (LJ) potential; however, unlike “nm/cut/“, “nm/cut/split” applies the standard LJ (n=12-m=6) potential above r=2^(1/6) (i.e., when forces are attractive). Note n modifies the strength of the repulsive component of the LJ potential, while m modifies the attractive component. 

Similarly, style “fene/nm/split” is a derived class of the standard bond style “fene” command which consists of two components: 1) an attractive logarithmic term and 2) a repulsive Lennard-Jones (n=12,m=6) term. 

The “fene/nm/split” style goes hand-in-hand with style “nm/cut/split”; it replaces the usual LJ (6-12) potential with the generalized LJ term, allowing users to modify the parameters n and m.

Used in parallel, “nm/cut/split” and “fene/nm/split” will allow users to combine double-bridging-like Monte-Carlo bond swapping with a DPD-like core-softening to equilibrate Kremer-Grest melts of any chain stiffness for which the melt remains isotropic [1].

Collectively, these new features are small modifications to the already existing pair and bond styles. For example, pair_nm_cut_split.cpp only modifies the compute and single functions to include the additional Lennard-Jones (n=12, m=6) term above r=2^(1/6), where sigma is set to 1.0. On the other hand, in addition to modifying the compute and single functions, bond_fene_nm_split.cpp  required additional modifications to all other functions to allow users to input values for n and m. 

Other modifications include changing the “rlogarg” cutoff from 0.1 to 0.02 so only bond lengths l greater than 1.485 give the warning and crash the run if rlogarg < -.21 instead of r > 2 r_0. The structure/syntax for these programs are consistent with the already existing pair and bond styles. 

**Related Issue(s)**

**Author(s)**
Joseph D. Dietz (josephfox@usf.edu) , Robert S. Hoy (rshoy@usf.edu), University of South Florida

**Licensing**

**Backward Compatibility**
“fene/nm/split” is compatible with the current unstable version of lammps. Because of this, commands such as “using MathConst::MY_CUBEROOT2;” must be modified to be compatible with the latest stable version.  In contrast, “nm/cut/split”  is backward compatible. 

**Implementation Notes**

In order to utilize these pair and bond styles, relevant packages: MOLECULE, EXTRA-PAIR, EXTRA-MOLECULE, must be installed. The calculated pair and bond energies were verified in our study [1], where these core-softened interactions were used with the MC package to prepare polymer melt samples. 

**Post Submission Checklist**

[1] J.D. Dietz, R. S. Hoy, “Facile equilibration of well-entangled semiflexible bead-spring polymer melts,”  arXiv:2109.11001

[DBHequilscript-kbend0.0.gz](https://github.com/lammps/lammps/files/7237594/DBHequilscript-kbend0.0.gz)
[FRInitState.kbend0.0.Nch1000.N400.bond0.961.rho0.850.theta105.8.gz](https://github.com/lammps/lammps/files/7237595/FRInitState.kbend0.0.Nch1000.N400.bond0.961.rho0.850.theta105.8.gz)


akohlmey:
@jddietz If I understand you correctly, a bond_coeff for fene/nm/split of `1 250 3   0.022 1.5 12 6` should produce the exact same value as `1 250 3   0.022 1.5` for bond style fene, should it not?

akohlmey:
@jddietz ok. I have converted the PR back to a draft. Please click on "Ready for review" when you are done and we should have another go at it.

jddietz:
We replaced MY_CUBEROOT2 with r0[itype][jtype]*r0[itype][jtype] everywhere in pair_nm_cut_split.cpp, and replaced MY_CUBEROOT2 with sigma[type]*sigma[type] everywhere in bond_fene_nm_split.cpp. This makes the code do the exact same thing it did before provided sigma = 2^{1/6} [as is usually the case], but is generalizable to other values (i.e. real rather than reduced units). 
 
We tested this version of the code on the same set of Kremer-Grest bead-spring polymer melts systems, where we equilibrated test samples using the method discussed in [1]. We then compared these results to previously verified values. The MC bond swap success rate was the primary metric we used to verify correctness, as this quantity is susceptible to the changes in the parameters n an m.

akohlmey:
@jddietz please note that you DOS style text files which makes it impossible to see changes because **every** line is different due to replacing LF with CR-LF. I've converted it back. Also updated the branch to the current development version.

You still also need to update the documentation to correctly describe the implemented behavior. I think it would be important in general to point out the r0 parameter is not the LJ sigma in LAMMPS conventions (which is the distance for E == 0) but the minimum of the potential (i.e. `2^(1/6)*sigma` for a 12-6 LJ).

jddietz:
> @jddietz please note that you DOS style text files which makes it impossible to see changes because **every** line is different due to replacing LF with CR-LF. I've converted it back. Also updated the branch to the current development version.
> 
> You still also need to update the documentation to correctly describe the implemented behavior. I think it would be important in general to point out the r0 parameter is not the LJ sigma in LAMMPS conventions (which is the distance for E == 0) but the minimum of the potential (i.e. `2^(1/6)*sigma` for a 12-6 LJ).

Hi Axel, Thank you for the update.  I finished drafting the updated documentation and I am waiting on my advisor to take a look at it before I commit the files. I included your comment about the r0 parameter.

akohlmey:
@jddietz I've been reviewing your latest changes and made some updates. I've also chosen to rename bond style `fene/nm/split` to bond style `fene/nm` since there is no "split" in there. I also clarified the documentation to be consistent with the documentation for lj/cut and nm/cut and made the differences between `r_0` and `sigma` more visible.

Also updated the unit tests accordingly

jddietz:
@akohlmey, thank you very much for taking the time to make those changes and updates. I was looking through your updates and I noticed that "semiflexible" was marked as a spelling error; however, this is the correct spelling. Most papers dealing with "semiflexible polymers" utilize this convention.

akohlmey:
> I was looking through your updates and I noticed that "semiflexible" was marked as a spelling error; however, this is the correct spelling. Most papers dealing with "semiflexible polymers" utilize this convention.

I see this similar to british versus american spelling. In the interest of keeping the false-positives file from growing more than needed we have adopted the conventions of the underlying spell checker. We also remove/replace extended ASCII or UTF-8 characters (like umlauts, characters with accents, non-ASCII punctuation and blanks).

Please note, that strictly enforcing the LAMMPS conventions would mean remove the title of the paper entirely. I can do that, too. ;-)

jddietz:
@akohlmey, Oh I see, thank you for the clarification, the hyphen doesn't seem so bad after all. In any case, we hope to have our paper published in J. Chem. Phys. soon, as It is currently under review. We can then remove the title and replace it with the proper journal citation. Thanks again for your help!

sjplimp (approved):

PR #2962, opened 2021-09-27
A few final updates to the LAMMPS manual

akohlmey:
**Summary**

A few minor updates to the manual.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

### DIFF
--- a/doc/src/Run_basics.rst
+++ b/doc/src/Run_basics.rst
@@ -2,15 +2,22 @@ Basics of running LAMMPS
 ========================
 
 LAMMPS is run from the command line, reading commands from a file via
-the -in command line flag, or from standard input.
-Using the "-in in.file" variant is recommended:
+the -in command line flag, or from standard input.  Using the "-in
+in.file" variant is recommended.  The name of the LAMMPS executable is
+either ``lmp`` or ``lmp_<machine>`` with `<machine>` being the machine
+string used when compiling LAMMPS.  This is required when compiling
+LAMMPS with the traditional build system (e.g. with ``make mpi``),
+but optional when using CMake to configure and build LAMMPS:
 
 .. code-block:: bash
 
    $ lmp_serial -in in.file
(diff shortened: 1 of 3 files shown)


doc/src/Run_basics.rst
```diff
@@ -2,15 +2,22 @@ Basics of running LAMMPS
...
 
 LAMMPS is run from the command line, reading commands from a file via
-the -in command line flag, or from standard input.
-Using the "-in in.file" variant is recommended:
+the -in command line flag, or from standard input.  Using the "-in
+in.file" variant is recommended.  The name of the LAMMPS executable is
+either ``lmp`` or ``lmp_<machine>`` with `<machine>` being the machine
+string used when compiling LAMMPS.  This is required when compiling
+LAMMPS with the traditional build system (e.g. with ``make mpi``),
+but optional when using CMake to configure and build LAMMPS:
```
stanmoore1, line 10: May want to explain why `-in in.file` is preferred: 

In https://docs.lammps.org/Run_basics.html it has this note:


`The redirection operator “<” will not always work when running in parallel with mpirun; for those systems the -in form is required.`

stanmoore1 (approved):

doc/src/Run_basics.rst
stanmoore1: Actually this is in the same section (didn't realize that), so could add something like `see note below`.

akohlmey: If it was my choice we would drop that entirely. This is effectively a leftover from the Fortran days when there was no portable way to process command line flags and thus people used I/O redirection to read input files. No longer supporting I/O redirection would make some of the more convoluted code in the LAMMPS class significantly simpler. But some people's habits are hard to break, especially if they go back > 25 years. ;-)

sjplimp (approved):

PR #2966, opened 2021-09-29
Tweaks to CMake build for portability and early detection of build problems

### LINKED ISSUE #2964: [BUG] LATTE.cmake yielding undefined reference... when building with MKL
**Summary**

When building LAMMPS with LATTE and MKL, the build command gives "undefined reference to `mkl_...'" error messages during the linking LATTE.

In the cmake options for LATTE in
  `lammps/cmake/Modules/Packages/LATTE.cmake`
LAMMPS passes the BLAS and LAPACK libraries to LATTE as following
  `-DBLAS_LIBRARIES=${BLAS_LIBRARIES} -DLAPACK_LIBRARIES=${LAPACK_LIBRARIES}`
where ${BLAS_LIBRARIES} and ${LAPACK_LIBRARIES} are lists of multiple ".so" files and link options such as -lm, separated by spaces.
During the LATTE link procedure, the linker takes only the first element in the list as the relevant library.
For example, if
  `-DBLAS_LIBRARIES=libmkl_intel_lp64.so libmkl_intel_thread.so libmkl_core.so libiomp5.so -lm -ldl -lm -ldl`
LATTE links only with "libmkl_intel_lp64.so".
As a result, it gives "undefined reference" errors for the functions defined in other libraries, such as libmkl_core.so.

The current master branch of LAMMPS. 

**Expected Behavior**

Build completion without error.

**Actual Behavior**

Build failure with the following error messages:
```
libmkl_intel_lp64.so: undefined reference to `mkl_lapack_zgemlqt'
libmkl_intel_lp64.so: undefined reference to `mkl_spblas_lp64_mkl_zcsrcsc'
libmkl_intel_lp64.so: undefined reference to `mkl_graph_descriptor_destroy_internal'
libmkl_intel_lp64.so: undefined reference to `mkl_spblas_lp64_mkl_scscsm'
...
```

**Steps to Reproduce**

I tested the build only on LANL Kodiak cluster.

```
git clone https://github.com/lammps/lammps.git
mkdir build
cd build

module purge
module load cmake python intel/19.1.3 intel-mpi/2019.9.304 mkl/2020.0.4

CC=mpiicc CXX=mpiicpc \
cmake ../lammps/cmake \
  -DCMAKE_INSTALL_PREFIX=${INSTALL_DIR} \
  -DCMAKE_BUILD_TYPE=RelWithDebInfo \
  -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=OFF \
  -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON \
  -DBUILD_MPI=ON \
  -DBUILD_OMP=ON \
  -DPKG_LATTE=ON
```

Attached are the cmake and make logs.

[make.log](https://github.com/lammps/lammps/files/7247093/make.log)
[cmake.log](https://github.com/lammps/lammps/files/7247094/cmake.log)

### LINKED ISSUE #2967: [BUG] const issues initializing lammps with the C API
**Summary**

C-API example in https://docs.lammps.org/Library_create.html is incompatible with modern (strict?) compilers with respect to const variables/arguments

github master 8d8c7109

**Expected Behavior**

Example program compiling as is

**Actual Behavior**

With the example program as given, attempting to pass the `const char *argv[]` to `lammps_open_no_mpi` gives the error
```
atoms.h:215:61: error: invalid conversion from ‘const char**’ to ‘char**’ [-fpermissive]
  215 |                 calc = (void *) lammps_open_no_mpi(lmpargc, lmpargv, NULL);
      |                                                             ^~~~~~~
      |                                                             |
      |                                                             const char**
```

Removing the `const` breaks (or at least causes a warning) the initializer
```
atoms.h:212:37: warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]
  212 |                 char *lmpargv[] = { "liblammps", "-log", "none"};
      |                                     ^~~~~~~~~~~
```

It would be nice if the example could be updated to work cleanly with modern (and strict) standards

akohlmey:
**Summary**

CMake script portability tweaks: 
- add checks for required commands to patch a downloaded source (patch and sed)
- use cmake command mode to copy file instead of 'cp'
- allow compilation of Tcl wrapper if only Tcl is available but not Tk, support Tcl stub library with newer Tcl versions
- Bonus changes: small corrections in the documentation ;-)

**Related Issue(s)**

Closes #2964 
Fixes #2967

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Implementation Notes**

The ExternalProject_add() function apparently cannot pass lists to variables for the external project's CMake configuration. The only viable solution has been to use add_subdirectory() as in PR #2940 which is also required to pass the configuration selection for multi-config build environment (like the visual studio builder). This will have to be addressed when we get ready to make LAMMPS portable to visual studio compilation.

**Post Submission Checklist**


cmake/Modules/Packages/LATTE.cmake
```diff
@@ -19,6 +19,14 @@ if(DOWNLOAD_LATTE)
   set(LATTE_MD5 "820e73a457ced178c08c71389a385de7" CACHE STRING "MD5 checksum of LATTE tarball")
   mark_as_advanced(LATTE_URL)
   mark_as_advanced(LATTE_MD5)
+
+  # CMake cannot pass BLAS or LAPACK library variable to external project if they are a list
+  list(LENGTH BLAS_LIBRARIES} NUM_BLAS)
+  list(LENGTH LAPACK_LIBRARIES NUM_LAPACK)
+  if((NUM_BLAS GREATER 1) OR (NUM_LAPACK GREATER 1))
+    message(FATAL_ERROR "Cannot compile downloaded LATTE library due to a technical limitation")
```
junghans, line 27: Not a super useful error message.

akohlmey: Suggest a better one.

akohlmey: That is just as confusing because this is what this is attempting to do. Since this is only a stopgap measure it is still significantly better than getting an error about not finding the command "-dl" or something similarly random. At least people will than know who and what to ask and an explanation can be provided. The hope is that this will affect only a small number of people. It wasn't a problem on Fedora until it introduced "flexiblas".

junghans: My problem is that the error message that no real explaination what is wrong, maybe just adding a link to https://github.com/lammps/lammps/issues/2964 would help.

akohlmey: I cannot think of a one line explanation. Otherwise it would be there.

rbberger (approved):

PR #2984, opened 2021-10-05
Add new group arg for delete_atoms porosity

sjplimp:
**Summary**

Add a groupID arg to delete_atoms porosity to allow more general definitions of eligible atoms to delete.
Now both a groupID and regionID are used to specify eligible atoms to delete a fraction of.  Using group = all with
a regionID reproduces the old behavior.  Using a group with region = NULL allows deletion of a fraction of group.
Using both a groupID and regionID is also now possible.

**Related Issue(s)**

N/A

**Author(s)**

Andre Archer (U Florida) and Steve

**Licensing**

**Backward Compatibility**

This adds a required arg to delete_atoms porosity, so will break existing scripts.  But it is easy to add the new arg (group = all) to
reproduce what existing scripts did. 

**Implementation Notes**

**Post Submission Checklist**


sjplimp:
@akohlmey Andre is going to do some testing with this before considering it for mergin

akohlmey (approved):

akohlmey:
@sjplimp FYI, I've updated the accessor functions for looking up region styles to use the new API.

PR #2996, opened 2021-10-15
Add compute ave/sphere/atom

stanmoore1:
**Summary**

Add new `compute ave/sphere/atom` (along with Kokkos version) that computes local density and temperature using neighbors of atom i inside a sphere (cutoff distance specified by user).

**Related Issue(s)**

None

**Author(s)**

Stan Moore (SNL), original concept from Mitch Wood (SNL)

**Licensing**

**Backward Compatibility**

Yes

**Implementation Notes**

Verified by comparing to OVITO

**Post Submission Checklist**


stanmoore1:
@sjplimp @akohlmey this is ready to merge from my point of view. Future work could include adding pressure or even making a general spherical averaging compute that can take in other computes/fixes, similar to `compute chunk/atom`.

akohlmey:
@sjplimp should this be in src or rather in EXTRA-COMPUTE? I am leaning toward the latter.

Also, I am not overly happy with the name of this compute. I am concerned that there would be significant potential for people misunderstanding what it does and then asking confusing questions or being confused by the output. On the other hand, I cannot think of anything that is short and would be better. The best I can think of would be "local/phase/atom".

akohlmey:
> The best I can think of would be "local/phase/atom".

I just need to complain and a better idea comes to mind. :-(
If this **will** be expanded to perform local averaging of data, then a) it could be called `localavg/atom` and then it would also be justified to leave it in src.

stanmoore1:
>If this will be expanded to perform local averaging of data, then a) it could be called localavg/atom and then it would also be justified to leave it in src.

I would prefer `spherical/atom` or something like that.

sjplimp:
How about compute ave/sphere/atom, similar to the fix ave/* commands.

akohlmey:
> How about compute ave/sphere/atom, similar to the fix ave/* commands.

That would work for me. A bit verbose, but allows to include variants that e.g. average over other shapes.

stanmoore1:
I'm fine with `ave/sphere/atom` and I'm fine if it goes into `EXTRA-COMPUTE` for now. Should I rename the files?

stanmoore1:
What about `compute property/sphere`?

sjplimp:
It has to end in /atom b/c it produces a per-atom quantity.  I think ave/sphere/atom is more descriptive.

stanmoore1:
Sounds good. Should it go in `EXTRA-COMPUTE` or stay in `src`? @akohlmey what is the criteria for `EXTRA-COMPUTE` anyway, seems sort of arbitrary, similar to USER packages...

akohlmey:
> Sounds good. Should it go in `EXTRA-COMPUTE` or stay in `src`? @akohlmey what is the criteria for `EXTRA-COMPUTE` anyway, seems sort of arbitrary, similar to USER packages...

It is a bit of a sliding scale and based on personal preferences. In src are base and commonly used functionality. In the EXTRA- folders more unusual ones. E.g. compute rdf is in src and compute adf is in EXTRA-COMPUTE.
This one close to the boundary but I think putting it in EXTRA-COMPITE will be better for now. If this becomes a frequently used feature, we can move it any time.

sjplimp:
The motivation for EXTRA was simply to keep the src dir small.  I'd say EXTRA-COMPUTE for now.  We can always
move it into src later, or make a AVE-SPHERE package if there becomes several related computes.

stanmoore1:
OK, this will create a new dependency between KOKKOS and EXTRA-COMPUTE, that is OK right?

stanmoore1:
@akohlmey @sjplimp I made the changes, can you please review? Thanks

sjplimp (approved):

PR #3001, opened 2021-10-20
Add accessor functions to `Modify` and `Domain` that do not require using class internal data structures

akohlmey:
**Summary**

This pull request adds multiple accessor functions that either return a pointer or a vector of pointers:
- `Modify::get_fix_by_id()`
- `Modify::get_fix_by_index()`
- `Modify::get_fix_by_style()`
- `Modify::get_fix_list()`
- `Modify::get_compute_by_id()`
- `Modify::get_compute_by_index()`
- `Modify::get_compute_by_style()`
- `Modify::get_compute_list()`
- `Domain::get_region_by_style()`

Those are intended to eventually replace the corresponding "find" functions, where meaningful. This typically reduces the number of lines of code that need to be written and the `_by_id()` and `_by_style()` names are more consistent, plus having to loop over internal data structures of the Modify and Domain classes is avoided thus producing more maintainable code.
The `_by_index()` functions are equivalent to accessing the corresponding arrays directly (and implemented as inline functions).

Since the `get_*_by_style()` functions return a vector of pointers, they resolve the ambiguity of the corresponding "find" functions that would only return only the index of the first instance and allow adding a check for how many instances are found and, if necessary, process them all instead of only whatever is found first.

The pull request also contains changes to existing code to take advantage of the new APIs and some related refactoring.
The `find_xxx_by_style()` uses have been completely replaced and thus those APIs have been removed.

**Related Issue(s)**

This continues and extends the changes from PR #2869 

**Author(s)**

Axel Kohlmeyer (Temple U)

**Licensing**

**Backward Compatibility**

All unit tests pass.

**Implementation Notes**

N/A

**Post Submission Checklist**


akohlmey:
> Creates unnecessary copies. Should use `auto&` instead of `auto`.

should be using `const auto &` now.

src/fix_deposit.cpp
```diff
@@ -227,37 +227,31 @@ void FixDeposit::init()
 {
   // set index and check validity of region
 
-  iregion = domain->find_region(idregion);
-  if (iregion == -1)
-    error->all(FLERR,"Region ID for fix deposit does not exist");
+  iregion = domain->get_region_by_id(idregion);
```
rbberger, line 230: rename to `region`?

rbberger (approved):

src/fix_deposit.cpp
akohlmey: I have thought about this (there are lots more cases like that for fixes and computes, often in loops, though), but decided to keep the variable name to avoid any confusion with how the Pointers class provides access to other "base" classes (doesn't apply to fixes, computes, or regions, because there can be multiple, but still).

PR #3003, opened 2021-10-21
Fix a work group/block reduction bug in build_ItemGPU

joeatodd:
**Summary**
If part of the work group returns early, the behaviour of team_barrier() & team_reduce() are not well defined. This commit replaces the early return with 'active' guards to ensure all threads participate in the barriers & reductions.

Recommend viewing the diff with 'ignore whitespace' - the changes are actually minor ;)

**Related Issue(s)**

**Author(s)**

Joe Todd (Codeplay)

**Licensing**


joeatodd:
Hi, sorry I missed this conversation. I hadn't thought about the potential performance implications of not being able to kill off whole warps. Your proposed solution sounds good.

PR #3021, opened 2021-11-05
Allow dump sort to work with more than 2 billion atoms

stanmoore1:
**Summary**

- Allow dump sort to work with more than 2 billion (2^31) atoms
- Fix bug with dump sort and multiproc that gave wrong atom count in dump header
- Fix incorrect error message related to dump sorting
- Add check to `dump h5md` for 32-bit integer overflow

**Related Issue(s)**

None

**Author(s)**

Stan Moore (SNL)

**Licensing**

**Backward Compatibility**

Yes

**Implementation Notes**

I ran 4 billion LJ atoms for one timestep and dumped out the coordinates sorted in the z-direction. I also checked for integer overflow using `-ftrapv` and `-Wconversion`.


stanmoore1:
I checked other dump styles and can't see any more issues.

stanmoore1:
I tested both sorting by id and by coordinate for 4 billion atoms. This is ready for review/merge from my point of view.

sjplimp (approved):

PR #3031, opened 2021-11-18
Second Moment Aproximation to the Tight Binding addes as pair style

Iximiel:
**Summary**

I added the Second Moment Aproximation to the Tight Binding (SMATB) potential as a pair style, along with some very simple examples and the documentation to start a simulation with this potential.

**Author(s)**

Daniele Rapetti - Politecnico di Torino - daniele.rapetti@polito.it (personal: iximiel(at)gmail(dot)com)

**Licensing**

**Post Submission Checklist**


akohlmey:
@Iximiel thanks for your submission. Before looking at the implementation and making some (required) suggestions, can you please discuss how your implementation differs from what functionality is available in the [SMTBQ package](https://docs.lammps.org/Packages_details.html#pkg-smtbq)
with its [pair style smtbq](https://docs.lammps.org/pair_smtbq.html)

Iximiel:
Thank you for the fast reply.
The main applicative difference between SMTB-Q and SMATB is that the SMTB-Q aims at simulating metal oxides while SMATB  aims at simulating metal and alloys from the transition group.
More on the physics side, even if they share partially a term of the potential, _i.e._, the covalent part of the SMTB-Q (eq. 4 in 10.1016/j.susc.2013.05.015) is similar to the bonding part of the SMATB (eq 18 in 10.1103/PhysRevB.23.6265). This similarity is obviously due to the fact we both use a second moment approximation of the TB. Another difference is the absence of the charge terms in SMATB.

Initially I thought that the acronym SMATB describes more correctly the idea behind the form of the potential. Anyway, I agree that maybe this name similarity can be misleading, and I am keen in accepting any kind of change in the keywords. A possible solution is given by the fact that the SMATB potential is also known as "Gupta" or "RGL" (according to the community that is using it).

Thank you again for your time
Daniele

akohlmey:
@Iximiel thanks for the info, that eliminates the concern that we would add duplicate functionality.
There is one general concern: we cannot accept adding your contribution to the core LAMMPS sources folder. It would have to be added to an existing package or become a new package. Due to the similarity with the smtbq pair style, it would make sense to join those pair styles in a single package. That however would require renaming the SMTBQ package (again, ugh!) so it would properly cover both implementations and also the pair styles for consistency.

@sjplimp what do you think?
- The package would be renamed from SMTBQ to SMATB
- The existing pair style from `smtbq` to `smatb/qeq` to more correctly represent that it is a second moment approximation to tight binding with charge equilibration added (of course the old pair style name would be retained as an alias for backward compatibility)
- The new pair styles in this pull request would be moved there
- The documentation for each of the package pair styles would include a note explaining the difference and referring to the other

@Iximiel once we have sorted out these "administrative matters", I will start looking at the code, examples and other details.

Please note that you added your changes to the "develop" branch in your fork. We usually encourage people to keep the branches from the upstream repository unchanged (and set the corresponding branch in the original LAMMPS repository as their upstream branch for simple synchronization with upstream) and instead work on a feature branch and submit that as a pull request. But no worries, you can correct that later. For the time being, please don't make any changes to your branch, since I plan to do the renaming and moving of files by myself and then push them into your repository. Thanks, Axel.

sjplimp:
I'm fine with a merged package and rename of the package and old pair style.  As a courtesy, running this by the original smtbq folks would be a good idea.

nsalles33:
Hi all, I implemented pair_SMTBQ in LAMMPS with Olivier Politano and Robert Tétot. We agree to merge the package SMTBQ to SMATB/QEq. 
Thanks
Nicolas Salles

Iximiel:
Hi, I am sorry for answering so late.
I agree with the proposed merging of the packages.
If anything else is needed from me, consider me at your disposal.

Thanks
Daniele

akohlmey:
> Hi, I am sorry for answering so late. I agree with the proposed merging of the packages. If anything else is needed from me, consider me at your disposal.
> 
> Thanks Daniele

@Iximiel 

No worries. We've been slow to process this as well. I've finally worked my way down the queue far enough to review this again. I've updated the code to the most recent upstream changes, made some cosmetic changes and also aligned its execution more with how things are done elsewhere in LAMMPS. Also, I've added some unit test inputs for the pair style tester tool. The good news is, that restarting works fine. The bad news is, that using newton pair off does not work. I've made some smal changes that resulted in consistent energies, but forces are off. For the time being, I have made using newton pair off an error and also told the unit test input to skip those. 

I've ultimately decided to not rename the SMTBQ package or its pair style, but just moved the new styles to the existing package and updated the documentation accordingly. This way there is less disruption and backward compatibility.

athomps (approved):
I approve.

PR #3032, opened 2021-11-18
Add dx, dy and dz computes to compute bond/local and property/local

GenieTim:
**Summary**

This PR enables outputting not only the norm of the distance between two atoms in a bond or pair, but also the three directional components.

**Related Issue(s)**

**Author(s)**

Me: Tim Bernhard, ETH Zürich, tim@bernhard.dev

**Licensing**

**Backward Compatibility**

No break.

**Implementation Notes**

The important note and possible request for feedback concerns the direction of the components. As I understand it, for the compute bond/local, the check on line 391 (and 271 for pair/local) should be sufficient to assert that the direction goes from the atom with the lower to the higher id. Is this correct?
Otherwise, the implementation relies on calculations that were already done, so not much room for error I guess.

**Post Submission Checklist**

The motivation to implement this feature is to measure end-to-end vectors (using the *new* `compute bond/local dx dy dz` with some "ghostbonds").

### DIFF
--- a/src/compute_pair_local.cpp
+++ b/src/compute_pair_local.cpp
@@ -268,7 +268,7 @@ int ComputePairLocal::compute_pairs(int flag)
         else ptr = alocal[m];
 
         // to make sure dx, dy and dz are always from the lower to the higher id
-        double directionCorrection = i > j ? -1.0 : 1.0;
+        double directionCorrection = itag > jtag ? -1.0 : 1.0;
 
         for (n = 0; n < nvalues; n++) {
           switch (pstyle[n]) {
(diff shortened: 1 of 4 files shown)


sjplimp (approved):

src/compute_pair_local.cpp
```diff
@@ -264,11 +267,20 @@ int ComputePairLocal::compute_pairs(int flag)
         if (nvalues == 1) ptr = &vlocal[m];
         else ptr = alocal[m];
 
+        // to make sure dx, dy and dz are always from the lower to the higher id
+        double directionCorrection = i > j ? -1.0 : 1.0;
```
akohlmey, line 271: Shouldn't this be using `itag` and `jtag`?

GenieTim: Thanks! As noted in the implementation notes, I was not sure about this, but makes sense. Thanks, fixed in 74577fa.

PR #3040, opened 2021-11-24

### LINKED ISSUE #3050: [BUG] Fatal error in PMPI_Wait error running local_density/benzene_water example with mpirun
**Summary**

Running `/examples/PACKAGES/local_density/benzene_water/benzene_water.in` with a longer production run results in Fatal error in PMPI_Wait. 

LAMMPS 29 Sep 2021 - Update 1
gcc/8.3.0
openmpi/4.0.4

CentOS 7

**Expected Behavior**

The simulation will successfully run for the longer run time. 

**Actual Behavior**

The following error occurs:
```
Fatal error in PMPI_Wait: Message truncated, error stack:
PMPI_Wait(219)....................: MPI_Wait(request=0x7ffdc7041888, status=0x1) failed
MPIR_Wait_impl(100)...............: fail failed
MPIDI_CH3U_Receive_data_found(131): Message from rank 30 and tag 0 truncated; 32 bytes received but buffer size is 16
```

And the following warnings appear in the standard out among regular LAMMPS output:
``` 
[1638920701.066148] [kp088:29249:0]          ib_md.c:1063 UCX  WARN  IB: ibv_fork_init() was disabled or failed, yet a fork() has been issued.
[1638920701.066165] [kp088:29249:0]          ib_md.c:1064 UCX  WARN  IB: data corruption might occur when using registered memory.

```

**Steps to Reproduce**

	1. Compile LAMMPS with MANYBODY and MOLECULE packages
	2. Cd into `/examples/PACKAGES/local_density/benzene_water/` or download and unzip benzene_water.zip below.
	3. Change the production run time (the very last line) from 1,000 steps to 1,000,000 steps. 
	4. Run the input file with mpirun with -np >= 12.
	5. Observe error.

[benzene_water.zip](https://github.com/lammps/lammps/files/7672310/benzene_water.zip)
[script_outputs.zip](https://github.com/lammps/lammps/files/7672317/script_outputs.zip)

akohlmey:
**Summary**

This pull request adds multiple small changes and bugfixes.

**Related Issue(s)**

Addresses report from lammps-users at: https://sourceforge.net/p/lammps/mailman/message/37390420/
Closes #3050

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

The following sets of changes are included:
- fix communication bug in pair style local/density, correct docs, refactor potential file parser
- make certain `Pair::did_mix` is initialized
- add restart support to fix charge/regulation
- explicitly disallow using `reset_timestep` when a `fix ave/*` is active
- create atom tags on inserted atoms, if tags are enabled
- add support for Plumed 2.7.3 and 2.6.5, update fix plumed to be compatible with Plumed 2.8.x
- small tweaks to a few manybody pair style documentation files for increased consistency

**Post Submission Checklist**


akohlmey:
@sjplimp can you please have a look at the changes to src/MANYBODY/pair_local_density.cpp?
I believe they are correct, but then I wonder how this could have worked in parallel previously at all, so a second pair of eyes would be much appreciated to make certain I am not overlooking something.
This fixes the bug(s) reported in #3050. It is much easier to reproduce the issue as in the description by adding `processors * 2 *`. Then the example will fail with 2 MPI ranks already.

src/fix_ave_atom.cpp
```diff
@@ -41,6 +41,7 @@ FixAveAtom::FixAveAtom(LAMMPS *lmp, int narg, char **arg) :
   nevery = utils::inumeric(FLERR,arg[3],false,lmp);
   nrepeat = utils::inumeric(FLERR,arg[4],false,lmp);
   peratom_freq = utils::inumeric(FLERR,arg[5],false,lmp);
+  time_depend = 1;
```
sjplimp, line 44: Should also remove the error check in end_of_step*( with message Invalid timestep reset ...
Ditto for other fix ave/*

sjplimp (changes requested):

sjplimp:
@akohlmey The pair local density changes look fine.  If the comm buffers are big enough, then understating the needed comm forward/reverse size may not cause a memory issue.  But it could, so good to fix it.

Where is the change to "create atom tags on inserted atoms, if tags are enabled".  Didn't see it.  Would like to check that.

akohlmey:
> @akohlmey The pair local density changes look fine. If the comm buffers are big enough, then understating the needed comm forward/reverse size may not cause a memory issue. But it could, so good to fix it.

but the incorrect (far too small) return values from the pack_forward() and pack_reverse() functions should also result in corrupted/missing data?
 
> Where is the change to "create atom tags on inserted atoms, if tags are enabled". Didn't see it. Would like to check that.

src/MC/fix_charge_regulation.cpp around line 992

src/fix_ave_atom.cpp
akohlmey: done

sjplimp:
Re: pair local density - I see your point now.  I wasn't seeing the code that packed the bufs in the PR view.
Yes. they do need to return m >> nLD, else missing data.

Also why isn't pack_comm named pack_forward_comm, ditto unpack?  Seems like the Comm method will not
be calling anything for forward comm?

Also, this isn't just a parallel issue.  But also PBC for ghost atoms on 1 proc.

sjplimp:
Now I see what you meant by adding tags to new atoms - looks good.  Thx

akohlmey:
> Also why isn't pack_comm named pack_forward_comm, ditto unpack? Seems like the Comm method will not be calling anything for forward comm?

didn't notice that. o dear. fixed that, too. this is one more example demonstrating why we should adopt the c++11 "override" keyword.  this pair style was **very** broken.

sjplimp (approved):

PR #1704, opened 2019-10-08
Revised implementation of a new atomic stress definition for correct computation of heat flux with many-body interactions

donatas-surblys:
**Summary**

This PR is a revised version of PR #1462 according to advice from @athomps. I created a new branch, as little code was shared with the previous PR.
Details and discussion can be found in PR #1462, but in brief, this implements a new "centroid atomic stress" definition, which enables compute heat/flux to produce correct values for many-body interactions, such as angles, dihedrals and impropers, while keeping the original atomic stress intact. This is based on the work by our group [1], and is also equivalent to a recent work and implementation by Boone et al. [2,3], although they did not refer to it as a new atomic stress definition.

**Related Issues**

Supersedes PR #1462

**Author(s)**

Donatas Surblys (Tohoku University) surblys.donatas at gmail dot com

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

Full backward compatibility.

**Implementation Notes**

`vatom` array is left untouched, and a new per-atom 3x3 "centroid" virial array, `cvatom`, is created in angle, dihedral and improper classes to store values computed by the new definition. The `pressatomflag` in compute classes can now have a new value "2", to indicate that the compute calculates per-atom centroid virial. The `vflag` in integrate.cpp or min.cpp can also have value "8" (and other combinations setting the appropriate bit), to indicate that per-atom centroid virial needs to be computed. Therefore, `cvatom` is only allocated and computed when needed.

Currently, only angles, dihedrals and impropers are implemented. This should work well enough, unless pair styles with many body interactions, such as AIREBO are used. The USER-OMP versions are also implemented. Because I needed to add an argument to the ThrOMP::ev_setup_thr, I had to update all of the pair/bond/angle/dihedral/improper files in USER-OMP, so the number of changed files is quite large.

A new compute centroid/stress/atom, that has 9 components per atom, has been constructed and is mostly identical to compute stress/atom. Compute heat/flux has been modified to accept the new centroid/stress/atom compute.

Before continuing to implement centroid virial for tersoff/AIREBO and similar, I would like to get some feedback from the developers if this approach is acceptable.

**Post Submission Checklist**

_Please check the fields below as they are completed **after** the pull request has been submitted. Delete lines that don't apply_

- [ ] The feature or features in this pull request is complete
- [ ] Licensing information is complete
[...]

PR #3046, opened 2021-12-06
Centroid atomic stress for shake, rattle and rigid/small

donatas-surblys:
**Summary**

This implements the centroid form atomic stress for constrained dynamics via shake and rattle fixes, as well as rigid body dynamics for fix rigid/small. The centroid form provides strictly correct heat flux values when used with the heat/flux compute for the whole system, and gives good approximation of local heat flux over control volumes. Without the centroid form, the original implementation gives heat flux values that don't quite match the input/output of thermal baths, like langevin and ehex fixes. Details can be found in our recently published paper [1].

If the changes look OK, I can also go on to update the documentation.

**Related Issue(s)**

This is a continuation of work done in  PR #1704.

**Author(s)**

Donatas Surblys (Tohoku University) surblys.donatas at gmail dot com

**Licensing**

**Backward Compatibility**

This might change the heat flux values obtained from shake, rattle, and rigid/small fixes if somebody was trying to use centroid/stress/atom and LAMMPS didn't properly error out.

**Implementation Notes**

New v_tally functions are defined for shake/rigid and rigid/small in fix.cpp, that internally call the original v_tally functions, and then do additional centroid stress computations if necessary.

For shake/rigid:
Constraint forces are treated as pair interactions and handled respectively.

For rigid/small:
The Body structure in fix_rigid_small.h is updated to the following variables:
```
    int natoms;            // total number of atoms in body
    double xgc[3];         // geometric center position
    double xgc_body[3];    // geometric center relative to xcm in body coords
```
This allows computation of the centroid position, enabling easily obtaining centroid form atomic stress, even when not all atoms are owned by the process. I'd like some feedback if I approached in an acceptable way, and have some slight doubts if wrapping was handled correctly.

What's still missing:
- fix rigid: Haven't really looked into it, but might be possible to use same approach as fix rigid/small. Don't know if there's much need.
- TIP4P: Need quite an involved approach. A naive implementation was not successful.
- Many-body potentials.

Notes about the original implementation of rigid/small:
Currently, fix rigid/small gives unphysical results for y an z components when used with the heat/flux compute. This is because the atomic stress virial is not averaged over all atoms in the molecule, i.e. the atomic stress tensor is not symmetric, but only the upper triangular is stored. Because of this

> jv[0]  = rijx fijx vjx + rijx fijy vjy + rijx fijz vjz
> jv[1]  = rijy fijx vjx + rijy fijy vjy + rijy fijz vjz
> jv[2]  = rijz fijx vjx + rijz fijy vjy + rijz fijz vjz

becomes

> jv[0]  = rijx fijx vjx + rijx fijy vjy + rijx fijz vjz
> jv[1]  = rijx fijy vjx + rijy fijy vjy + rijy fijz vjz
> jv[2]  = rijx fijz vjx + rijy fijz vjy + rijz fijz vjz

As you can see, only the x component is correct. In practice, the other components give obviously unphysical values.
I strongly believe there should be at least a warning cautioning against such use.

**Post Submission Checklist**

[1] https://doi.org/10.1063/5.0070930

### DIFF
--- a/src/RIGID/fix_rigid_small.cpp
+++ b/src/RIGID/fix_rigid_small.cpp
@@ -785,7 +785,7 @@ void FixRigidSmall::initial_integrate(int vflag)
   // forward communicate updated info of all bodies
 
   commflag = INITIAL;
-  comm->forward_comm_fix(this,26);
+  comm->forward_comm_fix(this,29);
 
   // set coords/orient and velocity/rotation of atoms in rigid bodies
 
@@ -879,6 +879,7 @@ void FixRigidSmall::enforce2d()
     b->xcm[2] = 0.0;
     b->vcm[2] = 0.0;
     b->fcm[2] = 0.0;
+    b->xgc[2] = 0.0;
     b->torque[0] = 0.0;
     b->torque[1] = 0.0;
     b->angmom[0] = 0.0;
@@ -1351,6 +1352,16 @@ void FixRigidSmall::set_xv()
 
       v_tally(1,&i,1.0,vr);
     }
+  }
+
+  // update the position of geometric center
+  for (int ibody = 0; ibody < nlocal_body + nghost_body; ibody++) {
+    Body *b = &body[ibody];
+    MathExtra::matvec(b->ex_space,b->ey_space,b->ez_space,
+                      b->xgc_body,b->xgc);
+    b->xgc[0] += b->xcm[0];
+    b->xgc[1] += b->xcm[1];
+    b->xgc[2] += b->xcm[2];
   }
 
   // set orientation, omega, angmom of each extended particle
@@ -1905,11 +1916,15 @@ void FixRigidSmall::setup_bodies_static()
   double **x = atom->x;
 
   double *xcm;
+  double *xgc;
 
   for (ibody = 0; ibody < nlocal_body+nghost_body; ibody++) {
     xcm = body[ibody].xcm;
+    xgc = body[ibody].xgc;
     xcm[0] = xcm[1] = xcm[2] = 0.0;
+    xgc[0] = xgc[1] = xgc[2] = 0.0;
     body[ibody].mass = 0.0;
+    body[ibody].natoms = 0;
   }
 
   double unwrap[3];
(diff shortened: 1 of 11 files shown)


athomps (approved):
I approve. Nice work!

src/RIGID/fix_rigid_small.cpp
```diff
@@ -73,6 +73,7 @@ FixRigidSmall::FixRigidSmall(LAMMPS *lmp, int narg, char **arg) :
   dof_flag = 1;
   enforce2d_flag = 1;
   stores_ids = 1;
+  centroidstressflag = CENTROID_AVAIL;
```
akohlmey, line 76: This flag will be inherited by the `FixRigidNHSmall` class and thus the corresponding derived classes, but I don't see the corresponding changes to enable the feature; similarly `FixRigidSmallOMP`. So it either needs to be disabled in the corresponding constructors or the equivalent changes added.

akohlmey (changes requested):
As mentioned in the PR description, updated docs are missing and the suggested warning should be added.

donatas-surblys:
Regarding, `FixRigidNHSmall`, I think it's OK as is,
because the modifications in `FixRigidSmall` were in the `setup`, `set_xv`, `set_v`
and various communication functions, which are all called either explicitly or
through inheritance by `FixRigidNHSmall`. Actually, our paper was done with the
`fix rigid/nve/small`, which I believe inherits from `FixRigidNHSmall`.
Regarding the OMP version, yeah, I missed it, as it didn't come out with "-sf omp"
when using `fix rigid/nve/small`.
I'd like to fix it up too eventually, but I'll disable it for the time being.
I've also disabled centroid stress for non-small rigid fixes.

I was looking into having LAMMPS emit a warning about using `compute heat/flux`
with standard atomic virial from `fix rigid` and `fix rigid/small`.
Unfortunately, I can't seem to think up a good way, as the warning message should be
emitted from `compute_heat_flux.cpp`, but at that point there's not enough
information about the underlying fixes that were used to compute the virial.
I guess it would be easiest to add a new flag to the `stress/atom` compute,
but this doesn't seem like a clean approach.
For the time being, I've added a separate warning to the manual.

I've added short explanations and a new citation to the manual.
I've also added a short clarification on how the standard virial is computed
for `fix rigid`.

I've also added citation information to the centroid/stress/atom,
as has been suggested in the past.

akohlmey (approved):

PR #3048, opened 2021-12-07
Add Lammps extension for vscode to the tools section of the documentation 

ThFriedrich:
**Summary**

This PR adds [Lammps extension for vscode](https://thfriedrich.github.io/lammps_vscode/) to the tools section of the documentation and fixes a typo in doc/src/pair_lj_cut_coul.rst

**Author(s)**

Thomas Friedrich, University of Antwerp
thomas.friedrich@uantwerpen.be

**Licensing**

**Backward Compatibility**

 - No known compatibility issues

**Implementation Notes**

 - verified the docs build using `make`

**Post Submission Checklist**

### DIFF
--- a/doc/src/Tools.rst
+++ b/doc/src/Tools.rst
@@ -97,7 +97,7 @@ Miscellaneous tools
    * :ref:`singularity <singularity_tool>`
    * :ref:`SWIG interface <swig>`
    * :ref:`vim <vim>`
-   * :ref:`VSCODE Extension <lammps_vscode>`
+   * :ref:`Visual Studio Code Extension <lammps_vscode>`
 
 ----------
 
--- a/doc/src/Tools.rst
+++ b/doc/src/Tools.rst
@@ -97,6 +97,7 @@ Miscellaneous tools
    * :ref:`singularity <singularity_tool>`
    * :ref:`SWIG interface <swig>`
    * :ref:`vim <vim>`
+   * :ref:`VSCODE Extension <lammps_vscode>`
 
 ----------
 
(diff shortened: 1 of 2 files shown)


doc/src/Tools.rst
```diff
@@ -97,6 +97,7 @@ Miscellaneous tools
    * :ref:`singularity <singularity_tool>`
    * :ref:`SWIG interface <swig>`
    * :ref:`vim <vim>`
+   * :ref:`VSCODE Extension <lammps_vscode>`
```
rbberger, line 100:
```suggestion
   * :ref:`Visual Studio Code Extension <lammps_vscode>`
```

rbberger (approved):

akohlmey:
@ThFriedrich Please note that the Tools page has the following information at the top.
```
LAMMPS is designed to be a computational kernel for performing molecular dynamics computations. Additional pre- and post-processing steps are often necessary to setup and analyze a simulation. A list of such tools can be found on the LAMMPS webpage at these links:

Pre/Post processing

External LAMMPS packages & tools

Pizza.py toolkit

The last link for Pizza.py is a Python-based tool developed at Sandia which provides tools for doing setup, analysis, plotting, and visualization for LAMMPS simulations.

Additional tools included in the LAMMPS distribution are described on this page.
```

I cannot see the actual extension included added to the tools folder of the distribution. If this is meant to be external, it should be included in the LAMMPS website under "External LAMMPS packages & tools" instead.

ThFriedrich:
@akohlmey Thanks for the comment. I could add the the package as binary (~30mb), but I guess it makes little sense as it's being updated regularly and the installation is actually easier through the editor. Adding the repository as a submodule would include the tool in a way I guess, but not sure how this would be helpful. So putting it under [External Packages](https://www.lammps.org/external.html) indeed seems more consistent. This webpage doesn't seem to be part of the documentation though, unless I'm missing it. Not sure how I would add it there. Any pointers? 
On the other hand, there are other editor plugins listed under tools and it may be less intuitive for users to look for editor plugins in different places.

akohlmey:
adding a 30Mb blob to the distribution is not desirable, also we've avoided submodules (because it is too challenging for inexperienced folks) and using the internal facility for vscode extensions to install it is most certainly the best option.

> This webpage doesn't seem to be part of the documentation though, unless I'm missing it. Not sure how I would add it there. Any pointers?

The www.lammps.org website is in a separate, private repository and uses a different markup (for historic reasons), but it would be easy to add based on the RST changes.

> On the other hand, there are other editor plugins listed under tools and it may be less intuitive for users to look for editor plugins in different places.

This is not easy to resolve. It essentially boils down to the question "which information do we put in to the website directly and which into the manual". There were good reasons to keep those separate, but the boundaries have been blurred over time.

It would be rather straightforward to have the Tools page split into "bundled" and "external" tools and similarly to add an "external" packages section to the Packages page.

@sjplimp This is something that we should discuss at the next developer call as this is related to the topic of how to manage/improve maintenance of www.lammps.org. It may be beneficial to move some content from the website into the manual to make it easier to maintain and to keep up-to-date since it allows people to submit updates as a pull request.
Simplifying the website by consolidating information shared between website and manual and reducing the number of items on the website is a suggestion that has come up repeatedly.

akohlmey:
@ThFriedrich I have just added the content of this PR to the LAMMPS website. Please check out https://lammps.org/external.html#vscode and report any issues or need for corrections here.

We can revisit integrating this information into the manual later after we have had a discussion among the LAMMPS developers about whether to "rebalance" what information is put on the website directly and what would be integrated into the manual.

ThFriedrich:
Thanks @akohlmey! Looks all good to me. :)

PR #2025, opened 2020-04-25
 Time-dependent dumps for variable timestep

mkanski:
**Summary**

This PR implements new dump feature which enables outputting data based on an increment of simulation time instead of timesteps. This will be useful for simulations with a variable timestep, see #2011 for more details.

**Related Issues**

Closes #2011 

**Author(s)**

Michał Kański, Jagiellonian University, michal.kanski@gmail.com

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

No issues

**Implementation Notes**

Outputting dumps every `M` time units is achieved by checking at every step if the simulation time will exceed a multiply of `M` at the next step. Since the check is done at the end of a step, the value of `dt` is already updated if `fix dt/reset` is used. I also added a support for `dump_modify time` to `xyz`, `xyz/gz`, and `xyz/zstd` dump styles. 

**Post Submission Checklist**

- [x] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [x] The source code follows the LAMMPS formatting guidelines
- [x] Suitable new documentation files and/or updates to the existing docs are included
- [x] The added/updated documentation is integrated and tested with the documentation build system
- [x] The feature has been verified to work with the conventional build system
- [x] The feature has been verified to work with the CMake based build system

**Further Information, Files, and Links**

Tested dump_styles:
- [x] `atom`
- [x] `atom/gz`
- [x] `atom/zstd`
- [x] `atom/mpiio`
- [ ] `atom/adios` - not tested
- [x] `cfg`
- [x] `cfg/gz`
- [x] `cfg/zstd`
- [x] `cfg/mpiio`
- [x] `custom`
- [ ] `custom/adios` - not tested
- [x] `custom/gz`
- [x] `custom/zstd`
- [x] `dcd` - disabled, because it assumes a constant dt
- [ ] `h5md` - I couldn't get it to work, even without time-dependent dumping (segmentation fault)
- [x] `image`
- [x] `local`
- [x] `molfile`
- [x] `movie`
- [ ] `netcdf` - I couldn't get it to work, even without time-dependent dumping. Exits with error:
[...]

PR #3052, opened 2021-12-08
Time-dependent dumps for variable timestep (alternate implementation)

### LINKED ISSUE #2011: [Feature Request] Dump written every n time units for simulations with variable timestep
**Summary**

Implement an option for a dump to be written every n time units (e.g. 100 fs). 

**Detailed Description**

Currently, it is impossible to write a dump every n time units if a variable timestep is used. Such an option would be useful for visualising and analysing simulations where the timestep varies, e.g. modelling of sputtering or combustion.
I am willing to implement this, but I am not sure what is the best way to do it. I have a partial workaround (see the snippet below) which works without any change in the source code, but it has two issues:
1. It writes a dump header every `N`.
1. The accuracy (how close a frame is to a correct dump time) depends on the `N` parameter of the dump. Low `N` gives a better accuracy but slows down simulation, especially if potential energy per atom is written.  

I think the first issue can be avoided by introducing a new flag, e.g. `skip_empty`, and modify the line below, so it only writes header if there are atoms to be dumped or the flag is set to `false`:
https://github.com/lammps/lammps/blob/0711232e5b31fcf727803146636816b234283f8d/src/dump.cpp#L389

i.e.:
`if (filewriter && (!skip_empty || (skip_empty && nheader != 0))) write_header(nheader);`

I am not sure how to solve the second issue. I am also open to different implementations of this feature.

The snippet below writes a dump only every `n_time_every` time units:
`variable n_time_every equal 100`
`variable N equal 1`
`variable thresh atom (floor(time)%${n_time_every})`
`dump dump all atom $N dump.lammpstrj`
`dump_modify dump thresh v_thresh != LAST time yes`

sjplimp:
**Summary**

This is a second attempt at a PR to enable writing dump files based on simulation time interfvals, rather than timestep intervals.
Useful when timestep size is varying, e.g. with the fix dt/reset command.

**Related Issue(s)**

This is an alternative implementation to and thus closes #2011 and closes #2025 .  It uses some of the same ideas, but is different in the following ways:
a) changes are mainly confined to the Output class, almost nothing in the Dump clasdes, except for dump_modify parsing
b) dump instances which are in time mode (not step mode) store both a next time and a next step for invocation,
    so that the logic for invoking dumps while timestepping does not need to be altered or checked every timestep
c) fix dt/reset calls a method in Output when it changes the timestep size, so that the values in (b) can be adjusted
d) issues with insuring computes are invoked on dump output steps are addressed, computes are only invoked on needed steps
e) name of new dump modify keyword is every/time
f) similar to every keyword (for timesteps), the every/time keyword allows use of a variable to set the next interval (in time units)
g) removed logic for time dumps for storing future timesteps for needed energy/virial computes --
    for the case of variable timestep models (e.g. fix dt/reset), this is to avoid many false predictions of when the eng/virial
    needs to be enabled, leading to inefficiency -- the current method will invoke all per-atom eng/virial computes on a timestep
    a time dump is performed -- can't think of a good way to avoid that. e.g. if the dump doesn't need those computes

**Author(s)**

Michal Kanski and Steve Plimpton

**Licensing**

**Backward Compatibility**

Behavior of dump output triggered by timestep settings (as opposed to simulation time) should be unaltered.
Changed reset_timestep commandt to zero out simulation time if the new timestep < old timestep.

**Implementation Notes**

**Post Submission Checklist**


sjplimp:
@mkanski Please take a look at this alternate PR.  I think it addresses some of the issues you flagged, e.g. with computes or needing to check every timestep for possible dump output.  Attached are some input scripts I used for testing.  They all seem to work as expected.  Some features of your PR need to be imported into this.  E.g. enabling some of the Dump styles like xyz to output real time.  All the ugly details are in the new Output::calculate_next_dump() method.

sjplimp:
[test_scripts.tar.gz](https://github.com/lammps/lammps/files/7681236/test_scripts.tar.gz)

sjplimp:
@akohlmey You may want to take a look at this as well.  There are a few issues still to resolve.  What happens when the user
a) uses the timestep command between runs to change timestep size
b) does the same during a run, using the run every command
c) uses the reset_timestep command between runs to change the timestep
d) does the same during a run, using the run every command
Possibly we should just disallow (b) and (d).  Need to think about it.

mkanski:
@sjplimp Thank you for taking the time to implement this. It works fine for me with two exceptions:

- if `dt/reset` is invoked on every timestep, the dump is written only on the very first timestep
- The dump is always written at the end of minimization

Should I close PR #2025 and create a new one for writing simulation time in xyz dumps after this PR is merged?

sjplimp:
@mkanski Thanks - I'll look at the dt/reset issue.  I also realized that if the dumps are far apart and the timestep is changed frequently by dt/reset, then this version may up predicting many timesteps at which the dump will occur (whenever timestep changes).  Which is OK, except that all those timesteps will invoke all the energy/virial computes.  B/c it has no way of knowing when they are not needed.  I have an idea to fix that.

Re: dump at end of min - was that not the way the code worked before?  This seems unrelated to the new time dumps.

sjplimp:
Also, re the other PR, can you just add the support for other dump styles to this PR?  E.g. If not directly, I think you can submit a PR to be merged to this one, which I can do.

sjplimp:
@mkanski Think I fixed the issue with fix dt/reset freq=1
@akohlmey I'm thinking we should make it an error to use the reset_timestep command when a dump command is defined and the dump has already written out at least one snapshot.  I'm looking at the code in Output::reset_timestep() for dumps and it already has some caveats.  It will be even more complicated for time dumps (vs step dumps).  Since the user could pick a new timestep before or after the current timestep, and also induce snapshots that would then be out of sequence in the dump file or overwrite previous ones, this seems like an illogical thing to allow.  I can't think of a use case. Your thoughts?

akohlmey:
> @akohlmey I'm thinking we should make it an error to use the reset_timestep command when a dump command is defined and

I am all for it. Please note that for this particular use case, it should already be disallowed because fix dt/reset sets `Fix::time_depend=1;` and that triggers refusing to reset the timestep.

akohlmey:
> > @akohlmey I'm thinking we should make it an error to use the reset_timestep command when a dump command is defined and
> 
> I am all for it. Please note that for this particular use case, it should already be disallowed because fix dt/reset sets `Fix::time_depend=1;` and that triggers refusing to reset the timestep.

Just looking at fixes that set this flag, and I think there should be more. Any of the fix ave/* fixes would also be inconsistent if the time step would be reset, same as dumps.

sjplimp:
@akohlmey Just to clarify - in my Q above I was not talking about new time dumps where fix dt/reset could be involved.  I'm talking about standard dumps, where for example, someone would use reset_timestep between runs (e.g. after equil) to reset
the timestep, and sayiing we should disallow that if the dump is already active.  Though it currently is allowed I believe.
Conversely, I think it should be OK to use the timestep command (change size of timestep) between runs or via run every.  The new time dumps will have to account for that, similar to the fix dt/reset command.

akohlmey:
> @akohlmey Just to clarify - in my Q above I was not talking about new time dumps where fix dt/reset could be involved. I'm talking about standard dumps, where for example, someone would use reset_timestep between runs (e.g. after equil) to reset the timestep, and sayiing we should disallow that if the dump is already active. Though it currently is allowed I believe. Conversely, I think it should be OK to use the timestep command (change size of timestep) between runs or via run every. The new time dumps will have to account for that, similar to the fix dt/reset command.

That is what I understood. I was just saying that a run with a standard dump **and** fix dt/reset will *already* not allow using reset_timestep already due to the setting in fix dt/reset.

sjplimp:
@akohlmey I agree the fix ave/* commands should set time_depend=1.  I'm amazed someone hasn't had an issue with that before.  Though there is a related error check in FixAveTime::end_of_step().  But I think time_depend=1 is better.
Just don't allow users to do it.

akohlmey:
Just added this in commit facb49fc27a0942252381181c2183e24c2dc015d to PR #3040

sjplimp:
@mkanski OK - I think this is ready to merge now, except for adding simulation time info to any dump commands your older PR added it to.  Can you add those changes into this PR?

src/verlet.cpp
```diff
@@ -252,6 +252,8 @@ void Verlet::run(int n)
     ntimestep = ++update->ntimestep;
     ev_set(ntimestep);
 
+    printf("VERLET %ld: %d %d\n",ntimestep,eflag,vflag);
```
stanmoore1: Debug output needs to be removed.

stanmoore1 (changes requested):

stanmoore1:
Two of the unit tests are failing with `"ERROR: Cannot reset timestep with active dump - must undump first (src/output.cpp:609)`

sjplimp:
@mkanski I think I have now addressed my most recent comment to you.  I added your logic to the various "xyz" dump styles to include output of the simulation time if time_flag is set.  I also think any style (except dcd and xtc) will be compatible with the new every/time option for output at simulation time intervals.  I will push the new version in a few minutes.

mkanski:
@sjplimp Sorry for not responding earlier, but I got COVID last week and I wasn't able to work on this.

sjplimp:
@mkanski - ouch, very sorry to hear you got Covid - I hope you are recovering ok - I think this PR is now complete, you can check it if you wish

@akohlmey ready to merge, as far as I can tell

mkanski:
@sjplimp I'm almost cured, but still my sense of smell hasn't returned.

About the PR, everything looks good with one exception: time is equal to 0.0 in all frames of a `dump xyz` which were written during a minimization.  It wasn't an issue with my implementation because I invoked `update->update_time()` on each timestep.

sjplimp:
@mkanski - re:Covid - good to hear.   I changed the xyz dumps to output the current "time".  However your comment about min
makes me think that time dumps (as opposed to step dumps) maybe should be not allowed to be defined during minimization.  Conceptually I don't think "time" should advance during a min ?  What do you and @akohlmey think?

akohlmey:
>  time is equal to 0.0 in all frames of a `dump xyz` which were written during a minimization. It wasn't an issue with my implementation because I invoked `update->update_time()` on each timestep.

A minimization doesn't advance time, so I don't think there should be an output of the current time when updating a dump except during an MD, otherwise time is ill-defined. So I'd rather suggest that time is only output if `update->whichflag == 2`.

@sjplimp you are re-introducing the BIGINT_FORMAT macro that we are/were working hard to eliminate. The fmtlib code is and the `fmt::format()` and `fmt::print()` functions are included to not need them.

sjplimp:
Re: fmt lines - I just copied those lline from Michal's older PR, so please edit to whatever is current syntax.
Re: time during min.  It's more subtle than that.  It's up to the user to include or not the sim time in any dump file
with dump_modify time yes.  My suggestion was to just disallow use of a time dump when a minimization is performed.
But just as the code advances the timestep during min, I think it will also update the sim "time" during min.  Both
are counter-intuitive.  But probably we should at least be consistently counter-intuitive.  I think the current version does that.
Michal can verify for his example.

mkanski:
If you output time via `thermo` command, it does increase during minimization and it is also correctly written to `dump custom` dumps with `time` flag. I believe `dump xyz` should be consistent with the rest of LAMMPS and also output updated time instead of 0.0.

I think time dumps should be allowed during minimization, because the simulation time does change. It's a little unphysical, but at least everything would be consistent.

sjplimp:
Agree with @mkanski comment - I changed the xyz dumps to use the compute_time() function like dump custom does.
Once the fmt output changes are made, this should be ready to go.

src/verlet.cpp
akohlmey: this is done

akohlmey:
> Two of the unit tests are failing with `"ERROR: Cannot reset timestep with active dump - must undump first (src/output.cpp:609)`

Unit tests are recovered with commit a653ee6b2ce37920c64596c03ed141d826933aa7

akohlmey (approved):

PR #3054, opened 2021-12-11
Support binary native dump files with read_dump and rerun

### LINKED ISSUE #2446: [Feature Request] Support binary native dump files with read_dump and rerun
**Summary**

Using binary files can be a space and time saver for reading/post-processing trajectories. It would be nice to have support for that. So far, this can only be done via VMD molfile plugins and the choice of a suitable file format.

nw13slx:
**Summary**

Add binary native dump file format to `read_dump` and `rerun`. 

**Related Issue(s)**

Implements and thus fixes #2446 

**Author(s)**

Lixin Sun (Harvard) nw13mifaso@gmail.com

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

A flag and binary file detection are added to the `Reader` class
- inside the `ReaderNative` class this binary file flag is use to switch between text and binary
- mostly adapted from `tools/binary2txt.cpp`

**Post Submission Checklist**


akohlmey:
@nw13slx thank you very much for your contribution. However, to be consistent with the writing of dump files, I would suggest a different implementation approach.
- add a flag "binary" to the Reader class and detect any ".bin" extension in the generic file open method
- fold the functionality of ReaderNativeBin into ReaderNative and just prefix the different code paths with an 'if (binary)' and 'else'
- make more use of the platform abstraction and avoid non-portable code

I am currently updating your branch accordingly. I will also add unit tests.
Please note that the dump classes will write compressed text files when using .bin.gz as an extension and thus it is not needed to support decompressing binary files. It doesn't make much sense anyway, since binary files do not compress much (and would only be created by manually compressing existing binary files, which is not very efficient to begin with).

nw13slx:
Awesome. Thank you @akohlmey for the prompt revision! 

- The class merging look great. Sorry that I didn't notice that they are in the same dump class. 
- When you said, "make more use of the platform abstraction and avoid non-portable code", did you mean the places I used `new` and `delete` instead of `memory->...` ? I'll remember that next time.
- Agree on the compressed binary. It's a bit weird to have binary and then compressed. That's why I was not sure... 
- The unit test looks great! Thanks again.

akohlmey:
> Awesome. Thank you @akohlmey for the prompt revision!
> 
> * The class merging look great. Sorry that I didn't notice that they are in the same dump class.

It was rather straightforward to do since you already did the bulk of the work.

> * When you said, "make more use of the platform abstraction and avoid non-portable code", did you mean the places I used `new` and `delete` instead of `memory->...` ? I'll remember that next time.

not so much. new/delete is fine, if this is a single allocation. memory->xxx() is beneficial for multi-dimensional arrays and for growing arrays.
What I am mostly referring to are the abstraction functions in the platform and utils namespaces. this is needed for example for fseek() and a few other file operations.

> * Agree on the compressed binary. It's a bit weird to have binary and then compressed. That's why I was not sure...

the main benefit of binary files (next to some space savings) is that reading is much faster since no parsing is required. having to (transparently) decompress adds computational effort and thus is counterproductive. if you want smaller files, actually compressed text modes may be smaller than compressed binary. 

> * The unit test looks great! Thanks again.

it is rather minimal. haven't checked coverage... there is much more than can be done (different kinds/types of fields, more unusual fields).

src/reader.cpp
```diff
@@ -25,24 +25,32 @@ using namespace LAMMPS_NS;
...
 ------------------------------------------------------------------------- */
 
 void Reader::open_file(const std::string &file)
 {
   if (fp != nullptr) close_file();
 
   if (platform::has_compress_extension(file)) {
-    compressed = 1;
+    compressed = true;
     fp = platform::compressed_read(file);
```
nw13slx, line 43: Should we add error info for not supporting the compressed binary?

src/reader.cpp
```diff
@@ -25,24 +25,32 @@ using namespace LAMMPS_NS;
...
     if (!fp) error->one(FLERR,"Cannot open compressed file for reading");
```
nw13slx, line 44: Should we add error info for not supporting the compressed binary?

akohlmey: if we change things here, we need to make the corresponding changes to the dump styles. Right now, compressed binary files are not detected nor generated; if a file ends in .bin.gz it will be the same as .gz. For reading, we should have the same behavior. If somebody externally compresses a binary file and tries to read that, it will create the same error as if you would rename a binary file to not end in .bin, i.e. incorrect format.

src/reader_native.cpp
```diff
@@ -36,14 +36,17 @@ ReaderNative::ReaderNative(LAMMPS *lmp) : Reader(lmp)
 {
   line = new char[MAXLINE];
   fieldindex = nullptr;
+  maxbuf = 0;
+  databuf = nullptr;
```
rbberger, line 40:
```suggestion
  databuf = nullptr;
  magic_string = nullptr;
  unit_style = nullptr;
```

src/reader_native.cpp
```diff
@@ -36,14 +36,17 @@ ReaderNative::ReaderNative(LAMMPS *lmp) : Reader(lmp)
...
 }
 
 /* ---------------------------------------------------------------------- */
 
 ReaderNative::~ReaderNative()
 {
-  delete [] line;
+  delete[] line;
```
rbberger, line 47:
```suggestion
  delete[] line;
  delete[] magic_string;
  delete[] unit_style;
```

src/reader_native.cpp
```diff
@@ -83,22 +117,73 @@ int ReaderNative::read_time(bigint &ntimestep)
...
+    read_buf(&nchunk, sizeof(int), 1);
+
+    int n;
+    for (int i = 0; i < nchunk; i++) {
+      read_buf(&n, sizeof(int), 1);
+      skip_buf(n*sizeof(double));
+    }
+
+    delete[] magic_string;
+    delete[] unit_style;
```
rbberger, line 145: if this is intentional, it should also add:
```suggestion
    delete[] unit_style;
    magic_string = nullptr;
    unit_style = nullptr;
```

src/reader_native.cpp
```diff
@@ -311,22 +454,52 @@ bigint ReaderNative::read_header(double box[3][3], int &boxinfo, int &triclinic,
...
+        {
+          double *words = &databuf[m];
+          for (int k = 0; k < nfield; k++)
+            fields[i_atom][k] = words[fieldindex[k]];
+          i_atom += 1;
+          m+=size_one;
+        }
+    }
+    delete[] magic_string;
+    delete[] unit_style;
```
rbberger, line 483: same here.
```suggestion
    delete[] unit_style;
    magic_string = nullptr;
    unit_style = nullptr;
```

src/reader_native.cpp
```diff
@@ -54,25 +57,56 @@ ReaderNative::~ReaderNative()
 
 int ReaderNative::read_time(bigint &ntimestep)
 {
-  char *eof = fgets(line,MAXLINE,fp);
-  if (eof == nullptr) return 1;
+  if (binary) {
+    int endian = 0x0001;
+    revision = 0x0001;
+    magic_string = nullptr;
```
rbberger, line 63:
```suggestion
    delete[] magic_string;
    delete[] unit_style;
    magic_string = nullptr;
```

rbberger (changes requested):
Please review the `delete[]` for `magic_string` and `unit_style`. If you're deleting it, you should also invalidate that pointer.

src/reader_native.cpp
```diff
@@ -54,25 +57,56 @@ ReaderNative::~ReaderNative()
...
+    if (feof(fp)) return 1;
 
-  if (utils::strmatch(line,"^\\s*ITEM: TIME\\s*$"))
-    read_lines(2);
+    // detect newer format
+    if (ntimestep < 0) {
+      // first bigint encodes negative format name length
+      bigint magic_string_len = -ntimestep;
+
+      delete[] magic_string;
```
rbberger, line 76: Not needed because of the above code.

rbberger (approved):

PR #3062, opened 2021-12-17
Follow Amber NetCDF standard more closely

### LINKED ISSUE #3061: [BUG] dump netcdf does not follow the Amber convention, creating `double` variables instead of `float`
**Summary**

The Amber convention for NetCDF trajectories (http://ambermd.org/netcdf/nctraj.xhtml) required all variables to use 32-bit float, but LAMMPS always creates files with the `cell_lengths`, `cell_angles`, `time` variable using double/64-bit float values.

From reading the code, it also seems to be possible to create variables for `coordinates`/`velocities`/`forces` (and the extensions `scaled_coordinates` and `unwrapped_coordinates`) using 64-bit variables with `dump_modify double yes`, although this is not documented in the website. The default behaviour here is following the standard.

I realised this was an issue when trying to re-implement a netcdf3 file parser without relying on the official `netcdf-c` library for [chemfiles](https://github.com/chemfiles/chemfiles). It seems likely that this issue have been hidden by `netcdf-c`dynamically  converting from one datatype to the other dynamically as required. The other implementation I know about which is not based on `netcdf-c` is [scipy's](https://docs.scipy.org/doc/scipy-1.7.1/reference/generated/scipy.io.netcdf_file.html), where the exact datatype matters less since everything is dynamic in Python.

This is happening on the latest `develop` branch: https://github.com/lammps/lammps/blob/2d4f030f113f179411bb3bc6afe97adec8bd62b5/src/NETCDF/dump_netcdf.cpp#L383-L391, and has been the case since at least 2015 (when I generated the file I'm using for testing).

**Expected Behavior**

I'm not sure what the best thing to do is here. I will have to write some code to handle `double` variables anyway in chemfiles to allow users to read existing files, so one possible resolution would be to acknowledge this divergence from the standard somewhere, and keep it as is.

Alternatively, the default could be changed to follow the standard, and users wanting to use double precision variable could still do this with `dump_modify`.

---

I'm willing to send PR implementing whatever change is decided, both to the code and documentation.

Luthaf:
**Summary**

Multiple small improvements (one per commit) to make LAMMPS netcdf dump follow the standard more closely:
- use float for variables by default
- use float for `scale_factor` attributes
- fix the dimension used in `spatial`, `cell_spatial`, and `cell_angular` variables
- define units for as many things as possible

**Related Issue(s)**

fixes #3061 

**Author(s)**

- Guillaume Fraux <guillaume.fraux@epfl.ch>

**Licensing**

**Backward Compatibility**

This code will strictly speaking create backward incompatible outputs, but I think it is still worth doing to follow the standard more closely. At the same time, there is very low risk for actual breakage, since netcdf-c (the most used library for reading these files) dynamically convert from one data type to another. I will check this by creating new files and checking if popular software are still able to read them

- [x] TODO: check new netcdf output in VMD
- [ ] TODO: check new netcdf output in PyMOl 

**Implementation Notes**

Only `dump netcdf` should be affected by these changes.

**Post Submission Checklist**

Standard documentation: http://ambermd.org/netcdf/nctraj.xhtml


src/NETCDF/dump_netcdf.h
```diff
@@ -40,12 +40,26 @@ class DumpNetCDF : public DumpCustom {
   virtual void write();
 
  private:
+   // type of quantity for per-atom values (used to get the unit)
+   enum Quantity {
+      Unknown = 0,
```
akohlmey, line 45: Please make the enumerator members upper case to indicate that those are constants.

src/NETCDF/dump_netcdf_mpiio.h
```diff
@@ -40,12 +40,26 @@ class DumpNetCDFMPIIO : public DumpCustom {
   virtual void write();
 
  private:
+   // type of quantity for per-atom values (used to get the unit)
+   enum Quantity {
```
akohlmey, line 44: We try to avoid redundant definitions of constants. So it would be better to have this in a separate header `netcdf_quantity.h` as `enum LAMMPS_NS::NetCDFQuantity` and, if possible, only use a forward declaration in the dump style headers (to reduce dependencies).

src/NETCDF/dump_netcdf.h
Luthaf: sure! do you prefer me adding new commits to address the review comments or rebasing the PR & editing the commits to ensure each commit reflects the latest state of the code?

akohlmey: thanks. just add new commits and push the branch. our preference is to follow the development and have individual commits recorded as changes happen (and information why) instead of having a "clean" commit history. once a branch is public any rebasing will lead to diverging commit histories and make small modifications to a pull request from the maintainers more complicated.

akohlmey:
@Luthaf @pastewka I have make some updates and cleanups. I've also tried to test the output with VMD. On my Fedora 35 machine with NetCDF 4.8.0 it was required to recompile VMD since the pre-compiled binaries seem to be using an incompatible, older NetCDF version that does not recognize the LAMMPS created files as NetCDF format. For the self-compiled VMD, however, the files are recognized and read. When adding 
`dump            1 peptide netcdf 10 peptide.nc id type x y z vx vy vz` to the peptide example, VMD outputs on reading:
```
netcdfplugin) conventions: 'AMBER'
netcdfplugin) trajectory follows AMBER conventions version '1.0'
netcdfplugin) AMBER: program 'LAMMPS'
netcdfplugin) AMBER: program version '14 Dec 2021'
netcdfplugin) AMBER: spatial dimension: 3
netcdfplugin) AMBER: atom dimension: 84
netcdfplugin) AMBER: frame dimension: 31
netcdfplugin) AMBER: coordinates units: 'angstrom'
netcdfplugin) AMBER: no coordinates scalefactor attribute, 1.0 assumed
netcdfplugin) AMBER: coordinates scalefactor: 1.000000
netcdfplugin) AMBER trajectory contains periodic cell information
netcdfplugin) AMBER: cell lengths units: 'angstrom'
netcdfplugin) AMBER: no cell lengths scalefactor attribute, 1.0 assumed
netcdfplugin) AMBER: cell lengths scalefactor: 1.000000
netcdfplugin) AMBER: cell angles units: 'degree'
netcdfplugin) AMBER: no cell angles scalefactor attribute, 1.0 assumed
netcdfplugin) AMBER: cell angles scalefactor: 1.000000
Info) Using plugin netcdf for structure file peptide.nc
Info) Using plugin netcdf for coordinates from file peptide.nc
Info) Finished with coordinate file peptide.nc.
```

Luthaf:
The changes you made looks good to me, however I will not have much time to work more on this until early January.

akohlmey:
> The changes you made looks good to me, however I will not have much time to work more on this until early January.

I am still chasing some stack corruption issue with the `netcdf/mpiio` dump style. Have you been able to use it?

pastewka:
It used to work at some point, but my guys also told me there is an issue with pnetcdf. I may have time between Christmas and New Years to debug this.

akohlmey:
> It used to work at some point, but my guys also told me there is an issue with pnetcdf. I may have time between Christmas and New Years to debug this.

When compiling with clang 13 I get these rather unusual warnings
```
lammps/src/NETCDF/dump_netcdf_mpiio.cpp:214:23: warning: stack frame size (1192) exceeds limit (4294967295) in function 'LAMMPS_NS::DumpNetCDFMPIIO::openfile' [-Wframe-larger-than]
void DumpNetCDFMPIIO::openfile()
                      ^
compile/lammps/src/NETCDF/dump_netcdf_mpiio.cpp:732:23: warning: stack frame size (248) exceeds limit (4294967295) in function 'LAMMPS_NS::DumpNetCDFMPIIO::write_data' [-Wframe-larger-than]
void DumpNetCDFMPIIO::write_data(int n, double *mybuf)
                      ^
```

The compiler detects a stack frame issue even though the diagnostic does not exceed the limit. Running with both valgrind and gdb confirms that the stack overflow/corruption and thus segfault happens with executing the `openfile()` function, which is the first of the two that is encountered.

```
==61507== Memcheck, a memory error detector
==61507== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==61507== Using Valgrind-3.18.1 and LibVEX; rerun with -h for copyright info
==61507== Command: ./lmp -in in.peptide
==61507== 
==61507== Warning: client switching stacks?  SP change: 0x1ffeffebd8 --> 0x15feffe760
==61507==          to suppress, use: --max-stackframe=42949674104 or greater
==61507== Invalid write of size 8
==61507==    at 0x567FBE: LAMMPS_NS::DumpNetCDFMPIIO::openfile() (src/NETCDF/dump_netcdf_mpiio.cpp:275)
==61507==  Address 0x15feffe758 is on thread 1's stack
==61507== 
==61507== 
==61507== Process terminating with default action of signal 11 (SIGSEGV): dumping core
==61507==  Access not within mapped region at address 0x15FEFFE758
==61507==    at 0x567FBE: LAMMPS_NS::DumpNetCDFMPIIO::openfile() (src/NETCDF/dump_netcdf_mpiio.cpp:275)
==61507==  If you believe this happened as a result of a stack
==61507==  overflow in your program's main thread (unlikely but
==61507==  possible), you can try to increase the size of the
==61507==  main thread stack using the --main-stacksize= flag.
==61507==  The main thread stack size used in this run was 8388608.
```

A side-by-side comparison of the source code shows very little differences of the two dump styles and most are integer constants or library calls. I've removed one inconsistency between the two styles where in the `netcdf/mpiio` style it was shadowing the "dims" array locally with an integer variable where `netcdf` uses "dim", but that should not and did not make a difference. I've used up pretty much my entire arsenal of C++ debugging tools, so any help on this would be highly welcome.
The only remaining method to narrow this down that I can think of would be rather laborious: comment out all class function bodies and put them back in small stages and see which change triggers the segfault/stack corruption.

pastewka:
Let me look at it over the holidays, maybe I see something. The stack frame does not exceed the limit, my guess is that the limit it somehow set wrongly - to the max of some unsigned int that is then interpreted as a signed negative integer. I don't know why this would lead to a segfault though.

pastewka:
Did you try playing with `ulimit` to see if the segfault disappears?

akohlmey:
> Did you try playing with `ulimit` to see if the segfault disappears?

I did the equivalent when using valgrind, but that only made it gobble up huge amounts of ram.
It must have to do something will call sequences and the cause may not be in the code in the NETCDF folder. Please recall that in C++ the "this" pointer is added as the first argument to all member functions and that things are more complex in LAMMPS because the `openfile()` member function is called from the `DumpCustom` class, so also virtual function tables are involved and errors there are always difficult to debug. :-(

akohlmey:
@Luthaf @pastewka I think the issue causing the stack corruption is resolved. In newer revisions of `pnetcdf.h` the constants `NC_MAX_*` are set to `NC_MAX_INT` instead of 1024. Thus I created a new constant `LMP_MAX_VAR_DIMS` in the `NetCDFUtils` namespace with a value of 1024 and use that for both dump styles.

akohlmey:
I've also added some tests to the unit test tree.

akohlmey:
@Luthaf @pastewka as far as I am concerned, this is now ready for merging. Please let me know what you think and whether you have any suggestions for further improvements or cleanups (and if yes, what they are :wink:).

pastewka:
Looks good to me, thanks for taking care of this @akohlmey - I like the new way of handling the unit conversion...

Luthaf:
The code looks good to me as well, thanks a lot @akohlmey for the cleanup & test cases!

sjplimp (approved):

PR #3065, opened 2021-12-20
Angle class2 update

sjplimp:
**Summary**

The angle class2 and variant class2/p6 were not using the specified r1/r2 distances for the BondAngle term.  They were
using the ones for the BondBond term.  This corrects that.  However it is likely a non-issue since it is likely in all cases that
the r1/r2 values in both terms are the same.  Also an important note was added to the doc page explaining that the order of I,J,K Angles in the data file needs to match the ordering of r1/r2 in those 2 sets of coefficients.

**Related Issue(s)**

N/A

**Author(s)**

Steve

**Licensing**

**Backward Compatibility**

This should only change the energy/forces of the class2 angle term if the bond/angle cross term uses different r1/r1 coeffs versus the bond/bond cross term.

**Implementation Notes**

**Post Submission Checklist**


akohlmey:
what about the KOKKOS and OPENMP versions?

sjplimp:
good catch - done

stanmoore1 (approved):

akohlmey (changes requested):
- KOKKOS version currently doesn't compile
- The changes **do** alter the forces for all of the force tests in the unit test tree so those will need to be reset after all other changes are complete

sjplimp:
Is the altered force test b/c the r1,r2 coeffs were different for the bond-bond and bond-angle cross terms?

akohlmey:
> Is the altered force test b/c the r1,r2 coeffs were different for the bond-bond and bond-angle cross terms?

probably. inputs are here: https://github.com/lammps/lammps/blob/develop/unittest/force-styles/tests/angle-class2.yaml and here: https://github.com/lammps/lammps/blob/develop/unittest/force-styles/tests/angle-class2_p6.yaml

PR #3072, opened 2021-12-30
Modernize parsing of topology data sections of data files

akohlmey:
**Summary**

This modernizes the parsing of several sections of data files using the ValueTokenizer class.
This enables stricter checking of the format (the number of tokens/words was previously only checked on the first line) and better reporting of errors since the error message now includes the offending line of text.
The parsing of the following sections is updated:
- Atoms
- Velocites
- Masses
- Bodies
- Bonus
- Bonds
- Angles
- Dihedrals
- Impropers

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues. Automated unit and regression tests pass.

**Implementation Notes**

The loop over the char buffer is unchanged instead of using `utils::split_lines()` to avoid extra copies of the data buffer.

To simplify parsing using the `utils::*numeric()` functions the handling of `char *` arguments (which  may be NULL) is now put in an overloaded function that will then call the original function with an empty string in that case. That way the original functions can safely use a `const std::string &` argument and thus avoid using the `c_str()` member function when converting of C++ strings.

**Post Submission Checklist**

### DIFF
--- a/src/atom.cpp
+++ b/src/atom.cpp
@@ -1209,31 +1209,24 @@ void Atom::data_vels(int n, char *buf, tagint id_offset)
   if (nwords != avec->size_data_vel)
     error->all(FLERR,"Incorrect velocity format in data file");
 
-  char **values = new char*[nwords];
-
   // loop over lines of atom velocities
   // tokenize the line into values
   // if I own atom tag, unpack its values
 
   for (int i = 0; i < n; i++) {
     next = strchr(buf,'\n');
+    *next = '\0';
+    auto values = Tokenizer(utils::trim_comment(buf)).as_vector();
+    if (values.size() != nwords)
+      error->all(FLERR, "Incorrect atom format in data file: {}", utils::trim(buf));
 
-    for (j = 0; j < nwords; j++) {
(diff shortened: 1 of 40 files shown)


src/atom.cpp
```diff
@@ -1381,19 +1382,27 @@ void Atom::data_angles(int n, char *buf, int *count, tagint id_offset,
...
+      ValueTokenizer values(utils::trim_comment(buf));
+      values.next_int();
+      itype = values.next_int();
+      atom1 = values.next_tagint();
+      atom2 = values.next_tagint();
+      atom3 = values.next_tagint();
+      atom4 = values.next_tagint();
+      if (values.has_next()) throw TokenizerException("Too many tokens","");
+    } catch (TokenizerException &e) {
+      error->one(FLERR,e.what() + std::string(" in ") + errtxt + utils::trim(buf));
```
rbberger, line 1404: How about `fmt::format("{} in {}: {}", e.what(), location, utils::trim(buf))`? Then you can also make `errtxt` to `auto location = "Dihedrals section of data file"`.

sjplimp (approved):

PR #3076, opened 2022-01-05
Fix bug of uf_lookup and u_lookup in angle_table

dynikon:
**Summary**

Fix bug of uf_lookup and u_lookup in angle_table.

tb->de and tb->df are array of length tlm1(tablength-1),
but original code tried to read tb->de[itable] and tb->de[itable] even if itable=tlm1.
So, the maximum of itable should be tablength-2.

In my environment, no error has occurred in original code,
but I thought this bug might cause some error in future.

**Licensing**

**Post Submission Checklist**

### DIFF
--- a/src/MOLECULE/angle_table.cpp
+++ b/src/MOLECULE/angle_table.cpp
@@ -635,7 +635,7 @@ void AngleTable::uf_lookup(int type, double x, double &u, double &f)
   int itable = static_cast<int> (x * tb->invdelta);
 
   if (itable < 0) itable = 0;
-  if (itable >= tablength) itable = tablength-1;
+  if (itable >= tablength-1) itable = tablength-2;
 
   if (tabstyle == LINEAR) {
     fraction = (x - tb->ang[itable]) * tb->invdelta;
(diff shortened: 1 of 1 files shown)


akohlmey:
@dynikon thanks for you submission. I have looked at the situation and your solution and after some thinking have decided, that the cleaner solution would be to allocate all elements for the linear interpolation table to the same size and then just do an extrapolation to get the final two elements. This is now implemented in commit d0f203127db0e53ab76214c8bea76e9c9efb0460

Please note that bond style table is affected by this issue in the same way (but not dihedral style table, since that is cyclic)

For the future, please also note that a) we no longer use "master" as the development branch but "develop" and b) that it is a best practice to have upstream branches (like "master" or "develop") unchanged from upstream (and update them with fast-forward merges from upstream only) and instead submit changes in feature branches.

dynikon:
@akohlmey thank you for your comments and commits.
I think your implementation will fix this problem, and thank you for mentioning about bond style table (I didn't notice about that).
And sorry for committing to master.

PR #2233, opened 2020-07-14
Dynamical_matrix and third_order features+update

charlessievers:
**Summary**

Further optimize third order, add folding to both commands, add output cutoff control for third order, add GPU support to both commands, reformat dynamical_matrix command to have 2 space indentation, and add force constant matrix option to dynamical_matrix command.

**Author(s)**

Charlie Sievers, UC Davis, charliesievers@cox.net

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Post Submission Checklist**

- [x] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [x] The source code follows the LAMMPS formatting guidelines
- [ ] The feature has been verified to work with the conventional build system
- [x] The feature has been verified to work with the CMake based build system
- [ ] Suitable tests have been added to the unittest tree.

**Further Information, Files, and Links**

Feature is not complete yet. 

Folding is discussed here: https://dannyvanpoucke.be/folding-phonons/
The fold options considers the interactions of the supercell (N atoms) w.r.t displacements of the original cell (n atoms). This produces a dynamical matrix of size (3nx3N). The original cell is designated by a user defined group, and the supercell is always considered to be the group all.

PR #3082, opened 2022-01-07
Improve PairReaxFFKokkos neigh list algorithm

stanmoore1:
**Summary**

Reduce the number of resizes needed for `PairReaxFFKokkos::PairReaxBuildListsHalf` by continuing to count if space is exceeded.

**Related Issue(s)**

None

**Author(s)**

Stan Moore (SNL), with helpful discussions from Evan Weinberg (NVIDIA)

**Licensing**

**Backward Compatibility**

Yes


stanmoore1:
@akohlmey this is ready to merge from my point of view.

akohlmey:
> @akohlmey this is ready to merge from my point of view.

yes, I only removed the label because having it present will trigger the extended regression and unit tests on every comment or other change to the PR (not just pushes). this way I can merge it right away after having one more approval without having to wait at least half an hour.

while on the subject on merging, can you please check out PR #2233? your change request is still blocking it.

stanmoore1:
OK thanks for the clarification, I will review #2233.

athomps (approved):
I approve.

PR #3101, opened 2022-01-26
Use setuptools instead of distutils in setup.py

### LINKED ISSUE #3128: [BUG] Possible segfault in Python initialisation
**Summary**

A low-prio corner case that I just stumbled upon:

If for some reason the call to `lammps_open` above fails, the following version query results in a segfault:

https://github.com/lammps/lammps/blob/2dfeb96fdabefd15c257ec8696554734a71a333b/python/lammps/core.py#L422

**Expected Behavior**

We're already in an invalid state then, but it is probably better to raise a Python exception instead of crashing.

**Steps to Reproduce**

This happened to me by providing a command line that contained package OpenMP instructions with a library that had not been built with that package:

```py
lmp = lammps.lammps(cmdargs=["-log", "none", "-sc", "none", "-pk", "omp", "1"])
```

```
LAMMPS Exception: ERROR: Package omp command without OPENMP package installed (src/input.cpp:1643)

Thread 1 "python3" received signal SIGSEGV, Segmentation fault.
lammps_version (handle=0x0) at lammps/src/library.cpp:4358
4358      return lmp->num_ver;
```

rbberger:
**Summary**

* `distutils` will be removed soon (Python 3.12). This upgrades the `setup.py` in the LAMMPS Python package.
* CMake: searches for Python `Interpreter` and `Development` components at the same time to ensure versions match (see [CMake docs](https://cmake.org/cmake/help/latest/module/FindPython.html))
* CMake: use same `find_package(Python)` for both LAMMPS and unit tests

This PR also contains some bug fixes:
* avoid segmentation fault when the instantiation of the LAMMPS class through the library interface fails (e.g. due to incompatible arguments); a RuntimeError exception is raised now. Fixes #3128 
* must reset value of BUILD_SHARED_LIBS if turned off in global namespace (e.g. when configuring compiling the embedded KOKKOS library)

**Author(s)**

@rbberger (Temple U)

**Licensing**

**Backward Compatibility**

**Implementation Notes**

LAMMPS also has an `install.py` that is used for some make targets. The main difference is that `install.py` also installs the shared library. Even if you update to `setuptools` in `install.py`, it still uses `easy_install` methods which are also on the outs. My recommendation is to get rid of `install.py` and use standard tools such as `pip` and `wheel` instead.

**Post Submission Checklist**


athomps (approved):
I approve.

PR #3102, opened 2022-01-27
Porting unit tests to run natively on Windows

akohlmey:
**Summary**

The pull request contains multiple changes that enable building and running (most) unit tests on windows when compiling with native compilers. Multiple bugs, portability errors, and other peripheral issues were detected in the process and corresponding changes are incorporated here as well.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

TBD

**Implementation Details**

The following sets of changes were applied to facilitate the objective:
- set CMAKE_RUNTIME_OUTPUT_DIRECTORY and CMAKE_LIBRARY_OUTPUT_DIRECTORY to the top level build folder, so that all executables and the LAMMPS shared library are there and not scattered through a directory tree equivalent to the source tree. This simplifies debugging of individual tester executables and works around issues with shared library lookup on Windows (since Windows has no equivalent of RPATH, and manipulations of the PATH environment variable is impossible before CMake 3.20 due to Windows' use of ';' as separator)
- make all added tests use the modern CMake syntax with NAME and COMMAND arguments to enable automatic look up of test executables after the previous change item was applied
- force "lib" prefix on all variants of liblammps. This forces Windows native compilers to produce liblammps.dll instead of lammps.dll and thus keep it consistent with cross-compilation and the expectations in the LAMMPS python module.
- support LAMMPSDLLPATH environment variable to override the liblammps.dll library location on Windows
- turn static class member variable in LAMMPS class that store git status info into functions. This enabled automatic export in the dll and avoids having to mess with `_declspec()` to export/import them from the shared library. By default everything is hidden on windows unlike on Linux/MacOS/FreeBSD
- add -Og when using gcc preset with Debug settings
- handle Windows specific inconsistency stemming from fread() dropping carriage return characters on reading and thus fread() calls may return fewer bytes than expected after repositioning the file pointer. This affect pair style eam/cd which needs to read a line from the end of the potential file
- fix uninitialized variable access bug in compute pair/local
- avoid integer overflow in dump balancing communication
- simplify fix deform searching for rigid fixes using new fix list access APIs.
- allow variable expansion for quoted arguments on the "include" command (to support variables containing paths with blanks)
- reduce number of reported memory issues in dump style testers
- Rather than setting up MPI STUBS to emulate an imported MPI target, we just incorporate the MPI object and explicitly add the include path in the few cases where otherwise we would add the MPI stubs library to the target to do that implicitly.
- do not explicitly set the test working directory to the default value
- cannot use symbolic links for test input files on Windows, must use explicit copies instead
- update handling of regular expressions in unit test matchers to suit the different regex semantics on Windows where '.' does not include '\n'. This predominantly means that MatchesRegex has to be replaced with ContainsRegex, but also some multi-line regex matchers must contain explicit '\n's and the correct number and location.
- skip PyLammps tests for now on Windows due to conflicts in I/O redirection
- add option to skip selected tests from YAML files if they contain a "no${CMAKE_SYSTEM_NAME}" tag, e.g. skip test entirely on windows with "noWindows" or MacOS with "noDarwin". not used (anymore), though.

**Post Submission Checklist**


cmake/CMakeLists.txt
```diff
@@ -290,25 +290,8 @@ if(BUILD_MPI)
...
-    target_include_directories(lammps INTERFACE $<BUILD_INTERFACE:${LAMMPS_SOURCE_DIR}/STUBS>)
-    target_compile_definitions(lammps INTERFACE $<INSTALL_INTERFACE:LAMMPS_LIB_NO_MPI>)
-  else()
-    target_include_directories(lammps INTERFACE $<BUILD_INTERFACE:${LAMMPS_SOURCE_DIR}/STUBS>)
-    target_compile_definitions(lammps INTERFACE $<INSTALL_INTERFACE:LAMMPS_LIB_NO_MPI>)
-    target_link_libraries(lammps PUBLIC mpi_stubs)
-  endif()
-  add_library(MPI::MPI_CXX ALIAS mpi_stubs)
+  target_sources(lammps PRIVATE ${LAMMPS_SOURCE_DIR}/STUBS/mpi.cpp)
+  target_include_directories(lammps PUBLIC ${LAMMPS_SOURCE_DIR}/STUBS)
```
junghans, line 294:
```suggestion
  target_include_directories(lammps PUBLIC $<BUILD_INTERFACE:${LAMMPS_SOURCE_DIR}/STUBS>)
```

cmake/CMakeLists.txt
```diff
@@ -594,7 +577,7 @@ if(PKG_ATC)
   if(BUILD_MPI)
     target_link_libraries(atc PRIVATE MPI::MPI_CXX)
   else()
-    target_link_libraries(atc PRIVATE mpi_stubs)
+    target_include_directories(atc PRIVATE ${LAMMPS_SOURCE_DIR}/STUBS)
```
junghans, line 580: I don't like the fact the we explicitly have to hard-code this directory everywhere.

I think slightly more CMake-ish way of doing this would be to create add target:
```cmake
add_library(mpi_flags INTERFACE)
target_include_directories(mpi_flags INTERFACE ${LAMMPS_SOURCE_DIR}/STUBS)
```
and then use
```
target_link_libraries(atc PRIVATE mpi_flags)
```

akohlmey: I don't like this indirect "CMake way" of specifying include paths by "faking" a library. Just adding an include path is more obvious to non-experts. However, if that "fake" library is called under the same target as the real MPI library, then this would allow to treat both cases as "need to add MPI support here" and reduce some redundant calls. Made some changes along that way and that seems to be working as expected.

cmake/CMakeLists.txt
akohlmey: Made the change because I trust you, but please explain why is this important to a non-guru person.

junghans: PUBLIC include directories will be added to the lammps exported target during install. But having a path from the source directory there makes not much sense as the source might be deleted after install.

cmake/CMakeLists.txt
junghans: Good point! mpi_flags makes it more clear it depends on mpi and not something else in stubs.

cmake/CMakeLists.txt
```diff
@@ -280,36 +280,20 @@ if(BUILD_MPI)
...
-    target_include_directories(lammps INTERFACE $<BUILD_INTERFACE:${LAMMPS_SOURCE_DIR}/STUBS>)
-    target_compile_definitions(lammps INTERFACE $<INSTALL_INTERFACE:LAMMPS_LIB_NO_MPI>)
-    target_link_libraries(lammps PUBLIC mpi_stubs)
-  endif()
+  target_sources(lammps PRIVATE ${LAMMPS_SOURCE_DIR}/STUBS/mpi.cpp)
+  add_library(mpi_stubs INTERFACE)
+  target_include_directories(mpi_stubs INTERFACE $<BUILD_INTERFACE:${LAMMPS_SOURCE_DIR}/STUBS>)
   add_library(MPI::MPI_CXX ALIAS mpi_stubs)
 endif()
+target_link_libraries(lammps PUBLIC MPI::MPI_CXX)
```
junghans, line 296: Some older CMake version didn't support this alias in the past, but if it works now that would be much cleaner.

cmake/CMakeLists.txt
akohlmey: Until we have made changes to the LAMMPS source tree to have a well defined C++ API for the library, use a version specific SONAME, and have a lammpsconfig.h header that records all defines instead of depending on using some method of passing required flags, the "make install" functionality should be considered "unsupported". I understand that this is useful for what you do (or want to do), but from my perspective this is subject to review and refactoring/redesign at a later point. Thus these kinds of adjustments and considerations are not going to be of any priority. It was easy enough in this particular case, so it was no problem this time.

cmake/CMakeLists.txt
akohlmey: Yes, I saw the comment, and will be subject to careful testing. Having spent some more time reading about CMake best practices, I am somewhat leaning toward requiring a newer version as minimum for CMake rather sooner than later (e.g. 3.12 would eliminate several workarounds), but Ubuntu 18.04LTS still has such a large installed base (and comes with 3.10) that I am too concerned that we would exclude too many people. The CMake build system is supposed to make it easier and more consistent for folks to compile LAMMPS, but if they first have to install a newer CMake version, that benefit goes right out the window. I just ran into an issue where a much newer CMake version would be a big help. See my next comment/question.

akohlmey:
@junghans here is a CMake challenge for you that has been bothering me for a bit now.
When building LAMMPS **with** testing enabled natively on Windows (this is part of what this PR is supposed to enable, plus refactoring of items I can across on the way), there is a problem when LAMMPS is compiled as a shared library:

The native VS build tool is a "multi-config" build system, that means, that executables are in a subfolder of CMAKE_CURRENT_BINARY_DIR corresponding to the selected configuration. While that works for running LAMMPS itself, it fails for the unit tests, because those cannot "find" lammps.dll. The dynamic loader cannot find it because the DLL is in a different folder and not covered by the PATH environment variable. I've tried to augment PATH, but ran into two problems: 1) I somehow cannot infer the current configuration (it look like I managed to expand it from $<CONFIG> in some other place, but that does seem to apply here) and 2) manipulation of the PATH environment variable is very tricky, since CMake interprets the PATH on Windows as a list (due to the use of ';' instead of ':') and functions to manipulate variables containing paths exist only in very recent versions of CMake (3.20 and later). Any suggestions for addressing either issue would be highly welcome.

FWIW, short of running on Windows, it is also possible to test/debug Multi-config specific issues on Linux when using -G 'Ninja Multi-Config'. Of course the DLL path resolution issue is a Windows thing... and there is apparently no equivalent of RPATH :-(

junghans:
Yeah, I don't have much experience with Windows nor multi-config. I have seen a solution where `CMAKE_LIBRARY_OUTPUT_DIRECTORY` is set to be the same as `CMAKE_RUNTIME_OUTPUT_DIRECTORY` and then the test are also executed with that as `WORKING_DIRECTORY`. However then there might a raise condition when running test in parallel when using the same file names, as we do!

For the multi-config case isn't the config stored in `CMAKE_CONFIGURATION_TYPE` or something?

akohlmey:
> Yeah, I don't have much experience with Windows nor multi-config. I have seen a solution where `CMAKE_LIBRARY_OUTPUT_DIRECTORY` is set to be the same as `CMAKE_RUNTIME_OUTPUT_DIRECTORY` and then the test are also executed with that as `WORKING_DIRECTORY`. However then there might a raise condition when running test in parallel when using the same file names, as we do!

Thanks for the hint. That is indeed helpful because it allows to "collect" all (shared) libraries and executables in the build folder and not have them in some subdirectory. Especially for the test executables, that has an additional benefit when one wants to run the tests in a standalone fashion, e.g. with a debugger. After some experimentation I have found that the WORKING_DIRECTORY can remain in place, but also found that we are using some backward compatible shortcut syntax which triggers an unwanted behavior. If add_test() is called with "NAME" and "WORKING_DIRECTORY" explicitly the tests will be found in the build folder yet run in the desired working directory. So the nasty PATH environment manipulations are not needed. CTest will take care of that and having the DLL file in the same folder as the executables will make the DLL lookup work on windows (still not yet tested).

> For the multi-config case isn't the config stored in `CMAKE_CONFIGURATION_TYPE` or something?

No. There is only CMAKE_CONFIGURATION_TYPES (plural) and that is to provide a list of available configurations. With multi-config, you cannot query the configuration at the time CMake is run but it changes later. It may be possible to use a generator expression.

Anyway, this is making progress again and allowing some overall improvement in consistency and removing some workarounds for identifying file locations (e.g. for plugin objects).

cmake/CMakeLists.txt
```diff
@@ -280,35 +283,19 @@ if(BUILD_MPI)
...
-  else()
-    target_include_directories(lammps INTERFACE $<BUILD_INTERFACE:${LAMMPS_SOURCE_DIR}/STUBS>)
-    target_compile_definitions(lammps INTERFACE $<INSTALL_INTERFACE:LAMMPS_LIB_NO_MPI>)
-    target_link_libraries(lammps PUBLIC mpi_stubs)
-  endif()
-  add_library(MPI::MPI_CXX ALIAS mpi_stubs)
+  target_sources(lammps PRIVATE ${LAMMPS_SOURCE_DIR}/STUBS/mpi.cpp)
+  add_library(mpi_stubs INTERFACE)
+  target_include_directories(mpi_stubs INTERFACE $<BUILD_INTERFACE:${LAMMPS_SOURCE_DIR}/STUBS>)
+  target_link_libraries(lammps PUBLIC mpi_stubs)
```
junghans, line 298: What happened to the alias?

unittest/c-library/CMakeLists.txt
```diff
@@ -1,26 +1,26 @@
 
 add_executable(test_library_open test_library_open.cpp test_main.cpp)
 target_link_libraries(test_library_open PRIVATE lammps GTest::GMock)
-add_test(LibraryOpen test_library_open)
+add_test(NAME LibraryOpen COMMAND test_library_open WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
```
junghans, line 4: Isn't `CMAKE_CURRENT_BINARY_DIR` the default?

junghans: https://cmake.org/cmake/help/latest/prop_test/WORKING_DIRECTORY.html#prop_test:WORKING_DIRECTORY

cmake/CMakeLists.txt
akohlmey: rolled that change back to minimize complications while working on other portability issues. will check it out again, now that everything else seems to be sorted out.

akohlmey:
@junghans @rbberger native compilation and unit testing is now mostly working on Windows with MSVC.
https://github.com/akohlmey/lammps/actions/workflows/compile-msvc.yml
Thanks a lot for your suggestions with this. Don't think it would have been possible otherwise. And there are some general improvements on top of just enabling this feature on windows.
There are some complications with Python that I am still looking into. For now those tests are skipped and I may need to postpone the resolution of those for a later pull request.

cmake/CMakeLists.txt
akohlmey: @junghans here is what I learned about this issue:
The alias as it was done before would not work because we already created the `MPI::MPI_CXX` imported target when using `find_package(MPI QUIET)` to initialize the `BUILD_MPI` option setting. CMake lets be create the alias, but it will use the imported target instead. Since there is no way to remove/delete/disable/rename an imported target, the only viable solution seems to be to create the same alias target for **both**. So I created a target `MPI::ANY_CXX` to represent either the real MPI or the stub MPI library and when I use *this* alias, it will correctly use the stub definitions as expected. At least as far as I could find out so far.

akohlmey: sadly, now ubuntu 18.04LTS (or rather CMake 3.10) is rearing its ugly head. While this works for serial compilation, the MPI compilation fails because CMake 3.10 does not let me create an alias to an imported target. For later CMake it will work, after the imported target is promoted to global scope, but that option does not exist. So back to square one and the old CMake code until ubuntu 18 is obsolete and we can require a later CMake version.

junghans: Too bad.

Could we fix that by using:
```
set_target_properties(MPI::MPI_CXX PROPERTIES IMPORTED_GLOBAL TRUE)
```
on CMake-3.10 or below ONLY?

A another option would be to do the alias only for 3.12 and above, and add that to the if block for `target_link_libraries()` that we have to carry around either way and then it is at least clear in the future what to do. Because it seems like every 6 months we are asking yourselves again why we didn't use an `ALIAS` target ;-)

akohlmey: No. The property exists since 3.11 only.
I think having those ifs for cmake versions are worse than the current situation.

akohlmey: There is not much more left to streamline. This version is still a significant improvement since the settings are carried with the LAMMPS library and we don't need to look for and link to the stubs library, which was causing the problems that got me started.

junghans: I would at least leave a comment recommending the future use of an ALIAS target.

unittest/c-library/CMakeLists.txt
```diff
@@ -1,26 +1,26 @@
 
 add_executable(test_library_open test_library_open.cpp test_main.cpp)
 target_link_libraries(test_library_open PRIVATE lammps GTest::GMock)
-add_test(LibraryOpen test_library_open)
+add_test(NAME LibraryOpen COMMAND test_library_open)
```
junghans, line 4: Minor nitpick: without any extra arguments these CMake changes are purely cosmetic now.

akohlmey: No they are not. Without them the test executables are not found (at least on windows) with a multi-config builder. I found that out empirically when comparing tests that would find them to those that would not.

junghans: That is interesting! I didn't know that!

Maybe this is part of the transformations: "Unlike the above NAME signature no transformation is performed on the command-line to support target names or generator expressions." (from https://cmake.org/cmake/help/latest/command/add_test.html)

Who knows, the new syntax is anyway more descriptive and we should use it everywhere.

akohlmey:
@rbberger are you planning to address the PyLammps issues in this PR? Otherwise please review so it can be merged soon.

rbberger (approved):

rbberger:
@akohlmey need more context to see what is wrong for myself, but that shouldn't stop this PR

PR #3115, opened 2022-02-03
PyLammps update

rbberger:
**Summary**

* Updates PyLammps to make better use of modern LAMMPS Python interface
* Add more properties to `pylammps.Atom`
* Modify how the output is captured to make it work across platforms
* IPython command auto-completion now includes command styles

**Related Issue(s)**

**Author(s)**

@rbberger (Temple U)

**Licensing**

**Backward Compatibility**

* PyLammps now requires `numpy`, vectors are returned as **writable** `numpy.array`
* `Atom` and `Atom2D` index property is now 0-based.

**Implementation Notes**

New implementation of `OutputCapture` now uses temporary files instead of `os.pipe()`. It also includes a few `fflush()` calls to make capturing output more deterministic.

**Post Submission Checklist**


akohlmey:
@rbberger we have to be careful with using fflush() so liberally. It can negatively impact performance for some parallel applications, as a flush on MPI rank 0 may trigger synchronization MPI calls by the MPI library.
In the long run it may be better to have a global option to enable/disable flushing i/o. right now it exists only for selected operations (e.g. thermo output).

Before merging this, I would like to get the feedback from @sjplimp on  this matter and whether he thinks the cases here are low impact (IMO they are not as low impact as I would be 100% comfortable with, but the impact is acceptable).

rbberger:
@akohlmey an alternative would be adding something like `lammps_flush_screen()` or `lammps_flush_buffers()` to the library interface which can be called only when needed.

akohlmey:
> @akohlmey an alternative would be adding something like `lammps_flush_screen()` or `lammps_flush_buffers()` to the library interface which can be called only when needed.

I'd like that a **lot** better.

akohlmey:
How about this: https://github.com/lammps/lammps/compare/develop...akohlmey:flushlog-util

rbberger:
@akohlmey I'm done with this PR. If you want, you can rename `flush_buffers` to `flushlog` if you like that more.

PR #3116, opened 2022-02-03
Build KOKKOS package updates for Python module and Windows compatibility

akohlmey:
**Summary**

This pull request contains a number of changes related to the KOKKOS package that enable (minimal) KOKKOS on Windows with native MSVC compilers and improve installation/handling/loading of the LAMMPS shared library from the LAMMPS python module on multiple platforms, but specifically on Windows as well.

**Related Issue(s)**

This extends changes from PR #3102 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

- always build kokkos libraries as static libraries (like the rest of the supporting libraries) so they get included into the LAMMPS shared library; tested with bundled and downloaded Kokkos lib
- when building the downloaded kokkos library, we need to import two library files; in the bundled library there is a combined target
- add check to disallow compiling KOKKOS with OpenMP enabled if the compiler does not at least support OpenMP 4.0
- windows has no `kill()` function. Call `exit()` in the kokkos segfault signal handler instead (for now).
- replace variable length array in pppm/kokkos with new/delete as MSVC does not allow VLAs
- remove non-portable custom (and unused) profiling support code from pair style exp6/rx/kk
- add introspection to LAMMPS class test for Kokkos features to avoid incorrect errors
- skip force-style test requiring a call to the `cp` command which is only available if compiling/running LAMMPS on windows using a bash shell.
- include building the PYTHON package when compiling and testing LAMMPS on Windows with GitHub Actions
- small change to kokkos lib to make it compile with a cross-compiler

**Post Submission Checklist**


stanmoore1:
Thanks @akohlmey!

akohlmey:
@stanmoore1 FYI with these changes I can now compile and run LAMMPS with KOKKOS enabled in "serial" mode natively on Windows with MSVC compilers. OpenMP does not work because MSVC only supports OpenMP 2.0 and KOKKOS doesn't seem to like that. Also KOKKOS works with the Linux-to-Windows cross-compiler, this time *with* OpenMP since the GCC-11 cross-compiler supports OpenMP 4.5.

Unfortunately, the native is triggering a limitation of the MSVC linker that apparently cannot handle more than 65535 symbols when generating a dll file.

One important change was the elimination of the variable length array declaration in pppm_kokkos.cpp. I don't (yet) have a way to check if this works on all supported KOKKOS backends.

Overall, this should take us one step closer to have a platform neutral code base.

cmake/Modules/Packages/ML-HDNNP.cmake
```diff
@@ -64,8 +62,8 @@ if(DOWNLOAD_N2P2)
   string(TOUPPER "${CMAKE_BUILD_TYPE}" BTYPE)
   set(N2P2_BUILD_FLAGS "${CMAKE_SHARED_LIBRARY_CXX_FLAGS} ${CMAKE_CXX_FLAGS} ${CMAKE_CXX_FLAGS_${BTYPE}} ${N2P2_CXX_STD}")
   set(N2P2_BUILD_OPTIONS INTERFACES=LAMMPS COMP=${N2P2_COMP} "PROJECT_OPTIONS=${N2P2_PROJECT_OPTIONS}" "PROJECT_DEBUG="
-    "PROJECT_CC=${CMAKE_CXX_COMPILER}" "PROJECT_MPICC=${MPI_CXX_COMPILER}" "PROJECT_CFLAGS=${N2P2_BUILD_FLAGS}"
-    "PROJECT_AR=${N2P2_AR}")
+    "PROJECT_CC=${CMAKE_CXX_COMPILER}" "PROJECT_MPICC=${CMAKE_CXX_COMPILER}" "PROJECT_CFLAGS=${N2P2_BUILD_FLAGS}"
+    "PROJECT_AR=${N2P2_AR}" "APP_CORE=nnp-convert" "APP_TRAIN=nnp-train" "APP=nnp-convert")
```
stanmoore1, line 66: Were these changes intentional? They seem a little out of of place in this PR.

stanmoore1 (approved):

lib/kokkos/core/src/desul/atomics/Generic.hpp
```diff
@@ -10,8 +10,10 @@ SPDX-License-Identifier: (BSD-3-Clause)
 #define DESUL_ATOMICS_GENERIC_HPP_
 
 #include <type_traits>
+#if defined(__GNUC__)
```
stanmoore1, line 13: Should I try to get these changed merged in upstream Kokkos?

stanmoore1:
@akohlmey I added two comments above.

cmake/Modules/Packages/ML-HDNNP.cmake
akohlmey: yes. this is to work around compilation issues that popped up when compiling with KOKKOS enabled using CMake. Basically, the makefile used in the external library is not very well done for inclusion in an "automated" build. Not sure why it didn't pop up before.

lib/kokkos/core/src/desul/atomics/Generic.hpp
akohlmey: What is currently confusing is the behavior when enabling OpenMP on MSVC for example. But also using the CLANG/LLVM variant bundled with Visual Studio is difficult since it follows the MSVC compiler conventions for flags.

crtrott: Kokkos does NOT support OpenMP with MSVC since MSVCs support is insufficient by a mile. This specific define is better protected in develop to get rid of additional warnings.

stanmoore1: Kokkos developers said this change is already in the Kokkos `develop` branch: https://github.com/kokkos/kokkos/blob/09a08b7bf3514114c71dde97e8f1c11050d6c733/core/src/desul/atomics/Generic.hpp#L13-L16

crtrott: Note: we protect with: 
```c++
#if defined(__GNUC__) && (!defined(__clang__))
```

akohlmey: > Kokkos does NOT support OpenMP with MSVC since MSVCs support is insufficient by a mile.

If you read the entire conversation, you should have seen that I am aware of that. However, what confused me without end is that the Kokkos CMake support seems to be removing the flag that would otherwise enable OpenMP, and that it does not produce either an error in the CMake script itself or uses the value of the _OPENMP macro to state that OpenMP support is insufficient.

We are using the latter for reporting the OpenMP version with code like this:
```c++
std::string platform::openmp_standard()
{

#if !defined(_OPENMP)
  return "OpenMP not enabled";
#else

  // Supported OpenMP version corresponds to the release date of the
  // specifications as posted at https://www.openmp.org/specifications/

#if _OPENMP > 202011
  return "OpenMP newer than version 5.1";
#elif _OPENMP == 202011
  return "OpenMP 5.1";
#elif _OPENMP == 201811
  return "OpenMP 5.0";
#elif _OPENMP == 201611
  return "OpenMP 5.0 preview 1";
#elif _OPENMP == 201511
  return "OpenMP 4.5";
#elif _OPENMP == 201307
  return "OpenMP 4.0";
#elif _OPENMP == 201107
  return "OpenMP 3.1";
#elif _OPENMP == 200805
  return "OpenMP 3.0";
#elif _OPENMP == 200505
  return "OpenMP 2.5";
#elif _OPENMP == 200203
  return "OpenMP 2.0";
#else
  return "unknown OpenMP version";
#endif

#endif
}
```

PR #3119, opened 2022-02-08
Feature create atoms overlap

erozic:
**Summary**

Added two options/keywords to the the create_atoms command: maxtries and exclude, both to be used with the "random" style of the command.
The use of the "exclude" keyword is for the new, randomly created particles not to be created too close to any preexisting particles by "excluding" the space around them, i.e. making the space around them not available for the creation of new particles. This way a minimization/equilibration procedure in the case of "overlaps"/high energies is avoided, which is crucial when such steps/procedures are not possible, for example when trying to add particles in the middle of an already running simulation.
The "maxtries" keyword limits the number of times the program will "try" and create a new particle that satisfies all conditions (given by other keywords like var/set, region, exclude, ...) before the program gives up and throws an error (signalling the conditions are too hard to satisfy).

**Author(s)**

Eugen Rožić, Institute Ruđer Bošković (previously UCL), erozic@zoho.eu

**Licensing**

**Backward Compatibility**

Fully backward compatible

**Implementation Notes**

Minor additions to the create_atoms.h and .cpp files (and the corresponding doc file). The only bigger change that influenced existing code (but only locally in the create_atoms.cpp) is the substitution of the "add_molecule" private method with two new ones: "gen_mol_coords" and "create_mol", which effectively do the same only split the process into two parts - the generation of new molecule coordinates and the creation/addition of the new molecule to the simulation, which was necessary for the implementation of the new keywords.

Testing was done for the creation of new atoms and new molecules for various values of parameters via both a python script and a standard input script - I can provide the testing scripts, but they showed no problems and demonstrated expected and documented behavior.

**Post Submission Checklist**


jrgissing:
thanks Eugen, this feature seems useful. you might consider using 'overlap' as the keyword, to be more in line with delete_atoms, fix bond/react, fix gcmc

sjplimp:
@erozic Thanks for submitting this PR.  I've looked over the code.  Two questions:
a) It appears this makes the insertion of N atoms an N^2 process?  Since each atom is
checked against all the previously inserted atoms?
b) I'm not seeing how this works in parallel.  Only one proc may detect an overlap
and attempt to laeve the "try" loop?  Doesn't there need to be some communication, e.g. MPI_Allreduce()
to insure all procs stay or leave the loop together?

erozic:
> @erozic Thanks for submitting this PR. I've looked over the code. Two questions: a) It appears this makes the insertion of N atoms an N^2 process? Since each atom is checked against all the previously inserted atoms? b) I'm not seeing how this works in parallel. Only one proc may detect an overlap and attempt to laeve the "try" loop? Doesn't there need to be some communication, e.g. MPI_Allreduce() to insure all procs stay or leave the loop together?

@sjplimp Yes, I believe you are right about both those things. I've updated the code now fixing the synchronisation issue (with an Allreduce), but I've also left a TODO comment in the code above giving an alternative possible solution that has some caveats.
In essence I don't think the N^2 complexity can be avoided but there is in principle no need to check against particles that are on processors not "responsible" for the space where the new coords are. However, it can be a problem if the overlap distance is bigger than the skin distance, and I think this would be a significant speed up only with highly inhomogenous systems (if one proc takes a much longer time then others; because others have to wait for all to finish anyway) so I'm not sure if it's worth the hassle and effort and complications? Especially since it's intended use isn't such that this would be the time-critical part of a simulation... At least as far as I can imagine (after all, it didn't even exist until now so it's obviously not very important xD)
In any case, you're welcome :)

sjplimp:
re: N^2 cost - there will be users who attempt to use this new option to insert a million particles in an empty box w/out overlap.  Which will mean 10^12 distance comparisons.  Or worse if 1000 attempts are made per insertion (e.g. in a dense system).  So what is the intended use case(s) for this option?  Probably there are better (more efficient) tools like PackMol which would be better for the kinds of use cases this option is not designed for?  Fix gcmc is another option for inserting molecules in a moderately dense system.  I think some explanation of all this is needed in the doc page.

erozic:
> re: N^2 cost - there will be users who attempt to use this new option to insert a million particles in an empty box w/out overlap. Which will mean 10^12 distance comparisons. Or worse if 1000 attempts are made per insertion (e.g. in a dense system). So what is the intended use case(s) for this option? Probably there are better (more efficient) tools like PackMol which would be better for the kinds of use cases this option is not designed for? Fix gcmc is another option for inserting molecules in a moderately dense system. I think some explanation of all this is needed in the doc page.

Well, the intended use case (what I needed it for) was to insert some (few) new atoms/molecules mid simulation in a relatively sparse system. That's why minimastion after creation of new particles was not an option (I wanted to continue the simulation where it paused), and I didn't want it to break without control when all it needed was to try once or twice more to find a position for the new atom/molecule that wouldn't overlap.
The maxtries option seemed to me useful on its own because maybe the var or other conditions are too hard to satisfy so maybe someone wants to control that, because currently it's a potentially endless loop with infinite tries...

So yeah, maybe you are right, maybe it should be noted in the doc file that it is an expensive option not intended to be used for dense systems and that the simultaneous use of "maxtries" with a low number is advised. Would that be OK?

erozic:
> ok - why don't you add some context explanation to the doc page, maybe in a Note. E.g. when this options is useful and efficient. And when you should not use it b/c inefficient. Then I will take another look at the PR. Thanks

@sjplimp Done! Hope this is good enough :) Also, I tested it again now, after the code change, and it works fine on multiple processors.

sjplimp:
@erozic I don't understand the logic in the check for overlaps.  It appears to just be checking for the new xone, which is fine for atoms.  But for molecules it isn't checking for each of the new atoms in the molecule?  I thought that was the point of separating the generation of atoms in the molecule versus adding the atoms in the molecule.  I.e. to do the overlap check before accepting the molecule insertion?  If that still needs to be done, then I can add it.  I am in the process of making some other changes to the PR, and can then push it back to you.

sjplimp:
@erozic OK, this is ready for you to take another look at.  I added some needed logic, like checking for PBC during overlaps, to the code, and reworked the doc page somewhat, adding an image and usage commands to show how the new overlap options work.

If you are satisfied with all this, the only remaining issue is this.  I don't see why the generate_molecule() and add_molecule() operations needs to be split up.  Before they were just one method, correct?  But I don't see that it matters to the overlap check since it uses molecule radius instead of a more careful atom-by-atom check (which would require the coords of each atom in the molecule).  So why don't we leave that method the way it was for simplicity?

erozic:
> @erozic I don't understand the logic in the check for overlaps. It appears to just be checking for the new xone, which is fine for atoms. But for molecules it isn't checking for each of the new atoms in the molecule? I thought that was the point of separating the generation of atoms in the molecule versus adding the atoms in the molecule. I.e. to do the overlap check before accepting the molecule insertion? If that still needs to be done, then I can add it. I am in the process of making some other changes to the PR, and can then push it back to you.

@sjplimp You are right, that was the point of separating generation of mol coords and creating the mol atoms, and you are right, it's not necessary since only the center (xone) is checked and molecule radius used. I think it's better/cleaner/faster/simpler that way, since this is not meant to be precise anyway, so I propose we keep that and do as you said, join back the methods for creating a new molecule into a single method (I took the liberty and did it already).

However, I think you made a mistake (that I fixed) in the code where overlaps are checked when you put the "mode == ATOM" condition, xone should be checked against all atom locations regardless of the mode (it's either a new atom location or a new molecule center location).

> @erozic OK, this is ready for you to take another look at. I added some needed logic, like checking for PBC during overlaps, to the code, and reworked the doc page somewhat, adding an image and usage commands to show how the new overlap options work.

Thank you for all the effort, I'm satisfied with everything if you are :) The only thing I would comment is that I thought it better for the command to fail if it didn't manage to create all the requested atoms, not just throw a warning, because user might want the exact amount created... but then again, they can make an effort to check that "manually" after the command so I guess it's OK, maybe even better this way.

sjplimp:
@erozic Thanks for the code updates.  Esp catching the mode = ATOM bug.  I had added that thinking there would be an else clause for mode = MOLECULE which would check all the new atoms in the molecule.  But I agree that it is fine to just check the molecule center and include its radius in the distance criterion.  I made any failure a warning, b/c I think with the maxtry dependence, there will potentially lots of use cases with failures,  The picture I put in the doc page is a moderately dense liquid, where I just used a big N and reasonable maxtry to get ~70% of N as sucesses.  I think that is a nice way to produce a liquid or amorphous solid which is ready-to-run with MD w/out danger of overlaps.

All I did in my final commit was edit a few of the comments.
@akohlmey this should now be ready to test and merge

sjplimp (approved):

erozic:
@akohlmey I tested it, it worked fine after this last bugfix of setting the initial value of quat_user (made huge problems, took me a while to figure it out, little bastard xD)

PR #3123, opened 2022-02-11
Remove incorrect error check in ReaxFF

stanmoore1:
**Summary**

https://github.com/lammps/lammps/commit/42e8a7613e144e86b8ce0eac302b5fc87323779e prevents running more than 2 billion atoms with ReaxFF, however:

- `bigN` is actually is defined as a `bigint` in `reax_types.h`, so the static_cast to `int` is wrong (and misleading)
- `bigN` actually isn't used anywhere and can just be removed

**Related Issue(s)**

https://github.com/lammps/lammps/commit/42e8a7613e144e86b8ce0eac302b5fc87323779e

**Author(s)**

Stan Moore (SNL), reported by Nick Curtis (AMD), @arghdos 

**Licensing**

**Backward Compatibility**

Yes


athomps (approved):
I approve

PR #3124, opened 2022-02-11
Reduce number of visible symbols in LAMMPS library by removing factory function templates from class definition

akohlmey:
**Summary**

The factory functions to create the various style class instances in LAMMPS are not accessed outside their respective implementation file. This PR makes them static local function of the translation units and thus should significantly reduce the
number of visible symbols in the LAMMPS library. This should help to speed up linking and to work around some linker limitations. We have over 1500 styles when LAMMPS is compiled with all features included.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

N/A

**Post Submission Checklist**

### DIFF
--- a/src/atom.cpp
+++ b/src/atom.cpp
@@ -54,6 +54,15 @@ using namespace MathConst;
 #define DELTA_PERATOM 64
 #define EPSILON 1.0e-6
 
+/* ----------------------------------------------------------------------
+   one instance per AtomVec style in style_atom.h
+------------------------------------------------------------------------- */
+
+template <typename T> static AtomVec *avec_creator(LAMMPS *lmp)
+{
+  return new T(lmp);
+}
+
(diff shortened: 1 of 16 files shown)


athomps (approved):
It seems like a pretty small saving, since I count about 26,000 text symbols in LAMMPS. Will removing 1500 of these really make much difference? I suppose it could in the future.

akohlmey:
> It seems like a pretty small saving, since I count about 26,000 text symbols in LAMMPS. Will removing 1500 of these really make much difference? I suppose it could in the future.

Agreed, it would be nice, if there was more. On the other hand, it is a rather significant saving considering the small number of changes. Also, 26k is what I see without KOKKOS. With KOKKOS this number triples.

stanmoore1 (approved):

PR #3090, opened 2022-01-18
Modernize to C++11: use override

rbberger:
**Summary**

This PR semi-automatically applies clang-tidy suggestions to make use of the C++11 `override` keyword.

The changes use `override` following the [C++ Core Guideline C.128](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-override), which is a collection of best practices.

It also removes the empty destructor of derived classes. (Option B in below discussion)

**Related Issue(s)**

Closes #2946

**Author(s)**

Richard Berger (Temple U)

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

All automated tests pass.

**Post Submission Checklist**

- [x] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [x] The source code follows the LAMMPS formatting guidelines
- [x] Suitable new documentation files and/or updates to the existing docs are included
- [x] The added/updated documentation is integrated and tested with the documentation build system
- [x] The feature has been verified to work with the conventional build system
- [x] The feature has been verified to work with the CMake based build system

PR #3125, opened 2022-02-12
Anisotropic Interlayer potential for TMDs and Metal Surfaces

oywg11:
**Summary**

Add two anisotropic Interlayer potentials:
1, for transition metal dichalcogenide and its heterostructures
2, for graphene/metal heterojunctions

**Related Issue(s)**

 None

**Author(s)**

Wengen Ouyang, Wuhan University, w.g.ouyang@gmail.com

**Licensing**

**Backward Compatibility**

Yes

**Implementation Notes**

The new ILPs are added as derived classes of pair ilp/graphene/hbn to avoid code duplication.
The main modifications are the definitions of atomic normal vectors for different materials.

**Post Submission Checklist**

https://pubs.acs.org/doi/10.1021/acs.jctc.1c00782
https://pubs.acs.org/doi/10.1021/acs.jctc.1c00622


akohlmey:
@oywg11 thanks for your contribution. I had it updated to the latest upstream code and reverted the changes that were undoing the updates from PR #3090 as well as update the new virtual functions accordingly. This also needed some additional changes in the documentation to have it fully integrated and spelling and anchor label issues removed. For the `ilp/graphene/hbn` pair style I also simplified the per style labels with a local map.
Please check if everything is still working as expected.

oywg11:
> @oywg11 thanks for your contribution. I had it updated to the latest upstream code and reverted the changes that were undoing the updates from PR #3090 as well as update the new virtual functions accordingly. This also needed some additional changes in the documentation to have it fully integrated and spelling and anchor label issues removed. For the `ilp/graphene/hbn` pair style I also simplified the per style labels with a local map. Please check if everything is still working as expected.

@akohlmey Thanks for the improvement. I recompiled the codes and confirmed that they give the same results as before.

src/INTERLAYER/pair_ilp_graphene_hbn.cpp
```diff
@@ -56,9 +57,15 @@ static const char cite_ilp[] =
     " year =    2018,\n"
     "}\n\n";
 
+// to indicate which potential style was used in outputs
+static std::map<int, std::string> variant_map = {
```
sjplimp, line 61: not clear why this should be a static variable?

sjplimp: what I mean is, is there a better way to do this w/out a static variable?

akohlmey: because it is relevant for and should be visible only in this particular file and this has less overhead than adding it to the class as a member object.
This way the class can use the map to customize error messages to print the correct style name when there are errors from derived classes.

sjplimp: ok - makes sense

akohlmey: > what I mean is, is there a better way to do this w/out a static variable?

I don't get your point. IMO a static variable (mind you this is actually a class instance) is the best solution that I can think of.
Why is 'static' bad in this case? Dropping the static would make it globally visible and then it would need to declared in a suitable namespace to avoid conflicts.

akohlmey:
> I realize some ILP files pre-date this PR, but now would be a good time to change the names if Axel agrees.

I had been pondering the same thing. So my first impulse was ask for something similar, but in the interest of having this merged for the patch release I thought it would be better to postpone this. My reason is threefold: 1) most of the pair styles and potential files are for rather specific cases, so it is not likely to have a large library of them (like for eam, tersoff, and similar), 2) when renaming the files for more consistent suffixes it should also be looked into more consistent names for the pair styles, if possible and meaningful, 3) it is probably necessary to do a more general audit of the potential files in the potentials folder.

Bottom line: I am undecided on whether to do this within this PR or later.

sjplimp:
I think it is fine to leave this PR as-is, i.e. ready to merge.

For a future PR on potential dir filenames, I think the key consistency to maintain (which suffix=ILP breaks) is that the suffix of the potential file be an exact match to the pair style name.  E.g. eam or eam.alloy.  All lowercase, like the pair style names.
So that if someone browses the potentials dir and wonders what this Carbon potential file is good for, they can use the suffix to go to the pair style doc page directly.  No ambiguity.

If some potential files can be used with multiple pair styles (e.g. ILP package variants), then possibly suffix = base name of multiple pair styles, and the pair style doc page can clarify that these 3 pair style variants use the same potential files.

sjplimp:
Also, changing pair style names to be more informative is a separate issue.  Anytime that is done, the corresponding potential file names should be changed accordingly.

akohlmey:
> I think it is fine to leave this PR as-is, i.e. ready to merge.

I won't be able to merge it, though, for as long as your request for changes is blocking it.

oywg11:
> I think it is fine to leave this PR as-is, i.e. ready to merge.
> 
> For a future PR on potential dir filenames, I think the key consistency to maintain (which suffix=ILP breaks) is that the suffix of the potential file be an exact match to the pair style name. E.g. eam or eam.alloy. All lowercase, like the pair style names. So that if someone browses the potentials dir and wonders what this Carbon potential file is good for, they can use the suffix to go to the pair style doc page directly. No ambiguity.
> 
> If some potential files can be used with multiple pair styles (e.g. ILP package variants), then possibly suffix = base name of multiple pair styles, and the pair style doc page can clarify that these 3 pair style variants use the same potential files.

@sjplimp Thanks for the suggestion. The reason why I didn't follow the rules is that the potential file names would be super long, for instance, BNCH.ILP should be named as BNCH.ilp.graphene.hbn. So I used the element to separate the potential files (i.e, the prefix). Another reason is that the interlayer potentials I submitted actually share the same formula. The main difference between them is the definition of the atomic normal vectors for different materials. If you think the long potential file name is not an issue, I'm also fine with renaming all the potential files.

oywg11:
> 

For your information, here I list the applicable scope of three ILPs:
1, pair ilp/graphene/hbn is applicable to the following heterojunctions: graphene/graphene, hBN/hBN, graphene/hBN and hydrocarbons/graphene, hydrocarbons/hBN;
2, pair ilp/tmd is applicable to the following heterojunctions: MX2/MX2 (M=Mo,W; X=S,Se,Te), graphene/MX2 and hBN/MX2;
3, pair saip/metal is applicable to the following heterojunction: graphene/metal(=Au,Cu,Ag,Ni) and hBN/metal(=Au,Cu,Ag,Ni).

sjplimp:
@oywg11 Long potential file names is not a problem.  As I mentioned previously, the more important consideration is clarity for users as to which potential file to use with which pair style.

But we can adjust the file names later.  I'll appove this PR.  Thanks again for submitting it to LAMMPS.

sjplimp (approved):

PR #3126, opened 2022-02-12
Added Python version of Zhou04_create_v2.f in eam_database

Bibobu:
**Summary**

In recent mail discussion, it was suggested Python translation of EAM input datafiles generator would be useful, given the old Fortran style of the original programs. Here is a Python translation of the NIST revised version corrected by Lucal Hale.

**Related Issue(s)**

None.

**Author(s)**

Original code: X. W. Zhou, xzhou@sandia.gov$                                  
Updated by: Lucas Hale lucas.hale@nist.gov$                            
Python version by: Germain Clavier g.m.g.c.clavier@tue.nl

**Licensing**

**Backward Compatibility**

No Lammps backward compatibility issue is to be expected. The generated input files are quasi exact match of the one obtained through the Fortran code. The exception being the format (engineer format is a bit tricky to get in Python and not interesting for 16th digits precision difference). Comparison has been done for Ta/Cu and W/Co couples and gave exact same energy and pressure.

**Implementation Notes**

A Python translation of the file EAM_code has been done in eamDatabase.py. The numpy library is needed to execute the code. An implementation without numpy dependence could be done if this represents a restriction.

**Post Submission Checklist**

### DIFF
--- a/tools/eam_database/test/in.lmp
+++ b/tools/eam_database/test/in.lmp
@@ -9,15 +9,15 @@ pair_coeff * * @FORTRANFILE@ @ELEM1@ @ELEM2@
 
 variable etot equal etotal
 variable press equal press
-thermo_style custom pe etotal press 
+thermo_style custom pe etotal press
 
-run 0
+run 0 post no
 
 variable efor equal ${etot}
 variable pfor equal ${press}
 thermo_style custom pe etotal v_efor press v_pfor
 
-run 0
+run 0 post no
 
 pair_coeff * * @PYTHONFILE@ @ELEM1@ @ELEM2@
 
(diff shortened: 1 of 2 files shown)


athomps:
The LAMMPS energy test is a good one.  This should really be done for all element pairs, to check that eamDatabase.py is faithful to original for all elements.

akohlmey (commented):
@Bibobu Sorry, but this is not how we add tests that can run with the automated testing and it is not sufficiently portable either. I will implement something later, when I am finished with some other cleanup/fixup work on a different pull request.

Bibobu:
@akohlmey No problem. As the documentation relative to the tools test part only contains info relative to the LAMMPS shell I was clearly unsure on how to proceed.

If it is useful, you can still use the shell loop and the test system I provided. It was essentially to confirm the match of both file as I got a flat zero relative difference for the pressure and energy computation of each pair.

athomps (approved):
I approve.

athomps:
> I got a flat zero relative difference for the pressure and energy computation of each pair.
@Bibobu Good.  That guarantees correctness.

akohlmey:
> @akohlmey No problem. As the documentation relative to the tools test part only contains info relative to the LAMMPS shell I was clearly unsure on how to proceed.

Yeah, testing is working progress in many places.

> If it is useful, you can still use the shell loop and the test system I provided. It was essentially to confirm the match of both file as I got a flat zero relative difference for the pressure and energy computation of each pair.

I tried to implement what I had in mind (run the create_eam.py script and the executable from a test script and then import the LAMMPS python interface to run the inputs and extract the results) and got to a point, where it was clear that I would get it to work, but also that it would be too much work (at least with my currently rather limited time) for the gain.
Thus I committed the changes to the create_eam.py script and made a few tweaks to the other scripts and deleted unwanted files. Since you **have** validated that everything works and this is not something that is modified a lot, it should be good enough to leave the test folder in there and rerun the tests when more elements are added.

Some point in the future, when the fortran code won't compile, it will just be the python code and then the test will be redundant.

PR #3127, opened 2022-02-12
Update compiling OpenCL loader lib

akohlmey:
**Summary**

This updates the CMake script code to build the OpenCL ICD loader library to be compatible with multi-config builds (like MSBuild from Visual Studio) and embed the build into the CMake hierarchy of LAMMPS' build process.
This also updates the version of the loader sources to the latest release (3.0.10).
To avoid undesired installation of the OpenCL library and headers due to embedding of the CMake scripting, an option was added to the loader that allows to skip installation (similar to what googletest uses).

A few minor inconsistencies in the function prototypes and issues pointed out by compiler warnings in the GPU library were addressed as well.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues

**Post Submission Checklist**

### DIFF
--- a/cmake/Modules/OpenCLLoader.cmake
+++ b/cmake/Modules/OpenCLLoader.cmake
@@ -1,6 +1,6 @@
 message(STATUS "Downloading and building OpenCL loader library")
-set(OPENCL_LOADER_URL "${LAMMPS_THIRDPARTY_URL}/opencl-loader-2021.09.18.tar.gz" CACHE STRING "URL for OpenCL loader tarball")
-set(OPENCL_LOADER_MD5 "3b3882627964bd02e5c3b02065daac3c" CACHE STRING "MD5 checksum of OpenCL loader tarball")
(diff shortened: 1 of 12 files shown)


ndtrung81 (approved):
thank you @akohlmey

junghans:
> Is there a public git repository where that tarball is generated from? I understand this is based on https://github.com/KhronosGroup/OpenCL-ICD-Loader, but with modifications.

Yeah, I would like to know that, too. a.) for copyright reason, but also because b.) Fedora would actually prefer to unbundle this.

akohlmey:
> Yeah, I would like to know that, too. a.) for copyright reason, but also because b.) Fedora would actually prefer to unbundle this.

No need to use this code, just set `-D USE_STATIC_OPENCL_LOADER=off` as documented here and use the system provided loader: https://docs.lammps.org/Build_extras.html#gpu

PR #3142, opened 2022-02-19
SNAP bik flag

charlessievers:
**Summary**

Added in flag for snap compute to provide per atom bispectrum descriptors instead of summed bispectrum descriptors.

**Author(s)**

Charlie Sievers - Sandia National Labs - charliesievers@cox.net

**Licensing**

**Implementation Notes**

Tested file examples/snap/in.snap.compute and produced the same output.

**Post Submission Checklist**


athomps (changes requested):
This looks nice and clean.   The only problem is that the treatment of the last column (reference potential) is not clearly defined in the code or in the documentation.  I suggest making a clear choice e.g. 1) `snap all[irow][lastcol] = 0;` for all b_ik rows 2) same, but with reference energy in first b_ik row 3) Put per-atom reference energy in every b_ik row.  Whatever choice is made should be explicitly documented.

doc/src/compute_sna_atom.rst
```diff
@@ -296,6 +299,14 @@ This option is typically used in conjunction with the *chem* keyword,
 and LAMMPS will generate a warning if both *chem* and *bnormflag*
 are not both set or not both unset.
 
+The keyword *bikflag* determines whether or not to expand the bispectrum
+rows of the global array returned by compute snap. If *bikflag* is set
+to *1* then the bispectrum row, which is typically the per-atom bispectrum
+descriptors :math:`B_{i,k}` summed over all atoms *i* to produce
+:math:`B_k`, becomes bispectrum rows equal to the number of atoms. Thus,
+the resulting bispectrum rows are :math:`B_{i,k}` instead of just
+:math:`B_k`.
```
athomps, line 308: Add "In this case, the entries in the final column for these rows are set to zero."

athomps (approved):
Still need to describe final column in documentation.

doc/src/compute_sna_atom.rst
charlessievers: Done

akohlmey: @athomps @charlessievers as a reminder. If you both consider this to be reading for merging, you need to assign the PR to me.

PR #3146, opened 2022-02-21
New developer page for comm functions

sjplimp:
**Summary**

New doc page for Programmer Guide which describes various communication operations provided in LAMMPS.  Which users writing new style classes can invoke from their new code.

**Related Issue(s)**

N/A

**Author(s)**

Steve

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/doc/src/Developer_comm_ops.rst
+++ b/doc/src/Developer_comm_ops.rst
@@ -1,8 +1,9 @@
-Communication operations
-------------------------
+Communication patterns
+----------------------
 
 This page describes various inter-processor communication operations
-encoded by LAMMPS, mostly in the core *Comm* class.  They are used by
+provided by LAMMPS, mostly in the core *Comm* class.  These are operations
+for common tasks implemented using MPI library calls.  They are used by
 other classes to perform communication of different kinds.  These
 operations are useful to know about when writing new code for LAMMPS
 that needs to communicate data between processors.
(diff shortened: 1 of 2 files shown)


sjplimp:
@akohlmey Here is draft of the new doc page we talked about.  Not sure where in the Developer section you envision it going.

akohlmey:
> @akohlmey Here is draft of the new doc page we talked about. Not sure where in the Developer section you envision it going.

thanks. How do you want to proceed?
- should I do a pass editing it and then you check it out again and finalize it?
- should I add comments with suggestions and you do the edits?

sjplimp:
1st bullet is fine

doc/src/Developer_comm_ops.rst
```diff
@@ -0,0 +1,209 @@
...
+to pack or unpack.  The *Pair* class also sets the *comm_forward* and
+*comm_reverse* variables which store the number of values it will
+store in the communication buffers for each operation.  This means it
+can choose to store multiple per-atom values in the buffer if desired,
+and they will be communicated all together.
+
+The *Fix*, *Compute*, and *Dump* classes can invoke the same forward
+and reverse communication operations using these *Comm* class methods:
+
+* forward_comm_fix()
```
akohlmey, line 57: @sjplimp Do you remember a particular reason why there is `forward_comm_pair(), `forward_comm_fix()` and so on instead of just `forward_comm()`?
With the exception of the `forward_comm_fix_variable()` those are all uniquely identifiable from the type of the first argument. So we could have them all named `forward_comm()` and the variable one `forward_comm_variable()`.

sjplimp:
You mean just change the method names, not template them?
That would be fine I guess.  What is the int *dummy* arg for forward_comm() ?
Maybe a Kokkos thing?  Will prototyping on the 1st arg mess up Kokkos?

akohlmey:
> You mean just change the method names, not template them?
yes. I don't think they can be templated. Won't save us anything. This is more a consistency thing, i.e. avoid the "hungarian notation", i.e. have the function name encodes the type of arguments. That is needed in C when you cast to `void *`.
I am not 100% convince this is a good idea. Having the same function do the same procedure is good and overloading useful, but those are also virtual functions and thus overloading them makes it harder to detect incorrect use.

> That would be fine I guess. What is the int _dummy_ arg for forward_comm() ? Maybe a Kokkos thing? Will prototyping on the 1st arg mess up Kokkos?

Should be OK, but for certain such a change would have to be a separate pull request with all uses updated and tested.

akohlmey:
> 1st bullet is fine

@sjplimp I have added my edits and also integrated it into the manual. For now, I just left it as a separate file and made it a new toc item. Looks quite good to me. Has pretty much everything I was looking for. I only added a few more details for better clarity. Fixed some spelling and formatting things (including your request for subheadings). Please check it out.

doc/src/Developer_comm_ops.rst
akohlmey: implemented in #3148. The text here is updated to reflect that PR.

doc/src/Developer_comm_ops.rst
```diff
@@ -0,0 +1,226 @@
...
+And second, *reverse communication* which sends ghost atom information
+from each processor to the owning processor to **accumulate** (sum) the
+values with the corresponding owned atoms.  The need for this arises
+when data is computed and also stored with ghost atoms (e.g. forces when
+using a "half" neighbor list) and thus those terms need to be added to
+their corresponding atoms on the process where they are "owned" atoms.
+Please note, that with the :doc:`newton off <newton>` setting this does
+not happen and the neighbor lists are constructed that these pairs are
+computed on both MPI processes containing one of the atoms and only the
+data pertaining to the local atom is stored.
```
athomps, line 40: 'stored' -> 'accumulated'

doc/src/Developer_comm_ops.rst
```diff
@@ -0,0 +1,226 @@
...
+The *Fix*, *Compute*, and *Dump* classes can also invoke the same kind of
+forward and reverse communication operations using *Comm* class methods.
+The same pack/unpack methods and comm_forward/comm_reverse variables
+must be defined by the *Fix* or *Compute* class.
+
+For *Fix* classes there is an optional second argument to the
+*forward_comm()* and *reverse_comm()* call which are used when there are
+communications with buffer sizes different from the value set in the
+comm_forward/comm_reverse variables.  For this to work a class member
+variable must be set to indicate to the pack/unpack routines which
```
athomps, line 88: `which.  which` -> `which`

athomps (approved):

athomps:
One additional thing that could be mentioned near the top of this section or elsewhere is that ghost atoms and forward/reverse communication are also used to implement periodic boundary conditions.

PR #3153, opened 2022-02-22
Log pizza updates

### LINKED ISSUE #3151: Log parsing and plotting in python 2
**Summary**

I am new to lammps and went to use the log parsing/plotting tools, specifically the files logplot.py and log2txt.py in lammps/tools/python/. Most of these tools are written in python 2 and some are called in a way that doesn't work on linux systems. I have a few commits that fix this, and I wanted to ask if I should open a PR for those.

I'm on commit 7ea26161 and running this on Fedora 35.

**Details**

I have 3 commits that I can submit as a PR but I first wanted to run them by you to make sure they are all reasonable. This is what's included:
1. The scripts are in python 2, so they need upgrading. They also have some confusing variable names (`list` and `map`), so I have fixed those, too. However, I have upgraded only logplot.py and log2txt.py. Is it okay to submit an upgrade to just some of the scripts in this folder?
2. logplot.py calls python using `#!/usr/bin/env python -i`, which I can't get to work in linux, so I removed the -i so that it can run (non-interactively) in linux. Is that okay?
3. log2txt.py also creates a file of unnamed columns. In all use cases I can think of, it's preferable to also include column names so that humans or plotting software can more easily read it later. I would also propose to write out column names, but if you don't think that's a generally useful change, I can keep that in my own version.

aliehlen:
**Summary**

This updates the files logplot.py and log2txt.py in lammps/tools/python/. Most of these tools are written in python 2 and some are called in a way that doesn't work on linux systems. To address this, these three commits do the following:
1. The scripts are in python 2, so they need upgrading. They also have some confusing variable names (`list` and `map`), so I have fixed those, too. However, I have upgraded only logplot.py and log2txt.py. Is it okay to submit an upgrade to just some of the scripts in this folder?
2. logplot.py calls python using `#!/usr/bin/env python -i`, which I can't get to work in linux, so I removed the -i so that it can run (non-interactively) in linux. Is that okay?
3. log2txt.py also creates a file of unnamed columns. In all use cases I can think of, it's preferable to also include column names so that humans or plotting software can more easily read it later. I would also propose to write out column names, but if you don't think that's a generally useful change, I can keep that in my own version.  

**Related Issue(s)**

closes #3151 

**Licensing**

**Backward Compatibility**

Yes, I think that this might break compatibility with python 2. 

**Implementation Notes**

Other than upgrading to python 3, this change behavior in 2 ways:
- logplot.py does not run in interactive mode
- log2txt.py will create a log file with column names (previous behavior was that no column names were included)

**Post Submission Checklist**

### DIFF
--- a/doc/src/Developer_comm_ops.rst
+++ b/doc/src/Developer_comm_ops.rst
@@ -90,7 +90,7 @@ For *Fix* classes there is an optional second argument to the
 fix performs multiple modes of communication, with different numbers
 of values per atom.  The fix should set the *comm_forward* and
 *comm_reverse* variables to the maximum value, but can invoke the
-communication for a particulard mode with a smaller value.  For this
+communication for a particular mode with a smaller value.  For this
 to work, the *pack_forward_comm()*, etc methods typically use a class
 member variable to choose which values to pack/unpack into/from the
 buffer.
--- a/tools/python/dump2cfg.py
+++ b/tools/python/dump2cfg.py
@@ -16,7 +16,7 @@ from dump import dump
 from cfg import cfg
 
 if len(sys.argv) != 8:
-  raise StandardError, "Syntax: dump2cfg.py dumpfile Nid Ntype Nx Ny Nz cfgfile"
+  sys.exit("Syntax: dump2cfg.py dumpfile Nid Ntype Nx Ny Nz cfgfile")
 
 dumpfile = sys.argv[1]
 nid = int(sys.argv[2])
--- a/tools/python/dump2pdb.py
+++ b/tools/python/dump2pdb.py
@@ -18,7 +18,7 @@ from dump import dump
 from pdbfile import pdbfile
 
 if len(sys.argv) != 8 and len(sys.argv) != 9:
-  raise StandardError, "Syntax: dump2pdb.py dumpfile Nid Ntype Nx Ny Nz pdbfile template"
+  sys.exit("Syntax: dump2pdb.py dumpfile Nid Ntype Nx Ny Nz pdbfile template")
 
 dumpfile = sys.argv[1]
 nid = int(sys.argv[2])
--- a/tools/python/dump2xyz.py
+++ b/tools/python/dump2xyz.py
@@ -16,7 +16,7 @@ from dump import dump
 from xyz import xyz
 
 if len(sys.argv) != 8:
-  raise StandardError, "Syntax: dump2xyz.py dumpfile Nid Ntype Nx Ny Nz xyzfile"
+  sys.exit("Syntax: dump2xyz.py dumpfile Nid Ntype Nx Ny Nz xyzfile")
 
 dumpfile = sys.argv[1]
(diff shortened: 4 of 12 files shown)


akohlmey:
@aliehlen thanks for the submission. I have made a couple of minor changes to retain compatibility with python 2.
I think this is good to be included, but let's wait for the verdict of @sjplimp, who wrote this code originally.

sjplimp:
@aliehlen Thanks for doing these updates for Py 3.  The code changes look basically fine.  @akohlmey Are you saying these changes now work for both Py2 and Py3.
That would be ideal if possible. 
Re: original 3 Qs:
(1) fine to only update some files
(2) The reason to launch logplot.py with -i, is that it displays a plot, which will immediately disappear if the script exits.  Maybe there is some more portable way to do this in the first line of the script?   But I think that functionality is important to preserve
(3) For log2txt.py I agree column header info is nice.  Is there a portable way to do this that can be read by all common plotting tools?  If not, I suggest doing it the way you think is best, but make it an option.  E.g. add an optional -h switch to the log2txt.py syntax (before the log.lammps arg) which also outputs the col headers.

sjplimp:
@allehlen - note there is also a lammps/pizza repo which has log.py in it - not sure if it is identical to the log.py in the lammps/tools repo - but it would be nice to have these Py 3 updates in it as well

akohlmey:
>  @akohlmey Are you saying these changes now work for both Py2 and Py3. That would be ideal if possible.
yes.

>  (2) The reason to launch logplot.py with -i, is that it displays a plot, which will immediately disappear if the script exits. Maybe there is some more portable way to do this in the first line of the script?

It does not disappear for me. but there is a way around it, too: 
just call [`code.interact()`](https://docs.python.org/3/library/code.html) at the end of the script.
I tested it successfully with Python3 and Python2.7.

akohlmey:
@aliehlen are you going to add (some of) the suggested changes to this pull request? Please let us know.

aliehlen:
@akohlmey Yes, sorry about that. I can do these, though it will be later this week or early next week, if that's okay.

akohlmey:
> @akohlmey Yes, sorry about that. I can do these, though it will be later this week or early next week, if that's okay.

No problem. We just need to know if a person will update a pull request of if we have to do it ourselves. I want to avoid that multiple people start working on this but also that this pull request sits there and waits for people to complete it.

aliehlen:
That makes sense!

aliehlen:
Okay, I have added an optional switch -n to turn on the writing of column names. (The argparse packages uses -h to display a help message, so I had to pick a different letter.) I'll look to see if it's easy to apply these changes to log.py in the lammps/pizza repo now, but that will be a separate thing, but I think this was the only missing action item for this PR.

akohlmey:
@sjplimp does this address all your concerns?

sjplimp (approved):
looks good - thanks to both of you for the changes

PR #3154, opened 2022-02-25
Refactor python/install.py and 'make install-python' to use setuptools and build a binary wheel

akohlmey:
**Summary**

Python distutils are deprecated in Python 3.11 and will be removed in 3.12. So scripts using it need to be updated to use setuptools instead. The `python/setup.py` script used by `make install` (only available with CMake build process) was already updated. This will create a *universal* package that is dependent on having the LAMMPS shared library installed in a system folder.

As an alternative there is the `python/install.py` script to install the LAMMPS python module *with* the corresponding shared library file into a suitable `site-packages` folder for transparent use of the module **without** having to set both the PYTHONPATH and the LD_LIBRARY_PATH (or DYLD_LIBRARY_PATH or PATH) environment variables. It is called from `make install-python` or can be called manually. This has been refactored to use setuptools as well.

**Related Issue(s)**

PR #3101 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

It is no longer possible to explicitly provide a directory to install the module and shared library to with the -d flag.
Instead using a virtual environment is recommended. The refactored procedure is compatible with that, while the old version was not.

**Implementation Notes**

The procedure has been changed to first create a virtual environment for the python version chosen, then installing required packages to build a so-called *binary wheel* into that and then build the wheel file itself. The install.py script can be called manually and is called from `make install-python` from either the CMake build folder or the src folder.  After building the wheel the virtual environment will be left and removed and the binary wheel file installed into the current site-packages folder. The binary `lammps-<version>-<pyversion>-<pyabi>-os>-<arch>.whl` file is moved to the build folder and can later be directly installed with pip (again) on the same or any compatible machine.

The procedure as of now is rather messy but surprisingly portable.

**Post Submission Checklist**


akohlmey:
@sjplimp @athomps this pull request has my proposed changes to the python/install.py scripts and python/setup.py to refactor it for switching from distutils to setuptools. I've tried to retain as much of the old behavior as possible, but the custom folder installation with `-d` is not possible anymore. Instead it now leaves the "binary wheel" file around which can be installed manually with pip (e.g. into a later created virtual environment). I am not proud of *how* this is implemented, but happy that I got everything somehow working for all tests I made.

Now it would be time to see if this still serves all needs that we need it to work for and on machines that are different from my desktop or containers or virtual machines.

akohlmey:
I will revise the manual and add some more comments to the script code describing the steps better, once we agree on whether to stick with this implementation or go a different route.

charlessievers:
@akohlmey Just tried this PR out on my local machine, MacOS 10.15.7. I use anaconda to manage my python environments. make install-python fails to add liblammps.* to the python site-packages of my current environment. I used cmake version 3.16.2. Where is the target location for the binary wheel?

akohlmey:
> Just tried this PR out on my local machine, MacOS 10.15.7. I use anaconda to manage my python environments.

You are using two things that are among the biggest headaches in portable software development: MacOS and anaconda.

In recent years, particularly Apple's software engineers have managed to achieve that MacOS is now "outweirding" Windows, which is not an easy feat. I tried the PR on Windows in a standard shell (not bash) and it worked out of the box...

> /Users/casieve/anaconda3/envs/stubs/bin/python3.8: No module named virtualenv

**Here** is the problem. The wheel build procedure creates a virtual environment (because it needs to install/update some packages via pip), but your Python is lacking virtualenv to create it. Looks like I need to add another sanity check.

charlessievers:
> You are using two things that are among the biggest headaches in portable software development: MacOS and anaconda.

Glad I could be of service then, at least I switched to cmake. Let me know if you need me to test out future implementations.

> **Here** is the problem. The wheel build procedure creates a virtual environment (because it needs to install/update some packages via pip), but your Python is lacking virtualenv to create it. Looks like I need to add another sanity check.

Makes sense, just thought you would want to know.

akohlmey:
> Where is the target location for the binary wheel?

The file should show up in the "build" folder, i.e. wherever you configured it for cmake or the src folder with GNU make.
Once you have the .whl file you won't need to use `make install-python` anymore, but could directly do `pip install ./lammps-2022.2.17-cp310-cp310-linux_x86_64.whl` (or whatever the file is called based on python version, OS and CPU).
Also uninstall should be simple with `pip uninstall lammps`.

akohlmey:
> Glad I could be of service then, at least I switched to cmake. Let me know if you need me to test out future implementations.
> 
> > **Here** is the problem. The wheel build procedure creates a virtual environment (because it needs to install/update some packages via pip), but your Python is lacking virtualenv to create it. Looks like I need to add another sanity check.
> 
> Makes sense, just thought you would want to know.

Absolutely. Will be pushing an update in a few minutes. MacOS is always a bit more difficult for me to test because the only test machines are either our jenkins workers or through GitHub Actions. 

I will also have to check in my FreeBSD virtual machines. That one has an unusual way to manage python versions so it is sometimes complicated to write scripts in a fully portable ways. 

The strangest thing is that this new version seems to be still compatible with Python 2.7 (I didn't aim for or expected it). There are some warnings about EOL and lack of updates, but yet it ran to completion and produced wheels on multiple virtual machines.

akohlmey:
> I think this behavior is excellent and is more clear to python users than make install-python.

"make install-python" is just a convenience wrapper. All it does is just run:
```
/path/to/my/python <LAMMPSHOME>/python/install.py -l <LAMMPSBUILD>/liblammps.so -p <LAMMPSHOME>/python/lammps
```
The make command just uses the existing knowledge about location of sources and binaries.

akohlmey:
> I think this behavior is excellent and is more clear to python users than make install-python

Come to think of it. I should add a '-n' (--noinstall) flag to the install.py script so it does not attempt to install the module.
That way one can just build the wheel and install it later. Won't change the behavior of `make install-python` though.

charlessievers:
`$ make install-python`
```
Scanning dependencies of target install-python
Installing LAMMPS Python module
Purging existing wheels...
Failed to create a virtualenv: /Users/casieve/anaconda3/envs/stubs/bin/python3.8: No module named virtualenv

make[3]: *** [CMakeFiles/install-python] Error 1
make[2]: *** [CMakeFiles/install-python.dir/all] Error 2
make[1]: *** [CMakeFiles/install-python.dir/rule] Error 2
make: *** [install-python] Error 2
```
>Come to think of it. I should add a '-n' (--noinstall) flag to the install.py script so it does not attempt to install the module.
That way one can just build the wheel and install it later. Won't change the behavior of make install-python though.

That is a good idea, also a try block that checks for no module name virtualenv and defaults to noinstall where you just end up with the wheel.

akohlmey:
> That is a good idea, also a try block that checks for no module name virtualenv and defaults to noinstall where you just end up with the wheel.

Not possible. The virtualenv is needed to build the wheel. I had initially used pip to install it into a user's local site-packages tree, but then decided against it. The purpose of using a virtualenv is to update/install pip, setuptools, build, and wheel to have
recent enough versions so the process works even on older operating systems. But then first installing virtualenv would just negate that consideration and I could just update all.

charlessievers:
> Not possible. The virtualenv is needed to build the wheel. I had initially used pip to install it into a user's local site-packages tree, but then decided against it. The purpose of using a virtualenv is to update/install pip, setuptools, build, and wheel to have recent enough versions so the process works even on older operating systems. But then first installing virtualenv would just negate that consideration and I could just update all.

I see, just read through `install.py`

akohlmey:
> I see, just read through `install.py`

Don't look too closely. It is **ugy** code. Particular the part where I load the virtual environment from python. which is why I need to run the wheel build script in a subprocess. but doing it in all-python helped to keep it portable...

charlessievers:
Maybe this is a dumb question, but what is the reason for not using a setup.py?

akohlmey:
> Maybe this is a dumb question, but what is the reason for not using a setup.py?

I **am** using it. But there is a whole lot of dark magic needed to make setuptools do something that is against their expectations. Also I wanted it to remain mostly backward compatible. Hence the wrapper script.

You can just do:
```
python -m build --wheel
```
However that will build a python-only wheel, without the shared library.
But the purpose of the whole exercise is to be able to bundle the shared library and thus make it build a binary wheel with the shared library included. But setuptools has no out-of-the-box provisions for a scenario where the compilation is already done. That is when I had to get down and dirty.

akohlmey:
@charlessievers now it should error out if virtualenv is not installed.

charlessievers:
Ah, thanks for the explanation. That makes sense.

Mind if I add in a flag to allow users to use the current python environment to build the wheel instead of creating a virtualenv?

akohlmey:
> Mind if I add in a flag to allow users to use the current python environment to build the wheel instead of creating a virtualenv?

Why? I don't see a reason why you want to do that.

charlessievers:
> Why? I don't see a reason why you want to do that.

Maybe I am a bit neurotic when it come to managing my python environments. Always having to delete files out of my ~/.local is something I would like to avoid.

akohlmey:
> Maybe I am a bit neurotic when it come to managing my python environments. Always having to delete files out of my ~/.local is something I would like to avoid.

But to avoid this *is* the reason for using a virtual environment. So you are contradicting yourself.

The steps are:
- create a virtual environment "buildwheel", i.e a folder buildwheel
- activate it and then use pip to update/install some required packages. this way the updates/installs are somewhere inside the buildwheel tree
- build the wheel with `python -m build --wheel`
- delete the "buildwheel" folder and everything that is in it
- use pip to install the previously created wheel file (the main step) into either the system site-packages or failing that the user site-packages folder
- move the generated wheel file to the build folder

If you don't want the installation part, you have to manually execute the install.py file with the required arguments (it will deduce the python version, and LAMMPS version from that) and --noinstall

charlessievers:
@akohlmey I see, that is clever. I just checked over all of the script and my problem is just that I still had a distutil lammps in my environment, so it defaulted to putting the build in my .local

```
/Users/casieve/anaconda3/envs/stubs/bin/python3.8 -m pip install --force-reinstall lammps-2022.2.17-cp38-cp38-macosx_10_15_x86_64.whl
Processing ./lammps-2022.2.17-cp38-cp38-macosx_10_15_x86_64.whl
Installing collected packages: lammps
  Attempting uninstall: lammps
    Found existing installation: lammps 2022.2.17
ERROR: Cannot uninstall 'lammps'. It is a distutils installed project and thus we cannot accurately determine which files belong to it which would lead to only a partial uninstall.
```

akohlmey:
> @akohlmey I see, that is clever. I just checked over all of the script and my problem is just that I still had a distutil lammps in my environment, so it defaulted to putting the build in my .local
> 
> ```
> /Users/casieve/anaconda3/envs/stubs/bin/python3.8 -m pip install --force-reinstall lammps-2022.2.17-cp38-cp38-macosx_10_15_x86_64.whl
> Processing ./lammps-2022.2.17-cp38-cp38-macosx_10_15_x86_64.whl
> Installing collected packages: lammps
>   Attempting uninstall: lammps
>     Found existing installation: lammps 2022.2.17
> ERROR: Cannot uninstall 'lammps'. It is a distutils installed project and thus we cannot accurately determine which files belong to it which would lead to only a partial uninstall.
> ```

Sorry, you still don't get it. The whole point of "make install-python" is to install the LAMMPS module. It *should* go into your ~/.local folder. If there is a LAMMPS module already present, then it will try to update it. But in your case, it cannot be updated, because your existing module was installed with the old script and that was using distutils and not from a wheel. That way there is no record of which files need to be removed and pip refuses to update.

charlessievers:
I think there is a miscommunication now. I understand that "If there is a LAMMPS module already present, then it will try to update it.", which is why I deleted my previous distutils version and reran `make install-python`. Anyway, it does not really matter.

> It should go into your ~/.local folder. 

The first try block defaults to what I thought was the intended behavior, updating the current environments site-package. In my case this adds it to my current anaconda environment, which is the behavior I expected. The second try block adds the module to .local, which is a nice back up behavior.

```
print("Installing wheel")
for wheel in glob.glob('lammps-*.whl'):
  try:
    print(sys.executable, '-m', 'pip', 'install', '--force-reinstall', wheel)
    txt = subprocess.check_output([sys.executable, '-m', 'pip', 'install', '--force-reinstall', wheel], stderr=subprocess.STDOUT, shell=False)
    print(txt.decode('UTF-8'))
    continue
  except:
    pass
  try:
    print('Installing wheel into standard site-packages folder failed. Trying user folder now')
    txt = subprocess.check_output([sys.executable, '-m', 'pip', 'install', '--user', '--force-reinstall', wheel], stderr=subprocess.STDOUT, shell=False)
    print(txt.decode('UTF-8'))
  except:
    sys.exit('Failed to install wheel ' + wheel)
  shutil.copy(wheel, olddir)
  os.remove(wheel)
```

akohlmey:
> Anyway, it does not really matter.

Actually, it **does** matter. A lot of the people using "make install-python" may have installed it the same way previously and expect that it will just work, but in the specific case of having an anaconda maintained environment (or using sudo and having it installed systemwide) with the old installation, then the new script should **not** fall back to enforcing a --user install.
Instead it should stop and tell the user to do a manual uninstall (just as you did without being recommended to do so).

athomps (approved):
I approve.

sjplimp (approved):

PR #3155, opened 2022-02-25
Collected small changes

akohlmey:
**Summary**

This pull request contains collected small changes and fixes

**Related Issue(s)**

Addresses the problem reported here: https://matsci.org/t/error-building-mpi-shared-lib-liblammps-mpi-so-undefined-reference-to-lammps-ns-connect-param-counts/40793

Addresses the issues reported here: https://matsci.org/t/segmentation-fault-every-time-i-try-to-run-intel-acceleration/40811

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

- update the detection of `cythonize` for the GNU make build system to be compatible with non-bash Bourne shells (like dash)
- be more careful when updating the settings for the ML-IAP package with the GNU make build system and always delete and restore settings to correctly handle detecting cythonize which also depends on the PYTHON package being installed
- let the `rerun` command bypass the time-dependent fix check when it resets the timestep; otherwise `rerun` would be rather useless for most analysis uses.
- use consistent naming scheme in force constant caching of the INTEL package. small refactor to avoid memory leaks.
- fix off-by-one bug in dihedral style fourier/intel
- fix bug in pair style lj/long/coul/long that wasn't triggering an intel style neighbor list which is required to have force field buffers allocated
- add support for recently released plumed versions 2.6.6, 2.7.4, and 2.8.0 (default is now 2.7.4)
- some refactoring to use newer and more compact APIs

**Post Submission Checklist**

### DIFF
--- a/cmake/Modules/Packages/PLUMED.cmake
+++ b/cmake/Modules/Packages/PLUMED.cmake
@@ -54,8 +54,8 @@ if(DOWNLOAD_PLUMED)
     set(PLUMED_BUILD_BYPRODUCTS "<INSTALL_DIR>/lib/libplumedWrapper.a")
   endif()
 
-  set(PLUMED_URL "https://github.com/plumed/plumed2/releases/download/v2.7.3/plumed-src-2.7.3.tgz" CACHE STRING "URL for PLUMED tarball")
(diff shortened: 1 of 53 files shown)


athomps (approved):
I approve.

sjplimp (approved):

PR #3156, opened 2022-02-25
Fix rare bug in Kokkos with CUDA UVM

stanmoore1:
**Summary**

Fix rare bug in Kokkos with CUDA UVM when some forces are computed on the host CPU--the forces were double counted when using UVM for this case.

**Related Issue(s)**

None

**Author(s)**

Stan Moore (SNL), reported by Matt Bettencourt (NVIDIA)

**Licensing**

**Backward Compatibility**

Yes


athomps (approved):
I approve.

sjplimp (approved):

PR #3147, opened 2022-02-21
Improvements from AMD to KOKKOS Package

stanmoore1:
**Summary**

Several performance improvements to Kokkos versions of `pair tersoff`, `fix qeq/reaxff`, and `pair reaxff`.

 - Reduced math overhead: enabled specialized calls (e.g., `cbrt` for a cube root instead of `pow`) and use power/exponential laws to reduce the number of exponentials evaluated, etc.

Pair Tersoff:
- Fused the `jj` loop in `TagPairTersoffComputeHalf` between the repulsive and attractive iterations
- Merged `ters_fc_k` with `ters_dfc`, `ters_bij_k` with `ters_dbij`, `ters_gijk` with `ters_dgijk`, and `ters_fa_k` with `ters_dfa`

Pair ReaxFF:
- Added blocking to the Torsion and (optionally) BuildLists kernels, to reduce thread divergence on GPUs

Fix QEq:
- Fused the CG solve for "s" and "t" vectors
- Improved the SpMV algorithm by using vector instead of team level parallelism on GPUs

**Related Issue(s)**

None.

**Author(s)**

Nicholas Curtis (AMD), Leopold Grinberd (AMD), Sriranjani Sitaraman (AMD)

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

Yes

**Implementation Notes**

**Post Submission Checklist**

- [ ] The feature or features in this pull request is complete
- [ ] Licensing information is complete
- [ ] Corresponding author information is complete
- [ ] The source code follows the LAMMPS formatting guidelines
- [ ] Suitable new documentation files and/or updates to the existing docs are included
- [ ] The added/updated documentation is integrated and tested with the documentation build system
- [ ] The feature has been verified to work with the conventional build system
- [ ] The feature has been verified to work with the CMake based build system
- [ ] Suitable tests have been added to the unittest tree.
- [ ] A package specific README file has been included or updated
- [ ] One or more example input decks are included

**Further Information, Files, and Links**

PR #3158, opened 2022-03-01
Split buildlists half kernel into "preview" and "finalize" kernels

arghdos:
**Summary**

Based on the 'preview' concept for ReaxFF Torsion evaluation introduced in #3147, this PR splits the `BuildListsHalf` kernel into two separate kernels.  The goal is to make the 'preview' kernel, which may be recalled repeatedly in the resize loop as short and quick as possible.  Then, the 'finalize' kernel evaluates the initial Bond Order values (e.g., `d_BO` and friends) in a "dense" manner, i.e., it no longer must do cutoff / bond-owner checks as that happens in the 'preview'.  

This yields a nice speedup from my testing:
  - ~11% on a single "GPU" of an MI250 (16^3 = 1245184 atoms)
  - ~20% on a very large case: 8 "GPUs" of an MI250 (22522752 atoms)

I haven't tested the impact on the Host or on NV devices, but I suspect it should help there as well

**Related Issue(s)**

N/A

**Author(s)**

Nicholas Curtis (AMD)

**Licensing**

**Backward Compatibility**

Yes

**Implementation Notes**

Validated against a `16^3` input case for both "list/blocking yes" and "list/blocking no" in Kokkos/HIP.

**Post Submission Checklist**

### DIFF
--- a/doc/src/pair_reaxff.rst
+++ b/doc/src/pair_reaxff.rst
@@ -163,9 +163,8 @@ The keyword *list/blocking* is only supported by the Kokkos version of
 ReaxFF and ignored otherwise. Setting the value to *yes* enables the
 "blocking" scheme (dynamically building interaction lists) for the
 ReaxFF bond neighbor list. This reduces the number of empty
-interactions and can improve performance in some cases (e.g. large
-number of atoms/GPU on AMD hardware). It is also enabled by default
-when running the CPU with Kokkos.
+interactions but can reduce performance in some cases (e.g. small
+number of atoms/GPU on AMD hardware).
 
 The thermo variable *evdwl* stores the sum of all the ReaxFF potential
 energy contributions, with the exception of the Coulombic and charge
@@ -370,7 +369,7 @@ Default
 """""""
 
 The keyword defaults are checkqeq = yes, enobonds = yes, lgvdw = no,
-safezone = 1.2, mincap = 50, minhbonds = 25, list/blocking = yes on CPU, no on GPU.
+safezone = 1.2, mincap = 50, minhbonds = 25, list/blocking = yes.
 
 ----------
 
--- a/doc/src/pair_reaxff.rst
+++ b/doc/src/pair_reaxff.rst
@@ -163,8 +163,9 @@ The keyword *list/blocking* is only supported by the Kokkos version of
 ReaxFF and ignored otherwise. Setting the value to *yes* enables the
 "blocking" scheme (dynamically building interaction lists) for the
 ReaxFF bond neighbor list. This reduces the number of empty
-interactions but can reduce performance in some cases (e.g. small
-number of atoms/GPU on AMD hardware).
+interactions and can improve performance in some cases (e.g. large
+number of atoms/GPU on AMD hardware). It is also enabled by default
+when running the CPU with Kokkos.
 
 The thermo variable *evdwl* stores the sum of all the ReaxFF potential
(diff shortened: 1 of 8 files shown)


stanmoore1:
Thanks @arghdos! Will take a look.

weinbe2:
Performance on A100-80GB looks good---smaller than the reported speed-ups on MI250X, but a perf boost is a perf boost! All of my runs below were just 300 timesteps, which is enough to amortize most Amdahl effects. All runs below are with `list/blocking yes`.

16^3 on 1xA100-80GB goes from 3.92 million atom timesteps / sec to 4.13 mat/s, a 3.6% speed-up.

The "very large case" is presumably 42^3 --- it looks like that problem size _requires_ running on 8 MPI ranks, otherwise this allocation fails: https://github.com/lammps/lammps/blob/develop/src/KOKKOS/fix_qeq_reaxff_kokkos.cpp#L340

Where presumably the issue is `m_cap` overflows (sufficiently confirmed by the error message that included "Allocation of size 1.718e+10 G failed").

So, alternatively, we can run a smaller problem size (33^3) to fit on the same number of physical cards, 4xA100-80GB, or we can run the same problem size (42^3) on 8xA100-80GB, matching the number of MPI ranks as what's possible when targeting 4x physical MI250. These two approaches give (again in millions of atom timesteps / second):

* 42^3 on 8xA100-80GB: 29.8 mat/s -> 30.9 mat/s, which (when you carry around all the digits) gives a ~3.5% speed-up
* 33^3 on 4xA100-80GB (10924848 atoms, roughly half): 29.7 mat/s -> 30.7 mat/s, or also a ~3.5% speed-up

Speed-ups with 1xGPU are roughly consistent, so it appears a local volume of ~16^3 repetitions is big enough to dominate scaling (in a weak scaling sense).

weinbe2 (approved):
Approved, this is a good, clean optimization.

stanmoore1:
I see a slowdown for CPU, will probably need to bifurcate code paths to keep the old version for CPU.

stanmoore1:
With these changes, I also see a small speedup for `list/blocking yes` on V100, so I think that option can be removed from the code.

stanmoore1:
Unfortunately I still see a slowdown for `list/blocking yes` on AMD for small problems. I think I may make that the default though, since it is faster on NVIDIA and CPU and AMD for large problems.

stanmoore1:
I'm running this PR through Kokkos regression testing.

stanmoore1:
Regression tests passed.

stanmoore1 (approved):

stanmoore1:
Actually I still see a slowdown for V100 and `list/blocking yes`.

stanmoore1:
@akohlmey do you see any reason why `fix_qeq_reaxff` would need ghosts in the neigh list? I think it was just copied/pasted over from the pair style.

athomps:
> @akohlmey do you see any reason why `fix_qeq_reaxff` would need ghosts in the neigh list? I think it was just copied/pasted over from the pair style.
Yes, it was probably inherited requirement from ReaxFF.  I see no reason. If it passes testing without, it is good.

stanmoore1:
Thanks @athomps!

stanmoore1:
I'm running this PR through testing again.

stanmoore1:
All the regression tests passed again. I also checked the per-atom energy and virial by hand. This PR is ready to merge from my point of view.

PR #3162, opened 2022-03-04
Optimized memory usage for pair style mesocnt

phankl:
**Summary**

Memory management of the mesocnt pair style was optimized. Now, memory is exactly calculated for the locally saved neighbor lists of nanotube chain segments. To support the memory management, 3D ragged array management was implemented. Also the potential file reader was modernized to use the LAMMPS potential file reader and tokenizer classes.

Additionally, a division by zero bug was fixed for co-planar tubes and an outdated paragraph from the documentation was removed.

**Related Issue(s)**

**Author(s)**

Philipp Kloza (U Cambridge)

**Licensing**

**Backward Compatibility**

All changes are fully backwards compatible.

**Implementation Notes**

3D ragged arrays were implemented in the same style as existing 2D ragged arrays, just with an additional layer. Memory management was optimized by splitting up the creation of local neighbor lists for chain segments into several passes where each pass calculates another layer of required memory. 

**Post Submission Checklist**


akohlmey:
@phankl thanks for your submission. I have one favor to ask: can you try deleting the "// clang-format off" line at the beginning of the pair_mesocnt.cpp file and then run `clang-format -i ` on this file (and the header) and then check if this is acceptable?
We are trying to convert the entire code to use clang-format to enforce some consistent formatting style. It works most of the time, but not for everything and the results then look much worse than without.
If clang-format only messes up some blocks with long lines you may also just bracket them with "// clang-format off" / "// clang-format on". See `src/ASPHERE/pair_resquared.cpp` for an example. Thanks in advance.

phankl:
@akohlmey thanks for the quick reply. I ran clang-format and both files look fine to me.

akohlmey:
> @akohlmey thanks for the quick reply. I ran clang-format and both files look fine to me.

Excellent. Thank you. It is always a bit awkward to do this with other people's code (I do it anyway if the changes are not too drastic). BTW: in the `Pair::allocate()` function there are often allocations of "ntypes+1" or "n+1" that become harder to read after clang-format. using a `const int np1 = atom->ntypes + 1;` and then np1 instead of n+1 can improve that. Check out for example: src/PERI/pair_peri.cpp or src/pair_table.cpp.

akohlmey:
@sjplimp please check out the change to `src/memory.h` in particular.

akohlmey:
@phankl FYI I have updated the reader/parser code in your pair style to use the bundled specialized classes. Please see https://docs.lammps.org/Developer_notes.html#reading-and-parsing-of-text-and-text-files and https://docs.lammps.org/Developer_utils.html#file-reader-classes for more info.

sjplimp (approved):

PR #3165, opened 2022-03-04
Computes for local pressure tensor in Cartesian and spherical coordinates

olavgal:
**Summary**

These computes the local diagonal components of the pressure tensor in Cartesian and spherical coordinates. This can be used to calculate the local pressure of flat and spherical interfaces.

**Author**
Olav Galteland, PoreLab, Department of Chemistry, Norwegian University of Science and Technology, olav.galteland@ntnu.no, olav.galteland@gmail.com

**Licensing**

**Backward Compatibility**

None that I am aware of

**Implementation Notes**

This implementation follows the article by Ikeshoji, Hafskjold and Furuholt closely with the Irving-Kirkwood contour [1] . I have taken the general structure of the code compute pressure/cylindrical by Addington, and adapted it to these coordinates.

I have tested this code for the following system
1. Flat liquid-vapor interface (compute pressure/cartesian)
2. Flat solid-fluid interface (compute pressure/cartesian)
3. Spherical droplet (compute pressure/spherical)
4. Spherical solid surrounded by fluid (compute pressure/spherical)

I have checked that momentum balance is obeyed. For the flat interfaces I have checked that the normal component is constant through the interfaces, and for the spherical interfaces I have checked that the relation P_T = P_N + r/2 dP_N/dr holds. I have ONLY used Lennard-Jones pair potentials. I do not know if this is compute works for more complicated potentials.

**Post Submission Checklist**

[1]  Ikeshoji, Hafskjold, Furuholt, Mol Sim, 29, 101-109, (2003).


doc/src/compute_pressure_cartesian.rst
```diff
@@ -0,0 +1,63 @@
+.. index:: compute pressure/cartesian
```
akohlmey, line 1: These files need to be fully integrated into the manual build by listing them in the `Commands_compute.rst` and the `compute.rst` pages. Also, src/.gitignore needs to be updated.

doc/src/compute_pressure_cartesian.rst
```diff
@@ -0,0 +1,63 @@
...
+:doc:`compute stress/atom <compute_stress_atom>`, :doc:`compute pressure/cylinder <compute_pressure_cylinder>`, :doc:`compute pressure/spherical <compute_pressure_spherical>`
+
+Default
+"""""""
+
+none
+
+----------
+
+.. _Ikeshoji2:
```
akohlmey, line 61: You must not have the same reference target twice across the entire manual.

akohlmey (changes requested):
Thanks for your contribution. I have added some comments with suggestions for improvement from just a cursory look and common things we look for in contributions.

athomps (commented):
Thanks for the nice additions.  The two new computes, `pressure/cartesian` and `pressure/spherical` seem to have a lot in common with the existing compute `pressure/cylindrical`.  It would be very helpful for users of LAMMPS, as well as maintainers of LAMMPS, if these three computes could be harmonized, so that they all work in the same way by default, and then any differences can be handled by additional optional keywords.  I would even go so far as to say that they should share a common doc page. I know this requires you to do some additional work, but it will be well worth it.  One obvious difference is the treatment of the kinetic contribution to the pressure. Either all three computes should include it by default, or they should all exclude it.  The opposite case can be handled by an optional keyword.

akohlmey:
> I would even go so far as to say that they should share a common doc page. I know this requires you to do some additional work, but it will be well worth it.

I fully support the suggestion of merging the documentation. I was tempted to make it myself right away but wanted to wait until other comments about the implementations were in. I had focused on simple technical changes thus far.

olavgal:
Thank you for all the comments! I will make the suggested changes. 

> I would even go so far as to say that they should share a common doc page. I know this requires you to do some additional work, but it will be well worth it.

I agree as well. I will combine the documentation for the three computes.

> One obvious difference is the treatment of the kinetic contribution to the pressure. Either all three computes should include it by default, or they should all exclude it. The opposite case can be handled by an optional keyword.

I think the simplest solution would be to add the kinetic calculation to the cylindrical coordinates. This should be straightforward, and I will give it a go if you agree that this would be a good solution.

athomps:
Yes, I agree this is a good solution. Please also add an option to omit the kinetic contribution.  This will allow the original authors of pressure/cyclindrical to recover their version without having to modify the new source code. 
> I think the simplest solution would be to add the kinetic calculation to the cylindrical coordinates. This should be straightforward, and I will give it a go if you agree that this would be a good solution.

olavgal:
I have pushed the commits that address your comments, and I have added the kinetic contribution to the pressure/cylinder, and merged the documentation pages.

src/EXTRA-COMPUTE/compute_pressure_cartesian.cpp
```diff
@@ -0,0 +1,537 @@
...
+    xi += (domain->boxhi[dir1]-domain->boxlo[dir1]); 
+  else if (xi > domain->boxhi[dir1])
+    xi -= (domain->boxhi[dir1]-domain->boxlo[dir1]); 
+  if (xj < domain->boxlo[dir1])
+    xj += (domain->boxhi[dir1]-domain->boxlo[dir1]); 
+  else if (xj > domain->boxhi[dir1])
+    xj -= (domain->boxhi[dir1]-domain->boxlo[dir1]); 
+
+  // Integrating contour from bin_s to bin_e
+  bin_s = lround((int)((xi - domain->boxlo[dir1])/bin_width1)%nbins1);
```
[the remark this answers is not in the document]
akohlmey: These were updated but still  not quite correct. The modulus should be taking **after** converting to integer. I've updated the code accordingly.

akohlmey:
@olavgal I have made some updates and corrections. There were some problems with stability, unwanted early aborts, and undetected floating point exceptions.

- you must **never** use an identity comparison for floating point to terminate a loop. floating point math is inexact (e.g. 0.1 + 0.2 != 0.3 if you look very, very closely)
- your checks for out of bounds data are "strict" so that it can become extremely difficult to provide the correct bin width. I have updated the code to compute the exact binwidth after determining the number of bins. it will print a warning if the change is larger than 1.0e-10. this applies to both new compute styles.
- the compute pressure/spherical would trigger floating-point exceptions (they are disabled on x86 by default, but not in my development builds) though either taking the square root of a negative number or a division by zero. I have added a check and tried to infer how to handle this.
- for compute pressure/cartesian, I ran into another out of box access error that I could not find a solution for.

I am using a very crude test example by adding the following lines to the peptide example input:

```
compute  1 all pressure/cartesian x 0.5
compute  2 all pressure/cartesian y 0.5 z 0.5
compute  3 all pressure/spherical 50.0 50.0 40.0 0.2 10.0

thermo_style custom step pe ke temp c_1[1][1] c_1[1][2] c_1[2][1] c_1[2][2] c_3[1][1] c_3[2][2] c_2[1][1]
thermo 10
```

This stops with: `ERROR: Bin outside: lb=1 (src/EXTRA-COMPUTE/compute_pressure_cartesian.cpp:516)`

Please see if you can resolve this and also double check that my changes didn't break anything.

Two more issues:
- we would very much appreciate if you could provide a couple of input example files that demonstrate the use of the computes
 please note that this would be done with a small system  and with just a few steps (< 1000) so that the input can finish on a single CPU within a few second (say less than 30).
- please note that the citation reminder in the C++ code should only be used if the cited paper is directly connected to you and your research. papers describing the algorithm, but where you and your group are otherwise not involved with should only be cited in the manual.

olavgal:
Thank you for the comments and corrections, @akohlmey. As far as I can see, the changes did not break anything.

> This stops with: `ERROR: Bin outside: lb=1 (src/EXTRA-COMPUTE/compute_pressure_cartesian.cpp:516)`
> 
> Please see if you can resolve this and also double check that my changes didn't break anything.

I was able to resolve this.

> Two more issues:
> 
>     * we would very much appreciate if you could provide a couple of input example files that demonstrate the use of the computes
>       please note that this would be done with a small system  and with just a few steps (< 1000) so that the input can finish on a single CPU within a few second (say less than 30).
> 
>     * please note that the citation reminder in the C++ code should only be used if the cited paper is directly connected to you and your research. papers describing the algorithm, but where you and your group are otherwise not involved with should only be cited in the manual.

I have added example files to a new folder examples/pressure, and I have changed the citation reminders in the code to papers where I have used the relevant codes. For the *pressure/spherical* the paper is only available as preprint so far. I assume that I can update the code once it (hopefully) has been accepted for publication?

akohlmey:
> I assume that I can update the code once it (hopefully) has been accepted for publication?

Yes.

Please note, I made some more updates because your examples were missing the matching log files and when running those, they triggered more floating point exceptions due to taking a square root of a negative number or division by zero.
You should perhaps make it a habit to compile with floating point exceptions enabled when programming LAMMPS:
https://docs.lammps.org/Build_settings.html#trigger-selected-floating-point-exceptions :wink:

athomps (approved):
This looks good to me. I made some additions to log output and updated compute stress/mop.  One suggestion for doc page would be to compare and contrast with compute stress/mop.

olavgal:
> You should perhaps make it a habit to compile with floating point exceptions enabled when programming LAMMPS: https://docs.lammps.org/Build_settings.html#trigger-selected-floating-point-exceptions wink

Thanks, I will do that :sweat_smile:

>  One suggestion for doc page would be to compare and contrast with compute stress/mop.

The paper by Ikeshoji et al. [1] compares the two methods (the method here is called cgIK). MOP by Todd et al. [2] does not compute the tangential components, i.e. it only computes xx, xy, xz when the x-surface is specified. cgIK in this code only computes the diagonal components (xx, yy, zz), but it is straight forward to also compute the off-diagonal components for pressure/cartesian. For pressure/spherical, one must solve eq. 23 in [1] for the off-diagonal components. This is not done in the paper. 

cgIK and MOP gives the same results for the normal component of a flat liquid-vapor surface, but the precision of cgIK was equal or better than MOP. This is likely because MOP considers particles crossing a surface, while cgIK averages over a small volume.

[1] Ikeshoji, Hafskjold, Furuholt, Mol Sim, 29, 101-109, (2003).
[2] Todd, Evans, Daivis, Phys. Rev. E 52, 1627 (1995).

akohlmey (approved):

sjplimp (approved):

athomps:
> cgIK and MOP gives the same results for the normal component of a flat liquid-vapor surface, but the precision of cgIK was equal or better than MOP. This is likely because MOP considers particles crossing a surface, while cgIK averages over a small volume.
> 

Thanks! I will add clarifying sentence to both doc pages, then it is ready to merge.

athomps:
After toggling back and forth between compute_stress_mop.rst and compute_pressure_cartesian.rst, I have one final suggestion. Rename the commands stress/cartesian, stress/spherical, stress/cylinder.  It will be consistent with stress/mop and also stress/atom.  It will also be less confusing for users and maintainers. If there are no objections, I can do it.

sjplimp:
I like that suggestion

olavgal:
I like that suggestion as well

athomps:
I switch names from pressure to stress. If I did not miss anything, this is now done. Thanks for the contribution @olavgal

akohlmey:
> I switch names from pressure to stress. If I did not miss anything, this is now done. Thanks for the contribution @olavgal

@athomps when renaming files in a package, we also need to update `src/Purge.list` for the "significant minority" that still uses the GNU build procedure. :wink:
I also made the class name for compute stress/cylinder fully consistent with our conventions and added backward compatibility for that style, so old inputs will keep working.

PR #3166, opened 2022-03-04
Make it easier to enable tabulation in ReaxFF

stanmoore1:
**Summary**

Tabulation in ReaxFF for the VdW and Coulomb forces gives a large speedup on CPUs, and a moderate speedup on some GPUs. This PR makes it easier to enable tabulation in ReaxFF by adding an option to the LAMMPS input script instead of having to create a separate `control` file.

**Related Issue(s)**

None.

**Author(s)**

Stan Moore (SNL)

**Licensing**

**Backward Compatibility**

Yes


athomps (approved):
I approve.

akohlmey:
@athomps do you think we should convert the remaining control settings (there cannot be many more after we removed the output file settings) to flags in a similar fashion and then eventually phase out the control file?

athomps:
> @athomps do you think we should convert the remaining control settings (there cannot be many more after we removed the output file settings) to flags in a similar fashion and then eventually phase out the control file?
Yes, getting rid of the control file entirely would be a good thing. Any new ReaxFF options that emerge could be added directly to the pair_style command.

PR #3167, opened 2022-03-04
Update latboltz package to new version

CDenniston:
This is a significant update of the lb/fluid package.  This was documented in a paper which can be accessed at https://doi.org/10.1016/j.cpc.2022.108318


akohlmey:
@CDenniston thanks for your submission. I just had a quick look at your contribution and noticed quite a few issues.
- you have a merge conflict with the development tree. you need to merge with the latest version of the develop branch from upstream
- your pull request reverts quite a few (minor) improvements and formal items from upstream. Please also note the requirements for headers to be compatible with running clang-format and consider using clang-format with the provided config file for your .cpp files. 
- it needs to follow new conventions for virtual functions that were implemented after the stable release
- there are opportunities to use (better) new APIs, e.g. to look up fixes by style or by ID
- there is a mechanism for flagging disabled/removed commands so that users get a suitable error message. please check out `src/fix_deprecated.cpp` and `src/fix_deprecated.h`
- there must be no trajectory or other generated files in the examples folders and the corresponding commands commented out in the input;  except for log files created with the -log flag and following the convention of other examples.

Most of the currently applied style conventions and requirements for submissions are explained here https://docs.lammps.org/Modify_style.html
I can assist with implementing them when I have more free time next week, but I just wanted to notify you right away.
If desired, I can also go through the entire set of changes and flag individual items requiring your attention.

CDenniston:
Ok, I will go through these and get back to you in a day or two.  I have never used clang before so I am not entirely sure how to do that but the others should be straightforward.

akohlmey:
> Ok, I will go through these and get back to you in a day or two. I have never used clang before so I am not entirely sure how to do that but the others should be straightforward.

About clang-format please see: https://docs.lammps.org/Build_development.html#clang-format-support
Please also note that currently using clang-format is not a strict requirement, but we are asking all contributors whether they are willing to adopt it so that eventually we can run clang-format automatically when merging changes to LAMMPS and thus have a consistent formatting throughout. Please also note that clang-format can sometimes mess things up or make code less readable than it was before. In those cases, one can place `// clang-format off/on` comment blocks around those statements.
Using clang-format is a convenient way to reformat the sources to be consistent with the rest of LAMMPS.

CDenniston:
@akohlmey , when you say "following the convention of the other examples", does that include the naming convention with input files starting with "in." rather then ending with ".in" and similar for data files?  Personally, I find that convention incredibly annoying as the system identifies endings with filetypes, not prefixes so every time I want to open a new example file I have to tell the system what app to use to open it (usually emacs in my case).  Is there an argument for using prefixes rather than suffixes (which would be standard in almost anything else)?

akohlmey:
> does that include the naming convention with input files starting with "in." rather then ending with ".in" and similar for data files
Yes.

> Is there an argument for using prefixes rather than suffixes (which would be standard in almost anything else)?

That is a discussion you have to have with @sjplimp.

I am personally not affected since I mainly work from the command line on a Linux system where the choice of command is always mine. For file extension association using .in isn't a good choice either. For the Windows packages and the Windows Store version of LAMMPS, where people are dependent on file extension, I have renamed all bundled examples to have .lmp as extension (to stick with the 3 letter DOS scheme, and to evoke LAMMPS (since the executable is called `lmp`). But I can do this because @sjplimp does not consider packaging as part of LAMMPS, but a personal effort.

But apart from that, when contributing to a software package (and I have done so for several different ones), I always try to adapt to the conventions (programming style, file names, etc.), even if it is causing discomfort, as I know how much it helps the people maintaining the software (my .emacs file for example switches between indentention style based on path names for different projects). On the other hand, for my own software I follow what my personal preferences are (e.g. 4 space indentation instead of 2 space).

That said, with good arguments, there is room for change. For example the convention of max 72 character wide lines has been dropped in favor of 100 characters when we developed the clang-format configuration (as it made look code look better on average, and we are confident to assume that nobody will be forced these days to work on a real VT100 terminal anymore).

sjplimp:
I like in.*, data.*, log.*, etc b/c when you have a dir with lots of input scripts, data files, etc, you can do
"ls" and see them grouped together.  I.e. they auto-alphabetize.

CDenniston:
It does lead to a vast number of filename extensions defined in a user's system unless they open using the command line (once you tell the system which app to identify with the file, that defines a new extension).   However, it is your call so I will just switch my extensions with prefixes.

What about whether there is a preference for log files in the examples or not?

sjplimp:
So long as the log files are small, yes include them.  If you look in the lowercase examples dirs, then all
have log files with a standard naming convention for each input script, with a data stamp, compiler, and proc count (often 1 proc or 4).  They are just meant to be reference files of a correct run for the user.

akohlmey:
> It does lead to a vast number of filename extensions defined in a user's system unless they open using the command line (once you tell the system which app to identify with the file, that defines a new extension). 

That is a fundamentally broken system anyway whatever conventions you stick to.

> What about whether there is a preference for log files in the examples or not?

The (oldest) example inputs like `examples/melt`, `examples/micelle`, or `examples/peptide` are the canonical templates. If the input file is `in.melt`, there there are two log files from from a serial run (log.<date>.melt.<compiler>.1), and one from a parallel run with 4 MPI ranks )(log.<date>.melt.<compiler>.4). No "log" or "screen", or "echo" commands in the input file and then run with the corresponding -log command line flag (so the log is complete).

akohlmey:
@CDenniston one more thing. You deleted the fill-out form during the submission of the pull request. While not everything is required, we need to put back the license consent statement, the author attribution and backward compatibility notice.
Please edit that text to reflect your submission.  thanks.

CDenniston:
Ok, I believe I have updated all the things you mentioned in your original list and updated the original comment where I had inadvertently deleted some of the form.  I see that it now did not get through all the automated checks.  However, the details of the failed test was not very instructive as to what was the problem...

akohlmey:
> Ok, I believe I have updated all the things you mentioned in your original list and updated the original comment where I had inadvertently deleted some of the form. I see that it now did not get through all the automated checks. However, the details of the failed test was not very instructive as to what was the problem...

Because of the complexity of the compile tests (we test many different combinations of settings, compilers, build system, and OS containers) it takes a lot of clicking to get to the actual errors (parallel compilation makes it even more difficult).
The problem appears to be due to using iostreams for error output (see below). Using iostreams causes problems, since it avoids the screen/logfile system used in LAMMPS and thus should not be done in the first place.
Instead, you can use `error->one()` or `error->all()` or `error->warning()`, e.g. like this:
```
error->one(FLERR, " file {} could not be opened: {}", datfile, utils::get_syserror());
```
Reading and parsing of inpute/potential/parameter files is something that we are actively working on to refactor. I am still working on improving the documentation for that (as a whole, the Programmer guide is not complete and is likely never going to be, but over the last couple of years we have made significant progress). Thus please see for example: https://docs.lammps.org/Developer_notes.html#reading-and-parsing-of-text-and-text-files
https://docs.lammps.org/Developer_utils.html#file-reader-classes

```
/mnt/lammps/workspace/dev/pull_requests/compilation_tests/lammps/src/LATBOLTZ/fix_lb_fluid.cpp:3827:12: error: 'cerr' is not a member of 'std'
       std::cerr << " file " << datfile << " could not be opened " << std::endl;
            ^~~~
/mnt/lammps/workspace/dev/pull_requests/compilation_tests/lammps/src/LATBOLTZ/fix_lb_fluid.cpp:3827:12: note: suggested alternative: 'errc'
       std::cerr << " file " << datfile << " could not be opened " << std::endl;
            ^~~~
            errc
/mnt/lammps/workspace/dev/pull_requests/compilation_tests/lammps/src/LATBOLTZ/fix_lb_fluid.cpp:3857:12: error: 'cerr' is not a member of 'std'
       std::cerr << " file " << datfile << " could not be opened " << std::endl;
            ^~~~
/mnt/lammps/workspace/dev/pull_requests/compilation_tests/lammps/src/LATBOLTZ/fix_lb_fluid.cpp:3857:12: note: suggested alternative: 'errc'
       std::cerr << " file " << datfile << " could not be opened " << std::endl;
            ^~~~
            errc
```

CDenniston:
Ok, I have removed the iostream errors. 

I can't seem to get it through the windows build though (works for the others).  The issue seems to be with your windows compiler not liking constants like M_SQRT2 from the <cmath> include.  Looking online it seemed that the problem might be needing to have _USE_MATH_DEFINES defined for windows but I have added that and it still seems to be unhappy.

akohlmey:
> not liking constants like M_SQRT2 from the cmath include

yeah, I ran across this myself some time ago when trying to port LAMMPS to visual C++. And when looking carefully through the standards, those constants (including M_PI) are actually extensions and thus compilers and c-libraries are not required to provide them.

for that reason, but also for improved consistency (especially for pi, people tend to put just some version of it with arbitrary precision into their code, or use acos(-1)), we have created the "math_const.h" header file.

so instead of using those constants, you can just add:

```
#include "math_const.h"
[...]
using namespace LAMMPS_NS;
using MathConst::MY_SQRT2;
using MathConst::MY_PI;
```
and then just replace `M_SQRT2` with `MY_SQRT2` and so on.

CDenniston:
Ok, that worked.  Let me know if anything else is needed.

akohlmey:
> Ok, that worked. Let me know if anything else is needed.

Thanks for taking the time and putting in the effort. You are making our job to integrate this much easier.
I currently don't know when I will have time for a closer look. Since this is a significant change, there may be more, mostly formal/cosmetic changes or attempts to better realign with ongoing refactoring work.

If I see something, would you prefer me to add notes to the source here on github and then submit a "change request"?
Otherwise, at least for changes unrelated to the physics of your model, I could also grab your branch and apply them myself and then either push them back into your submission or submit a pull request to the feature branch in your fork.
We have been trying hard to make all existing code in LAMMPS more consistent (and after we reached some minimum threshold there, we dropped dropped the USER- prefix for packages to reflect that), but now the challenge is to not go backwards on new submissions and updates to packages. Unfortunately, this means some extra work for the people submitting pull requests, but we hope that the result (a more maintainable, portable, and more consistent software) justifies it.

Since we are just a few people, we try to "outsource" as much of this as possible, but sometimes this is reaching the limit of what people can handle and also not everything is currently properly documented or consistently implemented (even in the core code). For some (mostly cosmetic) changes (like converting from old to new APIs) it can be much more effective to simply applying the change instead of describing what we would like to see. So please let us know when you are reaching a limit or prefer to do all changes on your own.

CDenniston:
At this point I would suggest doing whatever is easiest for you.  If it is easier for you to just make a change, go ahead and do it. 
 Otherwise, I am certainly willing to make more changes myself if needed.  I am completely flexible on how you do changes so do whatever is easier for you.

akohlmey:
> I am completely flexible on how you do changes so do whatever is easier for you.

What is easiest is probably a mix. I.e. I make some changes and then also add some comments and suggestions. For the bigger policy/design choices (if needed), it is usually beneficial to include @sjplimp in the discussion. He can see solutions that I don't always see because I have my nose too close to the ground. :wink:

I am asking mostly because some people are not so comfortable with others making changes to their branches (I am one of them, I prefer to get them as pull requests to my fork unless it is by a person that I know very well and trust) or with people adding 50+ change request comments to their pull request as they consider that as disrespecting their programming skills.

akohlmey:
@CDenniston I have updated your pull request to improve its alignment with the latest LAMMPS programming conventions. In particular:
- I moved the static constants from the fix_lb_fluid.h header to its own header
- I removed debug comments
- I converted code using iostreams and stringstream to use stdio and fmtlib
- I removed the documentation for the removed fixes
- I replaced direct access to LAMMPS internal data structures with accessor functions where available
- I added a check so that only one instance of fix lb/fluid can be used (it seemed implied from the code) and also flagged it to require continuous timestepping (a recently added flag)

I have not yet looked at the examples. If possible, I will also run them with valgrind to look for memory leaks and access to uninitialized data.

src/LATBOLTZ/fix_lb_fluid.h
```diff
@@ -42,123 +40,169 @@ class FixLbFluid : public Fix {
...
+  int dump_interval, dump_time_index;
+  std::string dump_file_name_xdmf;
+  std::string dump_file_name_raw;
+  FILE *dump_file_handle_xdmf;
+  MPI_File dump_file_handle_raw;
+  MPI_Datatype dump_file_mpitype;
+
+  int groupbit_viscouslb;
 
   double ***density_lb;    // fluid density
```
akohlmey, line 97: You may want to initialize all pointer variables declared in a class to `nullptr` in the class constructor.
This helps avoiding memory leaks, uninitialized memory access, and use after free cases.
If pointers are initialized to null in the constructor, you may always call `delete[]` on them or `memory->destroy()` before (re-)allocating them.

CDenniston: All the changes, including this one, you made seem like obvious things, some of which have thought about but got distracted on before.  

On this one (initialization to nullptr) I presume you mean doing this in the header rather than explicitly in the constructor (as done now)?

akohlmey: > On this one (initialization to nullptr) I presume you mean doing this in the header rather than explicitly in the constructor (as done now)?

No. We try to keep header files lean, specifically for headers of "styles", since they *all* are included in multiple places through the auto-generated `style_fix.h` header file. That is also why I am "on the warpath" to reduce potential namespace conflicts. Those caused multiple hard to reproduce bugs in the past.
Please see https://github.com/lammps/lammps/blob/develop/src/fix_nh.cpp#L57 for an example of what I mean.
This avoids issues that are often unlikely to get noticed by just running LAMMPS, since the first allocation of classes or arrays is using freshly reserved memory on Linux (which uses a malloc() implementation that reserves new memory by applying mmap() on /dev/zero with copy-on-write semantics and thus initialized to all zeros), but if people issue a `clear` command (and everything is freed and reallocated) or have workflow runs in python using the LAMMPS module with LAMMPS instances created either subsequently or concurrently then the freed memory is going to be recycled and will no longer be initialized to zeros but with random bytes left behind from the previous use.

CDenniston: Ah, the initializer list.  Somehow it never occurs to me to use these.  I have added a commit to change this and also I realized I somehow missed updating the reference in the README file in the LATBOLTZ directory so fixed that too.

akohlmey:
@CDenniston I've plugged some memory leaks due to not releasing all custom MPI types and did some more cosmetic changes.
This looks ready for merging now.

One possible consideration could be to add OpenMP parallelism to the "requilibriumdist*" functions. But for that, the loops need to be flattened and special precautions are needed for the random number generator classes (they are not reentrant, so there would need to be one instance per thread. So I am not sure if it is worth the effort.

sjplimp (approved):

sjplimp:
@CDenniston thanks for submitting this to LAMMPS
@akohlmey thanks for insuring consistency with the rest of LAMMPS

PR #3169, opened 2022-03-11
Reduce memory use in Kokkos package

stanmoore1:
**Summary**

Reduce memory use in Kokkos package when resizing views by explicitly deallocating large views (e.g. neighborlist) first to eliminate double allocations.

**Related Issue(s)**

#3158 which removed ghost atoms, which triggered a resize and a corresponding increase in memory

**Author(s)**

Stan Moore (SNL), reported by Nick Curtis (AMD)

**Licensing**

**Backward Compatibility**

Yes

### DIFF
--- a/src/KOKKOS/fix_acks2_reaxff_kokkos.cpp
+++ b/src/KOKKOS/fix_acks2_reaxff_kokkos.cpp
@@ -449,8 +449,24 @@ void FixACKS2ReaxFFKokkos<DeviceType>::allocate_matrix()
   // determine the total space for the H matrix
 
   m_cap = 0;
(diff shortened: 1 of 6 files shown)


stanmoore1:
@arghdos @weinbe2

stanmoore1:
There is more that could be done here to reduce memory, but this gets most of the largest views (i.e. neighborlist).

weinbe2:
Looks good, thanks @stanmoore1 !

arghdos:
Same here!

stanmoore1:
Eventually I want to use `Kokkos::realloc` instead of the empty view constructor, but we need Kokkos v3.6.0 which supports passing `Kokkos::WithoutInitializing` into `realloc` first.

athomps (approved):
I approve.

PR #3179, opened 2022-03-20
Refactoring of Thermo class and new thermo style yaml

akohlmey:
**Summary**

This pull request refactors and extends the Thermo class code with multiple goals:
- simplify string handling by using C++ std::string and std::vector classes. This allows to remove most heuristics about memory allocations and reduces the chances of memory leaks.  Several class members have been removed.
- improve correctness of output format processing.
- move redundant code to support functions and improve related error messages
- make output more readable:
  * for style "multi" update format and reserve extra space so "Step" and "Cpu" fields won't overflow as easily
  * for style "one" column headers are now centered over the output columns (with default format). Integers have more space.
- new thermo style "yaml" that allows printing YAML compatible thermo output, that can be easier extracted from log files than the "one" style output. it defaults to larger precision, too.  This makes it ideal for use with regression testing.
- Update comments in source and documentation for modifying the Thermo class and adding new thermo styles and keywords.

Revised "one" style output looks like this:
```
   Step        Temp        E_pair       E_mol        TotEng       Press    
         0         1.44   -6.7733681            0   -4.6134356   -5.0197073
        10    1.1253949   -6.3000527            0   -4.6120132   -2.5596852
        20    0.6257938   -5.5565565            0   -4.6178952   0.97351766
        30    0.7459273   -5.7395128            0   -4.6206568    0.3392841
```
Revised "multi" style output looks like this:
```
------------ Step             50 ----- CPU =     10.36881 (sec) -------------
TotEng   =    -25330.0317 KinEng   =     21501.0005 Temp     =       299.8229 
PotEng   =    -46831.0321 E_bond   =      2471.7035 E_angle  =     10836.5102 
E_dihed  =      5239.6319 E_impro  =       227.1218 E_vdwl   =     -1993.2873 
E_coul   =    206797.6802 E_long   =   -270410.3926 Press    =       237.6571 
Volume   =    308031.6762
```
New "yaml" style output looks like this:
```yaml
---
keywords: [Step, Temp, E_pair, E_mol, TotEng, Press, ]
data:
  - [100, 0.757453103239935, -5.7585054860159, 0, -4.62236133677021, 0.207261053624721, ]
  - [110, 0.759322359337036, -5.7614668389562, 0, -4.62251889318624, 0.194314975399602, ]
  - [120, 0.759372342462676, -5.76149365656489, 0, -4.62247073844943, 0.191600048851267, ]
  - [130, 0.756833027516501, -5.75777334823494, 0, -4.62255928350835, 0.208792327853067, ]
...
```
It can be extracted from log files e.g. with:
```bash
egrep  '^(keywords:|data:$|---$|\.\.\.$|  - \[)' log.lammps > log.yaml
```
Or read and parsed in python with:
```python
import re, yaml

docs = ""
with open("log.lammps") as f:
    for line in f:
        m = re.search(r"^(keywords:.*$|data:$|---$|\.\.\.$|  - \[.*\]$)", line)
        if m: docs += m.group(0) + '\n'

thermo = list(yaml.load_all(docs, Loader=yaml.SafeLoader))
```
After that `thermo` is a list containing a dictionary for each "run" where the tag "keywords" maps to the list of thermo header strings and "data" has a list of lists where the outer list represents the lines of output and the inner list the values of the columns matching the header keywords for that step.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

There are some changes to the output format of log files that may affect programs that depend on specific widths and formats.
Existing unit tests pass.

**Implementation Notes**

This so far retains the use of printf()-style format strings and its functionality is - thus far - unchanged.
With fmt formatting, however, we could simplify this further so that users only provide the desired width and precision (for floating point) for keywords. That would not be backward compatible, but would simplify changing formats at the expense of losing some flexibility.

Most of the difficult to read code processing C-style strings for that has been replaced by simpler and safer code and at multiple places already fmt formatting is used. When printf() style formatting is used, buffer overflows are now avoided by using snprintf() to print into a known size buffer and appending the result to a C++ style std::string class. So the benefit to switch to a non-backward compatible change from the maintainer's perspective is less. Question is what is better for the users?

**Post Submission Checklist**


athomps (commented):
How does this affect comparison of new log.lammps files with reference output?

akohlmey:
> How does this affect comparison of new log.lammps files with reference output?

There are only whitespace differences or for thermo style multi there are fewer '-' characters.

For example, the parser in the automated regression test have already been adapted by replacing string.startswith() with re.match() in a couple of places.

akohlmey:
> How does this affect comparison of new log.lammps files with reference output?

Looking forward, the basic idea is to pre-process input files to use thermo style yaml when running them for the purpose of collecting data for comparisons, which makes extracting the raw numbers much easier. Comparison would then be done based on just the extracted YAML format data and only that would need to stored locally as reference for future comparison. A few well placed comments in the input files that can be processed with "sed" would do the trick.

sjplimp:
I like the yaml option.  For the "one" style is it as easy to make the fields left-justified as right?
I'm wondering if that would look better.

akohlmey:
> For the "one" style is it as easy to make the fields left-justified as right? I'm wondering if that would look better.

Are you talking about the header or the data?

The column headers are currently centered (one of the features that fmt formating has in addition to printf formatting).
Changing between left, center, right justified is simply using "{:<12} ", {:^12} ", or {:>12} ", respectively at:
https://github.com/akohlmey/lammps/blob/356555418dfc8ab20099c9d44a230912738d2aa1/src/thermo.cpp#L329

Also the width can be easily made dynamic with something like this: `fmt::thermo("{:^{}} ", value, width);`

Left justified for data would in my opinion make sense for the floating point format. For integer it would look odd. But integer format could benefit from an extra space on the right. After some tweaking of the defaults, how does this look?

```
   Step         Temp        E_pair       E_mol        TotEng       Press    
         0   3           -6.7733681    0           -2.2744931   -3.7033504  
        50   1.6842865   -4.8082494    0           -2.2824513    5.5666131  
       100   1.6712577   -4.7875609    0           -2.281301     5.6613913  
       150   1.6444751   -4.7471034    0           -2.2810074    5.8614211  
       200   1.6471542   -4.7509053    0           -2.2807916    5.8805431  
       250   1.6645597   -4.7774327    0           -2.2812174    5.7526089  

```

sjplimp:
a left justified timestep might look better.  I.e. easier to see since its first.  Can you post an example?  Are there any other integer fields we output with thermo?

sjplimp:
ok - let's stick with right-justified for ints as well (timestep).  I'll approve the PR.

sjplimp (approved):

PR #3181, opened 2022-03-21
Damping fixes for finite-size particles simulations

jibril-b-coulibaly:
**Summary**

This PR introduces 2 fixes to help dissipate energy in finite-size (granular) simulations:
- `fix viscous/sphere` : the angular counterpart to `fix viscous`. Adds viscous torque opposing angular velocity
- `fix damping/cundall` : non-viscous damping scheme proposed by Cundall, 1987. Effective and popular in many DEM codes/communities.

**Related Issue(s)**

none

**Author(s)**

Jibril B. Coulibaly. Northwestern University. jibril.coulibaly@gmail.com

**Licensing**

**Backward Compatibility**

No change in backward compatibility

**Implementation Notes**

- `fix viscous/sphere` derived from `fix_viscous`, replacing force `f` by torque `torque` and velocity `v` by angular velocity `omega`
- `fix damping/cundall` based on `fix viscous` , adding elements from the [Yade-DEM implementation](https://www.gitlab.com/yade-dev/trunk/-/blob/master/pkg/dem/NewtonIntegrator.cpp)

These fixes are small enough deviations from `fix viscous` that I did not add specific tests. I've been using them successfully in my granular simulations. 

**Post Submission Checklist**

[PFC Documentation](http://docs.itascacg.com/pfc600/pfc/docproject/source/manual/numerical_simulations_with_pfc/pfc_formulation/energy_and_dissipation_mechanisms.html?node1219#cundall1987)
[Yade-DEM Documentation](https://yade-dem.org/doc/formulation.html#numerical-damping)


sjplimp:
I agree with Axel on the name change for cundall

dsbolin:
@akohlmey This is a useful feature. As stated in the accompanying doc page of this PR, this scheme doesn't represent a realistic physical dissipation process, but is a useful numerical damping scheme; hence I think no further need for scientific review.

@jibril-b-coulibaly I haven't tried, but this type of damping may be possible with the existing `fix addforce`, with appropriate combinations of atom-style variables for the different components of the force. There is currently no similar option for angular damping, which your fix enables. However, one nice thing about doing it via `fix addforce` is that scaling can be based on per-atom radii, rather than per-type scaling, which can be limiting in an arbitrarily polydisperse granular system. Is there value in having per-atom radius scaling (or more generally, scaling based on an atom-style variable input) with this type of damping? Even so, I would still recommend merging this PR, and leaving the per-atom radius/atom variable scaling for a possible future extension.

dsbolin:
@jibril-b-coulibaly Not sure if this will work as intended if `fix damping/cundall` is listed before other fixes that change forces (e.g. `fix gravity`) in the input script. That's probably fine, there are other cases like this in lammps. Maybe just a comment in the doc page about having this fix declared last?

akohlmey:
> @jibril-b-coulibaly Not sure if this will work as intended if `fix damping/cundall` is listed before other fixes that change forces (e.g. `fix gravity`) in the input script. That's probably fine, there are other cases like this in lammps. Maybe just a comment in the doc page about having this fix declared last?

@dsbolin I just added support for getting per-atom scaling from an atom style variable (which is pretty straightforward). It will be trivial to add a check for fix ordering.

jibril-b-coulibaly:
Thank you very much for your feedback @dsbolin @akohlmey :

- I kept the scaling option the way it is simply because I derived the `damping/cundall` from `fix viscous`. I have not heard of anyone scaling the Cundall damping to either type or radii ; most of the cases I am familiar with damp every particle the same. Because of the numerical nature of the scheme and the fact that damping is based the force, I am not sure there would be a particular need for it, or that it would be physically relevant, but it's nice to have the option.
- The ordering is an interesting point. There might be cases when users want to damp some forces, e.g., contact forces, and add others undamped forces later on. I believe writing a warning in the Doc could be more suitable than forcing this fix to be the last, I will let you be the judges on that before updating the doc.

akohlmey:
> * I believe writing a warning in the Doc could be more suitable than forcing this fix to be the last, I will let you be the judges on that before updating the doc.

Warnings in the documentation should provide an explanation of why, but it is important to have the code print a message to the log. Most people won't look at the documentation unless there is a warning in the output (and too many ignore those as well).
I agree that enforcing the fix order would be too strong a measure in this case.

I have added the warning.

Please review and test the code to make certain I have not messed up anything. I built some trivial tests only to validate and may have missed some details.

jibril-b-coulibaly:
Thank you @akohlmey, I have reviewed and tested the code and everything looks good and works well. I've made the following changes:

- I have added a Note about the fix order in the documentation for `fix damping/cundall`.
- I have also deleted the warning for fix order in `fix viscous/sphere`. The viscous damping torque is based on the angular velocity which no other fix changes, so fix order should not matter (like for the classical `fix viscous`).

jibril-b-coulibaly:
I have found an unrelated issue with some of my simulations involving many repeated runs. I get the "Too many warnings" message
"

This line goes to the log file between the thermo headers and the first values, e.g., something like:
```
Step ke
WARNING: Too many warnings: 106 vs 100. All future warnings will be suppressed (C:\lammps\src\thermo.cpp:461)
1000 102.50
```
Post-processing with Pizza.py (`log.py`) throws an error since it looks for floats right after the header line and cannot convert "WARNING". This would happen for any simulation with a lot of warnings. I also had similar issues in the past with the `fix halt` message printing in the middle of the thermo output. Fixing it would require to either:
1. Make Pizza.py ignore any line that starts with non-floats between lines starting by "Step" and "Loop time of".
2. Do not print thermo warnings between lines starting "Step" or after "Loop time of".

Please let me know what you think and if there are alternatives. If not, the former solution seems more convenient: @sjplimp, how do Pizza.py and LAMMPS interact and where would be the best place to submit such change in `log.py` ? Thanks

akohlmey:
> Please let me know what you think and if there are alternatives. If not, the former solution seems more convenient: @sjplimp, how do Pizza.py and LAMMPS interact and where would be the best place to submit such change in `log.py` ? Thanks

Lots of things may print text to screen/logfile during a run and that may confuse any heuristics that is parsing data from the "one" or "custom" style thermo output.

In the specific case, you have three options:
- change the warning behavior with thermo_modify warn
- use fix print to a file only for collecting "undisturbed" output
- we just merged a change to the development branch that supports a new thermo style "yaml" that should be easier to extract and won't need Pizza.py to parse the log. Please see: https://docs.lammps.org/latest/thermo_style.html and https://docs.lammps.org/latest/Howto_structured_data.html#yaml-format-thermo-style-output

akohlmey:
> Lots of things may print text to screen/logfile during a run and that may confuse any heuristics that is parsing data from the "one" or "custom" style thermo output.

I should clarify, that using thermo style yaml won't change the extra output to screen and logfile, but the specific format of the yaml output makes it much less likely that there is a conflict.

akohlmey:
@sjplimp can you please check this out, too, please?

PR #3182, opened 2022-03-22
ML-IAP updates

charlessievers:
**Summary**

Added different pytorch modules that allow different handling of element types and proposal for __init__.py bug fix

**Author(s)**

Charlie Sievers  charliesievers@cox.net

**Licensing**

**Backward Compatibility**

Retained old pytorch modules, just renamed some bispectrum to descriptors and added new modules.

**Implementation Notes**

Both IgnoreElems and UnpackElems work and run, more tests need to be run with ElemwiseModels.

### DIFF
--- a/python/lammps/mliap/__init__.py
+++ b/python/lammps/mliap/__init__.py
@@ -10,6 +10,7 @@ try:
 except OSError as e:
     if pylib.endswith(".a"):
         pylib.strip(".a") + ".so"
+        pylib = ctypes.CDLL(library)
     else:
         raise e
 if not pylib.Py_IsInitialized():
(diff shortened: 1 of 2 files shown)


charlessievers:
@lubbersnick please take a look over the changes.

athomps (approved):
I approve.

akohlmey:
@charlessievers @lubbersnick this is sitting here now for two weeks without anything happening to it. Please either assign to me for merging, or indicate that this is still work in progress and in that case possibly close it now and resubmit when ready in case it will take some more weeks/months. thanks.

lubbersnick:
Based on @charlessievers comment that "more tests need to be run with ElemwiseModels" I think we should mark that as experimental or something.

Otherwise I don't see a problem. I don't have any permissions to adjust who can merge it.

charlessievers:
It can be merged.

PR #3190, opened 2022-03-28
Ignore "processors" when nprocs == 1

srtee:
**Summary**

With this PR, LAMMPS ignores `processors` commands when running on a single MPI process. This removes one hurdle to running production scripts in testing or debugging LAMMPS.

**Related Issue(s)**

**Author(s)**

Shern Tee, University of Queensland

**Licensing**

**Backward Compatibility**

No issues expected. Previous `processors` commands that would have thrown errors because they came after box definition will still throw errors.

**Implementation Notes**

Added `comm->nprocs` check to `input->processors()` before it calls `comm->set_processors`.

**Post Submission Checklist**

### DIFF
--- a/doc/src/processors.rst
+++ b/doc/src/processors.rst
@@ -77,7 +77,9 @@ particular dimension over the course of the simulation.
 
 The product of Px, Py, Pz must equal P, the total # of processors
 LAMMPS is running on.  For a :doc:`2d simulation <dimension>`, Pz must
-equal 1.
+equal 1. As an exception, if LAMMPS is running on one processor,
+this command will be ignored, although a warning will be printed to the
+log file.
 
 Note that if you run on a prime number of processors P, then a grid
 such as 1 x P x 1 will be required, which may incur extra
@@ -342,6 +344,9 @@ This command cannot be used after the simulation box is defined by a
 :doc:`read_data <read_data>` or :doc:`create_box <create_box>` command.
 It can be used before a restart file is read to change the 3d
 processor grid from what is specified in the restart file.
+
+This command will be ignored if LAMMPS is running on a single processor.
+A warning will be printed to the log file when this occurs.
 
 The *grid numa* keyword only currently works with the *map cart*
 option.
--- a/src/input.cpp
+++ b/src/input.cpp
@@ -1739,7 +1739,9 @@ void Input::processors()
 {
   if (domain->box_exist)
     error->all(FLERR,"Processors command after simulation box is defined");
-  comm->set_processors(narg,arg);
+  if (comm->nprocs == 1)
+    error->warning(FLERR,"Processors command ignored on single MPI process");
+  else comm->set_processors(narg,arg);
 }
 
 /* ---------------------------------------------------------------------- */


akohlmey:
@srtee I don't think that it is a good idea to make commands behave differently depending on other settings or properties. We already have a case with the thermo output being normalized with units "lj" and not normalized otherwise which is causing some confusion. For the most part what you are trying to achieve can be also achieved with wildcards, e.g. `processors * * 1`.

srtee:
My use case is that my scripts frequently use `processors * * 2` / `processors * * 4` (electrode-electrolyte systems), for which I have to comment out the processors line for testing or debugging on one proc and then re-enable it for production.

What about a thermo variable `nprocs`? I could then use `if` to fire off a `processors` command only if `nprocs` is more than 1.

akohlmey:
> What about a thermo variable `nprocs`? I could then use `if` to fire off a `processors` command only if `nprocs` is more than 1.

You cannot use thermo keywords before the box is created, but you need to issue the `processors` command before then.
But there are other ways to access the communicator size. One is through the environment variables that `mpirun` or `mpiexec` commands set. Check out the following:
```
variable mpich getenv PMI_SIZE
variable openmpi getenv OMPI_COMM_WORLD_SIZE
variable nprocs equal v_mpich+v_openmpi
print "${nprocs}"
```
This should print `0` when you run serial without `mpirun` and the number of processors in the world communicator for MPICH and OpenMPI based MPI libraries (which is 99% of the implementations I know). So you could use `if ${nprocs} > 1` with this.

For a cleaner and fully portable access method, I have implemented an interface to the `lammps_extract_setting()` function of the library interface into the `variable` command in PR #3191. That will allow you to do `if $(extract_setting(world_size))`

akohlmey:
> What about a thermo variable `nprocs`? I could then use `if` to fire off a `processors` command only if `nprocs` is more than 1.

You cannot use thermo keywords before the box is created, but you need to issue the `processors` command before then.
But there are other ways to access the communicator size. One is through the environment variables that `mpirun` or `mpiexec` commands set. Check out the following:
```
variable mpich getenv PMI_SIZE
variable openmpi getenv OMPI_COMM_WORLD_SIZE
variable nprocs equal v_mpich+v_openmpi
print "${nprocs}"
```
This should print `0` when you run serial without `mpirun` and the number of processors in the world communicator for MPICH and OpenMPI based MPI libraries (which is 99% of the implementations I know). So you could use `if "${nprocs} > 1"` with this.

For a cleaner and fully portable access method, I have implemented an interface to the `lammps_extract_setting()` function of the library interface into the `variable` command in PR #3191. That will allow you to do `if "$(extract_setting(world_size)) > 1"`

akohlmey:
> What about a thermo variable `nprocs`? I could then use `if` to fire off a `processors` command only if `nprocs` is more than 1.

You cannot use thermo keywords before the box is created, but you need to issue the `processors` command before then.
But there are other ways to access the communicator size. One is through the environment variables that `mpirun` or `mpiexec` commands set. Check out the following:
```
variable mpich getenv PMI_SIZE
variable openmpi getenv OMPI_COMM_WORLD_SIZE
variable nprocs equal v_mpich+v_openmpi
print "${nprocs}"
```
This should print `0` when you run serial without `mpirun` and the number of processors in the world communicator for MPICH and OpenMPI based MPI libraries (which is 99% of the implementations I know). So you could use `if "${nprocs} > 1"` with this.

For a cleaner and fully portable access method, I have implemented an interface to the `lammps_extract_setting()` function of the library interface into the `variable` command in PR #3191. That will allow you to do `if "$(extract_setting(world_size)) > 1"`

akohlmey:
> What about a thermo variable `nprocs`? I could then use `if` to fire off a `processors` command only if `nprocs` is more than 1.

You cannot use thermo keywords before the box is created, but you need to issue the `processors` command before then.
But there are other ways to access the communicator size. One is through the environment variables that `mpirun` or `mpiexec` commands set. Check out the following:
```
variable mpich getenv PMI_SIZE
variable openmpi getenv OMPI_COMM_WORLD_SIZE
variable nprocs equal v_mpich+v_openmpi
print "${nprocs}"
```
This should print `0` when you run serial without `mpirun` and the number of processors in the world communicator for MPICH and OpenMPI based MPI libraries (which is 99% of the implementations I know). So you could use `if "${nprocs} > 1"` with this.

For a cleaner and fully portable access method, I have implemented an interface to the `lammps_extract_setting()` function of the library interface into the `variable` command in PR #3191. That will allow you to do `if "$(extract_setting(world_size)) > 1"`

akohlmey:
sorry about the multiple replies. github kept telling me, that I could "not comment at this time" even after reloading the page, but it *did* store the comments.

srtee:
That sounds great! No need for this PR then.

PR #3194, opened 2022-03-30
ELECTRODE package for constant potential method

ludwig-ahrens:
**Summary**

The constant potential method (CPM) (fix electrode/conp) allows specifying electrostatic potentials for a number of electrodes. The charges of electrode atoms will be changed at every time step to match the applied potential. In a similar way the constrained charge method (fix electrode/conq) allows the user to specify total charges for electrodes and the thermo-potentiostat (fix electrode/thermo) adds dissipation and fluctuation to the electric potential. In addition to the slab correction for the KSpaces, a wire correction and the 2D-Ewald method are implemented.

**Related Issue(s)**

**Author(s)**

Ludwig Ahrens-Iwers - Hamburg University of Technology
Shern Tee - s.tee@uq.edu.au - The University of Queensland
Robert Meißner - robert.meissner@tuhh.de - Hamburg University of Technology

**Licensing**

**Backward Compatibility**

**Implementation Notes**

The charge updates are performed by one of the three new fixes, electrode/conp, electrode/conq and electrode/thermo. Two new KSpaces have been implemented that inherit from the ewald and the pppm style and add methods that are needed for the CPM. The ElectrodeMatrix and ElectrodeVector classes are called by the fixes and handle the computation of the electrode-electrode and electrode-electrolyte interaction, respectively, with each class containing its own code for the short range Coulomb interactions. A BoundaryCorrection class was also added, which encapsulates the existing LAMMPS implementation of the EW3DC method and adds new implementations for a 2D Ewald summation and a wire correction method intended for 1D-periodic systems.

Four new tests are added to the unittest directory. During development more tests have been used that are currently in the examples/electrode directory.

**Post Submission Checklist**

[Preprint on ELECTRODE package](https://arxiv.org/abs/2203.15461)
[Paper on CPM with PPPM solver](https://aip.scitation.org/doi/full/10.1063/5.0063381)


akohlmey (changes requested):
Thanks for your submission. As you may have seen, there are build errors for some setups that need to be addressed. Please let us know, if you need additional pointers for those.

I have done a very cursory scan of the changes and added some suggestions for changes. Most of those apply to multiple files, but I stopped adding them after it became repetitive.
I don't have time for more detailed review, but there is quite a bit of suggestions to process already.

Please also go to the src folder and run: "make fix-homepage; make fix-whitespace; make fix-permissions" and commit the resulting changes.

akohlmey:
FYI, compilation fails because the neighbor list requests process has been revised (and we hope simplified). Please see: https://docs.lammps.org/Developer_notes.html#requesting-and-accessing-neighbor-lists
It is probably a good idea at this point to merge in the current "develop" branch from upstream.

akohlmey:
@ludwig-ahrens @robme @srtee I've done some significant refactoring of your pull request. A few things needed changes to the core code which are included here as well. In particular the following type of changes were made:
- update to the current development branch
- fix a few memory leaks
- revised neighbor list requests to follow the new interface
- correct file installation dependency issue for GNU make build system
- updates to presets for CMake and GNU make
- correct OpenMP directives to be properly compatible with pre-OpenMP 4 compilers
- silence compiler warnings
- refactor the matrix file reader/writers to avoid iostreams and be more compact through LAMMPS own' file reader and tokenizer/converter classes
- modernize code in some places as indicated by clang-tidy. this particularly applies to inconsistent and partially incorrect polymorphism. this was already noted before, but it applied to a whole lot more cases. also some core functions needed to me flagged as virtual
- fix pointer type mismatch bugs (no idea why those were not detected before)
- make timing output optional and off by default (implemented in one case, needs to be applied to the rest and also documented). even better would be to make this a compile time option (e.g. have code only included if -DLMP_ELECTRODE_TIMING is defined, which is off by default).
- apply clang-format
- fix issues in the documentation and address spelling issues
- add tags to yaml files to skip unsupported tests

Please test and check that all functionality is still working correctly or additional changes are needed (e.g. because things came out correctly by chance and now got broken).

Also please update the examples/tests to more closely resemble LAMMPS conventions.
- input files are `in.<somename>` and not `<somename>.in`
- provide log files for running with 1 and 4 MPI ranks in the format `log.<date>.<somename>.g++.<nproc>`
- use the -in flag and not I/O redirection in your scripts
- have inputs only generate dumps or other files when needed by tests (otherwise those lines, or arguments should be commented out) and have the tests delete them after the test succeeds

All the examples/tests should be run with valgrind and checked for further memory leaks (i only checked one tests so far).

Overall this is getting close to being ready for merge, but there is still some work needed.

Please also note that some of the looping constructs related to tags are going to become performance bottlenecks for large systems. Please consider using an atom map (array or hash) and the atom->map() function instead where possible.
It may also be needed to a) make some tests for system size limitations and strong/weak parallel scaling performance and b) provide some information in the manual about possible restrictions and performance impact. People generally expect that LAMMPS can handle very large systems with good strong and weak scaling.

sjplimp:
Axel's last point is important.  If you are looping over atoms, looking for an atom ID (tag), then you have to use the atom map
functionality instead.  If an atom map is enabled, then you can just look up the index of that atom directly.  You can also check
that an atom map is enabled (by the input script if necessary), and throw an error if it is not.

srtee:
> Axel's last point is important. If you are looping over atoms, looking for an atom ID (tag), then you have to use the atom map functionality instead. If an atom map is enabled, then you can just look up the index of that atom directly. You can also check that an atom map is enabled (by the input script if necessary), and throw an error if it is not.

Yup -- I'm in the middle of refactoring the code to use an atom array with callback to track our custom IDs, which should be done within the next day or two.

akohlmey:
@ludwig-ahrens @robme @srtee can you please provide some feedback on my comments and some ETA for updates?
Please note that we planning to do a stable release in a couple of months which means there will be a period where we won't integrate changes that are not self-contained or only bug fixes and that period will start at the end of this month. Since this PR changes some core files, any review and further processing would then be delayed until after the stable release.

Please note that I expect to have at least one more iteration of review and updates until this PR would be ready for merging.

sjplimp:
@srtee What is stored in the unordered map that doesn't change and how big is it?

srtee:
> @srtee What is stored in the unordered map that doesn't change and how big is it?

It's the same `taglist` which is in the current code but changed to an unordered map, since we never search, only access. Its use pattern (in the changed code) is "for each of my local atoms -- if atom is in electrode group (via `mask[i]`) look up the electrode ID as `taglist[tag[i]]`."

Size is the number of electrode atoms (typically 10-50% of particles in simulation) and it maps tag back to our internal numbering.

For example if our simulation had electrode particles with tags 142, 291, 311, and 201, `taglist` would be the unordered_map {142 -> 0, 291 -> 1, 311 -> 2, 201 -> 3}.

We need internal numbering because our final charge update is calculated using a (distributed) matrix-vector multiply for which we need to keep track of the ordering.

sjplimp:
So if you are running a huge simulation on 1000 cores, does this mean each core needs to store an unordered map for
up to half the atoms in the simulation?  If so, that seems like a scalability issue.  If I understand your use case, it seems like a better solution would be to define a per-atom vector where each electrode atoms stores its 1 to M index, where M = # of electrode atoms, otherwise the atom stores 0 or -1.  Then that single value per atom could migrate with the atom and always be known to whatever proc owned the atom.
Defining that vector can be done with fix property/atom (in the input script) or by using fix STORE (internal fix) from
within one of your fixes (or other style files in the ELECTRODE package) that sets up the data structure.

srtee:
> So if you are running a huge simulation on 1000 cores, does this mean each core needs to store an unordered map for
> up to half the atoms in the simulation?  If so, that seems like a scalability issue.  If I understand your use case, it seems like a better solution would be to define a per-atom vector where each electrode atoms stores its 1 to M index, where M = # of electrode atoms, otherwise the atom stores 0 or -1.  Then that single value per atom could migrate with the atom and always be known to whatever proc owned the atom.
> Defining that vector can be done with fix property/atom (in the input script) or by using fix STORE (internal fix) from
> within one of your fixes (or other style files in the ELECTRODE package) that sets up the data structure.

You'll be happy to hear we have much bigger scalability issues than that!

Right now _each_ proc stores a matrix of groupsize^2 doubles.

With work we could bring this down to (groupsize/nproc)*groupsize on each proc -- we'd need electrode atoms to carry their matrix row with them as they move (it's a symmetric matrix so column is equivalent) with less data movement for non-electrode atoms. (Shouldn't happen too often since electrode atoms must be stationary -- we can't assume an ordered arrangement though.) Still, compared to that memory requirement an additional map of length groupsize is not too bad in my opinion.

To scale our code to 1000 cores we would need a fundamentally matrix-less mode of operation -- such as conjugate gradient -- and also probably replace PPPM with msm.

I'll still take a look again at an atom array implementation -- I think I can make it more lightweight than I did first time around. Can the exchange operations pass a variable amount of information? Ideally it would pass nothing each time a non-electrode particle is exchanged and a key-value pair each time an electrode particle is exchanged. Then we could have a set of proc-local maps.

sjplimp:
yes, you can do an exchange where each atom stores a variable amount of data.  But you would have to write the pack/unpack
methods yourself in a fix.  And you would likely have to send at least one flag per atom, to indicate it has lots of data or none.
You can see a pack_exchange and unpack_exchange method in many existing fixes, like fix property/atom.  You also have
to register the fix via atom->add_callback(Atom::GROW) so that it gets invoked with an exchange.

However, I don't suggest you worry about this, or my earlier suggestion for replacing the unordered map with a per-atom vec, for this PR.  Since you have other memory/scalability issues, I suggest
instead you merge this PR and document in its doc pages what the non-scalability issues are.  So that users don't
attempt to run a huge problem that blows out memory or CPU time.  Then worry about optimizing things in subsequent PRs.

For a CG matrix solve, you can look at the QEQ package fixes, which do that for sparse matrices, i.e. IJ atom pairs within the cutoff are non-zero.  Basically one local communication per CG iteration.  If your matrix is dense, I don't know if CG will help.

akohlmey:
> However, I don't suggest you worry about this, or my earlier suggestion for replacing the unordered map with a per-atom vec, for this PR. Since you have other memory/scalability issues, I suggest
> instead you merge this PR and document in its doc pages what the non-scalability issues are. So that users don't
> attempt to run a huge problem that blows out memory or CPU time. Then worry about optimizing things in subsequent PRs.

It would still be worth doing some profiling to see how much, if any, *time* using the map can save over looping over atoms.

I would also recommend to program some little estimator that will print a warning, when the matrix gets large (*before* it is allocated!) so that people don't just get a silent crash or drive a machine into excessive swapping before they notice that this is an unfeasible calculation. What is a meaningful threshold is probably best empirically estimated (perhaps around 1GB?).

srtee:
> > However, I don't suggest you worry about this, or my earlier suggestion for replacing the unordered map with a per-atom vec, for this PR. Since you have other memory/scalability issues, I suggest
> > instead you merge this PR and document in its doc pages what the non-scalability issues are. So that users don't
> > attempt to run a huge problem that blows out memory or CPU time. Then worry about optimizing things in subsequent PRs.
> 
> It would still be worth doing some profiling to see how much, if any, _time_ using the map can save over looping over atoms.
> 
> I would also recommend to program some little estimator that will print a warning, when the matrix gets large (_before_ it is allocated!) so that people don't just get a silent crash or drive a machine into excessive swapping before they notice that this is an unfeasible calculation. What is a meaningful threshold is probably best empirically estimated (perhaps around 1GB?).

With versions being (1) the current version, (2) the initial atom vector attempt, (3) the new map-based refactor (which will be here in about a day), I find that (1) is a bit (~1-5%) faster than (2) and a bit (~1-5%) slower than (3).

I agree with warning the user about excessive electrode sizes. The newest version will contain a memory usage function -- I could call it pre-allocation and warn the user with how much memory the fix expects to use if it's >1 GB.

srtee:
> yes, you can do an exchange where each atom stores a variable amount of data. But you would have to write the pack/unpack methods yourself in a fix. And you would likely have to send at least one flag per atom, to indicate it has lots of data or none. You can see a pack_exchange and unpack_exchange method in many existing fixes, like fix property/atom. You also have to register the fix via atom->add_callback(Atom::GROW) so that it gets invoked with an exchange.

Looking at `AtomVec::unpack_exchange`, it looks like a `fix->unpack_exchange` method can test conditions like `if (atom->mask[nlocal] & groupbit)`, since it's only ever called after basic information has already been unpacked. We'll implement that next round.

akohlmey:
please let me know when this is ready for the next round of review and testing.

srtee:
> please let me know when this is ready for the next round of review and testing.

We're adding include guards that were missing from some headers and pulling out a common implementation of `erfc` and its derivative into a separate header. Other than that the code is ready for review and testing.

akohlmey (approved):

akohlmey:
@ludwig-ahrens @robme @srtee This looks nice now. I appreciate the updates to the examples. I updated one more time with upstream and made some minor cosmetic changes but I have no more requests.
@sjplimp do you have anything else to comment or would like to have changed?

robme:
I think you are tagging the wrong person because I have no idea what you are talking about. You probably want @robeme. Good luck with whatever it is you are doing.

sjplimp (approved):

Xian-qi:
I have successfully installed lammps-23Jun2022 with ELECTRODE package using parallel_studio_xe_2019_update4. But i can't repeat this example "graph-il" when i use this command "fix conp bot electrode/conp -1.0 1.979 couple top 1.0 symm on".  An error “KILLED BY SIGNAL: 11 (Segmentation fault)” will appear on my computer.  Anything is ok when i try to repeat other examples. Can you give me a hand?

srtee:
Could you try with the latest develop branch and see if that makes a difference? We have just pushed a number of major updates to the package.

If you have any further errors, it would be best to raise it either as an Issue on GitHub or to ask in the MatSci forums (https://matsci.org/c/lammps/40).

Xian-qi:
> 

Thank you  for your reply. I will try the latest version.

PR #3195, opened 2022-03-30
kk/reax: Preprocessing optimizations to ComputeAngular,Torsion; Modularity/Memory improvements in BuildLists

weinbe2:
**Summary**

This PR primarily provides performance improvements to the `ComputeAngular`/`ComputeTorsion` portions of evaluating the Reax forces/potential by iterative improvement on the ideas in #3147 --- partitioning the calculations into two segments:
* A low-intensity, high-divergence "preprocessing" stage which builds "triplets" and "quads" of atom interactions via a parallel "queue" mechanism which keeps this stage fully device-resident.
* A nearly perfectly convergent "compute" stage.

This PR also reduces memory overheads by eliminating the pre-computation of `dx`/`dy`/`dz` triplets, instead pre-computing just a `d`-differential term which can be rescaled by `delta x,y,z` later.

There is also some amount of code abstraction into reusable functions that can be shared across, for ex, the multiple versions of `BuildLists`.

These changes, in optimal combination (`OPT_ANGULAR_TORSION` and `OPT_REDUCE_DXDYDZ`, see definitions below), lead to a 2 to 10% performance improvements on A100-80GB-SXM4, measured with CUDA 11.4.2, where the size of the improvement increases as a function of problem size. The "best case" 1xGPU performance improvement was seen for 3,735,552 atoms; a boost from 4.14e6 atom timesteps / second to 4.56e6 atom timesteps / second, a ~10% improvement. In both cases `lists/blocking on` was used, with default values otherwise.

Changes lead to a small but measurable regression (sub-5%) with some cursory investigations on CPUs, likely due to some redundant work in the pre-processing described below. I think this could be addressed by divergent CPU/GPU paths unique to the pre-processing stage.

Correctness and performance has been verified against the HNS Reax problem; correctness checks included setting the CG residual tolerance to `1e-16` to eliminate that potential source of run-to-run variation.

**Related Issue(s)**

Improvements built on work done in #3147 , #3158 .

**Author(s)**

**Licensing**

**Backward Compatibility**

No issues.

**Implementation Notes**

As a mechanism to simplify review, three new changes are guarded by three new `#ifdef`. Ideally sub-optimal paths should be removed, and `#ifdef` eliminated, before the merge.

* `OPT_ANGULAR_TORSION`: Toggle the new codepaths for `ComputeAngular`/`ComputeTorsion`. This utilizes a three-pass approach.

1. Counting (`TagPairReaxCountAngularTorsion<false>`), which pre-computes the size of the queues. This uses modular functions `preprocess_angular<false>` and `preprocess_torsion<false>` to handle separate countings in a fused kernel.
2. Populating (`TagPairReaxCountAngularTorsion<true>`), which populates the queues with `int4`-esque quads of `i,j,k[,l]` values and offset indices that pass cutoff tests. This reuses `preprocess_[angular/torsion]<false>` and `<true>`. This kernel also handles pre-computing quantities like `SBO`, which only occur once per value of `i`, abstracted via `compute_angular_sbo`.
3. Computing (`TagPairReaxComputeAngularPreprocessed`, `TagPairReaxComputeTorsionPreprocessed`), which loads the quads and evaluates the relevant portions .

Of note: the "population" step generates a queue non-deterministically, in that the order of individual `i` can change from run to run. However, all values of `j`,`k`,`l` at _fixed_ `i` are contiguous, which promotes good cache reuse. While, on one hand, non-determinism is non-ideal, on the other hand, all forces in Angular/Torsion are accumulated via floating point atomics "anyway," so this isn't an additional "evil". 

Of course, addressing this is a point of discussion. Forming a guaranteed ordered list requires a `parallel_scan` on outputs from "counting" to generate ordered offsets for the "populate" step.

As a note: the "preprocessed" functions still have indentation that matches where the `for` loops _were_ in the non-preprocessed versions. This was left in, in my opinion, to help "guide the eye," but they do violate general code indentation rules and could be replaced by comments---I'm not married to my decision in the slightest.

* `OPT_SPLIT_COUNT_ANGULAR_TORSION`: Toggle fission of angular and torsion preprocessing (dependent on the above optimization), exploiting the modular functions for quick separate implementations. This fission actually leads to a sub-percentile reduction in performance on A100-80GB, but one could argue that the fission could reduce cache thrashing, so it could be worth investigating on other architectures before making a final decision. This splits `TagPairReaxCountAngularTorsion` into `CountAngular` and `CountTorsion` flavors.
* `OPT_REDUCE_DXDYDZ`: Trade memory for a little bit of extra compute by, for ex, replacing ` d_dln_BOp_pix, d_dln_BOp_piy, d_dln_BOp_piz` with just `d_dln_BOp_pi`; the `x,y,z` values can be re-generated by multiplying by `delta ij`, etc, as appropriate. In all relevant cases `delta ij`, etc, has already been computed in kernels so there is no additional memory i/o incurred from this optimization. This effects the `BuildLists` kernels on the write and `TagPairReaxComputeBond2` on the read, leading to a ~30% speed-up on `ComputeBond2`; ultimately a O(1%) performance boost for end-to-end runtime for the HNS system.

Some additional abstraction functions were introduced, as noted above, to eliminate redundant code across the various `BuildLists` functions. These are unguarded as the compiler should be able to inline these functions appropriately. The three additional functions are:

1. `build_hb_list`: abstracts the cutoff checking and building of the `hb` lists.
2. `build_bo_list`: abstracts the building of the `bo` lists, returns `true`/`false` if the `BO_*`, etc, values need to be subsequently stored so these functions can be reused for the preprocessed and non-preprocessed versions.
3. `compute_bo`: computes `BO_s`, `BO_pi`, `BO_pi2`, `C12`, `C34`, `C56`. 

There is some scope to create a modular function for the parts of `BuildLists` that write the `(i,j_index)` values, but it requires some decisions on how to handle "full" vs "half" cases that could ultimately lead to a comparable number of lines of code anyway.

**Post Submission Checklist**

None.

### DIFF
--- a/src/KOKKOS/kokkos_type.h
+++ b/src/KOKKOS/kokkos_type.h
@@ -1214,6 +1214,15 @@ struct params_lj_coul {
   F_FLOAT cut_ljsq,cut_coulsq,lj1,lj2,lj3,lj4,offset;
 };
 
+#ifdef OPT_ANGULAR_TORSION
+// ReaxFF
+
+struct alignas(4 * sizeof(int)) reax_int4 {
+  int i0, i1, i2, i3;
+};
+
+#endif
+
 // Pair SNAP
 
 #define SNAP_KOKKOS_REAL double
--- a/src/KOKKOS/kokkos_type.h
+++ b/src/KOKKOS/kokkos_type.h
@@ -1214,14 +1214,11 @@ struct params_lj_coul {
   F_FLOAT cut_ljsq,cut_coulsq,lj1,lj2,lj3,lj4,offset;
 };
 
-#ifdef OPT_ANGULAR_TORSION
 // ReaxFF
 
 struct alignas(4 * sizeof(int)) reax_int4 {
   int i0, i1, i2, i3;
 };
-
-#endif
 
 // Pair SNAP
 
--- a/src/KOKKOS/pair_reaxff_kokkos.cpp
+++ b/src/KOKKOS/pair_reaxff_kokkos.cpp
@@ -958,12 +958,10 @@ void PairReaxFFKokkos<DeviceType>::compute(int eflag_in, int vflag_in)
   count_angular = h_count_angular_torsion(0);
   count_torsion = h_count_angular_torsion(1);
 
-  if (count_angular > d_angular_pack.extent(0)) {
+  if (count_angular > d_angular_pack.extent(0))
     d_angular_pack = t_reax_int4_2d("reaxff:angular_pack",(int)(count_angular * 1.1),2);
-  }
-  if (count_torsion > d_torsion_pack.extent(0)) {
+  if (count_torsion > d_torsion_pack.extent(0))
     d_torsion_pack = t_reax_int4_2d("reaxff:torsion_pack",(int)(count_torsion * 1.1),2);
-  }
 
   // need to zero to re-count
   h_count_angular_torsion(0) = 0;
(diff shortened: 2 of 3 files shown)


stanmoore1:
Thanks @weinbe2, will test this on V100 and AMD.

@arghdos

arghdos:
> Of note: the "population" step generates a queue non-deterministically, in that the order of individual i can change from run to run ... While, on one hand, non-determinism is non-ideal, on the other hand, all forces in Angular/Torsion are accumulated via floating point atomics "anyway," so this isn't an additional "evil".

> Of course, addressing this is a point of discussion. Forming a guaranteed ordered list requires a parallel_scan on outputs from "counting" to generate ordered offsets for the "populate" step.

Nice, this was on my to-do list after we nix'd the non-coherent host-allocs during #3147.  I had come to a similar conclusion on the scan, it might be worthwhile to do as an optional step (for benchmarking / scale-out), but even doing a really hacky sort on the host didn't affect subsequent kernel timings too much.

Will take a look

arghdos:
Tested on our MI-250's, seeing a 12-13% uplift on a single GPU.  I get the best perf w/ `OPT_REDUCE_DXDYDZ` on, and `OPT_SPLIT_COUNT_ANGULAR_TORSION` off, but they're both about a 0.5% swing either way.

arghdos:
A similar bump for an 8-GPU case.  This is very nicely done @weinbe2.

weinbe2:
> Tested on our MI-250's, seeing a 12-13% uplift on a single GPU. I get the best perf w/ `OPT_REDUCE_DXDYDZ` on, and `OPT_SPLIT_COUNT_ANGULAR_TORSION` off, but they're both about a 0.5% swing either way.

I assume you mean with `OPT_ANGULAR_TORSION` as well --- in which case that means MI250, A100, and MI100 (per your other comment) are optimal with the same combo of two `#define`s, which is great!

> This is very nicely done @weinbe2.

Much appreciated!

arghdos:
> I assume you mean with `OPT_ANGULAR_TORSION` as well

Ah yes, this was on!

stanmoore1:
I see nearly 10% speedup for V100. `OPT_SPLIT_COUNT_ANGULAR_TORSION` seems to be in the noise. Will check the CPU next.

stanmoore1 (approved):
Looks great, thanks @weinbe2!

PR #3197, opened 2022-03-31
add support for custom header keywords with thermo output and dump styles

akohlmey:
**Summary**

This allows to customize the thermo output by replacing the default header keywords for the thermo data with custom strings.
This is implemented via thermo_modify. The header corresponding to the (default or custom) thermo keyword can be indexed from the left with a positive integer (starting from 1) or from the right with a negative integer (starting from -1) or by the thermo keyword (including compute, fix, property, or variable reference).

The indexing for the custom format string has been expanded to do reverse or by thermo keyword indexing as well.

For dump styles (atom, custom, cfg) the same feature exists and is implemented in a similar style through the dump_modify command with similar syntax.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

The negative indexing disregards the Volume keyword that is added when thermo style one or multi is used with a variable cell fix.

**Post Submission Checklist**


akohlmey:
@sjplimp looks like we need to find a different keyword instead of "header" since dump_modify has already a "header" keyword with a different meaning. Since the logic/syntax for the "dump_modify format" keyword is the same as for "thermo_modify format", I would like to have the same keywords with the same syntax for the custom column headers as well.

- "keyword" would match the variable name in the code, but that would probably be too confusing, since keyword is used a lot to describe those terms.
- "headername" or "columnname" are a bit long and the second easy to misspell
- "name" is too generic
- "keyname" perhaps? but it is an odd name

any ideas?

sjplimp:
colname

akohlmey:
@sjplimp dump_modify colname is now also available.

sjplimp (approved):

PR #3198, opened 2022-04-01
DPD-BASIC kokkosification 

bathmatt:
**Summary**

Developed Kokkos versions of the four routines in DPD-BASIC

**Author(s)**

Developed by Matt Bettencourt mattb@nvidia.com  Internal review by Evan Weinberg eweinberg@nvidia.com

**Licensing**

**Backward Compatibility**

no

**Implementation Notes**

Set the RNG within the modules and confirmed that the results were identical for the first many steps, 

**Post Submission Checklist**


src/DPD-BASIC/pair_dpd.h
```diff
@@ -28,11 +28,11 @@ class PairDPD : public Pair {
  public:
   PairDPD(class LAMMPS *);
   ~PairDPD() override;
-  void compute(int, int) override;
+  virtual void compute(int, int) override;
```
akohlmey, line 31: Why add "virtual" here? The "override" should be sufficient.

bathmatt: Habit mostly, I can remove all the virtuals in the DPD dir, I typically use virtual if someone builds on top of it.

akohlmey: As far as I understood the standard, functions don't lose their polymorphic status when "override" is used (they do without). So "virtual" is needed for the first definition and then "override" for all following unless you use "final" to prohibit further overrides.

bathmatt: That is the way I understand it as well.  I will remove the added virtuals as that is the style of LAMMPS as of a bit ago.

akohlmey: Perhaps some background info is helpful: we've run the code through clang-tidy to modernize some common constructs and would like to keep all new code consistent with that, so we don't have to do the clang-tidy processing all the time. :wink:

stanmoore1: >So "virtual" is needed for the first definition and then "override" for all following unless you use "final" to prohibit further overrides.

I wasn't aware of this--will keep this in mind for future code as well.

bathmatt: I pushed code without the virtual where it isn't needed, it is still needed in allocate cuz that is the start of the chain.

bathmatt:
@stanmoore1 @akohlmey how does this look?  tutti bene?  wanna close it out on my end if it is good else fix things before I forget.
Thanks!

akohlmey:
> @stanmoore1 @akohlmey how does this look? tutti bene? wanna close it out on my end if it is good else fix things before I forget. Thanks!

@bathmatt not yet *tutto bene*, but it is currently in the hands of @stanmoore1 and he'll eventually get to check it out.
As far as I am concerned, the things that need to be done are just minor changes to build system, documentation, etc. to fully integrate your changes. Those modifications are easier to do myself than to explain. So unless you explicitly ask be to point them all out and explain, I'll just wait and do them after the science has been checked and confirmed.

stanmoore1:
@bathmatt this is really close. I just need to do final checks and also need a few updates to the docs.

stanmoore1:
@akohlmey I think it would be good to add an example for `pair_dpd` and `pair_dpd_tstat`.

akohlmey:
> @akohlmey I think it would be good to add an example for `pair_dpd` and `pair_dpd_tstat`.

done

bathmatt:
Let me know what I can do to help out.

stanmoore1:
This passes all my tests now.

stanmoore1:
@akohlmey this PR is ready to merge from my point of view.

sjplimp (approved):

PR #3159, opened 2022-03-03
Refactor neighbor list request code

akohlmey:
**Summary**

This pull request refactors the neighbor list request code. The objective is to limit the need to access the list of requests directly.  Also these changes enforce that flags are set consistently. Some additional, rarely used changes of flags can be triggered through setter functions of the NeighRequest class. Many custom neighbor list requests collapse to one line.

Despite adding about 150 lines of documentation, this pull request results a net reduction of 2300 lines of code.

The old API is still used internally but public access to the NeighRequest member variables has been changed to protected.
Neighborlist related classes have been declared friend to retain access.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

Yes. Functionality is unchanged. Source code compatibility is currently preserved.

**Implementation Notes**

The NeighRequest class uses C++11 style "delegated constructors" to increase code reuse.

**Post Submission Checklist**

- [x] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [x] The source code follows the LAMMPS formatting guidelines
- [x] Suitable new documentation files and/or updates to the existing docs are included
- [x] The added/updated documentation is integrated and tested with the documentation build system
- [x] The feature has been verified to work with the conventional build system
- [x] The feature has been verified to work with the CMake based build system
- [ ] Suitable tests have been added to the unittest tree.
- [ ] A package specific README file has been included or updated
- [ ] One or more example input decks are included

PR #3199, opened 2022-04-01
Misc changes

stanmoore1:
**Summary**

- Fix numerical bug in Kokkos SNAP on GPUs
- Fix wrong name in Kokkos package docs
- Update fix ACKS2 docs
- Immediately error out when compute msd (and msd/nongauss) is used with a dynamic group and document it.
  This addresses https://matsci.org/t/calculating-msd-for-a-dynamic-group/41453
- Add "transrot" keyword to fix move to allow prescribing both linear translation and rotation of the fix group atoms at the same time. This addresses https://matsci.org/t/how-to-achieve-rotation-and-translation-of-a-sphere-at-the-same-time/41417
- enable and apply clang-format for fix move, compute msd and compute msd/nongauss

**Related Issue(s)**

#3159 

**Author(s)**

Stan Moore (SNL)
Axel Kohlmeyer (Temple U)

**Licensing**

**Backward Compatibility**

Yes


athomps (commented):
What is the effect of the bug/bugfix. Does it affect previous simulations?

stanmoore1:
> What is the effect of the bug/bugfix. Does it affect previous simulations?

@athomps #3159 caused that all the forces, energy, and pressure are all zero from Kokkos SNAP when running on GPUs. It does not affect versions before #3159 was merged.

athomps (approved):
I approve.

PR #3203, opened 2022-04-04
Fix issues in ReaxFF QEq and ACKS2

stanmoore1:
**Summary**

The removal of ghost atoms in the fix QEq and ACKS2 neigh lists (5888ba05146f12a618ef2125a29ac385c3167f5c) introduced a potential out of bounds access in the neigh list `ilist` and also caused the BiCGStab solver to fail with bad numerics when using a full list with Kokkos.

**Related Issue(s)**

5888ba05146f12a618ef2125a29ac385c3167f5c

**Author(s)**

Stan Moore (SNL), reported by Ray Shan (Materials Design)

**Licensing**

**Backward Compatibility**

Yes


athomps (approved):
Very good. I approve.

stanmoore1:
This passes all the tests I ran by hand. Full Kokkos regression testing is pending.

stanmoore1:
@akohlmey weird that the unit tests had a segfault, I didn't see that in my testing.

akohlmey:
> @akohlmey weird that the unit tests had a segfault, I didn't see that in my testing.

Not weird. The unittest procedure does the equivalent of multiple "clear" commands and "write/read_restart/data" and that is often not done in "normal" inputs. This regularly exposes when restarts don't store all relevant data or when data is not properly initialized when used the first time. Since Linux uses a malloc() that derives large memory blocks from using mmap(/dev/zero) with copy-on-write semantics, class members are initialized to 0 the first time around so many problems don't show, but after a "clear" memory is free()d and allocated and then will contain "random" data and then uninitialized variables can cause problems.

PR #2921, opened 2021-09-03
Enhance MDI support in LAMMPS in its MDI package

sjplimp:
**Summary**

This PR does several things wrt support in LAMMPS for the MDI code coupling library.  All in package MDI.

LAMMPS can now be used in all of the modes MDI supports.  As a driver code, coupled to engines that are either stand-alone codes or plugin libraries.  As an engine code, either in stand-alone mode or as a plugin library.

a) Refactors the mdi/engine and fix mdi/engine commands, so that all the MDI logic is in the former.  Also changes
    the name of the command to mdi engine, so that other variant mdi commands can be added.
b) Adds a fix mdi/aimd command so that LAMMPS can act as a driver for an AIMD calculation.
c) Adds custom MDI engine commands appropriate for using LAMMPS as an MD engine
d) Adds a mdi plugin command so that LAMMPS as an MDI driver can use an engine either as a standalone code or a plugin library
e) Adds an examples/mdi dir with several use cases with many input scripts and a README to explain how to use LAMMPS with MDI in all the various modes

**Related Issue(s)**

N/A

**Author(s)**

Steve and Taylor Barnes (MolSSI at Virginia Tech)

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

Should just add functionality, but haven't yet tested for backward compatibility.

**Implementation Notes**

**Post Submission Checklist**

- [ ] The feature or features in this pull request is complete
- [ ] Licensing information is complete
- [ ] Corresponding author information is complete
- [ ] The source code follows the LAMMPS formatting guidelines
- [ ] Suitable new documentation files and/or updates to the existing docs are included
- [ ] The added/updated documentation is integrated and tested with the documentation build system
- [ ] The feature has been verified to work with the conventional build system
- [ ] The feature has been verified to work with the CMake based build system
- [ ] Suitable tests have been added to the unittest tree.
- [ ] A package specific README file has been included or updated
- [ ] One or more example input decks are included

**Further Information, Files, and Links**

PR #3207, opened 2022-04-07
remove MESSAGE package

sjplimp:
**Summary**

Remove the MESSAGE package from LAMMPS, since it is now superceded by the MDI package (due to Taylor Barnes at MolSSI).   Which is superior and higher-level in several ways.  Also easier for users to write code for and/or use in conjunction with other codes which include MDI support.

**Related Issue(s)**

This PR should NOT be merged until PR #2921 is merged.

**Author(s)**

Steve

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**


sjplimp:
@akohmley See the note at the top about not processing this PR until after PR #2921.  Also
(a) Noticed that the -mpicolor command-line option was added to support the MDI package.  Don't know that we now have another use case for it.  But wouldn't hurt to leave it in. 
(b) Also removed 3 sub-dirs from examples/COUPLE.  All used the MESSAGE package. One for VASP, one for NWChem, one for an example toy MC code.  I think they would all be possible with the MDI package, but would require a re-write.

akohlmey (approved):

PR #3210, opened 2022-04-08
Add non-volatile version of join function to DomainKokkos

arghdos:
**Summary**

Adds non-volatile overload to fix build error w/ Kokkos commit (067c2593b58e1d9204d097432e221cfd524d7978)

**Related Issue(s)**

N/A

**Author(s)**

Nicholas Curtis

**Licensing**

**Backward Compatibility**

I am not 100% sure -- was there a time where Kokkos only had a volatile overload? If so, that might break.

**Post Submission Checklist**


arghdos:
@stanmoore1 -- this passes my validation tests for LJ/EAM/Tersoff/ReaxFF&SNAP, but I am not 100% confident on what this functor is doing.

stanmoore1:
@arghdos this fails to compile with the Intel compiler, I think we have to wait until we update the Kokkos version in LAMMPS too. :(

stanmoore1:
This will be fixed in Kokkos instead, see https://github.com/kokkos/kokkos/pull/4951.

arghdos:
FWIW, we might still want to have a non-volatile overload for join (hopefully that will make the Intel compiler happy)

stanmoore1:
@arghdos I agree, will do.

stanmoore1 (approved):

PhilMiller:
Once kokkos#4951 is merged, you'll ideally want to just want to change the parameters to your `join()` to not be `volatile` qualified. That should potentially offer better performance.

I don't know the code's policy/practices in terms of forward/backward Kokkos compatibility, though. Having both overloads will maintain operation with Kokkos 3.6 as well as `develop` (post 4951)

stanmoore1:
@PhilMiller thanks for the info. We'll plan to remove the `volatile` when the version of Kokkos bundled with with LAMMPS includes 4951 (probably Kokkos v3.7).

PR #3211, opened 2022-04-09
Collected small changes

akohlmey:
**Summary**

This pull request collects various small changes and fixes

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

N/A

**Post Submission Checklist**

N/A

### DIFF
--- a/python/lammps/formats.py
+++ b/python/lammps/formats.py
@@ -19,9 +19,9 @@
 
 import re, yaml
 try:
-  from yaml import CSafeLoader as Loader, CSafeDumper as Dumper
+  from yaml import CSafeLoader as Loader
 except ImportError:
-  from yaml import SafeLoader as Loader, SafeDumper as Dumper
+  from yaml import SafeLoader as Loader
 
 class LogFile:
   """Reads LAMMPS log files and extracts the thermo information


sjplimp (approved):

jibril-b-coulibaly (commented):
Hello,

I had issue with my code and isolated 2 erroneous expressions that would require correction:
- in `lib/gpu/lal_device.cpp`, line 206: `type_match == false` should be changed to `!type_match`, the `!` is missing.
- in `src/GRANULAR/pair_granular.cpp`, the old expression `(update->setupflag) ? false : true` and the new one `update->setupflag != 0` are logical opposite. The new expression should be `update->setupflag == 0`.

I would also have a question that might lead to some simplifications. it seems that `update->setupflag` and `field->threadflag` only take values 0 and 1. What is the rationale of having them as integer, rather than boolean? Are there any considerations of using other flag values (e.g., 2, 3...) in the future?

Thank you,
Jibril B. Coulibaly

akohlmey:
> Hello,
> 
> I had issue with my code and isolated 2 erroneous expressions that would require correction:
> 
> * in `lib/gpu/lal_device.cpp`, line 206: `type_match == false` should be changed to `!type_match`, the `!` is missing.
> * in `src/GRANULAR/pair_granular.cpp`, the old expression `(update->setupflag) ? false : true` and the new one `update->setupflag != 0` are logical opposite. The new expression should be `update->setupflag == 0`.

Thanks for the notes.

> I would also have a question that might lead to some simplifications. it seems that `update->setupflag` and `field->threadflag` only take values 0 and 1. What is the rationale of having them as integer, rather than boolean?

The reason is historical. LAMMPS started in a "C with classes" style of programming. And the prevalent C standard at the time didn't have a bool data type.  Starting with the requirement of the C++11 standard, we are slowly relaxing this style to allow more C++ and try to get rid of the more convoluted/complex parts of  C (like string handling).

> Are there any considerations of using other flag values (e.g., 2, 3...) in the future?

You never know. It would be using constants from an enumerator, though. Getting rid of "magic" numbers (e.g. for update->whichflag) is one of the refactoring targets.

PR #3212, opened 2022-04-12
Updates to the DIELECTRIC package

ndtrung81:
**Summary**

This PR contains several updates to the DIELECTRIC package such as adding new pair style (lj/cut/coul/debye/dielectric) and its omp variant, correcting and updating the doc page of fix polarize, and allowing users to set the per-atom dielectric constant via the set command 

**Related Issue(s)**

N/A

**Author(s)**

Trung Nguyen (U Chicago)

**Licensing**

**Backward Compatibility**

Should be maintained.

**Implementation Notes**

As users allow to change the per-particle dielectric constant, which will rescale the particle charge accordingly, care must be taken if they also use the set charge command for the same group of particles. This is documented in the doc page of the set command.

**Post Submission Checklist**


akohlmey:
@ndtrung81 remember to assign the PR to me when it is ready for me to check it out and merge it.

src/DIELECTRIC/fix_polarize_bem_gmres.cpp
```diff
@@ -276,6 +281,15 @@ void FixPolarizeBEMGMRES::setup(int /*vflag*/)
...
   }
 
+  // NOTE: qqrd2e = e^2/(4 pi e0) in kcal/mol
+  // converting epsilon0 [efield] from (F/m)(kcal/mol/e/A) into e/A^2
+  // epsilon0e2q = epsilon0 * (4184/6.023e+23) * (1/1.6e-19)^2 / (1e+10) / 332.06
+  //             = 0.000240258
+
+  epsilon0e2q = 1.0;
+  if (strcmp(update->unit_style,"real") == 0)
+    epsilon0e2q = 0.0795776 / force->qqrd2e;
```
akohlmey, line 287: Can we compute this constant with larger precision?
Should this perhaps be moved to conversion constants in the Force class?
Do I understand this correctly that this fix will work properly only with "lj" and "real" units?
I am adding a check and error messages for now.

akohlmey (changes requested):
@ndtrung81 I made one significant change and asked questions about the units dependent conversion factor in a few of the styles.
Please check it out and clarify/correct my changes if needed.
The rest is mostly cosmetic. It is high time also to move the "ewald" erfc() approximation constants into a header. Will apply this to other styles later in a more general PR.

src/DIELECTRIC/fix_polarize_bem_gmres.cpp
ndtrung81: @akohlmey Thanks for pointing out the units dependent factor and for adding the check and error message. In commit 7c64ff6, I have improved the precision of the constant epsilson0e2q. I have also computed this constant for "metal", "si" and "nano". The rest may come later, but these are the commonly used units relevant to these fixes in my opinion. The error message is updated to reflect the supported units. The doc page of the two fix polarize fixes is also updated to mention this restriction.

I can clean up the comments that explain how I calculated this factor for units other than lj, if necessary. For the moment, I only see this conversion factor present only in these two fixes, so no need to move it to the Force class.

akohlmey (approved):

PR #3222, opened 2022-04-20
Add fasthtml target for faster testing of .rst to .html translation

akohlmey:
**Summary**

This adds a "fasthtml" target to the `doc` folder make file. This performs an approximate but cached translation from reStructuredText to html so that after the first translation all subsequent translations will be fast since only modified
files are regenerated. Not all sphinx extensions are supported and checks of indices and cross-links may be incomplete.
This can significantly speed up the debugging of .rst files during development.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Post Submission Checklist**

### DIFF
--- a/doc/.gitignore
+++ b/doc/.gitignore
@@ -1,6 +1,6 @@
 /old
 /html
-/html-offline
+/fasthtml
 /epub
 /latex
 /mathjax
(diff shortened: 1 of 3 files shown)


athomps (approved):
This works great. Speedup is about 10x, even when using 4 cores.

akohlmey:
> This works great. Speedup is about 10x, even when using 4 cores.

FYI: make -j has no impact on sphinx processing.

sjplimp (approved):

PR #3223, opened 2022-04-21
Reduce excessive warnings from fmtlib and assorted small changes

akohlmey:
**Summary**

Some compilers produce excessive warnings with some code constructs, specifically in fmtlib.
Most of these warnings are inconsequential and thus very distracting but also problematic since the number of output lines they create can easily hide important and relevant warnings.

Also some small changes for improved portability and fixes for typos and inconsistencies in the manual are included.

This pull request aims to reduce them.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No problems expected.

**Implementation Notes**

Individual changes:
- propagate CMAKE_TUNE_FLAGS to compilation of the lammps executable, the colvars library, and the unittest tree
- PGI/NVHPC compilers by default warn about dead code which can happen with portable code that uses pre-processing for portability. Since they don't allow any selective reduction of warnings we set CMAKE_TUNE_DEFAULT to disable all warnings.
- Intel's new LLVM based compilers warn about `isinf()` always be true in fast floating point mode. Add flag to CMAKE_TUNE_DEFAULT to suppress this warning.
- comment out assertion in a fmtlib template specialization that may be always true based on the data type of the template argument
- allow the C++ standard to be set from the CMake command line and do not reset it. Make certain that c++11 or later is used.
- KOKKOS with CUDA no longer need to use `nvcc_wrapper` when compiling with CMake. Update preset.

**Post Submission Checklist**


sjplimp (approved):

cmake/CMakeLists.txt
```diff
@@ -468,6 +488,7 @@ set(CMAKE_TUNE_FLAGS "${CMAKE_TUNE_DEFAULT}" CACHE STRING "Compiler and machine
 separate_arguments(CMAKE_TUNE_FLAGS)
 foreach(_FLAG ${CMAKE_TUNE_FLAGS})
   target_compile_options(lammps PRIVATE ${_FLAG})
```
junghans, line 490: I think it is a bit cumbersome to have to add the flags in all these 3+ places. I think we could achieve the same by:
```suggestion
  target_compile_options(lammps PUBLIC $<BUILD_INTERFACE:${_FLAG}>)
```
here and then the `lammps` consumers such as `lmp`, unittest` would get them automatically.

Another options would be to create a `flags` target, e.g.:
```
add_library(LAMMPS::flags INTERFACE IMPORTED)
target_compile_options(LAMMPS::flags PUBLIC $<BUILD_INTERFACE:${_FLAG}>)
```
and then use `target_link_libraries(<foo> PRIVATE LAMMPS::flags)` to inject the flags for each target.
That would also work for `covals`, which isn't linked against the `lammps` target. With the `flags` target we could clean up the `LAMMPS_SIZE` defines all over the place, too.

junghans (approved):
All fine, just one comment.

src/MAKE/OPTIONS/Makefile.kokkos_cuda_mpi
```diff
@@ -10,7 +10,7 @@ KOKKOS_ABSOLUTE_PATH = $(shell cd $(KOKKOS_PATH); pwd)
 export MPICH_CXX = $(KOKKOS_ABSOLUTE_PATH)/bin/nvcc_wrapper
 export OMPI_CXX = $(KOKKOS_ABSOLUTE_PATH)/bin/nvcc_wrapper
 CC =		mpicxx
-CCFLAGS =	-g -O3 -DNDEBUG
+CCFLAGS =	-g -O3 -DNDEBUG -Xcudafe --diag_suppress=unrecognized_pragma
```
stanmoore1, line 13: @akohlmey can you please add this to `MAKE/MACHINES/Makefile.summit_kokkos` too?

stanmoore1 (approved):

PR #3224, opened 2022-04-21
change update of dynamic groups to post_force location in timestep

sjplimp:
**Summary**

For dynamic groups, change the point in the timestep where they are updated.  Previously this was post_intergrate().
Now it is post_force() and before any user-specified fixes which operate at post_force() are invoked.  This simplifies several issues:

(1) use of computes for the dynamic criterion which required ghost atom positions, e.g. compute cluster/atom and other similar computes.  This information is now current w/out the need to any extra communication.  Likewise a compute that uses pe/atom
or virial/atom can be used as a criterion, since pair styles have now been communicated.  There is also no longer an issue with the compute being invoked so early in the timestep that it's info is out-of-date by the end of the timestep.

**Related Issue(s)**

This solves the problem addressed by PR #3018, but in a simpler fashion and thus closes #3018

**Author(s)**

Steve, with an assist from Davie Quigley (U Warwick)

**Licensing**

**Backward Compatibility**

Don't think this will change the behavior of dynamic groups, unless they were invoking computes which should now work correctly.

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/doc/src/group.rst
+++ b/doc/src/group.rst
@@ -258,11 +258,17 @@ assignment is made at the beginning of the minimization, but not
 during the iterations of the minimizer.
 
 The point in the timestep at which atoms are assigned to a dynamic
-group is after the initial stage of velocity Verlet time integration
-has been performed, and before neighbor lists or forces are computed.
-This is the point in the timestep where atom positions have just
-changed due to the time integration, so the region criterion should be
-accurate, if applied.
+group is after interatomic forces have been computed, but before any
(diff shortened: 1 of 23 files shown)


sjplimp:
@akohlmey I think I fixed the test script issue

akohlmey:
> @akohlmey I think I fixed the test script issue

now compilation was broken, so had to propagate your changes to some more packages that were missing them.

akohlmey (approved):

PR #3228, opened 2022-04-22
A pull request for an optimized ILP(ilp/graphene/hbn) potential 

run-towards-the-future:
**Summary**

This pair style, ilp/graphene/hbn/opt contains our optimizations in SWLMFF. Since LMFF is not tidy enough to be integrated with LAMMPS, we currently can only include an optimized ILP(ilp/graphene/hbn) potential as a pull request.

**Related Issue(s)**

**Author(s)**

Optimization author 1: Ping Gao (National Supercomputing Center in Wuxi, China)
E-mail: qdgaoping@gmail.com
Optimization author 2: Xiaohui Duan (National Supercomputing Center in Wuxi, China)
E-mail: sunrise_duan@126.com

**Licensing**

**Backward Compatibility**

There is no changes in the pull request will break backward compatibility for inputs.

**Implementation Notes**

For the first committing, we add a new pair style, ilp/graphene/hbn/opt.
And this pair style updates to the OPT package.
For the second committing, we improve the performance of p
[SC21-LMFF.pdf](https://github.com/lammps/lammps/files/8541776/SC21-LMFF.pdf)
air style, ilp/graphene/hbn/opt.
And this improvement updates to the OPT package.

**Post Submission Checklist**

The optimizations have been included in our published paper, titled "LMFF: Efficient and Scalable Layered Materials Force Field on Heterogeneous Many-Core Processors". 
Url = {https://doi.org/10.1145/3458817.3476137},
Doi = {10.1145/3458817.3476137},

### DIFF
--- a/doc/src/Commands_pair.rst
+++ b/doc/src/Commands_pair.rst
@@ -124,7 +124,7 @@ OPT.
    * :doc:`hbond/dreiding/lj (o) <pair_hbond_dreiding>`
    * :doc:`hbond/dreiding/morse (o) <pair_hbond_dreiding>`
    * :doc:`hdnnp <pair_hdnnp>`
-   * :doc:`ilp/graphene/hbn <pair_ilp_graphene_hbn>`
+   * :doc:`ilp/graphene/hbn (t) <pair_ilp_graphene_hbn>`
    * :doc:`ilp/tmd <pair_ilp_tmd>`
    * :doc:`kolmogorov/crespi/full <pair_kolmogorov_crespi_full>`
    * :doc:`kolmogorov/crespi/z <pair_kolmogorov_crespi_z>`
--- a/doc/src/pair_ilp_graphene_hbn.rst
+++ b/doc/src/pair_ilp_graphene_hbn.rst
@@ -1,7 +1,10 @@
 .. index:: pair_style ilp/graphene/hbn
+.. index:: pair_style ilp/graphene/hbn/opt
 
 pair_style ilp/graphene/hbn command
 ===================================
+
+Accelerator Variant: *ilp/graphene/hbn/opt*
 
 Syntax
 """"""
@@ -122,6 +125,10 @@ headings) the following commands could be included in an input script:
    variable Evdw  equal c_0[1]
    variable Erep  equal c_0[2]
    thermo_style custom step temp epair v_Erep v_Evdw
+
+----------
+
+.. include:: accel_styles.rst
 
 ----------
 
--- a/src/.gitignore
+++ b/src/.gitignore
@@ -1091,6 +1091,8 @@
 /pair_hdnnp.h
 /pair_ilp_graphene_hbn.cpp
 /pair_ilp_graphene_hbn.h
+/pair_ilp_graphene_hbn_opt.cpp
+/pair_ilp_graphene_hbn_opt.h
 /pair_ilp_tmd.cpp
 /pair_ilp_tmd.h
 /pair_kolmogorov_crespi_full.cpp
--- a/src/Depend.sh
+++ b/src/Depend.sh
@@ -97,6 +97,10 @@ if (test $1 = "GRANULAR") then
   depend OPENMP
 fi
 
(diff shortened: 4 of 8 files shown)


akohlmey (changes requested):
@run-towards-the-future thanks for your contribution. Unfortunately, it will need some additional effort before a more thorough review will be done, and it can eventually be merged into the LAMMPS distribution.
1) It needs to be implemented as a derived class which derives from PairILPGrapheneHBN instead of Pair. That way only modified/optimized functions need to be provided and only additional data members are required in the optimized class. Please see other classes in the OPT and OPENMP package as examples. This should significantly reduce code redundancy and the resulting maintenance effort in case of bugfixes.
2) We have changed how we indicate polymorphism: only the base class function implementation is labeled virtual, all functions that override it in a derived class must be labeled with an override postfix. Again, please see other classes in OPT and OPENMP as examples.
3) only functions that are overridden from the base Pair class should be added as "public", everything else should be "protected".

I have already updated the integration into the GNU make based build system and the manual. So that is done for you, but it also means you should pull from your fork before you continue working.

run-towards-the-future:
Thank you for your updating some files.   
Now, we have made the modification you mentioned above. Please help us check whether the current work meets the merging requirements. 
Thank you very much.

run-towards-the-future:
> @run-towards-the-future thanks for your contribution. Unfortunately, it will need some additional effort before a more thorough review will be done, and it can eventually be merged into the LAMMPS distribution.
> 
> 1. It needs to be implemented as a derived class which derives from PairILPGrapheneHBN instead of Pair. That way only modified/optimized functions need to be provided and only additional data members are required in the optimized class. Please see other classes in the OPT and OPENMP package as examples. This should significantly reduce code redundancy and the resulting maintenance effort in case of bugfixes.
> 2. We have changed how we indicate polymorphism: only the base class function implementation is labeled virtual, all functions that override it in a derived class must be labeled with an override postfix. Again, please see other classes in OPT and OPENMP as examples.
> 3. only functions that are overridden from the base Pair class should be added as "public", everything else should be "protected".
> 
> I have already updated the integration into the GNU make based build system and the manual. So that is done for you, but it also means you should pull from your fork before you continue working.

Thank you for your updating some files.   
Now, we have made the modification you mentioned above. Please help us check whether the current work meets the merging requirements. 
Thank you very much.

akohlmey:
> Now, we have made the modification you mentioned above. Please help us check whether the current work meets the merging requirements

No. What you submitted is not correctly implementing derived classes in C++. For example you left the constructor and destructor unchanged and that leads to all kinds of memory corruption and inconsistent behavior. Because of that unit tests are failing with error messages about double free and corrupted memory from the malloc library.

Please consult your favorite C++ textbooks.

dxhisboy:
It seems after commit `ef51fb7`, the derivation from `PairILPGraphenehBN` is correct. Also, I find that there is no need of neighbor list for ghost atoms (I guess it is because of some historical reason), so I modified the code in `init_style` (commit `b55b9f5`).

Also, an additional comment:
Since the optimization has eliminated the usage of some data structures from `PairILPGraphenehBN`. That might be the initial reason to derive from `Pair`. Meanwhile, most of those data structures are pointers that allocated in `ILP_neigh` and initialized as `nullptr` in `PairILPGraphenehBN`. Thus those pointers does not hurt too much.

akohlmey (approved):

akohlmey:
@dxhisboy @run-towards-the-future thanks for the updates. I made some more cosmetic changes, ran tests with valgrind, and benchmarked the interlayer examples compatible with this optimized pair style. I didn't see any remaining issues and a substantial speedup. When paired with OPENMP package compatible pair styles, there is a bit more speed possible (even with 1 thread) due to optimized code paths in OPENMP using `-suffix hybrid opt omp`.

oywg11:
@run-towards-the-future @dxhisboy Thanks for the contribution! I compared the optimized ILP and the original one, they give the same results and the former gives a 2x speed up.

oywg11 (approved):

PR #3234, opened 2022-04-25
add support for wildcards in variable vectors

sjplimp:
**Summary**

Similar to computes and fixes, allow commands that take inputs from vector-style variables to use a wildcard asterisk character.
This can simplify input syntax if the vector length is large.  E.g. the thermo_style custom command can output all the vector values with one keyword instead of N.

This expands the inputs to encompass the wildcard range via utils::expand_args().  One difference is that vector-style variable lengths are not know until after they are evaluated, and the argument expansion is done much earlier.  So fully general wildcards are not allowed.  The user must specify a trailing number.  Commands that process vector-style variable typically use a zero value if the user has specified a vector index that exceeds the actual length.

**Related Issue(s)**

N/A

**Author(s)**

Steve

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/doc/src/thermo_style.rst
+++ b/doc/src/thermo_style.rst
@@ -85,7 +85,7 @@ Syntax
            f_ID[I] = Ith component of global vector calculated by a fix with ID, I can include wildcard (see below)
            f_ID[I][J] = I,J component of global array calculated by a fix with ID
            v_name = value calculated by an equal-style variable with name
-           v_name[I] = value calculated by a vector-style variable with name
+           v_name[I] = value calculated by a vector-style variable with name, I can include wildcard (see below)
 
 Examples
 """"""""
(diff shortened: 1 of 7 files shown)


sjplimp:
@akohlmey Can you please check the code I added in utils.cpp to the expand_args() method?
Also, here is a small test script that illustrates how the new feature works.

# test of global vector processing

units		lj
atom_style	atomic

lattice		fcc 0.8442
region		box block 0 5 0 5 0 5
create_box	1 box
create_atoms	1 box
mass		1 1.0

velocity	all create 1.44 87287 loop geom

pair_style	lj/cut 2.5
pair_coeff	1 1 1.0 1.0 2.5

neighbor	0.3 bin
neigh_modify	delay 0 every 20 check no

fix		1 all nve

compute         1 all temp
variable        ctemp1 vector C_1

#thermo_style    custom step temp press &
#                v_ctemp1[1] v_ctemp1[2] v_ctemp1[3] v_ctemp1[7] &
#                v_ctemp2[1] v_ctemp2[2] v_ctemp2[3]

thermo_style    custom step temp press v_ctemp1[*6]

thermo          50
run		100

sjplimp:
@athomps Please check this to see if it works as you are expecting

doc/src/thermo_style.rst
```diff
@@ -348,16 +348,16 @@ dimensions *lx*, *ly*, *lz*, *yz*, *xz*, *xy*\ .
...
-numeric values means all indices from 1 to N.  A leading asterisk
-means all indices from 1 to n (inclusive).  A trailing asterisk means
-all indices from n to N (inclusive).  A middle asterisk means all
-indices from m to n (inclusive).
+For output values from a compute or fix or variable, the bracketed
+index I used to index a vector, as in *c_ID[I]* or *f_ID[I]* or
+*v_name[I]*, can be specified using a wildcard asterisk with the index
+to effectively specify multiple values.  This takes the form "\*" or
+"\*n" or "n\*" or "m\*n".  If N = the size of the vector (for *mode* =
+scalar) or the number of columns in the array (for *mode* = vector),
```
athomps, line 356: This seems wrong, should be vector, and the term `mode` is not defined

athomps, line 356: This seems wrong, should be array, and the term `mode` is not defined

doc/src/thermo_style.rst
```diff
@@ -348,16 +348,16 @@ dimensions *lx*, *ly*, *lz*, *yz*, *xz*, *xy*\ .
...
-columns in the array (for *mode* = vector), then an asterisk with no
-numeric values means all indices from 1 to N.  A leading asterisk
-means all indices from 1 to n (inclusive).  A trailing asterisk means
-all indices from n to N (inclusive).  A middle asterisk means all
-indices from m to n (inclusive).
+For output values from a compute or fix or variable, the bracketed
+index I used to index a vector, as in *c_ID[I]* or *f_ID[I]* or
+*v_name[I]*, can be specified using a wildcard asterisk with the index
+to effectively specify multiple values.  This takes the form "\*" or
+"\*n" or "n\*" or "m\*n".  If N = the size of the vector (for *mode* =
```
athomps, line 355: This seems wrong, should be vector, and the term mode is not defined

athomps (approved):
I like it, except for mysterious text in thermo_style.rst that predates this PR. See comment.

doc/src/thermo_style.rst
sjplimp: Thanks for catching that - I just pushed a change that removes the mode disussion.  That was apparently text copied
from fix ave/time, where this a mode option in the command.

akohlmey:
> @akohlmey Can you please check the code I added in utils.cpp to the expand_args() method?

The best way to go about this would be to first write detailed unit tests for the existing functionality and then merge with your changes and check with the old and add new tests. I have started with the first part in PR #3233

akohlmey (approved):

akohlmey:
@sjplimp added the tests and the modified function is behaving as documented. The applied changes are small and straightforward enough.

PR #3242, opened 2022-04-29
Adding saip/metal/opt and ilp/tmd/opt

dxhisboy:
**Summary**

By modification of the chunk code of pair/ilp/graphene/opt, we can now adopt saip/metal/opt and ilp/tmd/opt into its chunk code. With methods described in #3228, both pair style can gain about 2x speedup.

**Author(s)**

Xiaohui Duan sunrise_duan@126.com (National Supercomputing Center in Wuxi, China)
Ping Gao qdgaoping@gmail.com (National Supercomputing Center in Wuxi, China)

**Licensing**

**Backward Compatibility**

There is no changes in the pull request will break backward compatibility for inputs.

**Implementation Notes**

The optimization methods are described in #3228 .
**Post Submission Checklist**

### DIFF
--- a/doc/src/Commands_pair.rst
+++ b/doc/src/Commands_pair.rst
@@ -125,7 +125,7 @@ OPT.
    * :doc:`hbond/dreiding/morse (o) <pair_hbond_dreiding>`
    * :doc:`hdnnp <pair_hdnnp>`
    * :doc:`ilp/graphene/hbn (t) <pair_ilp_graphene_hbn>`
-   * :doc:`ilp/tmd <pair_ilp_tmd>`
+   * :doc:`ilp/tmd (t) <pair_ilp_tmd>`
    * :doc:`kolmogorov/crespi/full <pair_kolmogorov_crespi_full>`
    * :doc:`kolmogorov/crespi/z <pair_kolmogorov_crespi_z>`
    * :doc:`lcbop <pair_lcbop>`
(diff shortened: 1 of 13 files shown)


akohlmey:
@dxhisboy thanks again for your contribution. I can reproduce the reference results and see no issues with coding style (I don't like the multiple inheritance, but - same as for the optimized EAM pair styles - it cannot be avoided in this case). On my desktop machine and with -suffix hybrid opt omp the speedup is closer to 2.5x than to 2x.

sjplimp (approved):

oywg11:
@dxhisboy Thanks for the contribution! I'm surprised that it took you only a few days to add these new optimized pair styles since our last discussion. I can confirm that the optimized codes give the same results as the old ones and the speedup is > 2x.

oywg11 (approved):

dxhisboy:
> @dxhisboy thanks again for your contribution. I can reproduce the reference results and see no issues with coding style (I don't like the multiple inheritance, but - same as for the optimized EAM pair styles - it cannot be avoided in this case). On my desktop machine and with -suffix hybrid opt omp the speedup is closer to 2.5x than to 2x.

My some thoughts in the multiple inheritance: It seems in my implementation, most implementations has been done in `PairILPGraphenehBNOpt`, other variants only need to setup the `variant` variable. Then, it is possible to derive those variants from PairILPGraphenehBNOpt directly, and avoid deriving from the variants.

PR #3244, opened 2022-04-30
Enable use of fix shake or fix rattle during minimization

akohlmey:
**Summary**

This PR changes fix shake (and thus also fix rattle) to be called during minimization. In that case the constraint algorithms are substituted by adding restraint forces to the 1, 2, or 3 atom pairs of the shake clusters to maintain the desired distances.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Post Submission Checklist**


akohlmey:
@sjplimp FYI, this is currently more a proof of concept than a ready-to-use and properly debugged implementation. I had this on my mind a long time but just recently saw how it could be conveniently implemented. We need to talk about it at some point. Not sure this should be in the upcoming stable release.

akohlmey:
@sjplimp this seems to be working as expected now. I still have to do detailed validation with multiple simple test systems for all different variants of shake clusters.

doc/src/fix_shake.rst
```diff
@@ -152,17 +156,26 @@ for.
...
 defines the molecule.  You must use the same *template-ID* that the
 command adding molecules uses.  The coordinates, atom types, special
-bond restrictions, and SHAKE info can be specified in the molecule
-file.  See the :doc:`molecule <molecule>` command for details.  The only
+bond restrictions, and SHAKE info can be specified in the molecule file.
+See the :doc:`molecule <molecule>` command for details.  The only
 settings required to be in this file (by this command) are the SHAKE
 info of atoms in the molecule.
 
+The *kbond* keyword allows to set the restraint force constant when fix
```
sjplimp: ... keyword sets the restraint force constant ...

doc/src/fix_shake.rst
```diff
@@ -152,17 +156,26 @@ for.
...
+shake or fix rattle are used during minimization. In that case the
+constraint algorithms are *not* applied and restraint forces are used
+instead to help maintaining the geometries.  How well the geometries
```
sjplimp: maintain

doc/src/fix_shake.rst
```diff
@@ -152,17 +156,26 @@ for.
...
+are maintained and how quickly a minimization will converge depends on
+the magnitude of the force constant (kbond).  If it is chosen too large
+the minimization may converge slowly.  The default is 1.0e6*k_B.
```
sjplimp: now a different default?

doc/src/fix_shake.rst
```diff
@@ -194,19 +207,25 @@ Restart, fix_modify, output, run start/stop, minimize info
 No information about these fixes is written to :doc:`binary restart
 files <restart>`.
 
+When used during minimization, the SHAKE or RATTLE algorithms are **not**
+applied.  Strong restraint forces are applied instead.
```
sjplimp: See the "note" at the very bottom of the fix addforce doc page about fix_modify energy. If that
is the case for this new feature, the same note should appear in this section of this doc page.

doc/src/fix_shake.rst
```diff
@@ -232,13 +251,14 @@ make a linear molecule rigid.
...
-none
+`fix rigid <fix_rigid>`, `fix ehex <fix_ehex>`,
+`fix nve/manifold/rattle <fix_nve_manifold_rattle>`
 
 
 Default
 """""""
 
-none
+kbond = 1.0e9*k_B
```
sjplimp, line 282: now different?

doc/src/fix_shake.rst
sjplimp: to help maintain the shake/rattle constraints.

doc/src/fix_shake.rst
akohlmey: yes. k_B seems to be scaling somewhat similar than the magnitude of force constants for different units settings.
For `real` units that leads to kbond~=2000 which is ok for the systems I tested so far. Larger values like 5000 or 10000 maintain the geometries tighter, but lead to many more iterations and thus slow down the minimization a lot.
Also with minimizers like "fire" one needs to reduce the timestep with such large restraint forces or you risk lost bond atoms. So a not so tight default is a safer choice.

akohlmey:
This is now confirmed to be fully equivalent to using explicit bonds with the same force constant than what fix shake/rattle use during minimization.

akohlmey:
@sjplimp after the discussion at https://matsci.org/t/peratom-energies-from-fix-addforce/42932
I have decided to revisit the topic of tallying energy and virial during minimization with fix shake and have implemented the equivalent workaround from fix cmap (it looks like we may need to add minimization support to fix cmap, BTW) to have access to the eflag variable.

For consistency, I've also included the virial, but since we have `fix_modify virial no` to turn it off (default is on) a user can choose to not include the restrain virial contributions when minimizing with fix box/relax. But per-atom energies and stress are available and confirmed to be consistent with the global energy and stress.

One yet untested issue is the behavior with KOKKOS. I'll work with @stanmoore1 to figure out the best approach when he is available.

stanmoore1:
@akohlmey do you have a test input script for this?

akohlmey:
> @akohlmey do you have a test input script for this?

I've been mostly using the peptide example with a `minimize 0 0 100 1000`
Added before the run statement.

akohlmey:
@sjplimp this now selectively blocks use of fix shake during minimization with Kokkos. This is still missing but will take time since I have failed figure out how to implement it on my own. After discussing with @stanmoore1 we decided to move forward with merging the PR this way now and add Kokkos support later.

doc/src/fix_shake.rst
```diff
@@ -194,19 +211,31 @@ Restart, fix_modify, output, run start/stop, minimize info
...
 files <restart>`.
 
-The :doc:`fix_modify <fix_modify>` *virial* option is supported by
-these fixes to add the contribution due to the added forces on atoms
-to both the global pressure and per-atom stress of the system via the
-:doc:`compute pressure <compute_pressure>` and :doc:`compute
-stress/atom <compute_stress_atom>` commands.  The former can be
-accessed by :doc:`thermodynamic output <thermo_style>`.  The default
-setting for this fix is :doc:`fix_modify virial yes <fix_modify>`.
+Fix *shake* and *rattle* behave differently during minimization and
```
sjplimp: and -> versus

doc/src/fix_shake.rst
```diff
@@ -152,17 +156,30 @@ for.
...
+algorithms are *not* applied and restraint forces are used instead to
+maintain the geometries similar to the constraints.  How well the
+geometries are maintained and how quickly a minimization converges,
+depends on the force constant *kbond*: larger values will reduce the
+deviation from the desired geometry, but can also lead to slower
+convergence of the minimization or lead to instabilities depending on
+the minimization algorithm requiring to reduce the value of
+:doc:`timestep <timestep>`.  The default value for *kbond* depends on
+the :doc:`units <units>` setting and is 1.0e6*k_B.
+
```
sjplimp, line 187: Add a line about why it is OK to not satisfy SHAKE exactly during a minimization.  I.e. b/c if you do dynamics after min, the dynamics will enforce the SHAKE constraint within a couple of steps (assuming not too far off).

sjplimp (changes requested):
couple of doc page suggestions

doc/src/fix_shake.rst
akohlmey: done

sjplimp (approved):

PR #3248, opened 2022-05-03
Add time option to the "reset_timestep" command and make the "time" property restartable

### LINKED ISSUE #3096: [Feature Request] Implement reset_time command
**Summary**

It would be nice to have a command `reset_time`, similar to `reset_timestep`, which would be able to set simulation time to any value.

**Detailed Description**

Simulation time always starts at 0, even if a restart file is used. This means that dumps which output simulation time also start from zero after a restart which leads to an inconsistency, i.e. multiple dump snapshots with the same (or very similar) simulation time. Adding a command which can set simulation time would solve this issue. An option to store and read simulation time from restart files would also be useful.

N/A

akohlmey:
**Summary**

This adds a `time` option to the `reset_timestep` command that allows to manually set the accumulated simulation time.
Also the current setting is added to binary restart files, so that restarts will continue at the stored time.
The change to restart files is backward compatible, i.e. old restarts can be read without a problem, but the simulation time in that case will be reset to 0.0.

**Related Issue(s)**

This implements and thus closes #3096 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Post Submission Checklist**


akohlmey:
@sjplimp this has the changes we discussed. At the moment it is possible to use NULL as the timestep to skip setting it.
But since this is a new option we could always require the timestep to be set. Then the question of side effects of setting the simulation time only is meaningless, because the restrictions for resetting the timestep apply.

akohlmey:
@sjplimp I have made some tests and the behavior with `reset_timestep NULL time ###` is not intuitive. So I will remove that option. To set the simulation time, it will be *required* to also set the timestep number.

sjplimp:
Is this wrt the operation of dump commands with the dump_modify every/time option?  So that requiring both timestep and time to be reset effectively does not allow use of reset_timestep time when those dumps exist?

akohlmey:
> Is this wrt the operation of dump commands with the dump_modify every/time option?

Yes. Output will happen but at strange multiples that only make sense when you know "time" is computed internally.

> So that requiring both timestep and time to be reset effectively does not allow use of reset_timestep time when those dumps exist?

Yes. It is not really a deficiency, since the primary use is to either reset timesteps *and* time after a minimization or set the time after a read_data (read_restart will work automatically with the changes included here), but in that case you usually also want to reset the timestep if you want to transparently continue from a data file instead of a restart.

akohlmey:
And interesting thought (but something for a future PR) would be to also read the TIME item from dump files, if available, and reset time along with the timestep.

sjplimp (approved):

sjplimp:
Agree with your comments - approved it

PR #3255, opened 2022-05-06
Add a timer-style variable

sjplimp:
**Summary**

Add timer-style variables to the variable class, for use in an input script.  They store the current elapsed CPU each time the
"variable name timer" command is processed in the script.  Thus using two timer-style variables to bracket an operation allows a print statement to output the difference of the 2 variable, to time the operation.  E.g.

variable start timer
other commands
variable stop timer
print "Elapsed time: $(v_stop-v_start:%.6f)"

**Related Issue(s)**

N/A

**Author(s)**

Steve

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

**Post Submission Checklist**


src/variable.cpp
```diff
@@ -500,6 +500,31 @@ void Variable::set(int narg, char **arg)
...
+  // num = 1, for string representation of dvalue, set by retrieve()
+  // dvalue = numeric initialization via platform::cputime()
+
+  } else if (strcmp(arg[1],"timer") == 0) {
+    if (narg != 2) error->all(FLERR,"Illegal variable command");
+    int ivar = find(arg[0]);
+    if (ivar >= 0) {
+      if (style[ivar] != TIMER)
+        error->all(FLERR,"Cannot redefine variable as a different style");
+      dvalue[ivar] = platform::cputime();
```
akohlmey, line 515: @sjplimp I am surprised you are using platform::cputime() here and not platform::walltime() or MPI_Wtime().

sjplimp:
@akohlmey I thought cputime() was the equivalent of MPI_Wtime().  I would have used MPI_Wtime() except I don't see it anywhere in the code anymore.  What I want is something that returns a double in secs and the difference of 2 of them is the elapsed wall-clock time (to high precision).

akohlmey:
> @akohlmey I thought cputime() was the equivalent of MPI_Wtime(). I would have used MPI_Wtime() except I don't see it anywhere in the code anymore. What I want is something that returns a double in secs and the difference of 2 of them is the elapsed wall-clock time (to high precision).

`MPI_Wtime()` has been superseded with `platform::walltime()`. This was an attempt to future-proof walltime measurements and avoid dependence on epoch. With storing the seconds in doubles the precision of the delta will significantly decline in the not so distant future due to the growing absolute value of the time since epoch in seconds as returned by MPI_Wtime(). platform::walltime() uses a high-precision timestamp when the executable was started as reference instead and thus there will be no loss of precision (unless the C++ library implementation of the clock functionality is bad).

Ok. I'll change it then.

akohlmey:
FYI, https://docs.lammps.org/Developer_platform.html#time-functions

akohlmey (approved):

PR #3256, opened 2022-05-06
Delete atoms exactly by fraction or count

### LINKED ISSUE #3249: [Feature Request] Enforce precise fractions/atom counts with `delete_atoms porosity`
The recent closing of #1425 made me realise I missed out a key aspect of the use case I mentioned there. The ability to use the `delete_atoms porosity` command on a group of atoms is extremely useful (thanks for implementing!), but this bit can be a deal breaker for the use case I was looking at:

> There is no guarantee that the exact fraction of atoms will be deleted, or that the same atoms will be deleted when running on different numbers of processors.

When modelling ion migration through Li-ion batteries, we may choose to charge compensate the removal of some fraction of the Li ions with some number of other defects to compensate (Schottky-pairs). If we are unable to ensure that an exact count/fraction of atoms is being deleted, then we are relying on luck to have the simulation be charge neutral after multiple deletion operations. If this constraint isn't met, the simulation obviously crashes.

I appreciate that enforcing this constraint may be slower so would not suggest it be included as default behaviour. Instead, I suggest the inclusion of an optional `precise` flag. When on, a check that the requested fraction leads to an integer number of atoms `n` being deleted should be performed, and then a `select_without_replacement(size(group),  n , seed)` type calculation could be used to both ensure that the deletion operation is consistent across system architectures/processor counts and can be reliably used to generate charge-neutral systems. (apologies for the pseudo-code!)

The absence of this feature usually required me to either pre-generate my inputs outside of lammps (leading to a large number of sizeable input files) which could instead have simply been replaced with the programmatic generation of my system and a single seed variable being passed in to `lmp_serial` to run several calculations of the same type with the same single file.

I appreciate this is achievable already, but the code required to do so can be quite unfriendly for new users and I imagine it is feature which would be commonly used if made available.

Happy to discuss further if helpful!

sjplimp:
**Summary**

The previous behavior of delete_atoms porosity was to delete an approximate number of the requested fraction of atoms.  The option has now been renamed delete_atoms partial and now has 3 options.  One of them is the same before.  Another is to delete the exact fraction of atoms requested.  The third is to delete the exact count of atoms requested.

This is similar to the behavior of the 3 set type/fraction variants though the syntax here is a bit different.  The two commands
(delete_atoms and set) use the same logic for flagging exactly the requested number of atoms.

**Related Issue(s)**

See issue #3249 - this addresses that features request and thus closes #3249

**Author(s)**

Steve

**Licensing**

**Backward Compatibility**

The delete_atoms porosity keyword has changed to partial, as has the syntax that follows the "partial" keyword.  So old
scripts with porosity will break.

**Post Submission Checklist**

### DIFF
--- a/doc/src/delete_atoms.rst
+++ b/doc/src/delete_atoms.rst
@@ -20,14 +20,14 @@ Syntax
          cutoff = delete one atom from pairs of atoms within the cutoff (distance units)
          group1-ID = one atom in pair must be in this group
          group2-ID = other atom in pair must be in this group
-       *partial* args = pstyle pvalue pflag group-ID region-ID seed
+       *partial* args = pstyle value flag group-ID region-ID seed
          pstyle = *fraction* or *count*
            for *fraction*:
-             pvalue = fraction (0.0 to 1.0) of eligible atoms to delete
-             pflag = 0 for approximate deletion, 1 for exact deletion
+             value = fraction (0.0 to 1.0) of eligible atoms to delete
+             flag = no/off for fast approximate deletion, yes/on for exact deletion
            for *count*:
-             pvalue = number of eligible atoms to delete
-             pflag = 0 for warning if count > eligible atoms, 1 for error
+             value = number of atoms to delete
+             flag = no/off for warning if count > eligible atoms, yes/on for error
          group-ID = group within which to perform deletions
          region-ID = region within which to perform deletions
                      or NULL to only impose the group criterion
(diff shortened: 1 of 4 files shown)


sjplimp:
@mjclarke94 - please see if this works as you expect wrt issue #3249

mjclarke94:
Will do! I'm away from my computer for the next week but will have a look when I get back.

sjplimp:
@akohlmey I added some output info when the old-style "delete_atoms porosity .." option is used.  The info explains how to do the equivalent operation with the new "partial" option.  Let me know if you think this is sufficiently clear?

akohlmey:
@sjplimp I think the message was way to detailed. Telling people that the name of the option has changed is sufficient.
I also think that using 0/1 integers when instead of parsing a "logical" is extra effort, so I changed that and also tweaked the error messages and documentation a little bit.

athomps (approved):
I approve as is.  However, I do think that the name `partial` still a bit obscure, albeit much better than `porosity`.  How about`random` ?

akohlmey:
>  I think you mean "no" or "off". Later it says no/off/0 and yes/on/1, so there is a 3rd option?
@sjplimp feel free to change as you see fit. About `utils::logical()` Please see: https://docs.lammps.org/Developer_utils.html#_CPPv4N9LAMMPS_NS5utils7logicalEPKciRKNSt6stringEbP6LAMMPS

Please also note that there are unit tests for the "delete_atoms" command in `unittest/commands/test_delete_atoms.cpp` so any changes to the syntax must be implemented there as well or else the tests will fail and thus block the PR from merging.

sjplimp:
@akohlmey ok - clarified the doc page and update the unit test file - I think this should be ready to merge

akohlmey (approved):

PR #3257, opened 2022-05-06
Add nprocs keyword to thermo output

sjplimp:
**Summary**

Add a "nprocs" keyword to thermo output.  The real motivation is not so much to include it with thermo output, but to make it accessible in input scripts in a variable.  E.g. to use in a filename or print statements or other variable calculations.

**Related Issue(s)**

N/A

**Author(s)**

Steve

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/doc/src/thermo_style.rst
+++ b/doc/src/thermo_style.rst
@@ -20,7 +20,8 @@ Syntax
        *yaml* args = none
        *custom* args = list of keywords
          possible keywords = step, elapsed, elaplong, dt, time,
-                             cpu, tpcpu, spcpu, cpuremain, part, timeremain,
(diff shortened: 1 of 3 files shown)


akohlmey:
@sjplimp this functionality (and then some) already exists. There is the "extract_setting()" special function for variables which is a wrapper for the corresponding library interface function which gives access to a large number of "internal" properties
https://docs.lammps.org/Library_properties.html#_CPPv422lammps_extract_settingPvPKc

We use this to make tests and inputs "self-adapting".

akohlmey:
superseded by #3259

PR #3258, opened 2022-05-09
Create atoms from STL mesh

akohlmey:
**Summary**

This adds a feature to the `create_atoms` command where it will build a set of atoms from a triangular mesh in STL format.
This is meant as an alternative to using regions with fix wall/region. which has some deficiencies. The weak point in this approach is its higher computational cost for high resolution representations.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

This offers two algorithms adjust the resolution of atoms to be placed on the mesh and produce a somewhat uniform distribution and reduce overlap with other mesh atoms.
1. placing an atom in the center of each triangle and do recursive bisectioning along the longest side in case that the average distance of the vertices from the center is large than a threshold
2.  use a quasi-random particle placement to achieve a desired minimum number density

If the atom style supports it, the average vertex distance from center is used as atom radius.
If the atom style supports it, the largest molecule ID is looked up and the new atoms are assigned a new, large molecule ID

Since repositories for STL files (e.g. for 3d printing) are often using binary format STL files, a converter is provided.

**Post Submission Checklist**


akohlmey:
This seems to be working reasonably well and has enough "knobs" to make certain that custom containers "hold water".
However, I am wondering if there are better algorithms to subdividing the triangles to "fill" them with spheres to generate a sufficiently flat uniform surface (if that is always desirable). And whether there should be a more sophisticated implementation with a different atom_style that stores the triangle data *and* their norm and then uses that information to compute the (granular) interaction with the atoms (to be implemented into pair style granular?)?

It looks ok for some example
[mixer.zip](https://github.com/lammps/lammps/files/8648348/mixer.zip) based on in.granregion.mixer (with an object obtained from https://www.thingiverse.com/thing:3497877 )

https://user-images.githubusercontent.com/289251/167328384-3ce320a7-33e4-4003-9b0d-9167d9677fd6.mp4

jtclemm:
This is a cool feature. I need to think about ways to use it in the future.

I took a stab at it and found fairly good results applying this  [quasi-random approach](http://extremelearning.com.au/evenly-distributing-points-in-a-triangle/). Using an equal number of mesh particles, it does a fairly nice job distributing monodisperse particles on the drum (particles rendered with radii of 1.5). However, narrow triangles still contain a lot of overlapping particles.

If triangle-particle interactions were added to the granular pair style, is there a good way in LAMMPS to identify contacts? Or would it still rely on distributing particles on the triangle to build neighbor lists and storing the triangle's geometry with each particle?

Mixer_comparison

akohlmey:
> This is a cool feature.

The main credit goes to the author of the MACHDYN package, who implemented something similar https://docs.lammps.org/fix_smd_wall_surface.html
I came across this recently because of some complaints of one of the static code analysis tools we use.
So I tried to use one of the examples to debug the issue and then visualized it, which looked nice, too.
https://www.youtube.com/shorts/gukDMSefD6k

Plus there was inspiration from a couple of questions on the LAMMPS forum: https://matsci.org/t/generating-a-spherical-shell-with-an-entry-hole/42064 which made me aware of the limitations of nested regions (where LAMMPS will always error out because of inside/outside requirements) which made me suggest to fill those regions with atoms.

> If triangle-particle interactions were added to the granular pair style, is there a good way in LAMMPS to identify contacts? Or would it still rely on distributing particles on the triangle to build neighbor lists and storing the triangle's geometry with each particle?

I have not fully thought this through, but if we also store the norm and vertices with the atom in the center, then you can model the entire triangle with one atom. You would then need to set the particle radius to the largest distance of a vertex from the center. The normal granular neighborlist should include the "mesh particles" for all possible contacts. This can probably further improved because for "large" triangles you would have many neighbors that are far away.

So it may still be better to subdivide the triangles to optimize the neighborlist. By assigning a different atom type to the mesh particles, interactions between themselves can be easily excluded. There will likely be an optimum between adding more atoms and thus having to process more pairs of (valid) interactions and using more subdivisions to reduce the number of non-interacting neighbors. Another open issue is the handling of edges and avoiding to double count those interactions.

If one would be able to identify the triangles surrounding a given triangle (e.g. as bonds), even more complex operations like having smooth, rounded transitions between triangles would be possible in principle...

akohlmey:
> I took a stab at it and found fairly good results applying this [quasi-random approach](http://extremelearning.com.au/evenly-distributing-points-in-a-triangle/). Using an equal number of mesh particles, it does a fairly nice job distributing monodisperse particles on the drum (particles rendered with radii of 1.5). However, narrow triangles still contain a lot of overlapping particles.

@jtclemm If you want me to include your changes, you could submit a pull request to this branch here: https://github.com/akohlmey/lammps/tree/create-atoms-mesh

Perhaps there should be a keyword choosing between subdivision methods?
E.g. for (near) equilateral triangles, it would be simpler/smoother to just divide them into 3 by placing an atom in the center and make replace one of the vertices of the original triangle, but the split-along-the-longest side worked better for irregular meshes in my tests.

dsbolin:
Agreed, this is a great feature, I look forward to using it. I think some of the issues with disparately sized and/or high aspect ratio triangles can also be circumvented by the user by pre-processing/re-meshing the STL file. There are many external, open-source tools to do this. If that's the case, maybe some comments to that end in the doc page would be helpful (i.e. 'avoid high-aspect ratio/disparately sized tris for efficiency reasons'). 

For a large triangle (or a perfectly flat surface represented by several triangles), would you end up with a perfectly regular triangle lattice 'corrugation' of the surface atoms? That's probably fine, but in some applications may be undesirable, as it could cause interlocking/ordering of particles near the surface. The easy fix is to introduce a small random perturbation to the positions of the spheres. Could be done via displace_atoms, or as an optional (in-plane?) perturbation of the spheres in the create_atoms command?

Somewhat related - I have an external bit of Python + C code to 'optimally' fill an arbitrary volume with clustered/overlapping spheres of different size such that the surface is conformed to. It's loosely based on [this paper](https://www.icevirtuallibrary.com/doi/epdf/10.1680/geot.8.T.037), and produces things like this:
image

The volume geometry representation in that case is voxel-based, but the concept and paper I linked are for a watertight STL (although that check is not trivial). The main use case there is for having arbitrarily-shaped particles, where the clustered/overlapping spheres are then assigned to rigid bodies. Could be a future addition to this capability if it's of interest.

dsbolin:
Also, the triangle particle/geometry representation that Steve mentioned is a fairly long story. 
@sjplimp : If it's helpful, I could post that code as a 'Work in Progress' PR along with some of our notes from that time, and folks could comment? It's fairly extensive, and I'm not sure I can commit to working on it in a meaningful way anytime soon, but perhaps some brave soul will be inspired?

sjplimp:
We do "watertight" checks for triangulated surface objects in SPARTA (DSMC particle low-density flow code).  The check is that
all tri edges appear exactly twice in the set of tris and with the 2 vertices in reverse order.  An unmatched edge on the simulation box face is also allowed, i.e. if the object is clipped by the simulation box.

ndtrung81:
@sjplimp @akohlmey To some extent the representation reminds me of the body/rounded/polyhdera in the BODY package where the whole STL mesh is considered a single body: the triangular faces and the edges are listed once in the data file. The interaction between the mesh and the granular particles (now modeled as individual spherical bodies or as rounded polyhedra) are computed as between the bodies. Not sure how relevant it is here, just wanted to chime in my thoughts.

jtclemm:
> @jtclemm If you want me to include your changes, you could submit a pull request to this branch here: https://github.com/akohlmey/lammps/tree/create-atoms-mesh

Sure, I can clean up the implementation and submit a PR. The method would address @dsbolin's comment about regular corrugation on large triangles since the quasi-randomness would prevent interlocking/super-lubricity.

akohlmey:
support for reading binary STL is included now as well (was rather easy).

jtclemm:
I just submitted another PR that fixed a few miscellaneous typos I found in the documentation and in the quasirandom method. 

Not directly relevant, but when running the examples I stumbled upon an inexact error message that I thought I'd mention: 
_Expected floating point parameter instead of 'bond' in input script or data file (../atom_vec_sphere.cpp:65)_
 I hadn't built the MOLECULE package so atom style hybrid didn't recognize bond as a valid style. Since atom style hybrid doesn't check unbuilt packages (as in `utils::check_packages_for_style()`), it just passed bond as an optional argument to atom style sphere.

akohlmey:
> I just submitted another PR that fixed a few miscellaneous typos I found in the documentation and in the quasirandom method.

thanks. already merged it.

> Not directly relevant, but when running the examples I stumbled upon an inexact error message that I thought I'd mention: _Expected floating point parameter instead of 'bond' in input script or data file (../atom_vec_sphere.cpp:65)_ I hadn't built the MOLECULE package so atom style hybrid didn't recognize bond as a valid style. Since atom style hybrid doesn't check unbuilt packages (as in `utils::check_packages_for_style()`), it just passed bond as an optional argument to atom style sphere.

That is an interesting observation. As it turns out, the AtomVecHybrid class was doing a bunch of things that are no longer needed and for which we have better solutions now. So by improving the check, I could also get rid of some ugly "pre-processor magic". Once alerted to some issue, I usually dig deeper to see if the same happens elsewhere, and this is now leading me to some updates for *all* hybrid styles. They did some error prone processing to handle that respective table styles accept arguments but others not and as a consequence unused arguments did not trigger errors. There even were comments that the current solution was inferior. Thanks to proper introspection it will be improved :wink:

jtclemm (approved):

doc/src/create_atoms.rst
```diff
@@ -69,21 +79,23 @@ Examples
...
-collection of atoms (or molecules).  It is an alternative to reading
-in atom coordinates explicitly via a :doc:`read_data <read_data>` or
-:doc:`read_restart <read_restart>` command.  A simulation box must
-already exist, which is typically created via the :doc:`create_box
-<create_box>` command.  Before using this command, a lattice must also
-be defined using the :doc:`lattice <lattice>` command, unless you
-specify the *single* style with units = box or the *random* style.
-For the remainder of this doc page, a created atom or molecule is
-referred to as a "particle".
+either on a lattice, or a single atom (or molecule), or from a triangle
```
sjplimp, line 89: or on a surface defined by a triangulated mesh

doc/src/create_atoms.rst
```diff
@@ -119,6 +131,49 @@ the specified coordinates.  This can be useful for debugging purposes
...
+            :target: _images/marble_race.jpg
+
+For the *mesh* style, a file with a triangle mesh in `STL format
+<https://en.wikipedia.org/wiki/STL_(file_format)>`_ is read and one or
+more particles are placed into the area of each triangle.  The reader
+supports both ASCII and binary files conforming to the format on the
+Wikipedia page.  Binary STL files (e.g. as frequently offered for
+3d-printing) can also be first converted to ASCII for editing with the
+:ref:`stl_bin2txt tool <stlconvert>`.  The use of the *units box* option
+is required. There are two algorithms for placing atoms available:
```
sjplimp, line 146: available for placing atoms

doc/src/create_atoms.rst
```diff
@@ -316,6 +371,12 @@ the atoms around the rotation axis is consistent with the right-hand
 rule: if your right-hand's thumb points along *R*, then your fingers
 wrap around the axis in the direction of rotation.
 
+The *radiusscale* keyword only applies to the *mesh* style and allows to
```
sjplimp, line 374: and adjusts

doc/src/create_atoms.rst
```diff
@@ -316,6 +371,12 @@ the atoms around the rotation axis is consistent with the right-hand
...
+adjust the radius of created particles, provided this is supported by
+the atom style.  Its value is a scaling factor (default: 1.0) that is
```
sjplimp, line 376: does scaling factor mean use a value from 0.0 to 1.0 ?  This should be stated.

doc/src/create_atoms.rst
```diff
@@ -316,6 +371,12 @@ the atoms around the rotation axis is consistent with the right-hand
...
+applied to the radius inferred from the size of the individual triangles
```
sjplimp, line 377: What is the "size" of the triangle?  Its longest or shortest edge length, or something else?

sjplimp: Also, possibly I missed it, but I think advice should be given to turn-off particle interactions between pairs of wall particles, e.g. via neigh_modify exclude, since it is likely there will be overlaps, either for finite-size or point particles.

sjplimp (changes requested):
see doc page comments

doc/src/create_atoms.rst
akohlmey: How the radius is inferred (average of distance from center) is discussed with the bisect algorithm.

sjplimp (approved):

PR #3262, opened 2022-05-11

akohlmey:
**Summary**

Various small changes and fixes

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- make `Output::add_dump()` return a pointer to the created dump
- make "write_dump" use `Output::add_dump()` this avoids segfaults with some dump_modify options
- fix "write_dump" internal dump style creation every option to avoid bogus "fix not computed at compatible times" errors
- fix out-of-bounds access bug with custom keywords for dump style cfg (detected by MSVC only!)
- refactor processing of arguments for hybrid styles to be similar to pair style hybrid
- arguments to bond/angle/dihedral/improper styles that don't use arguments are no longer silently ignored
- fix some minor documentation issues
- add "creator:" tag to YAML dump style
- correctly handle the case that the PyYAML module is not installed in the LAMMPS python module and the unit tests
- request that libyaml-cpp is always build as a static library so it won't need to be installed whan LAMMPS is installed
- request that the MDI library is always built with position independent code
- apply clang-tidy fixes to recently modified code and files that have been previously not included in "tidying"
- update singularity definition files for CUDA 11.6 and ROCm 5.1.2

**Post Submission Checklist**


src/DPD-REACT/random_external_state.h
```diff
@@ -111,7 +111,7 @@ int es_rand(es_RNG_t &state_)
 LAMMPS_INLINE
 double es_drand(es_RNG_t &state_)
 {
-  return 1.0 * es_urand64(state_) / MAX_URAND64;
+  return static_cast<double>(es_urand64(state_) / MAX_URAND64);
```
akohlmey: are you sure this should not be?
```
static_cast<double>(es_urand64(state_)) / MAX_URAND64;
```

stanmoore1: @akohlmey Good point, I tweaked the change a little, trying to fix this warning:

```
../random_external_state.h:114:52: warning: implicit conversion from 'const uint64_t' (aka 'const unsigned long') to 'double' changes value from 18446744073709551614 to 18446744073709551616 [-Wimplicit-const-int-float-conversion]
  return static_cast<double>(es_urand64(state_)) / MAX_URAND64;
                                                 ~ ^~~~~~~~~~~
  ```

stanmoore1: Done--thanks

stanmoore1 (approved):
I approve

sjplimp (approved):

PR #3267, opened 2022-05-15
Added Python wrapper that handles rotations to and from LAMMPS frame

athomps:
**Summary**

Added a Python wrapper for silicon elastic constant calculation that shows how to rotate to and from LAMMPS coordinate frame (upper-triangular cell matrix)

**Related Issue(s)**

None

**Author(s)**

Aidan Thompson, Germain Clavier

**Licensing**

**Backward Compatibility**

**Implementation Notes**

Uses basic linear transformations, including some Numpy code from FitSNAP (fitsnap3/scrapers/scrape.py) 

**Post Submission Checklist**


athomps:
Thanks @Bibobu @charlessievers

examples/ELASTIC_T/BORN_MATRIX/Silicon/elastic.py
```diff
@@ -0,0 +1,113 @@
...
+from __future__ import print_function
+import sys
+from elastic_utils import *
+
+np.set_printoptions(precision = 3, suppress=True)
+
+# setup MPI, if wanted
+me = 0
+# uncomment this if running in parallel via mpi4py
+from mpi4py import MPI
```
akohlmey, line 20: This can be made more portable by using `lammps.extract_setting("world_rank")` and `lammps.extract_setting("world_size")` from the LAMMPS python module which will make loading mpi4py directly redundant and should also work in parallel without mpi4py for as long as no MPI communicator needs to be passed from python to LAMMPS. Please see:
https://docs.lammps.org/Python_module.html#lammps.lammps.extract_setting
https://docs.lammps.org/Library_properties.html#_CPPv422lammps_extract_settingPvPKc

athomps: That's nice to know. I added it.

sjplimp (approved):

sjplimp:
@athomps Possibly a more general form of this script could be added to lammps/tools/python?  E.g. to rotate a box + atoms for an arbitrary triclinic box into a LAMPMS-compatible one?

akohlmey:
> @athomps Possibly a more general form of this script could be added to lammps/tools/python? E.g. to rotate a box + atoms for an arbitrary triclinic box into a LAMPMS-compatible one?

Rather than having a standalone script, I would suggest to add this to the lammps/python/lammps/ folder, e.g. in a "transforms.py" file, so it will become part of the LAMMPS python module. @rbberger had already moved a parser for logfiles there as "formats.py". The motivation behind this would be to eventually make Pizza.py obsolete and instead have reusable functionality updated of re-implemented and included in the LAMMPS module directly and thus avoiding having to maintain two different software packages that are closely connected.
Adding components to the LAMMPS python module would also make it much easier to add unit testing since the infrastructure for that is already in place.

sjplimp:
interesting suggestion for lammps/python/lammps - I didn't mean to say that this new script should be based on Pizza.py which I now see is the case for other scripts in tools/python.  Though it might be a convenient way to output the results of transform.py.  Rather that in the tools dir there are many dirs that contain python scripts related to LAMMPS.  It may be easier for users to find and use some of them if they are standalone scripts.  E.g. there could be a new tools/transform dir with a transform.py in it.

PR #3264, opened 2022-05-12
Add Kokkos version of ML-PACE

stanmoore1:
**Summary**

Add Kokkos version of ML-PACE. This version gives reasonable starting performance on the GPU: a single V100 GPU is about 2x faster than a full Intel Broadwell CPU node (36 cores), but this code is not yet as optimized as the Kokkos version of SNAP on the GPU. Further optimizations such as using shared memory on the GPU will be explored in the future. When running on the CPU, the Kokkos code falls back to the non-Kokkos version (same as SNAP). 

**Related Issue(s)**

None

**Author(s)**

Stan Moore (SNL), with helpful discussions with Evan Weinberg (NVIDIA) @weinbe2, and Yury Lysogorskiy (ICAMS) 
@yury-lysogorskiy

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

Yes

PR #3273, opened 2022-05-18
Initial refactor of view allocation in Kokkos package

stanmoore1:
**Summary**

Initial refactor of view allocation in Kokkos package to use convenient `MemoryKokkos` helper functions. This prevents double allocations when realloc'ing, reducing memory footprint in pair styles that use a lot of memory (SNAP, ReaxFF), and also prevents unnecessary view initialization.

**Related Issue(s)**

Started in #3264.

**Author(s)**

Stan Moore (SNL)

**Licensing**

**Backward Compatibility**

Yes

### DIFF
--- a/src/KOKKOS/memory_kokkos.h
+++ b/src/KOKKOS/memory_kokkos.h
@@ -284,7 +284,6 @@ template <typename TYPE>
 static void realloc_kokkos(TYPE &data, const char *name, int n1)
 {
   data = TYPE();
-  assert(data.use_count() == 0);
   data = TYPE(Kokkos::NoInit(std::string(name)),n1);
 }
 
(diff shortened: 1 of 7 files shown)


athomps (approved):
I approve.

akohlmey:
@stanmoore1 how strongly do you feel about having this merged before the stable vs. after?

I am already a bit concerned about the Kokkos update since it seemed to have more ripples than usual.
If you have some time, we should also discuss a few more things that are on my mind wrt to the stable release and KOKKOS that are probably easier to deal with over phone, or zoom.

stanmoore1:
>@stanmoore1 how strongly do you feel about having this merged before the stable vs. after?

It reduces memory footprint in SNAP and ReaxFF so it would be nice to have in the stable release but I'm not dead set on it. Can continue the discussion offline.

Issue #3265, opened 2022-05-13
[BUG] cmake install of lammps with ML-PACE fails to install libyaml-cpp-pace.so


bernstei:
**Summary**

Following the web site instructions for cmake-based installation of the python package for LAMMPS with ML-PACE active, the module cannot be loaded because it cannot find libyaml-cpp-pace.so

**LAMMPS Version and Platform**

CentOS 7, Intel 64 bit CPU, git latest 04537c9f22b

**Expected Behavior**

`python -c "import lammps; lammps.lammps()"` runs without error

**Actual Behavior**

```
tin 1525 : env LD_LIBRARY_PATH=$HOME/.local/lib64:$LD_LIBRARY_PATH python -c "import lammps; lammps.lammps()"
libpath liblammps.so
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/cluster2/bernstei/.local/lib/python3.8/site-packages/lammps/core.py", line 151, in __init__
    self.lib = CDLL(libpath,RTLD_GLOBAL)
  File "/home/Software/python/anaconda3/lib/python3.8/ctypes/__init__.py", line 373, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: libyaml-cpp-pace.so.0.6: cannot open shared object file: No such file or directory
```

**Steps to Reproduce**

compile lammps with "-D BUILD_SHARED_LIBS=ON  -D PKG_ML-PACE=ON -D LAMMPS_EXCEPTIONS=ON -D PKG_PYTHON=ON", build, then install with `cmake --install .` without root permission so it writes into ~/.local.
Run one liner above.

akohlmey:
@bernstei This is not a LAMMPS issue but due to the PACE source code which we don't maintain. A possible workaround would be to configure and install the PACE library as a standalone (shared) library and then don't use the automatic download through LAMMPS' CMake build.
@yury-lysogorskiy This is an issue of the CMake scripting that is part of the PACE library. To avoid complications with having to install many dependent shared libraries, we build them *always* as STATIC but with CMAKE_POSITION_INDEPENDENT_CODE active so that those static libs are compatible with building a shared  LAMMPS library. The support libraries are then included directly in liblammps.so (as is the PACE library) and problems like this won't show.

akohlmey:
@bernstei After looking at the pace library CMake code for yaml-cpp, there may be an easier workaround.
Please try configuring with -DYAML_BUILD_SHARED_LIBS=off
[...]

PR #3278, opened 2022-05-25
Feature/ml pace multispecies

yury-lysogorskiy:
**Summary**

* Update the version of lib-pace to  v.2021.10.25.fix2 that include:
[replace M_PI with user-defined pi const](https://github.com/ICAMS/lammps-user-pace/commit/0e70385d2c095931998312ee5cf9be77afb0ffca)  to address issue https://github.com/ICAMS/lammps-user-pace/issues/2
* BUGFIX initialization the maximum number of neighbours = 0 to avoid "bad array length" error in certain parallel computation, when no neighbours at all are found

**Author(s)**

Yury Lysogorskiy @yury-lysogorskiy 

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**


yury-lysogorskiy:
Before merge, I would like to bugfix another issue with  `pair_style hybrid`:
Namely, the following input:
```
pair_style  hybrid pace lj/cut 10
pair_coeff  1 1 pace Alyace Al
pair_coeff  1 2 lj/cut 0.006242 3.800 10.0
pair_coeff  2 2 lj/cut 0.01032 3.405 10.0
```
will cause error `ERROR: Incorrect args for pair coefficients (src/pair.cpp:813)`

The possible solution to that (but on older version of LAMMPS) was by using:
```c++
force->bounds(FLERR, type1, atom->ntypes, ilo, ihi);
force->bounds(FLERR, type2, atom->ntypes, jlo, jhi);
```
See here https://github.com/ICAMS/lammps-user-pace/commit/5c2a66dbf6b5768e19d0b3417bf06d556e309f2a?diff=unified#diff-51dc477f3cb68d22704e4381a065e15ccd5384efb4995323c441cf93b8cbcf15R323

But in the current LAMMPS version method `force->bounds` does not exist. 
How this should be solved properly ?

akohlmey:
@yury-lysogorskiy while on the subject of issues with ML-PACE, here are some more items that I have noticed recently:
- in order to fix LAMMPS issue #3265 we need to set CMake variables to change the build options of the embedded yaml-cpp library. This triggers some warnings when running CMake now that can be suppressed by the following change:
```
--- yaml-cpp/CMakeLists.txt	2022-05-25 11:23:07.000000000 -0400
+++ yaml-cpp/CMakeLists.txt	2022-05-25 13:43:40.382473723 -0400
@@ -2,6 +2,13 @@
 cmake_minimum_required(VERSION 3.4)
 project(YAML_CPP_PACE VERSION 0.6.3 LANGUAGES CXX)
 
+# set policy to honor set variables that match option() commands
+# this allows CMake scripts that include this file as a subproject
+# to enforce option variable settings.
+if(POLICY CMP0077)
+  cmake_policy(SET CMP0077 NEW)
+endif()
+
 include(CMakePackageConfigHelpers)
 include(CMakeDependentOption)
 include(CheckCXXCompilerFlag)

```

Furthermore, you are missing an "override" keyword:
```
--- ML-PACE/ace_radial.h	2022-05-25 14:09:48.705465223 -0400
+++ ML-PACE/ace_radial.h	2022-05-25 14:11:09.522531276 -0400
@@ -222,7 +222,7 @@
      * @param radbasename  type of radial basis function \f$ g_k(r) \f$ (default: "ChebExpCos")
      */
     void init(NS_TYPE nradb, LS_TYPE lmax, NS_TYPE nradial, DOUBLE_TYPE deltaSplineBins, SPECIES_TYPE nelements,
-              vector<vector<string>> radbasename);
+              vector<vector<string>> radbasename) override;
 
     /**
      * Destructor

```

I also see a bunch or warnings with compiling with Clang 14 about hiding variable declarations by importing a namespace, about constructor initializer lists listing class members in a different order than they were defined and thus initializing them in a different order and about  implicit copy constructors with a user defined copy assignment operator (which is deprecated in recent C++ standards). Those are too many to attach them here. I compile with: `cmake -S cmake -C cmake/preset/clang.cmake -C preset/cmake/basic.cmake -D PKG_ML-PACE=on -D PKG_KOKKOS=on -B build-pace; cmake --build build-pace`

stanmoore1:
>And you would have to teach the pace pair style to correctly skip NULL type assignments. Please see pair style sw, tersoff and many others for a reference.

PACE already uses `map_element2type` and `map[type[j]]`, so it should work?

akohlmey:
> PACE already uses `map_element2type` and `map[type[j]]`, so it should work?

That is only half of what is needed. It also needs to accept NULL as mapping input in the coeff() function, skip the type to map assignment for that, and keep setflag for that pair at 0.  Right now only known elements are accepted.

stanmoore1:
I don't see that code in Tersoff, looks like that is part of `map_element2type` in the parent `pair.cpp`?

akohlmey:
> I don't see that code in Tersoff, looks like that is part of `map_element2type` in the parent `pair.cpp`?

True, but you don't even get there. The following input:
```
units metal
region box block 0 1 0 1 0 1
create_box 2 box

pair_style      hybrid lj/cut 10 pace product
pair_coeff  * * pace Cu-PBE-core-rep.ace Cu NULL
pair_coeff  * 2 lj/cut 0.01 3.0
```
errors out with:
```
LAMMPS (4 May 2022)
Created orthogonal box = (0 0 0) to (1 1 1)
  1 by 1 by 1 MPI processor grid
ACE version: 2021.10.25
Product evaluator is used
Loading Cu-PBE-core-rep.ace
Total number of basis functions
	Cu: 16 (r=1) 726 (r>1)
Mapping LAMMPS atom type #1(Cu) -> ACE species type #0
ERROR: 'NULL' is not a valid element
 (src/ML-PACE/pair_pace.cpp:330)
Last command: pair_coeff  * * pace Cu-PBE-core-rep.ace Cu NULL
```

akohlmey:
It would be straightforward to adapt the coeff() function to not try mapping any NULL type to an element, but I have no idea what that will do to the code in the pace library because it has calls like the following:
```
aceimpl->ace->element_type_mapping.init(atom->ntypes + 1);
```
So it would seem like it would have to know as well that there are types that need to be ignored.

stanmoore1:
I see, thanks for the info

yury-lysogorskiy:
Sorry for running late... 
Regarding the  dealing with NULL element name, I checked the code implemented in 866391e8305dcea0d2b9f0603971957ec02d03e1 and compare some test hybrid lammps script with our internal  (seems to be non-conventional) implementation and it gave  the same energy results. I assume the comparison test is passed.

@akohlmey Just for my understanding, neighbour list with which element types: both #1 ACE and #2 lj/cut OR only with #1 ACE - will be passed to pair_pace::compute ?

yury-lysogorskiy:
Thank you for clarification. And the selection of the atom types is happening by setting up `map[i]=mu`, where mu>=0 and non-selection by `map[i] =-1` as done here
https://github.com/lammps/lammps/commit/866391e8305dcea0d2b9f0603971957ec02d03e1#diff-c6dda17054293e16f62cf0b3720248a5cf6f99f0f5254de830c64eebd0156262R335 ?

akohlmey:
> Thank you for clarification. And the selection of the atom types is happening by setting up `map[i]=mu`, where mu>=0 and non-selection by `map[i] =-1` as done here [866391e#diff-c6dda17054293e16f62cf0b3720248a5cf6f99f0f5254de830c64eebd0156262R335](https://github.com/lammps/lammps/commit/866391e8305dcea0d2b9f0603971957ec02d03e1#diff-c6dda17054293e16f62cf0b3720248a5cf6f99f0f5254de830c64eebd0156262R335) ?

Yes, although this is in some way redundant. The mapping is already done when you call `Pair::map_element2type()`.
So we could simplify this code a bit further by looping over the "elements" array instead of the arguments (again) and set the value of "map" (again). But since we are very close to a stable release, I didn't want to change that and risk overlooking some corner case considering that I have no knowledge of the internals of the ACE library. This would be a refactor to try in a month or so after the stable release is out.

akohlmey:
@yury-lysogorskiy Please let me do some more checking. When I look at the code again, there may be a mistake that just didn't show by accident because I put the pace pair style first and then the pace element index and LAMMPS atom type are the same.

akohlmey:
Everything is fine. Just made a test by swapping types around so that I need to use `pair_coeff * * pace Cu-PBE-core-rep.ace NULL Cu` instead of `pair_coeff * * pace Cu-PBE-core-rep.ace Cu NULL` but the output is identical. So the map is correct.

yury-lysogorskiy:
@akohlmey I'm planning to fix YAML CMakeLists.txt and ML-PACE/ace_radial.h override, just give me 30min

yury-lysogorskiy:
@akohlmey Ok, I see that this is acceptable numerical instabilities. 

Then I finally push the modification in  2a054c1

akohlmey:
> @akohlmey Ok, I see that this is acceptable numerical instabilities.

Great. I feel much more comfortable now that we have discussed and checked this in detail.
Thanks for helping to sort this out. At times like this, when everybody chips in, maintaining software is the most fun. ;-)

> Then I finally push the modification in [2a054c1](https://github.com/lammps/lammps/commit/2a054c17beacbbe0a16644699b2dcb99165bb98a)

Thanks.

> Regarding clang-14, there is really too much warnings to quick fix in short time. I will take a look on it later...

Please note that this is with a high warning level (-Wextra), so most of those are harmless. But some of them look like the may spell problems with future C++ standard revisions. I would be the most concerned about the "variable shadowing" as this can lead to confusion even though in most cases it is only local and from the context it is clear to the programmer which specific entity of a variable will be used. I've seen similar things in LAMMPS where local variables or function parameters hide class members. I am somewhat obsessive about avoiding such (harmless) compiler warnings because they can lead to not noticing important warnings that indicate a bug.

PR #3284, opened 2022-05-31
Port changes in #3280 to Kokkos

stanmoore1:
**Summary**

Port changes in #3280 to Kokkos
Also fix a small bug in Kokkos ReaxFF

**Related Issue(s)**

#3280 

**Author(s)**

Stan Moore (SNL)

**Licensing**

**Backward Compatibility**

Yes


stanmoore1:

stanmoore1:
I'm seeing what looks like a bug in the pressure computation for the Kokkos version using the test case in #3280.

stanmoore1:
Still debugging the pressure difference, getting closer...

stanmoore1:
Found the bug--I think it should be pretty rare. When it compared equal tags it had bad behavior in one spot.

athomps:
interesting. This tag check does not seem to exist in the non-kokkos code. A similar check exists in `fix_qeq_reax.cpp`, perhaps the some code pattern should be used in both places?

stanmoore1:
@athomps I copied that check from Stillinger-Weber. The non-Kokkos code does `if (j < i) continue` instead. If I recall correctly, using atom tags allows better load balancing when running on multiple procs. They both turn a full neigh list into a half neigh list with newton off, i.e. skip half the owned-owned interactions and duplicate the owned-ghost interactions on both procs.

stanmoore1:
The fix qeq/reax version adds this check as well `if (j < atom->nlocal)`, which turns a full neigh list into a half neigh list with newton _on_.

athomps (approved):
I approve.

PR #3289, opened 2022-06-02
A few last minutes updates and fixes for the next patch release

### LINKED ISSUE #3287: [BUG] Segfault in FixPropertyAtom::grow_arrays
**Summary**

In MPI parallel runs, using `create_atoms` twice in a specific way can cause a segfault in `fix property/atom` 's `grow_arrays`. Although this is where it happened to me, I actually think that something before that point is already off the rails.

`Stable 29 Sep 2021 - Update 3 -- 6fd8b2b1772`
also reproduced with `Unstable 4 May 2022 -- bd4bbbddb`

**Steps to Reproduce**

Run this input with at least 2 mpi processes: `mpirun -np 2 lmp_mpi -i test.lmp`:
[test.lmp.txt](https://github.com/lammps/lammps/files/8825457/test.lmp.txt)

Output will be similar to this (minus 2 debug statements, see below) with a segfault at the last command.
[test.log.txt](https://github.com/lammps/lammps/files/8825468/test.log.txt)

The first `create_atoms` goes through fine, but overestimates how many atoms it will create per process by the `LB_FACTOR`. `AtomVec->roundup` pads to a nearest page size, `FixPropertyAtom::grow_arrays` allocates that 10% too many entries.

The second `create_atoms` does not create atoms on one proc, therefore realizes (`nadd=0`, only `atom->nlocal` counts) that it actually has less atoms than assumed in the last call. In this case, `AtomVec->roundup` also rounds to a *lower* granularity than before. This results in a `grow_arrays` call that is actually shrinking the vectors. An integer underflow finally causes the segfault.

This basically can be triggered whenever a second create_atoms lattice call creates less than 10% of the amount of the first call in a domain (and this is not hidden by lucky rounding).

Since basically no `grow_*` call along this path looks like it expects shrinking, catching this in the fix seems not the correct place?

akohlmey:
**Summary**

A few small fixes and updates.

**Related Issue(s)**

Fixes #3287 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Post Submission Checklist**

### DIFF
--- a/cmake/Modules/Packages/ML-QUIP.cmake
+++ b/cmake/Modules/Packages/ML-QUIP.cmake
@@ -43,6 +43,7 @@ if(DOWNLOAD_QUIP)
   file(WRITE ${CMAKE_BINARY_DIR}/quip.config "${temp}")
 
   message(STATUS "QUIP download via git requested - we will build our own")
+  set(CMAKE_EP_GIT_REMOTE_UPDATE_STRATEGY CHECKOUT)
   # QUIP has no releases (except for a tag marking the end of Python 2 support). We use the current "public" branch
   # The LAMMPS interface wrapper has a compatibility constant that is being checked at runtime.
   include(ExternalProject)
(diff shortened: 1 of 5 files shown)


stanmoore1 (approved):
I approve.

PR #3294, opened 2022-06-07
added set eradius

cipitaua:
**Summary**

added the possibility to set eradius, which was missing

**Author(s)**

Roberto Guerra, roberto.guerra@unimi.it

**Licensing**

**Note**

documentation for set command should be updated


cipitaua:
for `atom_style electron`, the spin cannot be presently set as well. I did not add that because there's a name conflict between magnetic spin (atom->sp) and electronic spin (atom->spin). I'd suggest to use 'espin' for the latter in the set command.

akohlmey:
@cipitaua after some discussion with other LAMMPS developers we have decided to implement this slightly differently and also rename the SPIN package related flags. For the set command it seems more common to use the more verbose keywords. For compute property/atom we stuck with the shorter names (and renamed "spin" to "espin" for consistency. Please see PR #3420

cipitaua:
great, thank you

PR #3298, opened 2022-06-14
Reaxff species 'delete' keyword

jrgissing:
**Summary**

add 'delete' keyword to 'fix reaxff/species', which enables removal of molecules from system based on chemical formula or molecular weight

**Related Issue(s)**

none

**Author(s)**

JG

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/src/REAXFF/fix_reaxff_species.cpp
+++ b/src/REAXFF/fix_reaxff_species.cpp
@@ -219,10 +219,10 @@ FixReaxFFSpecies::FixReaxFFSpecies(LAMMPS *lmp, int narg, char **arg) : Fix(lmp,
             fprintf(fdel,"\t%s",del_species[i].c_str());
           fprintf(fdel,"\n");
           fflush(fdel);
-        } else error->all(FLERR, "Illegal fix reaxff/species command");
+        }
 
         iarg += ndelspec + 4;
-      }
+      } else error->all(FLERR, "Illegal fix reaxff/species command");
 
       // position of molecules
     } else if (strcmp(arg[iarg], "position") == 0) {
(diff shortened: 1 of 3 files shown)


athomps (approved):

stanmoore1 (changes requested):
I think this needs to error out for Kokkos (or be ported).

jrgissing:
it looks like only one major routine (FindMolecule) is overridden by Kokkos, and I do not modify that one. does that imply that it is still kokkosable, or are there other ways for it to break?

stanmoore1:
> it looks like only one major routine (FindMolecule) is overridden by Kokkos, and I do not modify that one. does that imply that it is still kokkosable, or are there other ways for it to break?

Yes good point, I believe this will work with Kokkos as is, thanks for checking

src/REAXFF/fix_reaxff_species.cpp
```diff
@@ -45,7 +47,8 @@ FixReaxFFSpecies::FixReaxFFSpecies(LAMMPS *lmp, int narg, char **arg) : Fix(lmp,
 {
   if (narg < 7) error->all(FLERR, "Illegal fix reaxff/species command");
 
-  force_reneighbor = 0;
+  force_reneighbor = 1;
```
stanmoore1, line 50: Should this be conditional on `delflag`?

jrgissing: Sure, it could be written that way. The reneighboring triggered by molecule deletion is technically redundant to the reneighboring settings already 'enforced' by fix reaxff/species. However, I like the idea of retaining it in the deletion routine for clarity, in case someone copies the code block elsewhere. I use quotations on 'enforced', because I think reneighboring settings can be easily modified by subsequent commands, perhaps rendering the reaxff/species calculation incorrect, but this is not checked for

stanmoore1 (approved):

PR #3305, opened 2022-06-17
Compute grid new

athomps:
**Summary**

Computes sna/grid and sna/grid/local calculate SNAP descriptors for a regular spatial grid of points.

**Related Issue(s)**

The initial purpose of these two computes is to characterize the fine-scale structure of atom configurations, for constructing a machine-learning surrogate finite temperature Kohn-Sham density functional theory.

**Author(s)**

Aidan Thompson, Lenz Fiedler

**Licensing**

**Backward Compatibility**

n/a

**Implementation Notes**

These computes are used by the MALA software via the Python library interface. https://github.com/casus/mala

**Post Submission Checklist**


akohlmey:
@athomps Just had a quick look and this PR is a case where a few programming style updates could be applied easily.
Also I would like to combine the last ~20 commits into a single commit and do a few formatting changes etc. How do you prefer to go about this?

- I can give you inline comments here on github and point to the relevant parts of the documentation and then you do everything
- We can discuss offline via email or phone or zoom and share the work
- You let me do a first pass and then look at what you would like to revert or change in addition

To combine commits a rebase is needed, which will change the history and require a forced update. So if you are still actively working on this branch, it is best, if you do it and nobody else is working on that branch. Then everybody makes a new checkout afterwards (or does a `git fetch` followed by a `git reset --hard` to get back in sync without re-adding the removed commits through a merge).

athomps:
Yeah, the multiple commits to remove files was not very nice.  I should have moved them to a new branch and then removed them before submitting the PR.  This branch is now frozen, so it’s fine if you do the rebase. I would prefer not to attempt it.

athomps:
The programming style changes I am happy to do, if you point them out. Public shaming is fine with me. 😁

akohlmey:
> The programming style changes I am happy to do, if you point them out. Public shaming is fine with me. grin

Ok. Actually, a bunch of formal things (like running "make fix-whitespace; make fix-homepage; make-fix-errordocs; make fix-permissions" in the src folder, some clang-format changes and a little header file hygiene I already did, because it is easy and quick for me). I did the force-push so now please do a `git branch -m compute-grid-new compute-grid-outdated`, `git fetch`, `git checkout compute-grid-new` and then check the commit log with `git log`: it should have one "Delete files" commit attributed to you and then several commits from me.

src/ML-SNAP/compute_sna_grid.cpp
```diff
@@ -0,0 +1,328 @@
...
+  wselfallflag = 0;
+  switchinnerflag = 0;
+  nelements = 1;
+
+  // process required arguments
+
+  memory->create(radelem, ntypes + 1, "sna/grid:radelem");    // offset by 1 to match up with types
+  memory->create(wjelem, ntypes + 1, "sna/grid:wjelem");
+
+  rcutfac = atof(arg[3]);
```
akohlmey: calls to `atof()` should be replaced with calls to utils::numeric(), and `atoi()` with utils::inumeric(), respectively.

src/ML-SNAP/compute_sna_grid.cpp
```diff
@@ -0,0 +1,328 @@
...
+#include "memory.h"
+#include "modify.h"
+#include "neigh_list.h"
+#include "neigh_request.h"
+#include "neighbor.h"
+#include "pair.h"
+#include "sna.h"
+#include "tokenizer.h"
+#include "update.h"
+
```
akohlmey, line 24: The following include files don't seem to be used:
- domain.h
- neighbor.h
- neigh_list.h (see related comment in class definition header)
- neigh_request.h
- pair.h
- tokenizer.h
and thus the include statement should be removed (unless I overlooked something).

src/ML-SNAP/compute_sna_grid.h
```diff
@@ -0,0 +1,56 @@
...
+#include "compute_grid.h"
+
+namespace LAMMPS_NS {
+
+class ComputeSNAGrid : public ComputeGrid {
+ public:
+  ComputeSNAGrid(class LAMMPS *, int, char **);
+  ~ComputeSNAGrid() override;
+  void init() override;
+  void init_list(int, class NeighList *) override;
```
akohlmey: There doesn't seem to be a neighbor list request coming from this compute, so the `init_list()` member function and the list pointer would be obsolete (and the neigh_list.h header not required).

athomps:
That worked.

src/ML-SNAP/compute_sna_grid_local.cpp
```diff
@@ -0,0 +1,316 @@
...
+  delete snaptr;
+
+  if (chemflag) memory->destroy(map);
+}
+
+/* ---------------------------------------------------------------------- */
+
+void ComputeSNAGridLocal::init()
+{
+  int count = 0;
```
akohlmey: We try to minimize direct access to lists of items that are part of other classes (e.g. Modify here).
This check can be replaced with:
```
  if ((modify->get_compute_by_style("snal/grid/local").size() > 1) && (comm->me == 0))
    error->warning(FLERR, "More than one instance of compute orientorder/atom");
```
The `get...by_style()` functions return a vector with the pointers of classes and you can get its size/length directly.
Similar functions already exist for fixes, regions. More to come. You can also request `get...by_id()` and then you get just the one pointer (since ID are unique).

src/compute_grid.cpp
```diff
@@ -0,0 +1,247 @@
...
+/* ---------------------------------------------------------------------- */
+
+ComputeGrid::~ComputeGrid()
+{
+  deallocate();
+}
+
+/* ---------------------------------------------------------------------- */
+
+void ComputeGrid::init() {}
```
akohlmey: If this function is empty it does not need to be provided and can also be removed from the header. It is already defined in the Compute base class.

akohlmey (changes requested):
Here are some suggestions for improvements. Since the grid and grid/local versions are quite similar, I usually have added comments only to one of them. They usually apply to both (except when they don't  :wink:).

akohlmey:
@athomps Finally, a word of praise. The example inputs came out valgrind clean. Well done!!

akohlmey:
> That worked.

Excellent! You can switch back to the compute-grid-outdated branch in case you suspect something is wrong or got broken, but at the end - after everything is done and fixed - you can delete it with `git branch -D compute-grid-outdated` (not the capital -D since this branch has become a "dead end" and is going to stay unmerged).

src/ML-SNAP/compute_sna_grid.cpp
athomps: > calls to `atof()` should be replaced with calls to utils::numeric(), and `atoi()` with utils::inumeric(), respectively.

Done. Also done for the 4 other SNAP compute, now they are all the same. I will submit those changes in a separate PR, I think.

src/ML-SNAP/compute_sna_grid.cpp
athomps: Done.

src/ML-SNAP/compute_sna_grid.h
athomps: Done

src/ML-SNAP/compute_sna_grid.cpp
akohlmey: They would fit in here just fine.

src/ML-SNAP/compute_sna_grid_local.cpp
athomps: I made this change. The only problem is, `sna/grid` matches `sna/grid/local`, not vice versa, so there is one false positive warning in the two examples. I assume this is related to matching style suffixes like `/kk`.

src/compute_grid.cpp
athomps: Done

src/ML-SNAP/compute_sna_grid_local.cpp
akohlmey: You can request an exact match by using `^sna/grid$` instead of `sna/grid` and similarly `^sna/grid/local$`. If there will be future versions with suffixes, there can be more complex regular expressions (although this is a minimal matcher that does not support the entire standard regex syntax).

athomps: That worked

athomps:
I'm all finished with my edits.  I will ask @sjplimp for review, as he may have some opinions about the over all `ComputeGrid` feature.

akohlmey:
> I'm all finished with my edits. I will ask @sjplimp for review, as he may have some opinions about the over all `ComputeGrid` feature.

I saw a way to get rid of the "SNAPCOMPUTENAME" defines by just using the "style" variable (errors can insert variables now) and re-ran clang-format one more time.

athomps:
@sjplimp I have addressed your suggestions.

sjplimp (approved):

akohlmey (approved):

PR #3307, opened 2022-06-17
Fix mdi aimd enhance

sjplimp:
**Summary**

Rename fix mdi/aimd to fix mdi/qm and make it more general.  It can now be used for AIMD, but also for triggering
quantum calculations every N steps during an MD run or processing a series of independent configurations passed 
to a quantum engine by LAMMPS.  Both the latter use cases are useful in a machine learning context.  New example scripts 
that illustrate these use cases were added to the examples dir.  This also required some new options be added to the "mdi" command.

**Related Issue(s)**

N/A

**Author(s)**

Steve

**Licensing**

**Backward Compatibility**

Command name was changed from fix mdi/aimd to fix mdi/qm

**Implementation Notes**

**Post Submission Checklist**


sjplimp:
@akohlmey I think is ready for you to look over and merge (post-stable).  Two last changes I made in lib/mdi.  One to include Python support in the MDI lib, by removing -Clanguage flags from the makefiles.  Two to use the current version of the MDI lib in Install.py.  That needs a checksum entry added I think.

sjplimp:
@akohlmey turned back on the WIP label, b/c I remembered another MDI feature I want to add

sjplimp:
@akohlmey @taylor-a-barnes ok, I think this PR is now ready to test and merge

sjplimp:
@akohlmey the checksum in lib/mdi/Install.py for the new version still needs to be added

akohlmey:
> @akohlmey @taylor-a-barnes ok, I think this PR is now ready to test and merge

@sjplimp @taylor-a-barnes I need some clarification about python support in the MDI library and LAMMPS.
When compiling the MDI library for LAMMPS from automatic download (i.e. not a systemwide installation), we prefer to compile it as a static library with position-independent-code (-fpic or equivalent) enabled. That simplifies installation and usage massively, since we only need to worry about having liblammps.so and the LAMMPS executable available and not worry about any add-on libraries.

If I want to configure/compile the MDI library, do I then need to enable Python support? or is enabling python plugins sufficient since the MDI library and headers are otherwise not installed as part of LAMMPS?

Right now we have a discrepancy about how things are compiled when LAMMPS is configured with CMake (static with C language support only and python plugins only if python is found) and GNU Make (shared with all languages enabled, if found).

I would very much like to have both compilation methods consistent.

taylor-a-barnes:
@akohlmey When the MDI is compiled with Python language support (which happens by default or by setting -Dlanguage=Python), it must be compiled as a shared library.  Python language support is only needed if you want to be able to do `import mdi` from a Python script.  LAMMPS doesn't actually need to do this, except in some of the MDI tests/example drivers.  Generally, if someone wants to use MDI as a Python package, I would recommend using conda or pip to install `pymdi` rather than manually compiling with Python language support from CMake.  For this reason, I think you would be fine sticking with a static build of MDI in all use cases, and tell users to separately install MDI as a Python package if they want to use the LAMMPS *.py MDI test scripts.

Enabling support for Python plugins is separate from Python language support.  The Python plugin support allows LAMMPS, when running as a driver, to use Python scripts as plugins.  You can compile MDI as a static library and still enable Python plugin support.

sjplimp:
@taylor-a-barnes Thanks for the explanation.  Since, as you say, the provided examples use Python, can you email me some instructions about installing pymdi, to include in the examples/mdi/README?  Ditto for the Python plugin support.   Also, since LAMMPS will use a specific version of MDI (set in the lib/mdi/Install.py file), how will the average user insure the MDI they get via pymdi, matches what LAMMPS is using?  Presumably that could cause issues for an actively developed code like MDI.

akohlmey:
> For this reason, I think you would be fine sticking with a static build of MDI in all use cases, and tell users to separately install MDI as a Python package if they want to use the LAMMPS *.py MDI test scripts.

Come to think of it, if LAMMPS itself is built as a shared library, one could add a symbolic link to liblammps.so to create a fake mdi lib and then import that into Python. Since it contains the static lib, it should work just as well. :wink:

sjplimp:
@akohlmey Please see the info from @taylor-a-barnes about Python usage that I added to the top of the examples/mdi/README file.  If you want to amend it with LAMMPS lib info, please do.

akohlmey:
@sjplimp this is now working for me and fully converted to compile/link with a static MDI lib and plugins - including python plugins - enabled. Please check if building from scratch and running still works for you or email me the failing commands and error messages

sjplimp:
@akohlmey My build and all the examples, including Python drivers, worked fine.  Ready to merge from my perspective.

akohlmey (approved):

PR #3314, opened 2022-06-25
Fortran karl

hammondkd:
**Summary**

Fixed one bug in fortran/lammps.f90: a completely empty LAMMPS constructor would result in warning messages about non-NULL pointers. lammps_open_no_mpi should have had "handle" passed as VALUE, not INTENT(OUT); the Fortran standard allows intent(out) arguments (including pointers) to be assigned arbitrary values at call time.

I also implemented alternative constructors that differentiate between MPI_F08's MPI_Comm and the MPI module's integer communicators and constructors that accept command-line strings instead of arrays of strings (e.g., "lmp -l log.special" instead of something like [character(len=11) :: "lmp","-l","log.special"]. This should allows users to access command line arguments the same way they would in the C interface.

Updates to the documentation are also included.

[I will not be put out if some or all of these changes are rejected, but I thought I'd share them.]

**Author**

Karl Hammmond, University of Missouri
hammondkd@missouri.edu

**Licensing**

**Backward Compatibility**

This should not break backward compatibility, with one exception: someone's compiler would have to have the MPI_F08 module, or they would need to provide a dummy module with a dummy MPI_comm type in it.

The changes should preserve the use of "use MPI" without the user noticing at all. I also ran a check on "include 'mpif.h'" and that seems to work, too, though obviously we are under no compulsion to support something the MPI standard has deprecated.

**Implementation Notes**

I tested this with a program (admittedly a very simple one) that called all five different variants on the LAMMPS constructor [i.e., lammps(args, comm), lammps(args), lammps(), lammps(str,comm), and lammps(str)] using the GNU, Intel, and Cray compilers at NERSC; all gave reasonable values.

I do get a warning using "-Wall" with gfortran that "args(:)" (an allocatable array) is unassigned, but that appears to be a bug in gfortran rather than a bug in this code.

**Post Submission Checklist**

### DIFF
--- a/doc/src/Fortran.rst
+++ b/doc/src/Fortran.rst
@@ -38,7 +38,7 @@ found together with equivalent examples in C and C++ in the
 
 .. note::
 
-   A contributed (and complete!) Fortran interface that more
+   A contributed (and more complete!) Fortran interface that more
    closely resembles the C-library interface is available
    in the ``examples/COUPLE/fortran2`` folder.  Please see the
    ``README`` file in that folder for more information about it
@@ -65,6 +65,7 @@ the optional logical argument set to ``.true.``. Here is a simple example:
 
    PROGRAM testlib
      USE LIBLAMMPS                 ! include the LAMMPS library interface
+     IMPLICIT NONE
      TYPE(lammps)     :: lmp       ! derived type to hold LAMMPS instance
      CHARACTER(len=*), DIMENSION(*), PARAMETER :: args = &
          [ CHARACTER(len=12) :: 'liblammps', '-log', 'none' ]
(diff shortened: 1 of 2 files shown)


akohlmey:
@hammondkd thanks for your submission with bug fixes and suggested features.
I have incorporated the bugfixes (well spotted!!) and the one example for passing command line flags into a WIP branch of mine.

I don't see much of a need to pass the argument list as a single string which also requires some nasty looking code to convert it to an array. If there ever is a convincing use case, I think that feature would be better implemented in the C library interface so that it is available from C and Python, too. And we already have a well tested utility function to split a command line, that also can handle quoted strings. So that would avoid code duplication and make the behavior more consistent across programming languages.

My thoughts about a Fortran 08 and later compatible interface (as may be needed by very new Fortran compilers like flang) were more along the lines of having a full binary module that is compiled with LAMMPS and installed like the mpi_f08 module.
The plan for `fortran/lammps.f90` on the other hand is to have everything needed in a single source file that can be included in the compilation of any project; and not just on the command line or in a make file as an additional file, but also with:
```
include 'lammps.f90'

PROGRAM testlib
  USE LIBLAMMPS                 ! include the LAMMPS library interface
  IMPLICIT NONE
  TYPE(lammps)     :: lmp       ! derived type to hold LAMMPS instance
[...]
```
I have to think about allowing to pass the mpi_f08 style communicator. I would have to be done in a different file since I don't want to make Fortran 2008 a requirement but stick with Fortran 2003.

Thanks again for your help and suggestions. It is much appreciated.

PR #3316, opened 2022-06-28
Add Stillinger-Weber Three-body loop optional keyword argument

### LINKED ISSUE #3297: [Feature Request] Two-Body-Only Stillinger-Weber Potential
**Summary**

Create a new pair style for a two-body-only Stillinger-Weber potential to avoid the slow down of calculating a three-body term which will always equal zero. 

**Detailed Description**

My group has a several models which use the Stillinger-Weber potential with `lambda=0`. Some examples include a water-methane model (1), a coarse grain model of silica (2), and some solvent-tetramethylammonium models (3). 

These models have a significant number of two-body-only SW interactions. They run slower than necessary due to `sw` computing a three-body term which will always equal zero. 

In order to speed things up, we created an in-house version of `sw` called `sw0`, which removes the entirety of the three-body loop. This results in significant speed ups of ~8x for highly two-body simulations.

We had also tested a one-line change to the `sw` style to try to speed up the three-body term without writing a whole new pair style. This line was added near the top of the three-body loop of `sw` to skip unnecessary calculations:
```cpp
if params[ijkparam].lambda == 0 continue;
```
This modification resulted in a ~5x speed improvement for highly two-body simulations over `sw`. It also caused a ~1.5% speed *decrease* in normal three-body Stilllinger-Weber simulations over `sw`. We considered these performance decreases over `sw0` to be unacceptable for our uses. 

For this reason, I'd like to submit a new pair style, `sw/2b`, which will compute only the two-body term of the Stillinger-Weber potential. 

(1) J. Phys. Chem. B 2010, 114, 21, 7302–7311
(2) J. Phys. Chem. C 2021, 125, 48, 26857–26868
(3) J. Chem. Theory Comput. 2017, 13, 1, 245–264

jkelowitt:
**Summary**

Adds a new keyword argument to the Stillinger-Weber potential. Using `pair_style sw threebody off` will skip the whole three body loop of the potential. This can be used to drastically increase the speed of interactions where lambda = 0. 

For example, this potential benefits massively from utilizing this kwarg:
```
#           eps    sig     a     *lam*   gam   cos          A            B            p  q    tol
mW  mW  mW  6.189  2.3925  1.8   23.15   1.2  -0.333333333  7.049556277  0.602224558  4  0.0  0.0
xL  xL  xL  0.34   4.5     1.8   0.0     1.2  -0.333333333  7.049556277  0.602224558  4  0.0  0.0
mW  xL  xL  0.8    4.5     1.8   0.0     1.2  -0.333333333  7.049556277  0.602224558  4  0.0  0.0
xL  mW  mW  0.8    4.5     1.8   0.0     1.2  -0.333333333  7.049556277  0.602224558  4  0.0  0.0
mW  mW  xL  0.8    4.5     1.8   0.0     1.2  -0.333333333  7.049556277  0.602224558  4  0.0  0.0
mW  xL  mW  0.8    4.5     1.8   0.0     1.2  -0.333333333  7.049556277  0.602224558  4  0.0  0.0
xL  mW  xL  0.8    4.5     1.8   0.0     1.2  -0.333333333  7.049556277  0.602224558  4  0.0  0.0
xL  xL  mW  0.8    4.5     1.8   0.0     1.2  -0.333333333  7.049556277  0.602224558  4  0.0  0.0
```

With 960 mW particles, and 80 xL particles, there is about a 5x speed improvement using:

```
pair_style hybrid sw sw threebody off
pair_coeff  * *  sw  1  mW_xL.sw  mW NULL
pair_coeff  1 2  sw  2  mW_xL.sw  mW xL
pair_coeff  2 2  sw  2  mW_xL.sw  mW xL
```
over the default SW:
```
pair_style sw
pair_coeff  * *  mW_xL.sw  mW xL
```

**Related Issue(s)**

Closes #3297

**Author(s)**

Jackson Elowitt, University of Utah, jkelowitt@protonmail.com

**Licensing**

**Backward Compatibility**

No issues that I'm aware of.

**Implementation Notes**

This adds a boolean flag which is set in the settings function. In compute, the three body loop will be skipped if this flag is set to false. If the three body interactions are set to 'off', the one_coeff flag is set to 0, to allow for multiple pair_coeff commands. 

I compared the above potential file against both the stock 'sw' and the mixed 'threebody' coefficients. While they don't equal step-for-step, the resulting energies, pressure, volume, enthalpy, etc. were equivalent. I have not compared across accelerated styles, only the default styles.

The sw pair style has a multitude of accelerated styles. As of writing, users who use both 'threebody on/off' and acceleration will get unexpected "Illegal pair_style command" errors if trying to use `threebody on/off`. 

**Post Submission Checklist**

Here are a few papers which utilize a three-body-skipping sw potential. This PR is a modification of the SW potential, rather than the separate potential (sw0) that was used in the papers below.

(1) J. Phys. Chem. B 2010, 114, 21, 7302–7311
(2) J. Phys. Chem. C 2021, 125, 48, 26857–26868
(3) J. Chem. Theory Comput. 2017, 13, 1, 245–264


src/MANYBODY/pair_sw.h
```diff
@@ -52,6 +52,7 @@ class PairSW : public Pair {
   Param *params;      // parameter set for an I-J-K interaction
   int maxshort;       // size of short neighbor list array
   int *neighshort;    // short neighbor list array
+  bool threebody_on;  // whether to run threebody loop
```
akohlmey: Please name this variable `threebody_flag`.

doc/src/pair_sw.rst
```diff
@@ -24,13 +24,17 @@ Syntax
...
+* keyword = *maxdelcs* or *threebody*
 
   .. parsed-literal::
 
        *maxdelcs* value = delta1 delta2 (optional)
          delta1 = The minimum thershold for the variation of cosine of three-body angle
          delta2 = The maximum threshold for the variation of cosine of three-body angle
+       *threebody* value = *on* or *off* (optional)
+         on (default) = Compute both the three-body and two-body terms of the potential
+         off = Compute both only the two-body terms of the potential
```
akohlmey: "Compute both only" -> "Compute only" and "terms" -> "term"

doc/src/pair_sw.rst
```diff
@@ -44,6 +48,10 @@ Examples
    pair_style sw/mod maxdelcs 0.25 0.35
    pair_coeff * * tmd.sw.mod Mo S S
 
+   pair_style hybrid sw sw threebody off
+   pair_coeff * * mW_xL.sw mW NULL
+   pair_coeff * 2 mW_xL.sw mW xL
+
```
akohlmey, line 54: This example is using incorrect syntax. Please see my post at issue #3297 instead.

doc/src/pair_sw.rst
```diff
@@ -141,6 +152,12 @@ potential is used as part of the *hybrid* pair style.  The NULL values
 are placeholders for atom types that will be used with other
 potentials.
 
+.. note::
+
+   When the *threebody on* keyword is used, multiple pair_coeff commands may
```
akohlmey: "threebody on" -> "threebody off"

doc/src/pair_sw.rst
```diff
@@ -111,6 +119,9 @@ This value enables the cut-off function to exclude unnecessary angles in the thr
    However, the angle variation is much smaller than the given threshold value for actual simulations,
    so the inconsistency between potential and force can be neglected in actual simulations.
 
+The *threebody* keyword determines whether or not the three-body term of the potential is calculated. Skipping this
+significantly increases the speed of the potential, but is only applicable when :math:`\lambda_{ijk} = 0`
```
akohlmey: "speed of the potential" -> "speed of the calculation".

Writing "but is only applicable..." implies that this is a check that only allows this flag only if `lambda_ijk` is actually zero. This check does not exist. So either you would add the check or the documentation should say "and is equivalent to setting :math:\`\lambda_{ijk}\` to zero".

src/MANYBODY/pair_sw.cpp
```diff
@@ -229,9 +230,26 @@ void PairSW::allocate()
...
-  if (narg != 0) error->all(FLERR,"Illegal pair_style command");
+  // Default
+  threebody_on = true;
+
+  int iarg = 0;
+
+  while (iarg < narg) {
+    if (strcmp(arg[iarg],"threebody") == 0) {
+      if (iarg+2 > narg) error->all(FLERR,"Illegal pair_style command");
+      if (strcmp(arg[iarg+1], "on") == 0) {
```
akohlmey: Please use [`utils::logical()`](https://docs.lammps.org/Developer_utils.html#_CPPv4N9LAMMPS_NS5utils7logicalEPKciRKNSt6stringEbP6LAMMPS) to parse arguments with truth values.

akohlmey (changes requested):
Please note, that this change also affects pair styles sw/angle/table, sw/gpu, sw/intel, and sw/kokkos which all derive from pair style sw. Either there needs to be a check disallowing the use of the flag, or when parsing the potential file, lambda_ijk must be forcibly set to zero for the affected pairs/triples. ...and this needs to be documented.

doc/src/pair_sw.rst
jkelowitt: > "and is equivalent to setting :math:`\lambda_{ijk}` to zero".

I feel like this implies that you get the same speed increase from setting lambda to zero as you would from using the keyword. 

I think I can sidestep this while knocking out the issue with the derived potentials you mentioned. I can set lambda to zero if `threebody_flag` is false, which should standardize the expected interaction across the derived classes. Although a it should be documented that `threebody on/off` won't work for sw/mod because it overrides `settings()`.

With that in place, I'd be able to write something like: "The *threebody* keyword determines whether or not the three-body term of the potential is calculated. Skipping this significantly increases the speed of the calculation, with the tradeoff that :math:`\lambda_{ijk}` is forcibly set to 0. If the keyword is used with the pair styles sw/angle/table, sw/gpu, sw/intel, or sw/kokkos, :math:`\lambda_{ijk}` will be forcibly set to 0, but no benefits will be gained. This keyword cannot be used for variants of sw/mod."

src/MANYBODY/pair_sw.h
jkelowitt: Resolved by 4a2182a

src/MANYBODY/pair_sw.cpp
jkelowitt: That makes things much cleaner, thank you. 

Resolved by 4db7f91

doc/src/pair_sw.rst
jkelowitt: Resolved by e8c2dcc

doc/src/pair_sw.rst
jkelowitt: Resolved by e8c2dcc

doc/src/pair_sw.rst
jkelowitt: Resolved by 9a05cd3

athomps (changes requested):
The example provided in the PR is confusing, because lambda is not zero.  Why would you want to get the wrong answer 5x faster? The reasons for using this option (and not using it) need to be explained clearly. 
[edit] I see now that the answer to the mystery is hidden in the rather strange usage of pair hybrid command.  The complexity of pair coeff commands is very awkward.  This could much better be handled behind the scenes in the C++ code, when  `threebody off` is specified.  For example, any atom that has lambda=0, the threebody part could be skipped. This could even be done silently without every exposing the `threebody off` option to the user. Alternatively, `threebody 1 0` could be used to select which types get threebody. There are many better ways to implement this performance optimization.

jkelowitt:
> any atom that has lambda=0, the threebody part could be skipped

To skip interaction in this manner, you need to do so within the three-body loop. The problem is that skipping interactions while inside the three-body loop is ~1.7x slower than skipping interactions from outside the three-body loop. 

> Alternatively, threebody 1 0 could be used to select which types get threebody. 

With this, you would be able to skip the loop from outside, but you wouldn't be able to skip the optimal number of interactions. For example, lets say we used `threebody 1 0`. This syntax would only allow us to skip one set of interactions, `xL xL xL`, because we've activated the three-body term for mW, which is present in every other interaction, and need to check every interaction. I haven't tested this, but I'd expect it to be even slower than the in-loop method. 

To get the most out of this optimization, it needs to skip all lambda==0 interactions, *and* do that from outside of the three-body loop. The two methods that I know of to do that are the keyword method in this PR, or to separate the optimization out into a new pair style. The keyword is awkward, but the separate pair style requires more maintenance in the long term. With the speed being equal, which matters more: clarity of use, or ease of maintenance?

athomps:
These are good points. Because lambda is typed not just on central I, but also neighbor atoms J/K, you can't just skip the loop entirely if some but not all lambda's are non-zero.  So I accept the current strategy of splitting the potential into parts for which lambda's are all zero and a part where all lambda's are non-zero.

stanmoore1:
@jkelowitt have you looked at the overhead of the if-test you added? A template parameter might be a little faster (if that is permissible with the other developers).

athomps:
My remaining complaint is that this change makes the largest and most important part of SW code explicitly conditional scope on `threebody_flag`, as if it is a special option, even though this condition is almost always satisfied. This misrepresents the intent of the code.  The intent would be clearer if you replaced `threebody_flag` with `skip_threebody_flag` and then put a small piece of conditional code that would have the effect of skipping the threebody calculation e.g. `if (skip_threebody_flag) jnumm1 = 0`

athomps (approved):
I approve. Thanks!

stanmoore1 (changes requested):
As @akohlmey said, these changes need to be ported to the SW suffix styles (OMP, Kokkos, etc) or it needs to error out for an unsupported keyword.

doc/src/pair_sw.rst
jkelowitt: This is partially solved by 419460f and d7c4a49. 

There is a weird interaction I'm not sure how I want to handle. sw/angle/table has its own read_file() function, which means it doesn't inherit my forcing lambda to be zero when the threebody flag is set. sw/angle/table does inherit the settings() function, however. 

This means that sw/angle/table will accept the threebody flag, but do nothing with it. I'm not sure how this should be handled, if I should just document it and call it a day, make the change to the sw/angle/table style to force lambda to be zero for consistency, or override the base sw settings function to reject the threebody keyword.

akohlmey: It makes no sense to turn threebody interactions off for pair style sw/angle/table since the whole point of the pair style is to *replace* the threebody term with a tabulated function (so changing the lambda values has no effect).

akohlmey:
> I approve. Thanks!

I disagree. This still needs work. I particularly dislike the "double negative" logic of the keyword (negating the internal flag variable is fine with me, but for the user it is weird to say "no" to turn a feature *on*).

There also is a need for adding unit tests and address current compilation failures etc. I will take a closer look tomorrow. 
@jkelowitt thank you so much for all your efforts to implement this. It is a big help that you got it this far with all the comments and requests. It looks like we're getting to a point now where some additional LAMMPS development experience is needed and one of the core developers needs to take care of making everything consistent and make sure that it remains backward compatible.
Also, for consistency, there should now be a `Pair::single()` function added and enabled when the threebody term is disabled. Again, that is something I will look into.

doc/src/pair_sw.rst
jkelowitt: Sounds good, I'll add a settings function to sw/angle/table to make it error out instead of blindly accepting a useless keyword.

jkelowitt:
@akohlmey Thank you for your help!

> it is weird to say "no" to turn a feature on

This was changed in 454c39a so that the keyword would match the internal flag. The keyword was changed to `skip_threebody on`, which avoids the double negative.

doc/src/pair_sw.rst
akohlmey: I have "massaged" the documentation a bit more to make it more succinct. Please check it out.

akohlmey:
> This was changed in [454c39a](https://github.com/lammps/lammps/commit/454c39ae21db4cdc49e962dd87b44fd6cd2fd5ca) so that the keyword would match the internal flag. The keyword was changed to `skip_threebody on`, which avoids the double negative.

What is meaningful for a developer when looking at the source code may not be useful to a user, especially since the majority of LAMMPS users these days do not look at the source code. Thus I restored the name of the keyword to be just "threebody" with the default to "on", which makes more sense to a (casual) user IMO.

akohlmey:
@jkelowitt @athomps @stanmoore1 I have checked and reviewed the source code in detail and ran some tests for a variety of scenarios.
I am now getting consistent results with threebody terms turned either on or off and also the hybrid setup with using both settings at the same time.

I also implemented an additional (significant) optimization of requesting only a half neighbor list for the "threebody off" case.
The code for the GPU, INTEL, KOKKOS package - as far as I was able to test it - produces the same results, but does not have the loop skipping and neighbor list optimization because of the different way the force kernels are written for those styles.
Perhaps somebody has an idea in the future to add an optimization there, too.
In addition, I implemented and enabled the `Pair::single()` function for the "threebody off" case and added yaml files for the force-style unit tests.

akohlmey (approved):

akohlmey:
I now have also tested with KOKKOS on a GPU and also saw correct and consistent results. Even for a hybrid setup with either threebody off on the CPU and regular sw on the GPU or both variants on the GPU (which is slower). I cannot think of anything else to test...

stanmoore1 (approved):
I approve

sjplimp (approved):
Aidan should have the final say on this one

athomps (approved):
I approve.

PR #3319, opened 2022-06-28
Boolean bug fix

### LINKED ISSUE #3318: [BUG] _Replace With Suitable Title_
**Summary**

Undefined behavior when Boolean expression in `if` command is single non-numeric text string, no spaces, no quotes
 
Mac OS

**Expected Behavior**

Exit with error message `ERROR: Invalid Boolean syntax in if command` or `ERROR: Invalid syntax in variable formula`

**Actual Behavior**

Evaluates as either TRUE or FALSE, I have seen both on consecutive tries

**Steps to Reproduce**

./lmp
if (False) then 'print "True"'

sjplimp:
**Summary**

Fix corner case with if-then-else Boolean logic where there is just a single string argument.  Also remove doc
page references to comparing strings with Boolean operaters like "<" or ">=".  This is not supported in the code.
The "==" and "!=" operators are supported for strings or numbers.

**Related Issue(s)**

Closes #3318

**Author(s)**

Steve

**Licensing**

**Backward Compatibility**

**Implementation Notes**

Code will now throw some errors which it didn't before.  See summary.

**Post Submission Checklist**


athomps (commented):
Please change result for single string to ERROR, not FALSE

sjplimp:
@athomps made the change

athomps (approved):
I approve.

akohlmey:
@sjplimp the changes needed to be reflected in the corresponding unit tests. I have done that and also added some more tests to catch the new errors and additional special cases discussed.

src/variable.cpp
```diff
@@ -4797,32 +4799,40 @@ double Variable::evaluate_boolean(char *str)
             delete[] str1;
             delete[] str2;
           }
+
         } else if (opprevious == LT) {
-          if (flag2) error->all(FLERR,"Invalid Boolean syntax in if command");
-          if (value1 < value2) argstack[nargstack].value = 1.0;
```
akohlmey, line 4802: @sjplimp this change introduces a bug (`$a < $b` is always false) that was flagged in extended regression testing. Will submit a PR with the fix (and a few others).

sjplimp: oops - must have whacked that line for LT when refactoring - good catch

PR #3334, opened 2022-07-05
Bug fix for bonded particles in coul/slater/long

2young-2simple-sometimes-naive:
**Summary**

The current version does not correctly calculate potentials and forces between bonded particles.

**Author(s)**

YH Wang, ywang268@jhu.edu

**Licensing**

**Backward Compatibility**

Input does not need to be changed

**Implementation Notes**
Verification:
The following forces and potentials are calculated under the same setup, for current version, fixed version, and coul/long. 
two particles are bonded by a bond with k=0, placed under various distance r, and the force and potentials are taken from the first timestep of the simulation. 

current force and potential between bonded particles:
image
image

fixed force and potential between bonded particles:
image
image

Intended force and potential between bonded particles, calculated  from coul/long:
image
image

**Post Submission Checklist**

- [x ] The feature or features in this pull request is complete
- [x ] Licensing information is complete
- [x ] Corresponding author information is complete
- [x ] The source code follows the LAMMPS formatting guidelines
- [x ] The feature has been verified to work with the CMake based build system

### DIFF
--- a/src/EXTRA-PAIR/pair_coul_slater_long.cpp
+++ b/src/EXTRA-PAIR/pair_coul_slater_long.cpp
@@ -13,8 +13,9 @@
 ------------------------------------------------------------------------- */
 
 /* ----------------------------------------------------------------------
- *     Contributing author:  Evangelos Voyiatzis (Royal DSM)
- * ------------------------------------------------------------------------- */
+   Contributing author:  Evangelos Voyiatzis (Royal DSM)
+   Bug fix on 07/05/2022: YH Wang (ywang268@jhu.edu)
+------------------------------------------------------------------------- */
 
 #include "pair_coul_slater_long.h"
 
@@ -32,19 +33,21 @@
 
 using namespace LAMMPS_NS;
 
-#define EWALD_F   1.12837917
-#define EWALD_P   0.3275911
-#define A1        0.254829592
-#define A2       -0.284496736
-#define A3        1.421413741
-#define A4       -1.453152027
-#define A5        1.061405429
+#define EWALD_F 1.12837917
+#define EWALD_P 0.3275911
+#define A1 0.254829592
+#define A2 -0.284496736
+#define A3 1.421413741
+#define A4 -1.453152027
+#define A5 1.061405429
+#define EPSILON   1.0e-10
 
 /* ---------------------------------------------------------------------- */
 
 PairCoulSlaterLong::PairCoulSlaterLong(LAMMPS *lmp) : Pair(lmp)
 {
   ewaldflag = pppmflag = 1;
+  ftable = nullptr;
   qdist = 0.0;
 }
 
@@ -52,30 +55,32 @@ PairCoulSlaterLong::PairCoulSlaterLong(LAMMPS *lmp) : Pair(lmp)
 
 PairCoulSlaterLong::~PairCoulSlaterLong()
 {
-  if (!copymode) {
-    if (allocated) {
-      memory->destroy(setflag);
-      memory->destroy(cutsq);
+  if (copymode) return;
 
-      memory->destroy(scale);
-    }
+  if (allocated) {
+    memory->destroy(setflag);
+    memory->destroy(cutsq);
+
+    memory->destroy(scale);
   }
+  if (ftable) free_tables();
 }
 
 /* ---------------------------------------------------------------------- */
 
 void PairCoulSlaterLong::compute(int eflag, int vflag)
 {
-  int i,j,ii,jj,inum,jnum,itype,jtype;
+  int i,j,ii,jj,inum,jnum,itable,itype,jtype;
   double qtmp,xtmp,ytmp,ztmp,delx,dely,delz,ecoul,fpair;
+  double fraction,table;
   double r,r2inv,forcecoul,factor_coul;
   double grij,expm2,prefactor,t,erfc;
   int *ilist,*jlist,*numneigh,**firstneigh;
   double rsq;
-  double slater_term;
+  double brr;
 
-  ev_init(eflag,vflag);
   ecoul = 0.0;
+  ev_init(eflag,vflag);
 
   double **x = atom->x;
   double **f = atom->f;
@@ -85,6 +90,7 @@ void PairCoulSlaterLong::compute(int eflag, int vflag)
   double *special_coul = force->special_coul;
   int newton_pair = force->newton_pair;
   double qqrd2e = force->qqrd2e;
+  double beta = 1.0 / lambda;
 
   inum = list->inum;
   ilist = list->ilist;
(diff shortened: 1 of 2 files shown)


evoyiatzis:
You are actually **introducing** bugs instead of fixing them. First of all, the expression for the potential energy for the coul/slater is different from the expression for coul/long. There is an additional parameter called \lambda which makes very easy to identify the differences. So, you should not get the same energy curves with coul/long and coul/slater. Note that the expression is given explicitly in https://docs.lammps.org/pair_coul_slater.html.  
Then, the idea behind coul/slater is to "smear" the partial charge in space and to not have point charges at in the case of coul/long. This is achieved by the exp(-2r/ $\lambda$) term in the potential energy expression implemented in lines 124 and 126 of the current code. These two lines you are deleting them and you convert the smeared charges to point charges.  
I would like to hear why the curves you are presenting  as "Intended force and potential " are the correct ones. Have they been published in the literature or it is based on your intuition?
Also, the current code has been verified to reproduce published data for ionic solutions (system with no bonds in the reference Gonzalez-Melchor, Mayoral, Velazquez, and Alejandre, J Chem Phys, 125, 224107 (2006). I can share the input script if needed ). Have you done a similar verification?
One last point: why the energy should be zero for short distances (distances of 2.0 in the case you present) ? There is no such inner cut-off in the publications where the slater approach has been presented and used. 
Side note: my native language is modern greek. The greek letter λ is pronounced as λάμδα which translates to lamda (matching greek and english characters). The lambda is more related to ancient greek and even at that times it was not the standard. So, you are not fixing any typo in the code... have a look at https://el.wikipedia.org/wiki/%CE%9B%CE%AC%CE%BC%CE%B4%CE%B1. Also, I should be able to give the names I would like to the variables I use in my code :)

2young-2simple-sometimes-naive:
> You are actually **introducing** bugs instead of fixing them. First of all, the expression for the potential energy for the coul/slater is different from the expression for coul/long. There is an additional parameter called \lambda which makes very easy to identify the differences. So, you should not get the same energy curves with coul/long and coul/slater. Note that the expression is given explicitly in https://docs.lammps.org/pair_coul_slater.html. Then, the idea behind coul/slater is to "smear" the partial charge in space and to not have point charges at in the case of coul/long. This is achieved by the exp(-2r/ λ) term in the potential energy expression implemented in lines 124 and 126 of the current code. These two lines you are deleting them and you convert the smeared charges to point charges. I would like to hear why the curves you are presenting as "Intended force and potential " are the correct ones. Have they been published in the literature or it is based on your intuition? Also, the current code has been verified to reproduce published data for ionic solutions (system with no bonds in the reference Gonzalez-Melchor, Mayoral, Velazquez, and Alejandre, J Chem Phys, 125, 224107 (2006). I can share the input script if needed ). Have you done a similar verification? One last point: why the energy should be zero for short distances (distances of 2.0 in the case you present) ? There is no such inner cut-off in the publications where the slater approach has been presented and used. Side note: my native language is modern greek. The greek letter λ is pronounced as λάμδα which translates to lamda (matching greek and english characters). The lambda is more related to ancient greek and even at that times it was not the standard. So, you are not fixing any typo in the code... have a look at https://el.wikipedia.org/wiki/%CE%9B%CE%AC%CE%BC%CE%B4%CE%B1. Also, I should be able to give the names I would like to the variables I use in my code :)

Please refer to [here](https://matsci.org/t/special-bonds-factor-coul-in-coul-long/37894) for the intended behavior for bonded particles. Your implementation works fine for non-bonded particles, but when the particles are bonded by a bond, they are classical coulomb interaction, as shown in the first 2 pictures (current force and potential between bonded particles). The intended force and potential for bonded particles should be 0, as shown in the last 2 pictures (Intended force and potential between bonded particles, calculated from coul/long:). I am correcting the calculation between bonded particles, the forces and potentials are still the same for non-bonded particles.

akohlmey:
> Please refer to [here](https://matsci.org/t/special-bonds-factor-coul-in-coul-long/37894) for the intended behavior for bonded particles. Your implementation works fine for non-bonded particles, but when the particles are bonded by a bond, they are classical coulomb interaction

This is not correct and you are reading my comments wrong. The potential function does not change with bonds. The problem that is referred to in the archived discussion, is the problem of how to correctly apply an exclusion. With long-range electrostatics, you cannot apply the exclusion factor to the erfc() damped coulomb interaction, but you must subtract the undamped interaction (as in the /cut variant) from that pair. 

Restoring the tabulation is definitely incorrect, since that is based on point charges.

2young-2simple-sometimes-naive:
> This is not correct and you are reading my comments wrong. 

Please disregard my citing then. 

> The potential function does not change with bonds.

After all, the measured force and potential are 0 for bonded particles in coul/long. Isn't that the intended behavior? 
Even if the potential stays the same for bonded and non-bonded, it should be slater instead of coulomb for the coul/slater/long. That means this bug fix request is invalid, but there should be another bug fix.

I've attached the test scripts. Edit plot.py at lin 40 to properly run lammps, then run plot.py to get result.[test-coul-bonded.zip](https://github.com/lammps/lammps/files/9056311/test-coul-bonded.zip) You can test it yourself if you do not trust my result.

```
2 atoms
1 bonds

1 atom types
1 bond types

-10.0 10.0 xlo xhi
-10.0 10.0 ylo yhi
-10.0 10.0 zlo zhi

Masses

1 1.0

Atoms

1 1 1 1 4.0 0.0 0.0
2 2 1 -1 0.0 0.0 0.0

Bonds

1 1 1 2

Bond Coeffs

1 0 1
```

evoyiatzis:
Let me briefly address the most serious point of your misunderstanding:
The potential energy between two charged particles modelled by the coul/slater is given by the expression:
V(r) = (C q_i q_j ) / (ε r) * [1 - (1 + r_{ij}/λ exp(-2r_{ij}/λ) ) ] 
The potential energy between two charged particles modelled by the coul//long is given by the expression:
V(r) = (C q_i q_j ) / (ε r)
The two expressions are given in the documentation (https://docs.lammps.org/pair_coul.html#pair-style-coul-long-command & https://docs.lammps.org/pair_coul_slater.html)
Assuming that ΄λ has a finite value, how is it mathematically possible to make the two expressions produce the same potential energy curves? Yet, the plots you are providing demonstrate that the code you have written for coul/slater reproduces exactly the same curve as coul/long. 
Moreover, I have already pointed out that the way you modify  lines 124 and 126 of the existing code wipes out the  (1 + r_{ij}/λ exp(-2r_{ij}/λ) ) term. So, you get the same mathematical expression V(r) = (C q_i q_j ) / (ε r) programmed in two different pair styles coul/long and coul/slater. 
You stated that "The intended force and potential for bonded particles should be 0": to my level of understanding this is dictated by the user using the special_bonds command - lj/coul or coul options (https://docs.lammps.org/special_bonds.html). It is not dictated from the pair style. In fact, for most of the cases, the DPD models include 1-2 non-bonded interactions (see for instance the seminal work of R.D. Groot and P. B. Warren https://aip.scitation.org/doi/10.1063/1.474784)
You also stated that "when the particles are bonded by a bond, they are classical coulomb interaction": can you justify this statement ? why coulombic interactions for non-bonded particles should be smeared and coulombic interactions for bonded particles should not be smeared? It sounds like such an approach is physically inconsistent.

akohlmey:
> After all, the measured force and potential are 0 for bonded particles in coul/long. Isn't that the intended behavior?

Not exactly zero. The interaction between the immediate pair is zero, but interactions with all periodic images of the pair are not.  Since you cannot handle individual pairs in with kspace you compute all interactions as if there were no exclusions and then specifically subtract the interaction for only the excluded pair.

akohlmey:
@evoyiatzis when looking at the code, I believe that @2young-2simple-sometimes-naive is correct that there is an inconsistency. The line
```
if (factor_coul < 1.0) forcecoul -= (1.0-factor_coul)*prefactor;
```
should be
```
if (factor_coul < 1.0) forcecoul -= (1.0-factor_coul)*prefactor*(1.0-slater_term);
```
I.e. you should subtract exactly the same term as if the special pair was a coul/slater/cut interaction.

akohlmey:
Here is a simple example input that illustrates my remarks:
```
units real
atom_style full
region box block -5 5 -5 5 -5 5
create_box 1 box bond/types 1 extra/bond/per/atom 4

special_bonds lj/coul 0.0 1.0 1.0

create_atoms 1 single  0.0 0.0 0.0
create_atoms 1 single  3.0 0.0 0.0
create_atoms 1 single  0.0 3.0 0.0
create_atoms 1 single  0.0 0.0 3.0

set atom 1 charge -1.0
set atom 2*4 charge  $(1.0/3.0)

mass 1 1.0
compute 1 all pe/atom

pair_style coul/slater/cut 0.5 25.0
pair_coeff * *
bond_style zero
bond_coeff * 3.0

dump 1 all custom 10 dump.lammpstrj id q x y z fx fy fz 

run 0 post no

create_bonds single/bond 1 1 2

run 10 post no

clear

units real
atom_style full
region box block -5 5 -5 5 -5 5
create_box 1 box bond/types 1 extra/bond/per/atom 4

special_bonds lj/coul 0.0 1.0 1.0

create_atoms 1 single  0.0 0.0 0.0
create_atoms 1 single  3.0 0.0 0.0
create_atoms 1 single  0.0 3.0 0.0
create_atoms 1 single  0.0 0.0 3.0

set atom 1 charge -1.0
set atom 2*4 charge  $(1.0/3.0)

mass 1 1.0
compute 1 all pe/atom

pair_style coul/slater/long 0.5 9.0
pair_coeff * *
bond_style zero
bond_coeff * 3.0
kspace_style ewald 1.0e-10

dump 1 all custom 10 dump.lammpstrj id q x y z fx fy fz 
dump_modify 1 append yes
run 0 post no

create_bonds single/bond 1 1 2

run 10 post no
```

a) due to using a cutoff larger than the system, in the coul/slater/cut case the force on the excluded atom pair will not be zero
b) by adding a bond (with no forces or energies) one direct pair should be excluded from the total energy and force. The the change in potential energy and force should be the same. With my proposed change, the difference between the coul/slater/cut and coul/slater/long case becomes smaller.

PR #3336, opened 2022-07-06
Minor improvements and bugfixes to the mdi/qm fix

taylor-a-barnes:
**Summary**

I've been running some tests with the updated `mdi/qm` fix, and have noticed a few bugs and/or portions of the code that could be made more general.  This PR makes the following changes to `fix_mdi_qm.cpp`:

- The use of the `>NATOMS`, `>ELEMENTS`, and `>TYPES` commands are now conditional on the engine having support for these commands.  The code checks whether the driver supports these commands with a call to `MDI_Check_command_exists`.  If the code does not support one of these commands, that command is not sent to the engine.  This allows the `mdi/qm` fix to be used with QM codes that do not support changing the number/types of atoms on-the-fly, as long as no atoms are added or removed over the course of the simulation.
- Similarly, I've made the `>CELL_DISPL` command conditional on the engine having support for it, as many QM codes have hardcoded values for this.  You can typically assume that if the driver sends coordinates to an engine that has a different value of CELL_DISPL, the engine will simply shift any atoms that fall outside of its box by an appropriate lattice vector.  The primary disadvantage to not sending `>CELL_DISPL` to the engine is that if the driver at some point requests coordinates from the engine (i.e., with `<COORDS`), some of those coordinates may need to be shifted back into the driver's box.  Because the `mdi/qm` fix doesn't use `<COORDS`, there shouldn't be much need for `>CELL_DISPL` anyway.
- The values sent alongside the `>CELL` command were not being converted to MDI units.  I've added the conversion.

**Related Issue(s)**

**Author(s)**

Taylor Barnes, MolSSI

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/src/MDI/fix_mdi_qm.cpp
+++ b/src/MDI/fix_mdi_qm.cpp
@@ -246,11 +246,22 @@ void FixMDIQM::init()
   int ierr = MDI_Check_command_exists("@DEFAULT", ">NATOMS", mdicomm, &natoms_exists);
   if (ierr) error->all(FLERR, "MDI: >NATOMS command check");
   if ( natoms_exists ) {
+
     ierr = MDI_Send_command(">NATOMS", mdicomm);
     if (ierr) error->all(FLERR, "MDI: >NATOMS command");
     int n = static_cast<int> (atom->natoms);
     ierr = MDI_Send(&n, 1, MDI_INT, mdicomm);
     if (ierr) error->all(FLERR, "MDI: >NATOMS data");
+
+  } else { // confirm that the engine's NATOMS is correct
+
+    ierr = MDI_Send_command("<NATOMS", mdicomm);
+    if (ierr) error->all(FLERR, "MDI: <NATOMS command");
+    int n;
+    ierr = MDI_Recv(&n, 1, MDI_INT, mdicomm);
+    if (ierr) error->all(FLERR, "MDI: <NATOMS data");
+    if ( n != atom->natoms ) error->all(FLERR, "MDI: Engine has the wrong number of atoms, and does not support the >NATOMS command.");
+
   }
 
   int elements_exists;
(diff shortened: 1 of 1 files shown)


sjplimp:
@taylor-a-barnes Thanks Taylor for looking this over and making changes.  Question about not sending >NATOMS if the engine doesn't accept it.  The use case from the LAMMPS side is to pass a series of different configs for the QM engine to evaluate.  Are you saying that is not possible if the QM engine does not accept NATOMS?  In which case that's OK, but maybe LAMMPS should throw an error, otherwise the user won't know the workflow is not working.  Or maybe the driver should query <NATOMS
after sending and throw an error if it didn't change?  Ditto maybe for TYPES and ELEMENTS.  If there are QM engines that simply don't support that usage mode, somehow LAMMPS needs to know I think.

taylor-a-barnes:
@sjplimp If the QM code doesn't support the `>NATOMS` command, LAMMPS can still send it new atomic coordinates for the atoms every timestep, it just can't change the number of atoms in the system.  In this case, the end-user is responsible for ensuring that the QM code is initialized with the correct number / types of atoms before establishing communication with LAMMPS.  So, for example, you could run AIMD in the microcanonical ensemble, but not the grand canonical ensemble.  You're correct that it would be preferable to confirm that the QM code has been initialized with the correct number of atoms in the event that >NATOMS isn't supported.  I'll make an update to the PR that validates `NATOMS` consistency every timestep - that way, if the end-user _does_ try to run in the grand canonical ensemble with a QM code that can't handle `>NATOMS`, they will get an appropriate error message.

taylor-a-barnes:
@sjplimp The latest commit adds a check for `<NATOMS` if `>NATOMS` isn't supported.

src/MDI/fix_mdi_qm.cpp
```diff
@@ -355,6 +380,9 @@ void FixMDIQM::post_force(int vflag)
     if (ierr) error->all(FLERR, "MDI: <STRESS command");
     ierr = MDI_Recv(qm_virial, 9, MDI_DOUBLE, mdicomm);
     if (ierr) error->all(FLERR, "MDI: <STRESS data");
+    //for (int i = 0; i < 9; i++) {
```
sjplimp: delete these lines?

sjplimp (approved):
just one comment on deleting what look like debug lines

sjplimp:
thanks @taylor-a-barnes - it the debug lines can be delete, this should be ready to merge @akohlmey

PR #3337, opened 2022-07-06
Adding pair style srp/react

vpalkar:
**Summary**
This pull request adds a pair style srp/react. This pair style interfaces the pair style srp (segmental repulsive potential) with the bond breaking and formation mechanisms provided by fix bond/break and fix bond/create, respectively. This is useful in modeling reactions in polymers at the mesoscale where minimizing topological violations is important. I presented about this at the 2021 lammps workshop.

**Author(s)**

Vaibhav Palkar
Kuksenok Lab, Clemson University
vpalkar@g.clemson.edu

**Licensing**

**Backward Compatibility**

**Implementation Notes**
The main implementation is in a new pair style and a new fix style which are based on the original pair srp and fix srp codes. In addition, there are minor changes to original srp, fix bond/break and bond/create codes (mostly only to be able to access variables). 

More implementation details are provided in files linked below under "Further information". I have provided a short file with function-wise diff w.r.t the original SRP code and a longer version that explains the rationale behind all changes.

**Post Submission Checklist**

short summary of changes: https://clemson.box.com/s/wbu8m6sivnesai1kdiv8foc94o95u4t8

longer implementation notes: https://clemson.box.com/s/oiojvojwp5x93264vle0isoqq5yqxjx8

compiled manual pages: https://clemson.box.com/s/i5odg1pby64twyp63ch8salqbmq8s7u6


akohlmey:
@vpalkar Thanks for your contribution and special thanks for making an effort to fully integrate your changes into the build system and the manual. I still did some minor adjustments and a few, mostly cosmetic changes. There were a few more opportunities to use recently added internal APIs and to increase code reuse. Still overall, this pull request made the integration very easy and little effort. Please double check that my changes did not break anything.

vpalkar:
@akohlmey thank you for your comments and the additional code changes. I got a "corrupted double-linked list" error when I pulled your changes to my local repo and ran a couple of small test lammps scripts. The idbreak and idcreate variables were getting deleted twice after your changes in the pair style code. My last commit just fixes this. I had a memory leak in the fix style code for the same variables which your changes fixed, thank you again for that! My test scripts are running successfully after the latest commit.

akohlmey:
Thanks for the the fast turnaround and for fixing the bugs in my bugfix.

sjplimp:
@vpalkar Thanks for contributing this feature.  I just have a hi-level Q.  Would it better to implement reactivity for this model via the newer, more powerful fix bond/react, rather than bond/break and bond/create?  Fix bond/break and bond/create have known problems (or limitations) when used together (a model with both breaking and creating).  Or possibly, if the SRP model is limited to linear chaing CG models, like bead-spring, then bond/break and bond/create are sufficient?

vpalkar:
@sjplimp thanks for your comment, the SRP is indeed used only in CG bead-spring like models (not necessarily linear chains) and hence for most of the use cases the simpler bond/break or bond/create is sufficient. When implementing this, we did also consider fix bond/react as some of its features could be potentially interesting, e.g. (as you pointed out) support for multiple simultaneous bond formation/breaking reactions and local temperature dependence of reaction probability. In the end we decided against using bond/react as these features would come at the cost of complexity, which is not required for several use cases. If there is significant interest, support for bond/react could perhaps be added later into the same pair style.

sjplimp (approved):

PR #3339, opened 2022-07-06
Redoing groups in compute contact/atom

### LINKED ISSUE #3338: [BUG] Incorrect handling of cross-group interactions in compute contact/atom
**Summary**

Compute contact/atom does not correctly filter cross-group contributions to contact numbers when the compute group-ID is not _all_. When looping through i-j neighbors, the compute filters all contributions where atom i is not in the compute group, seen [here](https://github.com/lammps/lammps/blob/b8acd2e31de3d19ac0a4629ef168e183335bcc74/src/GRANULAR/compute_contact_atom.cpp#L123), and never checks if atom j is a part of the group.

I have a patched copy of compute contact/atom I can submit shortly in a pull request to fix this bug. The patch also adds the ability to limit which groups contribute to the contact number, as can be done in compute coord/atom. 

Current development branch

**Expected Behavior**

From the documentation: "The value of the contact number will be 0.0 for atoms not in the specified compute group."

**Actual Behavior**

When the compute group is not _all_, atoms outside of the compute group may have a non-zero contact number if they are in contact with atoms within the group. Additionally, atoms inside of the compute group may not have the correct contact number if they are in contact with atoms outside of the group.

**Steps to Reproduce**

The attached input script creates a lattice of particles and separates the bottom half into its own group. Using this group, compute contact/atom should return an indentical contact number for every atom in this group (bottom half) and return a value of zero for every atom outside of this group (top half). Instead, atoms on the interface have intermediate contact numbers.

[in.contact.txt](https://github.com/lammps/lammps/files/9057246/in.contact.txt)

jtclemm:
**Summary**

This pull request fixes the incorrect cross-group handling incompute contact/atom when the compute group is not _all_ as reported in [issue #3338](https://github.com/lammps/lammps/issues/3338).

While fixing the bug, I also extended the functionality of compute contact/atom to match that of compute coord/atom allowing the user to specify a second group ID. The second ID, defaulted to _all_, restricts which atoms contribute to the contact number. 

**Related Issue(s)**

Fixes [#3338](https://github.com/lammps/lammps/issues/3338).

**Author(s)**

Joel Clemmer (SNL)

**Licensing**

**Backward Compatibility**

NA

**Implementation Notes**

I debated whether the input arguments should be:
`compute 1 all contact/atom mygroupID` (as implemented)
versus
`compute 1 all contact/atom group mygroupID` (as in compute coord/atom)
I'm not sure which is preferred.

**Post Submission Checklist**

NA

### DIFF
--- a/doc/src/compute_contact_atom.rst
+++ b/doc/src/compute_contact_atom.rst
@@ -8,10 +8,11 @@ Syntax
 
 .. parsed-literal::
 
-   compute ID group-ID contact/atom
+   compute ID group-ID contact/atom group2-ID
 
 * ID, group-ID are documented in :doc:`compute <compute>` command
 * contact/atom = style name of this compute command
+* group2-ID = optional argument to restrict which atoms to consider for contacts (see below)
 
 Examples
 """"""""
@@ -19,6 +20,7 @@ Examples
 .. code-block:: LAMMPS
 
    compute 1 all contact/atom
+   compute 1 all contact/atom mygroup
 
 Description
 """""""""""
@@ -45,6 +47,9 @@ overview of LAMMPS output options.
 The per-atom vector values will be a number >= 0.0, as explained
 above.
 
+The optional *group2-ID* argument allows to specify from which group atoms
+contribute to the coordination number. Default setting is group 'all'.
+
 Restrictions
 """"""""""""
 
@@ -62,5 +67,8 @@ Related commands
 
 Default
 """""""
+
+*group2-ID* = all
+
 
 none
--- a/src/GRANULAR/compute_contact_atom.cpp
+++ b/src/GRANULAR/compute_contact_atom.cpp
@@ -18,6 +18,7 @@
 #include "comm.h"
 #include "error.h"
 #include "force.h"
+#include "group.h"
 #include "memory.h"
 #include "modify.h"
 #include "neigh_list.h"
@@ -34,7 +35,16 @@ ComputeContactAtom::ComputeContactAtom(LAMMPS *lmp, int narg, char **arg) :
   Compute(lmp, narg, arg),
   contact(nullptr)
 {
-  if (narg != 3) error->all(FLERR,"Illegal compute contact/atom command");
+  if (narg != 3 && narg != 4) error->all(FLERR,"Illegal compute contact/atom command");
+
+  jgroup = group->find("all");
+  jgroupbit = group->bitmask[jgroup];
+  if (narg == 4) {
+    group2 = utils::strdup(arg[3]);
+    jgroup = group->find(group2);
+    if (jgroup == -1) error->all(FLERR, "Compute contact/atom group2 ID does not exist");
+    jgroupbit = group->bitmask[jgroup];
+  }
 
   peratom_flag = 1;
   size_peratom_cols = 0;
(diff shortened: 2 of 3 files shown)


src/GRANULAR/compute_contact_atom.cpp
```diff
@@ -120,28 +130,28 @@ void ComputeContactAtom::compute_peratom()
...
+      j &= NEIGHMASK;
+
+      delx = xtmp - x[j][0];
+      dely = ytmp - x[j][1];
+      delz = ztmp - x[j][2];
+      rsq = delx * delx + dely * dely + delz * delz;
+      radsum = radi + radius[j];
+      radsumsq = radsum * radsum;
+      if (rsq <= radsumsq) {
+        // Only tally for atoms in compute group (groupbit) if neighbor is in group2 (jgroupbit)
```
sjplimp: Not clear why these mask tests are here, rather than after the j &= NEIGHMASK line.  To avoid computing the distance when the 2 mask values will skip it anyway?

sjplimp: Could also do the mask test on atom i before the j loop, then only one test on atom J inside the j loop, after j is set.

doc/src/compute_contact_atom.rst
```diff
@@ -8,17 +8,19 @@ Syntax
 
 .. parsed-literal::
 
-   compute ID group-ID contact/atom
+   compute ID group-ID contact/atom group2-ID
```
sjplimp, line 11: I think this syntax is fine, since its the only option for now

sjplimp (changes requested):
See comments on mask checking

src/GRANULAR/compute_contact_atom.cpp
jtclemm: That's a good point, I can filter where i or j are not in either of the two groups to avoid calculating rsq. However, I think the compute will still need the final mask tests before incrementing `contact` to determine which atom(s) are updated. I'll push an update.

src/GRANULAR/compute_contact_atom.cpp
```diff
@@ -115,12 +125,15 @@ void ComputeContactAtom::compute_peratom()
...
+  bool update_i_flag, update_j_flag;
 
   for (i = 0; i < nall; i++) contact[i] = 0.0;
 
   for (ii = 0; ii < inum; ii++) {
     i = ilist[ii];
-    if (mask[i] & groupbit) {
+
+    // Only proceed if i is either part of the compute group or will contribute to contacts
+    if ((mask[i] & groupbit) || (mask[i] & jgroupbit)) {
```
sjplimp: minor point - I would make this if test and the j one below be "continue"
statements so the code is easier to read w/out too deep of nesting

src/GRANULAR/compute_contact_atom.cpp
```diff
@@ -132,15 +145,21 @@ void ComputeContactAtom::compute_peratom()
...
-        radsum = radi + radius[j];
-        radsumsq = radsum*radsum;
-        if (rsq <= radsumsq) {
-          contact[i] += 1.0;
-          contact[j] += 1.0;
+        // Only tally for atoms in compute group (groupbit) if neighbor is in group2 (jgroupbit)
+        update_i_flag = (mask[i] & groupbit) && (mask[j] & jgroupbit);
+        update_j_flag = (mask[j] & groupbit) && (mask[i] & jgroupbit);
+
+        if (update_i_flag || update_j_flag) {
```
sjplimp: ditto for this if statement being a continue

sjplimp (changes requested):
just requesting a minor syntax change

src/GRANULAR/compute_contact_atom.cpp
jtclemm: Just made the swap. Thanks for the input!

sjplimp (approved):

PR #3341, opened 2022-07-07
Add pair_modify option to allow separate cutoffs for each neigh list …

stanmoore1:
…with pair hybrid/overlay

**Summary**

This PR adds a `pair_modify` option: `neigh/trim` to allow separate cutoffs for each neigh list with `pair hybrid/overlay` and also adds "trim" methods for building a neighbor list with a shorter cutoff from one with a longer cutoff. This can give speedup when the cutoffs are significantly different.

For example, consider the following input command:

```
pair_style      hybrid/overlay lj/cut 2.5 coul/cut 5.0
pair_coeff      1 1 lj/cut 1.0 1.0
pair_coeff      1 1 coul/cut
```

With `pair_modify neigh/trim no` the code gives the same behavior as before:

```
  master list distance cutoff = 5.3
  ghost atom cutoff = 5.3
  binsize = 2.65, bins = 13 13 13
  2 neighbor lists, perpetual/occasional/extra = 2 0 0
  (1) pair lj/cut, perpetual
      attributes: half, newton on
      pair build: half/bin/atomonly/newton
      stencil: half/bin/3d
      bin: standard
  (2) pair coul/cut, perpetual, copy from (1)
      attributes: half, newton on
      pair build: copy
      stencil: none
      bin: none
```

While with `pair_modify neigh/trim yes` the code builds a separate neigh list for `lj/cut` by trimming the neigh list built for `coul/cut`:

```
  master list distance cutoff = 5.3
  ghost atom cutoff = 5.3
  binsize = 2.65, bins = 13 13 13
  2 neighbor lists, perpetual/occasional/extra = 2 0 0
  (1) pair lj/cut, perpetual, trim from (2)
      attributes: half, newton on, cut 2.8
      pair build: trim
      stencil: none
      bin: none
  (2) pair coul/cut, perpetual
      attributes: half, newton on
      pair build: half/bin/atomonly/newton
      stencil: half/bin/3d
      bin: standard
```

**Related Issue(s)**

None

**Author(s)**

Stan Moore (SNL), with helpful discussions from Steve Plimpton (SNL)

**Licensing**

**Backward Compatibility**

Yes

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/doc/src/neighbor.rst
+++ b/doc/src/neighbor.rst
@@ -49,29 +49,27 @@ sometimes be faster.  Either style should give the same answers.
 
 The *multi* style is a modified binning algorithm that is useful for
 systems with a wide range of cutoff distances, e.g. due to different
-size particles. For granular pair styles, cutoffs are set to the
-sum of the maximum atomic radii for each atom type.
-For the *bin* style, the bin size is set to 1/2 of
-the largest cutoff distance between any pair of atom types and a
-single set of bins is defined to search over for all atom types.  This
-can be inefficient if one pair of types has a very long cutoff, but
-other type pairs have a much shorter cutoff. The *multi* style uses
-different sized bins for collections of different sized particles, where
-"size" may mean the physical size of the particle or its cutoff
-distance for interacting with other particles. Different
+size particles. For granular pair styles, cutoffs are set to the sum of
+the maximum atomic radii for each atom type.  For the *bin* style, the
+bin size is set to 1/2 of the largest cutoff distance between any pair
+of atom types and a single set of bins is defined to search over for all
+atom types.  This can be inefficient if one pair of types has a very
+long cutoff, but other type pairs have a much shorter cutoff. The
+*multi* style uses different sized bins for collections of different
+sized particles, where "size" may mean the physical size of the particle
+or its cutoff distance for interacting with other particles. Different
 sets of bins are then used to construct the neighbor lists as as further
 described by Shire, Hanley, and Stratford :ref:`(Shire) <bytype-Shire>`.
-This imposes some extra setup overhead, but the searches themselves
-may be much faster. By default, each atom type defines a separate
-collection of particles. For systems where two or more atom types
-have the same size (either physical size or cutoff distance), the
-definition of collections can be customized, which can result in less
-overhead and faster performance. See the :doc:`neigh_modify <neigh_modify>`
-command for how to define custom collections. Whether the collection
-definition is customized or not, also see the
-:doc:`comm_modify mode multi <comm_modify>` command for communication
-options that further improve performance in a manner consistent with
-neighbor style multi.
+This imposes some extra setup overhead, but the searches themselves may
+be much faster. By default, each atom type defines a separate collection
+of particles. For systems where two or more atom types have the same
+size (either physical size or cutoff distance), the definition of
+collections can be customized, which can result in less overhead and
+faster performance. See the :doc:`neigh_modify <neigh_modify>` command
+for how to define custom collections. Whether the collection definition
+is customized or not, also see the :doc:`comm_modify mode multi
+<comm_modify>` command for communication options that further improve
+performance in a manner consistent with neighbor style multi.
 
 An alternate style, *multi/old*, sets the bin size to 1/2 of the shortest
 cutoff distance and multiple sets of bins are defined to search over for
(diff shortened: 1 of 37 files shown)


akohlmey:
FYI, there are three segmentation faults for unit tests. This often happens when there is some uninitialized data.

stanmoore1:
This is working in all my tests. One question, should `neigh/trim yes` be the default since it is normally faster, or should we keep the old behavior? @sjplimp @akohlmey

athomps (approved):
I approve.

akohlmey:
> This is working in all my tests. One question, should `neigh/trim yes` be the default since it is normally faster, or should we keep the old behavior? @sjplimp @akohlmey 

There still seem to be test failures with the embedded unit tests. I will need some time to reconfigure my phone to find a provider that supports roaming and the wifi hotspot of my cell phone before I can check it out in detail.

If we find and resolve the causes of the failures, I would suggest to make the new option the default. We are at the beginning of a development cycle and now is the best time for such changes and this way we increase the chances of getting bug reports.
If it turns out to be too risky, we can always roll it back and put a warning into the manual (faster but may not work for all cases).

akohlmey (changes requested):
There are many failed unit tests now:
https://ci.lammps.org/view/LAMMPS%20develop/job/dev/job/pull_requests/job/ubuntu/job/unit_tests/job/cmake_mpi_openmp_smallbig_shared/5027/

akohlmey:
@stanmoore1 I fixed two bugs, one due to an uninitialized variable and one in pair style hybrid/scaled due to an unallocated array (that style overrides the settings() functions from pair style hybrid).

akohlmey (approved):

akohlmey:
With the last two commits all unit tests pass.

akohlmey:
There are still issues running with OpenMP, e.g. the command `~/compile/lammps/build-test/lmp -in in.snap.InP.JCPA2020 -sf omp` in the examples/snap folder yields:
```
Hybrid pair style last /omp style zbl
Last active /omp style is pair_style zbl
ERROR: Requested neighbor pair method does not exist (src/neighbor.cpp:928)
Last command: run             ${nsteps}
```
Please see: https://ci.lammps.org/job/dev/job/pull_requests/job/ubuntu/job/run_tests/job/cmake_testing_openmp/563/
There may be similar issues when using the INTEL package as well.

akohlmey (changes requested):
The listed failing inputs need to be able to run without error.

stanmoore1:
@akohlmey thanks for working on this--I will fix the OMP/INTEL issue

akohlmey:
@stanmoore1 wouldn't it be cleaner to just define "forwarders" to the non-accelerated version like for the "skip" neighbor list classes in `src/OPENMP/npair_skip_omp.h`?

stanmoore1:
Yes, I didn't know that was an option.

akohlmey:
> Yes, I didn't know that was an option.

...and I didn't think of it until I saw your commit. :disappointed:

stanmoore1:
I don't think that will work for INTEL though, since the structure of the neigh list is different?

akohlmey:
> I don't think that will work for INTEL though, since the structure of the neigh list is different?

I don't know. Haven't looked at those in a long time. That would be a question for @wmbrownIntel

stanmoore1:
@akohlmey it was fairly trivial to add `npair trim` methods for OPENMP and INTEL. If you could test the new code that would be helpful. Thanks

stanmoore1:
@akohlmey there may be other corner cases that would error out, i.e. `trim + ssa` or `trim + respa`, any ideas what to do about that?

stanmoore1:
One other issue I noticed is that this code in `pair_hybrid` doesn't work if a pair_style has multiple neigh list requests (e.g. MEAM): https://github.com/lammps/lammps/blob/develop/src/pair_hybrid.cpp#L622, since it just finds the first request and then breaks out.

stanmoore1:
The Intel trim is broken in my test, debugging...

stanmoore1:
@akohlmey this is now passing my tests. Besides the issue of `pair_hybrid` not handling a pair_style that has multiple neigh list requests, this is ready to merge from my POV.

akohlmey:
> One other issue I noticed is that this code in `pair_hybrid` doesn't work if a pair_style has multiple neigh list requests (e.g. MEAM): https://github.com/lammps/lammps/blob/develop/src/pair_hybrid.cpp#L622, since it just finds the first request and then breaks out.
> 

This sounds like a job for @sjplimp to look into, and it may be a more general problem that could be reported as a separate issue and then addressed in a separate pull request.

akohlmey (approved):

PR #3345, opened 2022-07-12
Bond morse lattice fix

raynol-dsouza:
Modified morse_bond_lattice.cpp to run on lammps version 29Sept21


raynol-dsouza:
This was an accidental pull request. Kindly ignore.

PR #3347, opened 2022-07-14
New fix pair command

sjplimp:
**Summary**

Implement a fix pair command which can do 3 things:
a) trigger a pair style to perform extra per-atom computations on a timestep and/or make those quantities available
b) store the pair style per-atom values in the fix so they are accessible to other LAMMPS commands
c) trigger a dump to be invoked if an equal-style variable condition is met, the dump can output the fixes values

Use cases:
a) ML potentials which can compute a per-atom quantity like "uncertainty".  A dump of the config can be triggered
when the uncertainty is high, so that further DFT training can be performed
b) Access to the internal induced dipole moment for the new AMOEBA potentials, so it can be dumped or otherwise analyzed 

The triggered dumps by this fix should be specified with Nevery = 0, so they become "external" dumps not triggered by the
normal LAMMPS output.  Rather they are triggered externally, e.g. by this fix.

**Related Issue(s)**

**Author(s)**

Steve and Axel and Yury Lysogorskiy ([yury.lysogorskiy@rub.de](mailto:yury.lysogorskiy@rub.de))

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/doc/src/dump_modify.rst
+++ b/doc/src/dump_modify.rst
@@ -357,7 +357,7 @@ always occur if the current timestep is a multiple of $N$, the
 frequency specified in the :doc:`dump <dump>` command or
 :doc:`dump_modify every <dump_modify>` command, including timestep 0.
 It will also always occur if the current simulation time is a multiple
-of *Delta*, the time interval specified in the doc:`dump_modify
+of *Delta*, the time interval specified in the :doc:`dump_modify
 every/time <dump_modify>` command.
 
 But if this is not the case, a dump snapshot will only be written if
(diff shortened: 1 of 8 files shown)


sjplimp:
@akohlmey @yury-lysogorskiy This initial version of "fix pair" stores per-atom values and can trigger the dump output.  Next step is to add support for it to interact with the pair style.  No doc pages yet.
Sample input script:

# 3d Lennard-Jones melt

units		lj
atom_style	atomic

lattice		fcc 0.8442
region		box block 0 5 0 5 0 5
create_box	1 box
create_atoms	1 box
mass		1 1.0

velocity	all create 1.44 87287 loop geom

pair_style	lj/cut 2.5
pair_coeff	1 1 1.0 1.0 2.5

neighbor	0.3 bin
neigh_modify	delay 0 every 20 check no

variable        trigger equal 1

fix		1 all nve
fix             2 all pair 30 lj/cut foo dump 1 trigger # dump 2 trigger
#fix             3 all pair 40 lj/cut foo dump 2 trigger

dump            1 all custom 0 tmp1.dump id type x y z f_2[*]
dump_modify     1 delay 1

dump            2 all custom 0 tmp2.dump id type x y z f_2[*]
dump_modify     2 delay 1

run		100

sjplimp:
@akohlmey Another use case for these externally driven dumps (Nevery = 0) would be to have an additional option on the write_dump command, for "write_dump ID".  This assumes the dump has been previously defined as external.  Then the user could do a write_dump command whenever they wished.  E.g. after min, between run commands of varying length, etc.
However if the dump used per-atom energy or virial from the pair style, then the code would not be able to predict what step that would happen next.

akohlmey:
@sjplimp this all "all yours" now.

akohlmey:
Obsoleted by PR #3369

PR #3349, opened 2022-07-15
Collected small changes

akohlmey:
**Summary**

Various small changes and bug fixes

**Related Issue(s)**

- The change to the "ramp()" function was motivated by this discussion: https://matsci.org/t/expand-compress-region/43458/
- The change to the GPU library is based on the report here: https://matsci.org/t/problem-building-lammps-with-intel-icps/43380

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

The following changes are implemented:
- replace "auto" with explicit data type in the GPU lib since some Intel compiler versions seem to sometimes not able to deduce the type
- fix cut-n-paste error in `dump_modify` error message and improve the error message
- enable `dump_modify colname` support for `dump local` that was somehow forgotten
- allow the "ramp(x,y)" function to be used in between runs (it will return the value of x in this case), also avoid division by zero when used with "run 0"
- the variable functions vdisplace(), swiggle(), and cwiggle() may not be used with fix dt/reset since they assume a constant dt
- fix copy-n-paste bug in swiggle() variable function.
- silence warnings about integer overflows from CodeQL

**Post Submission Checklist**

### DIFF
--- a/doc/src/variable.rst
+++ b/doc/src/variable.rst
@@ -685,7 +685,9 @@ of a run, according to this formula:
 The run begins on startstep and ends on stopstep.  Startstep and
 stopstep can span multiple runs, using the *start* and *stop* keywords
 of the :doc:`run <run>` command.  See the :doc:`run <run>` command for
-details of how to do this.
+details of how to do this.  If called in between runs or during a
+:doc:`run 0 <run>` command, the ramp(x,y) function will return the
+value of x.
 
 The stagger(x,y) function uses the current timestep to generate a new
 timestep.  X,y > 0 and x > y are required.  The generated timesteps
--- a/doc/src/variable.rst
+++ b/doc/src/variable.rst
@@ -785,8 +785,11 @@ where dt = the timestep size.
 The run begins on startstep.  Startstep can span multiple runs, using
 the *start* keyword of the :doc:`run <run>` command.  See the :doc:`run
 <run>` command for details of how to do this.  Note that the
-:doc:`thermo_style <thermo_style>` keyword elaplong =
-timestep-startstep.  This function may not be used in combination with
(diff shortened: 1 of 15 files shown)


akohlmey:
@sjplimp this is PR is not complete, but please have a look at the changes to the variable code and the ramp() functions and add your comments (please also look at the use case). I am particularly interested why there was a hard error (but only in one of the three places where the function is programmed, why so?) when the behavior that I chose with this PR seems quite reasonable (at least to me).

sjplimp:
@akohlmey Similar betwee-run error for ramp is generated by vdisplace, swiggle, cwiggle

sjplimp:
@akohlmey Also, just noticed this typo:
This line appears in an swiggle code chunk, when it should be SWIGGLE
if (tree) newtree->type = CWIGGLE;

akohlmey:
> @akohlmey Also, just noticed this typo: This line appears in an swiggle code chunk, when it should be SWIGGLE if (tree) newtree->type = CWIGGLE;

This is now fixed (and the fix cherry-picked into the maintenance branch for the stable version)

akohlmey:
> @akohlmey Similar betwee-run error for ramp is generated by vdisplace, swiggle, cwiggle

@sjplimp I had a close look at those and there is a significant difference: they all take the increment as an argument while ramp() has to compute it. With that it seems to me that it would be better to leave them unchanged after removing the check for being in between runs. But what I **did** notice is that these three functions assume that *dt* is constant. So I added a check that disallows using them when fix dt/reset is in use.

sjplimp:
@akohlmey yes, removing the error check so they can be invoked between runs is what I was hoping

sjplimp (approved):

PR #3352, opened 2022-07-18
Add set diameter/range command

AkosSeres:
**Summary**

In my granular material simulations, I needed polydisperse spherical grains so I added the `diameter/range` option to the `set` command, which can be used to give a __random diameter to each particle within a range__.

Example usage:
```
set type 1 diameter/range 353423 0.5 1.5
```
This assigns a random diameter to each particle of type `1` between `0.5` and `1.5` with uniform distribution. `353423` is the seed.

I also added an entry to the documentation regarding the new option.

**Related Issue(s)**

None

**Author(s)**

Ákos Seres

**Licensing**

**Backward Compatibility**

No changes

**Implementation Notes**

There are other options for the `set` command for setting random properties, so implementing this feature was straightforward based on the pattern of the other options.

**Post Submission Checklist**

With the example command, I could create the following system:

Screenshot 2022-07-18 at 10 56 32

This random size distribution is needed because I want to avoid the crystallization of the sheared material which occurs when the spheres are the same size or too similar in size.

### DIFF
--- a/doc/src/set.rst
+++ b/doc/src/set.rst
@@ -16,7 +16,7 @@ Syntax
 * keyword = *type* or *type/fraction* or *type/ratio* or *type/subset*
   or *mol* or *x* or *y* or *z* or *charge* or *dipole* or
   *dipole/random* or *quat* or *spin* or *spin/random* or
-  *quat* or *quat/random* or *diameter* or *shape* or
+  *quat* or *quat/random* or *diameter* or *diameter/range* or *shape* or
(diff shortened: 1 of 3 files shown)


akohlmey:
Please note that this functionality already exists with:
```
variable diameter atom random(0.5,1.5,353423)
set type 1 diameter v_diameter
```
In addition with a variable other random distributions beyond a uniform distribution can be achieved or more complex settings can be realized.

AkosSeres:
Oh thank you, I wasn't aware of that. Sorry for the unnecessary pr.

PR #3355, opened 2022-07-19
Improve consistency in Python interpreter and libraries lookup in CMake

### LINKED ISSUE #3354: Set explicit `python` with `-D Python_EXECUTABLE=path`
**Summary**

Typo in the docs at  https://docs.lammps.org/Build_extras.html#python to set an explicit `python`

Version: 23Jun2022
Platform: AMD Zen2 with NVIDIA V100 GPU, Rocky Linux 8, EasyBuild recipe for LAMMPS

**Details**

The current docs say you can use
```
-D PYTHON_EXECUTABLE=path
```
to explicitly set the `python` to be used. This setting seems to be ignored though, when I used CMake 3.20.1 the correct setting was
```
-D Python_EXECUTABLE=path
```

akohlmey:
**Summary**

This pull request makes checking for the python interpreter and python libraries more consistent and portable across OSes and CMake versions.

**Related Issue(s)**

Fixes #3354 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

This pull request implements three sets of changes:
- The CMake variable PYTHON_EXECUTABLE can be used to override the automatically found python interpreter for all known CMake versions (previously 3.12 and later would require Python_EXECUTABLE instead). Internally both are set to the same value.
- When run in a python virtual environment (as indicated by the VIRTUAL_ENV environment variable), then the python interpreter in that virtual environment is set as default (can be overridded with -DPYTHON_EXECUTABLE)
- When looking for Python libraries with CMake pre-3.12 it is now always first searched for the interpreter and then the library search preferred to that version and the resulting versions compared. If library and interpreter have different versions CMake now stops with an error.

**Post Submission Checklist**

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -795,9 +795,13 @@ if(BUILD_SHARED_LIBS)
     set(Python_ADDITIONAL_VERSIONS 3.12 3.11 3.10 3.9 3.8 3.7 3.6)
     find_package(PythonInterp) # Deprecated since version 3.12
     if(PYTHONINTERP_FOUND)
-        set(Python_EXECUTABLE ${PYTHON_EXECUTABLE})
+      set(Python_EXECUTABLE ${PYTHON_EXECUTABLE})
     endif()
   else()
+    # backward compatibility
+    if(PYTHON_EXECUTABLE)
+      set(Python_EXECUTABLE ${PYTHON_EXECUTABLE})
+    endif()
     find_package(Python COMPONENTS Interpreter)
   endif()
   if(BUILD_IS_MULTI_CONFIG)
@@ -830,11 +834,17 @@ endif()
 ###############################################################################
 if(BUILD_SHARED_LIBS OR PKG_PYTHON)
   if(CMAKE_VERSION VERSION_LESS 3.12)
+    # adjust so we find Python 3 versions before Python 2 on old systems with old CMake
+    set(Python_ADDITIONAL_VERSIONS 3.12 3.11 3.10 3.9 3.8 3.7 3.6)
     find_package(PythonInterp) # Deprecated since version 3.12
     if(PYTHONINTERP_FOUND)
-        set(Python_EXECUTABLE ${PYTHON_EXECUTABLE})
+      set(Python_EXECUTABLE ${PYTHON_EXECUTABLE})
     endif()
   else()
+    # backward compatibility
+    if(PYTHON_EXECUTABLE)
+      set(Python_EXECUTABLE ${PYTHON_EXECUTABLE})
+    endif()
     find_package(Python COMPONENTS Interpreter)
   endif()
   if(Python_EXECUTABLE)
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -154,6 +154,19 @@ endif()
 ########################################################################
 # User input options                                                   #
 ########################################################################
+# set path to python interpreter and thus enforcing python version if
+# when in a virtual environment and PYTHON_EXECUTABLE is not set on command line
+if(DEFINED ENV{VIRTUAL_ENV} AND NOT PYTHON_EXECUTABLE)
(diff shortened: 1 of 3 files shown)


akohlmey:
@ocaisa this should hopefully address the inconsistency you reported in #3354 
To retain compatibility with CMake 3.10 (and thus Ubuntu18.04LTS) it turned out to be more convoluted than I would like it to be, but this is now behaving consistently in all tests that I made.
I am a bit hesitant to include this in the updates for the stable version right away. I think it would be better to observe if these changes cause any problems in the development version and then cherry-pick them into the stable branch maintenance patch later.

ocaisa:
For EasyBuild (for now), I've just taken a pragmatic approach and defined both `PYTHON_EXECUTABLE` and `Python_EXECUTABLE` knowing that one or the other will get ignored.

akohlmey:
> For EasyBuild (for now), I've just taken a pragmatic approach and defined both `PYTHON_EXECUTABLE` and `Python_EXECUTABLE` knowing that one or the other will get ignored.

Yeah, the changes to automatically pick up the python interpreter from a virtual environment if it is not set from the CMake command line are not likely to apply to your case.

But please also check whether CMake picks up the Python library version that matches the selected interpreter.
This PR includes changes that improve finding the consistent library version and checks that would error out, if they are not.
In my initial tests I came across cases where those would be different, but I don't remember whether this only applies to CMake versions older than 3.12 with the new find_package() module(s) for Python and thus also don't apply to your case.

akohlmey:
@sjplimp @taylor-a-barnes  FYI
While testing this pull request I noticed that the MDI library will not be able to detect (compatible) python support for CMake versions older than version 3.12 due to using a CMake module that was added in that version.
Thus people compiling on Ubuntu 18.04LTS (which ships with CMake 3.10 and is the reason why we have chosen CMake 3.10 as the minimum CMake version we support) will get some scary warnings and python support will not be included. Otherwise compilation does succeed, though.

ocaisa:
> But please also check whether CMake picks up the Python library version that matches the selected interpreter. This PR includes changes that improve finding the consistent library version and checks that would error out, if they are not. 

I don't see any inconsistencies in the builds I've done, thanks for the warning! For `3Mar2020`, it was reporting the `PythonLibs` found in the CMake output (and these are matching), for `23Jun2022` I don't see any similar output to check.

ocaisa:
I tell a lie, I checked on another system where the system default Python was newer than the one I want to use, and I indeed see the issue (for `3Mar2020`):
```
[ocaisa@gpunode1 ~]$ grep 'Found PythonI\|Found PythonL' /project/def-sponsor00/easybuild/build/LAMMPS/3Mar2020/foss-2020a-Python-3.8.2-kokkos/easybuild/easybuild-LAMMPS-3Mar2020-20220720.111635.log
-- Found PythonLibs: /cvmfs/pilot.eessi-hpc.org/versions/2021.12/compat/linux/x86_64/usr/lib64/libpython3.9.so (found version "3.9.9") 
-- Found PythonInterp: /cvmfs/pilot.eessi-hpc.org/versions/2021.12/software/linux/x86_64/amd/zen2/software/Python/3.8.2-GCCcore-9.3.0/bin/python (found version "3.8.2")
```
when configured with
```
-DPython_EXECUTABLE=/cvmfs/pilot.eessi-hpc.org/versions/2021.12/software/linux/x86_64/amd/zen2/software/Python/3.8.2-GCCcore-9.3.0/bin/python  \
-DPYTHON_EXECUTABLE=/cvmfs/pilot.eessi-hpc.org/versions/2021.12/software/linux/x86_64/amd/zen2/software/Python/3.8.2-GCCcore-9.3.0/bin/python
```
I've mitigated for this by also defining `PYTHON_LIBRARY` and `PYTHON_INCLUDE_DIR` (which are ignored by `23Jun2022`)

sjplimp (approved):

PR #3358, opened 2022-07-20
Add a section to manual with info on updating old source code for current LAMMPS versions

akohlmey:
**Summary**

This adds a section to the manual listing various changes to the internal APIs and conventions that code must be or should be updated for so it can be compiled with a current version of LAMMPS and possibly integrated into the distribution.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Post Submission Checklist**

### DIFF
--- a/doc/src/Developer_updating.rst
+++ b/doc/src/Developer_updating.rst
@@ -1,16 +1,21 @@
 Notes for updating code written for older LAMMPS versions
 ---------------------------------------------------------
 
-This section documents how C++ source files that were written for an
-older version of LAMMPS need to be updated to be compatible with the
-current and future version(s).  Due to the active development of LAMMPS
-it is likely to always be incomplete.  Please contact developer@lammps.org
-in case you run across an issue that is not (yet) listed here.  Please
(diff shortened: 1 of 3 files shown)


doc/src/Developer_updating.rst
```diff
@@ -0,0 +1,304 @@
+Notes for updating code written for older LAMMPS versions
+---------------------------------------------------------
+
+This section documents how C++ source files that were written for an
```
sjplimp: Clarify that this section is for users who have their own private LAMMPS extensions which they wish to keep synced with current LAMMPS.  I.e. not for code previously contributed to public LAMMPS.

sjplimp (changes requested):
see single comment I added

doc/src/Developer_updating.rst
akohlmey: done

rbberger (approved):

sjplimp (approved):

PR #3374, opened 2022-08-02
Small cleanup and performance tweak for spline evaluation in AIREBO pair style

akohlmey:
**Summary**

This removes some redundant code and optimizes the 5th order spline evaluation by using Horner's rule (which should also improve precision a little bit) and moving it to the class header to have it inlined.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Post Submission Checklist**

### DIFF
--- a/src/MANYBODY/pair_airebo.cpp
+++ b/src/MANYBODY/pair_airebo.cpp
@@ -66,7 +66,6 @@ PairAIREBO::PairAIREBO(LAMMPS *lmp)
   pgsize = oneatom = 0;
 
   nC = nH = nullptr;
-  map = nullptr;
   manybody_flag = 1;
   centroidstressflag = CENTROID_NOTAVAIL;
 
@@ -98,7 +97,6 @@ PairAIREBO::~PairAIREBO()
     memory->destroy(lj2);
     memory->destroy(lj3);
     memory->destroy(lj4);
-    delete[] map;
   }
 }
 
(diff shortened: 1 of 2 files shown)


sjplimp:
@akohlmey is this analytically identical code?

akohlmey:
> @akohlmey is this analytically identical code?

yes.

This uses "Horner's rule", i.e. a polynomial evaluation like `a_3 * x^3 + a_2 * x^2 + a_1 * x + a_0` becomes `(((a_3 * x) + a_2) * x + a_1) * x + a_0`.

This is the minimum number of evaluations and due to the incremental processing should reduce floating point errors for higher order terms if the coefficients have alternating signs as the magnitude of numbers differs less and thus there should be less truncation.
By alternating between the polynomial and the derivative, this also promotes vectorization by the compiler.

Unit and regression tests pass (we have multiple tests for AIREBO) and benchmarks are a bit faster (about 5% on my machine).

sjplimp (approved):

akohlmey:
@sjplimp fun fact: I originally learned this in middle school when learning how to efficiently compute numerical results alongside using logarithm tables. Electronic calculators were still quite expensive and became a standard tool only a few years later and were "outlawed" except in Physics classes where the school provided [TI-30](https://en.wikipedia.org/wiki/TI-30) calculators for use in the classroom. There the savings in absolute time were more significant, though. :smiley:

PR #3381, opened 2022-08-05
Update embedded fmtlib code to version 9.0.0

akohlmey:
**Summary**

This updates the embedded fmtlib code to version 9.0.0.
Also simplify and update the functions with added format parsing in `utils::` and `Error::`. Use `fmt::make_format_args()` instead of the removed `fmt::make_args_checked()`.

**Related Issue(s)**

Closes #3160 since this removes the previously included workaround.

**Author(s)**

Axel Kohlmeyer, Temple U and the fmtlib developers.

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

No known issues.

**Post Submission Checklist**

- [x] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [x] The source code follows the LAMMPS formatting guidelines
- [x] Suitable new documentation files and/or updates to the existing docs are included
- [x] The feature has been verified to work with the conventional build system
- [x] The feature has been verified to work with the CMake based build system
- [x] Suitable tests have been added to the unittest tree.

PR #3383, opened 2022-08-06
Documentation fixes

hammondkd:
**Summary**

I include a bunch of minor changes (and one or two not quite as minor changes) to documentation files. My goal in most cases was to make the documentation more consistent between files so they are more easily parsed by a non-human reader, without making any substantive changes to the documentation itself.

I also found a few missing or extra keywords. In the case of "extra" ones, I tried to ensure that the word was actually supposed to be removed (i.e., it's not in the source code), but for the most part they were keywords/parameters/arguments/styles that were in the longer description but not in the lines that look like "style = *a* or *b* or *c*". I also tried to make sure words that should be literally those characters (e.g., "yes" and "no") were in italics, and words that represent values were non-italicized, which is consistent with most of the documentation.

I include one typo-fix in fix_nh_uef.cpp ("npt/npt/uef") that I caught as well.

**Author(s)**

Karl Hammond
University of Missouri
hammondkd@missouri.edu

**Licensing**

**Backward Compatibility**

There are no changes to anything that will affect inputs in this pull request.

**Implementation Notes**

I checked that the HTML files build (both with "make fasthtml" and later with "make html"). They work, though when I pulled the latest "develop" from upstream, it gives me a warning:
``
[lammps_root]/doc/src/Developer_utils.rst:211: WARNING: doxygenfunction: Unable to resolve function "logmesg" with arguments (LAMMPS*, const S&, Args&&...) in doxygen xml output for project "progguide" from directory: [lammps_root]/doc/doxygen/xml.
Potential matches:
- template<typename ...Args> void logmesg(LAMMPS *lmp, const std::string &format, Args&&... args)
- void logmesg(LAMMPS *lmp, const std::string &mesg)
``
This didn't occur until I did a branch merge today.

**Post Submission Checklist**


akohlmey:
@hammondkd thanks for your submission. Don't worry about the warning. This is due to the changes in PR #3381 and a fix is already waiting in this commit https://github.com/lammps/lammps/commit/4421a793d9d3e6ada8315ec61fde620d2044f1a9

If you enjoy doing these kind of things, you may want to join the [LAMMPS Code Clinic effort](https://www.lammps.org/workshops/Aug22/) this month, that we are starting on Monday, August 8th.

sjplimp (approved):
Thanks Karl!

hammondkd:
@akohlmey I actually submitted these now in part because I saw the Code Clinic advertised and figured it would be less messy to submit these _before_ everyone started making changes rather than during or after. I have been debating whether or not to register for the clinic, but I suppose it can't hurt!

By the way, all of these edits were found while I was debugging a program I am currently writing that attempts to auto-generate my VIM syntax highlighting files for LAMMPS (which are much more detailed than the ones in tools/vim); I have been maintaining that file by hand for the last 10 years or so, but it's become nearly impossible to maintain due to the increasing complexity of LAMMPS, so my current iteration has to be at least five years behind in certain parts. The automated one is very much a work in progress, but if it ever works properly....

PR #3388, opened 2022-08-08
Mliap bug 3204

### LINKED ISSUE #3204: [BUG] python->LAMMPS->python patterns broken on some interpreters.
**Summary**

When opening lammps using the python interface to the shared library, commands that invoke the interpreter from inside lammps can cause fatal errors.

LAMMPS Versions: Probably many, but found in at least LAMMPS (24 Mar 2022), LAMMPS (29 Oct 2020) and LAMMPS (31 Aug 2021), LAMMPS (29 Aug 2021).
Operating system: MacOS, but this probably about the build of python rather than the OS per se.
Python: Various conda builds of python, including anaconda distributions as well as conda-forge distributions.

**Expected Behavior**

This feature should hopefully work?

**Actual Behavior**

The program crashes due to a failure to acquire references to the active python interpreter.

Depending on the python build, one may receive one of several error messages. I have omitted the normal output. It appears to fail when funcs.py intends to import the lammps python module.

Segfault:
```
<...>
LOOP ARGS 10 1.0 -4.0 <capsule object NULL at 0x10a2bb720>
Segmentation fault: 11
```
Fatal Python Error:
```
<...>
LOOP ARGS 10 1.0 -4.0 <capsule object NULL at 0x7fd8a813a780>
Fatal Python error: _PyInterpreterState_Get(): no current thread state
Python runtime state: initialized
```
Free without allocation error:
```
<...>
python(28375,0x1114f9e00) malloc: *** error for object 0x10422f080: pointer being freed was not allocated
python(28375,0x1114f9e00) malloc: *** set a breakpoint in malloc_error_break to debug
 *** Process received signal ***
 Signal: Abort trap: 6 (6)
 Signal code:  (0)
 [ 0] 0   libsystem_platform.dylib            0x00007fff20994d7d _sigtramp + 29
 [ 1] 0   ???                                 0x00007fc7ba70d830 0x0 + 140495803177008
 [ 2] 0   libsystem_c.dylib                   0x00007fff208a4406 abort + 125
 [ 3] 0   libsystem_malloc.dylib              0x00007fff20784165 has_default_zone0 + 0
 [ 4] 0   libsystem_malloc.dylib              0x00007fff207872aa malloc_report + 151
 [ 5] 0   python                              0x0000000103c8ccd3 _PyObject_Free + 115

```
**Steps to Reproduce**

The easiest way to reproduce the issue is to run the following script in the /examples/python directory of lammps:
```
import lammps
lmp=lammps.lammps()
lmp.file('in.python')
```

This error seems to occur because the python executable is compiled against its static library rather than its shared library. When python is the driving code, and LAMMPS attempts to check if the python interpreter is active, `Py_IsInitialized()` returns 0, even though it should return 1.

As such, the error can be detected -without- lammps at all:
```
import sysconfig
import ctypes
library = sysconfig.get_config_vars('INSTSONAME')[0]
pylib = ctypes.CDLL(library)
if not pylib.Py_IsInitialized():
    raise RuntimeError("This interpreter is not compatible with python->lammps->python controll flow..")
```
(On some builds INSTSONAME reports the wrong name for the python shared library, but this is a separate issue.)

I have also confirmed that from within lammps, Py_IsInitialized() returns 0, 

You can likewise reproduce the segmentation fault in the interpreter by doing something nontrivial such as attempting to import a module:
```
>>> library.PyImport_ImportModule(b"collections")
Segmentation fault: 11
```
Note: within python itself, there is `ctypes.pythonapi` which -can- access the C-API of the running interpreter, so the above is not really a bug, just the wrong way of doing things.

My conclusion so far is that when loading the python shared library in a static build of python, the symbols in the shared library are distinct and not the ones used by the python interpreter. What we should want is for the calls in LAMMPS to call back into the active interpreter's symbol space.

- This issue also affects the MLIAP package, and is the reason for the inclusion of the code [here](https://github.com/lammps/lammps/blob/fd80b1bd9b3ebe95a9054ae85c6eabfac82851da/python/lammps/mliap/__init__.py#L5-L11), but I want to be clear that this problem has nothing to do with the cython extensions used in MLIAP; the above shows that it is based on how the LAMMPS python module operates.

- https://github.com/upscale-project/pono/issues/92#issuecomment-703919663 says that the sysconfig variable "Py_ENABLE_SHARED" can be used to detect a similar-sounding problem, however, I have a working python+LAMMPS on linux where Py_ENABLE_SHARED is zero, so I don't know if we can trust that. 

- https://github.com/conda-forge/python-feedstock/issues/222#issuecomment-436507431 says that maybe one can load the python executable as a shared library to get around a similar-sounding problem. I attempted this and did not find it successful - I could fool ctypes into thinking that the executable was a library, but it returned 0 for Py_IsInitialized().

- In principle scikit-build has a function called [target_link_libraries_with_dynamic_lookup](https://github.com/scikit-build/scikit-build/blob/9ae49a2527d8dbe563d7a55f220da7c278aa94af/skbuild/resources/cmake/targetLinkLibrariesWithDynamicLookup.cmake) that is possibly supposed to handle this kind of 'lazy load'. I attempted to see if is possible to patch this into lammps cmake but I get an error that I don't know how to fix. `The test project needs language C which is not enabled.` 
This is also associated with some [scikit-build machinery](https://scikit-build.readthedocs.io/en/latest/cmake-modules/PythonExtensions.html) which seems to be aimed at helping to embed python or build extensions.

- I have some old conda environments laying around from late 3.6/early 3.7 that used to work and not crash, so it's not a fundamental problem with macos.

**Some potential ways to address this**

1) If the LAMMPS python module was built as a c-extension from python, this problem will likely not occur - in this case references to the python API in the lammps module python do not come through the python shared library. While this would most likely work, it might significantly change the build process and involve refactoring how python talks to lammps. From my limited experience, it is not too difficult to wrap the calls in liblammps using cython and compile an extension module for a given set of LAMMPS compiler settings, however, automatic integration of the various compiler flags that LAMMPS uses (serial vs parallel, BIGINT size) to the setup.py file was nontrivial to me.  While such a project could have other benefits, I find it highly unclear whether this would be worth the work -- so it's probably not the right place to start.

2) Potentially, changes in how linking works: it's possible that there are some differences in the compilation flags for embedded python vs python extension modules that could be worked into our the LAMMPS build process. It could require building a separate shared compiled library to use in the lammps python module? I don't know. The scikit-build machinery or parts of it might help here. 

3) Leave it alone - decide that this use case is outside of the scope of LAMMPS, at least for now.
3b) Raise an error or warning for the user ahead of crashing: One could add some state to the liblammps library which the python interpreter can set when opening lammps, when a command that requires the python interpreter runs, lammps can detect the flag is set and raise an error or warning instead of just dying - similar to the warning in lammps.mliap, but enforcing it in all scenarios where this can be a problem (through lammps/src/pythonimpl)

@rbberger Is this something you could take a look at?

@junghans Any idea if `target_link_libraries_with_dynamic_lookup` is something that might help? 

Thanks to @jan-janssen and @athomps for helping to confirm.

chemshift:
**Summary**

"import lammps.mliap" now succeeds with the python distributions from conda-forge
Tested on Linux and Windows with an Anaconda Environment and python=3.10.5 from conda-forge.

**Related Issue(s)**

fixes #3204 

**Author(s)**

Dionysios Sema, MIT | dsema@mit.edu

**Licensing**

**Backward Compatibility**

No other features affected.

**Post Submission Checklist**


akohlmey:
@lubbersnick can you please check this out?

akohlmey:
@chemshift thanks for your submission. I think part of the solution is that you are using Python 3.10.5. There have been issues with sub-interpreters for multiple 3.10.x versions that also affected several other software packages. The one where I was affected was "magically" corrected after python was updated to 3.10.5.

chemshift:
Hi @akohlmey. I also tested this with python=3.10.0, 3.9.13 and 3.8.10 (which is an older distribution) and all of these versions work on Linux. I think this has more to do with how python is compiled and distributed in the repos.

I asked @rohskopf to test this and it didn't work on his Mac (python v3.10.5), but was good on his Linux machine. Unfortunately, Mac users will have to compile python from source. If the repos provide appropriate builds in the future though, this code should work for all OSes.

The main problem is actually in lines 11-12:
```
if pylib.endswith(".a"):
        pylib.strip(".a") + ".so"
        pylib = ctypes.CDLL(library)
```
"pylib" doesn't change the name of "library" so it is still trying to load the static library. If the above is replaced with:

```
if library.endswith(".a"):
        library = library.strip(".a") + ".so"
        pylib = ctypes.CDLL(library)
```
it fixes the problem on Linux. The reason I changed the other parts of the code is because the key: 'INSTSONAME' doesn't exist on Windows, and doing it this way loads the mliap package for Windows too.

athomps (approved):
I approve

lubbersnick:
I tried to start a review but I am not sure if it is showing up here for others since I didn't see it from a non-logged in computer. My concern is with the blanket Except clause which doesn't handle all errors. Reposting:

I think this should be raising an OSError with a specific error message from whatever exception is caught here, no?

Something like:

```
except Exception as e:
    raise OSError("Unable to locate python shared library") from e
```

The current code doesn't do anything with the OSError.

An alternative would be to omit this blanket except clause.

@chemshift what do you think?

chemshift:
@lubbersnick Addressed.
The except clause should be there, since if the library is not found, it will raise an OSError anyway.

PR #3392, opened 2022-08-11
Update test for neighbor list settings during minimization

akohlmey:
**Summary**

This updates the check for when to change the neighbor list setting during minimization to accept `neigh_modify every 1 delay 0` regardless of the "check" setting.  The "check no" setting would be less efficient, but should be accepted while it is currently overwritten. There may be cases where the delayed neighbor list rebuild is unwanted (see link below).

**Related Issue(s)**

See https://matsci.org/t/minimization-changes-neighbor-modify-lammps-29-sept-2021/43879

**Author(s)**

Axel Kohlmeyer, Temple U.

**Licensing**

**Backward Compatibility**

No known issues.

**Post Submission Checklist**

### DIFF
--- a/doc/src/minimize.rst
+++ b/doc/src/minimize.rst
@@ -50,15 +50,17 @@ and forces) by pushing the atoms off of each other.
    The distance that atoms can move during individual minimization steps
    can be quite large, especially at the beginning of a minimization.
    Thus `neighbor list settings <neigh_modify>` of *every = 1* and
-   *delay = 0* are **required**.  This may be combined with *check = no*
-   (always update the neighbor list) or *check = yes* (only update the
-   neighbor list if at least one atom has moved more than half the
-   `neighbor list skin <neighbor>` distance since the last
+   *delay = 0* are **required**.  This may be combined with either
+   *check = no* (always update the neighbor list) or *check = yes* (only
+   update the neighbor list if at least one atom has moved more than
+   half the `neighbor list skin <neighbor>` distance since the last
    reneighboring).  Using *check = yes* is recommended since it avoids
    unneeded reneighboring steps when the system is closer to the minimum
-   and thus atoms move only small distances.
+   and thus atoms move only small distances.  Using *check = no* may
+   be required for debugging or when coupling LAMMPS with external
+   codes that require a predictable sequence of neighbor list updates.
 
-   If if the settings are **not** *every = 1* and *delay = 0*, LAMMPS
+   If the settings are **not** *every = 1* and *delay = 0*, LAMMPS
    will temporarily apply a `neigh_modify every 1 delay 0 check yes
    <neigh_modify>` setting during the minimization and restore the
    original setting at the end of the minimization.  A corresponding
--- a/doc/src/minimize.rst
+++ b/doc/src/minimize.rst
@@ -32,30 +32,49 @@ Description
 Perform an energy minimization of the system, by iteratively adjusting
 atom coordinates.  Iterations are terminated when one of the stopping
 criteria is satisfied.  At that point the configuration will hopefully
-be in local potential energy minimum.  More precisely, the
+be in a local potential energy minimum.  More precisely, the
 configuration should approximate a critical point for the objective
 function (see below), which may or may not be a local minimum.
 
-The minimization algorithm used is set by the
-:doc:`min_style <min_style>` command.  Other options are set by the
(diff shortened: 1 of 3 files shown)


stanmoore1 (commented):
This behavior is a little unpredictable. For example, a user will get a different neigh check setting if they use `every 5 delay 0 check no` vs `every 1 delay 0 check no` for minimization.

stanmoore1:
It may be more predicable to leave the neigh check setting unchanged in either case.

akohlmey:
> It may be more predicable to leave the neigh check setting unchanged in either case.

That is why a message is printed if the setting is changed.

For minimization "check yes" usually is a good idea since the size of the steps can change much more than during MD so you are not wasting a lot of time with neighbor list rebuilds when you are close to the minimum. But that is also why "every 1" and "delay 0" are required because at the beginning you may need to update the neighbor lists in every step.

The only change in behavior that this PR is introducing is that it accepts `every 1 delay 0 check no` as acceptable on top of the "check yes" version. This would mostly be for debugging so that you *can* enforce neighbor list rebuilds.

stanmoore1:
As a user, I would be confused why the code is changing my neigh check setting in one case but not the other. The message saying it is changing the setting would not help this confusion.

akohlmey:
Lets wait for @sjplimp to weigh in on this. I just checked with the manual and see that this behavior is not discussed in the manual currently. So at the very least, we also need to add an explanation to the manual, regardless of which behavior is implemented.

akohlmey:
@sjplimp @stanmoore1 @athomps @rbberger While messing with the neighbor list settings and reading through the related documentation it occurred to me that it would be "safer" to have as a default setting every = 1, delay = 0, check = yes instead of every = 1, delay = 10, check = yes.
That would a) avoid the message about changing the neighbor list settings for people that *only* want to run minimization and thus reduce confusion in inexperienced users b) would be more conservative and thus avoid dangerous builds and thus incorrect simulations by inexperienced users. The message about dangerous builds is easily overlooked. Assuming that people would choose an optimal time step, a choice of delay = 10 is *very* aggressive. For example the "melt" example reports 25 dangerous builds with this setting. Even with delay = 5 there are 8 dangerous builds. The peptide example reports 10 dangerous builds and no with delay 5, same for the rhodo input. A choice of delay = 10 would be acceptable for those two cases if you don't use fix shake and thus have to reduce the timestep significantly.
I think in general it would be better to have conservative defaults and expect the user to tweak the neighbor list settings for optimal performance, rather than using more aggressive settings and expect the user to adjust them to avoid inaccurate simulations due to missed neighbors from too infrequent neighbor list updates. Specifically inexperienced users and often also users with some experience will not be aware of that.

sjplimp:
@akohlmey I think we should discuss whether to change delay = 10 to delay = 0 at our next developer mtg.  Many trade-offs.
For this PR, I think allowing the user to override the minimize neigh list setting (for check) is fine, so long as documented.  Also, clarifying in the doc page how/why minimize changes neigh list settings is good.  I think choosing settings for speed over 100% safety is more important for dynamics, the opposite for minimization.

doc/src/minimize.rst
```diff
@@ -32,30 +32,49 @@ Description
...
+   The distance that atoms can move during individual minimization steps
+   can be quite large, especially at the beginning of a minimization.
+   Thus `neighbor list settings <neigh_modify>` of *every = 1* and
+   *delay = 0* are **required**.  This may be combined with *check = no*
+   (always update the neighbor list) or *check = yes* (only update the
+   neighbor list if at least one atom has moved more than half the
+   `neighbor list skin <neighbor>` distance since the last
+   reneighboring).  Using *check = yes* is recommended since it avoids
+   unneeded reneighboring steps when the system is closer to the minimum
+   and thus atoms move only small distances.
```
sjplimp: Add a sentence about why you might want to use check = no ?
Only for debugging in some context ?  I think that if every=1, the
utility of check = no is minimal.  Because you will (over time) get different
answers for check = no/yes, due to re-ordering atoms.

sjplimp (changes requested):
one suggestion on a doc page clarification

doc/src/minimize.rst
akohlmey: done

sjplimp (approved):

athomps (commented):
I approve.

PR #3396, opened 2022-08-12
Expanding the Fortran interface

hammondkd:
**Summary**

I have added implementations of get_thermo, extract_box, reset_box, memory_usage, get_mpi_comm, extract_setting, and extract_global to the Fortran API ([lammps_root]/fortran/lammps.f90).

**Author(s)**

Karl Hammond
University of Missouri
hammondkd at missouri.edu

**Licensing**

**Backward Compatibility**

This should not change any existing functions, only provide new ones.

**Implementation Notes**

I compiled this using GNU Fortran 11.3.1 using the "-std=f2003" and Intel Fortran 18.0.3 using the "-stand=f03" flag; no errors or warnings resulted.

This implementation of extract_global uses a derived type called "lammps_data" that contains pointers to a variety of data types, kinds, and ranks. The assignment operator is then overloaded so that when the user says "[pointer to integer(c_int)] = [type(lammps_data)]", the pointer is type/kind/rank-checked and then either exits with an error or associates that pointer with the internal LAMMPS data. A warning message has been added to the documentation cautioning users against modifying anything extract_global sends back to them.

Getting data back from LAMMPS required me to define constants such as LAMMPS_INT and LAMMPS_DOUBLE to match their definitions in library.h. This is somewhat of a maintenance nuisance (they must be maintained in library.h, constants.py, and lammps.f90), but I don't see a way around it short of including those constants in a header and then importing that header with the preprocessor.

Given that lammps_extract_global and lammps_extract_global_datatype currently tell me nothing about whether the pointer is to a scalar or a vector (and this makes a difference in Fortran), I had to handle the three-dimensional ones (boxlo, boxhi, sublo, subhi, sublo_lambda, subhi_lambda, and periodicity) as special cases. It might be helpful if library.h contained a function that returned this information for use in both Fortran and for Python, but I didn't write this function presently in case there's something I don't know about. I might later when/if we get around to extract_fix or extract_compute.

**Post Submission Checklist**


hammondkd:
Should I be putting `   .. versionadded:: TBD` on all of the new functions?

akohlmey:
> Should I be putting `   .. versionadded:: TBD` on all of the new functions?

Yes, that is a good idea. I have started adding those markers so that people reading the online manuals are less confused about errors when they use an older version of LAMMPS.

athomps (approved):
I approve. Thanks @hammondkd

akohlmey (approved):

PR #3399, opened 2022-08-14
Refactor access to molecule templates

akohlmey:
**Summary**

This pull request modernizes the access to molecule data structures as loaded into LAMMPS with the molecule command.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

Currently there are some inconsistencies in how data referred to by a molecule ID is processed. Also it requires to access internal data of the molecule data structure to determine whether one or more entries in the list of molecules needs to be used.
The intent of this pull request is to make this consistent and use C++ data structures instead of accessing internal lists.

**Post Submission Checklist**

### DIFF
--- a/src/GRANULAR/fix_pour.cpp
+++ b/src/GRANULAR/fix_pour.cpp
@@ -131,7 +131,7 @@ FixPour::FixPour(LAMMPS *lmp, int narg, char **arg) :
 
       if (atom->molecular == Atom::TEMPLATE && onemols != atom->avec->onemols)
         error->all(FLERR, "Fix pour molecule template ID must be same as atom style template ID");
-      onemols[i]->check_attributes(0);
+      onemols[i]->check_attributes();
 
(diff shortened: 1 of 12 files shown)


akohlmey:
@sjplimp This is the minimal change adding the new access function and applying it in very select cases. This now needs some thought on how far to take this and serve the multiple purposes of molecule IDs as outlined in my private email. I am making it public so you can look at the changes directly instead of me describing them to you.

I think the first question to address is the issue I mentioned about "atom style template" being different and requiring a different process. In other words, should we add a "template" command for it, or try to set it up to accept multiple molecule IDs as an argument (so that we can disallow loading multiple molecule files "into" the same molecule ID and simplify that processing), or keep this as is and then require more complex code to access molecules as it currently exists or based on the new accessor function in this pull request.

sjplimp (approved):

PR #3405, opened 2022-08-16
Support for distributed grids

### LINKED ISSUE #2269: [Feature Request] Implement dump style(s) for grid based data
**Summary**

For fixes and computes that compute properties on a spatial grid, there are currently several *ad hoc* file formats. Some of those are tricky to parse in a generic fashion, since they require knowledge from the input file. If the was a "grid" dump style, the output could be funneled to a generic interface which could then also implement multiple output file formats.

**Detailed Description**

Possible applications:
- output from chuck based spatial binning
- output from fix ttm and ttm/mod
- output of potential as computed in pppm

VTK is widely used in for storing/visualizing such data, so using the "native" VTK format would be a natural choice. This is all the more useful, as those could be converted to other formats with the help of VTK. However, the VTK libraries have many dependencies and may be complex to set up and install for just outputting one specific type of data. It may thus be worth considering to write a "mini-VTK" library that would only need to be able write (not read) data on a regular grid (plus some metadata) and only need to output a limited selection of kinds of data. The formats are well documented. It would be beneficial to do the implementation so that output can be done efficiently with distributed data without having to collect/buffer all data on MPI rank 0.

Resources about the VTK file formats:
- https://lorensen.github.io/VTKExamples/site/VTKFileFormats/
- https://vtk.org/wp-content/uploads/2015/04/file-formats.pdf

sjplimp:
**Summary**

Refactor the GridComm class into Grid2d and Grid3d classes which partition and communicate distributed grid data across and between processors for 2d and 3d simulations.  "Distributed" means the grid cells or points and the data they store are spread across processors.  Each processor owns the points within its subdomain.  Note that Grid2d/3d do not store any per-grid data.  The data is stored by the calling classes which use Grid2d/3d.

Existing classes that used distributed grids now use these grid classes:  PPPM, MSM, fix ttm/grid, pair amoeba/hippo.  

Added additional infrastructure to allow computes and fixes to create and output multiple grids (per compute or fix) and multiple data fields on each grid.  Added a few new commands to work with distributed per-grid data:

compute property/grid
fix ave/grid
dump grid
dump grid/vtk

Fix ave/grid can either time average per-atom data within each grid cell, or time-average per-grid data directly.  The former use case is similar to fix ave/chunk for regular 2d or 3d bins, except that that all processors own a copy of all the bins. So it is less scalable than the new fix ave/grid.

Also added options to dump image and dump_modify to color grid cells in output viz images by grid data.

The motivation for this PR is to standardize how per-grid data is used within LAMMPS and add infrastructure that makes it easier for developers to create their own pair styles or other models which need to use grids in conjunction with particles.

New commands being thought about for future PRs:
compute fft/grid
fix grid (internal command for use by other styles)
others to suggest ?

See the "Howto grid" doc page for an overview of per-grid (distributed) data.

**Related Issue(s)**

Implements and thus closes #2269 

**Author(s)**

Steve Plimpton, SNL

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**


sjplimp:
@akohlmey This still needs testing and debugging.  But there are 4 initial things you could look at:
(1) utils::grid_parse() method - maybe could be done better?
(2) wildcarding for grid ID columns, as used in fix ave/grid and dump/grid, does not appear to work
(3) use of templating in compute property/grid - I'm not sure if the "if" tests in the 2 templated methods are sufficiently simple to enable the compiler to create minimial-code versions of each instance of the template - I may be breaking some rules for use of template variables?
(4) use of extractor methods each compute or fix defines to "publish" its grid data to consumer classes

The Howto_grid.rst doc page is a good place to start for a hi-level overview, Likewise Howto_output.rst now talks about per-grid data as well.

akohlmey:
@sjplimp FYI, this currently fails compiling due to the DIELECTRIC package.

akohlmey:
> (1) utils::grid_parse() method - maybe could be done better?

yes. this has a memory leak and could be replaced with something like:
```
auto words = Tokenizer(name,":").as_vector();
if (words.size != 3) error->all(FLERR, ....);
```
with making this a `std::vector<std::string>` these are all automatic local memory allocations and do not require to be freed when they go out of scope.

Do you still want to keep this as a (2 line) function in utils or just do this inline?

sjplimp:
it's used several places, so I think it is good to be a utils method - you can change or rename it as you wish
to be consistent with utils - maybe gridID_parse() is more descriptive, as I think of the colon separated string as a gridID

sjplimp:
@akohlmey fixed the issue with DIELECTRIC

akohlmey:
> @akohlmey fixed the issue with DIELECTRIC

Unfortunately, there is more. Now it fails in the ELECTRODE package. :sob:

sjplimp:
@akohlmey ok, think I found all the package instances of GridComm --> Grid3d - except KOKKOS - @stanmoore1 will have to look at that one

akohlmey:
@sjplimp task (1) is done. The function is now called utils::gridid_parse().
FYI in the places where the function is currently used I have added const references to the "words" vector element returned from the call to keep the code readable (with words[0], words[1], and words[2] it would be easy to mix things up, already gname/dname is borderline).

As a general strategy: if strings used as function arguments are certain to not be null pointers, we should always try to use `const std::string &` as function argument. That makes for better readable string comparisons inside and is compatible with passing either a `char *` or `std::string` (const or non-const).

akohlmey:
@sjplimp task (2) is done. I also fixed a ton of small issues (spelling, typesetting, broken links) in the documentation.

Please note that `fix ttm/grid` was broken because it was missing the "override" keywords on the virtual functions.
Without override it would still compile despite changes in the function argument types in the base class and then the dummy functions of the base class were called instead of the ones in the `fix ttm/grid` class.

akohlmey:
> (3) use of templating in compute property/grid - I'm not sure if the "if" tests in the 2 templated methods are sufficiently simple to enable the compiler to create minimial-code versions of each instance of the template - I may be breaking some rules for use of template variables?

It is not about simplicity, but whether the result is known at compile time. In this case it is.
In addition, you could also make "dimension" and "nvalues" template parameters. The compiler will then create more function instances, but each will be smaller and only contain the collection of if branches active for a given set of parameters.

akohlmey:
> (4) use of extractor methods each compute or fix defines to "publish" its grid data to consumer classes

Not sure what you need from me in this case. Can you give me some more hints?
I probably would have created separate functions for 2d and 3d grids so that you won't need the typecast.

sjplimp:
Re (3): thanks - I'll think about making dimension and/or nvalues a template param as well - if it also required that code which invokes the template function must use template params that are known at compile time, e.g. not the variable "dimension"

sjplimp:
Re (4) - I just meant to check if the 4 extractor functions I added seem aligned with your other recent refactorings for avoiding requestors needing to access class data directly.  Each compute or fix that creates a Grid class needs to implement the 4 funcs.
4 already seemed a lot, so I didn't want more for 2d for 3d.

akohlmey:
> Re (3): thanks - I'll think about making dimension and/or nvalues a template param as well - if it also required that code which invokes the template function must use template params that are known at compile time, e.g. not the variable "dimension"

Yes, you need to enter a constant in the values. So all you gain is basically moving the runtime tests to a higher level. I think in performance this will not make a difference. The primary benefit would be if you call those functions from regions of code where you already have branched on a test on `dimension == 2` or `nvalues == 1`. The performance benefit is when you have the if statement inside loops, which is the case that you already have covered. So it depends on the calling code whether it is worth doing.

akohlmey:
> Re (4) - I just meant to check if the 4 extractor functions I added seem aligned with your other recent refactorings for avoiding requestors needing to access class data directly. Each compute or fix that creates a Grid class needs to implement the 4 funcs. 4 already seemed a lot, so I didn't want more for 2d for 3d.

Unfortunately, this is one of the things that you can really only have a qualified opinion on after making significant use of it.
For now, it is close enough.

I am more concerned about the "struct of arrays" situation versus using an "array of structs" in fix ave/grid (they are a common programming pattern currently in similar fixes and dump styles).
Right now we have code blocks like this:
```
  which = new int[nargnew];
  argindex = new int[nargnew];
  ids = new char*[nargnew];
  value2index = new int[nargnew];
  value2grid = new int[nargnew];
  value2data = new int[nargnew];
```
All of these arrays need to be kept in sync and managed. Instead we could define a struct that contains all properties for a single items. Then there is only one array to be managed. ...and if we ever want to add a property, we only need to add it to the struct and not add code to manage its array all over the place.

I have tried this in fix vector (which seemed to be the file with the fewest constraints) and now it has a `struct value_t` and stores everything in a std::vector<value_t>.

sjplimp:
@akohlmey This is now close to complete, so ready for regression testing if you want to trigger that.  There are some loose ends I will still finish.  Also a few non-implemented future features.  But they can be part of a future PR.

akohlmey:
> @akohlmey This is now close to complete, so ready for regression testing if you want to trigger that. There are some loose ends I will still finish. Also a few non-implemented future features. But they can be part of a future PR.

Unless this compiles with KOKKOS, there is not much testing that will run.
Also it has merge conflicts with some of the recently updated documentation that need to be resolved.

sjplimp:
@akohlmey OK - I think this is now ready for you to look over a final time and then merge

akohlmey:
> @akohlmey OK - I think this is now ready for you to look over a final time and then merge

@sjplimp this still fails to compile with KOKKOS enabled. I cannot merge unless that is resolved and thus there is not much point in reviewing before that.

sjplimp:
@stanmoore1 Can you take a look at the Kokkos issues?  Thanks

sjplimp:
@akohlmey Also, I like your struct of arrays suggestion above.  But I thought it should be a later PR and could include other
fixes or computes with similar patterns.

sjplimp:
@akohlmey I re-labeled this work-in-progress, b/c I have started adding methods to allow the distributed grid to be rebalanced.
It's incomplete, so won't be ready to merge yet even if Stan addresses the Kokkos issues.

stanmoore1:
@sjplimp should I work on porting this to Kokkos, or wait for your rebalance changes?

sjplimp:
@stanmoore1 I think those 2 things are independent so you can work on the Kokkos part

stanmoore1:
@sjplimp I finished porting this to Kokkos.

sjplimp:
@akohlmey @stanmoore1 This PR is nearly complete now I think.  I see there is an issue with the ELECTRODE pkg PPPM which I will look at.  It would be good for you to run this new version on any scripts (unit tests, examples, others) that exercise PPPM and MSM flavors.  I also need to add a few items to the doc pages.  And make a new examples dir.  But I think this is close enough for Axel to take a look at the coding and Stan for any Kokkos issues with this newer version.  Thanks

sjplimp:
@srtee Hi Shern - I needed to make some mods to pppm_electrode.cpp to make it work with the refactored Grid3d.cpp in this PR. Can you check your tests cases for the ELECTRODE package to see if I've broken any thing, or changed any answers.  The methods in pppm_electrode.cpp I changed were mostly allocate() and deallocate() and init() - thanks

srtee:
@sjplimp all my ELECTRODE examples work fine as of commit f684cd5, thanks!

sjplimp:
@stanmoore1 I think the unit test errors are all Kokkos related.  I have tested the PPPM and MSM variants with my own scripts.  You should test them with yours and also any Kokkos tests you have for KSpace.

sjplimp:
@stanmoore1 @akohlmey I have finished all the code + doc work for this PR. I think the only thing lacking is any needed Kokkos changes.  Also I will add a small new examples/grid dir in the next day or so.

sjplimp:
@akohlmey @stanmoore1 I added the new examples dir plus an updated fix ttm/grid example in the examples/ttm dir.  When Stan says the Kokkos part is done, this PR is ready to be considered for merge.

stanmoore1:
@sjplimp thanks I will take a look.

sjplimp:
@akohlmey I don't see this conflict with pppm_dielectric.cpp in my local branch, so not sure how to fix it ?  I pulled this PR into
my local branch before my latest commit (to dump image), and it created no conflicts.

akohlmey:
> @akohlmey I don't see this conflict with pppm_dielectric.cpp in my local branch, so not sure how to fix it ? I pulled this PR into my local branch before my latest commit (to dump image), and it created no conflicts.

The conflict was between the `distributed-grids` branch and the current `develop` branch. Github checks whether one can be merged into the other. You had only merged changes from Stan to `distributed-grids` on github.

I have fixed the issue.

The workflow for resolving such conflicts is:
```
git checkout develop
git pull
git checkout distributed-grids
git pull
git merge develop
# resolve eventual merge conflicts
# followed by git add / git commit
git push
```

sjplimp:
thanks @akohlmey

stanmoore1:
@sjplimp FYI, started porting this to Kokkos yesterday. I think I'm about 2/3 done.

sjplimp:
@akohlmey I found one other bug today with fix ave/grid, so even if Stan finishes, please don't merge yet

sjplimp:
@akohlmey ok, fixed some valgrind issues - I think this is ready to merge when Stan is finished with Kokkos changes

stanmoore1:
I think this is working now for Kokkos. I'm running regression tests now.

stanmoore1:
Internal regression tests passed. This is ready to merge from my POV.

stanmoore1 (approved):

PR #3406, opened 2022-08-17
Fix vim syntax to handle switching color themes

eskalVAR:
**Summary**
- filetype.vim includes *.in files as lammps input files
- syntax.vim defines the syntax links rather than just creating them once

This fix allows switching color schemes in vim while retaining lammps syntax highlighting which was previously not the case. 

**Author(s)**
Eryk Skalinski : eskalinski at protonmail dot com

**Licensing**

**Backward Compatibility**
No

**Implementation Notes**
Please see summary

**Post Submission Checklist**

### DIFF
--- a/tools/vim/filetype.vim
+++ b/tools/vim/filetype.vim
@@ -1,4 +1,5 @@
 augroup filetypedetect
  au! BufRead,BufNewFile in.*           setfiletype lammps
  au! BufRead,BufNewFile *.lmp          setfiletype lammps
+ au! BufRead,BufNewFile in.*           setfiletype lammps
 augroup END
--- a/tools/vim/filetype.vim
+++ b/tools/vim/filetype.vim
@@ -1,5 +1,5 @@
 augroup filetypedetect
  au! BufRead,BufNewFile in.*           setfiletype lammps
  au! BufRead,BufNewFile *.lmp          setfiletype lammps
- au! BufRead,BufNewFile in.*           setfiletype lammps
+ au! BufRead,BufNewFile *.in           setfiletype lammps
 augroup END
--- a/tools/vim/README.txt
+++ b/tools/vim/README.txt
@@ -3,50 +3,50 @@
 
 The files provided in this directory will enable syntax highlighting
 for the lammps script syntax in vim. The simulation scripts have to
-end on *.lmp or start with in.* (see mysyntax.vim).  
-By far not all commands are included
-in the syntax file (lammps.vim). You can easily add new ones.
+end on .lmp or .in or start with in. (see mysyntax.vim or filetype.vim).
+By far not all commands are included in the syntax file (lammps.vim).
+You can easily add new ones.
 
 =To enable the highlighting (compatible with old versions of VIM):
 ============================
-(0)   Create a ~/.vimrc
+(1)   Create a ~/.vimrc
       You can have a look in /usr/share/vim/vim*/vimrc_example.vim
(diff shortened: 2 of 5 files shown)


tools/vim/filetype.vim
```diff
@@ -1,4 +1,5 @@
 augroup filetypedetect
  au! BufRead,BufNewFile in.*           setfiletype lammps
  au! BufRead,BufNewFile *.lmp          setfiletype lammps
+ au! BufRead,BufNewFile in.*           setfiletype lammps
```
akohlmey: This added line seems redundant

akohlmey:
@Vikingat-RAGE do you have a sense for what is the minimum VIM version required for your change?

tools/vim/filetype.vim
eskalVAR: Sorry it was meant to be *.in, I was copying my local changes over to a fork.

eskalVAR:
> @Vikingat-RAGE do you have a sense for what is the minimum VIM version required for your change?

I believe 7.2

akohlmey:
@Vikingat-RAGE please leave the reviewer assignments. We should have proper VI users do the review. I am at heart a die-hard emacs user. :grimacing: 
Also we have the policy that at least two core LAMMPS developers should have "looked" at pull requests. Since I am the assignee and will be merging this, I am one, but we need one more...
Thanks for understanding.

athomps:
I'm also vi semi-literate

hammondkd:
My two cents, though I'm hardly a "core developer" of course:

The change to filetype.vim merely makes it so "foo.in" will be recognized as a LAMMPS input file, just as "in.foo" and "foo.lmp" currently are.

The changes to lammps.vim are a good idea, though someone should probably replace the tabs with spaces.

The reason it's a good idea: setting
```
    highlight link lammpsOutput Function
```
links the styles whether they are already set or not, while
```
    highlight default link lammpsOutput Function
```
only makes the link if it doesn't already exist. If a user adds
```
    highlight link lammpsOutput Underlined
```
into his/her .vimrc, say, the "default" keyword in lammps.vim lets the user's preference override what is in the syntax file. It also, as @Vikingat-RAGE mentions, allows the user to change the color scheme without having to re-load the syntax highlighting file (though I had not thought about that side effect until reading this).

Note also that VI allows abbreviations of everything, hence its name (shortening of "visual [mode]"), so "hi def" and "highlight default" do the same thing in this context.

The changes suggested here should work in any recent version of VIM, possibly all the way back to 5.0 when syntax highlighting was introduced. Certainly it will work in 7.0 and greater.

-Karl Hammond [VIM user for 18 years or so]

akohlmey:
> The changes to lammps.vim are a good idea, though someone should probably replace the tabs with spaces.

done. I've also updated the README file and added a reminder to use automatic tab expansion.
There simply cannot be enough reminders considering that "tabs are evil(tm)" :wink:

akohlmey:
> [VIM user for 18 years or so]

That is more than sufficient to offer you additional privileges in the repository.

hammondkd:
@akohlmey If you want, you can add the following as the last line of tools/vim/lammps.vim:
```
   " vim: set expandtab
```
This sets a "modeline" that tells VIM to expand tabs to spaces when editing this file, even if it's not set in that user's vimrc file. (This can be shortened to "vim: set et" if desired.)

sjplimp (approved):

PR #3261, opened 2022-05-11
Lammps 'MEAM' reference structure for 'sc' (simple cubic)

HhhhZhou:
**Summary**

With the recent publication of bismuth potential, the simple cubic was added as the MEAM reference structure. This code is the modified MEAM package based on the stable LAMMPS version released on 29 September 2021.

**Related Issue(s)**

MEAM interaction for Bi.

**Author(s)**

Henan Zhou hz185@msstate.edu
Doyl Dickel doyl@me.msstate.edu
Sungkwang Mun sungkwan@cavs.msstate.edu

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

Yes, the modification was only added to the MEAM package.

**Implementation Notes**

Follow the steps below to compile the package.

1. Download the most recent stable version. The version that we tested is 29 September 2021 (https://github.com/lammps/lammps/releases/tag/stable_29Sep2021).

2. Replace the files in LAMMPS src/ folder with the files extracted from the link above.
3. Type "make yes-meam" in the command line to enable the package.

4. Type "make pu" to reflect the changes in the source file.
5. Compile with "make <makefile name> -j <number of processors>". The detail of the procedures are specified in the following page. https://docs.lammps.org/Build_make.html

**Post Submission Checklist**

- [x] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [x] The source code follows the LAMMPS formatting guidelines
- [x] Suitable new documentation files and/or updates to the existing docs are included
- [x] The added/updated documentation is integrated and tested with the documentation build system
- [x] The feature has been verified to work with the conventional build system
- [x] The feature has been verified to work with the CMake based build system
- [x] Suitable tests have been added to the unittest tree.
- [x] One or more example input decks are included

**Further Information, Files, and Links**

[MEAM_BQZ_SC.zip](https://github.com/lammps/lammps/files/8666694/MEAM_BQZ_SC.zip)

[test files.zip](https://github.com/lammps/lammps/files/8666753/test.files.zip)

PR #3409, opened 2022-08-18

akohlmey:
**Summary**

This pull request contains multiple small changes and bugfixes

**Related Issue(s)**

Includes PR #3412 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- additional Python 2/3 compatibility porting for the dump.py and log.py files from Pizza.py, synchronize the two copies in the tree through symbolic links, add unit tests
- silence compiler warnings
- "resurrect" sc lattice support and Bi parameters from unfinished PR #3261, synchronized with PR #3412 also port to KOKKOS
- bug fixes for DIELECTRIC package:
  - temporarily revert to unscaled charges before writing data file with atom style dielectric
  - correctly tally stress and energy with newton pair on/off considering we are using a full neighbor list

**Post Submission Checklist**


python/examples/pizza/dump.py
```diff
@@ -288,9 +299,9 @@ def read_all(self):
 
     # if snapshots are scaled, unscale them
 
-    if (not self.names.has_key("x")) or \
-       (not self.names.has_key("y")) or \
-       (not self.names.has_key("z")):
+    if (not "x" in self.names) or \
```
rbberger: A more Python way of writing this would be `("x" not in self.names)`.

python/examples/pizza/dump.py
```diff
@@ -572,10 +580,10 @@ def owrap(self,other):
   def names2str(self):
     ncol = len(self.snaps[0].atoms[0])
     pairs = self.names.items()
-    values = self.names.values()
     str = ""
     for i in range(ncol):
-      if i in values: str += pairs[values.index(i)][0] + ' '
+      for p in pairs:
```
rbberger: `for k, v in pairs:`

rbberger (commented):
The python changes look ok, overall could be improved as mentioned in the comments. Removing the decode() should be fine, but maybe be explicit about opening the file in text read mode.

python/examples/pizza/dump.py
```diff
@@ -707,15 +716,15 @@ def set(self,eq):
     list = re.findall(pattern,eq)
 
     lhs = list[0][1:]
-    if not self.names.has_key(lhs):
+    if not lhs in self.names:
```
rbberger: same here

python/examples/pizza/dump.py
```diff
@@ -727,7 +736,7 @@ def set(self,eq):
 
   def setv(self,colname,vec):
     print("Setting ...")
-    if not self.names.has_key(colname):
+    if not colname in self.names:
```
rbberger: and here

python/examples/pizza/dump.py
```diff
@@ -765,7 +774,7 @@ def clone(self,nstep,col):
 
   def spread(self,old,n,new):
     iold = self.names[old]
-    if not self.names.has_key(new): self.newcolumn(new)
+    if not new in self.names: self.newcolumn(new)
```
rbberger: and here

rbberger:
My comment about read text mode is about what happens during `next()`. You'll likely run into issues with `rb` there when using multiple files and `read_snapshot` without `decode()`

hammondkd (changes requested):
The line
```
vim: set expandtab
```
should be
```
" vim: set expandtab
```
The line is a comment; it's not interpreted by the program that actually runs the file (which in this case is VIM when loading syntax highlighting). It's only acted on by VIM when it's editing the file. You can do the same trick in C++, Python, Fortran, etc. files to tell VIM "I want certain settings every time I edit this file."

hammondkd (approved):
Approving the tools/vim/lammps.vim two-character edit that I suggested (and only that one).

unittest/python/python-pizza.py
```diff
@@ -0,0 +1,149 @@
+import os
+import sys
+import unittest
+from lammps import lammps
+
+EXAMPLES_DIR=os.path.abspath(os.path.join(__file__, '..', '..', '..', 'examples'))
+PIZZA_DIR=os.path.abspath(os.path.join(__file__, '..', '..', '..', 'tools', 'python', 'pizza'))
```
rbberger, line 7: @akohlmey What do you think about making `pizza` a real python package with its own `setup.py`? And then change its usage to:
```python
from pizza import dump, log
```

unittest/python/CMakeLists.txt
```diff
@@ -108,6 +108,11 @@ if(Python_EXECUTABLE)
            WORKING_DIRECTORY ${EXECUTABLE_OUTPUT_PATH})
   set_tests_properties(PythonFormats PROPERTIES ENVIRONMENT "${PYTHON_TEST_ENVIRONMENT}")
 
+  add_test(NAME PythonPizza
+           COMMAND ${PYTHON_TEST_RUNNER} ${CMAKE_CURRENT_SOURCE_DIR}/python-pizza.py -v
+           WORKING_DIRECTORY ${EXECUTABLE_OUTPUT_PATH})
+  set_tests_properties(PythonPizza PROPERTIES ENVIRONMENT "${PYTHON_TEST_ENVIRONMENT}")
```
rbberger, line 114: How about extending `PYTHONPATH` in `PYTHON_TEST_ENVIRONMENT` to include the pizza directory? Instead of adding the relative path in `python-pizza.py`?

unittest/python/python-pizza.py
akohlmey: I don't really want to sink more time into this. It already is more than I wanted.
I would suggest to do this in the https://github.com/lammps/pizza repository. The Python 3 changes here need to be carried over there anyway. But you'll have to discuss this with @sjplimp. So far the design goal has been to make this more a collection of bits an pieces that can be used as needed, but without being a regular package.

The other idea that had been kicked around would be to move the pizza folder(s) and content to `python/lammps/pizza` and make it part of the LAMMPS module.

As mentioned above, I think both exceeds the scope of this PR and is a job for somebody else with more interest in Python programming and LAMMPS pre/post processing.

unittest/python/CMakeLists.txt
akohlmey: Could be done. But on the other hand, this is closer to the suggested way of using the Pizza.py files.

sjplimp (approved):

Issue #3371, opened 2022-07-29
LAMMPS Code Clinic 2022


akohlmey:
Hi all,

We will be holding a LAMMPS Code Clinic event as an online activity between August 8th, 2022 and September 2nd, 2022

This is an opportunity to learn about programming with LAMMPS. improve LAMMPS itself or your own code for LAMMPS, and collaborate with other LAMMPS developers.

As always, the price is free!

For more details, please go to the [Code Clinic event homepage](https://www.lammps.org/workshops/Aug22/)

PR #3411, opened 2022-08-23
Documentation edits (round 2)

hammondkd:
**Summary**

I have made many changes to clean up the documentation. Primarily these include mathematical expressions, but I also replaced parsed-literal with code-block LAMMPS in the Syntax sections. Any missing style descriptions for computes and fixes were added. I also included some changes that make the documentation more consistent and/or easier to parse for a non-human reader (e.g., replacing "[nocoeff]" with "keyword" and "zero or more keywords may be appended" and so forth).

I do not think I made any substantive changes to the documentation, only cleaning up math and wordsmithing edits plus those described above. However, the number of these edits is getting extreme, so I'm pushing these now so it is easier for the core developers to read them and take them in. I will continue working on this, continuing with fix styles, later.

These changes build cleanly with no warnings or errors using both "make html" and "make fasthtml" in the doc directory.

**Related Issue(s)**

Works toward #3371 (code clinic 2022, Task 4)

**Author(s)**

Karl Hammond, University of Missouri
hammondkd at missouri.edu

**Licensing**

**Backward Compatibility**

No change to the program itself are included, only to the documentation.

**Post Submission Checklist**


akohlmey:
@hammondkd thanks for this massive update. Ran this through the spell checker and fixed a few minor issues.
Found a misspelled Boltzmann (and removed the corresponding incorrect fixup in the false positives list).
With that I also changed some (more) Boltzmann constants from "k" to "k_B". I think with "k" often referring to force constants, using "k_B" is the better choice.

doc/src/compute_modify.rst
```diff
@@ -20,7 +20,7 @@ Syntax
          N = # of extra degrees of freedom to subtract
        *extra* syntax is identical to *extra/dof*, will be disabled at some point
        *dynamic/dof* value = *yes* or *no*
-         yes/no = do or do not re-compute the number of degrees of freedom (DOF) contributing to the temperature
+         yes/no = do or do not re-compute the number of degrees of freedom (dof) contributing to the temperature
```
akohlmey: Since "DOF" is an initialism, I prefer it to be in uppercase or at least "DoF".

hammondkd: The documentation was split on this; some used "dof" while others used "DOF"; I went with the first one I found, which was "dof," but I could change them all to DOF if that is preferable.

akohlmey: It is certainly *my* preference. I cannot speak for the other LAMMPS developers.

sjplimp (approved):

PR #3403, opened 2022-08-16
improve error messages

crackboy88:
**Summary**
improve error messages

**Author(s)**
Jiancheng Chen

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

- [ ] The feature or features in this pull request is complete
- [ ] Licensing information is complete
- [ ] Corresponding author information is complete
- [ ] The source code follows the LAMMPS formatting guidelines
- [ ] The feature has been verified to work with the CMake based build system

**Further Information, Files, and Links**

PR #3394, opened 2022-08-12
Improve errors

crackboy88:
**Summary**
I have improved some error messages in minimize.cpp and neighbor.cpp.

**Related Issue(s)**

**Author(s)**
Jiancheng Chen

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

- [ ] The feature or features in this pull request is complete
- [ ] Licensing information is complete
- [ ] Corresponding author information is complete
- [ ] The source code follows the LAMMPS formatting guidelines
- [ ] Suitable new documentation files and/or updates to the existing docs are included
- [ ] The added/updated documentation is integrated and tested with the documentation build system
- [ ] The feature has been verified to work with the conventional build system
- [ ] The feature has been verified to work with the CMake based build system
- [ ] Suitable tests have been added to the unittest tree.
- [ ] A package specific README file has been included or updated
- [ ] One or more example input decks are included

**Further Information, Files, and Links**

PR #3421, opened 2022-08-29
Major update to mesocnt styles

phankl:
**Summary**

This is a major update to the MESONT package, mainly regarding pair_style mesocnt and the new styles pair_style mesocnt/viscous, bond_style mesocnt and angle_style mesocnt.

This update introduces the following new features:
- segment-segment interaction mode in addition to segment-chain mode in pair_style mesocnt
- new bond_style mesocnt to automatically parameterize harmonic stretching of nanotube segments with presets for carbon nanotubes
- new angle_style mesocnt to automatically parameterize harmonic and anharmonic bending + buckling of nanotube segments with presets for carbon nanotubes
- new pair_style mesocnt/viscous that implements viscous friction between nanotube segments
- pair_style mesocnt now supports neighbor list construction based on bond topology which no longer requires bonded atoms to have consecutive atom IDs
- Gauss-Legendre quadrature is now used rather than the trapezoidal rule, giving much better convergence of the numerical integration used by the potential
- some performance improvements due to better logic 

In addition to the new features, a bug in identifying nanotube ends has been fixed that could lead to instabilities.

Functionalities of the mesocnt styles now match (and in some cases exceed) the mesont/tpm pair_style. Benchmarks show that the mesocnt styles are typically at least 2x faster than mesont/tpm using identical simulation parameters and also show better energy conservation in NVE simulations (see attached PDFs for plots). The mesocnt styles further don't need a separate library to be built and linked to LAMMPS, and do not rely on special computes to extract the individual contributions to the potential. They also work with any molecular atom style, and don't need their own atom style to function.

**Author(s)**

Philipp Kloza (U Cambridge)

**Licensing**

**Backward Compatibility**

Command structure of pair_style mesocnt has changed, so old input files are no longer compatible with the new version of the mesocnt styles.

**Implementation Notes**

When neighbor lists are rebuilt, atom->nspecial and atom->special are communicated for ghost atoms to ensure that all relevant atoms are available for the many-body potential in pair_style mesocnt. This was discussed in the following forum post: https://matsci.org/t/molecule-topology-for-ghost-atoms/42051

The styles bond_style mesocnt and angle_style mesocnt are essentially directly copied over from bond_style harmonic and angle_style harmonic respectively, and just added wrappers for the energy constants. Angle_style mesocnt also slightly changes the potential itself to account for anharmonic bending buckling.

Buckling information set by angle_style mesocnt is used by the mesocnt pair_styles to automatically switch from segment-chain to segment-segment mode. This is done by adding an atom property 'buckled' via a fix called 'angle_mesocnt_buckled' when angle_style mesocnt is initialized. 

**Post Submission Checklist**

[benchmark_nvt.pdf](https://github.com/lammps/lammps/files/9445041/benchmark_nvt.pdf)
[benchmark_nve.pdf](https://github.com/lammps/lammps/files/9445044/benchmark_nve.pdf)


doc/src/Packages_details.rst
```diff
@@ -1563,11 +1563,18 @@ segments. Internal interactions within a NT and the van der Waals
...
+
+This package contains two independent implementations of this model:
+:doc:`pair_style mesont/tpm <pair_mesont_tpm>` is the original implemention
+of the model based on a Fortran library in the ``lib/mesont`` folder. The
+second implementation is given by the mesocnt styles (:doc:`bond_style mesocnt <bond_mesocnt>`, 
+:doc:`angle_style mesocnt <angle_mesocnt>` and :doc:`pair_style mesocnt <pair_mesocnt>`).
+The mesocnt implemenation has the same features as the original implementation with the
+addition of friction,
+but is directly implemented in C++, interfaces more cleanly with general LAMMPS
+functionality and is typically faster. It also doesn't require its own atom type
```
akohlmey:
```suggestion
functionality and is typically faster. It also doesn't require its own atom style
```

src/MESONT/angle_mesocnt.cpp
```diff
@@ -0,0 +1,392 @@
+/* ----------------------------------------------------------------------
```
akohlmey, line 1: Since you are adding new styles in a package, those files must be added to `src/.gitignore` so that git will ignore the copy when using the traditional build system.

src/MESONT/angle_mesocnt.cpp
```diff
@@ -0,0 +1,392 @@
...
+#include "math_const.h"
+#include "memory.h"
+#include "modify.h"
+#include "neighbor.h"
+#include "update.h"
+
+#include <cmath>
+
+using namespace LAMMPS_NS;
+using namespace MathConst;
```
akohlmey:
```suggestion
using MathConst::MY_PI;
using MathConst::MY_2PI;
```

src/MESONT/angle_mesocnt.cpp
```diff
@@ -0,0 +1,392 @@
...
+  for (int i = 1; i < np1; i++) setflag[i] = 0;
+}
+
+/* ----------------------------------------------------------------------
+   set coeffs for one or more types
+------------------------------------------------------------------------- */
+
+void AngleMesoCNT::coeff(int narg, char **arg)
+{
+  if (narg < 1) error->all(FLERR, "Incorrect args for angle coefficients");
```
akohlmey: We have recently changed how we prefer to handle errors. Rather than having terse and non-specific error messages in the source code, we want to include relevant details. However, they should be just enough to figure out the rest from the documentation rather than suggesting what to change already in the error message.

Some more detailed explanations and examples are here: https://lammps.org/workshops/Aug22/project1/ and recent pull requests: PR #3394, PR #3403, PR #3408

src/MESONT/angle_mesocnt.cpp
```diff
@@ -0,0 +1,392 @@
...
+
+  int buckling_one;
+  if (strcmp(arg[1], "buckling") == 0)
+    buckling_one = 1;
+  else if (strcmp(arg[1], "harmonic") == 0)
+    buckling_one = 0;
+  else
+    error->all(FLERR,
```
akohlmey, line 215: This line, for example, could become:
```
error->all(FLERR, "Unknown first argument {} for angle coefficients", arg[1]);
```
Please keep in mind that LAMMPS will usually print the last line of input where the error happened to provide context.

src/MESONT/angle_mesocnt.cpp
```diff
@@ -0,0 +1,392 @@
...
+  else if (strcmp(update->unit_style, "cgs") == 0)
+    eunit = 1.6021765e-12;
+  else if (strcmp(update->unit_style, "electron") == 0)
+    eunit = 3.674932248e-2;
+  else if (strcmp(update->unit_style, "micro") == 0)
+    eunit = 1.6021765e-4;
+  else if (strcmp(update->unit_style, "nano") == 0)
+    eunit = 1.6021765e2;
+  else
+    error->all(FLERR, "Angle style mesocnt does not recognize this units style");
```
akohlmey:
```suggestion
    error->all(FLERR, "Angle style mesocnt does not support {} units", update->unit_style);
```

With this the first if statement could be removed, too, since this line will provide the same error message.

src/MESONT/bond_mesocnt.cpp
```diff
@@ -0,0 +1,119 @@
+/* ----------------------------------------------------------------------
```
akohlmey, line 1: Please see my suggestions for the angle style.

akohlmey (changes requested):
Thanks for your submission. This overall looks good and follows the published style suggestions rather closely. Thanks for that.
Please go to the src folder and type `make fix-whitespace` and commit those changes.
Similarly, you probably want to run `clang-format -i` on all your files after implementing the suggested cosmetic changes. I've added a few quick comments and suggestions. I will run some more checks and compile and run the code later on.

phankl:
Thanks for the feedback! I made all the requested changes to the error messages, the whitespaces and the formatting.
Is there a way to auto-format the doc .rst files?
I also wanted to update the potential file included with LAMMPS (new version: [C_10_10.mesocnt.zip](https://github.com/lammps/lammps/files/9453607/C_10_10.mesocnt.zip)) which is too large to be included in the direct git repository.

sjplimp (approved):

sjplimp:
@akohlmey Remind how we now handle very large potential files w/out including them in the Git repo or the tarball?

akohlmey:
> Is there a way to auto-format the doc .rst files?

what do you mean by "auto-format"?

akohlmey:
> @akohlmey Remind how we now handle very large potential files w/out including them in the Git repo or the tarball?

We put them on https://downloads.hpc.temple.edu and they are downloaded transparently when you do either `cmake -DPKG_MESONT=on` or `make yes-mesont`.

phankl:
> > Is there a way to auto-format the doc .rst files?
> 
> what do you mean by "auto-format"?

Similar to clang-format, just for .rst files to automatically format line length, indentation, etc. This probably isn't as critical as for actual code though.

akohlmey:
> Similar to clang-format, just for .rst files to automatically format line length, indentation, etc. This probably isn't as critical as for actual code though.

I am not aware of anything fully automatic. I am using `rst-mode` in emacs and reformatting text blocks works with `Alt-q`.

phankl:
> > Similar to clang-format, just for .rst files to automatically format line length, indentation, etc. This probably isn't as critical as for actual code though.
> 
> I am not aware of anything fully automatic. I am using `rst-mode` in emacs and reformatting text blocks works with `Alt-q`.

Thanks, this is really useful. Just fixed the formatting of all relevant doc files. When I build the documentation however, I get the following warnings:
```
Pair style entry mesocnt / viscous is missing or incomplete in Commands_pair.rst
Pair style entry mesocnt / viscous is missing or incomplete in pair_style.rst
pair_style index entry mesocnt / viscous is missing
Found 1 issue(s) with style index
```
I did add entries for mesocnt/viscous in both Commands_pair.rst and pair_style.rst, so what's missing here?

akohlmey:
> I did add entries for mesocnt/viscous in both Commands_pair.rst and pair_style.rst, so what's missing here?

No worries. I'll take care of this. I just checked out the PR branch and am making some tests now.

One thing that worries me is that my output does not match yours and also 1 MPI does not match 4 MPI.
The differences are small but still significant and appear immediately. This suggests some small differences in the forces.
Any idea where this could be coming from?

phankl:
This is indeed worrying, but only seems to occur with the 'chain' keyword. When 'segment' is used instead, the output is identical for 1 MPI and 4 MPI. I'll look into this issue, there probably is some error with the way neighbor chains are handled.

akohlmey:
> We put them on https://downloads.hpc.temple.edu and they are downloaded transparently when you do either `cmake -DPKG_MESONT=on` or `make yes-mesont`.

@phankl @sjplimp FYI the new potential file is now deposited on the download server and the md5sum hash in `src/MESONT/potentials.txt` updated.

akohlmey:
@phankl one final suggestion for the documentation. We have started to follow the example of the Python documentation and put `.. versionadded::` and `.. versionchanged::` directives into the documentation to indicate when some command/style was added and when (breaking) syntax changes happened. This is to address the behavior that people using older versions of LAMMPS quite often consult the online version of the documentation (which represents the latest patch) and then get confused if their executable does not accept the syntax or know the command. If you want, you can insert those as well into your pull request. Just put as date TBD, this will be replaced by the actual release date when that date is known. You can find some examples of those already in the documentation.

phankl:
> One thing that worries me is that my output does not match yours and also 1 MPI does not match 4 MPI.
> The differences are small but still significant and appear immediately. This suggests some small differences in the forces.
> Any idea where this could be coming from?

I believe to have identified the issue, or rather a combination of two issues.
1. Changing the `SWITCH` variable in pair_mesocnt.cpp to 1.0e-4 rather than 1.0e-6 reduces the sensitivity of the potential to small perturbations. This parameter influences how the potential behaves when two nanotube segments are nearly parallel and division by a small number may lead to errors. High values of `SWITCH` are good for numerical stability, but possibly bad for energy conservation.
2. If one atom in a segment is a ghost atom and the cutoff is slightly too small, then interacting neighbors of the ghost atom are missing in the neighbor list for the full segment. Increasing the cutoff by one segment length (10 Angstrom) remedies that.

With both changes in place, 1 MPI should be identical to 4 MPI to 4 decimal places. I'll take a look tomorrow if there is a clever way to reduce the error of the division.

phankl:
I added versions to the documentation as suggested. Also, I checked and setting `SWITCH` to 1.0e-4 actually improves energy conservation, so I'll keep that value. Another change is to the potential file (new new version: [C_10_10.mesocnt.zip](https://github.com/lammps/lammps/files/9461189/C_10_10.mesocnt.zip)), tweaking one of the cutoffs to improve numerical precision. Sorry for the extra effort having to replace the potential file again, but now the example results should be the same with 1/4 MPI and hopefully reproducible on different machines.

akohlmey (approved):

akohlmey:
@phankl thank you so much for making all those changes. I can confirm that with the latest updates and including the updated potential file I can match the reference results with two different compilers and with 1, 4, and 8 MPI tasks. This is looking very good now. I also make a few cosmetic changes. This is now good to go int. Thanks again for your support with this. You have saved me quite some time and effort.

Issue #2759, opened 2021-05-10
[BUG] in rare cases roundoff makes neighbor list fail to find pairs


bernstei:
**Summary**

If atom initial positions are outside the [0..1] scaled coordinate range, LAMMPS's wrapping can break the symmetry between local and ghost atom pairs, and that makes the neighbor lists's spatial ordering reject both times it checks the pair of atoms that should be neighbors.  See also more detailed discussion at [this matsci.org thread](https://matsci.org/t/apparently-inconsistent-forces-before-after-dump/36305/28).

**LAMMPS Version and Platform**

`LAMMPS (8 Apr 2021)`

CentOS 7 on an Intel CPU, gcc 10.3.0, compiled with -O0 or -O1.  It also appears to be the same for gcc 7.3.

**Expected Behavior**

When running attached input script, a neighbor list containing one pair of atoms (i.e. output contains `Total # of neighbors = 1`)

**Actual Behavior**

output contains `Total # of neighbors = 0`

**Steps to Reproduce**

`lmp_serial -in two_atoms.in | grep 'Total # of neigh'`

**Further Information, Files, and Links**

[two_atoms.in.gz](https://github.com/lammps/lammps/files/6444471/two_atoms.in.gz)

sjplimp:
@bernstei Can you explain the geometry for where these 2 atoms are wrt the prism box?  I'm not clear on why there are 14 ghost atoms when I run on one proc.  Is each of the in-the-box atoms a neighbor of the other in a ghost sense?  I think the problem is that the neigh alg for triclinic relies on the (inbox 1, ghost 2) pair having consistent coords with the (inbox 2, ghost 1) pair.  In the sense of one atom being above/below the other in some dim.  And the box coords --> lamda coords transformation can mess that up by epsilon due the complexity of the math and the fact that all 3 coord dims are coupled together.  Unlike for orthogonal boxes.

bernstei:
I think you're right about the problem.  As for where they are, I'd have to pop them in in some visualizer.  They are interacting across some periodic boundary, yes.  Do you really have 14 ghost atoms?  I recall one ending up with 8 images, and the other with 4, so a total of 12 local+ghost.  But I'm not sure of this - I'll have to check.

sjplimp:
~7 images/atom is fine.  I was just surprised b/c the cutoff was so short and the tilt does not seem extreme.  Mabye this is b/c each atoms are in one of the corners of the prism?

bernstei:
[...]

PR #3430, opened 2022-09-04
Reducing the number of npair and nstencil classes

jtclemm:
**Summary**

There are roughly 100 npair and 20 nstencil classes which are mostly just riffs on the same class, duplicating the majority of their code. A primary cause is the growing combinatorics of options (2D vs 3D, half vs full, newton vs newtoff, etc).   Taking a page out of the Kokkos book, I took a stab at creating a  `NStencilBin`  template which which combines 6 stencils (2D vs 3D, full vs half, and tri vs ortho) and a `NPairBin` template which (initially) combines 8 classes (full vs half, newton vs newtoff, ortho vs tri, and size vs no size) but could include atom_only as well.

At the moment, this pull request is unfinished and I’m hoping to get some feedback on whether this kind of an approach, or another approach, might be an appropriate way to consolidate code.

Initial LJ melt benchmarks found that on average building a neighbor list using the refactored code might be anywhere from 2% faster or slower depending on the specific neighbor list options. This is comparable to the typical variation between runs. More rigorous benchmarking will, of course, be necessary.

**Author(s)**

Joel Clemmer (SNL)

**Licensing**

**Implementation Notes**

Only one conceptual change was made to the construction of neighbor lists. For half+newton+ortho settings, the stencil will now include an atom’s own bin. Previously, the npair class would separately search an atom's own bin before looping through bins in the stencil. This change requires an extra logic check for `stencil[k] == 0` which leads to extra tests to see if the neighboring atom is on the ‘top right’. From initial tests, this change didn’t have a noticeable impact on performance.

Per Steve’s suggestion, in these circumstances an atom's own bin was kept at the front of the stencil such that it is processed first and the order of neighbors remains unchanged.

**Post Submission Checklist**

NA


akohlmey:
@jtclemm if you want feedback on the viability of your refactoring, you should explicitly notify @sjplimp and discuss with him. The neighborlist code and specifically the many classes are his design and work. I doubt that anybody else feels ready for an in-depth discussion on details.

jtclemm:
> @jtclemm if you want feedback on the viability of your refactoring, you should explicitly notify @sjplimp and discuss with him. The neighborlist code and specifically the many classes are his design and work. I doubt that anybody else feels ready for an in-depth discussion on details.

I spoke with Steve at the beginning of this PR but I will definitely need to follow up with him after I finish the last few details of this PR. I mostly need to run more rigorous performance benchmarks although I see there are still some errors in the unit/compilation test builds that I'll also need to fix. I should have more time to wrap this up in February.

jtclemm:
I just updated this branch then ran a collection of more systematic benchmarks. I selected a few examples with various neighbor styles (including a 2d version of in.melt), increased the runtime to at least 10k timesteps, and ran them on serial and parallel (4 processors) using the old and new code. The relative performance of the new code vs the old code is plotted below:
image

For each example, I ran 20 copies of the exact same job on different HPC nodes to estimate variations between jobs. The standard deviation is indicated by the error bars. For most examples, the new code does take about ~5% longer to create a neighbor list than the old code (left panel). A smaller distinction is then seen in the total run time (right panel). No differences were identified in any trajectories.

So bottom line, there does appear to be a measurable cost, albeit slight, to the newly templated neighbor classes that will affect users. Are there any opinions on whether or not the consolidation of neighbor classes is worth it?

Unless the unit tests turn up any new errors, this PR is complete and could be merged if its changes are considered to be worth the cost. If anyone has suggestions, I can also easily expand this analysis to more examples.

Since reviewing this PR might be a herculean task, I also plan on doing another detailed read through to double check all of the changes.

stanmoore1:
A few comments: I like this approach. I'm really confused why this would have up to 10-15% slowdown in the neigh list build though, since using templated if tests should have no runtime penalty because the compiler makes separate code paths. The only thing I can think of is the compiler is having a harder time optimizing or vectorizing the code. Seems like we should look into this a bit more.

jtclemm:
> A few comments: I like this approach. I'm really confused why this would have up to 10-15% slowdown in the neigh list build though, since using templated if tests should have no runtime penalty because the compiler makes separate code paths. The only thing I can think of is the compiler is having a harder time optimizing or vectorizing the code. Seems like we should look into this a bit more.

I cannot say for sure, but beyond the templates there is one other significant change that affects half + ortho +newton neighbor lists (the majority of examples). For example, seen here on line 132: https://github.com/lammps/lammps/blob/c3192313954c166d24f530d236a27e740f7e9283/src/OPENMP/npair_bin_atomonly_omp.cpp#L132
In the old files, the atom's own bin (stencil[k] = 0) is not included in the stencil list. Instead, the npair class would separately iterate over all atoms in that stencil (duplicating a lot of code), as seen starting here:
https://github.com/lammps/lammps/blob/b8ff478d7874a2ae0ef9402fc4728a77e87214a4/src/npair_half_bin_newton.cpp#L81
To me, a single if statement doesn't seem like it would be that expensive, but maybe there are other consequences that emerge.

Interestingly, you can see the one example with a full neighbor list (water_collapse from the SPH package) is one of the exceptions where performance is on par (if not slightly better). Maybe I should dig up some other examples with full lists or triclinic boundaries or newtoff to test this possibility

akohlmey:
> Unless the unit tests turn up any new errors, this PR is complete and could be merged if its changes are considered to be worth the cost. If anyone has suggestions, I can also easily expand this analysis to more examples.

I see unit tests failing to run pair styles in the INTEL package: https://ci.lammps.org/job/dev/job/pull_requests/job/ubuntu/job/unit_tests/job/cmake_mpi_openmp_smallbig_shared/6662/#showFailuresLink e.g. this one:
https://ci.lammps.org/job/dev/job/pull_requests/job/ubuntu/job/unit_tests/job/cmake_mpi_openmp_smallbig_shared/6662/testReport/junit/projectroot.unittest/force-styles/MolPairStyle_buck/
And on Windows there is an unexpected NaN popping up where a 0.0 is expected: https://github.com/lammps/lammps/actions/runs/4536535566/jobs/7993345512?pr=3430#step:6:4477

jtclemm:
> I see unit tests failing to run pair styles in the INTEL package: https://ci.lammps.org/job/dev/job/pull_requests/job/ubuntu/job/unit_tests/job/cmake_mpi_openmp_smallbig_shared/6662/#showFailuresLink e.g. this one: https://ci.lammps.org/job/dev/job/pull_requests/job/ubuntu/job/unit_tests/job/cmake_mpi_openmp_smallbig_shared/6662/testReport/junit/projectroot.unittest/force-styles/MolPairStyle_buck/ And on Windows there is an unexpected NaN popping up where a 0.0 is expected: https://github.com/lammps/lammps/actions/runs/4536535566/jobs/7993345512?pr=3430#step:6:4477

Thank you for the info. For some of the checks, I have a hard time finding the log files on GitHub to figure out which unittest failed. The INTEL issue definitely looks like it is a serious error and I'll work on isolating what went awry. The Python issue looks odd since it doesn't seem to involve neighbor lists (computes a KE), but I'll take a look at it too.

akohlmey:
> The Python issue looks odd since it doesn't seem to involve neighbor lists (computes a KE), but I'll take a look at it too.

It would not be the first time that a refactoring exposes a previously undetected bug.

jtclemm:
@akohlmey A big issue was that cmake could not find most of the `npair_*_intel.h` files without the corresponding non-intel files. E.g. `npair_full_bin.h` no longer existed so cmake didn't look for `npair_full_bin_intel.h`. I crudely fixed this by collapsing `npair_*_intel.h/cpp` files as needed. After the unittest errors are resolved, I will try to clean this up with a few more templates. I also added an NstencilBinIntel class to reproduce the old behavior (excluding the self bin) to avoid mucking around too much in a package I don't understand.

That being said, the same unit tests are still failing. I ran one of them (MolPairStyle_buck) on my local machine using Intel and dumped atom forces. I find identical results between this branch and the develop branch (~16 digits) but find considerable differences with the non-Intel results (~7th digit). The non-Intel results agree with the yaml file. Is the Intel package supposed to have this much deviation? I could also be making a trivial mistake as I have no experience with the Intel package. 

Thanks in advance for any suggestions. Currently, this has me a bit stuck.

akohlmey:
> Is the Intel package supposed to have this much deviation?

That depends on whether you run the INTEL package calculation in single precision, mixed precision or double precision mode. See: https://docs.lammps.org/package.html in double precision mode the difference should be close to the base class results. There is a fudge factor that accounts for floating point non-associativity and big(ger) fudge factors when using single or mixed precision (same for GPU package tests).

I'm currently busy with teaching classes (via Zoom in Europe, so *very* early morning until noon) and completing the next feature release for LAMMPS. Hopefully by next week I am done with lectures and the first round of grading assignments, so that I can have a closer look. There are some other unit tests that are a bit "funny" occasionally. So this would be an opportunity to have a good look at all of them.

jtclemm:
> See: https://docs.lammps.org/package.html in double precision

Ah thanks! That made a huge improvement. Makes sense in retrospect given the ~7 vs ~16 digits of precision. The intel results are now much closer to the yaml file and there are now noticeable differences between forces computed in this branch and the develop branch. Hopefully I can now isolate the bug.

jtclemm:
I ported the changes from #3846 so this PR should be up-to-date. I tested the example in #2759 w/ and w/o openmp and confirmed Steve's patch still works. I also added atomonly variants to the multi npair class.

However, there still seem to be issues with the Intel package that I haven't pinned down.

sjplimp:
@jtclemm Thanks for working more on this.  Are the Intel issues with accuracy, i.e. the neighbor lists are different (wrong) ?
I assume the CPU (non-Intel) results are essentially identical before and after this PR ?  And that the previous Intel results
matched the CPU (non-Intel)?

jtclemm:
> @jtclemm Thanks for working more on this. Are the Intel issues with accuracy, i.e. the neighbor lists are different (wrong) ?

Yes, the neighbor lists were definitely wrong, but I had to figure out which unit tests were failing. Should work now.

The INTEL package required distinct nstencil classes that reproduced the old behavior: not appending the self-bin to the stencil. Alternatively, I could have edited the intel npair classes, but I didn't want to poke around too much in the intel package since I'm unfamiliar with it.

akohlmey:
@jtclemm I've been looking through the static code analysis output for the develop branch and found this (i.e. in the code *before* this pull request):
coverity-neighlist-issue

Difficult for me to say if this is intentional and whether it is still present after your changes, but since you have already your hands dirty from digging through this code, it may be simpler to check this out as well. The same report exists for the 3d version of this file as well.

jtclemm:
> Difficult for me to say if this is intentional and whether it is still present after your changes, but since you have already your hands dirty from digging through this code, it may be simpler to check this out as well. The same report exists for the 3d version of this file as well.

Yes that was intentional. It was a kludge introduced in #3846. I mentioned it here: https://github.com/lammps/lammps/pull/3846#discussion_r1261430888. The plan was to fix it in this PR.

stanmoore1:
@jtclemm did you ever figure out the root cause of the 10-15% slowdown from this PR you were seeing previously?

jtclemm:
> @jtclemm did you ever figure out the root cause of the 10-15% slowdown from this PR you were seeing previously?

No, unfortunately I did not. I speculated it could result from an extra if statement, but I can try and investigate more when I get the chance.

sjplimp:
Re: Stan's comments on the slowdown, I agree it would be good to understand better. I looked at the earlier discussion of it above.
(1) The code example pointed to was for an OpenMP variant.  I assume in the plot that all the examples were for vanilla CPU runs (not OpenMP) ?  

(2) The code construct with the extra if tests in the main loop over bins is the same for CPU and OpenMP?

(3) Are all the slowdowns (e.g. in the plot) for the common case of HALF, NEWTON, (not) TRI ?
   Do tests with any of the opposite settings show no slowdown - FULL, NO-NEWTON, TRI ?

(4) There are actually 2 extra if tests in the bin loop for the common case, one to set bin_start, one when the atom is a ghost.

One Q: why isn't the if test in both cases for where (k == 0) and not (stencil[k] == 0) ??
I.e. the first bin in the stencil

Also, it seems like the "extra work" in the templating context might be only for the setting of bin_start.
I wonder if this altered code construct might help:
      //  bin_start = binhead[ibin + stencil[k]];   // MOVE this line inside the template test so bin_start is not set twice
      if (HALF && NEWTON && (!TRI)) {
        if (stencil[k] == 0) {  // use k == 0 instead?
          // Half neighbor list, newton on, orthonormal
          // loop over rest of atoms in i's bin, ghosts are at end of linked list
          bin_start = bins[i];
        } else {
          bin_start = binhead[ibin + stencil[k]];
        }
        // OR replace this inner if test with ternary operator, possibly the compiler optimzes this better
        // bin_start =  stencil[k] ? binhead[ibin + stencil[k]] : bins[i];   // again, can stencil[k] before the ? be just k
      } else {
        bin_start = binhead[ibin + stencil[k]];
      }

sjplimp:
I just realized the code construct I suggested is not in the inner loop over atoms of each bin, so less important.
Maybe just try k == 0 instead of stencil[k] == 0 in the inner loop (ghost atom test)

jtclemm:
> I just realized the code construct I suggested is not in the inner loop over atoms of each bin, so less important. Maybe just try k == 0 instead of stencil[k] == 0 in the inner loop (ghost atom test)

Focusing on the melt example, I tried running newton+ortho, newtoff, and newton+tri for a variety of scenarios including:
(a) the current develop branch
(b) this branch
(c) this branch with your proposed change (replacing stencil[k] == 0 with k == 0)
(d) this branch with the self-bin check moved outside of the k-loop, removing the if statement and duplicating a big chunk of code (i.e. like the current branch)

Running a handful of iterations, the average nlist build time is pretty much the same across all 4 builds with newtoff and newton+tri but I did see differences for Newton+ortho.  I didn't test a full neighbor list, but previously I saw no performance drop for the water collapse SPH demo. So yes, this seems to suggest that the only slow down is in newton+ortho which has that extra if statement. If helpful, I can run a more tests and give more rigorous timings.

Running Newton+ortho across the four builds, (a) and (d) are the same, (b) is ~10% slower, and (c) is ~5% slower. Of course, differences in the total simulation run time were much smaller since building nlists was only ~12% of the total time.

stanmoore1:
@jtclemm I can reproduce the slowdown on my Linux box. You could write another templated function for the inner j-loop, but that seems like it would be pretty ugly/hard to read compared to the current code.

jtclemm:
> @jtclemm I can reproduce the slowdown on my Linux box. You could write another templated function for the inner j-loop, but that seems like it would be pretty ugly/hard to read compared to the current code.

Thanks for verifying the slow down. I agree that that solution could work but might get messy. 

There also appear to be a few other ways to template out other if statements and improve performance beyond the original. For instance, removing "if (exclude && ...)" statement by introducing an NP_EXCLUDE variety gives a 5-10% speed up over the original. Probably could accomplish something similar by separating NP_MOLONLY into a templated and non-templated variety. But making thse changes would triple the number of npair classes - which I presume would increase compile time and binary size.

It isn't obvious to me how to balance performance vs. code complexity vs. proliferation of npair classes.

sjplimp:
@jtclemm Thanks for the additional testing.  I think including only optimations to the current templating strategy is fine.
I.e. don't worry about 5% slowdown in the neighbor building.  Since it is a small fraction of total run time.  And the
large reduction in NPair, NStencil classes due to your consolidation is very nice.
 
So I think that means change stencil[k] == 0 to k == 0 everywhere if the self-bin is always first in the list.
Possibly the other outer-loop change I suggested (don't double-calclulate bin_start) will speed-up problems like granular where there are few atoms/bin and many bins.

If other developers disagree, we can discuss at the next developer's conf call ?

jtclemm:
> @jtclemm Thanks for the additional testing. I think including only optimations to the current templating strategy is fine. I.e. don't worry about 5% slowdown in the neighbor building. Since it is a small fraction of total run time. And the large reduction in NPair, NStencil classes due to your consolidation is very nice.
> 
> So I think that means change stencil[k] == 0 to k == 0 everywhere if the self-bin is always first in the list. Possibly the other outer-loop change I suggested (don't double-calclulate bin_start) will speed-up problems like granular where there are few atoms/bin and many bins.

That makes sense. As it's a clear improvement with no downside, I made the 1st change. I'll probably hold off on benchmarking/implementing the 2nd for a future PR since this PR is already quite complicated. I also had a few less impactful changes (e.g. the NPairSkipX classes could be templated) which could be returned to in a future PR.

jtclemm:
Hey @stanmoore1, I incorporated the changes from #3974. Outside of INTEL, all of the skip classes now have a template variable TRIM. Do you have extra tests to check the changes?

Afterwards, this PR should be done.

stanmoore1:
@jtclemm I have been using the test case here: https://github.com/lammps/lammps/issues/3942#issuecomment-1773460432. Some of the regression tests should also have hybrid, @akohlmey can you trigger those?

stanmoore1:
@jtclemm actually the OxDNA test case still has a few outstanding issues we are working on. Any hybrid case with multiple cutoffs should work to test the trim.

jtclemm:
> @jtclemm I have been using the test case here: [#3942 (comment)](https://github.com/lammps/lammps/issues/3942#issuecomment-1773460432). Some of the regression tests should also have hybrid, @akohlmey can you trigger those?

Thanks. As an aside, do you know why there are NP_SIZE variants of NPairSkip and NPairSkipTrim? It looked unnecessary.

akohlmey:
> Some of the regression tests should also have hybrid, @akohlmey can you trigger those?

@stanmoore1 those extra tests can be added by adding the labels: `test_runs`, `test_for_regression`, and `gpu_unit_tests` to the PR. So anybody with "Triage" permission or higher can trigger those. Beware, for technical reasons those tests are retriggered with *every* activity on the pull request (like comments) once the label is added, so it is usually best to remove the labels as soon as the corresponding tests are started. Since those tests are time consuming and block resources for other tests.

akohlmey:
Failing regression tests are for the following examples:
- `examples/dreiding/in.dreiding`
- `examples/PACKAGES/gauss_diel/in.gauss-diel`
- `examples/PACKAGES/gauss_diel/in.gauss-diel-cg`
- `examples/PACKAGES/gauss_diel/in.gauss-diel-split`

See: https://ci.lammps.org/view/LAMMPS%20develop/job/dev/job/pull_requests/job/ubuntu/job/regression_tests/job/cmake_regression_mpi/683/

stanmoore1:
> Thanks. As an aside, do you know why there are NP_SIZE variants of NPairSkip and NPairSkipTrim? It looked unnecessary.

@jtclemm I agree, it looks redundant.

jtclemm:
I don't see any other failed regression tests, so I believe this PR is ready for review.

There are still some pending tasks:
1) possible optimization
2) adding templates for the Intel classes
3) consolidating some of the more niche npair classes (e.g. off2on)

But I think those are best left for a separate PR.

stanmoore1:
I'm running this through my internal regression tests just to double check, should be done soon.

stanmoore1:
Internal regression tests passed too.

stanmoore1 (approved):

PR #3432, opened 2022-09-05
Use primary context in CUDA GPU code.

### LINKED ISSUE #3431: [BUG] CUDA kernel launch failure with OpenMPI 4.1.4

```
LAMMPS (3 Aug 2022)
```

**Summary**

With OpenMPI 4.1.4, LAMMPS crashes due to `cuLaunchKernel` returning an error (400 = Invalid resource handle). With earlier versions, or with the UCC library disabled in OpenMPI, LAMMPS functions correctly.

This is a result of the UCC library making calls to the CUDA runtime library during some MPI calls, which results in the CUDA context being switched to the "primary" context (i.e. the context that the runtime library uses). Since LAMMPS is using the low-lever driver API, it needs to ensure that it is in the correct context being making such calls, however the current code fails to do this: it switches to its own context once during `UCL_Device::set` but then never checks that it's still in that context before e.g. allocating memory or launching kernels.

A simple workaround that allows LAMMPS to work with OpenMPI 4.1.4 + UCC is in benmenadue/lammps@294a1c216893cd6439ad6236be4822d95c101dea; I will PR it across shortly.

Error message:
```
Device 0: NVIDIA A100-SXM4-80GB, 108 CUs, 74/79 GB, 1.4 GHZ (Mixed Precision)
Device 1: NVIDIA A100-SXM4-80GB, 108 CUs, 1.4 GHZ (Mixed Precision)

Initializing Device and compiling on process 0...Done.
Initializing Devices 0-1 on core 0...Done.
Initializing Devices 0-1 on core 1...Done.
Initializing Devices 0-1 on core 2...Done.
Initializing Devices 0-1 on core 3...Done.

Cuda driver error 400 in call at file 'geryon/nvd_kernel.h' in line 340.
Cuda driver error 400 in call at file 'geryon/nvd_kernel.h' in line 340.
Cuda driver error 400 in call at file 'geryon/nvd_kernel.h' in line 340.
Cuda driver error 400 in call at file 'geryon/nvd_kernel.h' in line 340.
MPI_ABORT was invoked on rank 3 in communicator MPI_COMM_WORLD
with errorcode -1.

NOTE: invoking MPI_ABORT causes Open MPI to kill all MPI processes.
You may or may not see output from other processes, depending on
exactly when Open MPI kills them.
Cuda driver error 400 in call at file 'geryon/nvd_kernel.h' in line 340.
Cuda driver error 400 in call at file 'geryon/nvd_kernel.h' in line 340.
Cuda driver error 400 in call at file 'geryon/nvd_kernel.h' in line 340.
Cuda driver error 400 in call at file 'geryon/nvd_kernel.h' in line 340.
[gadi-dgx-a100-0001.gadi.nci.org.au:00928] 7 more processes have sent help message help-mpi-api.txt / mpi-abort
[gadi-dgx-a100-0001.gadi.nci.org.au:00928] Set MCA parameter "orte_base_help_aggregate" to 0 to see all help / error messages```

benmenadue:
**Summary**

This patch changes the CUDA context creation inside LAMMPS to instead use
the "primary" context for the requested device; this then makes LAMMPS
compatible with other libraries that also use CUDA via its runtime API.

It's not a complete fix, in that other libraries that use their own context
via the driver API could still break LAMMPS (e.g. if they switch to their
own context after LAMMPS has initialised its and don't switch it back before
returning), but it at least covers packages that use the CUDA runtime
library.

Unfortunately, to make all driver calls in LAMMPS "context aware" might
require more significant changes to the interfaces as the current context
may not be available at the calls (for example, the `_device_alloc` functions
in `nvd_memory.h` that don't take a `UCL_Device` argument).

**Related Issue(s)**

Fixes: #3431

**Author(s)**

Ben Menadue, National Computational Infrastructure, Australian National University

**Licensing**

**Backward Compatibility**

These changes should not have any impact on backwards compatibility.

**Implementation Notes**

As recommended in the CUDA documentation, switches it to use `cuDevicePrimaryCtxRetain` and `cuDevicePrimaryCtxRelease` instead of `cuCtxCreate` and `cuCtxDestroy`.

**Post Submission Checklist**


wmbrownIntel (commented):
@benmenadue  I think it is OK to make this the default behavior, but please add preprocessor option for original behavior (e.g. #ifndef GERYON_UNIQUE_CONTEXT). Thanks!

akohlmey:
> @benmenadue I think it is OK to make this the default behavior, but please add preprocessor option for original behavior (e.g. #ifndef GERYON_UNIQUE_CONTEXT). Thanks!

@wmbrownIntel implemented your request

akohlmey (approved):

wmbrownIntel (approved):

Issue #3282, opened 2022-05-30
[Feature Request] Refactor offline compilation download of files


akohlmey:
**Summary**

The current handling of transparent downloads is in several cases dependent on the compilation host being connected to the internet. There are some ways to provide already downloaded archives, but this has known problems, e.g. updates are not applied due to the variables being cached. Also there should be a test whether downloads are possible and CMake should not attempt a download in that case. If offline compilation is intended, this should be indicated and a folder path with previously downloaded source archives provided. The offline use path can then be inferred from the original download URL.

In summary:
- download URLs and hashes should not be stored in cached variables
- implement a test to detect if downloads are possible or not. the CMake scripting should react accordingly and error out or skip steps
- use a single variable to point to a folder with pre-downloaded sources instead of requiring to override all URLs. URLs will be adapted by CMake in offline mode

ocaisa:
Just to add to this, the current approach for building the docs also requires access to the internet...there is no (offline) way to provide the necessary packages to the `venv` used.

akohlmey:
> Just to add to this, the current approach for building the docs also requires access to the internet...there is no (offline) way to provide the necessary packages to the `venv` used.

This is not likely to change (that part of the build system is already complex as it is to be (more) portable). The documentation can be built anywhere and then only the resulting files (html folder, pdf or epub file) need to be transferred to the target location (which may be without internet access).

This is different from compiling executables for which it may only be possible to compile them on a disconnected machine.

mcepl:
So, when the build for openSUSE in our network isolated environment fails with “Downloading external potential C_10_10.mesocnt from https://download.lammps.org/potentials” is it this bug or not?

akohlmey:
> So, when the build for openSUSE in our network isolated environment fails with “Downloading external potential C_10_10.mesocnt from [https://download.lammps.org/potentials”](https://download.lammps.org/potentials%E2%80%9D) is it this bug or not?

You can prevent the download by using `-D DOWNLOAD_POTENTIALS=off` when configuring LAMMPS with CMake.
It appears that this setting is currently undocumented, so I will have to update the documentation. It currently affects only one potential file for a rarely used package, AFAICR.

PR #3439, opened 2022-09-09
Draft: Make CMake Doc build docenv optional

rbberger:
**Summary**

This PR allows the CMake documentation build to use an external Sphinx installation by disabling the automatic virtual environment generation.

Adds an advanced option `BUILD_DOC_VENV` which defaults to `ON`. If disabled, `find_package(Sphinx)` will try to find `sphinx-build` in your PATH. It also assumes your Python installation has all other requirements in `doc/utils/requirements.txt`.

**Related Issue(s)**

#3282

**Author(s)**

@rbberger

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/cmake/Modules/Documentation.cmake
+++ b/cmake/Modules/Documentation.cmake
@@ -4,6 +4,9 @@
 option(BUILD_DOC "Build LAMMPS HTML documentation" OFF)
 
 if(BUILD_DOC)
+  option(BUILD_DOC_VENV "Build LAMMPS documentation virtual environment" ON)
+  mark_as_advanced(BUILD_DOC_VENV)
+
   # Sphinx 3.x requires at least Python 3.5
   if(CMAKE_VERSION VERSION_LESS 3.12)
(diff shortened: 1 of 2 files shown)


akohlmey:
I've been thinking about this for some time now, and I don't think it is worth it and there are two reasons for that
1. Unlike the LAMMPS executables and libraries, which may need to be built on a specific machine with specific tools/compilers, the manual is not tied to such an environment and thus can be build on a different machine (*with* live internet access) without a difference and the resulting output can be moved, if needed.
2. With recent changes in Sphinx affecting the output and requiring some adjustments to some of the "hacks" we do to get the outputs we want, it is not likely that translation will work with an arbitrary version of sphinx and a system installed version will not likely be up-to-date, so using a virtual environment and/or installing custom python packages into user site-packages folders will likely be needed anyway. Right now the version "window" is between 5.2 and 5.3.
3. The sphinx-fortran extension would need to be installed manually over internet somehow anyway.

PR #3440, opened 2022-09-09
Enforce dump snapshots on multiples of N steps in a run after minimize

sjplimp:
**Summary**

Fix an issue with the timesteps for dump snapshots in a run following a minimization not syncing to the specified frequency of multiples of the specified N timesteps.  This may have used to work, but code for dump snapshot output (by step or by time)
has been refactored a couple of times recently, and this case no longer works as it should.

**Related Issue(s)**

N/A

**Author(s)**

Steve

**Licensing**

**Backward Compatibility**

This may change when snapshots are written in a run if it follows a minimize which ended on an arbitrary timestep.

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/doc/src/restart.rst
+++ b/doc/src/restart.rst
@@ -12,7 +12,7 @@ Syntax
    restart N root keyword value ...
    restart N file1 file2 keyword value ...
 
-* N = write a restart file every this many timesteps
+* N = write a restart file on timesteps which are multipls of N
 * N can be a variable (see below)
 * root = filename to which timestep # is appended
 * file1,file2 = two full filenames, toggle between them when writing file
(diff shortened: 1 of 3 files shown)


sjplimp:
@akohlmey Portions of output.cpp were not indented correctly, which makes the file diff look as if big changes were made.
Which is not the case.

akohlmey:
> @akohlmey Portions of output.cpp were not indented correctly, which makes the file diff look as if big changes were made. Which is not the case.

Not a problem. One can tell github to suppress whitespace-only changes or I can check out the branch and do `git diff -w develop..HEAD`.

I think I will spend some time this afternoon to add some more tests to the dump style unit tests for the various cases to make sure we retain consistent behavior in the future.  There are already some tests for `dump_modify every` and `dump_modify every/time`, but not yet for the default behavior with run only or minimize plus run or run plus minimize and also with either starting from step 0 or a different timestep.

sjplimp:
>> add some more tests to the dump style unit tests for the various cases
That's a good idea.  I removed some logic that was preventing the run setup() from resetting the next dump timestep correctly (after min).  I think the logic was no longer applicable due to previous refactoring of all the per-step and per-time stuff.  But not 100% sure of that.  It was certainly causing a problem for this new issue.

akohlmey (approved):

PR #3445, opened 2022-09-12
Reorder punctuation and quotation characters in manual for clarity

akohlmey:
**Summary**

There are several cases where using `."` or `,"` can lead to misunderstandings about command names and settings and input file syntax. This PR changes these to `".` and `",` respectively.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Post Submission Checklist**


hammondkd:
Most (if not all) of these undo changes I made during my three "doc fixes" pull requests. I will make note of this style sheet preference for future edits, but I had made those changes because putting periods and commas inside quotes is preferred in every style sheet I've ever seen. (Sorry....)

While we're on the subject of style preferences, would it be preferable to have some of these things that could potentially be confused inside back-quotes so they are clearly "as is" words, such as the `c_` and `f_` entities, rather than inside double quotes?

sjplimp (approved):

akohlmey:
> Most (if not all) of these undo changes I made during my three "doc fixes" pull requests. I will make note of this style sheet preference for future edits, but I had made those changes because putting periods and commas inside quotes is preferred in every style sheet I've ever seen. (Sorry....)

Yes, I am aware, that it is gramatically correct to have this order. But that refers to direct speech while here the quotation marks are used for highlighting. My concern is that many non-native speakers are not aware of the english language punctuation rules and thus make wrong assumptions about what is meant. Of course that raises the question whether there would be a better alternative to highlight such references to keywords and filenames. The challenge here is that most of this is inherited from the original typesetting, which had much fewer options than what is available with sphinx and thus any change would be a major effort unless it can be automated. But then again, automation does not always work since this is such a large amount of text, there are always exceptions and complications. So you are either in danger of making too many and thus some undesired changes and also missing some cases.

> While we're on the subject of style preferences, would it be preferable to have some of these things that could potentially be confused inside back-quotes so they are clearly "as is" words, such as the `c_` and `f_` entities, rather than inside double quotes?

Yes, there would be a benefit to have some clearly define "rules", a document describing them, and a complete review and reassessment of the typesetting of the manual. At the moment, there is so much work still that can be done on the source code and - while not perfect - the documentation is for the most part quite clean and relatively consistent, when having to choose between code and documentation, the preference is - obviously - on updating and modernizing the code. The challenge with changes to the manual typesetting is always: will it work equally well with HTML and LaTeX? and will it be accepted by the other LAMMPS developers. People have a way of settling into the status quo and resist change unless there is a large benefit.

PR #3448, opened 2022-09-15
Updates to BPM package

jtclemm:
**Summary**

This pull request includes an assortment of document updates, bug fixes, and new features aimed at the BPM package including:
- Addition of an ev_tally_xyz() method to the bond class so tangential forces can contribute to the virial calculation
- New methods to create bonds in the FixUpdateSpecialBonds and Neighbor classes (see implementation notes for details)
- Compute bond/local can now output custom per-bond data such as reference lengths for BPM bonds (identical to the use of `pair->svector` to output extra granular pairstyle data)
- Estimates of bond lengths and bond settings are now restarted correctly for BPM bond styles
- Bond coefficients and reference states are now only written to restart files, not data files. This removes complications with initializing a fix and knowing its ID before reading the data file and aligns with the behavior of the GRANULAR package. This required adding an extra optional keyword to fix property/atom to toggle the `wd_section` flag.
- Fixing typos and formatting in documentation

**Author(s)**

Joel Clemmer (SNL)

**Licensing**

**Backward Compatibility**

NA

**Implementation Notes**

Although BPM bonds can be broken in the middle of a timestep, there isn't a method to efficiently create new bonds given certain conditions (e.g. particles in contact for X timesteps or the temperature drops below Y) which is needed for new mesoscale models in development. To create bonds without triggering a neighbor list rebuild, one needs to:
1) update special neighbor lists and add necessary special bond bits to neighbor list (accomplished using the new `FixUpdateSpecialBonds->create_bond()` method)
2) add new data to the `atom->bond` arrays (easy to do)
3) add a new entry to the `neighbor->bondlist` array which persists until the next nlist rebuild (tricky)

To accomplish task 3, I created a method `neighbor->add_temporary_bond(i,j, btype)` which would append a bond between the i-j atoms to `neighbor->bondlist`, growing the array if necessary. This method would let other classes create a temporary bond as a placeholder until the next neighbor list build without editing protected variables of the neighbor/ntopo classes. This works but I'm not convinced it is the cleanest approach.

**Post Submission Checklist**


sjplimp (approved):

src/fix_update_special_bonds.cpp
```diff
@@ -163,7 +197,37 @@ void FixUpdateSpecialBonds::pre_force(int /*vflag*/)
...
+    i1 = atom->map(tag1);
+    i2 = atom->map(tag2);
+
+    // Loop through atoms of owned atoms i j and update SB bits
+    if (i1 < nlocal) {
+      jlist = firstneigh[i1];
+      jnum = numneigh[i1];
+      for (jj = 0; jj < jnum; jj++) {
+        j = jlist[jj];
+        if (j >> SBBITS & 3 != 0) continue;                  // Skip bonded pairs
```
akohlmey: We should always use parentheses to indicate precedence in all but the most trivial if statements. In this case the precedence rules state that the "!=" is evaluated first which is probably not what you want.

```suggestion
        if (((j >> SBBITS) & 3) != 0) continue;                  // Skip bonded pairs
```

src/fix_update_special_bonds.cpp
```diff
@@ -163,7 +197,37 @@ void FixUpdateSpecialBonds::pre_force(int /*vflag*/)
...
+        if (tag[j] == tag2) jlist[jj] = j ^ (1 << SBBITS);   // Add 1-2 special bond bits
+      }
+    }
+
+    if (i2 < nlocal) {
+      jlist = firstneigh[i2];
+      jnum = numneigh[i2];
+      for (jj = 0; jj < jnum; jj++) {
+        j = jlist[jj];
+        if (j >> SBBITS & 3 != 0) continue;                  // Skip bonded pairs
```
akohlmey: ditto
```suggestion
        if (((j >> SBBITS) & 3) != 0) continue;                  // Skip bonded pairs
```

src/fix_update_special_bonds.cpp
jtclemm: Thanks Axel for spotting that, you are right. That was not what I intended.

akohlmey: @jtclemm I didn't spot it, it was flagged by compiling with Clang++ 14 using -Wall -Wextra

Do you prefer to apply the necessary changes yourself, or do you want me to apply them? Or do them differently?

jtclemm: I just added the change.

PR #3449, opened 2022-09-17
ML potentials with proper orthogonal descriptors

exapde:
**Summary**

This pull request implements the fitting and calculation of a new machine-learning interatomic potential (ML-IAP) based on proper orthogonal descriptors (POD). The implementation is located in the folder src/ML-POD, while the documentation is located in the folder doc/src.  The new features of this pull request include:

- Users can train POD potentials inside LAMMPS. 
- Users can perform MD simulations with the trained POD potentials.
- Users can combine POD and SNAP to create hybrid POD-SNAP potentials. 
- Users can make linear and quadratic POD, POD-SNAP potentials. 

**Related Issue(s)**

This pull request does not address any open GitHub issue. 

**Author(s)**

Ngoc Cuong Nguyen,  Department of Aeronautics and Astronautics, Massachusetts Institute of Technology, 
77 Massachusetts Avenue, Cambridge, 02139, MA. Email: cuongng@mit.edu and exapde@gmail.com

Andrew Rohskopf, Sandia National Laboratories. Email: adrohsk@sandia.gov.  

**Licensing**

**Backward Compatibility**

This pull request should not break backward compatibility for inputs. 

**Implementation Notes**

In this pull request, three new C++ classes were created, including mlpod, pair_mlpod, and fitpod_command, to implement the fitting command and pair style for POD. Examples are provided in the examples/mlpod folder to demonstrate the implementation and its performance.  

This pull request should not affect other features in LAMMPS. 

**Post Submission Checklist**

The original code of this pull request is hosted at https://github.com/cesmix-mit. The Github repository https://github.com/cesmix-mit/pod_examples contains additional examples for this pull request. 

More information about the authors can be found on their personal websites  https://www.mit.edu/~cuongng/ and https://rohskopf.github.io/. Technical details about the methodology implemented in this pull request can be found in the paper https://arxiv.org/abs/2209.02362. 

The research of Dr. Nguyen is supported by Center for the Exascale Simulation of Materials in Extreme Environments https://computing.mit.edu/cesmix/.


cmake/CMakeLists.txt
```diff
@@ -431,8 +432,8 @@ if(PKG_MSCG OR PKG_ATC OR PKG_AWPMD OR PKG_ML-QUIP OR PKG_LATTE OR PKG_ELECTRODE
     set_target_properties(linalg PROPERTIES OUTPUT_NAME lammps_linalg${LAMMPS_MACHINE})
     set(BLAS_LIBRARIES "$<TARGET_FILE:linalg>")
     set(LAPACK_LIBRARIES "$<TARGET_FILE:linalg>")
-  else()
-    list(APPEND LAPACK_LIBRARIES ${BLAS_LIBRARIES})
+  else() 
+    list(APPEND LAPACK_LIBRARIES ${BLAS_LIBRARIES})
```
junghans:
```suggestion
  else()
    list(APPEND LAPACK_LIBRARIES ${BLAS_LIBRARIES})
```

cmake/Modules/StyleHeaderUtils.cmake
```diff
@@ -168,7 +168,7 @@ function(DetectBuildSystemConflict lammps_src_dir)
   if(ARGC GREATER 1)
     list(REMOVE_AT ARGV 0)
     foreach(SRC_FILE ${ARGV})
-        get_filename_component(FILENAME ${SRC_FILE} NAME)
+        get_filename_component(FILENAME ${SRC_FILE} NAME)
```
junghans:
```suggestion
        get_filename_component(FILENAME ${SRC_FILE} NAME)
```

cmake/CMakeLists.txt
```diff
@@ -624,6 +625,10 @@ if(PKG_ELECTRODE)
   target_link_libraries(lammps PRIVATE ${LAPACK_LIBRARIES})
 endif()
 
+if(PKG_ML-POD)
```
junghans: I would prefere:
```suggestion
if(PKG_ELECTRODE OR PKG_ML-POD)
```
and then dropping the duplication.

cmake/CMakeLists.txt
junghans:
```suggestion
  else()
    list(APPEND LAPACK_LIBRARIES ${BLAS_LIBRARIES})
```

akohlmey:
@exapde @rohskopf how close do you think this PR is to ready for review from your perspective?
From just a cursory look, it seems the coding style and conventions used are still rather far from what it common practice in LAMMPS, so you may want to check out the recommendations at https://docs.lammps.org/Modify_style.html

Some highlights:
- we don't use iostreams and stringstreams, instead there are the Tokenizer/ValueTokenizer and TextFileReader classes for reading and fmt::format and util::logmesg for output
- any output to the screen/logfile should be  protected by `if (comm->me == 0)`
- the examples seem to need significant streamlining and cleanup
- there should be no `using std::*`. Anything in namespace std:: should be visible as being there.
- class internal methods should be just called as `method()`, not as `this->method()`

rohskopf:
Thanks @akohlmey. We'll make the suggested code style and convention changes before this is ready for review, and we'll make sure that everything at https://docs.lammps.org/Modify_style.html is satisfied.

akohlmey:
> Thanks @akohlmey. We'll make the suggested code style and convention changes before this is ready for a review, and that everything at https://docs.lammps.org/Modify_style.html is satisfied.

Don't hesitate to ask if you have questions on addressed in the manual or need clarifications. Not everything is set in stone and not everything is documented to the last detail. If I have the time, I can also have a pass of converting code if this concerns LAMMPS internals like file parsing. There are some additional suggestions here: https://docs.lammps.org/Developer_notes.html and with the individual class/function documentation in the developer guide section.

akohlmey:
FYI, I have changed the status of the PR to "Draft". You can click on "Ready for review" when you want me to take a look.

cmake/CMakeLists.txt
```diff
@@ -624,6 +625,10 @@ if(PKG_ELECTRODE)
   target_link_libraries(lammps PRIVATE ${LAPACK_LIBRARIES})
 endif()
 
+if(PKG_ELECTRODE OR PKG_ML-POD)
```
junghans, line 642: Now you can drop the block above.

exapde:
Hi @akohlmey, 

Drew and I made many changes for our package to meet LAMMPS's style requirements. We think this is ready for review but have some questions/concerns:

- For collating and allocating LAPACK arrays it was convenient to build our own neighbor list in compute_podfit, which fits potentials. This is only done when fitting potentials with `compute_podfit`. This fitting will only involve systems of hundreds of atom at most, so we hope our O(N^2) neighbor list is not problematic. The pair style of course uses LAMMPS neigh lists. Is it okay if we use our own neighbor list in compute_podfit?

- For similar convenience we sometimes use our own memory allocator TemplateMalloc which is just a malloc template, instead of the LAMMPS memory allocator. Is it okay if we use this for our custom arrays?

We appreciate your time reviewing our code. Many thanks!

akohlmey:
> * For collating and allocating LAPACK arrays it was convenient to build our own neighbor list in compute_podfit, which fits potentials. This is only done when fitting potentials with `compute_podfit`. This fitting will only involve systems of hundreds of atom at most, so we hope our O(N^2) neighbor list is not problematic. The pair style of course uses LAMMPS neigh lists. Is it okay if we use our own neighbor list in compute_podfit?

I would suggest to add a test in the code that prints a warning (with error->warning()) on MPI rank 0, if there are more than, say, 1000 atoms in the compute group. The warning could say that the calculation may be slow with so many atoms due to using an O(N**2) algorithm.

> * For similar convenience we sometimes use our own memory allocator TemplateMalloc which is just a malloc template, instead of the LAMMPS memory allocator. Is it okay if we use this for our custom arrays?

This seems very similar to what the code in memory.h does, so it would make it easier for others to read, study, and debug or extend your code if it uses the create/destroy/grow functions from the Memory class.

What worries me particularly about your template code is that it contains references to using CUDA and HIP, but I cannot see any use of that anywhere in the code. Is this something that is planned to add or already exists but is not yet included?

> We appreciate your time reviewing our code. Many thanks!

I will take a look later today.

cmake/CMakeLists.txt
```diff
@@ -633,6 +634,10 @@ if(PKG_ELECTRODE)
   target_link_libraries(lammps PRIVATE ${LAPACK_LIBRARIES})
```
akohlmey: This block is redundant and needs to be removed.

cmake/CMakeLists.txt
```diff
@@ -633,6 +634,10 @@ if(PKG_ELECTRODE)
...
 endif()
 
+if(PKG_ELECTRODE OR PKG_ML-POD)
```
akohlmey, line 642: Same as for the ELECTRODE package, there also needs to be support for the GNU make based build system so that the desired BLAS/LAPACK version is used.  You can use the contents from lib/electrode as template. Since the needs are equivalent.

examples/pod/README.md
```diff
@@ -0,0 +1,21 @@
...
+    mkdir build-pod
+    cd build-pod
+    cmake -C ../cmake/presets/basic.cmake -D PKG_ML-POD=on ../cmake
+    cmake --build .
+
+### Fit a POD potential for tantalum
+
+Go to `lammps/examples/pod/Ta` directory and run 
+
+    lmp < in.podfit
```
akohlmey: We are trying to phase out recommending I/O redirection as a way to read input files. It has several disadvantages and is mostly a leftover from when LAMMPS was written in Fortran. Please use the -in flag.

examples/pod/Ta/README.md
```diff
@@ -0,0 +1,14 @@
+### POD example for Ta
+
+We will fit a potential to the `Ta` training data in the `XYZ` directory, which houses `.xyz` files 
+of the training data taken from [the FitSNAP repo](https://github.com/FitSNAP/FitSNAP/tree/master/examples/Ta_XYZ/XYZ)
+
+Fit POD with
+
+    lmp < in.podfit
```
akohlmey: see the other comment about the -in flag and I/O redirection.

src/ML-POD/pair_pod.h
```diff
@@ -0,0 +1,177 @@
...
+// clang-format off
+PairStyle(pod,CPairPOD);
+// clang-format on
+#else
+
+#ifndef LMP_PAIR_POD_H
+#define LMP_PAIR_POD_H
+
+#include "pair.h"
+#include "pod.h"
```
akohlmey: There should be no #include statement in for pod.h in the header for pair style pod.
Instead forward declarations should be used, if needed and pod.h only included from the .cpp file.
This one is particularly problematic since it include podcommon.h which has unprotected defines.
The style_pair.h file will include all pair_<style>.h header files at the same time and thus we have to be **extremely** careful to avoid including non-system and non-core LAMMPS headers because there is a high risk of conflicts and unexpected side effects. This has happened in the past multiple times already.

src/ML-POD/pair_pod.h
```diff
@@ -0,0 +1,177 @@
+#ifdef PAIR_CLASS
+// clang-format off
+PairStyle(pod,CPairPOD);
```
akohlmey: I am very concerned about using names that are initialisms, especially if the number of characters is small.
So the name "pod" worries me and if at all possible we should be using something longer to reduce the probability of conflicts and confusion. We have a few legacy cases ("lj", "sw") where "lj" is acceptable because it is a widely used abbreviation and "sw" because it has been used for a very, very long time. But for new contributions, we prefer something longer. I don't have a really good suggestion, but already using "mlpod" instead of "pod" would bother me less.

src/ML-POD/pair_pod.h
```diff
@@ -0,0 +1,177 @@
...
+
+#ifndef LMP_PAIR_POD_H
+#define LMP_PAIR_POD_H
+
+#include "pair.h"
+#include "pod.h"
+
+namespace LAMMPS_NS {
+
+  class CPairPOD : public Pair {
```
akohlmey: The convention for pair style class names is: PairName, *not* CPairName. Please change.

src/ML-POD/pair_pod.h
```diff
@@ -0,0 +1,177 @@
...
+  private:
+      std::vector<std::string> globVector(const std::string& pattern, std::vector<std::string> & files);
```
akohlmey: These member functions do not seem to be implemented or used in the pair style. So the prototypes should be removed.

src/ML-POD/pair_pod.h
```diff
@@ -0,0 +1,177 @@
...
+
+      int podneighborcount(double *r, double rcutsq, int nx, int N, int dim);
+      int podneighborlist(int *neighlist, int *numneigh, double *r, double rcutsq, int nx, int N, int dim);
+
+      void get_data(std::vector<std::string> species);
+
+      void read_data_files(std::string data_file, std::vector<std::string> species);
+
+  public:
+      struct datastruct {
```
akohlmey: This struct should be declared in the common header and the pair style should hold a pointer to it. This way only a forward declaration is needed (PIMPL!).

src/ML-POD/pair_pod.h
```diff
@@ -0,0 +1,177 @@
...
+      void compute(int, int) override;
+
+      void settings(int, char **) override;
+      void coeff(int, char **) override;
+      void init_style() override;
+      double init_one(int, int) override;
+      double memory_usage() override;
+      void *extract(const char *, int &) override;
+
+      void InitPairPOD(std::string pod_file, std::string coeff_file);
```
akohlmey: Pass string classes as const references to avoid useless copies.

src/ML-POD/pair_pod.h
```diff
@@ -0,0 +1,177 @@
...
+
+      int backend=1;
```
akohlmey: The LAMMPS conventions is that the initialization of class members should be in the constructor. Pointers usually in the initializer list and variables in the constructor itself.

src/ML-POD/pair_pod.cpp
```diff
@@ -0,0 +1,370 @@
...
+#include "atom.h"
+#include "comm.h"
+#include "error.h"
+#include "force.h"
+#include "memory.h"
+#include "neigh_list.h"
+#include "neighbor.h"
+#include "tokenizer.h"
+
+#include <glob.h>
```
akohlmey: This header doesn't seem to be needed. Only required headers should be included.

src/ML-POD/pair_pod.cpp
```diff
@@ -0,0 +1,370 @@
...
+
+#include <cmath>
+#include <cstring>
+#include <sys/time.h>
```
akohlmey: What is sys/time.h included for?

src/ML-POD/pod.cpp
```diff
@@ -0,0 +1,3763 @@
...
+
+#include <string>
+#include <vector>
+#include <iomanip>
+#include <limits>
+#include <glob.h>
+#include <random>
+#include <algorithm>
+
+#define _USE_MATH_DEFINES
```
akohlmey: Please do not use this define and do not include math.h but cmath.
For math constants, you should be including the LAMMPS provided "math_const.h" header and the constants included therein. This way we have consistent math constants across platforms and C-library implementations.

src/ML-POD/compute_podfit.cpp
```diff
@@ -0,0 +1,1564 @@
+/***************************************************************************
+               CESMIX-MIT Project
+
+ Contributing authors: Ngoc-Cuong Nguyen (cuongng@mit.edu, exapde@gmail.com)
+ ***************************************************************************/
+
+#ifndef __COMPUTEPODFIT
```
akohlmey: what is this construct used for?

src/ML-POD/compute_podfit.cpp
```diff
@@ -0,0 +1,1564 @@
...
+    }
+
+    out.write( reinterpret_cast<char*>( &a[0] ), sizeof(T) * N );
+
+    out.close();
+  }
+  */
+}
+
+bool CPODFIT::is_a_number(std::string line)
```
akohlmey: This is redundant. There are "utils::is_integer()" and "utils::is_double()" and for consistent conversion to numbers you can use either the "ValueTokenizer" class or the "utils::numeric()" or "utils:inumeric()" functions.

src/ML-POD/compute_podfit.cpp
```diff
@@ -0,0 +1,1564 @@
...
+  utils::logmesg(lmp, "error analysis for test data set: {}\n", fitting_weights[4]);
+  utils::logmesg(lmp, "energy/force calculation for training data set: {}\n", fitting_weights[5]);
+  utils::logmesg(lmp, "energy/force calculation for test data set: {}\n", fitting_weights[6]);
+  utils::logmesg(lmp, "fitting weight for energy: {}\n", fitting_weights[0]);
+  utils::logmesg(lmp, "fitting weight for force: {}\n", fitting_weights[1]);
+  utils::logmesg(lmp, "fitting weight for stress: {}\n", fitting_weights[2]);
+  utils::logmesg(lmp, "**************** End of Data File ****************\n");
+}
+
+std::vector<std::string> CPODFIT::globVector(const std::string& pattern, std::vector<std::string> & files)
```
akohlmey: Convenience functions like this should not be class members. If there is only one application, they can be defined as static functions at the top of the file.  Also, the same functionality can be achieved with using "platform::list_directory()" and looping over the resulting vector of strings and processing only files that would match a "regex" pattern with "utils::strmatch()". That would be better code reuse and is platform neutral.

src/ML-POD/compute_podfit.h
```diff
@@ -0,0 +1,214 @@
...
+#include "compute.h"
+#include "pod.h"
+
+namespace LAMMPS_NS {
+
+class CPODFIT : public Compute {
+private:
+
+  class NeighList *list;
+  std::vector<std::string> globVector(const std::string& pattern, std::vector<std::string> & files);
```
akohlmey: Convenience functions like this and the following should not be class members. Rather they should be static functions at the top of the implementation, or - if possible - replaced by existing convenience functions in the "utils" and "platform" namespaces of LAMMPS.

src/ML-POD/compute_podfit.h
```diff
@@ -0,0 +1,214 @@
...
+
+  bool is_a_number(std::string line);
+
+public:
+  struct datastruct {
```
akohlmey: This seems to be almost the same struct as used by the pair style. Why not use one common definition?

src/ML-POD/compute_podfit.h
```diff
@@ -0,0 +1,214 @@
...
+      TemplateFree(lattice, backend);
+      TemplateFree(energy, backend);
+      TemplateFree(stress, backend);
+      TemplateFree(position, backend);
+      TemplateFree(force, backend);
+      TemplateFree(atomtype, backend);
+    }
+  };
+
+  struct neighborstruct {
```
akohlmey: Same as for the pair style, headers for compute styles should not be defining and implementing structs. Instead there should be a pointer to it and a forward declaration.

src/ML-POD/compute_podfit.h
```diff
@@ -0,0 +1,214 @@
...
+
+#ifndef LMP_COMPUTE_PODFIT_H
+#define LMP_COMPUTE_PODFIT_H
+
+#include "compute.h"
+#include "pod.h"
+
+namespace LAMMPS_NS {
+
+class CPODFIT : public Compute {
```
akohlmey: The naming convention for compute style classes is ComputeName.

I don't see any "compute" member function declaration here. What is the "consumable property" that this compute style computes, is it a global scalar, a global vector, or a local vector/array or a per-atom vector/array?
How is the compute triggered and how the results accessed? I cannot find anything useful in the documentation either. That one reads like this "compute" should rather be a CommandStyle class. Please explain.

src/ML-POD/pair_pod.cpp
```diff
@@ -0,0 +1,370 @@
...
+  podptr->podArraySetValue(gd, 0.0, nd1234);
+
+  for (int ii = 0; ii < inum; ii++) {
+    int i = ilist[ii];
+    int jnum = numneigh[i];
+
+    // allocate temporary memory
+
+    if (nijmax < jnum) {
+      nijmax = PODMAX(nijmax, jnum);
```
akohlmey: Why use PODMAX and not the MAX macro defined in pointers.h

src/ML-POD/pair_pod.cpp
```diff
@@ -0,0 +1,370 @@
...
+  for (int ii = 0; ii < n+1; ii++)
+    for (int jj = 0; jj < n+1; jj++)
+      scale[ii][jj] = 1.0;
+  allocated = 1;
+
+  if (narg != 4 + atom->ntypes) error->all(FLERR,"Incorrect args for pair coefficients");
+
+  map_element2type(narg-4,arg+4);
+
+  //std::cout<<map[0]<<std::endl;
```
akohlmey: Please remove commented out debug code unless there is a permanent legitimate need for such blocks (this doesn't look like it). Then either put them into a local/static convenience function, or use #if 0 / #endif.  Or make the #if depend on some local define at the top (say:  #define POD_DEBUG_MAP 0) where you can change the value to trigger getting debug info. You also want to avoid multiple outputs for parallel runs, so check for comm->me == 0.

src/ML-POD/pair_pod.cpp
```diff
@@ -0,0 +1,370 @@
...
+  TemplateMalloc(&typeai, nablockmax, backend);
+  TemplateMalloc(&tmpmem, szd, backend);
+}
+
+void CPairPOD::allocate_atommemory()
+{
+  TemplateMalloc(&forces, dim*nmaxatom, backend);
+  TemplateMalloc(&stress, 9, backend);
+  if (atommemory) {
+  TemplateMalloc(&atomtype, nmaxatom, backend);
```
akohlmey: Please check for proper indentation. LAMMPS has a clang-format configuration file in the src folder. You can use that with clang-format to re-indent and reformat your files according to LAMMPS conventions.

src/ML-POD/pod.cpp
```diff
@@ -0,0 +1,3763 @@
...
+// LAMMPS header files
+
+#include "comm.h"
+#include "error.h"
+#include "memory.h"
+#include "tokenizer.h"
+
+#include <string>
+#include <vector>
+#include <iomanip>
```
akohlmey: Please remove unused includes and sort them alphabetically.

src/ML-POD/compute_podfit.cpp
```diff
@@ -0,0 +1,1564 @@
+/***************************************************************************
```
akohlmey: All source filed need to have the LAMMPS copyright message that is commonly used for all source files.
Please note that there are two variants. One for headers, one for .cpp files. More project/source specific comments can follow.

src/ML-POD/compute_podfit.cpp
```diff
@@ -0,0 +1,1564 @@
...
+}
+
+template <typename T> void CPODFIT::writearray2file(const char* filename, T *a, int N, int backend)
+{
+
+  if (N > 0){
+
+    FILE *fp = fopen(filename, "wb");
+
+    fwrite( reinterpret_cast<char*>( &a[0] ), sizeof(T) * N, 1, fp);
```
akohlmey: Why does this have to be template function? There is just one use of it. In fact, the function should be removed and its equivalent code put right where it is called.

akohlmey (changes requested):
This pull request still looks like it needs a bit of work. I may have more suggestions, but it is getting difficult to put in more comments and anticipate what this would look like after all existing suggestions have been processed. So I'd rather stop now and wait until we have resolved all comments and then do another pass.

As before, please keep in mind that only some requested changes are "hard" requests, while others are just suggestions. So don't hesitate to ask, if you disagree or have a better alternative in mind.

exapde:
> I would suggest to add a test in the code that prints a warning (with error->warning()) on MPI rank 0, if there are more than, say, 1000 atoms in the compute group. The warning could say that the calculation may be slow with so many atoms due to using an O(N**2) algorithm.

We will add the warning message. 

> This seems very similar to what the code in memory.h does, so it would make it easier for others to read, study, and debug or extend your code if it uses the create/destroy/grow functions from the Memory class.

We will use create/destroy/grow functions from the Memory class to allocate and delete memory. 

> What worries me particularly about your template code is that it contains references to using CUDA and HIP, but I cannot see any use of that anywhere in the code. Is this something that is planned to add or already exists but is not yet included?

Currently, there is no CUDA/HIP in the code. The code has been written in such way that the GPU implementation is relatively easy. In fact, the computation is done over a block of atoms simultaneously. For the current version of the code, we set the block size to 1. We plan to write a GPU implementation using Kokkos to allow the code to run on GPUs.

akohlmey:
> Currently, there is no CUDA/HIP in the code. The code has been written in such way that the GPU implementation is relatively easy. In fact, the computation is done over a block of atoms simultaneously. For the current version of the code, we set the block size to 1. We plan to write a GPU implementation using Kokkos to allow the code to run on GPUs.

If you do a KOKKOS implementation, then you will be implementing this as a derived class and you'll have to use the device/host memory access methods that the Kokkos library provides. This means that you are not very likely able to keep this code anyway.

src/ML-POD/pair_pod.h
rohskopf: Fixed. Also used the same forward declaration in the new command style header `fitpod_command.h`.

src/ML-POD/pair_pod.h
rohskopf: Thanks for the suggestion - changed to `mlpod`.

src/ML-POD/pair_pod.h
rohskopf: Changed.

src/ML-POD/pair_pod.h
rohskopf: Removed this struct from the pair style entirely.
In general should we also use forward declaration of structs in all LAMMPS classes (e.g. command styles)?

src/ML-POD/pair_pod.h
rohskopf: Removed this `InitPairPOD` function and initialized variables and pointers in the constructor and `coeff()` function, respectively. Also fixes https://github.com/lammps/lammps/pull/3449#discussion_r1018241939

src/ML-POD/compute_podfit.cpp
rohskopf: I don't think it was used for anything.
We converted this compute to `fitpod_command.cpp` and removed the construct.

src/ML-POD/compute_podfit.h
rohskopf: This is declared in `fitpod_command.h` now, and it is only used for that command style.

cmake/CMakeLists.txt
exapde: Removed.

cmake/CMakeLists.txt
exapde: Thanks for the tip. We create the GNU Makefile for ML-POD based on lib/electrode.

src/ML-POD/pair_pod.h
exapde: Removed.

src/ML-POD/pair_pod.h
exapde: The InitPairPOD function was removed.

src/ML-POD/pair_pod.cpp
exapde: The header was removed.

src/ML-POD/pair_pod.cpp
exapde: The header was removed.

src/ML-POD/pod.cpp
exapde: Removed _USE_MATH_DEFINES and math.h. We now use cmath and math_const.h.

src/ML-POD/compute_podfit.cpp
exapde: Removed this function.

src/ML-POD/compute_podfit.cpp
exapde: It is defined as a static function at the top of the file.

src/ML-POD/compute_podfit.h
exapde: Function declaration is removed as It is defined as a static function at the top of the file.

src/ML-POD/pair_pod.cpp
exapde: PODMAX is replaced with MAX

src/ML-POD/pair_pod.cpp
exapde: Removed.

src/ML-POD/pod.cpp
exapde: Removed.

src/ML-POD/compute_podfit.cpp
exapde: Included the LAMMPS copyright message for all source files.

src/ML-POD/compute_podfit.cpp
exapde: The function is removed.

examples/pod/Ta/README.md
exapde: -in is used.

src/ML-POD/compute_podfit.h
exapde: We use CommandStyle class for fitpod.

src/ML-POD/pair_pod.cpp
exapde: The code is re-indented.

src/ML-POD/fitpod_command.cpp
```diff
@@ -0,0 +1,1554 @@
...
+    if (keywd == "randomize_training_data_set") fitting_weights[9] = utils::numeric(FLERR,words[1],false,lmp);
+    if (keywd == "randomize_test_data_set") fitting_weights[10] = utils::numeric(FLERR,words[1],false,lmp);
+
+    // other settings
+
+    if (keywd == "file_format") file_format = words[1];
+    if (keywd == "file_extension") file_extension = words[1];
+    if (keywd == "path_to_training_data_set"){
+      utils::logmesg(lmp, "{}\n", words[1]);
+      training_path = words[1];
```
akohlmey: FYI, the functionality of joining words in a vector of strings exists as `utils::join_words()`.

src/ML-POD/fitpod_command.cpp
```diff
@@ -0,0 +1,1554 @@
...
+  data.num_atom_cumsum.resize(len+1);
+  podptr->podCumsum(&data.num_atom_cumsum[0], &data.num_atom[0], len+1);
+
+  data.num_config_cumsum.resize(nfiles+1);
+  podptr->podCumsum(&data.num_config_cumsum[0], &data.num_config[0], nfiles+1);
+
+  // convert all structures to triclinic system
+
+  int dim = 3;
+  double Qmat[dim*dim];
```
akohlmey: This is a so-called "variable length array" and thus not valid C++ (it is common in Fortran and tolerated by compilers like GCC in their standard "forgiving" modes since they are allowed in C98). Changing the preceding line into: `constexpr int dim = 3;`  would make it valid again.
An alternate approach to support variable length arrays is to use a `std::vector<double> Qmat(dim*dim);` declaration and then pass `Qmat.data()` when the array would be passed to a function that expects a pointer.

src/ML-POD/fitpod_command.cpp
```diff
@@ -0,0 +1,1554 @@
...
+
+std::vector<int> CFITPOD::shuffle(int start_in, int end_in, int num_in)
+{
+  int sz = end_in - start_in + 1;
+  std::vector<int> myvector(sz);
+
+  for (int i = 0; i<sz; i++)
+    myvector[i] = start_in + i;
+
+  unsigned seed = std::chrono::system_clock::now().time_since_epoch().count();
```
akohlmey: Please use `platform::walltime()` multiplied by a suitable factor for increased portability and consistency.

src/ML-POD/fitpod_command.cpp
```diff
@@ -0,0 +1,1554 @@
...
+  fclose(fp_errors);
+  fclose(fp_analysis);
+
+}
+
+void CFITPOD::error_analysis(datastruct data, double *coeff)
+{
+  int dim = 3;
+  double energy;
+  double force[dim*data.num_atom_max];
```
akohlmey: This is another variable length array. (Hint: g++ -Wvla will show them all)

src/ML-POD/fitpod_command.cpp
```diff
@@ -0,0 +1,1554 @@
...
+
+  int nfiles = data.data_files.size();  // number of files
+  int num_configs = data.num_atom.size(); // number of configurations in all files
+  //int nd12 = podptr->pod.nd1 + podptr->pod.nd2;
+  //int nd123 = podptr->pod.nd1 + podptr->pod.nd2 + podptr->pod.nd3;
+  //int nd1234 = podptr->pod.nd1 + podptr->pod.nd2 + podptr->pod.nd3 + podptr->pod.nd4;
+  //double effectivecoeff[nd1234];
+
+  int m = 8;
+  double outarray[m*num_configs];
```
akohlmey: and two more.

src/ML-POD/fitpod_command.cpp
```diff
@@ -0,0 +1,1554 @@
...
+  int nd22 = podptr->pod.nd22;
+  int nd23 = podptr->pod.nd23;
+  int nd24 = podptr->pod.nd24;
+  int nd33 = podptr->pod.nd33;
+  int nd34 = podptr->pod.nd34;
+  int nd44 = podptr->pod.nd44;
+  int nd1234 = nd1+nd2+nd3+nd4;
+  int nd = podptr->pod.nd;
+
+  double newcoeff[nd];
```
akohlmey: and another VLA

src/ML-POD/fitpod_command.cpp
```diff
@@ -0,0 +1,1554 @@
...
+  utils::logmesg(lmp, "**************** End of Error Calculation ****************\n");
+
+  print_analysis(data, outarray, errors);
+}
+
+void CFITPOD::energyforce_calculation(datastruct data, double *coeff)
+{
+  int dim = 3;
+  double energy;
+  double force[1+dim*data.num_atom_max];
```
akohlmey: This seems to be the last VLA in this file.

src/ML-POD/pair_pod.h
```diff
@@ -0,0 +1,125 @@
...
+   DE-AC04-94AL85000 with Sandia Corporation, the U.S. Government retains
+   certain rights in this software.  This software is distributed under
+   the GNU General Public License.
+
+   See the README file in the top-level LAMMPS directory.
+------------------------------------------------------------------------- */
+
+#ifdef PAIR_CLASS
+// clang-format off
+PairStyle(mlpod,PairPOD);
```
akohlmey: When you change the name of the pair style, all "derived" names (name of source files, name of Class, include guards, name of documentation file, etc.) should be changed as well so that they are consistent.

src/ML-POD/pair_pod.h
```diff
@@ -0,0 +1,125 @@
...
+#ifndef LMP_PAIR_POD_H
+#define LMP_PAIR_POD_H
+
+#include "pair.h"
+
+namespace LAMMPS_NS {
+
+  class PairPOD : public Pair {
+  public:
+      PairPOD(class LAMMPS *);
```
akohlmey: Indentation seems inconsistent here. You can just run `clang-format -i` on the header files to have them reformatted.

src/ML-POD/pair_pod.h
```diff
@@ -0,0 +1,125 @@
...
+      void compute(int, int) override;
+
+    void settings(int, char **) override;
+    void coeff(int, char **) override;
+    void init_style() override;
+    double init_one(int, int) override;
+    double memory_usage() override;
+    void *extract(const char *, int &) override;
+
+    int dim = 3;
```
akohlmey: Please move all initializations from the header files to either initializer lists of the constructor (for pointers) or explicit statements in the constructor.

src/ML-POD/pair_pod.h
```diff
@@ -0,0 +1,125 @@
...
+    int atommemory = 0;
+    int podpairlist=0;
+    int analysis = 0;
+    int runMD = 0;
+    int savecalculation = 0;
+    int savefrequency = 0;
+    int randomrotation = 0;
+
+    double energy=0.0;    // potential energy
+    double *forces=NULL;   // atomic forces
```
akohlmey: Please use `nullptr` instead of `NULL`

akohlmey (changes requested):
@exapde @rohskopf Thanks for your many and diligent updates.

Unfortunately, when looking over the improved branch, I still have a few more requests for changes. Specifically, the variable length arrays need to be replaced and the class initialization changed to the LAMMPS conventions. Please also note, that when enabling warnings, I see quite a few "unused variable" warnings. Those variables should be removed. There also are a few "unused parameter" warnings. If those cannot be safely removed, you should put the name of the variable in C-style comments (`double /*var*/`) to silence those warnings.

akohlmey:
@exapde @rohskopf I have a little extra time this week. So I could help implementing some of the suggested changes.

akohlmey:
> @exapde @rohskopf I have a little extra time this week. So I could help implementing some of the suggested changes.

I have submitted a pull request in the fork with a set of (simple) changes.

src/ML-POD/pair_pod.h
exapde: Replaced NULL with nullptr in all source files.

src/ML-POD/pair_pod.h
exapde: Moved  initializations to the constructor.

src/ML-POD/pair_pod.h
exapde: > > @exapde @rohskopf I have a little extra time this week. So I could help implementing some of the suggested changes.
> 
> I have submitted a pull request in the fork with a set of (simple) changes.

Hi @ akohlmey! Thanks for your help. It is much  appreciated.

exapde:
> >

src/ML-POD/pair_pod.h
exapde: Source files are reformatted.

src/ML-POD/pair_pod.h
exapde: Changed the pair style name from pod to mlpod and updated POD's documentation files to reflect the name change. Changed the example folder name from pod to mlpod.

exapde:
Sorry I accidentally closed the pull request.

exapde:
> > @exapde @rohskopf I have a little extra time this week. So I could help implementing some of the suggested changes.
> 
> I have submitted a pull request in the fork with a set of (simple) changes.

Hi @akohlmey! Many thanks for your suggestions and changing ML-POD. We appreciate your time reviewing our code and giving us your comments. We made a number of changes for ML-POD based on your recent comments. Please let us know if there are other things we need to change ML-POD.

src/ML-POD/fitpod_command.cpp
```diff
@@ -0,0 +1,1617 @@
...
+
+/* ---------------------------------------------------------------------- */
+
+void CFITPOD::read_data_file(double *fitting_weights, std::string &file_format,
+                             std::string &file_extension, std::string &test_path,
+                             std::string &training_path, const std::string &data_file)
+{
+  std::string datafilename = data_file;
+  FILE *fpdata;
+  //if (comm->me == 0){
```
akohlmey: This commented out code should be reinstated. When running in parallel you do not want every parallel process to open the same file. This can wreak havoc on parallel file system and be worse on non-parallel networked file systems. Rather you want only MPI rank 0 to read it and then broadcast the data to all MPI ranks, or have MPI rank 0 read *and* parse the file and then broadcast the resulting parsed data.

akohlmey:
> Please let us know if there are other things we need to change ML-POD.

@exapde @rohskopf 

Yes, there are more changes needed to address issues I've come across. I've added a unit test for the pair style and noticed some memory leaks, uninitialized data accesses, file pointer leaks, and some programming style issues. Also, I am trying to make the screen output a bit nicer looking. Some of it is done with a lot of complex code, which can be simplified using fmt library features.

In the process I noticed that the pair style does not support per-atom energies and per-atom stress. There should be some code added that will print a warning. I will look into that, too. This also needs to be documented, if it is not. The same applies to `pair_modify nofdotr`.

There also are some more build system and general LAMMPS "maintenance" changes needed. All of those are easier for me to just do them than to explain how to do them to you.

I'm adding comments for other changes that I think you can do or where you have to make a decision/change (or not).

src/ML-POD/fitpod_command.h
```diff
@@ -0,0 +1,155 @@
...
+    int *atomtype;
+
+    int training = 1;
+    int normalizeenergy = 1;
+    int training_analysis = 1;
+    int test_analysis = 1;
+    int training_calculation = 0;
+    int test_calculation = 0;
+    int randomize = 1;
+    double percentage = 1.0;
```
akohlmey: This seems to be a *fraction* (range 0.0-1.0) and not a *percentage* (range 0.0 to 100.0).

src/ML-POD/fitpod_command.cpp
```diff
@@ -0,0 +1,1617 @@
...
+    // settings for fitting weights
+
+    if (keywd == "fitting_weight_energy") fitting_weights[0] = utils::numeric(FLERR,words[1],false,lmp);
+    if (keywd == "fitting_weight_force") fitting_weights[1] = utils::numeric(FLERR,words[1],false,lmp);
+    if (keywd == "fitting_weight_stress") fitting_weights[2] = utils::numeric(FLERR,words[1],false,lmp);
+    if (keywd == "error_analysis_for_training_data_set") fitting_weights[3] = utils::numeric(FLERR,words[1],false,lmp);
+    if (keywd == "error_analysis_for_test_data_set") fitting_weights[4] = utils::numeric(FLERR,words[1],false,lmp);
+    if (keywd == "energy_force_calculation_for_training_data_set") fitting_weights[5] = utils::numeric(FLERR,words[1],false,lmp);
+    if (keywd == "energy_force_calculation_for_test_data_set") fitting_weights[6] = utils::numeric(FLERR,words[1],false,lmp);
+    if (keywd == "percentage_training_data_set") fitting_weights[7] = utils::numeric(FLERR,words[1],false,lmp);
```
akohlmey: This seems to be a *fraction* and not a *percentage*. See my other comment.

src/ML-POD/fitpod_command.cpp
exapde: Thanks for spotting them out. These lines of code are reinstated.

src/ML-POD/fitpod_command.h
exapde: _Percentage_ is changed to _fraction_.

src/ML-POD/fitpod_command.cpp
exapde: _percentage_ is replaced with _fraction_ in all source files.

exapde:
> Yes, there are more changes needed to address issues I've come across. I've added a unit test for the pair style and noticed some memory leaks, uninitialized data accesses, file pointer leaks, and some programming style issues. Also, I am trying to make the screen output a bit nicer looking. Some of it is done with a lot of complex code, which can be simplified using fmt library features.

Thanks a lot for adding the unit test for the pair style. It is helpful indeed. How do we run the unit test to spot out memory leaks, uninitialized data accesses, file pointer leaks? We would like to fix those issues with the help of the unit test.

exapde:
> In the process I noticed that the pair style does not support per-atom energies and per-atom stress. There should be some code added that will print a warning. I will look into that, too. This also needs to be documented, if it is not. The same applies to `pair_modify nofdotr`.

Is it okay to add a warning message in the _coeff_ function of the pair style?

We updated the pair style documentation to indicate that the pair style does not support per-atom energies and per-atom stress.

exapde:
> There also are some more build system and general LAMMPS "maintenance" changes needed. All of those are easier for me to just do them than to explain how to do them to you.

Thank you so much for your help.

akohlmey:
> Is it okay to add a warning message in the _coeff_ function of the pair style?

At that point you cannot know whether per-atom energies or stress are requested.
Don't worry about this one. I have an idea and will implement in a feature branch with my collected updates and fixes to this PR.

akohlmey:
> Thanks a lot for adding the unit test for the pair style. It is helpful indeed. How do we run the unit test to spot out memory leaks, uninitialized data accesses, file pointer leaks? We would like to fix those issues with the help of the unit test.

I have already done this. You don't have access to those changes yet. This is another of these issues that are easier to do for me than to explain. The whole pair style testing procedure is a bit ad hoc and was created to test a *lot* of input variations and features of pair styles (and other force styles and some fixes) very efficiently and enforce consistency. Because of that, the procedure is a bit convoluted and requires some experience to understand and then fix issues.

Available documentation is here: https://docs.lammps.org/Build_development.html and here: https://docs.lammps.org/Developer_unittest.html
It probably will make more sense to you, once you see the changes in my branch.
I will submit it soon, even though I am not yet finished trying out and checking things, but that way we avoid that I have to deal with too many merge conflicts whenever you make some updates.

akohlmey:
@exapde @rohskopf I think that after the changes from https://github.com/cesmix-mit/lammps/pull/2 are merged, this PR is ready for a general review by the other LAMMPS developers. There are still several parts of the code, that could be done differently and more according to my personal preferences, but - as far as I am concerned - this second bunch of changes brings the branch to a state where - at least with the checks and tools that we usually use - it meets the required coding style requirements and works well enough on all tested platforms.

Further updates/cleanups/refactoring can be done later.

Please note, that I am mostly looking at technical issues, consistency, and other formal details.
Any scientific assessment and overall LAMMPS code design concerns are the forte of @athomps and @sjplimp

exapde:
> @exapde @rohskopf I think that after the changes from [cesmix-mit#2](https://github.com/cesmix-mit/lammps/pull/2) are merged, this PR is ready for a general review by the other LAMMPS developers. There are still several parts of the code, that could be done differently and more according to my personal preferences, but - as far as I am concerned - this second bunch of changes brings the branch to a state where - at least with the checks and tools that we usually use - it meets the required coding style requirements and works well enough on all tested platforms.

Hi @akohlmey! Many thanks for your comments and your changes to improve ML-POD, so that it meets the style requirements and works well on all tested platforms.

exapde:
> Further updates/cleanups/refactoring can be done later.

Totally agree. 

> Please note, that I am mostly looking at technical issues, consistency, and other formal details. Any scientific assessment and overall LAMMPS code design concerns are the forte of @athomps and @sjplimp

We wish to receive comments from @athomps, @sjplimp, and other lammps developers as well.

akohlmey (approved):

sjplimp:
@exapde Thanks for submitting this new packages and ML potential.  @athomps is better to comment on the potential and fitting command itself.  My two comments are about compatibility.
1) I think the examples dir would be better as examples/PACKAGES/mlpod (or just pod), rather than examples/mlpod.
The dirs under examples tend to be very simple ones that the developers of LAMMPS have created, which includes the
ones that have potential names (reaxff, meam, comb, snap, etc) which LAMMPS developers worked on.
2) Is the name of the potential MLPOD or just POD?  If the latter, then I suggest that it be a "pair_styled pod" command
and src/ML-POD/pair_pod.cpp/h files.  LIkewise the name of the fitting command (and files) should just be fitpod, not fitpod_command.  I assume mlpod.cpp/h is an auxiliary supporting class.  That name is fine but it could also just be pod.cpp/h

akohlmey:
@sjplimp we changed the style and file names from pod to mlpod to reduce the risk of possible name conflicts. Mlpod follows the naming pattern of mliap. I would very much like to keep this.

sjplimp:
Well, pair mliap is not really a potential by itself, it is an umbrella for plugging different descriptors and models into, to create a potential.  The other 5 ML-* dirs are stand-alone potentials, and their names are pair hdnnp, pair pace, pair quip, pair rann, and pair snap. I.e. not prefixed by ML.  Having said that, "pod" is a pretty short, generic name.  So I say let @athomps make the decision on the naming.  What's the argument for fitpod_command instead of just fitpod as a command name?

akohlmey:
> Well, pair mliap is not really a potential by itself, it is an umbrella for plugging different descriptors and models into, to create a potential. The other 5 ML-* dirs are stand-alone potentials, and their names are pair hdnnp, pair pace, pair quip, pair rann, and pair snap.

Well, the subject of the PR says "ML potentials with proper orthogonal descriptors", so I think "mlpod" is just as justified as "pod".

That said, my main worry is about a "mlpod.h" -> "pod.h" rename. *That* is where the big risk for conflict is. So if that file (and class) can keep its name, I can live with changing the name of the pair style back.

> I.e. not prefixed by ML. Having said that, "pod" is a pretty short, generic name. So I say let @athomps make the decision on the naming.
> What's the argument for fitpod_command instead of just fitpod as a command name?

The name of the command *is* "fitpod". Since I already had so many requests for changes, I didn't ask for renaming that, too.

lib/linalg/dgelss.f
```diff
@@ -254,11 +254,11 @@ SUBROUTINE DGELSS( M, N, NRHS, A, LDA, B, LDB, S, RCOND, RANK,
 *
 *              Compute space needed for DGEQRF
                CALL DGEQRF( M, N, A, LDA, DUM(1), DUM(1), -1, INFO )
-               LWORK_DGEQRF=DUM(1)
+               LWORK_DGEQRF = INT( DUM(1) )
```
athomps: Are these changes necessary? If not, better to avoid changing files unnecessarily. Is this like an upgrade to BLAS. If so, can we add a note to the README file with a BLAS version number?

akohlmey: I cherry-picked this from a different branch. This adds two new LAPACK functions to the linalg library that are used by the code here. The LAPACK version update was part of the change set. Including the entire change set avoids a merge conflict.

examples/mlpod/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+    cd path/to/lammps
+    mkdir build-pod
+    cd build-pod
+    cmake -C ../cmake/presets/basic.cmake -D PKG_ML-POD=on ../cmake
+    cmake --build .
+
+### Fit a POD potential for tantalum
+
+Go to `lammps/examples/pod/Ta` directory and run
```
athomps: This file will need to be edited to reflect final agreed path to Ta dir.

lib/linalg/dgelss.f
akohlmey: > can we add a note to the README file with a BLAS version number?

Done in 9b7b45bdea6b4ab2359c72d94cc63fe79fb6cb52

examples/mlpod/README.md
exapde: examples/mlpod/README.md includes a link to another Github repository that hosts a number of other examples for fitting and running POD. These examples are not included in ML-POD because the training data is considerably bigger than the Ta data. There is a separate README.md in the Ta dir for instructions on how to fit POD and run MD.

exapde: Yes, we'll do that when we agree on the example path.

sjplimp:
ok - re:naming conventions, I think we are converging.  I am suggesting:
(1) Leaving the auxiliary class files mlpod.cpp/h instead of pod.cpp/h is fine.  I agree "pod" might have conflicts
(2) Please rename the pair style file (and its class) to pair_pod.cpp/h --> input script command is "pair_style pod"
(3) Please rename the fitting command file (and its class) to fitpod.cpp/h  --> input script command is "fitpod"
(4) Doc page names and content should reflect pair_pod.rst and fitpod.rst
Sounds like both @akohlmey and me would be happy with this

exapde:
> @exapde Thanks for submitting this new packages and ML potential. @athomps is better to comment on the potential and fitting command itself. My two comments are about compatibility.

@sjplimp Thanks for your helpful comments. 
 
> 1. I think the examples dir would be better as examples/PACKAGES/mlpod (or just pod), rather than examples/mlpod.
>    The dirs under examples tend to be very simple ones that the developers of LAMMPS have created, which includes the
>    ones that have potential names (reaxff, meam, comb, snap, etc) which LAMMPS developers worked on.

We agree the example dir should be better as examples/PACKAGES/mlpod. We will transfer our examples to examples/PACKAGES/mlpod.

> 2. Is the name of the potential MLPOD or just POD?  If the latter, then I suggest that it be a "pair_styled pod" command
>    and src/ML-POD/pair_pod.cpp/h files.  LIkewise the name of the fitting command (and files) should just be fitpod, not fitpod_command.  I assume mlpod.cpp/h is an auxiliary supporting class.  That name is fine but it could also just be pod.cpp/h

The name of the potential is POD. Since "pair_styled pod" command does not conflict with other pair styles, it will be fine to use "pair_styled pod" command and src/ML-POD/pair_pod.cpp/h files. As mlpod.cpp/h is an auxiliary supporting class, it will be fine to keep their name to avoid potential conflict in the future.

athomps:
I don't feel quite as converged as @sjplimp.  I like the command names `pair_style mlpod` and `fitpod`.  I like the filenames `*mlpod.h` and 'fitpod_command.h'.  There is precedence for this with `kim_command.h` and `mdi_command.h`. I like moving examples/mlpod to examples/PACKAGES/mlpod

athomps:
I think the two filenames `pod.txt` and ` data.txt` used in `fitpod_command.rst` and in `in.fitpod` are too generic and need a bit more thought.  Probably they should both refer to `Ta` in some way. I see that `pod.txt` content is identical to `Ta_param.mlpod`. I have not run the example, so what are the names of the output files generated by the `fitpod` command?  This should be documented in README and `fitpod_command.rst`.  Should they not be user-specified, perhaps using a user-specified basename for all output files e.g. base name = `Ta2` yields output file `Ta2_coeff.mlpod` identical to `Ta_coeff.mlpod`?  All of this needs to be straightened out in the code, then demonstrated in a suitable example, with reference output files, timestamped, similar to the reference log files, and finally documented in `fitpod_command.rst`.   By providing reference output, you can get rid of `original` directory, which contains unnecessary duplicate input files that are not quite the same i.e. `pod.txt` is not the same in Ta and Ta/original.

exapde:
> I think the two filenames `pod.txt` and ` data.txt` used in `fitpod_command.rst` and in `in.fitpod` are too generic and need a bit more thought. Probably they should both refer to `Ta` in some way. I see that `pod.txt` content is identical to `Ta_param.mlpod`. I have not run the example, so what are the names of the output files generated by the `fitpod` command? This should be documented in README and `fitpod_command.rst`. Should they not be user-specified, perhaps using a user-specified basename for all output files e.g. base name = `Ta2` yields output file `Ta2_coeff.mlpod` identical to `Ta_coeff.mlpod`? All of this needs to be straightened out in the code, then demonstrated in a suitable example, with reference output files, timestamped, similar to the reference log files, and finally documented in `fitpod_command.rst`. By providing reference output, you can get rid of `original` directory, which contains unnecessary duplicate input files that are not quite the same i.e. `pod.txt` is not the same in Ta and Ta/original.

Hi @athomps, many thanks for your helpful comments. 

At present, the output files generated by the fitpod command are coefficients.txt, training_errors.txt, training_analysis.txt, test_errors.txt, and test_analysis.txt. The names of the output files are generic. We will allow user-specified names using  a base name. For instance, if basename_for_output_files = `Ta', then  the fitpod command will generate Ta_coeff.mlpod, Ta_training_errors.mlpod, Ta_training_analysis.mlpod, Ta_test_errors.mlpod, and Ta_test_analysis.mlpod.  

How does that sound to you?

exapde:
> I don't feel quite as converged as @sjplimp. I like the command names `pair_style mlpod` and `fitpod`. I like the filenames `*mlpod.h` and 'fitpod_command.h'. There is precedence for this with `kim_command.h` and `mdi_command.h`. I like moving examples/mlpod to examples/PACKAGES/mlpod

I prefer `pair_style pod` and `fitpod` since fitpod and pod sound good together  in the sense that the first is the fitting of the second (similar to snap and fitsnap). I think  `mlpod` and `fitpod` do not sound they are related. (`mlpod` and `fitmlpod` are somewhat related, but `fitmlpod` is not a nice command name.) 

I like `mlpod.h/cpp`, 'fitpod_command.h/cpp', and moving examples/mlpod to examples/PACKAGES/mlpod. I prefer `fitpod_command.h/cpp' because we can immediately see fitpod is a Command Style. Similarly, we can see `pair_pod.h/cpp' is a Pair Style right away without looking at the contents of the files. The file name 'fitpod.h/cpp' is not specific enough, so we don't know if it is a style class or an auxiliary supporting class unless we look into their implementation. 

These are just my two cents. I am fine with whatever we converge on.

sjplimp:
@athomps I based by suggestion about fitpod b/c of what @akohlmey said.  Which was that the name of the command (in an input script) is fitpod, but the filenames were fitpod_command.cpp/h.  I think those 2 need to be consistent like they are for all other commands.  Now that I think about it, the only 2 commands where they are not consistent is the kim and mdi LAMMPS commands. Which are encoded in filenames kim_command.h and mdi_command.h.  But that is because the filenames kim.h and mdi.h already existed in the KIM and MDI libraries themselves.  In the case of the fitpod command, is there a fitpod.h filenname (perhaps in a POD lib) that is already taken?  If so, using fitpod_command.h within LAMMPS would be the right thing to do, but if not, I am suggesting the code for the fitpod command in LAMMPS should be in fitpod.cpp/h files.

akohlmey:
> By providing reference output, you can get rid of `original` directory, which contains unnecessary duplicate input files that are not quite the same i.e. `pod.txt` is not the same in Ta and Ta/original.

@athomps There is one problem with the current implementation of the "fitpod": it randomizes the list of files used for fitting and uses the walltime as random seed. Thus the output changes with every invocation.

athomps:
> > I don't feel quite as converged as @sjplimp. I like the command names `pair_style mlpod` and `fitpod`. I like the filenames `*mlpod.h` and 'fitpod_command.h'. There is precedence for this with `kim_command.h` and `mdi_command.h`. I like moving examples/mlpod to examples/PACKAGES/mlpod
> 
> I prefer `pair_style pod` and `fitpod` since fitpod and pod sound good together in the sense that the first is the fitting of the second (similar to snap and fitsnap). I think `mlpod` and `fitpod` do not sound they are related. (`mlpod` and `fitmlpod` are somewhat related, but `fitmlpod` is not a nice command name.)
> 
> I like `mlpod.h/cpp`, 'fitpod_command.h/cpp', and moving examples/mlpod to examples/PACKAGES/mlpod. I prefer `fitpod_command.h/cpp' because we can immediately see fitpod is a Command Style. Similarly, we can see `pair_pod.h/cpp' is a Pair Style right away without looking at the contents of the files. The file name 'fitpod.h/cpp' is not specific enough, so we don't know if it is a style class or an auxiliary supporting class unless we look into their implementation.
> 
> These are just my two cents. I am fine with whatever we converge on.

I like these arguments and I agree with all these proposals.  @sjplimp is against `fitpod_command.cpp/h`, and explained the specific reason for `kim_command.cpp/h` and `mdi_command.cpp/h`. The situation here is a little different, but the same reasoning applies, so I am for `fitpod_command.cpp/h`

athomps:
> > By providing reference output, you can get rid of `original` directory, which contains unnecessary duplicate input files that are not quite the same i.e. `pod.txt` is not the same in Ta and Ta/original.
> 
> @athomps There is one problem with the current implementation of the "fitpod": it randomizes the list of files used for fitting and uses the walltime as random seed. Thus the output changes with every invocation.

Irreproducible randomization by default defeats a lot of standard software testing measures and should be turned off.

stanmoore1:
>Randomization by default defeats a lot of standard software testing measures and should be turned off.

100% agree with this.

sjplimp:
>> I like these arguments and I agree with all these proposals. @sjplimp is against fitpod_command.cpp/h, and explained the specific reason for kim_command.cpp/h and mdi_command.cpp/h. The situation here is a little different, but the same reasoning applies, so I am for fitpod_command.cpp/h

Sorry @athomps, can you clarify what you mean by "same reasoning applies" ?  Are you saying there is in fact another fitpod.h file somewhere,therefore the LAMMPS package should not use fitpod.h ?  (which is why we couldn't use mdi.h and kim.h)  Otherwise (to my way of thinking), there is not a good reason to call it fitpod_command.h in LAMMPS.  Or maybe you mean something else ... either way, I'm fine with what you decide

athomps:
> Sorry @athomps, can you clarify what you mean by "same reasoning applies" ? 

@sjplimp The reasoning was explained by @exapde:

>  I prefer fitpod_command.h/cpp' because we can immediately see fitpod is a Command Style. Similarly, we can see pair_pod.h/cpp' is a Pair Style right away without looking at the contents of the files. The file name 'fitpod.h/cpp' is not specific enough, so we don't know if it is a style class or an auxiliary supporting class unless we look into their implementation.

sjplimp:
ok - thanks for explaning

athomps:
Okay. It sounds like once those changes to are made to `fitpod` command (input and output filenames, and non-random execution), this will be good to go.

exapde:
> > @athomps There is one problem with the current implementation of the "fitpod": it randomizes the list of files used for fitting and uses the walltime as random seed. Thus the output changes with every invocation.
> 
> Irreproducible randomization by default defeats a lot of standard software testing measures and should be turned off.

Hi  @athomps and @akohlmey, I would like to get your suggestion on randomization of the list of training input configurations in fitpod. At present, this randomization feature is only enabled when (1) fraction of the training dataset is set to a value less than 1 by users and (2) the optional flag randomize_training_data_set is set to 1 by users. The normal behavior of the training process is all deterministic unless users want to randomize the list of training configurations on purpose. I think the feature does not affect software testing measures. However, I will disable the randomization feature if you think it is not useful. Please let me know what you think.

athomps:
> 

As long as the examples that are provided (which will get used for regression testing) are deterministic, there is no issue.

akohlmey:
> As long as the examples that are provided (which will get used for regression testing) are deterministic, there is no issue.

I think I have identified the source of randomness when using all files: the order of files in the directory can change between checkouts on different machines and thus leading to slightly different fitting results.
I've prepared a PR to the forked repo where that list of files is now sorted alphabetically and thus should be (more) consistent across platforms.

exapde:
> I think I have identified the source of randomness when using all files: the order of files in the directory can change between checkouts on different machines and thus leading to slightly different fitting results.
> I've prepared a PR to the forked repo where that list of files is now sorted alphabetically and thus should be (more) consistent across platforms.

Hi @akohlmey, many thanks for making several updates to ML-POD and sorting the list of files. 

I have a question. Below is the output of podfit command for the Ta example on my system. 
 
----------------------------------------------------------------------------------------
  File    | # configs | # atoms  | MAE energy | RMSE energy | MAE force | RMSE force |
----------------------------------------------------------------------------------------
Displaced_A15.xyz     9        576       0.011326     0.011373     0.113223     0.141496  
Displaced_BCC.xyz     9        486       0.012158     0.013983     0.240350     0.311834  
Displaced_FCC.xyz     9        432       0.001459     0.001605     0.082874     0.104038  
Elastic_BCC.xyz       100      200       0.004442     0.004774     0.000010     0.000013  
Elastic_FCC.xyz       100      400       0.002855     0.002914     0.000147     0.000208  
GSF_110.xyz           22       528       0.005794     0.006857     0.047304     0.097750  
GSF_112.xyz           22       660       0.010592     0.011566     0.123316     0.191078  
Liquid.xyz            3        300       0.013230     0.015407     0.567180     0.757188  
Surface.xyz           7        236       0.025392     0.037556     0.096037     0.295185  
Volume_A15.xyz        30       240       0.039262     0.048655     0.000000     0.000000  
Volume_BCC.xyz        21       42        0.049805     0.067608     0.000000     0.000000  
Volume_FCC.xyz        31       124       0.030071     0.041758     0.000000     0.000000  
----------------------------------------------------------------------------------------
All files             363      4224      0.012915     0.025804     0.122413     0.259832 
----------------------------------------------------------------------------------------

Changing the order of the files means swapping the corresponding rows of the above table, but the numbers should stay the same. Furthermore, the last row should be the same across all platforms regardless of the order of the files. If the output table does not behave in that way, there may be potential bugs in the code.

Can you please let me see the output of the podfit command on different platforms? Thanks so much again for your help which I am grateful to.

akohlmey:
> Can you please let me see the output of the podfit command on different platforms? Thanks so much again for your help which I am grateful to.

I am currently working on a bunch of other issues.

I get consistent screen outputs, but not the detailed files where output is printed with more digits.
BTW: outputting double precision floating point numbers with more than 15 digits is outputting some garbage digits since those numbers only have about 15 1/2 digits precision.

However, in the coefficients file - even after sorting - there is consistency only for up to 8 digits precision.

exapde:
> I am currently working on a bunch of other issues.

Totally understand.

> I get consistent screen outputs, but not the detailed files where output is printed with more digits. BTW: outputting double precision floating point numbers with more than 15 digits is outputting some garbage digits since those numbers only have about 15 1/2 digits precision.
> 

I agree. 

> However, in the coefficients file - even after sorting - there is consistency only for up to 8 digits precision.

It makes sense to have consistently up to 8 digits precision. That is an expected behavior due to round-off errors. I feel better now!.

akohlmey:
> It makes sense to have consistently up to 8 digits precision. That is an expected behavior due to round-off errors. I feel better now!.

No, it doesn't. There is a difference between precision of the calculation and accuracy of the result.

If I run the same operations in the same order, then the precision of the output should be reproduced to the full precision of the calculation. Even if the input data has low accuracy, it has always the *same* error and thus the resulting error should also always be the *same*. There has to be some source of randomness or inconsistency that enters into the calculation that I have not yet been able to identify.

In other words, there may be some small differences between a serial and a parallel calculation, but the same calculation with the same LAPACK library on the same machine compiled with the same compiler using the same compiler optimization should produce identical output.

exapde:
> In other words, there may be some small differences between a serial and a parallel calculation, but the same calculation with the same LAPACK library on the same machine compiled with the same compiler using the same compiler optimization should produce identical output.

Hi @akohlmey! When running podfit in serial on my machine, I consistently got the same coefficient file for several different runs. When running podfit in parallel on my machine, I also got the same coefficient file for several different runs.  However, the coefficient file of the serial runs is different from the coefficient file of the parallel runs up to 8 digits. The fitting errors are the same for both the serial and parallel runs though. It seems that I get the expected behaviors on my machine. 

Now I understand you made Ta_coeff.pod and Ta_param.pod for the unit test to ensure that the output of MD is consistently the same. The coefficient file generated by the fitpod command is not used for the unit test since it can change.

exapde:
Hi @akohlmey, @athomps, @sjplimp, we made changes to the fitpod command (input and output filenames), update documentation, and README files, as suggested. Please let us know if there are other changes to be made. 

We appreciate your time reviewing the code and suggesting comments. We are thankful to @akohlmey for making so many updates to improve ML-POD.

akohlmey:
> However, the coefficient file of the serial runs is different from the coefficient file of the parallel runs up to 8 digits.

If the fit is only "good" to 8 digits precision, the output to the coefficients file should be rounded accordingly. The additional digits would be just random, so this rounding would produce more consistent results across platforms.

athomps:
`[athomps@s1062580 Ta]$ diff Ta_coeff.pod Ta_coefficients.pod | wc -l
      64
`
It would be nice if this returned 0

athomps:
Reference log files need to be updated to reflect modified Ta_coeff.pod file

athomps (approved):
> Reference log files need to be updated to reflect modified Ta_coeff.pod file

Once that is done, I am satisfied. Nice work!

exapde:
Hi @akohlmey! I used a Linux machine to compile the LAMMPS source code and ran the Ta examples on Linux.  I updated the Ta examples. I also made the computation of the pod coefficients more robust, so that it is less prone to round-off errors.  

The updated code does not pass the unit test check. Could you please let me know how to rebuild the unit test?  

As always, many thanks for your kind help.

akohlmey:
> Could you please let me know how to rebuild the unit test?

I've just done this myself since there were some other small changes that could/should be done.
The principle is simple: You can run the pod pair style test selectively in the build folder with
```
ctest -R ManybodyPairStyle:pod
```
To update a special flag must be set via an environment variable (since there is no portable way to tell ctest to pass flags to the individual test program) like this:
```
TEST_ARGS=-u ctest -R ManybodyPairStyle:pod
```

More information on testing is here: https://docs.lammps.org/Build_development.html#code-coverage-and-unit-testing-cmake-only and here: https://docs.lammps.org/Developer_unittest.html

exapde:
> Once that is done, I am satisfied. Nice work!

@athomps Thanks very much for the time you took to review and give us helpful comments.

akohlmey:
@exapde @rohskopf are there any more pending/planned changes for this pull request?

exapde:
> @exapde @rohskopf are there any more pending/planned changes for this pull request?

Hi @akohlmey, there are no more changes from my side for this pull request. Thanks so much for your kind help that allows  us to complete this pull request.

PR #3454, opened 2022-09-21
Fixing typo in pair_gayberne.rst

evoyiatzis:
Fixing typo in the documentation

**Licensing**


evoyiatzis:
also the A1 and A2  right after the first set of equations should be bold and the numbers  subscripts

akohlmey:
> also the A1 and A2 right after the first set of equations should be bold and the numbers subscripts

I have added this PR and the suggested format change (and a couple more) into PR #3450

PR #3456, opened 2022-09-23
Update "make install" for CMake to use pip for building/installing the LAMMPS python module

### LINKED ISSUE #3429: [BUG] Wrong Python module install path on Fedora36
**Summary**

When building lammps on Fedora36 with `-D BUILD_SHARED_LIBS=on -D LAMMPS_EXCEPTIONS=on -D PKG_PYTHON=on` lammps Python module installs into `.local/local/lib` instead of `.local/lib`

Fedora linux version 36
lammps release branch commit: 883a25bbd847f66d663dbb61381bdbdb9922c574

**Expected Behavior**

Python moudule installs into `.local/lib/pythonX.X/site-packages/lammps` and it works on `import lammps`

**Actual Behavior**

For some reason it installs into `.local/local/lib/pythonX.X/site-packages/lammps` and Python can't find the module.

```shell
running install
/usr/lib/python3.10/site-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
  warnings.warn(
running install_lib
creating /home/denis/.local/local
creating /home/denis/.local/local/lib
creating /home/denis/.local/local/lib/python3.10
creating /home/denis/.local/local/lib/python3.10/site-packages
creating /home/denis/.local/local/lib/python3.10/site-packages/lammps
...
```

Though, if I use `make install-python` command it installs properly with `pip` into `.local/lib/pythonX.X/site-packages/lammps`

akohlmey:
**Summary**

This updates the "make install" target for CMake to modernize the LAMMPS python module installation and attempts to address the incorrect path used.

**Related Issue(s)**

Fixes #3429 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

This is likely not portable beyond Linux, but at least a start and something worth testing and - if needed - improving.
The wrong path with the extra "local" happens when using --prefix, so this replaces it with --target and an explicit path.

**Post Submission Checklist**


junghans (commented):
I am unsure of the new method has `DESTDIR` support, which is what all the lammps package use.

@rbberger or I will need to test that first.

cmake/CMakeLists.txt
```diff
@@ -849,8 +851,11 @@ if(BUILD_SHARED_LIBS OR PKG_PYTHON)
     find_package(Python COMPONENTS Interpreter)
   endif()
   if(Python_EXECUTABLE)
-    file(MAKE_DIRECTORY ${CMAKE_BINARY_DIR}/python)
-    install(CODE "execute_process(COMMAND ${Python_EXECUTABLE} setup.py build -b ${CMAKE_BINARY_DIR}/python install --prefix=${CMAKE_INSTALL_PREFIX} --root=\$ENV{DESTDIR}/ WORKING_DIRECTORY ${LAMMPS_PYTHON_DIR})")
+    file(MAKE_DIRECTORY ${CMAKE_BINARY_DIR}/python/lib)
+    file(MAKE_DIRECTORY ${CMAKE_BINARY_DIR}/python/src)
+    file(COPY ${LAMMPS_SOURCE_DIR}/version.h  DESTINATION ${CMAKE_BINARY_DIR}/python/src)
+    file(COPY ${LAMMPS_PYTHON_DIR}/README ${LAMMPS_PYTHON_DIR}/pyproject.toml ${LAMMPS_PYTHON_DIR}/setup.py ${LAMMPS_PYTHON_DIR}/lammps  DESTINATION ${CMAKE_BINARY_DIR}/python/lib)
+    install(CODE "execute_process(COMMAND ${Python_EXECUTABLE} -m pip install ${CMAKE_BINARY_DIR}/python/lib --target=${CMAKE_INSTALL_PREFIX}/lib/python${Python_VERSION_MAJOR}.${Python_VERSION_MINOR}/site-packages)")
```
junghans: It would be better to not hard coded the target path as some distros install mpi-enabled python modules into specific locations.

akohlmey: Feel free to propose/implement some alternate way to set the destination and that also addresses the problems with `--prefix`. ...or go back to `--prefix` and forget about #3429. This is why it is flagged as a draft. My main intent was to try transfer the method of building/installing the LAMMPS python module from `make install-python` to `make install`. This is simpler, since none of the trickery to locate, copy, and bundle the pre-compiled shared library with the wheel package is required. I thought this could help to move this forward before executing `setup.py` will fail.

rbberger: Probably the only "clean" and general way of solving this is either using `Python_SITELIB` and stripping the common prefix or writing your own `execute_process` that runs a short Python snippet to compute the relative path for you.

Alternatively use what Axel is suggesting, but make the location of `--target` an advanced CMake cache variable that defaults to what is done now. Something like `CMAKE_PIP_INSTALL_TARGET`.

akohlmey: BTW: I think at this point we should perhaps stop worrying about compatibility with older CMake versions for certain targets and features and thus avoid having to program all kinds of backward compatibility hacks and instead just have gracefully degraded performance. For example in this case, there could be simply a warning message that to have the python module installed, CMake 3.12 or newer is required (which has the better way to query Python support).

junghans: `LAMMPS_PIP_INSTALL_TARGET`, but yeah that works.

And pip has some `DESTDIR` support these days through `--root $DESTDIR --installed-root /`, see https://github.com/pypa/pip/issues/1351, but to not make our logic in CMAke overly complicated, we could just add something like `LAMMPS_PIP_EXTRA_INSTALL_OPTIONS` at the end of the command and then an expert user can inject whatever extra option incl. an overwrite for the default `--target`.

junghans: And to Axel's other point, I am all for dropping the support for CMake<3.12 for some of these more expert feature.

stanmoore1 (approved):

cmake/CMakeLists.txt
akohlmey: After some experiments, it appears the unconditional use of the --root flag seems to trigger the unexpected path for installing the Python module. So the hard-coded path can go away and with the latest modification it seems to be installing the LAMMPS Python module as expected. If one wants to install the python module for local use only, and not all/most of LAMMPS, people need to use "make install-python" as that also includes the shared library with the wheel package, while 'make install' depends on the installed shared library and, if needed, setting LD_LIBRARY_PATH accordingly.

PR #3458, opened 2022-09-23
MLIAP Unified Interface

Boogie3D:
**Summary**

This PR introduces new features and improvements to the ML-IAP Python component, most prominently including a new "Unified" pair style. It also fixes problems raised in #2934.

**Related Issue(s)**

Fixes an [issue](https://github.com/lammps/lammps/pull/2934#issuecomment-918628329) where ML-IAP loads `lammps` library separately from `lammps-python`.

A NumPy example for ML-IAP is also included.

**Author(s)**

Steven Anaya, LANL | anayastevenr@gmail.com
Nicholas Lubbers, LANL | nlubbers@lanl.gov

**Licensing**

**Backward Compatibility**

No known backward compatibility problems are introduced by this PR.

**Implementation Notes**

The main improvement is the addition of the ML-IAP "Unified" pair style, which allows Python libraries to build ML-IAP components for LAMMPS, taking control of both the descriptor and model protocols. This feature allows, among other possibilities, the ability to use Deep Neural Network models which are end-to-end (taking positions as input, rather than features) and those that view the environment of each atom recursively. A pedagogical example showing how Lennard-Jones can be implemented efficiently in Python is [provided](https://github.com/Boogie3D/lammps/blob/mliappy_unified/python/lammps/mliap/mliap_unified_lj.py). There is a corresponding [example](https://github.com/Boogie3D/lammps/blob/mliappy_unified/examples/mliap/in.mliap.unified.lj.Ar), which reproduces the classic [lj/melt example](https://github.com/Boogie3D/lammps/tree/mliappy_unified/examples/melt) using this programming technique. There are also several non-trivial examples which give a notion of more practical usage using the HIP-NN Deep Neural Network via the `hippynn` python [library](https://github.com/lanl/hippynn).

Secondly, this PR introduces improvements to the ML-IAP Python coupling:
1) It includes a version of the "reverse" Python-driving-LAMMPS [code](https://github.com/Boogie3D/lammps/blob/mliappy_unified/examples/mliap/mliap_unified_lj_Ar.py) which only depends on NumPy and is thus suitable for automated testing. Also note [examples/mliap/mliap_numpy_Ta06.py](https://github.com/Boogie3D/lammps/blob/mliappy_unified/examples/mliap/mliap_numpy_Ta06A.py) which tests the reverse python->lammps coupling for mliap model python.
2) Python-driven codes will extract the embedded ML-IAP modules from LAMMPS directly using the loaded `lammps` library object, rather than seeking out the `lammps` library. This prevents any possibility that `lammps.lammps()` objects and `lammps.mliap` module.

Lastly, this PR introduces a few extra features for the ML-IAP pair style and data structure. The ML-IAP pair style can be toggled to use neighbors of ghost atoms, which is required for some kinds of potentials.

It also introduces the following members to the `MLIAPData` class:
* `double **f` - pointer to `atom->f`, updated on each call to `generate_neighdata()`
* `int ntotal` - stores total number of owned and ghost atoms on this proc
* `int natomneigh` - current number of atoms and ghosts in atom neighbor arrays
* `int npairs` - number of `ij` neighbor pairs
* `int *pair_i` - array storing index of each i atom for each ij pair
* `int *elems` - array storing element of each atom in or not in the neighborlist

The following member is added to the `MLIAPDescriptor` class:
* `double ** cutghost` - array storing cutoff for each ghost pair

We used Cython similarly to `mliap_model_python`.

**Post Submission Checklist**

https://www.lanl.gov/org/ddste/aldsc/hpc/_assets/docs/posters/22HPC-showcase-poster-anaya.pdf


doc/src/pair_mliap.rst
```diff
@@ -167,6 +173,34 @@ to specify the path for these *model* and *descriptor* files.
    larger systems since their size depends on the total number of
    neighbors per MPI process.
 
+The *unified* keyword is followed by an argument specifying the
+filename containing the serialized unified Python object and the "ghostneigh" toggle
+(0/1) to disable/enable the construction of neighbors lists including
+neighbors of ghost atoms. If the filename ends in '.pt', or '.pth', it will be loaded
+using pytorch; otherwise, it will be loaded as a pickle file.
+If ghostneigh is enabled, it is recommended to set :doc:`comm_modify <comm_modify>`
```
akohlmey, line 185: FYI: when I enable the "ghostneigh" flag in the `in.mliap.unified.lj.Ar` example, I get a segmentation fault. Is this expected behavior? Can this be checked for and an error triggered?
Increasing the communication cutoff as suggested below did not make a difference.

examples/mliap/README
```diff
@@ -109,4 +109,70 @@ in.mliap.nn.cu
...
+but using the ``unified`` keyword and
+the Python coupling to PyTorch.
+
+This example can be run in two different ways:
+
+1: Running a LAMMPS executable: in.mliap.unified.lj.Ar
+
+You can run the example as follows
+
+`lmp -in in.mliap.unified.lj.Ar -echo both`
```
akohlmey, line 124: FYI the input now includes the python call to create the pickle via the LAMMPS shell command.

python/lammps/mliap/__init__.py
```diff
@@ -25,8 +26,10 @@
     raise OSError("Unable to locate python shared library") from e
 
 if not pylib.Py_IsInitialized():
-    raise RuntimeError("This interpreter is not compatible with python-based mliap for LAMMPS.")
+    warnings.warn("This interpreter is not compatible with python-based MLIAP for LAMMPS. "
```
akohlmey, line 29: Why this change to a warning instead of an error?

akohlmey (commented):
@Boogie3D @lubbersnick Thanks for your submission. This looks quite nicely done. I've made some - mostly cosmetic - minor changes to improve consistency with the rest of LAMMPS and some (yet undocumented) conventions.
There are a couple notes/questions added in different locations, but neither is a big concern.

doc/src/pair_mliap.rst
Boogie3D: I don't believe this is expected behavior. The error likely comes from neighbor list memory allocation as this segfault occurs while calculating Per MPI Rank memory allocation (see backtrace below). If I reduce the problem size from 4000 atoms to 256 I get a neighbor list overflow instead of a segfault. This error does not happen with the hippynn Al or InP examples.
```
Core was generated by `lmp -i in.mliap.unified.lj.Ar'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x0000149cd25883d8 in int_mallinfo () from /lib64/libc.so.6
Missing separate debuginfos, use: yum debuginfo-install glibc-2.28-151.el8.x86_64
(gdb) bt
#0  0x0000149cd25883d8 in int_mallinfo () from /lib64/libc.so.6
#1  0x0000149cd258e09c in mallinfo () from /lib64/libc.so.6
#2  0x0000149cd32247c6 in LAMMPS_NS::Info::get_memory_info (this=this@entry=0x7ffc14762a90, meminfo=meminfo@entry=0x7ffc14762a20)
    at /projects/ml4chem/sanaya/lammps/src/info.cpp:1300
#3  0x0000149cd3310b6b in LAMMPS_NS::Output::memory_usage (this=0x744a30) at /projects/ml4chem/sanaya/lammps/src/output.cpp:1018
#4  0x0000149cd33110d1 in LAMMPS_NS::Output::setup (this=0x744a30, memflag=memflag@entry=1) at /projects/ml4chem/sanaya/lammps/src/output.cpp:299
#5  0x0000149cd33f3ae2 in LAMMPS_NS::Verlet::setup (this=0x73ca20, flag=1) at /projects/ml4chem/sanaya/lammps/src/verlet.cpp:160
#6  0x0000149cd33a0c6f in LAMMPS_NS::Run::command (this=0x91bbc0, narg=1, arg=0x749180) at /projects/ml4chem/sanaya/lammps/src/run.cpp:171
#7  0x0000149cd323d1b8 in LAMMPS_NS::Input::execute_command (this=0x72d6f0) at /projects/ml4chem/sanaya/lammps/src/input.cpp:847
#8  0x0000149cd323d58d in LAMMPS_NS::Input::file (this=0x72d6f0) at /projects/ml4chem/sanaya/lammps/src/input.cpp:302
#9  0x0000000000400e85 in main (argc=<optimized out>, argv=<optimized out>) at /projects/ml4chem/sanaya/lammps/src/main.cpp:76
```

python/lammps/mliap/__init__.py
Boogie3D: The idea here was that users could run LAMMPS with a pickled model even without having mliappy properly installed. The exception was getting in the way of this so it was changed to a warning and if mliappy isn't installed then `activate_mliappy`, `load_model`, and `load_unified` are not imported into the `lammps.mliap` module (they could still potentially be imported from the `loader` submodule but using them would result in undefined behavior).

doc/src/pair_mliap.rst
akohlmey: The gdb trace is bogus. Something has written to an allocated storage block beyond the allocated size and thus messed up the internal malloc data structures.

I have run the code with valgrind and identified the location of the problem at this location:
https://github.com/lammps/lammps/pull/3458/files#diff-cb6b51f0e9ebcb4ecda7e8a34b450fb05f24b7de8a8bfbac614cbfd21af03b6dR256

The index `i` overflows the dimension of the `data->eatoms` array which is of size `data->nlistatoms`.
I am implementing a hack that avoids the crash, but will likely lead to incorrect per-atom energy data.
I would expect that `data->eatoms` would need to be allocated (and cleared) so it includes ghost atoms and then a reverse communication would be needed to sum the contributions to ghost atoms to their principal owned atoms.

akohlmey: Also found and fixed a memory leak, an initialization issue and noticed that destructors of some polymorph classes were not make virtutal and general our conventions for using virtual/override were not fully followed.

akohlmey: @Boogie3D I think I have sorted this out for the LJ example. I could not test the hippynn examples, since I did not see any of the .pt files.
I also added a test that creates two LAMMPS instances with mliap.unified.lj and lj/cut pair styles, respectively and check some results after a couple of MD steps.

examples/mliap/README
akohlmey: I've refactored this some more. Now it is using the LAMMPS python command so no external file is needed.

python/lammps/mliap/__init__.py
akohlmey: Not sure if that is a good idea. I personally prefer a clean error and forcing people to do a proper installation.
But this is "your" code that you will have to support and thus the choice is yours.

doc/src/pair_mliap.rst
Boogie3D: > @Boogie3D I think I have sorted this out for the LJ example. I could not test the hippynn examples, since I did not see any of the .pt files. I also added a test that creates two LAMMPS instances with mliap.unified.lj and lj/cut pair styles, respectively and check some results after a couple of MD steps.

The scripts to create the hippynn .pt files are in the hippynn [repo](https://github.com/Boogie3D/hippynn/tree/lammps_mliap/examples/lammps) but unfortunately require HIPNN training to be performed first; the .pt files themselves are not uploaded anywhere.

Thank you for your various improvements and for solving the problem with memory allocation and LJ so quickly. I am beginning graduate school soon and am not currently working at LANL so I may not have had the time to do so myself.

python/lammps/mliap/__init__.py
Boogie3D: We can speak with @lubbersnick about it as it was his rationale. I agree that the implementation here could be cleaner; it should probably raise an error if someone does try to invoke one of the `mliappy` functions without the right installation rather than segfaulting, etc.

doc/src/pair_mliap.rst
akohlmey: > the .pt files themselves are not uploaded anywhere

Can you send me a copy of those files via email? I would very much like to make certain that my changes did not break anything.

akohlmey:
@lubbersnick How do propose to continue with this PR?
I cannot confirm that my changes did not break anything (but they prevent out-of-bounds memory accesses, so if they change results, probably the cause needs to be addressed as well) because the necessary software is not yet publicly available.

python/lammps/mliap/__init__.py
lubbersnick: Hey, we can make it error if someone tries to active the MLIAP module. But there is no need to make it error to prevent people from accessing -other- things in lammps.mliap. Like @Boogie3D said the other objects can be used to export models for lammps even when the 'reverse' python driving lammps mode doesn't work (which is on most macs right now). From what I know fixing this can require installing python from source to adjust the build options, which is a few steps beyond what we should expect from a typical user.

akohlmey: Honestly, I don't care much, I mostly was curious. Also, I am not going to make any more comments on the Python/Mac/Conda/etc situation in public.

What worries me much more is that it is not possible for me to run most of the new example inputs currently and thus verify my modifications to avoid memory corruption.

athomps:
@lubbersnick I think @akohlmey has a point about the unrunnable HIPNN examples.  Either we should make them runnable or take them out.

lubbersnick:
@akohlmey @athomps I would like an eventual future where we have an example in the lammps repository because it shows a much more clear reason and use case compared to reimplementing LJ. In some sense the difficulty with including networks in the repository is the number of things to validate in order to prove a network is good/useful for an actual application. Although I don't think it is possible to release a practically useful network from the very limited InP dataset, we can maybe make a standardized one for aluminum. 

I am slightly confused at the comment that the code is not public - it is, and will very soon [be merged](https://github.com/lanl/hippynn/pull/10) into the main hippynn repository.  Building your own network for aluminum is our basic example in hippynn. It is probably not worth the effort to do so for the sake of checking the lammps codebase; it requires going to get the training dataset and so on. I will email you a network so you can before/after test your changes if you like.  However I am pretty sure that the change to `update_pair_energies` shouldn't affect the hippynn library because it does not use that function, as can be verified [here](https://github.com/Boogie3D/hippynn/blob/26b75ca5ebbb3952c32176597a70bc7fdfb9b5ed/hippynn/interfaces/lammps_interface/mliap_interface.py) . 

However, it was good to identify this difficulty of mixing update_pair_energies with ghostneigh. We should probably be clear that ghostneigh is connected to the semantics of the model -- models can give different answers with it on or of and so ghostneigh needs to be set as appropriate for that model. So the fact that ghostneigh gives a segfault for LJ does not surprise me. It is better that we have this check instead of corrupting memory though, and very sharp of you to find it.  A good 'future work' thing for us would be to set up the mliap interface class to store the ghostneigh information as needed for that model so that the setting cannot be wrong.

akohlmey:
> @akohlmey @athomps I would like an eventual future where we have an example in the lammps repository because it shows a much more clear reason and use case compared to reimplementing LJ. In some sense the difficulty with including networks in the repository is the number of things to validate in order to prove a network is good/useful for an actual application. Although I don't think it is possible to release a practically useful network from the very limited InP dataset, we can maybe make a standardized one for aluminum.

The examples in the LAMMPS repository are *not* meant to be blueprints for excellent research. Their purpose is to a) demonstrate the syntax and structure of inputs for a given package and b) to provide a reference that can be checked against in case changes to the code base are made that may (or may not) affect a package.  In both cases, the input does not have to produce a high accuracy result. If you are concerned about people misusing the network, just make up a new element/compound and for example call it Bovinexcrementum (element symbol BS) and then there is no danger of somebody using it for real research and blaming bad results on you (or rather the main LAMMPS developers, since most people in first approximation blame us for everything that goes wrong, regardless of who actually authored the code).

> I am slightly confused at the comment that the code is not public - it is, and will very soon [be merged](https://github.com/lanl/hippynn/pull/10) into the main hippynn repository. Building your own network for aluminum is our basic 

I followed the instructions given in the pull request description (and in your private email) and in both cases it points to a repository that *at that moment* did *not* contain the necessary changes. You cannot expect people to know that this is not correct.

> function, as can be verified [here](https://github.com/Boogie3D/hippynn/blob/26b75ca5ebbb3952c32176597a70bc7fdfb9b5ed/hippynn/interfaces/lammps_interface/mliap_interface.py) .

> However, it was good to identify this difficulty of mixing update_pair_energies with ghostneigh. We should probably be clear that ghostneigh is connected to the semantics of the model -- models can give different answers with it on or of and so ghostneigh needs to be set as appropriate for that model.

Then a check whether this is required or not **should** be in the model and a proper error should be thrown if the wrong setting is used, or - even better - the C++ code should have a way to query the model for the correct setting and then there need not be a flag. Same as some manybody potentials already contain the pairwise cutoff and set it accordingly. The available documentation suggests (at least the way how it appears to me) that the flag can be chosen rather freely.

> So the fact that ghostneigh gives a segfault for LJ does not surprise me. It is better that we have this check instead of corrupting memory though, and very sharp of you to find it. A good 'future work' thing for us would be to set up the mliap interface class to store the ghostneigh information as needed for that model so that the setting cannot be wrong.

Exactly. For the LJ case, it causes a significant slowdown, for obvious reasons. I just enabled it, since there was nothing saying in the docs that it is not supported and can cause problems, and as part of my test procedure for new features, I try to check different flags added to commands and also run with valgrind and other tools to check for memory issues etc (where possible, the python part is next to impossible to debug that way though).

akohlmey:
> @lubbersnick I think @akohlmey has a point about the unrunnable HIPNN examples. Either we should make them runnable or take them out.

Since there seems to be no mention of hippynn in the manual itself and no documentation for how to obtain the code and the potential files (the only reference is `Requires the HIPNN code.` in the README), I second the removal of those examples. We cannot expect a user that looks at the examples to figure out where to find the code, how to install it, how to obtain/generate the potential file etc. Beyond that, the PR seems suitable to be merged.

lubbersnick:
Okay, let's just remove those examples for this PR. At some point I will write an example with more complete documentation and a dummy network.

Boogie3D:
> Okay, let's just remove those examples for this PR. At some point I will write an example with more complete documentation and a dummy network.

I second this for now.

athomps:
I agree that it is sufficient for now to provide only the LJ example (in two different forms).

athomps (approved):
I approve.

PR #3427, opened 2022-09-02
Doc fixes (round 3)

hammondkd:
**Summary**

Edits to documentation sources to insert math and other formatting, replacing parsed-literal with code-block LAMMPS where appropriate, and other minor edits. There should not be any substantive changes to the documentation. I also added "elif" and "else" to LAMMPSparser.py

These only got up through "fix balance," but given that the code clinic ends soon, I wanted to get these in before it ends.

**Related Issue(s)**

Contributes to #3371.

**Author(s)**

Karl Hammond, University of Missouri

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

This should change nothing at all in terms of input or program execution; only documentation.

**Post Submission Checklist**

- [ ] The feature or features in this pull request is complete
- [ ] Licensing information is complete
- [ ] Corresponding author information is complete
- [ ] The source code follows the LAMMPS formatting guidelines
- [ ] Suitable new documentation files and/or updates to the existing docs are included
- [ ] The added/updated documentation is integrated and tested with the documentation build system
- [ ] The feature has been verified to work with the conventional build system
- [ ] The feature has been verified to work with the CMake based build system
- [ ] Suitable tests have been added to the unittest tree.
- [ ] A package specific README file has been included or updated
- [ ] One or more example input decks are included

PR #3461, opened 2022-09-26
Correction of some LL features in the SPIN package

julient31:
**Summary**

This PR aims at correcting and improving the SPIN package. It contains:
- improvement of the compute/spin routines, with a better treatment of multiple fix precession/spin,
- correction of an error in the langevin/spin routines. This error was only visible when large-damping simulations were performed (which is a physically rare situation),
- addition of a new test problem corresponding to the "large damping situations" stated above,
- corrections of the documentation (the stochastic equation we use is referred to as sLL, and not sLLG).

**Related Issue(s)**

None

**Author(s)**

Julien Tranchida, 
julien.tranchida@cea.fr or julien.tranchida1@gmail.com

**Licensing**

**Backward Compatibility**

Those changes are limited to the SPIN package, and have to reason to break the backward compatibility. 

**Implementation Notes**

The modified code successfully passed all the test problems included in the SPIN package. They were compiled with traditional make and with cmake.

**Post Submission Checklist**

None.


julient31:
@akohlmey Hello Axel!
Some small corrections / improvements of the SPIN package in this PR. 

I hesitated to include the spin versus spin/atom modification that we discussed in this PR. It would need me to modify atom_vec_spin.h a tiny bit, and change all the examples. That's not a lot of work. 
That being said, I believe it may be cleaner to do it in a separate PR. What do you think?

Best,
Julien.

akohlmey:
> I hesitated to include the spin versus spin/atom modification that we discussed in this PR. It would need me to modify atom_vec_spin.h a tiny bit, and change all the examples. That's not a lot of work. That being said, I believe it may be cleaner to do it in a separate PR. What do you think?

Yes, this should and can wait. I first need to contact the EFF package author to get the corresponding change done without changing the SPIN package would only be half measures. Also, with the moderate existing changes, we have been able to keep backward compatibility which we would have to break. That is something to weigh against the benefits, which are mostly important for software maintenance and development and will pay dividends mostly in the future if there would be additional packages and atom styles with similar features. Given the somewhat less-than-mainstream nature of both the SPIN and the EFF package, that has not a high probability, so I am now more hesitant compared to my initial impulse.

What *could* be done now (but probably also best as separate pull requests) is the application of some new/modified conventions for error messages and the documentation. Please see  https://www.lammps.org/workshops/Aug22/project1/, https://www.lammps.org/workshops/Aug22/project4/ for some background and the resulting pull requests as examples (PR #3394, PR #3403, PR #3408, PR #3428 for error messages, and PR #3411, PR #3427 for the documentation)

A final step to consider would be to come up with some ways to implement your library of tests within the unittest framework.
It is quite easy to run calculations through the LAMMPS library interface and retrieve data for comparison. I have recently implemented a test comparing two different inputs with different pair styles that should still give the same results this way.
https://github.com/lammps/lammps/pull/3458/files#diff-c657bea437220a727e735458f510c453aeb70d58a771345149bcd1fb82898cd1
This demonstrates how recurring input file fragments can be stored and reused as strings and then how to make comparisons with accommodation for numerical errors.

athomps (approved):
I approve.

akohlmey:
@julient31 FYI, I have simplified the lookup of the precession fixes using newer APIs in the Modify class.

PR #3470, opened 2022-10-03
Further extending the Fortran interface

hammondkd:
**Summary**

Fortran interface implementations of ``get_mpi_comm``, ``extract_atom``, ``extract_compute``, ``extract_fix``, ``extract_variable``, ``gather_atoms``, ``gather_atoms_concat``, ``gather_atoms_subset``, ``scatter_atoms``, ``scatter_atoms_subset``, ``get_os_info``, ``config_has_mpi_support``, ``config_has_gzip_support``, ``config_has_png_support``, ``config_has_jpeg_support``, ``config_has_ffmpeg_support``, ``config_has_exceptions``, ``config_has_package``, ``config_package_count``, ``config_package_name``, ``installed_packages``, ``flush_buffers``, ``is_running``, ``has_error``, and ``get_last_error_message``. Fixed a missing warning in library.cpp (``lammps_gather_atoms_subset`` and ``lammps_scatter_atoms_subset`` require an atom map).

I also created ``lammps_extract_variable_datatype`` in library.cpp and modified the Python and Fortran interfaces to use it. Python, C, and Fortran library interfaces also now work with *vector* style variables and variables that return strings.

**Author**

Karl D. Hammond
hammondkd at missouri.edu
University of Missouri

**Licensing**

**Backward Compatibility**

The Python interface should maintain backward compatibility, but the *vartype* optional argument to ``extract_variable`` is now obsolete. In this implementation, the user can provide *vartype* to override what LAMMPS finds, but if it's absent or ``None``, LAMMPS will "guess" what the type is based on the return value from ``lammps_extract_variable_datatype``. The Fortran interface also uses this.

**Implementation Notes**

The use of ``lammps_extract_variable_datatype`` required two new constants in the Python interface and four in library.h and lammps.f90; new ones are ``LMP_VAR_VECTOR`` and ``LMP_VAR_STRING``. Documentation for this `enum` (``LMP_VAR_CONST``) is included.

Rather than creating new user-space global constants, I implemented ``LMP_STYLE_GLOBAL`` and friends as a derived type bound to the ``lammps`` derived type, so the user will type something like ``lmp%style%global`` to access that constant, assuming ``lmp`` is the name of the LAMMPS instance being used (and since Fortran is case-insensitive, they could just as easily use ``LMP%STYLE%GLOBAL``).

Return values from ``extract_atom``, ``extract_compute``, ``extract_fix``, and ``extract_variable`` are actually (private) derived types; they either bind user-space pointers to LAMMPS data or allocate and/or assign user-space arrays and scalars to copies, depending on how LAMMPS stores those data. Basically, if LAMMPS allocates space for it in the library interface, the Fortran interface makes copies, deallocates the C pointers, and assigns the data to the user's array. If the C library assigns pointers to actual LAMMPS data, so does the Fortran interface. Warnings about modifying those data, if applicable, are given in the documentation.

``extract_fix`` has *nrows* and *ncols* values for global data. The indices are transposed, so as to be consistent with the per-atom arrays and local arrays (which are necessarily transposed due to Fortran's column-major bookkeeping convention and C's row-major convention). These changes are documented with an example for the user.

Unit tests and documentation for all functions are included, including "new in version TBD" tags where appropriate. I also updated the documentation for ``lammps_scatter_atoms`` and ``lammps_gather_atoms``, which were blank (?). Documentation builds without or spelling errors (I found some false positives, also included). I also made miscellaneous edits to the documentation and the README.

**Post Submission Checklist**


hammondkd:
It appears to be failing at the kokkos initialization tests. I can reproduce this if I compile with kokkos and the default settings (i.e., set PKG-KOKKOS in ccmake, then configure and generate). It goes away if I set "t 1" in the command line instead of 2, but this definitely is not something I touched in this pull request, and I'm not sure how to fix it myself.

akohlmey:
> It appears to be failing at the kokkos initialization tests. I can reproduce this if I compile with kokkos and the default settings (i.e., set PKG-KOKKOS in ccmake, then configure and generate). It goes away if I set "t 1" in the command line instead of 2, but this definitely is not something I touched in this pull request, and I'm not sure how to fix it myself.

Can you elaborate some more on this? I only found one test failure that was not at all KOKKOS specific, but rather due to a non-portable assumption about a "USERNAME" environment variable. I've bypassed that by using `platform::putenv()` to explicitly set a custom environment variable, that can then be queried.

akohlmey:
This also includes updates to `tools/swig/lammps.i` for the new additions in the pull request plus forgotten updates for other recent  library interface changes. To make it harder to overlook this, I've also updated the comments describing which files need to be kept in sync.

hammondkd:
> I only found one test failure that was not at all KOKKOS specific, but rather due to a non-portable assumption about a "USERNAME" environment variable.

That makes a lot more sense! I thought USERNAME was set on Linux, MacOS, and Windows, but perhaps not (and I should have checked for that). Using "putenv" is much more portable, though, I agree.

The kokkos comment was because I saw something in the unit tests log that said it failed when doing one of the unit tests (there isn't much information in the log files from the auto-generated unit tests), and I thought it was pointing to the "TEST(lammps_open, with_kokkos)" line. This did indeed fail on my machine once I enabled the KOKKOS package, and if I changed the following line from
```
    const char *args[] = {"liblammps", "-k", "on", "t", "2", "-sf", "kk", "-log", "none"};
```
to
```
    const char *args[] = {"liblammps", "-k", "on", "t", "1", "-sf", "kk", "-log", "none"};
```
that test worked. However, I suspect this is a problem with how I built kokkos (having never done so before), so if fixing the USERNAME environment variable problem fixes it, let's assume all is well unless we encounter it again.

akohlmey:
> Using "putenv" is much more portable, though, I agree.

well, it only became portable after I wrote a portable wrapper for it. :grin:

> (there isn't much information in the log files from the auto-generated unit tests)

There are two ways to view the automated unit test results:
- https://ci.lammps.org/blue/organizations/jenkins/dev%2Fpull_requests%2Funit_tests/activity
- https://ci.lammps.org/job/dev/job/pull_requests/job/unit_tests/

The default (Blue Ocean) is not very informative because we have "nested" tests so that we don't exceed the GitHub API rate for "free" projects with requests from ci.lammps.org. The second is more informative (you can get there from the link on the pull request by clicking on the "arrow" icon next to the "login" button on the to right. Then you have to click on the Console output and then link of the sub-job and again on the Console output. You would end up for example here:  https://ci.lammps.org/job/dev/job/pull_requests/job/fedora/job/unit_tests/job/cmake_serial_kokkos_smallbig_shared/126/
Which tells you that the `FortranExtractVariable` test failed. Click on "Console output" and you get a verbose log of the entire test run.

> However, I suspect this is a problem with how I built kokkos (having never done so before)

This looks like you didn't set `-DKokkos_ENABLE_OPENMP=on`. By default KOKKOS only enables serial execution. If you want to enable KOKKOS but are unsure how, using/adapting one of the kokkos cmake presets in cmake/preset/ can help.

akohlmey:
@hammondkd there are likely more portability issues with the file names. There is the Windows situation with the different kind of path separator (we have portability functions for joining and splitting paths similar to those in Pythons "os" module) and also when having to support multi-configuration builds like they are created by Visual Studio and the "Ninja Multi-Config" builder.

akohlmey:
@hammondkd a more fine-grained display of the automated test is available in the "github" channel of the LAMMPS developer (not workshop) slack. I can send you an invite, if you want. It is rather verbose though (there are 40+ messages from individual runs per push or merge), so you don't want it to create notifications or sounds.

hammondkd:
> There is the Windows situation with the different kind of path separator

Working on this now. Thank you for the help finding the output from unit tests---this will make it much easier to track down things that went wrong.

akohlmey:
commit  02158516f21b31fe0321d6a71b3d8a542cfdd4cb should take care of the unit test errors due to reading files.

hammondkd:
> commit [0215851](https://github.com/lammps/lammps/commit/02158516f21b31fe0321d6a71b3d8a542cfdd4cb) should take care of the unit test errors due to reading files.

Good catch. I would have been stumbling on that one for a while longer before I figured it out....

hammondkd:
I found some additional errors when I compiled while checking for Fortran 2008 compliance; fixing them now. One may take more time.

The failed ScatterGather test is due to something I didn't realize was barred by the standard, the fixes above may fix that, too.

akohlmey:
> I found some additional errors when I compiled while checking for Fortran 2008 compliance; fixing them now. One may take more time.
> 
> The failed ScatterGather test is due to something I didn't realize was barred by the standard, the fixes above may fix that, too.

actually scatter/gather is not supported by the library interface for 64-bit atom IDs, so those tests must be skipped when -DLAMMPS_BIGBIG is defined, or - if you want to see that they are skipped - you can add a line:
```
if (lammps_extract_setting(nullptr, "tagint") == 8) GTEST_SKIP();
```
At the beginning of each test fixture.

hammondkd:
> actually scatter/gather is not supported by the library interface for 64-bit atom IDs, so those tests must be skipped when -DLAMMPS_BIGBIG is defined

That too; fixing now. The other problem turned out to be a typo, which I fixed (the compiler gave very strange error/warning messages leading me to believe that I had done something else).

hammondkd:
Yay! Thanks, @akohlmey, for help fixing some of these issues.

akohlmey:
@hammondkd removed the TODO item about testing for error messages. Needed to refactor some of the code for that. Also made programming style conventions consistent.

akohlmey:
@lammps/core is anybody going to review and possibly approve this?

Since I "own" the fortran module, it is primarily my call to say yay or nay, but I don't want to go against our agreed upon procedures to have a second person look at it without an urgent reason.

So please let me know. Thanks, Axel.

sjplimp (approved):
Thanks Karl and Axel for all your work on this.  Looks like a nice upgrade to the Fortran interface to make it compatible with support for other languages.

PR #3473, opened 2022-10-04
Added quartic function for explicit pair interaction in pair_style list.

### LINKED ISSUE #3472: [BUG] Wrong reading of pair-specific cutoff in pair_style list
**Summary**

The pair-specific cutoff specified in pair list data file required by pair_style list is read and used incorrectly.

LAMMPS SEP 15 2022
Source code

**Expected Behavior**

The pair-specific cutoff is correctly read from the explicit pair list as cutoff, not squared cutoff, so that these explicitly specified pair interactions are in effect.

**Actual Behavior**

The pair-specific cutoff is read as squared cutoff, which makes the cutoff significantly smaller than it should. In my case with real units, the cutoff is much shorter than the equilibrium pair distances. These explicitly specified pair interactions are never in effect.

**Steps to Reproduce**

In the explicit pair list data file, which is required by pair_style list, the pair-specific cutoff is explicitly given in the last column.

At line 268 of src/MISC/pair_list.cpp, the pair-specific cutoff is read as "cutsq", which is the squared cutoff. This effectively shortens the cutoff if it's larger than 1.

TerrySuun:
**Summary**

A new functional form for pair_style list, using quartic function.

**Related Issue(s)**

Also fixes #3472 
While refining our model, we decided to use additional pair specific interactions. After testing with morse poential or LJ potential, we decided that a fast-decaying function is needed. The quartic function allows greater control of the shape of the energy function. It decays to zero at r_c conveniently. Hence we'd like to make this function available to pair_style list. 

**Author(s)**

Tiedong Sun
terrencesun@gmail.com

**Licensing**

**Backward Compatibility**

This changes to pair list energy for morse interactions to be similar to bond style morse (minimum at 0.0) instead of pair style morse (minimum at D_0) for simpler testing. Forces are unchanged.

**Implementation Notes**

The code is adapted from bond_style quartic with minor adjustments. To ensure zero potential at r_c, the parameter U_0 in bond_style quartic is omitted. As r_c is an intrinsic parameter to the quartic function, r_c is required to be specified with each individual entry in the interaction list.

This PR also contains a unit test that compares results from pair style list with a system containing lj/cut and harmonic or morse bonds.

**Post Submission Checklist**

### DIFF
--- a/doc/src/pair_list.rst
+++ b/doc/src/pair_list.rst
@@ -120,7 +120,7 @@ and the coefficients:
 * :math:`b_2` (distance units)
 * :math:`r_c` (distance units)
 
-Note that the cutoff :math:`r_c` should always be specified to ensure zero energy and smooth force at cutoff.
+Note that the global cutoff specified by *pair_style list* command is ignored, unless :math:`r_c` is not specified. In this case, :math:`r_c` equals the sqare root of the globle cutoff.
 
 ----------
 
--- a/doc/src/pair_list.rst
+++ b/doc/src/pair_list.rst
@@ -30,15 +30,17 @@ Description
 """""""""""
 
 Style *list* computes interactions between explicitly listed pairs of
-atoms with the option to select functional form and parameters for
-each individual pair.  Because the parameters are set in the list
-file, the pair_coeff command has no parameters (but still needs to be
-provided).  The *check* and *nocheck* keywords enable/disable a test
-that checks whether all listed bonds were present and computed.
+atoms with the option to select functional form and parameters for each
+individual pair.  Because the parameters are set in the list file, the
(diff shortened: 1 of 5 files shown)


akohlmey:
@TerrySuun thanks for your submission. While reviewing the code, I notice that it was easy to trigger segmentation faults with invalid inputs, so I've improve some checks and also simplified some parts and improved error messages.
That includes a test to require the individual cutoff for "quartic" style list entries.

In addition, I've added a unit test that compares results from pair style list with an equivalent system using pair style lj/cut and bond styles harmonic and morse.

sjplimp (approved):

TerrySuun:
@sjplimp @akohlmey Thanks for the swift integration and improvement.

PR #3475, opened 2022-10-05
Refactor computes and fixes to use array of structs instead of many index arrays

akohlmey:
**Summary**

This refactors several computes and fixes that access data from computes, fixes, or variables to store information in an array of a struct instead of multiple arrays of the same dimension. Also store references to computes and fix class instances instead of their index in the internal lists. That avoids having to access the internal data structures in the Modify class.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

This introduces a struct `value_t` which holds the necessary data for individual references and then maintains a STL vector of them. This allows to use the simplified loop constructs introduced in C++11 and avoids all complications with memory management and synchonization/indexing of multiple arrays. Adding a new property is simplified by just adding it to the struct.

**Post Submission Checklist**


doc/src/dump.rst
```diff
@@ -202,7 +202,7 @@ or multiple smaller files).
    to a dump file may be slightly outside the simulation box.
    Re-neighbor timesteps will not typically coincide with the
    timesteps dump snapshots are written.  See the :doc:`dump_modify
-   pbc <dump_modify>` command if you with to force coordinates to be
+   pbc <dump_modify>` command if you wish to force coordinates to be
```
stanmoore1: This doesn't make sense, I think we need to remove "to", so it says "command if you wish force coordinates to be ..."

stanmoore1 (commented):
I'm not convinced that the new struct is cleaner than the original code with arrays, but my feelings aren't that strong. I think @sjplimp or @athomps should review.

akohlmey:
> I'm not convinced that the new struct is cleaner than the original code with arrays, but my feelings aren't that strong. I think @sjplimp or @athomps should review.

I agree that clarity remains about the same. That is mostly due to the complexity of the problem and using `val.which` or `which[i]` is not that much different.

There are two major problems with the arrays I am trying to address:
- when adding new features/flags/properties you need to add more arrays and do memory management for them. This is error prone and has been occasionally causing problems. Now you just add an entry to the struct and no other changes in constructor, destructor, or elsewhere are needed. 
- you can store pointers to class instances and have no longer to depend on knowing how the Modify class stores its lists of them. this is good OO programming practice and has the potential for significant benefits in the future when all such indexed accesses have been replaced.

doc/src/dump.rst
akohlmey: According to my understanding of English grammar, the "to" is required. Alternative, one might use "forcing" instead of "to force", but that sounds less "natural" to me.

But since I am not a native speaker, we probably need somebody else to weigh in on this. @sjplimp ?

sjplimp: yes, the "to" is needed

sjplimp (approved):

PR #3480, opened 2022-10-10
Documentation updates for DIELECTRIC package

aliehlen:
**Summary**

This PR updates and fills in missing information in the documentation of the DIELECTRIC package. It also changes a kspace default setting for one of the solver options, such that kspace defaults are the same and sensible for all solver options, and explicitly states this default in the documentation.

There is not an associated issue, but I have discussed these updates with @ndtrung81 separately.

**Licensing**

**Backward Compatibility**

There are no issues with backward compatibility.

### DIFF
--- a/doc/src/atom_style.rst
+++ b/doc/src/atom_style.rst
@@ -91,7 +91,7 @@ quantities.
 +--------------+-----------------------------------------------------+--------------------------------------+
 | *charge*     | charge                                              | atomic system with charges           |
 +--------------+-----------------------------------------------------+--------------------------------------+
-| *dielectric* | normx normy normz area/patch ed em epsilon curv | system with surface polarization     |
+| *dielectric* | normx normy normz area/patch ed em epsilon curv     | system with surface polarization     |
 +--------------+-----------------------------------------------------+--------------------------------------+
 | *dipole*     | charge and dipole moment                            | system with dipolar particles        |
 +--------------+-----------------------------------------------------+--------------------------------------+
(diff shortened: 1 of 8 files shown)


doc/src/atom_style.rst
```diff
@@ -91,7 +91,7 @@ quantities.
 +--------------+-----------------------------------------------------+--------------------------------------+
 | *charge*     | charge                                              | atomic system with charges           |
 +--------------+-----------------------------------------------------+--------------------------------------+
-| *dielectric* | dipole, area, curvature                             | system with surface polarization     |
+| *dielectric* | normx normy normz area_per_patch ed em epsilon curv | system with surface polarization     |
```
stanmoore1: We typically don't like "_" (underscores) in the documentation, I would suggest changing `area_per_patch` to `area/patch`

examples/PACKAGES/dielectric/README
```diff
@@ -40,3 +40,7 @@ where
 * curvature: surface mean curvature at the patch.
   For example, for spherical interfaces, curvature = 1/spherical radius.
   For planar interfaces, curvature = 0.
+
+Note that the properties normx, normy, normz, area_per_patch, ed, em, and curvature are not
```
stanmoore1: See above

doc/src/atom_style.rst
```diff
@@ -180,16 +180,21 @@ vector with the 3 diameters of the ellipsoid and a quaternion 4-vector
...
-dielectric constants of two sides of the interface, and the local
-dielectric constant at the boundary element.  The distinction between
-the physical and interface particles is only meaningful when :doc:`fix
-polarize <fix_polarize>` commands are applied to the interface
-particles.
+particle (e.g. an ion), or an interface particle representing a boundary
+element between two regions of different dielectric constant. For
+interface particles, in addition to the properties associated with
+atom_style full, each particle also should be assigned a normal unit vector
+(defined by normx, normy, normz), an area (area_per_patch), the difference and
```
stanmoore1: change to `area/patch`

doc/src/atom_style.rst
aliehlen: All right, sorry about that. I can fix it now.

stanmoore1 (approved):

PR #3482, opened 2022-10-11
Fix issue with KSpace slab correction energy with non-neutral systems

### LINKED ISSUE #3413: [BUG] Incorrect background charge correction term for charged slab systems
**Summary**

When using the electrostatic slab correction from Yeh & Berkowitz and the background charge correction from Ballenegger for charged systems, the background-background charge interaction term is calculated incorrectly. In pppm.cpp in the slabcorr() and slabcorr_groups() functions, this term is calculated as:

e_slabcorr = MY_2PI\*(dipole_all\*dipole_all - qsum\*dipole_r2 - qsum\*qsum\*zprd\*zprd/12.0)/volume;

Whereas it should use the system extent in the z direction plus vacuum padding:

e_slabcorr = MY_2PI\*(dipole_all\*dipole_all - qsum\*dipole_r2 - qsum\*qsum\*zprd_slab\*zprd_slab/12.0)/volume;

This does not affect the forces.

I am using lammps-29Oct20 on Linux,
but the bug is also present in lammps-23Jun22

**Expected Behavior**

I expect the energy of the system does not change significantly when the vacuum padding of the slab is varied since the energies are corrected.

**Actual Behavior**

When the slab volfactor value is changed, the energy of the system changes significantly. 

**Steps to Reproduce**

Run the attached file and see how etotal varies before and after I set 
kspace_modify slab 5.0
I verified that my code suggestion at the top fixes the problem.

[slabtest.zip](https://github.com/lammps/lammps/files/9417134/slabtest.zip)

stanmoore1:
**Summary**

KSpace slab correction energy with non-neutral systems was using the wrong length (`zprd` instead of `zprd_slab`), so the energy varied with the `volfactor` parameter. Forces were not affected.

**Related Issue(s)**

Fixes #3413

**Author(s)**

Stan Moore (SNL)

**Licensing**

**Backward Compatibility**

Yes

### DIFF
--- a/src/DIELECTRIC/pppm_dielectric.cpp
+++ b/src/DIELECTRIC/pppm_dielectric.cpp
@@ -412,7 +412,7 @@ void PPPMDielectric::slabcorr()
   double **x = atom->x;
   double *eps = atom->epsilon;
 
-  double zprd = domain->zprd;
+  double zprd_slab = domain->zprd*slab_volfactor;
   int nlocal = atom->nlocal;
 
   double dipole = 0.0;
--- a/src/DIELECTRIC/pppm_dielectric.cpp
+++ b/src/DIELECTRIC/pppm_dielectric.cpp
@@ -442,7 +442,7 @@ void PPPMDielectric::slabcorr()
   // compute corrections
 
   const double e_slabcorr = MY_2PI*(dipole_all*dipole_all -
-    qsum*dipole_r2 - qsum*qsum*zprd*zprd/12.0)/volume;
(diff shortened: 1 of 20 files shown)


stanmoore1:
Running the test in #3413, I do see a more constant energy when changing `volfactor`.

stanmoore1:
@jonwils @srtee

stanmoore1:
All the regression tests passed, and all the unit tests passed, so we must not have any tests with non-neutral systems and KSpace and slab correction?

stanmoore1:
@akohlmey this should go into the stable release (#3460)

sjplimp (approved):

Issue #3467, opened 2022-09-30
[Feature Request] Requesting a new bond_style quartic/cut


TerrySuun:
**Summary**

A new bond_style quartic/cut

**Detailed Description**

In our modeling, to improve an existing molecule model, we need a bond type which behaves similar to LJ potential: the energy goes to zero when r goes to infinity. We need it for specific pair of atoms, not between two pair types, hence adding another pair interaction would involve pairs of atoms that are excluded from this interaction.

So far, bond_style quartic is the closest one but not ideal for two reasons. 1, there's an LJ term in energy of bond_style quartic, which is not necessary and complicates the model. 2, the bond_style quartic bond can be broken but cannot be rebuilt when bond atoms are within short distance again. 

Hence the ideal bond type would be only using the quartic terms in the current bond_style quartic, when r within cutoff. It should take constant energy value (can be U0 as in bond_style quartic) when r is beyond cutoff.  

**Further Information, Files, and Links**

Though this seems like an unusual bond to be useful, it will really make our modeling work easier.

akohlmey:
Please note that most features in LAMMPS are not added by the LAMMPS developers, but contributed by LAMMPS users that had a specific need for a specific feature in LAMMPS. Please see https://docs.lammps.org/Modify.html for more information on how to add features to LAMMPS.

akohlmey:
BTW: What you describe sounds a lot like what pair style `list` implements: https://docs.lammps.org/pair_list.html

TerrySuun:
Yes, indeed. I'll try implement with pair_style list to see if it works for us.

akohlmey:
After re-reading the description, I am thinking that a new bond style is not going to solve the problem, since the atoms may not move beyond the communication cutoff or else there will be bond atom lost errors or the communication cutoff must be extended and the calculation will slow down. Pair style `list` is more suited for the given conditions. If needed an additional functional form can be added.

I will close this issue now. Please submit a new issue if you need changes to Pair style `list` (no promises, however, how quickly they will be done. As already mentioned, it may be simpler that you make the change yourself and then submit a pull request).

PR #3486, opened 2022-10-13
Add optional unbreakable bond_style quartic.

TerrySuun:
**Summary**

Related to my earlier request #3467 
https://github.com/lammps/lammps/issues/3467

For our model, we require not only specific pairwise interactions, but also suitable exclusion rules for 1-2 and 1-3 neighbors. In this case, a bond interaction is more convenient than pair_style list.

The original bond_style quartic requires special_bonds 1.0 1.0 1.0, to turn on the nonbonded pair interaction after bond breaking. 

In this PR, a optional "breakable" keyword is added to bond_style quartic, so that a more conventional bond definition with quartic potential is possible. Such bonds are treated as normal bonds, and don't require specific special_bonds option. 

The default is kept the same as the original "bond_style quartic" bonds.

**Related Issue(s)**

Fulfills request #3467 

**Author(s)**

Tiedong Sun
School of Biological Sciences, Nanyang Technological University, Singapore

**Licensing**

**Backward Compatibility**

backward compatible. 

**Implementation Notes**

A global switch (BondQuartic::isBreakable_glob) and bond type-specific switches (BondQuartic::isBreakable) are added to control whether certain bond type is "breakable". The global option is set by the newly added BondQuartic::settings method.
The bond type-specific option is set when bond coefficients are read (BondQuartic::coeff). 

If any bond type is breakable, special_bonds 1 1 1 is enforced. 

Two simulations are compared and verified to produce the same trajectory:
1) bond_style quartic breakable no
2) bond_style zero + pair_style list (same quartic potential as in 1) 

**Post Submission Checklist**


akohlmey:
@TerrySuun thanks for your submission. However, I don't like how it is implemented as an option and that the breakable/unbreakable flag has to be checked all the time. I would prefer that we would have a bond style "quartic/breakable" and "quartic/unbreakable" where one is a derived class of the other. there still could be only one implementation of the compute and single functions, the compute function could become a wrapper around a templated "eval" function where the breakable/unbreakable flag is a template parameter and thus will become a compile time constant. The single function is not time critical, so the if statement could remain. For consistency with the programming style of the rest of LAMMPS there should be a class variable `breakable_flag` that is set in the respective constructors to 0 or 1 as needed. For backward compatibility there could be an alias "quartic" for the breakable variant. You can see examples for templated functions in the OPENMP package and also the "shared" implementation of a derived class with the airebo and airebo/morse pair styles.

TerrySuun:
Hi @akohlmey, I tried a new implementation with a derived class for quartic/breakable. The default bond_style quartic is aliased to quartic/breakable. The unittest for quartic/unbreakable uses the same data for quartic/breakable, which might need updates. The manual page is updated as well. Hope this one is better than the previous attempt.

src/MOLECULE/bond_quartic.cpp
```diff
@@ -76,11 +73,34 @@ void BondQuartic::compute(int eflag, int vflag)
...
+    if (evflag) {
+      if (eflag) eval<0,1,1>(cutsq, x, f, bondlist, nbondlist, nlocal, newton_bond);
+      else eval<0,0,1>(cutsq, x, f, bondlist, nbondlist, nlocal, newton_bond);
+    } else eval<0,0,0>(cutsq, x, f, bondlist, nbondlist, nlocal, newton_bond);
+  }  
+}
+
+/* ---------------------------------------------------------------------- */
+template <int BREAKABLE, int EFLAG, int EVFLAG>
+void BondQuartic::eval(double **cutsq, double **x, double **f, int **bondlist, int nbondlist, int nlocal, int newton_bond)
```
akohlmey: What is the point of passing those arguments? As a member function of the class, the eval() template has access to class members and the Atom and other class pointers from the Pointers class just like the compute() function.

src/MOLECULE/bond_quartic.h
```diff
@@ -14,6 +14,7 @@
 #ifdef BOND_CLASS
```
akohlmey, line 14: The file(s) and class name should be renamed to be consistent with the (new) principal command name.
OR bond/breakable and bond/unbreakable should both be derived classes and have the BondStyle macro, while it would be removed from the `bond_quartic.{cpp,h}` files.

If you go for the first option (= rename files), the old file names need to be listed in the `src/Purge.list` file so they can be purged with `make purge` when somebody updates the git checkout and uses the conventional build system.  All new files need to be added to `src/.gitignore` so they won't be easily added to git by accident in the same case.

src/MOLECULE/bond_quartic_unbreakable.cpp
```diff
@@ -0,0 +1,56 @@
...
+  partial_flag = 1;
+  breakable_flag = 0;
+}
+
+/* ---------------------------------------------------------------------- */
+
+BondQuarticUnbreakable::~BondQuarticUnbreakable()
+{
+  if (allocated) {
+    memory->destroy(setflag);
```
akohlmey: Why delete those arrays here? (and then attempt a second time in the base class?
In general, memory->destroy() is safe to call on unallocated arrays, for as long as they are properly initialized to a null pointer in the constructor's initializer list of the class where they are defined.

src/OPENMP/bond_quartic_omp.h
```diff
@@ -17,7 +17,7 @@
 
 #ifdef BOND_CLASS
 // clang-format off
-BondStyle(quartic/omp,BondQuarticOMP);
+BondStyle(quartic/breakable/omp,BondQuarticOMP);
```
akohlmey, line 21: Why not have an alias to bond_style quartic/omp here? This will break using the `omp` suffix for old inputs.

src/MOLECULE/bond_quartic.cpp
```diff
@@ -229,26 +253,33 @@ void BondQuartic::coeff(int narg, char **arg)
...
-double BondQuartic::equilibrium_distance(int /*i*/)
+double BondQuartic::equilibrium_distance(int i)
 {
-  return 0.97;
+  // equilibrium distance of the quartic potential, excluding LJ
+  double b2_4ac = 9.0 * (b1[i] * b1[i] + b2[i] *  b2[i]) - 14.0 * b1[i] * b2[i];
+
+  if (b2_4ac >= 0.0) {
+    return (8.0 * rc[i] + 3.0 * (b1[i] + b2[i]) - sqrt(b2_4ac) ) / 8.0;
+  } else return rc[i];
```
akohlmey: I don't think this is correct for the breakable case. Well, neither is 0.97, which is only true for k=30 (and sigma=1.0 epsilon=1.0 which are implied by the bond style).

akohlmey (changes requested):
A few more updates are required.

It may be worth waiting for @sjplimp to weigh in on some of the suggestions and comments in case his preferences differ from mine.

src/OPENMP/bond_quartic_omp.h
TerrySuun: Overlooked. thanks for pointing out.

src/MOLECULE/bond_quartic.cpp
TerrySuun: Yes, this is only valid for rc > r_eqilibrium > 2^(1/6), to pass the unit test. It is the equilibrium length if the energy function is without the LJ term. I'll update it.

src/MOLECULE/bond_quartic_unbreakable.cpp
TerrySuun: So in this case, I should use an empty destructor?

src/MOLECULE/bond_quartic.cpp
TerrySuun: Thanks for pointing it out. This is silly.

src/MOLECULE/bond_quartic_unbreakable.cpp
akohlmey: > So in this case, I should use an empty destructor?

Just go with the default destructor and don't provide an explicit one.

src/MOLECULE/bond_quartic.h
TerrySuun: I went with the renaming option.

src/MOLECULE/bond_quartic_breakable.cpp
```diff
@@ -92,42 +112,44 @@ void BondQuartic::compute(int eflag, int vflag)
...
+        bondlist[n][2] = 0;
+        for (m = 0; m < atom->num_bond[i1]; m++)
+          if (atom->bond_atom[i1][m] == atom->tag[i2]) atom->bond_type[i1][m] = 0;
+        if (i2 < atom->nlocal)
+          for (m = 0; m < atom->num_bond[i2]; m++)
+            if (atom->bond_atom[i2][m] == atom->tag[i1]) atom->bond_type[i2][m] = 0;
+        fbond = 0.0;
+        if (EFLAG) ebond = 0.0;
+      } else {
+        // unbreakable bond
```
akohlmey, line 132: This part does not make much sense to me and neither the addition of the LJ repulsion below.
Why not have the quartic potential only for the "unbreakable" bond, regardless of its length?
Having no force and constant energy does not make much sense.
And then do the switch and LJ contribution only for the breakable version.

doc/src/bond_quartic.rst
```diff
@@ -1,45 +1,92 @@
...
 """""""""""
 
-The *quartic* bond style uses the potential
+The *quartic/breakable* and *quartic/unbreakable* bond styles use the potential
 
 .. math::
 
    E      & = E_q + E_{LJ} \\
-   E_q    & = K (r - R_c)^ 2 (r - R_c - B_1) (r - R_c - B_2) + U_0 \\
+   E_q    & = \left\{ \begin{array} {l@{\quad:\quad}l}
```
akohlmey, line 47: It would be cleaner to have two `.. math::` blocks. One for each variant.

src/MOLECULE/bond_quartic_breakable.cpp
TerrySuun: Oops, my mistake here. Breakable bonds should be broken here. The force and energy calculation below should be skipped. For unbreakable bonds, we still need to check with cutoff to decide if it's constant energy.

doc/src/bond_quartic.rst
TerrySuun: Updated.

src/MOLECULE/bond_quartic_breakable.cpp
akohlmey: You are still not making sense. For "unbreakable" bonds there should be no "cutoff". That is what makes them unbreakable. You seem to be confusing the meaning of R_c.  R_c is no "cutoff". Just plot the potential function for a few choices of parameters and you should see.

TerrySuun: @akohlmey 
Hmm, seems we are indeed talking about different things. The "breakable" bonds are broken once r > rc, and never "rebond", which means fbond=0 and ebond=0 once r>rc. I'm trying to introduce the "rebond" capability of this bond type, hence used fbond=0 and ebond=U_0 beyond rc, so that the force due to the bond is the same in two variants. And bondlist[i][2] is never set to 0. That's what "unbreakable" means to me. The potential in this case is with single minimum, and ebond = U_0 when r goes to infinity (assuming B1*B2 < 0, as in the example in the doc). 

By the way, it is producing the same forces with pair_style list quartic, where fbond=0 and ebond=0 when r>rc (differ with bond quartic by U_0 only). (Maybe you didn't intend to include it in this form?)

I guess your potential function for the unbreakable bond is the real double-well quartic function, where the bonds don't break even without a cutoff. But this form is never available in LAMMPS, to my knowledge.

Now when thinking about it, it is confusing to use a quartic functional form but only use the r<rc part. But it provides an effectively breaking bond, with smooth energy, at a specified distance with "rebond" capability, which I cannot find in other bond types. 

If it is decided that this type of bond is not desirable, users can always use bond_style table, of course, if such bonds are needed. 

Please advise how to proceed. Thanks.

akohlmey: > That's what "unbreakable" means to me

That makes no sense to me. And it is meaningless in LAMMPS because this way the bonded atoms can move an arbitrary distance, which conflicts with the domain decomposition parallelization.

Also, it should be possible to reform a bond with some atom from a *different* bond that was broken and that is not possible with this style. To model bonds that can break and then form again, you can use fix bond/react.

> By the way, it is producing the same forces with pair_style list quartic, where fbond=0 and ebond=0 when r>rc (differ with bond quartic by U_0 only). (Maybe you didn't intend to include it in this form?)

Yes, that was not intended. I will have to revise that implementation.

> Please advise how to proceed. Thanks.

Let me discuss with the other LAMMPS developers. There are things that are strange about the upstream implementation and others that make no sense to me, so I need to figure out who is using it for what and then look into how to go on. The sad part is that it now looks quite likely that we need to abandon this pull request and our combined efforts will be mostly useful toward our own better understanding of the situation.

TerrySuun: @akohlmey 

> That makes no sense to me. And it is meaningless in LAMMPS because this way the bonded atoms can move an arbitrary distance, which conflicts with the domain decomposition parallelization.

I realize that from the beginning, I focused on this special case of using this bond type for specific atom pairs that belong to the same molecule. In such cases, the distance between these two atoms is bounded because they still belong to the same molecule. Hence if the communication cutoff is large enough, for example larger than the size of the molecule, the simulation can proceed, though slower.

> The sad part is that it now looks quite likely that we need to abandon this pull request and our combined efforts will be mostly useful toward our own better understanding of the situation.

It's not sad for me at all :)  Looking forward to your decisions.

akohlmey: @TerrySuun as expected, after my discussion with other LAMMPS developers, the consensus is that you or others with a similar interest should look into using fix bond/react. Of course you are free to keep using your modifications, but we will not integrate them into the LAMMPS distribution.

TerrySuun: @akohlmey 
OK, understand. But I'm not going to use fix bond/react. I don't want to change my topology during simulation. I just need another functional form for my bonds. Thanks for all the help.

sjplimp: @TerrySuun I suggest you talk to Jake Gissinger (author of fix bond/react) to see if what you have in mind is possible wirh fix bond/react.  It can certainly allow bonds to break and reform.  I'm not sure what you mean by "dont want to change topology".  When the bond breaks there has to be some topological change.  For example, I assume you don't want to
continue to compute 3- and 4-body interactions that include the broken bond.  What is computationally implausible about your model is that you want to check the distance between 2 broken-bond atoms no matter how far apart they are,  That's difficult to do correctly in a parallel code.  You say they won't be too far apart, but how can that be the case for any usage of the model, e.g. one which strains the system until bonds break?  Also it seems physically implausible to require the same 2 atoms rebond (requiring some memory of previous toplogy).  How would that be the case in a real material.  Why wouldn't rebonding occur when any two atoms with the correct local topology got close enough together (e.g. atoms which were previously in 2 different molecules)?  All of that is something I believe fix bond/react can do more naturally that what you are proposing, both computationally and physically.

TerrySuun: @sjplimp Thanks for the suggestion. My situation is that we have a model for a complex, and we're trying to "patch" it to fit better to some data. Now our model is not "flexible" enough, hence the idea of replacing certain harmonic bonds with "diminishing" bonds that provide zero force at long bond length. I agree this is a special, even weird model, even though it is quite similar to bond_style morse. (BTW, I've tried bond_style morse, which is too soft on the attractive part.) But this is probably the most straightforward way of doing it, as re-calibrate the whole model is quite time-consuming. This is indeed not a good model for general use, especially when considering the problem in parallel simulations. But I still don't see fix bond/react suitable for our need, as we'd like to keep the 1-2, 1-3 and 1-4 pair interaction constant throughout the simulation.

> I'm not sure what you mean by "dont want to change topology". When the bond breaks there has to be some topological change. For example, I assume you don't want to continue to compute 3- and 4-body interactions that include the broken bond. 

I don't want the bond to break, just to provide zero force at long distance (I know it's technically broken. Weird.). Something "bond_style harmonic/shift/cut" provides, but with continuous force at cutoff. All other interactions, 1-2, 1-3 pairs, 3-,4-body interactions, are still calculated the same way all the time. Just the bond energy can go to a specific constant, and bond force can be zero.

> What is computationally implausible about your model is that you want to check the distance between 2 broken-bond atoms no matter how far apart they are, That's difficult to do correctly in a parallel code. 

Yes, this is not good. But I don't think it is much different from bond_style morse, where you can also have very long bond lengths.

> Also it seems physically implausible to require the same 2 atoms rebond (requiring some memory of previous toplogy). 

The memory is already there, in bondlist. And hey, are harmonic bonds *physically plausible* at all times? How about bond_style harmonic/shift/cut? They can "rebond" if I understand the doc correctly.

> How would that be the case in a real material. Why wouldn't rebonding occur when any two atoms with the correct local topology got close enough together (e.g. atoms which were previously in 2 different molecules)? 

It's just a model, to hopefully give rise to two states. And it's supposed to be just a "patch". These questions are equally valid for bond_style harmonic/shift/cut. 

******
I think this PR just provides one more functional form for bonds. The raised problems based on physical and practical reasons are already there with other bond types. Maybe just let the users judge and decide, with proper cautions and warnings?

Anyway, thanks for your time and effort. I'll try to see if there's other way around. Now bond_style table is on the top of my list. Hopefully we can find some way to fit the experimental data with a small set of parameters.

sjplimp: ok - thanks for the clarification - I see more clearly now what you are trying to do.  I think bond_style table could do what you want, but you will have to parameterize it yourself obviously.  If you want to do something like bond morse or bond harmonic/shift/cut, just with a different functional form, then that would be fine.  But I suggest you divorce it from the existing bond quartic (not an option, not a derived class, not the same name).  This is b/c bond quartic is associated with breakable bonds for LAMMPS users and they are conceptually different than static bonds, both in the way they are used, and internally.   And as you said you really want a regular static bond (never breaks), just one whose force goes to zero at sufficient distance.  I agree that could be documented so users understand its limitations when atoms get too far apart, and LAMMPS will throw an error in that case anyway, if a processor can't find both atoms.

I've suggested to Axel that we rename bond quartic to bond quartic/break to emphasize it is is a breakable bond.
So you could name your new style something like bond quartic/cut (similar to harmonic/shift/cut) if you want to use a quartic formula with an added cutoff feature.

TerrySuun: @sjplimp Thanks for the reply.
 
> I've suggested to Axel that we rename bond quartic to bond quartic/break to emphasize it is is a breakable bond. So you could name your new style something like bond quartic/cut (similar to harmonic/shift/cut) if you want to use a quartic formula with an added cutoff feature.

This makes sense. I started with quartic bond only because the functional form is flexible enough for my needs, yet with a small number of parameters.

PR #3488, opened 2022-10-14
Improving stability of the performance of Tersoff/Kokkos across compiler and Kokkos versions

weinbe2:
**Summary**

This PR addresses fluctuations in the performance of Tersoff across CUDA (and in theory HIP, though unverified) between compiler versions and Kokkos versions by constraining the "chunk size" of work, which translates to holding the blocksize fixed "under the hood." This removes version-to-version performance variations due to Kokkos's heuristics either reacting differently to how code is generated for different compiler versions, or due to potential changes to Kokkos's heuristics themselves.

**Author(s)**

**Licensing**

**Backward Compatibility**

No issues.

**Implementation Notes**

In brief, the _ideal_ way to fix this would have been able to specify a tiling/chunking/blocking size to simple Kokkos 1-d `RangePolicy` launches, but that functionality is not there (yet) --- this is feedback I've handed off to the Kokkos developers.

The alternative, more natural way to expose work batching is via `TeamPolicy` parallelism, where the number of threads per team (in Kokkos language) can be tuned by hand. This approach has been implemented in this PR. However, due to Kokkos heuristics w/regards to specifying L1 vs shared memory carveouts on NV GPUs, this leads to a large performance regression. This will in principle be addressed once https://github.com/kokkos/kokkos/pull/4295 is refined and merged.

A less natural but functional approach is to take advantage of `MDRangePolicy`s in Kokkos, which does expose the ability to set a tiling size. However, `MDRangePolicy` does not expose a "rank 1" launch (a piece of feedback I've also shared with the Kokkos devs; it's a person-hour problem). The workaround for this problem is to use a "rank 2" (two dimensional) launch, with the bounds of the second dimension trivially set to one. In practice the overhead to this is negligible, it's just textually messy. Nonetheless, this approach has _also_ been implemented.

All three versions of the Tersoff potential calculations now exist: `RangePolicy`, `MDRangePolicy`, and `TeamPolicy`. To avoid a mess of code reuse, the potential calculation has been abstracted into a single `tersoff_compute` function that is reused everywhere.

The behavior in this PR, as written, is as follows:
 * Host is unchanged (as toggled by `HALF`); it still uses a `RangePolicy`.
 * Device (as toggled by `HALFTHREAD`) currently uses `MDRangePolicy` for the force calculation (`Kokkos::parallel_for`) and `RangePolicy` for the energy calculation (`Kokkos::parallel_reduce`). Unfortunately, switching to an `MDRangePolicy` for the energy calculation led to a ~5% regression.

As noted above, once https://github.com/kokkos/kokkos/pull/4295 is merged, the `MDRangePolicy` path can be removed and the `TeamPolicy` path can be used --- an early hacked test with what the Kokkos PR is supposed to accomplish shows performance parity between the two approaches. That said, the `TeamPolicy` path can quickly be tested today by commenting out the line `#define KOKKOS_TERSOFF_MDRANGEPOLICY_WORKAROUND`. (Of note --- it could make more sense to run the `TeamPolicy` implementation on other architectures; there's no reason why `Team` vs `MDRange` couldn't be toggled as a function of architecture at this time.)

Of note: all of this is documented within the code.

Performance and correctness were verified by compiling and running the code, checking CUDA 11.2, 11.6, and 11.8. This PR closes the performance gaps between different CUDA versions, and in all three cases led to a performance uplift of various magnitude.

**Post Submission Checklist**

None.


stanmoore1:
Performance looks good on V100, AMD, and CPU.

stanmoore1 (approved):

PR #3489, opened 2022-10-14
Doc fixes (round 4)

hammondkd:
**Summary**

This PR contains edits to several documentation files. This should, assuming my probably-not-perfect AWK script found all of them, fix all remaining discrepancies of the form
```
* keyword = *any* or *undo* or *special*

  .. parsed-literal::

       *any* arg = none = turn off interactions if any atoms are in the group (or on if *undo* is also used)
       *undo* arg = none = turn specified bonds on instead of off
       *remove* arg = permanently remove bonds that have been turned off
       *special* arg = recompute pairwise 1-2, 1-3, and 1-4 lists
```
(note that *remove* is missing on the first line) in the documentation.

I also made some math-related edits and more generally cleaned up the documentation for several fix commands, leaving off at fix bond/react. fix "bocs" in particular had some missing keywords and did not match the source code (it does now).

I added a few lines to the LaTeX preamble in the Sphinx configuration file (doc/utils/sphinx-config/conf.py.in) that eliminate the problem of section numbers overlapping in the table of contents when there are more than 100 subsections or more than 10 subsections and more than 10 subsubsections in a particular section. Those changes should only affect the PDF version of the documentation.

**Author(s)**

Karl Hammond
University of Missouri

**Licensing**

**Backward Compatibility**

I do not think there are any changes to actual code in this one, so nothing should change other than how it's documented.

**Implementation Notes**

I noted in the PDF that the body text seems to be in Times but the math is still in Computer Modern Roman; the following line added to the LaTeX preamble would make the math into Times Roman as well, at the expense of roping in another required package:
```
\usepackage{mathptmx}
```
I couldn't find where Times was brought in (probably through Sphinx itself, and obviously not through the usual methods, such as including one of the three or four packages I know of for Times, or I would have found it in the log file), so I didn't dare bring in something like newtxtext/newtxmath or txfonts. The documentation builds correctly for the fasthtml, html, and pdf targets.

**Post Submission Checklist**


akohlmey:
@hammondkd re: math fonts. I have copied over some LaTeX "magic" that I once used to get consistent postscript fonts in my PhD thesis. This was, however, for using the "New Century Schoolbook" font, which I liked better than "Times New Roman" to have better readability for both printed and specifically the electronic version. However, that font needs more space and thus will blow up the manual by several hundred pages and has more lines where the last few words are wrapped.

I've adapted it for "Times New Roman" and stuck this into your branch for testing.

hammondkd:
> This was, however, for using the "New Century Schoolbook" font, which I liked better than "Times New Roman" to have better readability for both printed and specifically the electronic version.

The "fouriernc" package gives New Century Schoolbook with comparable math fonts, I think; what you proposed is essentially a manual version of what something like times+mathptmx, times+stix, txfonts, or newtxtext+newtxmath do. You are absolutely correct, though: New Century Schoolbook is a "wider" typeface than Times, so the manual would be a fair bit longer. Something like Bitstream Charter, Garamond, or Utopia (`\usepackage[charter]{mathdesign}`, `\usepackage[garamond]{mathdesign}`, and `\usepackage[utopia]{mathdesign}`, respectively) might be similar in length to the Times version but not in Times (which I don't mind, but I understand some people get tired of seeing).

athomps (approved):
I approve. Thanks @hammondkd for the thorough doc checking.

PR #3503, opened 2022-10-28
Refactor handling of pair/only flag for GPU and KOKKOS

akohlmey:
**Summary**

This changes the way how the "pair/only" keyword is handled for both the GPU and KOKKOS package in a way that will recover the currently broken KOKKOS support.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

This restores lost functionality.

**Implementation Notes**

Rather than swapping away the suffix string, this now uses a true/false flag and then adjusts the code that create the force styles (all others remain unchanged).

**Post Submission Checklist**


stanmoore1:
@akohlmey this appears to be working except the output of the neighborlist build has the `kk/host` suffix in the output of the pair style, which is incorrect:

```
  (1) pair sw/kk/host, perpetual
      attributes: full, newton on, kokkos_device
      pair build: full/bin/kk/device
      stencil: full/bin/3d
      bin: kk/device
```

stanmoore1:
I did more testing and it looks like more than just the pair style is running on the GPU with `pair/only on`. For the LJ benchmark, at least the `fix nve` is also running on the GPU.

```patch 
diff --git a/src/KOKKOS/fix_nve_kokkos.cpp b/src/KOKKOS/fix_nve_kokkos.cpp
index 5dcb611d41..f6554b9ffb 100644
--- a/src/KOKKOS/fix_nve_kokkos.cpp
+++ b/src/KOKKOS/fix_nve_kokkos.cpp
@@ -32,6 +32,9 @@ FixNVEKokkos<DeviceType>::FixNVEKokkos(LAMMPS *lmp, int narg, char **arg) :
   atomKK = (AtomKokkos *) atom;
   execution_space = ExecutionSpaceFromDevice<DeviceType>::space;
 
+  if (execution_space == Host) printf("Fix Host!!!!\n");
+  else if (execution_space == Device) printf("Fix Device!!!!\n");
+
   datamask_read = X_MASK | V_MASK | F_MASK | MASK_MASK | RMASS_MASK | TYPE_MASK;
   datamask_modify = X_MASK | V_MASK;
 }
diff --git a/src/KOKKOS/pair_lj_cut_kokkos.cpp b/src/KOKKOS/pair_lj_cut_kokkos.cpp
index e818375762..9878a90683 100644
--- a/src/KOKKOS/pair_lj_cut_kokkos.cpp
+++ b/src/KOKKOS/pair_lj_cut_kokkos.cpp
@@ -40,6 +40,8 @@ PairLJCutKokkos<DeviceType>::PairLJCutKokkos(LAMMPS *lmp) : PairLJCut(lmp)
   kokkosable = 1;
   atomKK = (AtomKokkos *) atom;
   execution_space = ExecutionSpaceFromDevice<DeviceType>::space;
+  if (execution_space == Host) printf("Pair Host!!!!\n");
+  else if (execution_space == Device) printf("Pair Device!!!!\n");
   datamask_read = X_MASK | F_MASK | TYPE_MASK | ENERGY_MASK | VIRIAL_MASK;
   datamask_modify = F_MASK | ENERGY_MASK | VIRIAL_MASK;
 }
```

```
./lmp_summit_kokkos -in in.lj -k on g 1 -sf kk -pk kokkos pair/only on

Pair Device!!!!
Fix Device!!!!
```

akohlmey:
> I did more testing and it looks like more than just the pair style is running on the GPU with `pair/only on`.

Yes. The pair/only setting is only applied for force computation.

stanmoore1:
>Yes. The pair/only setting is only applied for force computation.

I'm a little confused. The `pair/only` option is supposed to use the `kk` suffix for pair style (only) and `kk/host` for everything else.

stanmoore1:
```
    pair/only = off or on
      off = use device acceleration (e.g. GPU) for all available styles in the KOKKOS package (default)
      on  = use device acceleration only for pair styles (and host acceleration for others)

The pair/only keyword can change how the KOKKOS suffix “kk” is applied when using an accelerator device. By default
device acceleration is always used for all available styles. With pair/only set to on the suffix setting will choose device
acceleration only for pair styles and run all other force computations on the host CPU. The comm flags will also
automatically be changed to no. This can result in better performance for certain configurations and system sizes.
```

akohlmey:
> > Yes. The pair/only setting is only applied for force computation.
> 
> I'm a little confused. The `pair/only` option is supposed to use the `kk` suffix for pair style (only) and `kk/host` for everything else.

That requires much more extensive changes. Not much of a difference for the GPU package, although to use the /gpu integrators may be beneficial.

Well, this is why I flagged this as draft...

stanmoore1 (approved):
This now passes my test. I approve.

PR #3504, opened 2022-10-29
Update example for coupling LAMMPS as a dynamically loaded library

akohlmey:
**Summary**

This updates the example in `examples/COUPLE/plugin` to the current state of the LAMMPS library interface.
It also adds a CMake build script for the example application (in C).

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

This now follows the convention from library.h, that `-DLAMMPS_LIB_MPI` needs to be defined so that the `lammps_open()` function that uses an MPI Communicator as argument is exposed. This allows to use the interface without MPI being available/used in the calling code. The LAMMPS library may still be compiled with MPI support.

**Post Submission Checklist**


stanmoore1 (approved):
I approve

sjplimp (approved):

akohlmey:
@stanmoore1 @sjplimp with this pull request, we are turning this plugin loader effectively into a supported feature (we test if it compiles and can run the example input, the loader is ported to Windows), it should probably also be documented somewhere beyond the README file. Any suggestion where to best add it?

sjplimp:
Not sure I understand the feature.  But somewhere in Sections 3 and/or 4 of
the manual?
Possibly also a one-liner in Section 1 - Features
Or possibly add a paragraph (or bullet section) in Section 1 about all the
different ways LAMMPS can be used: exe, standard lib, Python, plugin, etc

Steve

akohlmey:
> Not sure I understand the feature.

This is like using the LAMMPS C-library interface, but instead of using library.h and linking to liblammps.so/a you link to the "plugin object" and the dynamic loader library (if required). That way the executable is independent from the liblammps.so file. Instead you provide the name of the file you want to load at runtime, and then it is loaded similar to what ctypes does for the LAMMPS python module. This means, you can recompile and change the LAMMPS shared library without having to change or recompile/relink your program. For as long as the number and type of arguments does not change for the library functions you are using, it can work for a wide range of LAMMPS versions.

stanmoore1:
@akohlmey I'm trying to build `examples/COUPLE/plugin` but I get this error, any ideas? I'm using g++ 7.3.0. 

```
liblammpsplugin.c:38:43: error: invalid conversion from 'void*' to 'void (*)(void*)' [-fpermissive]
 #define ADDSYM(symbol) lmp->symbol = dlsym(handle,"lammps_" #symbol)
                                      ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~
```

stanmoore1:
Clang 14 gives this error:
```
liblammpsplugin.c:96:3: error: assigning to 'int (*)(void *, const char *)' from 'void *' converts between void pointer and function pointer
  ADDSYM(extract_setting);
  ^~~~~~~~~~~~~~~~~~~~~~~
```

akohlmey:
> I'm using g++ 7.3.0.

This is C code, you should use gcc.

stanmoore1:
>This is C code, you should use gcc.

Ah yes, that works, thank you!

stanmoore1:
This gives a warning with `-pedantic` flag:

```
liblammpsplugin.c:65:36: warning: ISO C forbids assignment between function pointer and 'void *' [-Wpedantic]
 #define ADDSYM(symbol) lmp->symbol = dlsym(handle,"lammps_" #symbol)
```

akohlmey:
> This gives a warning with `-pedantic` flag:
> 
> ```
> liblammpsplugin.c:65:36: warning: ISO C forbids assignment between function pointer and 'void *' [-Wpedantic]
>  #define ADDSYM(symbol) lmp->symbol = dlsym(handle,"lammps_" #symbol)
> ```

Yes, it does. If you want that to pass (and compilation with C++) you have to get rid of the ADDSYM macro and write some script/program that generates the explicit typecasts for all function pointer assignments instead.

stanmoore1:
@akohlmey this change removes the ISO C warning and also compiles with C++:

```patch
diff --git a/examples/COUPLE/plugin/liblammpsplugin.c b/examples/COUPLE/plugin/liblammpsplugin.c
index 1350a16..deb8d3a 100644
--- a/examples/COUPLE/plugin/liblammpsplugin.c
+++ b/examples/COUPLE/plugin/liblammpsplugin.c
@@ -62,7 +62,7 @@ liblammpsplugin_t *liblammpsplugin_load(const char *lib)
 #ifdef _WIN32
 #define ADDSYM(symbol) lmp->symbol = (void *) GetProcAddress((HINSTANCE) handle, "lammps_" #symbol)
 #else
-#define ADDSYM(symbol) lmp->symbol = dlsym(handle,"lammps_" #symbol)
+#define ADDSYM(symbol) *(void **) (&lmp->symbol) = dlsym(handle,"lammps_" #symbol)
 #endif
 
 #if defined(LAMMPS_LIB_MPI)
```

stanmoore1:
@akohlmey looks great, I approve, thanks

PR #3507, opened 2022-10-30
Thermo_modify format ranges

hammondkd:
**Summary**

I made minor modifications to `thermo.cpp` so that the ``thermo_modify format`` command now accepts ranges. For example,
```
thermo_style custom step temp press etotal c_1[1*20]
thermo_modify format 5* "%5f"
```
will output the first 20 elements of compute ID `1` (this is existing functionality), but the "5*" (equivalent in this case to "5*24") format ID in ``thermo_modify`` will now cause fields 5 through 24 to have "%8f" as their format specifier. This makes it easier to format multiple very similar columns without reformatting all floats that way: previously, the user would have to specify either all floats at once, then override the ones that the user wanted to be different, or specify the 20 that the user wanted to be different in different "format" lines.

**Author(s)**

Karl Hammmond, University of Missouri
hammondkd at missouri.edu
@hammondkd

**Licensing**

**Backward Compatibility**

This should not affect backward compatibility (and I performed a few checks to ensure that it does not).

**Implementation Notes**

This existing utility functions to check formatting (is_integer and is_type). The is_type function currently does not check whether the range given constitutes a valid type, it just checks that it's a valid integer range specifier; if this functionality ever changes, this routine would need to be modified.

This has been checked that existing functionality isn't disrupted (as far as I can tell). I did not write a separate unit test to check this new feature, but there isn't one dedicated to ``thermo_modify format`` that I see anyway (and if there is one, I'll add a quick test for this feature). There *are* documentation updates included.

**Post Submission Checklist**


hammondkd:
Just to make sure this gets documented somewhere: my original implementation used `not`, `and`, and `or` in thermo.cpp, but the Windows build didn't work. I changed them to `!`, `&&`, and `||` instead and it worked. I'm pretty sure that's not supposed to happen (i.e., the C++ standard calls them synonyms, or so I thought), but there it is all the same.

sjplimp (approved):

akohlmey:
@hammondkd support for negative ranges (`-2*6` *not* `-2*-6`) could be added with something like the following:
```
      } else if (utils::strmatch(arg[iarg + 1], "^-?\\d*\\*\\d*$")) {
        int nlo, nhi, neg=0;
        if (arg[iarg][0] == '-') neg = 1;
        utils::bounds(FLERR, arg[iarg + 1] + neg, 1, nfield_initial, nlo, nhi, error);
        if (neg) {
          nlo = -nlo;
          nhi = -nhi;
        }
        int icol = -1;
        [...]
```

sjplimp:
> @hammondkd support for negative ranges (`-2*6` _not_ `-2*-6`) could be added with something like the following:
>       } else if (utils::strmatch(arg[iarg + 1], "^-?\\d*\\*\\d*$")) {
>         int nlo, nhi, neg=0;
>         if (arg[iarg][0] == '-') neg = 1;
>         utils::bounds(FLERR, arg[iarg + 1] + neg, 1, nfield_initial, nlo, nhi, error);
>         if (neg) {
>           nlo = -nlo;
>           nhi = -nhi;
>         }
>         int icol = -1;
>         [...]

Did this get added to the merge?  I'm hoping not.

akohlmey:
> Did this get added to the merge? I'm hoping not.

no.

PR #3512, opened 2022-11-01
Refactoring the GRANULAR package

### LINKED ISSUE #3022: [BUG] Neighbor history overflow in restart granular pairstyles
**Summary**

When a restart file is written via the "write_restart" command with a granular pair style and Newton on, it can trigger the error "Neighbor history overflow, boost neigh_modify one" in the `pre_exchange()` method of `FixNeighHistory`. This occurs because WriteRestart moves atoms to new processors, changing values of `atom->nghost` and `atom->nlocal`, before calling `FixNeighHistory->write_restart()`. However, in `FixNeighHistory` these values are stored after every neighbor list build as `nlocal_neigh` and `nall_neigh`. Since these variables are out of date, `FixNeighHistory` fails to zero certain elements of the npartner array (lines 354 and 409). When the total number of contacts is then tallied in `FixNeighHistory` using a reverse comm call, bad values of `npartner` are packed.

Fundamentally, this issue occurs because LAMMPS never runs `FixNeighHistory->post_neighbor()` between `comm->exchange()/comm->borders()` in `WriteRestart` and `FixNeighHistory->pre_exchange()` in `FixNeighHistory->write_restart()`. I'm tempted to say this could be fixed by simply running `post_neighbor()` before `pre_exchange()` in `FixNeighHistory->write_restart()`, but I haven't confirmed it. I'll try to test it, but I'm a little puzzled by why `nlocal_neigh` and `nall_neigh` are needed in the first place. 

Note that regular 'restart' commands work fine since LAMMPS rebuilds the neighbor list in anticipation of the restart file being written. So it's easy to work around the issue but LAMMPS should at least issue a warning.

LAMMPS stable release on 29 Sep 2021

**Expected Behavior**

LAMMPS should correctly identify the number of neighbors and not error.

**Actual Behavior**

LAMMPS errors despite there being plenty of room in page files for the neighbor list and fails to write the restart file.

**Steps to Reproduce**

This bug is unfortunately (or maybe fortunately) hard to trigger but I included an input script and restart file for a moderately sized system (100k atoms) which can reproduce it if run on 8 processors. The restart file does contain some information from custom fixes, but they are unused and irrelevant to the bug.

[example.zip](https://github.com/lammps/lammps/files/7489420/example.zip)

jtclemm:
**Summary**

To modularize the GRANULAR package, we moved force/torque calculations from pair_granular.cpp and fix_gran_wall.cpp into a new GranularModel class which handles the details of interaction forces. Force calculations within the GranularModel class are then further broken down into a set of granular submodel classes (GSM) which each define a specific normal, damping, tangential, rolling, and sliding contact model. Adding a new contact model now just requires a new GSM class.

These changes should have no effect on input scripts although simulation results might vary due to:
1) Some changes in the order of operations will accumulate floating point errors
2) A few bug fixes involving memory errors or typos in algebra

This pull request also includes a new heat conduction model for the granular package which uses a new atom style (sphere/temp) and integrator (temp/integrate).

Minor changes were made to fix neigh/history. Non-default transfers now have a type dependence and the npartner array is fully zeroed before counting partners to resolve issue https://github.com/lammps/lammps/issues/3022

**Related Issue(s)**

Fixes #3022 

**Author(s)**

Joel Clemmer (SNL)
Dan Bolintineanu (SNL)

**Licensing**

**Backward Compatibility**

Old restart files will no longer work.

**Implementation Notes**

Aside from creating an atom array for temperature and the rate of heat flow, changes were localized to the GRANULAR package and fix neigh/history.

**Post Submission Checklist**

NA


jtclemm:
This pull request is mostly finished and ready for reviewer feedback. Dan and I mostly just need to check a block of algebra.

akohlmey:
FWIW, this PR fails currently because pair style tracker in the MISC package has not been updated.

sjplimp:
@jtclemm and @dsbolin - Thanks for submitting this - looks like a more flexible refactoring.  I haven't yet looked thru the code details but 2 Qs:
(1) This appears not to affect the older gran/hooke, gran/hooke/history, etc styles.  But it does affect fix wall/region.  Does that mean you can use fix wall/region with both the old and new granular pair styles?
(2) What is the effect on run speed of pulling the force kernel out of the pair style and its innermost double loop over atoms & neighbors?  E.g. if you run the standard granalar benchmark in the bench dir, is it faster with the old style vs the refactored pair granular with the separate force class?

jtclemm:
Hi Steve, thanks for taking a look at the changes. Regarding your questions:

> (1) This appears not to affect the older gran/hooke, gran/hooke/history, etc styles. But it does affect fix wall/region. Does that mean you can use fix wall/region with both the old and new granular pair styles? 

It does not affect the old pair gran/X styles but does affect fix wall/gran. In the new granular force class, we added a method to construct a "classic" model so you can still use the old and new granular pair style models with fix wall/gran.

> (2) What is the effect on run speed of pulling the force kernel out of the pair style and its innermost double loop over atoms & neighbors? E.g. if you run the standard granalar benchmark in the bench dir, is it faster with the old style vs the refactored pair granular with the separate force class?

I've run a few benchmark tests and the relative performance depended on the force models. Specifically testing the benchmark problem in LAMMPS and comparing average pair times, I first see a 50% increase going from pair gran/hooke/history to the equivalent settings in the current pair granular. I then see an additional 25% increase going from the current pair granular to this refactored pair granular. So as expected it is slower, but it's not as big of a jump as going from pair gran -> granular.

As a quick test, I tried passing the neighbor list to the granular force class and letting it perform the loop. In that case, I only found a slow down of a few percent (if any).

src/GRANULAR/gsm.h
```diff
@@ -0,0 +1,62 @@
...
+
+   Copyright (2003) Sandia Corporation.  Under the terms of Contract
+   DE-AC04-94AL85000 with Sandia Corporation, the U.S. Government retains
+   certain rights in this software.  This software is distributed under
+   the GNU General Public License.
+
+   See the README file in the top-level LAMMPS directory.
+------------------------------------------------------------------------- */
+
+#ifndef LMP_GSM_H_
```
akohlmey: Include guards should always be `LMP_AAA_BBB_E` if the file name is `aaa_bbb.h`

What worries me is the use of the "gsm" initialism. While it is certainly convenient, it would be better to have something "longer" and more verbose/specific instead as that reduces the risk that there is an include file with the same name elsewhere. We already had some problems in the past with, e.g. "memory.h" and "atom.h" and a few other "obvious" names.

Since this is not something that a user will have to enter, it should not be a big problem to use something like GranSubModel or GranSubM as base class name and thus `gran_sub_model.h` or `gran_subm.h` as corresponding file name. I am open for alternate suggestions.

src/GRANULAR/fix_temp_integrate.cpp
```diff
@@ -0,0 +1,136 @@
...
+      if (cp_type[i] < 0.0) error->all(FLERR,"Illegal fix command");
+    }
+  } else {
+    error->all(FLERR,"Illegal fix command");
+  }
+
+  if (cp_style == NONE)
+    error->all(FLERR, "Must specify specific heat in fix temp/integrate");
+  dynamic_group_allow = 1;
+  time_integrate = 1;
```
akohlmey: I have some formal concerns about this fix.
- It is flagged as doing time integration, but particle positions are unchanged. If a position update is done with a different fix, there will be warnings about atoms being time integrated twice. But that is not really true in this case.
- All fixes that have names starting with "temp/" are thermostats, so having the "temp/integrate" name is in conflict with that convention
- For time integration fixes that are specific to some atoms styles, we have a convention to use something like fix nve/atom-style (see. fix nve/sphere, fix nvt/asphere, fix nve/bpm/sphere etc.) so this fix's name should end in sphere/temp.

So something like `fix heat/flow/sphere/temp` or `fix heat/transport/sphere/temp` could be more in line with the existing conventions.

src/GRANULAR/granular_model.h
```diff
@@ -0,0 +1,114 @@
...
+
+#ifndef LMP_GRANULAR_MODEL_H
+#define LMP_GRANULAR_MODEL_H
+
+#include "pointers.h"    // IWYU pragma: export
+
+namespace LAMMPS_NS {
+namespace Granular_NS {
+
+#define EPSILON 1e-10
```
akohlmey: This either should not be a define but a `static constexpr double` type or not in the header, but only in the .cpp files.

src/GRANULAR/pair_granular.h
```diff
@@ -20,7 +20,9 @@ PairStyle(granular,PairGranular);
 #ifndef LMP_PAIR_GRANULAR_H
 #define LMP_PAIR_GRANULAR_H
 
+#include "granular_model.h"
```
akohlmey: Including this header should not be needed here.
You should be able to get away with a forward declaration like
```
namespace GRANULAR_NS {
   class GranularModel;
}
```
right after `namespace LAMMPS_NS {`

src/GRANULAR/gsm.h
jtclemm: Originally, Dan and I debated fully spelling out GranularSubmodel to avoid the acronym. But I much prefer your shortened suggestions as an in-between. I'm happy with any of your suggestions or GranSubMod. I can make the changes.

src/GRANULAR/granular_model.h
jtclemm: Looking at this again, I don't think EPSILON should even be defined at this level but rather in the specific submodel files its used. Presumably one could define a separate force model that uses an epsilon-like value other than 1e-10. I'll move it to gsm_rolling.cpp and gsm_tangential.cpp.

src/GRANULAR/pair_granular.h
jtclemm: Good point, I'll make the change.

src/GRANULAR/fix_temp_integrate.cpp
jtclemm: I'll swap it to one of those names and fix the incorrect flag. Thanks for pointing those out!

src/GRANULAR/gsm.h
akohlmey: I don't have a strong feeling about any of those. Either will work for me as either will address the concern.

jtclemm: I just pushed changes which should include all of these changes. For the include guard I have something like:
`#ifndef LMP_GRAN_SUB_MOD_H_`
for gran_sub_mod.h. Is this the correct format? Your example ended in `_E_` but I wasn't sure whether or not that was a typo.

akohlmey: Yes, this was a typo. It was supposed to be `LMP_AAA_BBB_H` (note no trailing underscore).

Sorry about being so anal retentive about such minor details, but having consistent names for these can be a massive time saver when we have to make bulk changes, as those can be automated for all conforming files. I have been through such processes a few times and finding and fixing the non-conforming files has always been a bit of a hassle. :-(

jtclemm: Ah, I totally overlooked the trailing underscore. I'll prune those. It also looks like the styles aren't being found so I'll look at that again and patch it.

Thanks!

akohlmey: > It also looks like the styles aren't being found so I'll look at that again and patch it.

Did you change `cmake/Modules/StyleHeaderUtils.cmake` and `src/Make.sh` as needed for the change in file names?

jtclemm: Yep, I missed some changes to the cmake style utils file. Think it's fixed now.

src/GRANULAR/granular_model.cpp
```diff
@@ -0,0 +1,493 @@
...
+  scaleadd3(radi, omegai, radj, omegaj, wr);
+
+  // relative tangential velocities
+  double temp[3];
+  cross3(wr, nx, temp);
+  sub3(vt, temp, vtr);
+  vrel = len3(vtr);
+
+  // calculate forces/torques
+  forces[0] = 0.0;
```
akohlmey: Is this supposed to clear out the entire "forces" array? Or just the first element?

akohlmey:
FYI, the documentation needs some more updates for changed names:
```
[...]
updating environment: [new config] 951 added, 0 changed, 0 removed
reading sources... [100%] region .. write_restart                                                                                                                                               
/home/akohlmey/compile/lammps/doc/src/fix_heat_flow_sphere_temp.rst:4: WARNING: Title underline too short.

fix heat/flow/sphere/temp command
==========================
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
/home/akohlmey/compile/lammps/doc/src/Commands_fix.rst:238: WARNING: unknown document: 'fix_temp_integrate'
/home/akohlmey/compile/lammps/doc/src/Howto_granular.rst:46: WARNING: unknown document: 'fix_temp_integrate'                                                                                    
/home/akohlmey/compile/lammps/doc/src/Howto_granular.rst:67: WARNING: unknown document: 'Modify_gsm'
/home/akohlmey/compile/lammps/doc/src/fix.rst:385: WARNING: unknown document: 'fix_temp_integrate'                                                                                              
/home/akohlmey/compile/lammps/doc/src/pair_granular.rst:661: WARNING: unknown document: 'fix_temp_integrate' 
[...]
Total number of styles (including suffixes): 1646
Total number of style index entries: 1210
Fix style entry heat/flow/sphere/temp is missing or incomplete in Commands_fix.rst
Fix style entry heat/flow/sphere/temp is missing or incomplete in fix.rst
```

src/GRANULAR/granular_model.cpp
jtclemm: No, that looks like an overlooked line that should have been deleted earlier. Removed now.

jtclemm:
> FYI, the documentation needs some more updates for changed names:

Thank you for the reminder. I keep forgetting to update all the higher level doc pages.

jtclemm:
Dan and I just finished checking the last few sections of algebra (some was amended) so there are no more features we plan to add to this PR. 

I'm not entirely sure what caused some checks to fail - seems like some issues with python for windows?

akohlmey:
> I'm not entirely sure what caused some checks to fail - seems like some issues with python for windows?

This is due to some (incompatible) software changes in the Azure Windows instances used by GitHub actions.
I have cherry-picked the change to the workflow file that should address it from a different PR.

We are currently a bit backed up with merging pull requests and the one that has this change is still pending.

akohlmey:
@jtclemm automated integration and unit tests pass now.
Have you checked whether your changes need to be ported to KOKKOS?

jtclemm:
> Have you checked whether your changes need to be ported to KOKKOS?

At some point, it will be good to port the granular pair style to KOKKOS. But currently only pair gran/hooke and maybe gran/hertz has KOKKOS support which weren't affected by these changes.

akohlmey:
> > Have you checked whether your changes need to be ported to KOKKOS?
> 
> At some point, it will be good to port the granular pair style to KOKKOS. But currently only pair gran/hooke and maybe gran/hertz has KOKKOS support which weren't affected by these changes.

This PR has changes to `src/fix_neigh_history.cpp` which has a KOKKOS version. Those two should behave consistently, even if you have no pair style granular/kk yet.

jtclemm:
> This PR has changes to `src/fix_neigh_history.cpp` which has a KOKKOS version. Those two should behave consistently, even if you have no pair style granular/kk yet.

The important changes to fix neigh/history were to the nondefault transfer method, but the KOKKOS version does not yet have a nondefault transfer method. This is something that will need to be addressed before porting pair granular.

apsantos:
The fix temp/integrate should be updated in `examples/granular/in.pour.heat` from 
`fix             2 all temp/integrate constant 1.0`
to
`fix             2 all heat/flow/sphere/temp constant 1.0`

I know you all are still working on this PR, and would have caught it anyway. Thanks!

jtclemm:
> The fix temp/integrate should be updated in `examples/granular/in.pour.heat` from `fix 2 all temp/integrate constant 1.0` to `fix 2 all heat/flow/sphere/temp constant 1.0`
> 
> I know you all are still working on this PR, and would have caught it anyway. Thanks!

Thanks for spotting that, should be fixed now.

akohlmey:
@jtclemm this looks like it should be updated to the latest changes from upstream and then reviewers need to be reminded to re-review it.

jtclemm:
> @jtclemm this looks like it should be updated to the latest changes from upstream and then reviewers need to be reminded to re-review it.

Thank you for the reminder. Just updated.

@sjplimp, do you have any opinions on the pros/cons of this reorganization of pair granular? As mentioned above, these changes do increase the pair time by about 25% compared to the original pair granular. For context, the original pair granular already increased runtime 50% from pair gran/hooke[hertz].  

While I typically prioritize performance, in this case I think the revised code is worth the costs since its much easier to ensure models are accurately implemented and don't conflict with others (particularly given the number of subtle bugs we just discovered in the original pair granular). Although it isn't an option for every user, if an application demanded performance it's relatively easy to copy the force calculations from a specific set of sub classes directly into a pair style to create a customized pair style even faster than the original pair granular.

jtclemm:
I emailed with Steve about the new atom style and decided to alternatively use fix property/atom to add the new variables as needed. This will simplify any future efforts to expand heat conduction beyond the GRANULAR package, in particular the BPM package.

@sjplimp  Please let me know whether or not the changes in the latest commit are consistent with your understanding of our discussion.
@stanmoore1 This should be ready for a final review.

Thanks again!

sjplimp (approved):

sjplimp:
@jtclemm Nice job - looks comprehensive to me.  The new Modify gran sub mod doc page seems particularly useful.

PR #3514, opened 2022-11-05
bond/react feature updates

jrgissing:
**Summary**

Several new features added to fix bond/react: 
  1) Support for per-bond variables for custom constraint: 
       - For example, allows bond force to determine when reactions occur
  2) _rate_limit_ keyword: enforce a reaction rate upper limit for each reaction
  3) _rescale_charges_ keyword: ensure the total charge of reaction site remains constant

Thanks to Sam Root (Stanford), Emily Lin (UT Austin), and Benjamin Jensen (NASA Langley) for these suggestions

**Related Issue(s)**

none

**Author(s)**

JG

**Licensing**

**Backward Compatibility**

yes!

**Implementation Notes**

**Post Submission Checklist**


src/compute_bond_local.cpp
```diff
@@ -390,6 +391,11 @@ int ComputeBondLocal::compute_bonds(int flag)
           engrot *= mvv2e;
         }
 
+        // populate map to access per-bond values from atom tags
```
akohlmey: I don't like this change (including the change to compute.h). It provides no functionality for compute bond/local and is not controlled by some flag (and thus always consumes resources whether it is used or not) and whoever needs to use it needs to set up compute bond/local in a suitable fashion so that the data is computed. That is rather indirect and difficult to understand from the source code. Why not make computation of the map a self-contained function in fix bond/react?

jrgissing: good points. yes, that should work

akohlmey: Thanks for making the change.

jrgissing:
I am thinking of making this PR more backward compatible using the lmprestart revision number, which would enable it to be backwards compatible for all stable releases except the most recent, when the revision number was updated. I might be able to do better if the LAMMPS version of the restart is also remembered and made accessible. Any thoughts on if implementing that would be acceptable?

akohlmey:
> Any thoughts on if implementing that would be acceptable?

I think that would be useful. The file revision is just for stopping if the file format has changed and thus an old version file cannot be parsed without crashing or getting into an inconsistent state. Storing the specific version would allow individual commands and restart functions to implement more fine grained backward compatibility.
I have thus implemented making the version number (as integer so it is easier to make comparisons) in this commit https://github.com/lammps/lammps/pull/3516/commits/4392b9c8cbc0c2b09827074509e9cbc9419d86a8

jrgissing:
> I I have thus implemented making the version number (as integer so it is easier to make comparisons) 

Great! I will wait for that to be merged before updating this PR

akohlmey:
> > I I have thus implemented making the version number (as integer so it is easier to make comparisons)
> 
> Great! I will wait for that to be merged before updating this PR

It is merged now.

jrgissing:
> It is merged now.

Thanks! One minor issue, is that this development branch will be 'incorrect' until the version number is stepped past the last patch (because I am comparing against that patch version for backward compatibility). would it make sense to increment the version with the first commit after a patch release?

akohlmey:
> > It is merged now.
> 
> Thanks! One minor issue, is that this development branch will be 'incorrect' until the version number is stepped past the last patch (because I am comparing against that patch version for backward compatibility). would it make sense to increment the version with the first commit after a patch release?

Well, either choice would be "incorrect" in that sense. Strictly speaking, you would have to be able to tell which is before and what is after the merge commit. This is something that could be done with subversion as a SCM, but I know nothing else.
We have established a convention that the pull request changing the version string is going to be the *last* PR merged before a  release.

BTW: for added/changed features you can add in the documentation for fix bond/react `.. versionadded:: TBD` or `.. versionchanged:: TBD` and then in the version update PR, those TBDs will be replaced by the actual version date. Since LAMMPS uses the date and not a major.minor numbering scheme we don't really know up front what the date will be.

I've been also retro-actively adding such version markers to the documentation to make it more obvious to people reading the online docs while running an older version of LAMMPS, which parts are relevant to them. More changes of that kind are thus much appreciated.

jrgissing:
Okay. To clarify, by incorrect I also mean that the new code not testable without other uncommitted changes, such as incrementing the current LAMMPS version by at least a day. Otherwise, reading the new restart format is not triggered. not a problem for me, but I imagine this could cause confusion if other people use this new lmp->restart_ver value, especially if it gets worked into unit tests

> BTW: for added/changed features you can add in the documentation for fix bond/react .. versionadded:: TBD or .. versionchanged:: TBD

thanks for pointing this out, I'll add some of those version tags

akohlmey:
> Okay. To clarify, by incorrect I also mean that the new code not testable without other uncommitted changes, such as incrementing the current LAMMPS version by at least a day. Otherwise, reading the new restart format is not triggered. not a problem for me, but I imagine this could cause confusion if other people use this new lmp->restart_ver value, especially if it gets worked into unit tests

We don't test against restart files created by other than the current LAMMPS version. In general, restarts are only guaranteed to be compatible with the same version. You can add a more fine grained control into your code, by adding a "restart_revision" number into the section of the restart written and read by your own fix. Best as the first item. every time you add a new data entry or section to the restart, you increment that number, and upon reading, you can adjust accordingly.

If you change the restart code of your fix so that the fix specific restart revision number is written and read only when the LAMMPS restart_ver is > 20221103 and the default is the value compatible with the current format, then you will have control over backward compatibility once the next LAMMPS patch release is out.

jrgissing:
thanks, that makes sense

stanmoore1 (approved):

PR #3516, opened 2022-11-07

akohlmey:
**Summary**

**Related Issue(s)**

Addresses issues reported in https://matsci.org/t/problems-with-building-with-adios2/45328 and https://matsci.org/t/problems-with-adios-dump-style/45327/9

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

- add -y flag to add-apt-repository commands in apptainer definition files for ubuntu to avoid them stopping for input when confirming outdated gpg signing
- update .github/CODEOWNERS file: add/change a few entries and sort list of assigned packages/folders
- remove obsolete FindZMQ.cmake file
- add support to build cross-compiled Windows binaries against MS-MPI (only 64-bit) with -DUSE_MSMPI=on (default is off). Eventually support for the old MPICH2 package and 32-bit will be removed.
- install MS-MPI into GitHub Action for Windows via nuget, so it will test native compilation with MPI
- add check to CMake support for ADIOS to check ADIOS configuration for MPI support. Error out if not compatible with MPI support in LAMMPS
- add support for changed APIs (w/o MPI communicator argument) when compiling ADIOS package with -DBUILD_MPI=no
- store LAMMPS version id of restart file when initialized from restart to allow more fine grained backward compatibility of individual commands.
- silence compiler warnings

**Post Submission Checklist**


akohlmey:
@pnorbert please check out the changes to the ADIOS package support. This will a) check if the ADIOS2 library has been configured with/without MPI as needed and b) adjust the calls to ADIOS according to when MPI is used in LAMMPS and when not (i.e the MPI stub library is used which defines MPI_STUBS when its header is included).

The second fix is to skip initialization if the input has multiple "run" or "minimize" commands. Without this change, there is an exception inside ADIOS.

sjplimp (approved):

sjplimp:
@akohlmey You mention silence compiler warnings.  With my g++ 11.3 (actually mpicc) I see this warning:

./variable.cpp: In member function ‘int LAMMPS_NS::Variable::next(int, char**)’:
../variable.cpp:760:14: warning: ignoring return value of ‘size_t fread(void*, size_t, size_t, FILE*)’ declared with attribute ‘warn_unused_result’ [-Wunused-result]
  760 |         fread(buf,1,64,fp);

This is with no packages installed.  I think I've seen similar warnings with other packages
installed, but I can't recall which ones.  All related to funcs like fread().

akohlmey:
> @akohlmey You mention silence compiler warnings. With my g++ 11.3 (actually mpicc) I see this warning:

Yes, I see it too, when I turn on extra warnings or compile on Ubuntu. Ubuntu has some "special" default settings when compiling source code that adds GCC specific "decorations" to some subset of functions. This includes functions like `fread()` and `fwrite()` where the Ubuntu developers/maintainers consider it good practice to check if the read or write was successful or incomplete or caused an error.

For the cases where it matters, we have mostly replaced these calls with functions like `utils::sfread()` or at least captured the return value. This is an ongoing process and needs to be regularly applied to newly added or changed code. Sometimes, like in the change in the AMOEBA package in this PR, the warnings indicate some unused code and then it is a question of either removing this "dead code" or making it more obvious why this is (not yet) used and more consistently so. Sometimes there are bugs. Sometimes there is nothing wrong.

If you want to see some more warnings, you can compile LAMMPS using (in the main LAMMPS folder):
```
cmake -S cmake -B build-warnings -C cmake/presets/gcc.cmake -C cmake/presets/most.cmake -D PKG_AWPMD=on -D PKG_ML-RANN=on -D PKG_MGPT=on -D PKG_SMTBQ=on 
cmake --build build-warnings -j16
```
Some of those still need to be addressed or are indications of code using obsolescent constructs (POEMS is an example).

Not for all warnings is it meaningful to silence them. Best example for that is the "variable maybe uninitialized" warning in GCC. It is easy to silence those, but doing so by enforcing initialization would make it **much** harder to find initialization errors with valgrind, so I prefer to turn that warning off and ignore it rather than following the suggestion from the compiler warning.

Overall, this is a fairly interesting topic (at least for me) since it has helped us - together with a variety of static code analysis tools like coverity scan or clang-tidy - to identify bugs or bad programming practices and replace them with improved code and thus have significantly improved the code quality in LAMMPS over the years since we started doing it. At the beginning the number of warnings was *HUGE*.

PR #3524, opened 2022-11-15
Fix segfault when using dynamic groups with r-RESPA

ssande7:
**Summary**

`run_style respa` segfaults when using dynamic groups. This is caused by the loop over `modify->list_post_force_respa` iterating to `modify->n_post_force_respa_any` instead of just `modify->n_post_force_respa` (the former also counts the number of `FixGroup` styles).

Minimal example:
```
units		lj
atom_style	atomic

lattice		fcc 0.8
region		box block 0 10 0 10 0 10
create_box	1 box
create_atoms	1 box
mass		1 1.0

velocity	all create 1.0 87287

pair_style	lj/cut 2.5
pair_coeff	1 1 1.0 1.0 2.5

neighbor	0.3 bin
neigh_modify	every 20 delay 0 check no

fix		1 all nve

# This line causes a segfault
group all_dyn dynamic all region box

run_style respa 2 2
run		1

```

**Related Issue(s)**

None

**Author(s)**

Stephen Sanderson (UQ)

**Licensing**

**Backward Compatibility**

No issues expected.

**Implementation Notes**

I have assumed that only user-defined fixes, and not FixGroup fixes, should be printed during r-RESPA setup. If this is not the case, a separate loop over `modify->list_post_force_group` would also be needed.

**Post Submission Checklist**

### DIFF
--- a/src/respa.cpp
+++ b/src/respa.cpp
@@ -372,7 +372,7 @@ void Respa::setup(int flag)
         mesg += fmt::format(" {}:{}", ilevel + 1, step[ilevel]);
 
       mesg += "\n  r-RESPA fixes :";
-      for (int l = 0; l < modify->n_post_force_respa_any; ++l) {
+      for (int l = 0; l < modify->n_post_force_respa; ++l) {
         Fix *f = modify->get_fix_by_index(modify->list_post_force_respa[l]);
         if (f->respa_level >= 0)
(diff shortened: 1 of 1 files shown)


akohlmey:
@ssande7 Thanks for your submission. I have reviewed your change and agree that it is the correct bugfix.

stanmoore1 (approved):
I approve.

PR #3527, opened 2022-11-16

akohlmey:
**Summary**

This pull request collects multiple small changes and bug fixes.

**Related Issue(s)**

It expands on the changes in PR #3525 and includes PRs #3524 and #3529

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

Yes. Any "box" command in the input will be ignored and a warning printed that the command was removed.

**Implementation Notes**

- enhancement of the pair style suffix handling in compute fep to also handle hybrid suffixes
- small programming style adjustments to fix bond/react
- allow a much wider range of valid values for bond and angle style gaussian before truncation is applied
- improve/correct/clarify documentation for bond/angle style gaussian
- fix bugs in `pair_coeff` parsing of pair styles nm/cut/coul/cut and nm/cut/coul/long (reported by Bertrand Siboulet, CEA)
- remove "box" command; print warning instead of hard error when initializing simulation box with large tilt.
- add keywords to allow querying if Kokkos is active and how many threads/gpus are configured via the library interface
- reduce redundant code in `Domain::minimum_image()` and add (fast) heuristic check to error out if atoms have moved apart too far and thus the code would get "stuck" finding the minimum image distance (motivated by problem reported in https://matsci.org/t/simulation-failed-to-finish-or-produce-an-error-message/45440)
- add a few more versionadded/versionchanged tags to the documentation
- symlink data files in FEP examples where needed
- fix spelling issues in fix bond/react docs
- fix unitialized data issues in pair style pace/extrapolation

**Post Submission Checklist**


src/domain.cpp
```diff
@@ -984,22 +984,34 @@ void Domain::subbox_too_small_check(double thresh)
      use minimum_image_once() instead
 ------------------------------------------------------------------------- */
 
+#ifdef LAMMPS_BIGBIG
+static constexpr double MAXIMGCOUNT = 1<<21;
```
sjplimp: Why not make MAXIMGCOUNT much smaller, e.g. 1024 or even 4.
Are there cases when a value > 4 between a pair of atoms is not
an indication of a bad dynamics ?

akohlmey: Fine with me.

I made a conservative choice by looking at the maximum possible distance due to image flags. Anything beyond that is not representable in LAMMPS.

> Are there cases when a value > 4 between a pair of atoms is not an indication of a bad dynamics ?

I was trying to come up with the worst possible case, e.g. people that put together systems from individually equilibrated data files and then try to cross-link monomers into polymers.

sjplimp: let's use 16
the comment for the method indicates it can be needed when the cutoff > boxsize,
  16 seems a reasonable limit on that

sjplimp: also in the comment header for this method, on the last line can you
change "use" to "it uses" and indent it another 2 spaces
it is referring to fix_shake using minimum_image_once()

akohlmey: @sjplimp updates are made. Please also note the new "reset_image_flags" command, which should help folks to reset their systems so that the image flags are zero or close it it. A good test is the "peptide" example input which has water molecules that have passed through the box multiple times. This resets them so that at least one atom of a water has all zero image flags.

akohlmey: The command is nearly a macro, since most of the work is done by several existing computes and variables.

sjplimp: Useful functionality but I'm not keen on adding new commands if we can avoid it.  Doesn't have to be in this PR,
but what if we combined reset_atom_ids, reset_mol_ids, reset_image_flags into a single reset command.  Similar to
"set" it (re)sets attributes of atoms.  Reset_timestep could stay, b/c it is widely used and is not resetting atom properties.

akohlmey: OK. I will take the commits out of this PR and create a new one. While we are looking at combining commands, could we perhaps add the "box" command functionality somewhere, too? It has only one option and that is a bit obscure. I only had to use it once recently when `replicate` on a triclinic box mysteriously triggered a "too large tilt" error.

sjplimp (approved):

PR #3529, opened 2022-11-17
correct a grammatical error

mehdibghk:


akohlmey:
There is no need to submit a whole pull request for such a trivial change. You could just send an email to the LAMMPS developers at developers@lammps.org or add a comment to a "collection pull request" like PR #3527.
I have pulled your branch into the later and will thus close this one now.

PR #3530, opened 2022-11-18
Add reset_atoms meta-command and reset_atoms image command

akohlmey:
**Summary**

This adds a new command `reset_atoms image` and converts `reset_atom_ids`  into `reset_atoms id` as well as `reset_mol_ids` into `reset_atoms mol`.

**Related Issue(s)**

Spun off from PR #3527

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

Yes, but deprecation warning is printed when using `reset_atoms <sub-command>`.

**Implementation Notes**

The reset of image flags strictly follows the bond topology, even if there are multiple molecule IDs for a fragment.
On the other hand, it also strictly only applies changes to atoms in the selected group.
In addition to functional changes, clang-format is enabled and applied for all affected reset sub-commands.

**Post Submission Checklist**


sjplimp:
I thought about this some more - here is what I suggest:
(1) Add "box" command functionality to "change_box" command as a new option, remove "box" command
(2) Add a "change_atoms" command with 3 options: id, mol, image, instead of a "reset_atoms" command
This way we have create_box/atoms, displace_box/atoms, change_box/atoms commands.
And id, mol, image are the same keywords used in other commands.
What do you think?

akohlmey:
> I thought about this some more - here is what I suggest: (1) Add "box" command functionality to "change_box" command as a new option, remove "box" command

I'll take a look.

> (2) Add a "change_atoms" command with 3 options: id, mol, image, instead of a "reset_atoms" command This way we have create_box/atoms, displace_box/atoms, change_box/atoms commands. And id, mol, image are the same keywords used in other commands. What do you think?

I see the appeal of this change because of the symmetry, but semantically the "set" command resembles much more closely what I would expect from a "change_atoms" command than these "reset" sub-commands. With "change_<item>" you expect to change the item according to some input data (which is what happens with change_box) while the three "reset" sub-commands don't accept much of an input, but rather "recompute" the properties in question (and dependent properties, if needed) according to some algorithm. So while "reset" is not a perfect choice, it feels closer than "change_atoms".  Another motivation for leaving things as it is currently in the PR is that the change in syntax is rather small and easy to memorize (and it is easy to maintain backward compatibility for some time).  Similarly, it requires only small changes to the documentation. I think that has some appeal, too, and for me personally, that one is stronger.

akohlmey:
@athomps @stanmoore1 
It looks like @sjplimp and I cannot find a compromise that works for both of us. We do agree on subsuming multiple commands into a "meta" command with sub-commands, but not on the name. Arguments are here and in PR #3527

Do you have preferences or objections on using "change_atoms" or "reset" as prefix (which this PR implements)?

akohlmey:
> (1) Add "box" command functionality to "change_box" command as a new option, remove "box" command

@sjplimp I've looked into merging the "box" and "change_box" commands. It is a bit more complex than anticipated. "box" can *only* be used *before* a box is created and "change_box" only *after*.  So merging those will result in one command with two different "modes" and keywords/syntax depending on whether it is used before or after a box is created. That can result in a bit of confusion with easily confused users.

So I am wondering, perhaps it would be simpler to remove the need for the "box" command altogether and turn the hard error when tilting the box too far into a warning and implement it so that it will only be printed once per run.

sjplimp:
Hmm.  My worry is users will not see the warning and be unaware they are running with a very inefficient geometry.
Remind me the various ways the box can exceed the tilt limit?

sjplimp:
If it is only an issue when a box is created, via read_data or create_box, then I suggest adding an option to those commanda which the user is required to specify to allow a large-tilt box.   But what does fix deform do if the box is a very large tilt to start with?  Ditto for read_restart?

akohlmey:
> Hmm. My worry is users will not see the warning and be unaware they are running with a very inefficient geometry. 

In my experience, the users that may overlook the warning are the same that will just blindly use the "box" command to avoid the error. So my expectation is that the net effect is the same.

> Remind me the various ways the box can exceed the tilt limit?

The only check for the flag set by the "box" command is in `Domain::set_initial_box()`. This function is called from "change_box", "create_box", "read_data", "read_dump", "read_restart", "replicate", and "fix ipi". 

Also, for the "replicate" command, the check seems to be not working as expected. I recently tried to replicate a system with a 'small' tilt and still had to use the "box" command to avoid the check to bypass an unexpected error exit.

akohlmey:
> But what does fix deform do if the box is a very large tilt to start with?

"fix deform" has a "flip on/off" keyword (on by default), that will either flip the tilt or not when it is over the given ratio.

> Ditto for read_restart?

I would have expected that "read_restart" will read back the "Domain::tiltsmall" flag and thus behave as if the "box" command is included, but that does not seem to be the case. So "read_restart" will currently require re-issuing the "box" command beforehand to read restarts with large tilts.

akohlmey:
> be unaware they are running with a very inefficient geometry.

BTW, I don't recall any other case in LAMMPS where we have a hard error because of inefficiency.

akohlmey:
In summary, my current thinking is that keeping the "box" command is the 'lesser evil' compared to adding a flag to multiple other commands. The alternative would be to drop the command entirely, remove the `Domain::tiltsmall` flag, and downgrade the error to a warning.

athomps:
I think we should change the ERROR to a WARNING and get rid of the `box` command. I did my own testing on examples/in.melt, increasing xy=xz=yz by integer multiples of lx=ly=lz.  This gave the correct trajectory up to xy/lx=5, running on 8 procs for 250 time steps, at which point the communication cost had increased by 35x, increasing sharply as O(xy^4), but only for 1 < xy/lx < 5.  Beyond xy/lx=5, I got lost atoms.  I think generating an error for xy/lx > 0.5 is inappropriate, since most people will be doing it for a valid reason, and will be paying attention to any performance loss.  Many of the people who choose a box with excessive tilt and don't notice the performance loss will either suppress the error using `box tilt large` or get a `lost atom` error.

athomps:
> The alternative would be to drop the command entirely, remove the `Domain::tiltsmall` flag, and downgrade the error to a warning.

I think we should do this alternative, see post above..

akohlmey:
This PR has been changed to implement `reset_atoms` as outlined.
The removal of "box" has been added to PR #3527

sjplimp (commented):
Is having 3 additional doc pages necessary while we are still supporting
the old syntax?  I'm not clear why there isn't one doc page with the content for
all 3 options, as for all other commands I can think of.  The total volume of content
is not large.

akohlmey:
> I'm not clear why there isn't one doc page with the content for
> all 3 options, as for all other commands I can think of. The total volume of content
> is not large.

Three reasons:
1) Convenience. The three sub-command pages were already written and needed only small adjustments and a small wrapper for the "meta-command". It is also easier to refer to the different functionality. This somewhat follows what we do for dumps.
2) Clarity. I've looked at the page for the "kim" command - which has a similar function as "reset_atoms" and find it a bit confusing and quite convoluted. I would rather prefer to split that one into multiple sub-command doc pages, too, than combining the pages of these previously separate commands.
3) User behavior. Based on my interactions with users on the mailing list and forum, I noticed that people mostly read the beginning part of the page when they look up the documentation for commands. So long pages with many different sections with explanations are going contrary to that behavior. This is similar to why there is such a big business in search engine optimizations (SEO). If your page is not shown in the first 10 matches on Google, it will not be easily seen.

sjplimp (approved):

sjplimp:
@akohlmey Does this mean all the issues relative to doc page changes were resolved ?

akohlmey:
> @akohlmey Does this mean all the issues relative to doc page changes were resolved ?

Yes, that was in commit 9ba759b. The other changes were needed to get through integration testing.

PR #3536, opened 2022-11-23
support LJ units in fix ipi

Yi-FanLi:
**Summary**

Choose the unit conversion constants based on whether force->boltz is 1.0 or not. This makes fix ipi support units lj.

**Related Issue(s)**

**Author(s)**

Yifan Li, Princeton University, yifanl0716@gmail.com

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/src/MISC/fix_ipi.cpp
+++ b/src/MISC/fix_ipi.cpp
@@ -406,9 +406,9 @@ void FixIPI::final_integrate()
     potconv=3.1668152e-06/force->boltz;
     posconv=0.52917721*force->angstrom;
   }
-    posconv3=posconv*posconv*posconv;
-    forceconv=potconv*posconv;
-    pressconv=1/force->nktv2p*potconv*posconv3;
+  posconv3=posconv*posconv*posconv;
+  forceconv=potconv*posconv;
+  pressconv=1/force->nktv2p*potconv*posconv3;
 
   // compute for potential energy
   pot=modify->compute[modify->find_compute("thermo_pe")]->compute_scalar();
--- a/src/MISC/fix_ipi.cpp
+++ b/src/MISC/fix_ipi.cpp
@@ -401,7 +401,7 @@ void FixIPI::final_integrate()
   char retstr[1024];
 
   // conversions from LAMMPS units to atomic units, which are used by i-PI
-  if (force->boltz == 1.0 ) potconv = posconv = 1.0;  // if one uses LJ units, posconv and potconv should be 1
+  if (strcmp(update->unit_style,"lj") == 0) posconv = 1.0; // if one uses LJ units, posconv should be 1
   else { // if one does not use LJ units, convert energy to eV and length to Angstrom
     potconv=3.1668152e-06/force->boltz;
     posconv=0.52917721*force->angstrom;
--- a/src/MISC/fix_ipi.cpp
+++ b/src/MISC/fix_ipi.cpp
@@ -322,7 +322,8 @@ void FixIPI::initial_integrate(int /*vflag*/)
   double *boxhi = domain->boxhi;
   double *boxlo = domain->boxlo;
   double posconv;
-  posconv=0.52917721*force->angstrom;
+  if (force->boltz == 1.0) posconv = 1.0; // if one uses LJ units, posconv should be 1
+  else posconv=0.52917721*force->angstrom; // if one does not use LJ units, convert length to Angstrom
   boxlo[0] = -0.5*cellh[0]*posconv;
   boxlo[1] = -0.5*cellh[4]*posconv;
   boxlo[2] = -0.5*cellh[8]*posconv;
@@ -400,11 +401,14 @@ void FixIPI::final_integrate()
   char retstr[1024];
 
   // conversions from LAMMPS units to atomic units, which are used by i-PI
-  potconv=3.1668152e-06/force->boltz;
-  posconv=0.52917721*force->angstrom;
-  posconv3=posconv*posconv*posconv;
-  forceconv=potconv*posconv;
-  pressconv=1/force->nktv2p*potconv*posconv3;
+  if (force->boltz == 1.0 ) potconv = posconv = 1.0;  // if one uses LJ units, posconv and potconv should be 1
+  else { // if one does not use LJ units, convert energy to eV and length to Angstrom
+    potconv=3.1668152e-06/force->boltz;
+    posconv=0.52917721*force->angstrom;
+  }
+    posconv3=posconv*posconv*posconv;
+    forceconv=potconv*posconv;
+    pressconv=1/force->nktv2p*potconv*posconv3;
 
   // compute for potential energy
   pot=modify->compute[modify->find_compute("thermo_pe")]->compute_scalar();


src/MISC/fix_ipi.cpp
```diff
@@ -322,7 +322,8 @@ void FixIPI::initial_integrate(int /*vflag*/)
   double *boxhi = domain->boxhi;
   double *boxlo = domain->boxlo;
   double posconv;
-  posconv=0.52917721*force->angstrom;
+  if (force->boltz == 1.0) posconv = 1.0; // if one uses LJ units, posconv should be 1
```
akohlmey, line 325: To check for the current units setting, you should do a string compare of `update->unit_style`:

```suggestion
  if (strcmp(update->unit_style,"lj") == 0) posconv = 1.0; // if one uses LJ units, posconv should be 1
```

src/MISC/fix_ipi.cpp
```diff
@@ -400,8 +401,11 @@ void FixIPI::final_integrate()
   char retstr[1024];
 
   // conversions from LAMMPS units to atomic units, which are used by i-PI
-  potconv=3.1668152e-06/force->boltz;
-  posconv=0.52917721*force->angstrom;
+  if (force->boltz == 1.0 ) potconv = posconv = 1.0;  // if one uses LJ units, posconv and potconv should be 1
```
akohlmey: Same as above:

```suggestion
  if (strcmp(update->unit_style,"lj") == 0) posconv = 1.0; // if one uses LJ units, posconv should be 1
```

akohlmey:
@Yi-FanLi since i-PI (at least the version included here) does implicitly assume that you are using atomic units.
There should be a warning printed in the init() function, if the conversion is not done. Also a corresponding note should be added to the documentation in doc/src/fix_ipi.rst.

ceriottm:
Hi @Yi-FanLi, thanks for contributing to `fix_ipi`.  I want to make sure I understand what you're trying to do here. 
As @akohlmey mentioned, i-PI has a very hardcore "everything is atomic units" convention, internally, and so at the interface LAMMPS has to convert to whatever internal units it's using. As far as I know, this is currently working as intended. Now, essentially what you do here is to interpret what i-PI is passing as if it were LJ units, but it really uses atomic units!

If I haven't missed something, in all honesty, I am not very happy about this change because it's kind of hacky: you are basically say "ok, I'll pretend i-PI is using LJ units too" but this is dangerous, because e.g. then you also have to make assumptions on the mass of the atoms, the temperature that is used within i-PI both for thermostatting and to define the path integral spring constant.... 

Could you perhaps explain what you are trying to achieve? There may be a better way to achieve the same without introducing a confusing special case that implies you are hijacking i-PI to do something it's not meant to do. Cheers

Yi-FanLi:
@ceriottm Hi Michele, thanks for your kind reply.
Let me explain what I am doing here:
The point is to run a path integral MD for noble gas systems with the Lennard-Jones potential **using the LJ reduced units**.  Inside LAMMPS, this is done by setting all unit conversion factors equal to 1 (i.e., force->boltz, force->angstrom, etc.) 

**The unit style lj is different from others (like metal and real)**, since there is no direct conversion of its length to Angstrom, its energy to eV, and its time to ps. No unit conversion should be applied in the molecular dynamics run when the LJ reduced units are used. This is different from the other units. Your previous implementation is correct for all the other units, but if I want to use units lj in LAMMPS, it does not work correctly. For example, it uses posconv=0.52917721*force->angstrom=0.52917721 when units=lj. But in this case my length unit is not Angstrom but sigma in the lj pair style. So this conversion does not make sense.

When one is using the LJ reduced units, the conversion between the reduced units and the real unit also depends on the system. For example, mass = 1.0 corresponds to 6.64e-27 kg for He, and epsilon = 1.0 corresponds to 6.03 K. One should not do the unit conversion explicitly during the MD simulation if LJ units are used.

If one chooses units lj in LAMMPS, then in the i-PI input script all units should be set to "atomic_unit" (to make sure all the unit conversion constants to be 1), and fix_ipi.cpp should pass the coordinates and forces exactly as calculated by LAMMPS. Therefore, all the unit conversion constants should be 1, which is distinct from other units.

Did I clearly describe what I mean?

akohlmey:
> The point is to run a path integral MD for noble gas systems with the Lennard-Jones potential

What keeps you from using "units real" and the corresponding parameters?
If you want results in reduced units, you can always convert the results after the fact.
There really is no advantage to using reduced units in LAMMPS for such a system.

Yi-FanLi:
> What keeps you from using "units real" and the corresponding parameters?

Nothing keeps me from using "units real", but sometimes it is also convenient to use "units lj". After all LAMMPS supports "units lj". I think this modification is just to add support of the "lj" unit style in fix ipi at almost no cost. It does not introduce any change to the behavior of fix ipi to the currently support unit styles, and also benefits people who want to use "units lj". So I think it should be doable.

akohlmey:
> So I think it should be doable.

But since i-PI does *not* support reduced units but rather expects atomic units (as @ceriottm has confirmed), it is *not* a good idea. You will have to hack i-PI and to give it bogus input so that it operates in the way you need it to be operating.

LAMMPS is open source, so you are free to keep using this hack for your own purposes, and anybody else that has this urge can do so, but I don't think that we should do this modification against the preferences of the author of the fix and the i-PI code. Sorry.

ceriottm:
hi @Yi-FanLi - I agree with @akohlmey on this. units are a pain to work with, but what you are doing here, while harmless at the surface, opens up the door to all sorts of nasty mistakes in simulations. I think that the easiest way to achieve what you want is actually to use `units electron` in lammps, that if I'm not mistaken uses atomic units for all the quantities we care about here. of course you can keep using your approach if you like - you seem to know what you are doing so hopefully you will avoid these pitfalls. thanks for using i-PI

Yi-FanLi:
Thanks for your kind reply. @ceriottm  I appreciate this opportunity to communicate with you.
As I understand, the "atomic unit" in i-PI operating by avoiding unit conversions in the equations of motion, and converting the user's input quantities with constants in utils/units.py. Is it correct?

ceriottm:
yes, all the conversion are done at the I/O level. internally literally everything is in a.u.; even the temperature is measured in hartree internally. anyway if you want to continue the conversation on i-PI I suggest you move to the [i-PI user forum](https://groups.google.com/g/ipi-users) so we don't pollute the github of LAMMPS, and we don't spam the maintainers.

Yi-FanLi:
> all the conversion are done at the I/O level

This behaves the same as LAMMPS's "lj" units do. Your current unit conversion in "fix_ipi.cpp" will lead to wrong behavior if someone uses "units lj" in LAMMPS. You should either tell people not to use lj units in the document of fix ipi, or set the conversion constants to be 1 for lj units in fix_ipi.cpp.

akohlmey:
> This behaves the same as LAMMPS's "lj" units do.

Atomic units are not the same as reduced units. In atomic units you have a specific reference, in reduced units it is arbitrary. As @ceriottm stated, atomic units are equivalent to the "units electron" setting in LAMMPS.

We will add a check to fix ipi in LAMMPS that it will refuse to run with incompatible "units" settings.

PR #3540, opened 2022-11-27
cmake: update GTest detection

rbberger:
**Summary**

LAMMPS currently forces either a download or an offload tarball of GTest to be present. This change allows an existing installation to be used. I'd also really like to see this in the next stable, so I already rebased it in https://github.com/rbberger/lammps/tree/stable_cmake_updates in case there is another maintenance release to save some work when cherry-picking.

- Use CMake's find_package(GTest) starting with v3.20
- Use existing logic as a fallback if not found
- Update the following targets to match the newer CMake targets
  * GTest::GTest -> GTest::gtest
  * GTest::GMock -> GTest::gmock
  * GTest::GTestMain -> GTest::gtest_main
  * GTest::GMockMain -> GTest::gmock_main

**Related Issue(s)**

**Author(s)**

@rbberger 

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**


rbberger:
@akohlmey looks like the Windows test is failing due to unrelated updates on the runners. Python seems to be detected to be v3.11 and the MSVC compiler also had a minor bump. **Update:** Forcing `Python_EXECUTABLE` to the selected Python3 (3.10) fixes the compilation issue but still segfaults. **Update 2**: Here is the windows image changelog: https://github.com/actions/runner-images/commit/9b06b852013a0b207a0f73fc2b9aea75c29287f5

rbberger:
@akohlmey Found it. It's something with auto-detected PNG and/or JPEG libraries. Setting WITH_PNG/JPEG to OFF removes the failure.

akohlmey:
> @akohlmey Found it. It's something with auto-detected PNG and/or JPEG libraries. Setting WITH_PNG/JPEG to OFF removes the failure.

Thanks.

akohlmey:
@rbberger I am not sure what to think of this change.

I currently have no intention to do another update to the stable version. Backporting changes is becoming increasingly laborious and so far there have been no major problems overall. Besides, this is not a kind of change that I would be interested to backport in the first place.

In fact, I am not yet convinced that it is a desirable change. What is the benefit of using an existing googletest installation and thus having to be compatible to a wider variety of code versions. More importantly, what if googletest will drop support for C++11? Since it is not likely that LAMMPS will make C++14 or later the minimum standard soon, that would result to even more complex CMake code and maintenance effort.

Since I do not consider testing as a crucial part of LAMMPS that is relevant to LAMMPS *users* and thus required packaging or anything similar, I prefer the simpler approach of just requiring one specific version. This is also in line with what the googletest docs recommend, if I remember correctly.

rbberger:
> @rbberger I am not sure what to think of this change.
> 
> I currently have no intention to do another update to the stable version. Backporting changes is becoming increasingly laborious and so far there have been no major problems overall. Besides, this is not a kind of change that I would be interested to backport in the first place.
> 

If there is no further stable update in the current cycle, fine.

> In fact, I am not yet convinced that it is a desirable change. What is the benefit of using an existing googletest installation and thus having to be compatible to a wider variety of code versions. More importantly, what if googletest will drop support for C++11? Since it is not likely that LAMMPS will make C++14 or later the minimum standard soon, that would result to even more complex CMake code and maintenance effort.
> 

If there is a version requirement, you can use the `find_package` mechanism to limit what versions are supported by LAMMPS. The change makes sure it's at least 1.12. We could add `EXACT` to force that version, similar what the download does. 

The main benefit of this is change is making pulling this test dependency **optional**, like anywhere else we do in CMake. What we have in LAMMPS right now is becoming more and more of an attempt of what an actual package manager does. Let's **allow** actual package managers worry about dependencies and how to work in offline scenarios. Right now, even if you have all the right dependencies installed on an offline machine, this logic will try to pull in something that isn't necessary.

> Since I do not consider testing as a crucial part of LAMMPS that is relevant to LAMMPS _users_ and thus required packaging or anything similar, I prefer the simpler approach of just requiring one specific version. This is also in line with what the googletest docs recommend, if I remember correctly.

Things evolve. LAMMPS doesn't live in a bubble.

akohlmey:
> Things evolve. LAMMPS doesn't live in a bubble.

On the other hand, we don't need to add changes because it is technically possible without looking for a significant benefit. I fail to see the benefit. Very few people use the testing features anyway.

If adding googletest as library would be a new feature, I might think differently, but since the current mechanism exists for quite some time, has served its purpose well, and since it required some significant "massaging" to work consistently across all platforms with all the different compilation settings and modes in LAMMPS, I  am extremely hesitant to make changes that take away knowing exactly what will be used. 

The fact that this requires adding a test for a CMake version is another concern. I'd rather not add more CMake scripting that is dependent specific CMake versions.

rbberger:
This is not a change just for change. In this particular case it allows someone on a restricted system to install LAMMPS dependencies via e.g. Spack and then do a CMake compilation.

akohlmey:
Testing is an optional feature and not a requirement to install LAMMPS. So that is not something requiring a change that adds complexity.

rbberger:
You're ignoring people who are developing/working with LAMMPS and need to test as well on such restricted systems. Anyway, I understand this is not going anywhere with you. @stanmoore1, @athomps, or @sjplimp please chime in on this. If I'm literally the only one who sees the benefit of this, I'll close it myself and find a workaround for my use case.

stanmoore1:
Here are my 2 cents: I think it is valuable to provide a way for users who cannot connect to the internet to still run testing in LAMMPS. The actual code changes in this PR look reasonable, but I can also understand @akohlmey's concern about code branches that depend on CMake version. I personally don't have very strong feelings on this either way.

>We could add EXACT to force that version, similar what the download does.

I think this would be necessary to ensure the same version is used in either case, so multiple versions don't need to be supported.

rbberger:
@stanmoore1 added `EXACT`. I'm also not a fan of CMake version specific branches myself. If it were all up to me I would just increase the minimum required CMake version. The arguments for using such an old CMake version are already close to zero, but I already know that is "too extreme" for some people to stomach. So I didn't even attempt that.

All I'm doing instead is adding an optional support. It still falls back to the existing behavior if its CMake version is too old and/or it doesn't find that exact version of Google Test installed.

akohlmey:
> Here are my 2 cents: I think it is valuable to provide a way for users who cannot connect to the internet to still run testing in LAMMPS.

What @rbberger is omitting here is that there *is* already a mechanism for offline access to compiling the testing code. Only it needs to provide the source tarball instead of a pre-compiled and installed library. Using the source for this has the benefit that it inherits all compilation settings from LAMMPS and thus is more likely compatible than an external compilation.

> > We could add EXACT to force that version, similar what the download does.
> I think this would be necessary to ensure the same version is used in either case, so multiple versions don't need to be supported.

Requiring an exact version pretty much goes contrary to the "let CMake find the package" mechanism and the argument about using a packaging system to provide the library. With requiring a fixed version, chances are high that either extra packaging work is needed (i.e. multiple versioned packages must be provided) or that the requirement breaks the dependencies and the provided version is incompatible. Then you end up having to compile from source anyway.

rbberger:
> LAMMPS currently forces either a download **or an offload tarball of GTest to be present**. This change allows an existing installation to be used.

I'm not omitting anything. All I want is a standard way of getting a dependency into CMake that doesn't involve having to handle dependency tarballs yourself.

sjplimp:
I read thru the comments, but don't understand all the technical issues.  So I'm not sure I should have a vote. But this comment by @rbberger was compelling:
>> Things evolve. LAMMPS doesn't live in a bubble.

And as a general rule I think it is good to support as many use cases as possible and practical.

It also seems you are saying that updating the CMake version LAMMPS uses could make this simpler.
We discussed that last month in the developer's mtg, and as I recall were in general agreement to do
that.  However there is one or two Linux distributions which only distribute an older version,
thus we might wait until they upgrade, to make it easier on LAMMPS users.  I may be confusing this
with compiler versions, but @akohlmey will likely recall the details.

rbberger:
> It also seems you are saying that updating the CMake version LAMMPS uses could make this simpler.

In this particular case it only really removes the if statement that checks if the CMake version is new enough. 

Long term updating the CMake version will have benefits in more places since it allows us to remove some workarounds. But it's not important for this PR and I don't want it to get side tracked.

badshah400:
When building RPMs for openSUSE, we build our packages in a sandboxed VM with no network provisions. So if we want to run tests, we have to either provide a downloaded gtest tarball or, if possible, use a system installed version of it. The latter option is preferable to us for, among other reasons, to reduce src rpm sizes, reduce download/upload data when forking and committing, and to allow a consistent gtest version to be used for testing different packages across the distro.

I overlooked this PR (sorry @rbberger) and ended up writing my own patch that does something similar to what is here, albeit in a slightly different form. The patch sets up a new `cmake` option `EXTERNAL_CMAKE` (`OFF` by default) to allow looking for `gtest` using `find_package` when turned on during configure. I note the patch below for what it is worth. If this is something the devs may be interested in, I would be happy to open a new PR.

I have tested the patch on our builds, both when turning the new option `ON` and `OFF`. 

Thanks in advance for any suggestions.

```diff
---
 unittest/CMakeLists.txt |   41 +++++++++++++++++++++++++----------------
 1 file changed, 25 insertions(+), 16 deletions(-)

Index: lammps-stable_29Aug2024_update2/unittest/CMakeLists.txt
===================================================================
--- lammps-stable_29Aug2024_update2.orig/unittest/CMakeLists.txt
+++ lammps-stable_29Aug2024_update2/unittest/CMakeLists.txt
@@ -9,24 +9,33 @@ if((CMAKE_CXX_COMPILER_ID STREQUAL "GNU"
   message(FATAL_ERROR "Need GNU C++ compiler version 6.x or later for unit testing")
 endif()
 
-message(STATUS "Downloading and building googletest framework")
-set(GTEST_URL "https://github.com/google/googletest/archive/release-1.12.1.tar.gz" CACHE STRING "URL of googletest source")
-set(GTEST_MD5 "e82199374acdfda3f425331028eb4e2a" CACHE STRING "MD5 sum for googletest source")
-mark_as_advanced(GTEST_URL)
-mark_as_advanced(GTEST_MD5)
-set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
+set(EXTERNAL_GTEST OFF CACHE BOOL "Use system gtest")
+if (EXTERNAL_GTEST)
+  find_package(GTest REQUIRED)
+  add_library(GTest::GTest ALIAS GTest::gtest)
+  add_library(GTest::GMock ALIAS GTest::gmock)
+  add_library(GTest::GTestMain ALIAS GTest::gtest_main)
+  add_library(GTest::GMockMain ALIAS GTest::gmock_main)
+else()
+  message(STATUS "Downloading and building googletest framework")
+  set(GTEST_URL "https://github.com/google/googletest/archive/release-1.12.1.tar.gz" CACHE STRING "URL of googletest source")
+  set(GTEST_MD5 "e82199374acdfda3f425331028eb4e2a" CACHE STRING "MD5 sum for googletest source")
+  mark_as_advanced(GTEST_URL)
+  mark_as_advanced(GTEST_MD5)
+  set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
 
-include(ExternalCMakeProject)
-if(CMAKE_CXX_COMPILER_ID STREQUAL "IntelLLVM")
-  set(cxx_strict_flags "-Wno-error -Wno-unused-command-line-argument")
+  include(ExternalCMakeProject)
+  if(CMAKE_CXX_COMPILER_ID STREQUAL "IntelLLVM")
+    set(cxx_strict_flags "-Wno-error -Wno-unused-command-line-argument")
+  endif()
+  set(INSTALL_GTEST OFF CACHE BOOL "" FORCE)
+  set(INSTALL_GMOCK OFF CACHE BOOL "" FORCE)
+  ExternalCMakeProject(googletest ${GTEST_URL} ${GTEST_MD5} googletest . "")
+  add_library(GTest::GTest ALIAS gtest)
+  add_library(GTest::GMock ALIAS gmock)
+  add_library(GTest::GTestMain ALIAS gtest_main)
+  add_library(GTest::GMockMain ALIAS gmock_main)
 endif()
-set(INSTALL_GTEST OFF CACHE BOOL "" FORCE)
-set(INSTALL_GMOCK OFF CACHE BOOL "" FORCE)
-ExternalCMakeProject(googletest ${GTEST_URL} ${GTEST_MD5} googletest . "")
-add_library(GTest::GTest ALIAS gtest)
-add_library(GTest::GMock ALIAS gmock)
-add_library(GTest::GTestMain ALIAS gtest_main)
-add_library(GTest::GMockMain ALIAS gmock_main)
 
 option(SKIP_DEATH_TESTS "Do not run 'death tests' to reduce false positives in valgrind" OFF)
 mark_as_advanced(SKIP_DEATH_TESTS)
```

PR #3541, opened 2022-11-29
Fix bug in Kokkos ReaxFF, Tersoff, and Fix Neigh History on GPUs

### LINKED ISSUE #3535: [BUG] ReaxFF + Kokkos + NPH/NPT problem
**Summary**

Simulations of ReaxFF + Kokkos present problems when performing simulations with NPH or NPT. Even when using independent temperature control when there is an expansion of the box due to melting, the system temperature also jumps. The script works fine without kokkos.

Lammps version 3Nov22, OS: Pop!_OS 22.04  
KOKKOS_ARCH = Volta70 
AMD Ryzen Threadripper 3970X 32-Core Processor

**Expected Behavior and Actual Behavior**

Using the same input file (except minor modifications between EAM and ReaxFF as units) I ran NPH + Langevin Thermostat simulations to model the iron melting point (In other simulations I used NPT with equivalent results).

In the simulation, I heat the system up to 1500 K in a few steps and then I heat it more slowly. When the system melts, an abrupt increase in the box volume is observed. At the same time, the temperature increases linearly. This happens with EAM+Kokkos without problems. However, for ReaxFF + Kokkos, once an abrupt rise in the case occurs, there is also an abrupt jump in temperature that should not occur. When running the same script for ReaxFF without kokkos (usual mpirun), no melting is observed in the tested window, only linear expansion of the volume. The temperature evolution is similar to what happened with EAM (black and blue lines overlapped in the plot)

The graph shows what happened.

Volandtemp

**Steps to Reproduce**

Attached are 3 folders with the input scripts and the log.lammps files 

commands: 

EAM + Kokkos /ReaxFF +Kokkos

lmp  -k on g 1 -sf kk -pk kokkos newton on neigh half -in in.melting

ReaxFF (usual mpirun)

mpirun -np 32 lmp -in in.melting 

[Test-ReaxFF-EAM-Kokkos.zip](https://github.com/lammps/lammps/files/10079501/Test-ReaxFF-EAM-Kokkos.zip)

stanmoore1:
**Summary**

The atom `tag` for Kokkos ReaxFF and some other styles (Tersoff variants and fix neigh history) wasn't being updated on GPUs correctly, and somehow this slipped through all the regression tests.

**Related Issue(s)**

Fixes #3535

**Author(s)**

Stan Moore (SNL)

**Licensing**

**Backward Compatibility**

Yes

### DIFF
--- a/.github/workflows/compile-msvc.yml
+++ b/.github/workflows/compile-msvc.yml
@@ -26,7 +26,7 @@ jobs:
     - name: Select Python version
       uses: actions/setup-python@v4
       with:
-        python-version: '3.10'
+        python-version: '3.11'
 
     - name: Building LAMMPS via CMake
       shell: bash
@@ -35,6 +35,8 @@ jobs:
         python3 -m pip install pyyaml
         cmake -C cmake/presets/windows.cmake \
               -D PKG_PYTHON=on \
+              -D WITH_PNG=off \
+              -D WITH_JPEG=off \
               -S cmake -B build \
               -D BUILD_SHARED_LIBS=on \
               -D LAMMPS_EXCEPTIONS=on \
--- a/src/KOKKOS/fix_neigh_history_kokkos.cpp
+++ b/src/KOKKOS/fix_neigh_history_kokkos.cpp
@@ -20,6 +20,7 @@
 #include "modify.h"
 #include "neigh_list_kokkos.h"
 #include "pair_kokkos.h"
+#include "atom_masks.h"
 
 using namespace LAMMPS_NS;
 
@@ -170,6 +171,7 @@ template <class DeviceType>
 void FixNeighHistoryKokkos<DeviceType>::post_neighbor()
 {
   tag = atomKK->k_tag.view<DeviceType>();
+  atomKK->sync(execution_space,TAG_MASK);
 
   k_firstflag.sync<DeviceType>();
   k_firstvalue.sync<DeviceType>();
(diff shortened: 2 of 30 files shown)


stanmoore1:
Windows failure looks unrelated to the changes here.

akohlmey:
> Windows failure looks unrelated to the changes here.

Yes, it is due to software changes in the Windows Azure instances used by GitHub actions.

athomps (approved):
I approve. Is there something to be learned from the fact that this did not affect any results?

stanmoore1:
>Is there something to be learned from the fact that this did not affect any results?

Our regression tests typically run short trajectories to minimize the exponential divergence found in MD, which apparently wasn't enough time steps to expose issues with atom tags. Running on the GPU makes it even worse since it has a larger amount of non-determinism, which can mask bugs. Longer physics runs showed the issue, so perhaps we should add the test cases that exposed the issues to a "heavy" set of regression tests that only run nightly or weekly or so.

stanmoore1:
This also fixes the bug reported here: https://matsci.org/t/issues-with-combination-of-fix-move-periodic-boundary-and-kokkos/45346

athomps:
I think the reason it was not detected is that atoms tags are hardly used at all in ReaxFF.  But they are used in Tersoff, so I would expect that to show a clear effect:

`if (itag > jtag) {`

athomps:
I think there should be possible to construct a light-weight test, maybe using Tersoff, that exposes this bug. It might be useful for other things too.

PR #3549, opened 2022-12-09
Add python scripts for creating tables, pair style table/mod

akohlmey:
**Summary**

This adds python classes and functions that in the tools folder that can be used to create tabulated potential files for LAMMPS.  The user needs to write a custom python code that contains the function for the potential energy and -optionally - the force. If the force function is missing, its values will be constructed from numerical differentiation (error to analytical force typically about 1.0e-10 to 1.0e-8). This custom code then instantiates a tabulation class to which it passes the function(s).

In addition, a modified version of pair style table is included: pair style table/mod. This style differs from pair style table by using linear tabulation internally compared to r-squared based tabulation in pair style table.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Post Submission Checklist**


akohlmey:
Question: should pair style table/mod remain in src or be moved to src/EXTRA-PAIR?

Technically, it would be possible to merge the two styles, but then it would become even more convoluted than it already is.

evoyiatzis:
Useless info of the day  : the add_argument() method has a 'type' keyword for performing any necessary type-checking and type conversion. It might be used for i.e. num-points & inner.

akohlmey:
> Useless info of the day : the add_argument() method has a 'type' keyword for performing any necessary type-checking and type conversion. It might be used for i.e. num-points & inner.

Thanks for the tip. Yes, command line argument checking (also for meaningful/supported values) was missing and this is a neat way of adding it.

akohlmey:
will split this PR into two parts: 1) add python tabulation tools 2) refactor pair style table with the addition of linear spacing (and probably also other tabulation styles including their documentation).

PR #3551, opened 2022-12-09
Add new compute efield/wolf/atom command

akohlmey:
**Summary**

This adds a compute efield/wolf/atom command to the EXTRA-COMPUTE package

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

This adds a compute style that can be used to compute the electric field vector and strength at the position of an atom.
Unlike the corresponding command in the DIELECTRIC package, this command does an independent calculation and thus is rather slow. It uses the Wolf summation approximation and thus will assume that long-range contributions cancel.

**Post Submission Checklist**


doc/src/compute_efield_wolf_atom.rst
```diff
@@ -0,0 +1,127 @@
...
+"""""""""""
+
+This compute calculates a per-atom vector, which can be accessed by
+any command that uses per-atom values from a compute as input.  See
+the :doc:`Howto output <Howto_output>` page for an overview of
+LAMMPS output options.
+
+The vector contains 4 values per atom. The first 3 per-atom vector
+values will be the x-, y-, and z-direction electric field components
+in force units the 4th per-atom vector value is the field magnitude.
```
athomps: ...units. The 4th...

doc/src/compute_efield_wolf_atom.rst
```diff
@@ -0,0 +1,127 @@
...
+enough :math:`\alpha` damping parameter, the second term shrinks and the
+potential becomes a rapidly-converging real-space summation.  With a
+long enough cutoff and small enough :math:`\alpha` parameter, the
+electric field calculated by the Wolf summation method approaches that
+computed using the Ewald sum.
+
+The value of the electric field components will be 0.0 for atoms not in
+the specified compute group.
+
+When the *limit* keyword is used, only contributions from atoms in the
```
athomps, line 72: The keyword name *limit* is not very descriptive for this option. Is there a similar option in some existing fix or compute that could be copied.  If not, what about *include*, following *neigh_modify*? Or just plain *group*?

athomps (approved):
I approve, apart from comments. Also, one idea for the future might be to add this to the Pair() class interface with optional support, similar to single(), born_matrix(), etc.

doc/src/compute_efield_wolf_atom.rst
akohlmey: @athomps I just did a bit of a survey and in most cases either the second group argument is mandatory or optional but without a keyword. Compute coord/atom uses the keyword "group" which is not really representative of what it does, either. Perhaps "restrict" or "select" could be an option, but if we want to go for consistency, then "group" would be the precedence. The terms "limit" or "restrict" or "select" are indicating that the second group is limiting which atoms are contributing to the field while the compute group is selecting for which atoms the property is computed.

akohlmey:
@sjplimp I have been thinking a bit about how to adapt this also for QM/MM applications and proper long-range Coulomb. The benefit of this kind of compute would be that it won't modify the force or energy arrays/accumulators.
- I think it should be possible to make a variant with long-range Coulomb by using the relevant parts of pair_coul_long.cpp and adding a suitably adapted version of the "fieldforce()" method to Kspace styles.
- It should also be possible to compute the potential instead of the field vector (basically it would be epot/q_i same as E is f/q. The self-energy term would be tricky, but my intuition is to just use the unmodified term.

doc/src/compute_efield_wolf_atom.rst
athomps: I like `group`, as it is fairly obvious what it does when people need to use it.  As with `coord/atom`, the default should be `group all`

doc/src/compute_efield_wolf_atom.rst
```diff
@@ -0,0 +1,126 @@
...
+   \frac{q_i q_j {\rm erf}(\alpha r_{ij})}{r_{ij}} \qquad r < r_c
+
+where :math:`\alpha` is the damping parameter, and *erf()* and *erfc()*
+are error-function and complementary error-function terms.  This
+potential is essentially a short-range, spherically-truncated,
+charge-neutralized, force-shifted, pairwise *1/r* summation.  With a
+manipulation of adding and subtracting a self term (for i = j) to the
+first and second term on the right-hand-side, respectively, and a small
+enough :math:`\alpha` damping parameter, the second term shrinks and the
+potential becomes a rapidly-converging real-space summation.  With a
```
sjplimp, line 64: Is this statement about accuracy of Wolf correct or are there additional caveats?
I thought Wolf summation might only be appropriate for certain kinds of materials,
e.g. solids, and not liquids or soft matter (biomoleulce, biomembranes) with large asymmetries?  If so, some mention of those caveats would be useful.  Otherwise users
might think this always gives good Efield values.

akohlmey: @sjplimp This was taken verbatim from the documentation of the corresponding pair style. I have been careful to put the statement "Define a computation that approximates the electric field" at the beginning of the doc page and thus already alerting people. For as long as it is not used for computing forces it is actually not so bad since errors do not accumulate and the structure will still follow the (hopefully more accurate) choice of pair/kspace style.

PR #3552, opened 2022-12-09
Fix bug in Kokkos minimize on GPUs

stanmoore1:
**Summary**

Fix bug in Kokkos minimize on GPUs. Data transfer between GPU and CPU was missing, so the atom positions were not updated correctly, which led to `Stopping criterion = linesearch alpha is zero` when it should be `Stopping criterion = energy tolerance`. 

**Related Issue(s)**

None

**Author(s)**

Stan Moore (SNL), reported by Megan McCarthy (SNL)

**Licensing**

**Backward Compatibility**

Yes


athomps (approved):
I approve.

PR #3538, opened 2022-11-24
Remove suffix from style names when writing restart files

akohlmey:
**Summary**

By removing the style suffix from style names when writing binary restarts, the resulting restart file should be readable with suffixes enabled or without. This also includes some minor changes to the restart procedure to better match stored restart data from fixes.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

For restarts written prior to this change, the same accelerator/suffix must be enabled to read the restart.
This is the old behavior and thus this change should not introduce any compatibility problems.
Behavior of restarting without a suffix of with a different suffix from a restart written with suffixes remove has not be fully verified.

**Implementation Notes**

This introduces a new utility function `utils::strip_style_suffix()` providing a string operation needed by this change multiple times.

**Post Submission Checklist**

- [x] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [x] The source code follows the LAMMPS formatting guidelines
- [x] Suitable new documentation files and/or updates to the existing docs are included
- [x] The added/updated documentation is integrated and tested with the documentation build system
- [x] The feature has been verified to work with the conventional build system
- [x] The feature has been verified to work with the CMake based build system
- [x] Suitable tests have been added to the unittest tree.

PR #3555, opened 2022-12-13
Refactor Kokkos `AtomVec`

stanmoore1:
**Summary**

Greatly reduce duplicated code in Kokkos `AtomVec` styles by inheriting from non-Kokkos atom_vec styles and removing Kokkos host methods, falling back to original CPU (non-Kokkos) methods instead. This requires virtual inheritance for non-Kokkos atom_vec styles.

**Related Issue(s)**

Should address restart read/write issues in #3538.

**Author(s)**

Stan Moore (SNL)

**Licensing**

**Backward Compatibility**

Yes


stanmoore1:
Kokkos regression tests pass now.

stanmoore1:
This is ready to merge from my POV.

stanmoore1:
@akohlmey Steve was wondering if there is any potential performance impact from using virtual inheritance in the `AtomVec` styles, i.e. if calling the class functions would be slower since it inherits from a virtual base. As far as I know, the only impact is that a `dynamic_cast` is required, which is slower that `static_cast`, but this is typically called once during `init()` so no performance impact. Can you confirm?

sjplimp (approved):

stanmoore1:
Quick search suggests: "Common implementations will make access to data members of virtual base classes use an additional indirection." Maybe we want to do a quick performance check before merging.

https://stackoverflow.com/questions/4899657/performance-impact-of-virtual-inheritance

akohlmey:
> Can you confirm?

I don't expect that it will matter much.
- most of the access to the stored pointers are through the copied pointer variables in the Atom class
- almost all data access within the AtomVec classes is in the base class itself or to local data members
- affected functions in the AtomVec classes are virtual already

stanmoore1:
I ran this on my Linux box with the LJ benchmark (vanilla CPU) and I don't see any noticeable slowdown.

stanmoore1:
So I think it can be merged now.

PR #2424, opened 2020-10-15
Patching neighbor history restarts

jtclemm:
**Summary**

When writing a restart file, fix neighbor/history writes data stored in per-atom arrays. However, this data is one timestep out of date. In the prior pair->compute call, only the data in page files is updated. This PR addresses this bug by calling an extra instance of pre_exchange (which copies page file data into per-atom arrays) before packing restart data.

**Related Issue(s)**

None

**Author(s)**

Joel Clemmer (SNL) - jtclemm@sandia.gov

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

Old restart files may not work if they saved neighbor/history data.

**Implementation Notes**

Added new write_restart method in fix neighbor/history to call pre_exchange(). It doesn't actually write data (other than the size of the data written, 0).

**Post Submission Checklist**

- [ ] The feature or features in this pull request is complete
- [ ] Licensing information is complete
- [ ] Corresponding author information is complete
- [ ] The source code follows the LAMMPS formatting guidelines
- [ ] Suitable new documentation files and/or updates to the existing docs are included
- [ ] The added/updated documentation is integrated and tested with the documentation build system
- [ ] The feature has been verified to work with the conventional build system
- [ ] The feature has been verified to work with the CMake based build system
- [ ] Suitable tests have been added to the unittest tree.
- [ ] A package specific README file has been included or updated
- [ ] One or more example input decks are included

**Further Information, Files, and Links**

None

PR #3558, opened 2022-12-14
Small patch to BPM package

jtclemm:
**Summary**

This PR fixes an incorrect formatting in the doc page for bpm/rotational bonds and also removes some unnecessary clears in fix update/special/bonds. Clearing the arrays of broken bonds in FixUpdateSpecialBonds->setup() could prevent some broken bonds from being properly processed if they break after the last neighbor list build of a run command (such that pre_exchange() never processes the bonds). If another run statement is then called, the information that these bonds broke is then wiped from arrays before special bonds are updated.

**Related Issue(s)**

NA

**Author(s)**

Joel Clemmer (SNL)

**Licensing**

**Backward Compatibility**

NA

**Implementation Notes**

NA

**Post Submission Checklist**

NA

### DIFF
--- a/doc/src/bond_bpm_rotational.rst
+++ b/doc/src/bond_bpm_rotational.rst
@@ -67,7 +67,7 @@ which is proportional to the tangential shear displacement with a
 stiffness of :math:`k_s`. This tangential force also induces a torque.
 In addition, bending and twisting torques are also applied to
 particles which are proportional to angular bending and twisting
-displacements with stiffnesses of :math`k_b` and :math:`k_t',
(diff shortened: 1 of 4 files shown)


stanmoore1 (approved):

jtclemm:
I just realized a similar issue could occur when writing restarts. The pre_exchange() method has to be called before the restart is written to ensure all per atom data is updated. To do this, I added pre_exchange() calls to the write_restart() method of both update/sepcial/bonds and bond/history. A similar bug was fixed in fix neigh/history using this approach: https://github.com/lammps/lammps/pull/2424 .

PR #3561, opened 2022-12-16
Fix bug when atoms are added after run

stanmoore1:
**Summary**

- Prevent segfault when atoms are added with `read_data` command after a run, by zeroing out ghost atoms. Without this change, `atom->map_set()` could read uninitialized data in the `atom->tag` when velocities are read in, leading to a crash.
- Also clear atom map for `create_atoms` and `read_data`

**Related Issue(s)**

None

**Author(s)**

Stan Moore (SNL), issue reported by Jonathan Willman (USF)

**Licensing**

**Backward Compatibility**

Yes

### DIFF
--- a/src/create_atoms.cpp
+++ b/src/create_atoms.cpp
@@ -441,10 +441,12 @@ void CreateAtoms::command(int narg, char **arg)
   MPI_Barrier(world);
   double time1 = platform::walltime();
 
+  // clear global->local map for owned and ghost atoms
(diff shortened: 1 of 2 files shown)


stanmoore1:
This is ready to merge from my POV. There is another known issue with `create_atoms` after a run, but I don't have a bugfix yet, so I will submit a subsequent PR when that gets fixed.

akohlmey (approved):

PR #3532, opened 2022-11-21
Update Kokkos version in LAMMPS to 3.7.1

stanmoore1:
**Summary**

- Update Kokkos version bundled with LAMMPS to v3.7.01
- Backport Kokkos PR #5624 to Kokkos version in LAMMPS to correct occupancy calculations for Kokkos CUDA, see https://github.com/kokkos/kokkos/pull/5624 for a more detailed description. This may hurt performance a little in some cases, need to check.
- Fix init arguments and some join stuff for Kokkos 4

Closes #3568
Closes #3584

**Related Issue(s)**

#3568, #3584

**Author(s)**

Stan Moore (SNL), Christian Trott (SNL) Axel Kohlmeyer (Temple U), and Kokkos developers, thanks to @weinbe2 for helping backport Kokkos changes

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

Yes

PR #3568, opened 2022-12-20
Fix init arguments and some join stuff for Kokkos 4

crtrott:
**Summary**

Update LAMMPS to work with upcoming Kokkos 4.0 release

**Author(s)**

Christian Trott crtrott@sandia.gov

**Licensing**

**Backward Compatibility**

Should be backward compatible, including to pre Kokkos 4 releases of Kokkos. 

**Implementation Notes**

Did a bunch of compilation and run tests with bench and examples directory

**Post Submission Checklist**


crtrott:
Few comments: 

- Kokkos is now using non-volatile overloads of join/init in reductions. I added them in places missing, but LAMMPS developers should consider just removing the volatile overloads.
- I did a backward compatible (Pre Kokkos 3.7) ifdef path for initialization. One could get rid of that path if you don't intend to still compile with Kokkos 3.6 etc. 
- In the initialization path the numa option has been semi ignored for a long time. LAMMPS developers could consider getting rid of the numa option on the LAMMPS command line args. I just used it as a multipler on number of threads.

akohlmey:
> * I did a backward compatible (Pre Kokkos 3.7) ifdef path for initialization. One could get rid of that path if you don't intend to still compile with Kokkos 3.6 etc.

At least with the CMake build system, the LAMMPS KOKKOS package is set up to require the same version (or later) for an external Kokkos library than what is bundled. Currently that is 3.7.0, so with a Kokkos 3.6.x installation CMake will already fail at the configuration step.

akohlmey:
Support code for Kokkos prior to 3.7.0 is now removed and support for the "numa" or "n" option to the -kokkos flag.

stanmoore1:
I merged this into #3532

PR #3571, opened 2022-12-22
LEPTON package using the Lepton library to compute forces from expression strings

akohlmey:
**Summary**

This pull request adds a new package LEPTON that interfaces LAMMPS styles  with the [Lepton library](https://simtk.org/projects/lepton) from computing expressions and their derivatives given in strings.
These styles can function as an alternative to tabulated potentials or evaluating python functions/classes. The expressions are pre-parsed and optimized internally. For x86 there is also the option to use a JIT compiler.

Using Lepton for lj/cut is typically about 2-5 times slower than compiled code (depends a bit on the compiler and optimization flags). In comparison lj/cut with pair style python is over 10 times slower.
The pair style exists in two variants when the second has also access to the atomic charges of the two interacting atoms.

Corresponding multi-threaded versions of the styles are added to the OPENMP package, too.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

With support from the Colvars developers this pull request also contains changes to the COLVARS package and its library, so it can share the Lepton library with the LEPTON package.

Currently implemented styles in the LEPTON and OPENMP packages:
- [x] pair style lepton
- [x] pair style lepton/omp 
- [x] pair style lepton/coul
- [x] pair style lepton/coul/omp 
- [x] bond style lepton
- [x] bond style lepton/omp
- [x] angle style lepton
- [x] angle style lepton/omp
- [x] dihedral style lepton
- [x] dihedral style lepton/omp

**Post Submission Checklist**


giacomofiorin:
@akohlmey I haven't reviewed the contents of the PR in detail, but overall this is a welcome change. @jhenin have so far avoided bundling Lepton together with Colvars, so moving it to a shared location is the best approach.

For the same reason, I think it would be much simpler to keep the Lepton library unchanged, without changing or editing its files. Colvars can begin using this copy of Lepton immediately. I pushed a commit to that sense here:
https://github.com/Colvars/lammps/tree/lepton-package
With the necessary changes, i.e. eliminating the additional namespace and renaming the main header back to `Lepton.h` it should just work.

A couple of changes may be needed to make this work on Windows, but that's an issue with Lepton itself, and I can submit them as review after you decide what to do with the copy that is bundled with Colvars.

akohlmey:
> I think it would be much simpler to keep the Lepton library unchanged, without changing or editing its files.

Compiling the Lepton library produces an excessive amount of compiler warnings. I won't take the changes back that produce them. Also, mid-term I would like to have its class structure set up to be properly abstract the details of the implementation from the use case and then avoid the memory corruption issues that I can into. Of course those would need to be eventually submitted to upstream.

> A couple of changes may be needed to make this work on Windows, but that's an issue with Lepton itself

I was not having any problems on Windows. Both the cross-compilation from Fedora Linux and the native compilation on Windows server via GitHub actions complete and the latter also can successfully run the unit tests. As you can see from the changes to the preset files, the LEPTON package is included in both cases when using CMake.
https://ci.lammps.org/job/dev/job/pull_requests/job/compilation_tests/6643/display/redirect
https://github.com/lammps/lammps/actions/runs/3783646379

giacomofiorin:
> > I think it would be much simpler to keep the Lepton library unchanged, without changing or editing its files.
> 
> Compiling the Lepton library produces an excessive amount of compiler warnings. I won't take the changes back that produce them. Also, mid-term I would like to have its class structure set up to be properly abstract the details of the implementation from the use case and then avoid the memory corruption issues that I can into. Of course those would need to be eventually submitted to upstream.

I think that Peter would be open to merging those upstream changes fairly quickly (see [here](https://github.com/openmm/openmm/pull/2642), [here](https://github.com/Colvars/colvars/issues/231) and [here](https://github.com/openmm/openmm/pull/2290)).

If "eventually" means sometime before the next LAMMPS stable release, I don't have any major concerns. However, it would _not_ be a good idea to have multiple versions of the Lepton library around long-term.

> > A couple of changes may be needed to make this work on Windows, but that's an issue with Lepton itself
> 
> I was not having any problems on Windows. 

You're right: the problems would arise with dynamic linkage, but you have hard-coded the building of a static Lepton library in the build recipes, so if anybody wants to tweak them to use a DLL they're on their own.

akohlmey:
> You're right: the problems would arise with dynamic linkage, but you have hard-coded the building of a static Lepton library in the build recipes, so if anybody wants to tweak them to use a DLL they're on their own.

Not true. We test compilation with static and shared library and typically build on Windows creating a DLL. However, we're using a CMake feature to build the export lists automatically, so indicating static linkage works in either case. Traditional make does not support building with MSVC compilers. anyway.  GCC via Cygwin might work, but is unsupported.

akohlmey:
> If "eventually" means sometime before the next LAMMPS stable release, I don't have any major concerns. However, it would _not_ be a good idea to have multiple versions of the Lepton library around long-term.

First we need to get it to work well with LAMMPS. No point in pushing changes to upstream, if we still need to make changes.

giacomofiorin:
> > You're right: the problems would arise with dynamic linkage, but you have hard-coded the building of a static Lepton library in the build recipes, so if anybody wants to tweak them to use a DLL they're on their own.
> 
> Not true. We test compilation with static and shared library and typically build on Windows creating a DLL. However, we're using a CMake feature to build the export lists automatically, so indicating static linkage works in either case. Traditional make does not support building with MSVC compilers. anyway. GCC via Cygwin might work, but is unsupported.

But both the Lepton library and the LAMMPS LEPTON package code are in the same DLL, are they not?

BTW I see that you're redoing from scratch the changes to make Colvars use the Lepton copy from lib/lepton, so I deleted my branch at https://github.com/Colvars/lammps/tree/lepton-package.

akohlmey:
> But both the Lepton library and the LAMMPS LEPTON package code are in the same DLL, are they not?

Yes, they are. Building all the support libraries as DLLs (or .so or .dylib elsewhere) would be a giant mess.
Bascially, we don't provide suitable headers and link library stubs so you can compile and link your own application against the LAMMPS library on Windows, so there is not a problem. The DLL has to be sufficient to be imported into Python for the LAMMPS python module and the library DLL file approach is best suited for that.

akohlmey:
FWIW, the support for GNU make builds still has a few issues. This needs some more thinking and testing. I think the CMake approach of just including the LEPTON.cmake file and then protecting that from being included twice is superior.
In addition, one can make lepton a "link library" for colvars which imports all the defines including those for JIT support.

giacomofiorin:
> FWIW, the support for GNU make builds still has a few issues. This needs some more thinking and testing. I think the CMake approach of just including the LEPTON.cmake file and then protecting that from being included twice is superior. 

The traditional make route would be nice to have, but it also requires a good mechanism to deal with cross-dependency between libraries, not just between LAMMPS and them.

> In addition, one can make lepton a "link library" for colvars which imports all the defines including those for JIT support.

Exactly! Since you're making the effort to get asmjit to work in LAMMPS, there is really no reason not to reuse it.

akohlmey:
> The traditional make route would be nice to have, but it also requires a good mechanism to deal with cross-dependency between libraries, not just between LAMMPS and them.

It is not so bad since it is a) a one way dependency, i.e. colvars requires lepton, but not the other way around, and b) we know that that the Lepton sources are in place. When using "make lib-colvars" is is possible to define an explicit dependency on "make lib-lepton" and that triggers building both. The src/COLVARS/Install.sh script can include the settings for both lib/colvars and lib/lepton.

akohlmey:
@giacomofiorin I think I have the conventional build issues under control now. The only glitches are due to the automatic tests, which use "manual" compilation of the libraries instead of "make lib-<name>". I will try to put a workaround in place until this PR is merged.

giacomofiorin (approved):
Colvars library interoperability is preserved (regtests from https://github.com/Colvars/colvars pass). Approving.

akohlmey:
@giacomofiorin it looks like I have now found a solution to handle all permutations of installing or uninstalling the LEPTON and COLVARS package to update the Makefile.package settings that they always contain the required references to the include directories and libraries in the lib/lepton and lib/colvars folders. Also, the compilation settings are kept in sync via using a single included makefile. This is pushing the mechanisms in place to their limits, though. CMake is rather straightforward by comparison. :wink:

akohlmey:
> I think that Peter would be open to merging those upstream changes fairly quickly.

Just submitted a small change (strict aliasing rule violation workaround) as a test: https://github.com/openmm/openmm/pull/3882

Will submit others when this goes through.

akohlmey:
@giacomofiorin do you want *me* to update the COLVARS package build docs in the LAMMPS manual to reflect the changes here or do you prefer to do it yourself?

giacomofiorin:
> @giacomofiorin do you want _me_ to update the COLVARS package build docs in the LAMMPS manual to reflect the changes here or do you prefer to do it yourself?

I can do it, unless you'd rather do it for other reasons (in that case please state so clearly so that I don't do something that you'd end up redoing anyway)

akohlmey:
> I can do it, unless you'd rather do it for other reasons (in that case please state so clearly so that I don't do something that you'd end up redoing anyway)

I don't mind either way. In some sense, it will be easier for me, since I made the changes, but I want to make certain I am not stepping on your toes. On the other hand, you have a better sense for what needs to be communicated to the Colvars users and how you want to steer them toward compiling it. This is a significant change, so I am asking for your preference and will offer to do it, but not touch the colvars docs unless you give me the ok.

giacomofiorin:
> I don't mind either way. In some sense, it will be easier for me, since I made the changes, but I want to make certain I am not stepping on your toes. On the other hand, you have a better sense for what needs to be communicated to the Colvars users and how you want to steer them toward compiling it. This is a significant change, so I am asking for your preference and will offer to do it, but not touch the colvars docs unless you give me the ok.

Done. The fix doc needed some general refreshing as well anyway.

Absolutely no stepping on toes involved. From my experience the users of Colvars with LAMMPS are people who read the doc when they need. So the main guideline would be simply to make their life easier especially, when they are sysadmins that don't use the software themselves but are nonetheless asked to install it :-)

doc/src/pair_lepton.rst
```diff
@@ -0,0 +1,125 @@
...
+   pair_coeff  * *  "k*((r-r0)^2*step(r0-r)); k=200; r0=1.5" 2.0
+   pair_coeff  1 2  "4.0*eps*((sig/r)^12 - (sig/r)^6);eps=1.0;sig=1.0" 1.12246204830937
+   pair_coeff  2 2  "eps*(2.0*(sig/r)^9 - 3.0*(sig/r)^6);eps=1.0;sig=1.0"
+
+Description
+"""""""""""
+
+.. versionadded:: TBD
+
+Pair style *lepton* computes spherical pairwise interactions based on
```
sjplimp: replace "computes spherical pairwise ..." with something like

computes pairwise interactions between particles which depend solely on the distance
between the particles, for pairs of particles within the specified cutoff.  The potential function
(energy) must be specified as an expression ...

Am I correct this style cannot be used (for example) with potentials that depend on charge?

doc/src/pair_lepton.rst
```diff
@@ -0,0 +1,125 @@
...
+evaluating strings between neighboring atoms within the given cutoff.
+The potential function must be provided as an expression string using
+"r" as the distance variable. For example `"200.0*(r-1.5)^2"` represents
```
sjplimp: .. as the distance variable.  Note that additional constants in the expression can be defined in the string using semicolon separators; see the examples above.   For example "200 ...
(maybe have two versions of the same string with and without the additional constants)

sjplimp:
@akohlmey I think this is a cool new feature.

doc/src/pair_lepton.rst
akohlmey: > Am I correct this style cannot be used (for example) with potentials that depend on charge?

This specific style not, but I can imagine doing a variant that does. It is possible to have multiple "external variables" in a Lepton expression that can be set on a per-evaluation basis. But that comes with significant overhead for cases that do not use it. I have been thinking about interfacing other properties. For equal style variables, I have decided on doing a string replacement instead because that only needs to be done once and then the expression compiler can do more optimizations. Interfacing charges would be somewhat easier, but I can also imagine an interface to per-atom custom properties or atom-style variables. This mostly requires some conventions and possibly writing some extra C++ classes for the Lepton library implementing functions that access the data.

doc/src/pair_lepton.rst
akohlmey: done

doc/src/pair_lepton.rst
akohlmey: > Am I correct this style cannot be used (for example) with potentials that depend on charge?

Support for implementing coulomb interactions has been added as a new pair style `lepton/coul`. For this pair style you can reference either "qi" or "qj" within the expression to get access to the charge. More details and the differences between this style and the plain `lepton` one are in the docs. As you may see from the code, while Lepton has convenient ways to set values in such expressions, they come with some overhead and because the expression is evaluated for each pair, there is only a limited amount of the processing that can be done outside the innerloop. Trying to access per-atom variables or properties would be even more complex (global variables are simple in comparison and can be done with a string replacement just as like in input files and `${variable}` expansions).

akohlmey: @sjplimp FYI compatibility with kspace styles is implemented the same way as for pair style table. I have been able to create tables with this pair style that agree with those for coul/cut and coul/long within 1.0e-14 in forces and energies.

sjplimp (approved):

PR #3574, opened 2022-12-25

akohlmey:
**Summary**

This pull request collects various small changes and fixes

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

Calculations using dihedral style table will report different stress/pressure.

**Implementation Notes**

- fix bug in stress tally for dihedral style table
- Small documentation improvements and corrections
- improved error messages
- enable and apply clang-format to selected files or parts of files
- make created table files better suitable for human consumption
- add unit test section for the extract() function in angle style tester
- fix unit test for angle style amoeba
- fix issue in the ML-POD package that led to inconsistent results on different platforms
- add logic to `rerun` command, so that dump are triggered on expected steps in case the sequence of timesteps read in advances faster than what the dump has set up. This is needed due to the new logic to allow for time based outputs. This includes variable based output, which was never properly implemented for rerun.
- add a few more unit tests for timestepping fixes
- update LAMMPS' Sphinx base theme from the read-the-docs theme v1.0.0 to v1.1.1
- prevent installation of Sphinx version 6.0.0 for building the manual. This will break out theme so we'll need to wait until there is a compatible update and then adapt the LAMMPS theme
- set canonical URL for the manual to always point to https://docs.lammps.org to make Google & Co. happy.
- provide explicit force function for harmonic angle tabulation example to show need for applying unit conversion to energy per radian squared.

**Post Submission Checklist**


src/dump.cpp
```diff
@@ -231,16 +232,22 @@ void Dump::init()
     ids = idsort = nullptr;
     index = proclist = nullptr;
     irregular = nullptr;
+    if ((has_id == 0) && (me == 0))
+      error->warning(FLERR,"Dump {} is without Atom-IDs and not sorted by id. This may lead to data "
```
sjplimp: I suggest this wording (here and below):
Dump {} includes no atom IDs and is not sorted by ID.  This may complicate post-processing tasks or visualization

akohlmey:
@stanmoore1 this should now be easier to digest.

stanmoore1 (approved):
I approve

PR #3575, opened 2022-12-26
Correct handling of random force with exclusions in DPD pair styles

### LINKED ISSUE #3563: DPD with special bonds
**Summary**

Using non-standard special_bonds weights in DPD packages breaks the fluctuation-dissipation theorem and thus samples wrong temperatures.

**Expected Behavior**

Correct temperatures with special bond weights which are not 0 or 1.

The problem is, that the weighting factor (called dpd_factor within the code) scales the complete forces. This is consisted with how normal LJ interactions are treated, but if the dpd_factor is not 0.0 or 1.0, it breaks the fluctuation dissipation theorem, as dissipative forces and random forces would have to be scaled differetly. This can be very easily solved by  not scaling the total forces, but the dissipative force by dpd_factor and the random force by sqrt(dpd_factor). 

While this case seems to be niche, it is easily solvable and would improve the usage of special bonds in DPD packages.

akohlmey:
**Summary**

This addresses the concern about the random force scaling factor for exclusion factors other than 0.0 and 1.0.

**Related Issue(s)**

Fixes #3563 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

The results for the following DPD pair styles (dpd, dpd/tstat, dpd/ext, dpd/ext/tstat) and their accelerated variants will change for exclusion factors other than 0.0 or 1.0 (where sqrt(x) = x).

**Implementation Notes**

To avoid having to call sqrt() many times during a force compute, the corresponding scaling factors are pre-computed and stored with the pair style.

Packages completed:
- [x] DPD-BASIC
- [x] OPENMP
- [x] GPU
- [x] KOKKOS
- [x] INTEL

**Post Submission Checklist**


ndtrung81 (approved):
@akohlmey everything looks good to me. Thanks!

src/KOKKOS/pair_dpd_ext_kokkos.cpp
```diff
@@ -41,8 +41,8 @@ using namespace LAMMPS_NS;
 
 
 template<class DeviceType>
-PairDPDExtKokkos<DeviceType>::PairDPDExtKokkos(class LAMMPS *lmp) :
-  PairDPDExt(lmp) ,
+PairDPDExtKokkos<DeviceType>::PairDPDExtKokkos(class LAMMPS *_lmp) :
+  PairDPDExt(_lmp) ,
 #ifdef DPD_USE_RAN_MARS
   rand_pool(0 /* unused */, lmp)
```
stanmoore1: @akohlmey is changing `lmp` to `_lmp` necessary? Looks like this change was missed here (for `#ifdef DPD_USE_RAN_MARS`)

stanmoore1 (approved):

src/KOKKOS/pair_dpd_ext_kokkos.cpp
akohlmey: It is not a crucial change, but it reduces false positives with several code analysis tools that complain about the "lmp" parameter hiding the "lmp" class member. In 99.9% of the cases in LAMMPS it does not affect functionality, but as we are using the initializer lists more the risk to have undefined or unexpected behavior grows. So it is a change that I do whenever I notice it (it is too boring to do for the entire code base in one big change). Sorry about missing the additional entries.

PR #3576, opened 2022-12-27
Updates to the DIELECTRIC package

ndtrung81:
**Summary**

This PR includes maintenance changes to the pair and kspace dielectric styles to ensure consistency with non-dielectric counterparts.

**Related Issue(s)**

N/A

**Author(s)**

Trung Nguyen (UChicago)

**Licensing**

**Backward Compatibility**

Should maintain backward compatibility.

**Implementation Notes**

Previously the pair and kspace styles use the atom charges (atom->q) scaled by local epsilon's. This PR renames the variables so that atom->q has consistent meaning with other packages as unscaled values. The pair and kspace dielectric styles now operate on the scaled values instead. There are also changes to the pppm/dielectric to match the elong value given by the scaled charges with the elong value given by regular pppm with a global dielectric constant. This is a nontrivial issue as electrostatic energy is not linearly dependent upon charge density.

The documentation stays unchanged with the changes in this PR, unless suggested otherwise.

**Post Submission Checklist**


akohlmey:
@ndtrung81 This PR needs updates to the unit tests. If you don't mind, I will do them.

ndtrung81:
@akohlmey not at all, please help me with the unit tests updates. Thanks!

akohlmey:
> @akohlmey not at all, please help me with the unit tests updates. Thanks!

@ndtrung81 turns out the situation is not as simple as I thought. There are some conditions, e.g. after a restart, where the results differ where they should not. Will need to take a closer look.

ndtrung81:
@akohlmey were the failed unit tests on MacOS and Windows, or regardless? I tried to run after restart, and read from a restart file, on my Fedora box, the results seem to match in my tests.

akohlmey:
> @akohlmey were the failed unit tests on MacOS and Windows, or regardless? I tried to run after restart, and read from a restart file, on my Fedora box, the results seem to match in my tests.

There were failures everywhere. What the unit test programs do is a bit more complex: they run for a few steps, then use `clear` and read a restart or data file and start over with different settings and do this multiple times. The failures were in pppm/dielectric and pppm/disp/dielectric. The runs are special also in that they either turn off pair or kspace completely (with either `pair_modify compute no` or `kspace_modify compute no` so that even if a matching pair/kspace style is required to run the tests, it will not contribute to forces, energies, and stress. I have checked with valgrind and there was no uninitialized memory access. So I am a bit puzzled about how to trigger the situation outside the test tools.

akohlmey:
@ndtrung81 I think I may have identified the cause of the unit test failures. Please see commit 52b84c97768ba0aabb9a5a4b34157f95430e06fc

I now understand why I was so confused by the failures, since it was the original computation that was off due to the uninitialized class member, while the restart were resulting in correct data.

akohlmey:
@ndtrung81 do you have any pending changes or do you want to make more tests?
If not, please assign the pull request to me and post a note that reviewers should have a look at it.

ndtrung81:
@akohlmey thanks, I just pushed the commit for some last-minute edits. The PR is ready for review now.

stanmoore1 (approved):
I approve

PR #3577, opened 2022-12-28
Have PyTorch interface for MLIAP working in Kokkos.  This uses cuPy a…

bathmatt:
…nd a simple example is provided

**Summary**

The ML-IAP interface had a python interface and code to connect it to PyTorch.  This was only on the CPU,  Since the ML-IAP pair style was ported to Kokkos, this model was also ported to Kokkos in this PR. 

**Related Issue(s)**

**Author(s)**
Matt Bettencourt NVIDIA mattb@nvidia.com

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/python/lammps/mliap/__init__.py
+++ b/python/lammps/mliap/__init__.py
@@ -31,5 +31,8 @@ if not pylib.Py_IsInitialized():
                   "in undefined behavior.")
 else:
     from .loader import load_model, load_unified, activate_mliappy
-
+    try:
+         from .loader import  load_model_kokkos,  activate_mliappy_kokkos
+    except: 
+        pass
 del sysconfig, ctypes, library, pylib
--- a/python/lammps/mliap/__init__.py
+++ b/python/lammps/mliap/__init__.py
@@ -33,6 +33,6 @@ else:
     from .loader import load_model, load_unified, activate_mliappy
     try:
          from .loader import  load_model_kokkos,  activate_mliappy_kokkos
-    except: 
+    except:
         pass
 del sysconfig, ctypes, library, pylib
--- a/python/lammps/mliap/loader.py
+++ b/python/lammps/mliap/loader.py
@@ -56,7 +56,6 @@ class DynamicLoader(importlib.abc.Loader):
 
 def activate_mliappy(lmp):
     try:
-        print("activate_mliappy")
         library = lmp.lib
         module_names = ["mliap_model_python_couple", "mliap_unified_couple"]
         api_version = library.lammps_python_api_version()
@@ -75,7 +74,6 @@ def activate_mliappy(lmp):
 
 def activate_mliappy_kokkos(lmp):
     try:
-        print("activate_mliappy_kokkos")
         library = lmp.lib
         module_names = ["mliap_model_python_couple_kokkos"]
         api_version = library.lammps_python_api_version()
@@ -94,7 +92,6 @@ def activate_mliappy_kokkos(lmp):
 
 def load_model(model):
     try:
-        print("load_model")
         import mliap_model_python_couple
     except ImportError as ie:
         raise ImportError("ML-IAP python module must be activated before loading\n"
@@ -104,7 +101,6 @@ def load_model(model):
 
 def load_model_kokkos(model):
     try:
-        print("load_model_kokkos")
         import mliap_model_python_couple_kokkos
     except ImportError as ie:
         raise ImportError("ML-IAP python module must be activated before loading\n"
--- a/python/lammps/mliap/loader.py
+++ b/python/lammps/mliap/loader.py
@@ -56,6 +56,7 @@ class DynamicLoader(importlib.abc.Loader):
 
 def activate_mliappy(lmp):
     try:
+        print("activate_mliappy")
         library = lmp.lib
         module_names = ["mliap_model_python_couple", "mliap_unified_couple"]
         api_version = library.lammps_python_api_version()
@@ -72,14 +73,45 @@ def activate_mliappy(lmp):
     except Exception as ee:
         raise ImportError("Could not load ML-IAP python coupling module.") from ee
 
+def activate_mliappy_kokkos(lmp):
+    try:
+        print("activate_mliappy_kokkos")
+        library = lmp.lib
+        module_names = ["mliap_model_python_couple_kokkos"]
+        api_version = library.lammps_python_api_version()
+
+        for module_name in module_names:
+            # Make Machinery
+            loader = DynamicLoader(module_name,library,api_version)
+            spec = importlib.util.spec_from_loader(module_name,loader)
+
+            # Do the import
+            module = importlib.util.module_from_spec(spec)
+            sys.modules[module_name] = module
+            spec.loader.exec_module(module)
+    except Exception as ee:
+        raise ImportError("Could not load ML-IAP python coupling module.") from ee
+
 def load_model(model):
     try:
+        print("load_model")
         import mliap_model_python_couple
     except ImportError as ie:
         raise ImportError("ML-IAP python module must be activated before loading\n"
                           "the pair style. Call lammps.mliap.activate_mliappy(lmp)."
                           ) from ie
     mliap_model_python_couple.load_from_python(model)
+
+def load_model_kokkos(model):
+    try:
+        print("load_model_kokkos")
+        import mliap_model_python_couple_kokkos
+    except ImportError as ie:
+        raise ImportError("ML-IAP python module must be activated before loading\n"
+                          "the pair style. Call lammps.mliap.activate_mliappy(lmp)."
+                          ) from ie
+    mliap_model_python_couple_kokkos.load_from_python(model)
+
 
 def load_unified(model):
     try:
@@ -89,3 +121,4 @@ def load_unified(model):
                           "the pair style. Call lammps.mliap.activate_mliappy(lmp)."
                           ) from ie
     mliap_unified_couple.load_from_python(model)
+
--- a/src/KOKKOS/mliap_model_python_couple_kokkos.pyx
+++ b/src/KOKKOS/mliap_model_python_couple_kokkos.pyx
@@ -1,0 +1,157 @@
+# cython: language_level=3
+# distutils: language = c++
+
+cimport cython
+
+from libc.stdint cimport uintptr_t
+
+import pickle
+
+# For converting C arrays to numpy arrays
+import numpy
+import torch 
+import cupy
+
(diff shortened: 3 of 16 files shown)


bathmatt:
I built this code using the nvcr.io/nvidia/pytorch:22.04-py3 container.

bathmatt:
@lubbersnick finally got approval, you might wanna take a look

src/KOKKOS/mliap_model_python_couple_kokkos.pyx
```diff
@@ -0,0 +1,157 @@
...
+cimport cython
+
+from libc.stdint cimport uintptr_t
+
+import pickle
+
+# For converting C arrays to numpy arrays
+import numpy
+import torch 
+import cupy
```
akohlmey: This unconditionally loads "cupy". I don't see any check for it anywhere.
What will happen if KOKKOS was compiled without CUDA? E.g. configured for HIP or SYCL or OpenMP?

bathmatt: Good catch, Will fix.

> E.g. configured for HIP or SYCL or OpenMP
Well, I could say "then you're doing it wrong" :)

bathmatt: Added try block for importing cupy.  It will work on GPU for Cuda and do all the CPU/GPU data copies for HIP/SYCL as this is out of my expertise.  To be fair, python is already out of my expertise

stanmoore1 (approved):
I approve

akohlmey:
@stanmoore1 have you been able to confirm compilation with traditional make and CUDA? I will have to take a closer look at the CMake parts. There are some details that I am not happy after a cursory look.

stanmoore1:
> @stanmoore1 have you been able to confirm compilation with traditional make and CUDA? I will have to take a closer look at the CMake parts. There are some details that I am not happy after a cursory look.

I have not tried running this yet, just scanned the code changes.

python/lammps/mliap/__init__.py
```diff
@@ -31,5 +31,8 @@
                   "in undefined behavior.")
 else:
     from .loader import load_model, load_unified, activate_mliappy
-
+    try:
```
akohlmey, line 34: Rather than depending on the package load failing and catching the exception, this should use the available introspection to see if the KOKKOS package is included in LAMMPS and only then try to load the modules.

python/lammps/mliap/loader.py
```diff
@@ -56,6 +56,7 @@ def exec_module(self, module):
 
 def activate_mliappy(lmp):
     try:
+        print("activate_mliappy")
```
akohlmey: Please remove debug code.

python/lammps/mliap/loader.py
```diff
@@ -72,15 +73,46 @@ def activate_mliappy(lmp):
     except Exception as ee:
         raise ImportError("Could not load ML-IAP python coupling module.") from ee
 
+def activate_mliappy_kokkos(lmp):
+    try:
+        print("activate_mliappy_kokkos")
```
akohlmey: More debug code

python/lammps/mliap/loader.py
```diff
@@ -72,15 +73,46 @@ def activate_mliappy(lmp):
...
         import mliap_model_python_couple
     except ImportError as ie:
         raise ImportError("ML-IAP python module must be activated before loading\n"
                           "the pair style. Call lammps.mliap.activate_mliappy(lmp)."
                           ) from ie
     mliap_model_python_couple.load_from_python(model)
 
+def load_model_kokkos(model):
+    try:
+        print("load_model_kokkos")
```
akohlmey: and another debug line.

src/KOKKOS/pair_mliap_kokkos.cpp
```diff
@@ -165,6 +176,8 @@ void PairMLIAPKokkos<DeviceType>::settings(int narg, char ** arg)
     } else
       iarg++;
   }
+  PairMLIAP::settings(narg, arg);
```
akohlmey: I am a bit confused by seeing this statement here with unmodified arguments. I would have expected that any options that are parsed in this class would be removed from the list of arguments before passing it to the base class for further processing.

src/ML-IAP/mliap_model_python.cpp
```diff
@@ -66,17 +70,18 @@ MLIAPModelPython::MLIAPModelPython(LAMMPS *lmp, char *coefffilename) :
...
 }
 
 /* ---------------------------------------------------------------------- */
 
 MLIAPModelPython::~MLIAPModelPython()
 {
-  MLIAPPY_unload_model(this);
+  if (model_loaded)
+    MLIAPPY_unload_model(this);
+  model_loaded=false;
```
akohlmey: model_loaded is an integer variable not a boolean.

src/ML-IAP/pair_mliap.cpp
```diff
@@ -131,52 +131,53 @@ void PairMLIAP::settings(int narg, char ** arg)
...
-  delete model;
-  model = nullptr;
-  delete descriptor;
-  descriptor = nullptr;
-
   // process keywords
 
   int iarg = 0;
 
+  //Check to see if there are more than one model or descriptor
```
akohlmey: This looks very "hack-ish" and I am not convinced that this is a good approach. See my comment for the ::settings() function in the KOKKOS version.

src/lammps.cpp
```diff
@@ -442,7 +442,11 @@ LAMMPS::LAMMPS(int narg, char **arg, MPI_Comm communicator) :
       iarg += 3;
       while (iarg < narg && arg[iarg][0] != '-') iarg++;
 
-    } else error->universe_all(FLERR,"Invalid command-line argument");
+    } else {
+      std::string errmsg("Invalid command-line argument");
```
akohlmey:
```suggestion
      std::string errmsg("Unknown pair style mliap argument: ");
```

src/lammps.cpp
```diff
@@ -442,7 +442,11 @@ LAMMPS::LAMMPS(int narg, char **arg, MPI_Comm communicator) :
...
+      errmsg += arg[iarg];
+      error->universe_all(FLERR,errmsg.c_str());
```
akohlmey:
```suggestion
      error->universe_all(FLERR,errmsg);
```

Alternately, it should be possible to use just a single line like this:
 `error->universe_all(FLERR, fmt::format("Unknown pair style mliap argument: {}", arg[iarg]));`

akohlmey (changes requested):
I've simplified, updated, and corrected the CMake and GNU make build support so that it should now work as expected.
There are some comments about code that looked like it could be done more cleanly.
I have not yet compiled and tested in detail with a debugger. But I thought it would be better to just show my concerns and allow for updates before I have to dig in deep myself, which is always more effort for a person that has not written the code in question.

Also, there are no updates to the documentation that document what this enables and what its restrictions are.

python/lammps/mliap/__init__.py
bathmatt: Do we have an example of this somewhere I can copy?

python/lammps/mliap/loader.py
bathmatt: removed

python/lammps/mliap/loader.py
bathmatt: removed

python/lammps/mliap/__init__.py
akohlmey: There is an example in `unittest/python/python-pizza.py`. But after having a closer look at this and making some tests, I have to conclude that you have no alternative here. This code is run *before* you have access to an instance of the LAMMPS class and thus access to the loaded LAMMPS shared library and that does not give you access to the compiled in features. The cure would be worse than the disease. :-(

src/lammps.cpp
bathmatt: Went with alternate.  cleaner.  Old habits die hard.

src/ML-IAP/mliap_model_python.cpp
bathmatt: fixed

src/KOKKOS/pair_mliap_kokkos.cpp
bathmatt: I have cleaned up the parsing,  Getting approval on my latest changes from a NVIDIA standpoint.  Should have a the last two issues resolved by early PST.

bathmatt: This has been pushed.  @akohlmey if you are happy I think if you remove changes requested I can push once the tests pass.

bathmatt:
Fixed cut-n-paste error.  Failing test now passes

```
Test project /home/mattb/lammps/LAMMPS/omp-test
    Start 3: InvalidFlag
1/1 Test #3: InvalidFlag ......................   Passed    0.35 sec
```

akohlmey (approved):

bathmatt:
I can't merge, is merging into develop the correct branch?  It says blocked  

The base branch restricts merging to authorized users. Learn more about protected branches.

and 

You’re not authorized to merge this pull request

@akohlmey or @stanmoore1 can you press the button?  Or let me know what branch to merge into?

akohlmey:
> I can't merge, is merging into develop the correct branch? It says blocked

You are not supposed to merge anything. "develop" is the correct branch, but it is protected. Only core developers are permitted to merge.  

There is a "policy" document here describing the process (for users and for us to have it written down in case there is a disagreement about what is compliant and not) https://github.com/lammps/lammps/blob/develop/doc/github-development-workflow.md

akohlmey:
@bathmatt just to clarify the process further. Since the pull request is currently assigned to @stanmoore1, he has the choice of what to do with it next. Usually this means that he needs to run some (manual) tests to confirm that all existing functionality is preserved and that the new code compiles and runs on the KOKKOS relevant platforms. Once he is statisfied (or has concluded that the change does not justify the time and effort) the request will be assigned to me and then I will merge it at the time when it is convenient to merge. We had for a while the policy that anybody could merge, but that would occasionally cause problems because requests would be merged in an order that would create avoidable merge conflicts or would affect the stability of the "develop" branch. As a consequence, pull requests may be sitting in the queue for longer than strictly necessary because we all have high priority projects besides LAMMPS. It is worth it since it helps us to keep our sanity. :-)

stanmoore1:
@bathmatt I removed my assignment and assigned @akohlmey.

PR #3579, opened 2022-12-28
Convert linalg library from Fortran to C++

akohlmey:
**Summary**

The lib/linalg folder contains a collection of BLAS/LAPACK routines that are used by various packages in LAMMPS.
This library serves as a fallback when the OS does not have an optimized BLAS/LAPACK library available.
However, it would still limit portability of LAMMPS because not all target platforms have a compatible Fortran compiler installed or do not support Fortran at all. By translating the sources to C++ with the help of f2c it can be compiled with the same C++ compiler that is required to compile LAMMPS.  This applies for example to Windows with MSVC and MacOS with the native compilers.

**Related Issue(s)**

N/A.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues (for now).

**Implementation Notes**

BLAS/LAPACK files are organized to be highly portable and are mostly written in a plain Fortran 77 style. Code requiring system routines or calling out to the Fortran runtime are few. Those functions were reimplemented in C++ using the corresponding C++ runtime features. A few files had to be modified (use of 'recursive subroutine' and 'exit' which are not f2c compatible) so that f2c can translate them. F2c can make code compatible with recursive calls via a command line flag, and 'exit' can be replaced with 'goto'.  A few of the f2c runtime functions are adapted and included. Since libgfortran is also based on f2c it uses some of the same functions. To avoid conflicts during linking with Fortran libraries, our versions have been renamed.

**Post Submission Checklist**


akohlmey:
Question for the reviewers: should we keep the files in the "lib/linalg/fortran" and "lib/linalg/static" folders? Technically, they are only required to recreate the C++ sources. This means they are in some way redundant and not really needed unless somebody needs to add a missing LAPACK function and add/convert it to C++. I could just build a separate git repository from those and add a link to that to the README. Dropping them should make the (every growing) tarball a bit smaller... :wink:

Update: I just went ahead and did it. Conversion and original Fortran is now available at [the linalg repository](https://github.com/lammps/linalg/)

sjplimp:
@akohlmey can you clarify ...
a) you are saying you replaced the Fortran files in lib/linalg with C versions you created yourself?
b) the original Fortran versions are available somewhere else (user is pointed to them) and the on-line
    versions are identical to what LAMMPS had, or there might have been mods in the Fortran versions ?
c) the make and CMake which needs LAPACK source (as opposed to available libs) now uses the C versions
d) is there any chance the new C files will have bugs or different behavior than the previous Fortran ?
e) perf is the same (or if not, doesn't matter since using a native lib is the best) ?

akohlmey:
> @akohlmey can you clarify ... 

> a) you are saying you replaced the Fortran files in lib/linalg with C versions you created yourself?

Yes, created by converting them from Fortran to C++ with the f2c program, plus write a few support functions in C++ directly (like XERBLA, which prints error messages).

> b) the original Fortran versions are available somewhere else (user is pointed to them) and the on-line versions are identical to what LAMMPS had, or there might have been mods in the Fortran versions ?

The C++ files are exact replacements of the corresponding Fortran sources. In a few cases I had to make minor changes because the Fortran sources were using constructs that f2c does not support (E.g. EXIT in a loop, so I replaced it with GOTO/CONTINUE). The Fortran sources *with* the conversion scripts are in https://github.com/lammps/linalg.  So if we want to update the LAPACK version or add a new function, we can just check out that repo, add or replace the fortran file and copy the resulting C++ sources back to LAMMPS.

> c) the make and CMake which needs LAPACK source (as opposed to available libs) now uses the C versions

Once you have the binary of the compiled library, it will look to the linker exactly like the compiled Fortran version, only that it does not depend on the Fortran runtime library. It is pure C++/C.

> d) is there any chance the new C files will have bugs or different behavior than the previous Fortran ?

f2c is a very mature project. It was all there was in terms of Fortran on Linux where the was not Portland Group Fortran compiler or GNU g77. It has been reliable for a lot of software back at the time (some 25 years ago) and I don't think it got worse. Not considering the kind of people that maintain it. On top of that, I have adapted some of the tester programs that come with BLAS/LAPACK to only use functions we provide. Those tests have passed on multiple platforms. BTW, the testers are still using Fortran. ;-)

> e) perf is the same (or if not, doesn't matter since using a native lib is the best) ?

I am not aware of any place in LAMMPS where the BLAS/LAPACK performance matters.  It is usually used in places where people need an Eigensolver and don't want to write one of their own.  I would expect that for sufficiently large problems, the overhead versus the Fortran files compiled with a native Fortran compiler will be in the neighborhood of 10-15%. This is against the reference BLAS/LAPACK implementation, which is significantly slower than OpenBLAS or MKL or equivalent. In the end the same argument that I usually make when people are concerned about FFTW vs. KISS FFT. 10-20% of something that is a small fraction of the total is not worth worrying about.  Besides, the point of even bundling the Fortran version is to enable people to use the packages that depend on BLAS/LAPACK, even if they don't have a suitable optimized version available. With translating them to C++ we grew the number of people that can use those packages to those that also do not have a Fortran compiler.

sjplimp:
@akohlmey thanks

sjplimp (approved):

PR #3582, opened 2022-12-31
Sllod variants

athomps:
**Summary**

Added psllod keyword to toggle between SLLOD and p-SLLOD variants.

**Related Issue(s)**

This distinction between SLLOD and p-SLLOD  was originally raised and partially explained here https://matsci.org/t/question-about-sllod-implementation/32166 

**Author(s)**

Aidan Thompson, Pieter J. in 't Veld, James P. Ewen

**Licensing**

**Backward Compatibility**

No change for pure shear flow, including `examples/nemd`

**Implementation Notes**

n/a

**Post Submission Checklist**


akohlmey:
There are SLLOD fixes in the EFF, INTEL, KOKKOS, and OPENMP packages. Don't they need to be updated accordingly?

athomps:
> There are SLLOD fixes in the EFF, INTEL, KOKKOS, and OPENMP packages. Don't they need to be updated accordingly?

Done, not compiled or tested.

athomps:
@akohlmey Thanks for completing the package updates.  I am able to compile packages now and it all looks good.

akohlmey:
> @akohlmey Thanks for completing the package updates. I am able to compile packages now and it all looks good.

You're welcome. I've made some small code modernizations and added unit test inputs. Strangely enough the output for the two settings is identical. What conditions or input settings beyond the different value of the keyword are required to make a difference?

The test inputs are currently using:
```
  change_box all triclinic
  fix all all deform 2 xy final 2.0 remap v
  fix test all nvt/sllod temp 50.0 ${t_target} 1.0
```
and
```
  change_box all triclinic
  fix all all deform 2 xy final 2.0 remap v
  fix test all nvt/sllod temp 50.0 ${t_target} 1.0 psllod yes
```

athomps:
The effect of the call to `remove_bias()` on `vdelu` calculation is mathematically and numerically zero for pure shear strain rate tensor, as explained [here](https://matsci.org/t/question-about-sllod-implementation/32166).  I did my own check and found for `examples/nemd`:
`h_two = 0 0 0 0 0 0.01`
`MathExtra::multiply_shape_shape(h_two,h_two) = 0 0 0 0 0 0`
A non-zero effect occurs only if there is an extensional component e.g.:
`h_two = e -e 0 0 0 0`
`MathExtra::multiply_shape_shape(h_two,h_two) = e^2 e^2 0 0 0 0`

athomps:
The effect of this change for extensional flow was also mentioned by @ssande7 here:

https://github.com/CTCMS-UQ/lammps/pull/11

Maybe @ssande7 or some other interested party could provide a suitable modification of `examples/nemd` that shows the effect, perhaps also a measure of the difference in energy dissipation.

athomps:
For a simple test, this change to `examples/nemd` will be sufficient:
```
24,25c24,25
< fix           1 all nvt/sllod temp 1.0 1.0 1.0 tchain 1 psllod 1
< fix           2 all deform 1 xy erate 0.01 yz erate 0.01 remap v
---
> fix           1 all nvt/sllod temp 1.0 1.0 1.0 tchain 1
> fix           2 all deform 1 xy erate 0.01 remap v
```

ssande7:
@athomps here's an input script modified from the nemd example that keeps track of energy dissipation for either shear or elongational flow (the `psllod` flag should only make a difference in the latter case). There are some minor issues with storing the initial energy that I haven't had time to dig into properly, but it should work as a starting point at least. I get better accuracy writing out the expected energy dissipation each timestep and integrating that in post processing.

I should note though that in my own testing I found that the implementation in this PR (and in current stable LAMMPS) doesn't quite give the correct energy dissipation even under simple shear flow, since the equations of motion are not reversible. The position at time `t` is used to calculate the velocity bias, and then this is added back at time `t+dt/2` (under the Verlet integrator), but the actual streaming velocity at that time would depend on the position at `t+dt/2`. I've solved this in our working branch in the PR you linked by switching to treating the velocities as peculiar and using a reversible integration scheme, and this has the added bonus of no longer needing to manually perform the initial velocity kick.

There are also some issues with fix deform that become particularly problematic under mixed flows. Since fix deform does not account for the interplay between different shear/elongation rates, the `h_rate` values it sets are incorrect in these cases, meaning that when `MathExtra::multiply_shape_shape(h_rate,h_inv,h_two)` is called to recover the input deformation rates, those are also incorrect. I've fixed this in our working branch as well, but it's rather unwieldy and in my opinion would be better achieved by integrating the box directly (similar to `fix nvt/uef`) rather than re-solving for its shape every timestep (and this would also open up opportunity for a `fix npt/sllod` implementation).

I was planning to separate the core SLLOD fixes from the molecular SLLOD code once I get back from holidays next week and submit that as a PR (it also currently includes a psllod flag), but happy to submit it to this branch instead if you'd prefer.

```
# 2d NEMD simulation

units		lj
timestep 0.001
atom_style	atomic
dimension	2

lattice		sq2 0.8442
region		box prism 0 10 0 30 -0.5 0.5 0 0 0
create_box	2 box
create_atoms	1 box
mass		* 1.0

velocity	all create 1.0 87287 loop geom

region		slice block 4 6 INF INF INF INF
set		region slice type 2

pair_style	lj/cut 2.5
pair_coeff	* * 1.0 1.0
pair_modify shift yes

neighbor	0.3 bin
neigh_modify	delay 0 every 1


fix		1 all nvt/sllod temp 1.0 1.0 0.1 tchain 1 psllod no

# ELONGATION
# variable eps index 0.01
# fix		2 all deform 1 x trate ${eps} y trate -${eps} remap v
# variable vx_new atom vx+x*${eps}
# variable vy_new atom vy-y*${eps}
# velocity all set v_vx_new v_vy_new NULL

# SHEAR
variable gamma index 0.01
fix		2 all deform 1 xy erate ${gamma} remap v
variable vx_new atom vx+y*${gamma}
velocity all set v_vx_new NULL NULL

# Get initial energy
thermo_style custom step temp pxx pyy pxy vol etotal
thermo_modify norm no temp 1_temp
run 1 # can't use run 0 since grad u isn't set correctly during init
variable Einitial index $(etotal)
reset_timestep 0

#dump		1 all custom 1000 dump.nemd id type x y z

#dump		2 all image 1000 image.*.jpg type type adiam 1.2
#dump_modify	2 pad 5

#dump		3 all movie 1000 movie.mpg type type adiam 1.2
#dump_modify	3 pad 5

# Calculate expected energy dissipation rate based on -vol * P^T:∇u
# ELONGATION:
# variable Edot_sllod equal -vol*${eps}*(c_thermo_press[1]-c_thermo_press[2])

# SHEAR:
variable Edot_sllod equal -vol*${gamma}*c_thermo_press[4]

# Thermostat component
variable Edot_alpha equal -(c_1_temp[1]+c_1_temp[2])*f_1[2]

# Total energy dissipation rate
variable Edot_expected equal v_Edot_sllod+v_Edot_alpha

# Integrate expected energy dissipation to calculate expected total energy.
fix running all ave/time 1 1 1 v_Edot_expected ave running
variable TotEng_expected equal v_Einitial+f_running*elapsed*dt

thermo_style custom step temp pxx pyy pxy vol etotal v_TotEng_expected
thermo_modify norm no temp 1_temp
thermo		1000

run		50000
```

athomps:
@ssande7 Thanks for the quick response.  I am hoping that the additional issues with energy dissipation in fix nvt/sllod (half-step errors in velocity, box dimensions) are small enough that the method is still useful without further correction.  If you see an opportunity to make simple changes to address some of these issues, it would be good to include them in this PR.   I like the test that you have provided. I have not tried it yet, but I have a couple of questions.

1. Is the energy dissipation expected to match a known macroscopic value of -vol * P^T:∇u?  If so, how close should the match be?
2. I don't think the units of `Edot_sllod` and `Edot_alpha` having been converted to energy/time. Is this because no conversion is needed for LJ units?

stanmoore1 (approved):

ssande7:
@athomps no worries. The fixes aren't super simple unfortunately, and are required for correctness (see below). I'll aim to submit them as a separate PR with some more details later next week/the week after and we can take things from there (I'll be without internet for the rest of this week).

With regards to your questions:
1. One of the important properties of the sllod algorithm is that the energy dissipation rate exactly matches that macroscopic value. In practice there will be some numerical error, and in that input script with my corrections applied I see variations of around 0.5 order of magnitude, but I think that's mostly due to the rudimentary integration of the energy dissipation rate since I get a much closer agreement when integrating it in post processing. Without reversible equations of motion, the energy dissipation will drift away from the expected value, which is a major issue for those (for example) wanting to perform time-transient correlation function (TTCF) calculations that are quite sensitive to this. It also matters for viscosity calculations, although the difference in that case is probably minor. In the case of p-sllod, the energy dissipation rate will only match that macroscopic value for pure shear flow, and this is one of the criticisms of that algorithm. Additionally, without the corrections to fix deform (which are the most complicated of the changes), the simulation box does not deform correctly under mixed flows (even some pure shear flows with multiple shear directions), and the h_rate matrix can also be incorrect, both of which cause (sometimes very) incorrect energy dissipation, and can lead to the sllod code using a shear rate that is significantly different from the one specified by the user.
2. That's correct that no conversion factor is required for LJ units.

akohlmey:
@athomps How do you want to proceed with this PR?

I have updated the commands in the unit test inputs according to the quoted examples and the two variants still come out the same. If this is supposed to be like this, then it is good enough for a unit test, even though having a modification that creates a difference between the two variants would be preferred.

Otherwise, it seems that from the technical part of implementing the psllod flag itself, this seems to be ready now.

stanmoore1:
I fixed a bug in the Kokkos version on GPUs, so that needs to get merged in any case (even if I need to submit a new PR).

akohlmey:
> I fixed a bug in the Kokkos version on GPUs, so that needs to get merged in any case (even if I need to submit a new PR).

We can always cherry-pick it into #3574, it it is still around by that time. But I don't mind merging this as is and then having further changes in a separate pull request. There are obviously more changes coming from @ssande7, if I understood correctly.

At the moment, I just want to get as much of the pending stuff merged as possible quickly, since we have almost 10 pull requests that are either ready and just need an approval or are close to ready and just need a few small tweaks and checks. Since there is a bunch of new code among it, I would like to run it through static code analysis and clang-tidy afterwards and apply the necessary fixes (there always are some small oversights that are caught by that). Christmas break has been rather productive this time... ;-)

athomps:
I am happy with the current test, even though `psllod` has no effect on it. This PR is complete.

PR #3585, opened 2023-01-03
Feature/pace extrapolation kokkos

yury-lysogorskiy:
**Summary**

Implementation of extrapolation grade gamma (CPU implementation is here: https://github.com/lammps/lammps/pull/3315)  with KOKKOS

**Related Issue(s)**

**Author(s)**

Yury Lysogorskiy (ICAMS, Ruhr-University Bochum)

**Licensing**

**Backward Compatibility**

**Implementation Notes**

* new KOKKOS/pair_pace_extrapolation_kokkos.h/cpp  files, copy of  KOKKOS/pair_pace_kokkos.h/cpp
* modified kernel `TagPairPACEComputeRho` to mimic ACEBEvaluator ([here](https://github.com/ICAMS/lammps-user-pace/blob/main/ML-PACE/ace/ace_b_evaluator.cpp#L332) and  [here](https://github.com/ICAMS/lammps-user-pace/blob/main/ML-PACE/ace/ace_b_evaluator.cpp#L397)) and also to collect B-projections
* new kernel `TagPairPACEComputeGamma`

**Post Submission Checklist**

### DIFF
--- a/src/KOKKOS/pair_pace_extrapolation_kokkos.h
+++ b/src/KOKKOS/pair_pace_extrapolation_kokkos.h
@@ -24,9 +24,10 @@ PairStyle(pace/extrapolation/kk/host,PairPACEExtrapolationKokkos<LMPHostType>);
 #define LMP_PAIR_PACE_EXTRAPOLATION_KOKKOS_H
 
 #include "pair_pace_extrapolation.h"
-#include "ace-evaluator/ace_radial.h"
 #include "kokkos_type.h"
 #include "pair_kokkos.h"
+
+class SplineInterpolator;
 
 namespace LAMMPS_NS {
 
@@ -313,23 +314,7 @@ class PairPACEExtrapolationKokkos : public PairPACEExtrapolation {
 
     t_ace_3d4 lookupTable;
 
-    void operator=(const SplineInterpolator &spline) {
-      cutoff = spline.cutoff;
-      deltaSplineBins = spline.deltaSplineBins;
-      ntot = spline.ntot;
-      nlut = spline.nlut;
-      invrscalelookup = spline.invrscalelookup;
-      rscalelookup = spline.rscalelookup;
-      num_of_functions = spline.num_of_functions;
-
-      lookupTable = t_ace_3d4("lookupTable", ntot+1, num_of_functions);
-      auto h_lookupTable = Kokkos::create_mirror_view(lookupTable);
-      for (int i = 0; i < ntot+1; i++)
-        for (int j = 0; j < num_of_functions; j++)
-          for (int k = 0; k < 4; k++)
-            h_lookupTable(i, j, k) = spline.lookupTable(i, j, k);
-      Kokkos::deep_copy(lookupTable, h_lookupTable);
-    }
+    void operator=(const SplineInterpolator &spline);
 
     void deallocate() {
       lookupTable = t_ace_3d4();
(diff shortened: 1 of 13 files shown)


src/KOKKOS/pair_pace_extrapolation_kokkos.h
```diff
@@ -0,0 +1,354 @@
...
+PairStyle(pace/extrapolation/kk/host,PairPACEExtrapolationKokkos<LMPHostType>);
+// clang-format on
+#else
+
+// clang-format off
+#ifndef LMP_PAIR_PACE_EXTRAPOLATION_KOKKOS_H
+#define LMP_PAIR_PACE_EXTRAPOLATION_KOKKOS_H
+
+#include "pair_pace_extrapolation.h"
+#include "ace-evaluator/ace_radial.h"
```
akohlmey: This include should be avoided. Any header declaring a "style" should not include any implementation specific headers.

akohlmey: Upon closer inspection, this change **must** be done. The include file in question is in violation of the programming style requirement for LAMMPS. The standout issue is the use of "using namespace std;" in a header file. This effectively eliminates the benefits of namespaces and because all style files of the same kind are included at the same time in LAMMPS, will be implicitly applied to many other headers and classes and thus potentially cause namespace conflicts or unexpected access ambiguities beyond the style where it is applied. We've had very hard to debug problems of this kind before where adding a package would unexpectedly break some other packages.

In most of LAMMPS we go one step further and do not even use "using namespace std;" in the implementation to make it more obvious to people reading the code which are functions/classes of the C++ standard library and which are functions/members/instances of LAMMPS classes.

src/fix_pair.cpp
```diff
@@ -36,6 +36,10 @@ FixPair::FixPair(LAMMPS *lmp, int narg, char **arg) :
   if (nevery < 1) error->all(FLERR,"Illegal fix pair every value: {}", nevery);
 
   pairname = utils::strdup(arg[4]);
+  if(lmp->suffix) {
```
stanmoore1: In other places we use the `^` regex. This logic will not work for `suffix2`.

stanmoore1 (approved):

stanmoore1:
@yury-lysogorskiy nice job, thanks for working on this!

src/fix_pair.cpp
stanmoore1: @akohlmey what do you suggest for this?

stanmoore1: @akohlmey thanks, seems like it may be better to have a `force->pair_match_suffix` that pulls in all the suffix logic?

akohlmey: That would be a larger refactoring project. The challenge is to update all use cases and be mindful of exceptions.
The current API has grown over time. There is suffix handling and the look up of hybrid sub-styles. Both were added incrementally and I noticed that not all applications do both and do it consistently.

akohlmey (changes requested):
None of the "style" headers or any of the headers they include may use "using namespace std;"
If at all possible the implementation should be hidden from the interface and forward declarations or a pointer to a struct with the actual implementation stored (PIMPL). That way the headers for a style only need to include LAMMPS headers.

yury-lysogorskiy:
`fix pair` doesn't work anymore,  trying to fix it, please, do not merge

yury-lysogorskiy:
I fixed the issue with `fix pair` in [20b6355](https://github.com/lammps/lammps/pull/3585/commits/20b6355888d3b005601b62439270cd749dae24e9)

akohlmey (approved):

sjplimp (approved):

akohlmey:
@stanmoore1 this is still assigned to you. Are you still testing?

stanmoore1:
@akohlmey I approve the code changes and believe it is ready to be merged.

PR #3589, opened 2023-01-07

akohlmey:
**Summary**

The pull request combines multiple small changes and fixes

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- small cleanup in the mliappy code for Kokkos support to address reports from static analysis with CodeQL
- always build the C++-only "mesocnt" styles even if no Fortran is available and mesont/tpm cannot be built
- handle dependency of bond style mesont on bond style harmonic
- small refactoring of minimizer styles to use symbolic constants for integrator and linesearch styles, modernize output formatting
- silence bogus warning about atom IDs for image and movie style dumps
- avoid usage of sprintf(). replace with C++ string features or fmtlib
- address some minor issues flagged by static code analysis with coverity scan
- conditionally add `CONFIGURE_DEPENDS` keyword to CMake globbing. We do this by defining the variable `${CONFIGURE_DEPENDS}` to either by the keyword or empty for CMake prior to 3.12. This will make it easy to turn it always on when we increase the minimum CMake version required. With this change CMake now automatically recognizes when files are added to or removed from the a source folder and re-runs CMake to update the build rules.

**Post Submission Checklist**


hammondkd:
Line 54 of `doc/src/Fortran.rst` has a typo:
```
   A contributed Fortran interface interface is available in the
```
("interface" appears twice)

akohlmey:
> ("interface" appears twice)

Thanks for the note. Should be fixed now.

athomps:
It's confusing that MESONT package contains `mesocnt` styles. It would be better to make everything `mesont`, and then use CNT only as a buzzword in docs.

akohlmey:
> It's confusing that MESONT package contains `mesocnt` styles. It would be better to make everything `mesont`, and then use CNT only as a buzzword in docs.

What is even more confusing is that those are independent implementations of effectively the same model. One in pure C++, the other with some (explicit comment removed) Fortran library support functions. The one letter difference helps me to tell them and the related support code apart.

athomps:
> What is even more confusing is that those are independent implementations of effectively the same model. One in pure C++, the other with some (explicit comment removed) Fortran library support functions. The one letter difference helps me to tell them and the related support code apart.

I see.  Well, in that case, maybe `pair_mesocnt.cpp` should be called `pair_mesont_c.cpp`?

akohlmey:
> > What is even more confusing is that those are independent implementations of effectively the same model. One in pure C++, the other with some (explicit comment removed) Fortran library support functions. The one letter difference helps me to tell them and the related support code apart.
> 
> I see. Well, in that case, maybe `pair_mesocnt.cpp` should be called `pair_mesont_c.cpp`?

That is a question for @phankl

It would also be interesting to hear from @iafoss whether there is still active development on the Fortran based implementation. 

We have a bit of a REAX vs. USER-REAXC kind of situation here. According to the latest changes by @phankl the C++ implementation now supports all features of the Fortran based one.
So the best long-term choice depends on which parts get continued development and updates and which have features that are not available in the other.

From the mainenance (and user perspective) it would be preferable to have just one implementation and thus avoid confusions.

akohlmey:
@athomps the discussion about the style names and whether to deprecate the Fortran based variant is probably better moved to a "Feature Request" issue. The changes in this pull request only enable compiling the parts of the package that do not require a Fortran compiler to be compiled on machines that do not have one.

athomps:
> We have a bit of a REAX vs. USER-REAXC kind of situation here. According to the latest changes by @phankl the C++ implementation now supports all features of the Fortran based one. So the best long-term choice depends on which parts get continued development and updates and which have features that are not available in the other.

If this analogy is true, then it might be best to anticipate the eventual retirement of the Fortran version.  One way to do this would be to move the source files that depend on the Fortran library to a separate package called MESONT_F, give all the related commands a `_f` suffix, and then drop the `c` in the non-Fortran commands that remain in MESONT.

akohlmey:
> > We have a bit of a REAX vs. USER-REAXC kind of situation here. According to the latest changes by @phankl the C++ implementation now supports all features of the Fortran based one. So the best long-term choice depends on which parts get continued development and updates and which have features that are not available in the other.
> 
> If this analogy is true, then it might be best to anticipate the eventual retirement of the Fortran version. One way to do this would be to move the source files that depend on the Fortran library to a separate package called MESONT_F, give all the related commands a `_f` suffix, and then drop the `c` in the non-Fortran commands that remain in MESONT.

Creating a new package and/or (temporarily) renaming stuff is what I am trying to avoid. It would have made the included changes much simpler. But since these are not models with the same degree of popularity as ReaxFF, I think it would be OK to live with an inconsistency for a little bit. For a transition period, we would have to support alias names for the pair styles, too. And that is yet another effort.

sjplimp (approved):

iafoss:
@akohlmey,
Currently, I work at a different place. 
Professor Alexey Volkov, the main developer of the Fortran code (and the mesoscopic CNT model itself), as far as I know, is mainly using a standalone code in his group, but that version is actively developing. The Fortran version of the mesoscopic NT code added to LAMMPS is an adaptation of this Fortran code. It includes the advanced things related to the potential itself, but the latest features, such as CNT-polymer interaction for a description of crosslinking or an additional heatbath for accurate accounting for missing degrees of freedom in the mesoscopic simulation, were not included in this adaptation. The work on the adaptation of the Fortran code for LAMMPS was done in the group of Professor Leonid Zhigilei (previously Alexey Volkov worked in his group). However, at the moment, as far as I know, Leonid Zhigilei does not have any active projects involving CNT modeling. So that is why recently there was no active development of this package adopting Alexey's Fortran code to LAMMPS. 

Meanwhile, the C++ version of the package is an independent implementation of Alexey's paper(s). Unfortunately, there was a miscommunication when we worked on the package, and we were not aware of another (C++) LAMMPS CNT package developing exactly at the same time. As far as I remember, the initial C++ version of the package at the moment of including into LAMMPS in early 2020 was a very basic one, not including even such things as CNT buckling, for example.  However, if this package was under active development recently, it could include extra features (I'm not aware of the current situation, unfortunately). 

I you have any specific questions to Alexey Volkov or Leonid Zhigilei, I can contact them.

akohlmey:
@iafoxx Thanks for the quick response.

According to the comments for the rather recently submitted pull request to significantly update the C++ variant https://github.com/lammps/lammps/pull/3421 its implementation is now feature complete in comparison with the Fortran based one, it is said to be up to 2x faster, and it is more portable (since a C++ compiler is required to compile LAMMPS anyway, but not all platforms have a Fortran compiler when using their native compilers (e.g. macOS and Windows)).

Which is why the question whether to keep the Fortran based implementation included is a valid one. From the maintainer's view it is difficult to judge the feature equivalence or superiority of both variants, but replacing Fortran code with equivalent C++ code is certainly making things easier (see https://github.com/lammps/lammps/pull/3579 for example). It would also allow to enable the package by default in some more presets.

This would not be the first case of this kind. We had similar situations twice already (for the ReaxFF and MEAM implementations) and after having two variants included concurrently for some time, we ultimately removed the Fortran versions after confirming the equivalence or superiority of the C++ versions. There is a mechanism in place that will make people aware of the alternative if an old input is used.

phankl:
@akohlmey, thanks for raising this issue. Maybe I'll add my two cents, but keep in mind that I'm naturally biased towards by own implementation. As far as I'm concerned, my implementation is indeed functionally equivalent to the Fortran implementation (i.e. it includes the same segment-chain + segment-segment simulation modes + bending buckling as well as a convenient automatic parametrisation interface). As noted in #3421, the functionality now also exceeds the Fortran implementation in a few places as the mesocnt/viscous pair-style implements inter-tube friction. I've also implemented Gauss-Legendre quadrature compared to the trapezoid rule in the Fortran implementation for better numerical convergence. I've also attached the benchmarks I ran to confirm the improved performance in #3421. As a result of the improved performance (and some stability issues), I know of two research groups I am in contact with that have moved away from the Fortran version and have started using the C++ implementation. 

As @iafoss mentioned, Alexey has added lots of functionality to his group's version of the Fortran library, but I don't know whether he intends to port it to the LAMMPS version. 

As for the original question, I don't have any preference in terms of names for the package styles. I think the differences between the implementations are documented clearly enough, and realistically, there probably aren't too many people using this package as of now. While I don't think this package will reach ReaxFF levels of popularity, the number of users may increase in the future, so it might be worth resolving this issue sooner rather than later.

akohlmey:
> As @iafoss mentioned, Alexey has added lots of functionality to his group's version of the Fortran library, but I don't know whether he intends to port it to the LAMMPS version.

@iafoss This leads to two key questions. Is there a) an intention to maintain and update the Fortran based code in LAMMPS. If I remember correctly, this was not an easy step, since the original code was doing a few things that were not very portable across platforms and it is still not a very "clean" implementation, but there are still parts of LAMMPS that are worse in terms of programming style. The second question would be if there is an objection to drop the Fortran code in favor of the C++ implementation.

As a software maintainer, I cannot judge the science, but I am looking at this from the software engineering point. Having only the pure C++ code has three advantages: 1) it does not have any additional requirements (like a Fortran compiler) to be compiled on any platform that LAMMPS can be compiled on 2) it is by construction compatible with using multiple LAMMPS instances concurrently through the C/Python library interface which is becoming a growing use case of LAMMPS with high-throughput workflow frameworks growing in popularity. Finally, 3) having a Fortran library coupled to C++ is always more complex, so dropping it would make maintenance of the package *significantly* simpler.

Since we all are aware that this model and package is *not* one of the frequently used ones in LAMMPS, I am rather relaxed about how to go about it. But then again, since we started the discussion, I would rather bring it to a conclusion now instead of postponing a decision on what to keep and how until later.

iafoss:
@akohlmey I discussed this question with Alexey Volkov and got the following. At the current moment, he doesn't have sufficient capacity to incorporate new features into the package and fully rewrite the remaining Fortran parts as C++ (we agree with your arguments regarding getting rid of Fortran). If you have to make a decision right now, we would agree with your suggestion.

However, if you can wait for a year or so, Alexey may get extra funding and workforce to continue developing the package (get rid of Fortran and add more functionality to the updated force filed).

akohlmey:
@iafoss Thanks for the update. I apologize to take so long to respond, but I had to think a bit about what is the best way to proceed. Initially, I was leaning toward just leaving everything *as is* and wait.  But the point about rewriting the Fortran library in C++ kept stirring in my mind and with that in mind, I think it would be just as well to remove the Fortran version now.

Assuming there would be additional funding for Alexey, it would not make much sense to me to replicate the efforts of @phankl and create a *second* C++ version of the same functionality and then to add to it.  Rather it seems to me much more effective to **collaborate** with @phankl and improve and extend his version. And this is not just from the perspective of development and using time efficiently, I believe that this would also be in the interest of the users so they would not have to make a choice between the two competing implementations where the differences will be hard to tell for most of them.

phankl:
It took me more than 6 months of total development time to get the C++ implementation to work as well as it does, so I agree with @akohlmey that it does not make sense to replicate my work. I actually originally suggested to Alexey a while ago to collaborate on a C++ implementation, but nothing came of it. I'd be more than happy to carry this forward and expand the functionality of the package in the future, including implementing some features that Alexey and colleagues have come up with in the meantime.

iafoss:
It is quite unfortunate, but this option may be reasonable at the moment. I'll contact Alexey Volkov and ask him for the final confirmation before accepting the removal.

As I remember, @phankl instead of contacting Alexey in 2019 and letting him know that he is developing the package just asked several questions related to the potential and silently worked on the package. Meanwhile, in one of our projects, it was explicitly requested to add the mesoscopic CNT simulation capabilities to LAMMPS. If @phankl let Alexey know that he is doing this work and is planning to support the package in the future, we would not spend efforts adopting Fortran code to LAMMPS. So we ended up with two versions of the package( 
I see @akohlmey's point that switching to C++ implementation might be reasonable, and if @phankl is planning to support his package in the future, it may be suboptimal to put more effort into rewriting the Fortran code.

phankl:
@iafoss, whilst I do not think this is the place for such discussions, I believe your claim that I did not reach out to Alexey and worked 'silently' on my implementation is not true and unfair. I emailed him in July 2019, stating that I had started working on a LAMMPS implementation. I had made a lot of progress on the code, and Alexey suggested we collaborate. I agreed, further suggesting to use my codebase for any future work as the base functionality was nearly complete. He replied he was busy working on other things in mid-August 2019 but would get back to CNT work in September. I didn't hear anything from him, but still later shared my near-finished first version of the code in early December. Alexey indeed helped by answering a few of my questions, but then stated that you were working on your own LAMMPS version a few weeks later. I suggested we write a paper together in January 2020, which he declined, stating he would get back to me soon after, which he never did. 

Maybe there was some kind of miscommunication, but my intention to collaborate was clear from the start and I openly shared my code with Alexey well before the Fortran version was published. Regardless, I think we should let this rest. 

I'm happy to support the C++ package in the future, and if @iafoss or Alexey would like to implement new features I would gladly collaborate on that.

iafoss:
I do not know the details of your and Alexey's communications, but the thing that was told me by Alexey when in October 2019 I was asked to do work on this project (which is completely unrelated to my previous work) is "We have a project requirement to have mesoscopic CNT implemented in LAMMPS. One student contacted me with several questions regarding the potential and mentioned that he may be interested in implementing it in LAMMPS **but then just disappeared**". Then Alexey got an email from @phankl about at the middle of December that he implemented the basic version of the potential. At that point, I already worked on this project for several months, and since it was unclear if this basic alternative C++ version will be improved in the future, we decided to proceed with adopting the Fortran code, which as I remember had many more features at that moment. 

It is quite unfortunate that @phankl's and Alexey's miscommunication caused these troubles, but it is good that @phankl continued working on his package and implemented things included into the Fortran package. Since @phankl is planning to support his package in the future, it may be better for LAMMPS and for users to have a pure C++ version.

akohlmey:
Ok. This discussion has gone far enough. It is quite common that there are parallel developments and it is not automatically a bad thing, as frustrating as it may be for the involved parties at some time or another.

Sometimes having a reference to compare to can make *both* implementations better. Also, sometimes developers have different priorities. Please see for example, the interfaces that LAMMPS has to Colvars and Plumed2, which are both collective variable toolkists.  Finally, sometimes it just happens that an implementation has technical advantages over the other. That does not automatically diminish the scientific value or appreciation for the other. If you look at PR #3605 you should see that it keeps the credits and mentions that in older releases, two versions were available (so it is always possible for people to go back, if needed).

As a person focusing on maintaining the overall package my primary interests are: 1) what is better for the users and 2) what makes it easier to compile/install and maintain. This should be addressed in PR #3605. Any technical suggestions or suggestions for improving the documentation should go there. I am closing *this* discussion now.

PR #3591, opened 2023-01-09
Implementation of ABC-FIRE

sebastianecheverrir:
**Summary**

Implementation of an improved version of FIRE, called ABC-FIRE.  Details of the method can be found in the article "ABC-FIRE: Accelerated Bias-Corrected Fast Inertial Relaxation Engine" doi.org/10.1016/j.commatsci.2022.111978

**Related Issue(s)**

This does not address any open issue

**Author(s)**

Sebastian Echeverri Restrepo
SKF and King's College London
sebastianecheverrir@gmail.com

Predrag Andric
SKF

**Licensing**

**Backward Compatibility**

Changes will not break backward compatibility

**Implementation Notes**

Two new files were added called min_abcfire.cpp and min_abcfire.h. This make the implementation completely independent of the current fire implementation. 
Correctness was verified with various systems and potential types, as described in the publication doi.org/10.1016/j.commatsci.2022.111978.

**Post Submission Checklist**

doi.org/10.1016/j.commatsci.2022.111978

### DIFF
--- a/doc/src/min_modify.rst
+++ b/doc/src/min_modify.rst
@@ -28,9 +28,9 @@ Syntax
        *discrete_factor* value = factor
          factor = discretization factor for adaptive spin timestep (adim)
        *integrator* value = *eulerimplicit* or *verlet*
-         time integration scheme for fire minimization
+         time integration scheme for fire or abcfire minimization
        *tmax* value = factor
-         factor = maximum adaptive timestep for fire minimization (adim)
+         factor = maximum adaptive timestep for fire or abcfire minimization (adim)
 
 Examples
 """"""""
@@ -53,7 +53,7 @@ The *cg* and *sd* minimization styles have an outer iteration and an
 inner iteration which is steps along a one-dimensional line search in
 a particular search direction.  The *dmax* parameter is how far any
 atom can move in a single line search in any dimension (x, y, or z).
-For the *quickmin* and *fire* minimization styles, the *dmax* setting
+For the *quickmin*, *fire* and *abcfire* minimization styles, the *dmax* setting
 is how far any atom can move in a single iteration (timestep).  Thus a
 value of 0.1 in real :doc:`units <units>` means no atom will move
 further than 0.1 Angstroms in a single outer iteration.  This prevents
@@ -82,7 +82,7 @@ even if atoms could move in the gradient direction to reduce forces
 further.
 
 The choice of a norm can be modified for the min styles *cg*, *sd*\
-, *quickmin*, *fire*, *fire/old*, *spin*, *spin/cg* and
+, *quickmin*, *fire*, *fire/old*, *abcfire*, *spin*, *spin/cg* and
 *spin/lbfgs* using the *norm* keyword.  The default *two* norm computes
 the 2-norm (Euclidean length) of the global force vector:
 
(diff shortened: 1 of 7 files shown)


akohlmey:
@sebastianecheverrir thank you for your submission.
I have a couple of questions:
- can you elaborate a bit more for which kind of scenario this minimizer variant is an improvement? When I replace "fire" with "abcfire" in the inputs in the examples/fire folder, it uses more iterations than "fire" and "fire" has the lower energy final geometry. This should be reflected in the documentation (more than just a link to a publication).
- when I compare the code to the fire implementation. there is a large amount of shared code. I wonder if it would not be better for maintaining this by having "abcfire" variant as a derived class that just sets a flag and then branch out in the base "fire" variant as needed. See airebo and airebo/morse for an example.

akohlmey:
@sebastianecheverrir please see below for how your code could be folded into the regular fire code. Due to using a templated function the if statements come at no cost as the compiler will optimize them away.

https://github.com/lammps/lammps/compare/develop...akohlmey:lammps:abc-fire-alternative

It is also possible to drop  `min_style abcfire` altogether and instead have something like `min_modify abcfire yes/no`.

sebastianecheverrir:
@akohlmey Thanks for the fast review and reply 

abcfire should be in general more efficient than fire. I have re-run the examples in the folder _examples/fire_ and I get the following:

**Input file: in.fire**
With fire 2360 iterations are needed to converge. It reaches an energy (TotEng) of  -2.895392
With abcfire 2841 iterations are needed for convergence, but the energy  (TotEng)  is lower -2.9203989. abcfire reaches the energy -2.895392 after about 1600 iterations

**Input file: in.neb.sivac.fire (run with -partition 16x1)**
In this example the default value of  _timestep_ was changed to 0.01, which  improves the performance of fire for this specific simulation. If we use the default value for _timestep_ , abcfire is faster than fire (84 vs121 iterations)
As with this example, there can be cases where, if fire is fine tuned, it will be faster than abc fire. Nevertheless, our tests indicate that, in general, abcfire should be faster than fire.

Thanks for the example for folding the code. I will make some modifications to the files.

akohlmey:
@sebastianecheverrir please note that a blanket statement like `should be in general more efficient` is not very useful advice, when the results are so varied and specifically when you don't provide a good description for *when* using the ABC variant has an advantage and why. Users need some guidelines in the documentation and ideally some suggestions about which settings to tweak how for the best outcome.

After sleeping over it and having some more time to consider the impact of the changes, I am now even more convinced that this is too small a change to justify a new style. It is not really a new algorithm but more like a variant. It should just be implemented via `min_modify` since it is quite similar to other options offered there. Adding this change to the code in my suggested branch would be rather straightforward.

jguenole:
@akohlmey I also believe that an implementation via `min_modify` would be the best. It will then be straightforward to use this additional tweak to the `fire` algorithm. 

Regarding the addition itself, it is IMHO an additional tweak to the `fire` algo. As it is already full of tweak and this one remains simple and yet show in some cases impressive improvement, I find very interesting to implement it directly. @sebastianecheverrir thanks for your contribution to the algo!

Assuming that it perform "better" (iteration and/or forces) than the original `fire` _in most cases_, it is indeed difficult to say in which condition it could perform worst that the original `fire`: I will try to run some of the test cases from the original fire paper in the next days to see if I can help on that matter.

akohlmey:
@sebastianecheverrir @jguenole I have updated my "alternative" branch to now fully implement the min_modify version and also adapt the documentation. Please test and check carefully that I have not introduced any errors. I have also updated a couple of the examples. I am content with this version, but I am open to suggestions. One possible change is to take back the template function "optimization". It is fairly easy to do and undo. Not sure how much of a difference this makes here. I will now sit back and let you guys figure out the final solution and which pull request to merge. Just ping me if you need me to check something or give my opinion.

sebastianecheverrir:
@akohlmey   FIRE has a latency time whenever the velocities are reset to zero. ABC-FIRE aims to reduce that latency time. So, _in theory_,  ABC-FIRE should _always_ converge faster. In practice, it is harder to tell. The tests that we have done show that it is always faster, but there could be specific cases where it does not behave as expected.  Let's see what the additional tests that @jguenole will do give.

@akohlmey  Thanks for taking the time to implement the min-modify version. I tested it with some of the cases from the ABC-FIRE  article. I get exactly the same results. My suggestion would be to merge your version. 

@jguenole Thanks for offering to make some additional tests, I look forward to seeing the results.

akohlmey:
> Thanks for taking the time to implement the min-modify version. I tested it with some of the cases from the ABC-FIRE article. I get exactly the same results. My suggestion would be to merge your version.

Thanks for checking this out. I have flagged this PR as draft and it will be automatically closed when the other gets merged. Please note the documentation is still minimal. If you have a small example where ABC-FIRE is particularly efficient, please attach to #3594 so I can add it to examples/fire. I have so far duplicated the existing examples and added the min_modify line.

PR #3596, opened 2023-01-12
NEB routine to target equal energy difference between knots

tomswinburne:
**Summary**

NEB 'ideal' aims to give equidistant knots; this PR implements a NEB style 'equal' which aims to give knots equidistant in energy. From the proposed fix_neb.rst:

With a value of ideal, the spring force is computed as suggested in ref`(WeinanE) <WeinanE>`

Fnudge_parallel = -Kspring * (RD-RDideal) / (2 * meanDist)

where RD is the "reaction coordinate" see [:doc:`neb <neb>`](https://github.com/tomswinburne/lammps/blob/energy_spacing_neb/doc/src/fix_neb.rst#id15) section, and RDideal is the ideal RD for which all the images are equally spaced. I.e. RDideal = (I-1)*meanDist when the climbing replica is off, where I is the replica number). The meanDist is the average distance between replicas. Note that in this case the specified Kspring is in force units.

Note that the ideal form of nudging can often be more effective at keeping the replicas equally spaced.

With a value of equal the spring force is computed as for ideal before the climbing stage, then is computed to promote equidistant spacing in energy rather than distance:

Fnudge_parallel = -Kspring * (ED-EDideal) / (2 * meanEDist)

where ED is the cumulative sum of absolute energy differences ED=sum(i<I)|E(Ri+1)-E(Ri)|, EDideal = (I-1)*meanEdist and meanEdist is the average absolute energy difference between replicas. This form of nudging is to aid schemes which integrate forces along NEB pathways such as [:doc:`fix_pafi <fix_pafi>`], by providing optimal quadrature points.

**Related Issue(s)**

**Author(s)**
 Tom D Swinburne, CNRS / Aix-Marseille University, thomas.swinburne at cnrs.fr

**Licensing**

**Backward Compatibility**
no

**Implementation Notes**
Fully checked in lammps, minimal changes to code

**Post Submission Checklist**

- [ ] Licensing information is [complete]
- [X] The added/updated documentation is integrated and tested with the documentation build system
- [X] A package specific README file has been included or updated

Test image:
neb_test_c15


doc/src/fix_neb.rst
```diff
@@ -124,6 +125,20 @@ in force units.
 Note that the *ideal* form of nudging can often be more effective at
 keeping the replicas equally spaced.
 
+With a value of *equal* the spring force is computed as for *ideal*
+before the climbing stage, then is computed to promote equidistant 
+spacing in energy rather than distance:
+
+.. parsed-literal::
```
akohlmey: Mathematical expressions are better typeset with a `.. math::` block or a `:math:` role. We should use the opportunity of this pull request to revise the entire NEB documentation accordingly. I am happy to take a stab at it unless you prefer to do it yourself.

src/REPLICA/fix_neb.cpp
```diff
@@ -62,22 +62,29 @@ FixNEB::FixNEB(LAMMPS *lmp, int narg, char **arg) :
...
   kspringIni = 1.0;
   kspringFinal = 1.0;
-
   int iarg = 4;
   while (iarg < narg) {
     if (strcmp(arg[iarg],"parallel") == 0) {
       if (iarg+2 > narg) error->all(FLERR,"Illegal fix neb command");
       if (strcmp(arg[iarg+1],"ideal") == 0) {
         NEBLongRange = true;
+        EqualForceNEB = false;
```
akohlmey: Rather than having a growing list of booleans with constraints of which may be true or false, we should convert this to have just one integer variable that selects the "NEB mode" and then an enumerator with symbolic constants to choose one of the available ones. This would make it more straightforward to add a fourth or fifth, if that need ever arises.

src/REPLICA/fix_neb.cpp
```diff
@@ -135,7 +142,8 @@ FixNEB::FixNEB(LAMMPS *lmp, int narg, char **arg) :
   else procnext = -1;
 
   uworld = universe->uworld;
-  if (NEBLongRange) {
+
+  if (NEBLongRange or EqualForceNEB) {
```
akohlmey: Using "or" as logical operator is not portable (it fails to compile with Microsoft Visual C++ by default). You should be using "||". The latter form is the most common, used throughout LAMMPS, and a familiar choice in C/C++ anyway.

src/REPLICA/fix_neb.cpp
```diff
@@ -849,6 +818,66 @@ void FixNEB::inter_replica_comm()
   }
 }
 
+/*
+Calculate ideal positions for parallel "ideal" or "equal"
+*/
+void FixNEB::calculate_ideal_positions()
+{
+  // Skip unless "ideal" or "equal"
+  if (not (EqualForceNEB or NEBLongRange)) return;
```
akohlmey: Same as above, also "not" is not portable and not following the established convention in LAMMPS.

src/REPLICA/neb.cpp
```diff
@@ -166,8 +168,12 @@ void NEB::command(int narg, char **arg)
     } else if (strcmp(arg[iarg],"verbose") == 0) {
       verbose=true;
       ++iarg;
+    } else if (strcmp(arg[iarg],"terse") == 0) {
```
akohlmey: This is an undocumented addition. No feature gets added to LAMMPS without documentation.
Furthermore, the same situation as above applies here. This turns a binary choice into a ternary, so we should have an enumerator with symbolic constants (VERBOSE, NORMAL, TERSE).

But before even going there, please give the rationale for the "terse" option.

akohlmey (changes requested):
As indicated in individual comments, this needs changes to the code and documentation and some explanation. It would also be an opportunity to do some refactoring beyond your changes, especially for the documentation to make it better aligned with the latest style and conventions that we are trying to establish for better readability and visual appeal.

doc/src/fix_neb.rst
tomswinburne: I would be very happy to take up the second iteration as I am confident you will be much better aligned in style with the rest of the doc!

src/REPLICA/fix_neb.cpp
tomswinburne: Agreed, I avoided this to change the existing code as little as possible. I can implement this

src/REPLICA/fix_neb.cpp
tomswinburne: Thank you for this, I was not aware, it will be changed

src/REPLICA/fix_neb.cpp
tomswinburne: Noted. Thank you!

src/REPLICA/neb.cpp
tomswinburne: The "terse" option was to give a basic readout which fits into one terminal line- the current image-by-image readout spans multiple lines and is not ideal visually. 

But happy to remove this for this PR.

doc/src/fix_neb.rst
akohlmey: Ok. Please ping me, when you are done making changes, so I would step on your feet when I grab the branch and make modifications.

src/REPLICA/fix_neb.cpp
akohlmey: I suggest to have the constants reflect the name of the keywords. Something like:
```
enum {NEIGHBOR, IDEAL, EQUAL};
```

src/REPLICA/neb.cpp
akohlmey: I am not opposed to this kind of change. Having only summary output and skipping per-replica output is a sufficient justification for a third output verbosity option. I just wanted to know that and also it needs to be documented.

tomswinburne:
I have implemented the changes @akohlmey , I added a similar `enum` for the verbose keyword also.

tomswinburne:
To confirm @akohlmey , I will not edit anymore until future notice

akohlmey (approved):

akohlmey:
> To confirm @akohlmey , I will not edit anymore until future notice

I am done with my changes

tomswinburne:
Many thanks for the quick changes @akohlmey

athomps (changes requested):
1. I suggest replacing `terse` and `verbose` keywords with a single optional keyword `verbosity` that is followed by values `normal` or `terse` or `verbose`.  This will fully reflect the values of `print_mode` in the implementation.  The doc page should be updated accordingly, including in Default section at the bottom  "{verbosity} = {normal}" 
2. The plot included in the PR shows that `EDideal` is different on the left and right sides of the climbing image.  This should be reflected in the doc page definition of Fparallel, at least in the form of a comment, but ideally in the form of a mathematical expression.

akohlmey:
> 1. I suggest replacing `terse` and `verbose` keywords with a single optional keyword `verbosity` that is followed by values `normal` or `terse` or `verbose`.  This will fully reflect the values of `print_mode` in the implementation.  The doc page should be updated accordingly, including in Default section at the bottom  "{verbosity} = {normal}"

This is done now.

tomswinburne:
>     2. The plot included in the PR shows that `EDideal` is different on the left and right sides of the climbing image.  This should be reflected in the doc page definition of Fparallel, at least in the form of a comment, but ideally in the form of a mathematical expression.

Updated doc page to reflect the operation of *equal* and *ideal*. During climbing, the *ideal* style no longer targets equal spacing, but equal spacing each side of the climbing image. The *equal* style is identical to *ideal* before climbing, then during climbing the target is equal absolute energy differences each side of the climbing image.

athomps (approved):
I approve.

athomps:
> Updated doc page to reflect the operation of _equal_ and _ideal_. During climbing, the _ideal_ style no longer targets equal spacing, but equal spacing each side of the climbing image. The _equal_ style is identical to _ideal_ before climbing, then during climbing the target is equal absolute energy differences each side of the climbing image.
@tomswinburne Thanks for also clarifying the effect of _ideal_ style in climbing mode.

tomswinburne:
Many thanks @akohlmey and @athomps !

PR #3597, opened 2023-01-13
Add ITEM: DIMENSION to dump grid output

sjplimp:
**Summary**

Add these lines to the dump grid output file for each snapshot:

ITEM: DIMENSION
2 or 3

This is so that post-processing tools like OVITO can determine how to visualize the grid. 

**Related Issue(s)**

N/A

**Author(s)**

Steve

**Licensing**

**Backward Compatibility**

Change the format of the dump grid snapshots, but should not be a problem if a post-processing tool simply skips
ITEMs it does not recognize.

**Implementation Notes**

**Post Submission Checklist**


doc/src/dump.rst
```diff
@@ -363,22 +363,25 @@ attributes is given below.
 
 .. versionadded:: 22Dec2022
 
-For style *grid* the extent of the Nx by Ny by Nz grid that overlays
-the simulation domain is output with each snapshot:
+For style *grid* the dimension and extent of the Nx by Ny by Nz grid
```
akohlmey: I think the terms "dimension" and "extent" need additional qualifiers here or alternate terms to avoid confusion.
People often use the term "grid dimensions" to describe what you have called "extent" here and renamed to "size" in the dump output. After all "ITEM: DIMENSION" does not seem to refer to the grid but to the dimensionality of the simulation cell.

There is some mention of details below, but this may be a bit too late.

sjplimp:
@akohlmey please see if this version is more clear

akohlmey:
> @akohlmey please see if this version is more clear

Yes. Small change, big impact.

akohlmey (approved):

PR #3603, opened 2023-01-20
reaxff/species delete rate limit

jrgissing:
**Summary**

add 'delete_rate_limit' keyword to set upper limit on rate of molecule deletion

**Related Issue(s)**

none

**Author(s)**

Jake Gissinger (NASA)

**Licensing**

**Backward Compatibility**

**Implementation Notes**

unlike the related 'fix bond/react' keyword, smooth restart support has not been implemented

**Post Submission Checklist**


src/REAXFF/fix_reaxff_species.cpp
```diff
@@ -50,6 +64,8 @@ FixReaxFFSpecies::FixReaxFFSpecies(LAMMPS *lmp, int narg, char **arg) :
     x0(nullptr), BOCut(nullptr), fp(nullptr), pos(nullptr), fdel(nullptr), ele(nullptr),
     eletype(nullptr), filepos(nullptr), filedel(nullptr)
 {
+  if (lmp->citeme) lmp->citeme->add(cite_reaxff_species_delete);
```
akohlmey: Please see: https://docs.lammps.org/stable/Modify_style.html#citation-reminder-suggested

akohlmey (changes requested):
I didn't notice before, but since we have a KOKKOS version of `fix reaxff/species`, this feature and support for the "delete" keyword, which it depends on and that was added previously, need to be ported to the KOKKOS version.

jrgissing:
> I didn't notice before, but since we have a KOKKOS version of `fix reaxff/species`, this feature and support for the "delete" keyword, which it depends on and that was added previously, need to be ported to the KOKKOS version.

last time, we decided that the 'delete' keyword does not affect KOKKOS, because only one major routine (FindMolecule) is overridden by KOKKOS, and I do not modify that one.

akohlmey:
> last time, we decided that the 'delete' keyword does not affect KOKKOS, because only one major routine (FindMolecule) is overridden by KOKKOS, and I do not modify that one.

I apologize. This is my bad. Didn't notice and didn't recall.

Too many pull requests and too much hacking on different things for an old and abused brain like mine, I suppose. 

Jumping between rather different tasks like CMake scripting, grammar and overview checks of the manual, portability checks (and bootstrapping all kinds of virtual machines and cross-compilers for that), and testing GPU kernels on different OpenCL hardware seems to be taking its toll. :-(

akohlmey (approved):

sjplimp (approved):

PR #3608, opened 2023-01-24
Multi-state MEAM with Kokkos support

rohskopf:
**Summary**

Add multi-state MEAM (MS-MEAM) support in the MEAM pair style.

This is similar to normal MEAM but we have 6 extra parameters in the MEAM library file, along with a few extra arrays in the potential calculation.

Kokkos implementation is included.

**Related Issue(s)**

None

**Author(s)**

Drew Rohskopf (SNL)
Aidan Thompson (SNL)
Mike Baskes (LANL, UNT)

**Licensing**

**Backward Compatibility**

Backward compatible.

**Implementation Notes**

Added optional `ms` flag when declaring the MEAM pair style, which activates MS-MEAM. This allows us to read the extra MS-MEAM library file parameters in the pair style. The flag is also used in the MEAM class in other files, similar to other flags like `ialloy`. Unlike `ialloy`, however, we need to use `msmeamflag` in the pair style for reverse/forward communication and library file reading.

**Post Submission Checklist**

MS-MEAM paper: https://journals.aps.org/prb/abstract/10.1103/PhysRevB.75.094113


rohskopf:
Running tests on my Linux machine:
```
[... build-testing]$ ctest -R AtomicPairStyle:meam
Test project /home/adrohsk/lammps/build-testing
    Start 244: AtomicPairStyle:meam
1/4 Test #244: AtomicPairStyle:meam .............   Passed    5.46 sec
    Start 245: AtomicPairStyle:meam_sc
2/4 Test #245: AtomicPairStyle:meam_sc ..........   Passed    1.65 sec
    Start 246: AtomicPairStyle:meam_spline
3/4 Test #246: AtomicPairStyle:meam_spline ......   Passed    1.42 sec
    Start 247: AtomicPairStyle:meam_sw_spline
4/4 Test #247: AtomicPairStyle:meam_sw_spline ...   Passed    1.47 sec

100% tests passed, 0 tests failed out of 4

Label Time Summary:
slow        =  10.00 sec*proc (4 tests)
unstable    =   1.47 sec*proc (1 test)

Total Test time (real) =  10.02 sec
```

Looking into why it's failing on Windows.

akohlmey:
> Looking into why it's failing on Windows.

Here is a suggestion: make sure you have valgrind installed on your Linux box and then run: `ctest -T memcheck -R AtomicPairStyle:meam$ -V`

doc/src/pair_meam.rst
```diff
@@ -11,7 +11,9 @@ Syntax
 
 .. code-block:: LAMMPS
 
-   pair_style meam
+   pair_style meam [ms]
+
+* ms = use multi-state MEAM (optional)
```
akohlmey: Using "ms" as a keyword is not a very user friendly choice. This should be more descriptive like "multistate".
An alternative (and my personal preference in this case) would be to have a "wrapper" pair style "msmeam". Where the derived class has no added functionality except for initializing the "msmeam" flag to 1
This has the advantage of the pair style name matching the published name of the method and retaining the property that MEAM pair style commands do not have arguments.

examples/meam/msmeam/README.md
```diff
@@ -0,0 +1,9 @@
+To run Baske's test, do
```
akohlmey, line 1: We don't want the examples folder "littered" with test inputs like this. There are some exceptions when the author of the package has written elaborate test/validation scripts. But in general, the better place for running consistency tests would be the unittest tree. See for example the unittest/force-styles/test_pair_list.cpp file.

examples/meam/msmeam/library.MSmeam
```diff
@@ -0,0 +1,14 @@
+# DATE: 2018-09-22  MB DATE: 2016-09-21 CONTRIBUTOR: Steve Valone, smv@lanl.gov CITATION: Baskes, PRB 1992; smv, sr, mib, JNM 2010
```
akohlmey: There should be a UNITS: tag here and the file should be moved to `potentials/library.msmeam` (note the case).

examples/meam/msmeam/HGaMS.meam
```diff
@@ -0,0 +1,30 @@
+bkgd_dyn        =       1
```
akohlmey: this file should be called `HGa.msmeam` and moved to the `potentials` folder.

src/MEAM/pair_meam.cpp
```diff
@@ -429,13 +449,23 @@ void PairMEAM::read_global_meam_file(const std::string &globalfile)
         b1[index] = values.next_double();
         b2[index] = values.next_double();
         b3[index] = values.next_double();
+        if (msmeamflag){
```
akohlmey: In the LAMMPS programming style conventions there is always a blank before opening braces.

akohlmey (changes requested):
There are a few formal/cosmetic changes that should be done.
On the issue with the "ms" keyword, I would wait until there is feedback from other LAMMPS developers before making changes.

akohlmey:
@rohskopf found another uninitialized data access on Linux, but that does not seem to address the problem on Windows. It looks like I need to compile LAMMPS manually on a Windows machine with Visual Studio to have a closer look.
I have a suitable virtual machine set up on my desktop in the office and will try to take a look while we are getting ready for making a new patch release today.

akohlmey:
@rohskopf Just ran the tester on Windows and it prints the following message before it crashes:

```
Run-Time Check Failure #2 - Stack around the variable 'drho3mdrm2' was corrupted.

```

akohlmey:
Based on the message from Windows, I compiled on Linux with bounds checking enabled and found the block that I changed in commit 16354d02622390c6c758ff152d4f7a51843607c0 was accessing the 3-element arrays at index 3, since this block is not inside a loop over m. I added the loop.

rohskopf:
Thank you @akohlmey. Made the corresponding fix in Kokkos.

doc/src/pair_meam.rst
athomps: @akohlmey Good point.  Perhaps we should follow the example of the PairEAM, which has many `eam/*` styles implemented as child classes.  The new command should be either `pair ms/meam` or `pair meam/ms`, I prefer the first one.

akohlmey: Hmmm... if we want to be consistent with EAM, it should be meam/ms. I like msmeam a little better than ms/meam. With style names that habr suffixes, it usually the suffix and not the prefix that indicates the modification; msmeam avoids that.
@sjplimp would you mind being the tie-breaker here? :wink:

sjplimp: I think meam/ms is consistent with other pair styles variants.  Also easier for users to understand and find
since pair style meam and meam/ms will be adjacent in the alphabetic listing.

akohlmey: Ok. I'll implement the change.

examples/meam/msmeam/README.md
akohlmey: Unit test file is added

akohlmey (approved):

doc/src/pair_meam.rst
akohlmey: done

akohlmey:
@rohskopf @athomps I think this may be ready to go in now. Please check and confirm.

src/KOKKOS/meam_dens_init_kokkos.h
```diff
@@ -43,11 +43,23 @@ void MEAMKokkos<DeviceType>::operator()(TagMEAMZero, const int &i) const {
   d_rho0[i] = 0.0;
   d_arho2b[i] = 0.0;
   d_arho1(i,0) = d_arho1(i,1) = d_arho1(i,2) = 0.0;
-  for (int j = 0; j < 6; j++)
+  if (this->msmeamflag) {
```
stanmoore1: This will crash on GPUs (need to get rid of `this->` since it is a host pointer)

stanmoore1 (changes requested):
@rohskopf thanks for porting this to Kokkos, nice job! Just had one change request.

stanmoore1:
Also, can we please rename the MSMEAM example to `in.test` instead of `test.in`? This is standard for LAMMPS and needed for regression testing.

athomps (approved):
I approve.

stanmoore1:
I'm testing this on a GPU now

stanmoore1:
GPU results match CPU for the example.

stanmoore1 (approved):

src/KOKKOS/meam_dens_init_kokkos.h
rohskopf: Thanks @stanmoore1 .
Does this also apply to the Kokkos pair cpp files?
E.g. found another one here: https://github.com/rohskopf/lammps/blob/2f8ba9dd324aecdba9cd2426203bf777642b78ec/src/KOKKOS/pair_meam_ms_kokkos.cpp#L23

rohskopf: Oops never mind, I see why `this->` is necessary in the wrapper.

rohskopf:
Thanks for all the help @akohlmey and @stanmoore1 !
Confirmed it's ready from my end.

src/KOKKOS/meam_dens_init_kokkos.h
stanmoore1: Yes just can't use `this->` inside a Kokkos parallel region, e.g. `parallel_for`

PR #3612, opened 2023-01-25
Fix PIMD: bugfix in spring energy, and adding virial estimator

yotamfe:
**Summary**
Hi,

Two proposed changes to the quantities that Fix PIMD exposes for each step:

1. Fix a bug in the calculation of the spring energy, using the position difference rather than the position, and including the spring constant factor.
2. Provide also the current value of the virial, which is useful as an estimator for the kinetic energy of the original quantum system. It is necessary to add this as an additional quantity because the estimator cannot otherwise be recovered from the already available *total* force.

**Related Issue(s)**

None.

**Author(s)**

Barak Hirshberg
Sagi Meir

**Licensing**

**Backward Compatibility**

No issues expected.

**Implementation Notes**

1. We tested the fix for the spring energy by comparing the reported values to values we computed post-processing from the positions, in a simulation of several particles in a harmonic trap.
2. We tested the virial by estimating the energy of several particles in a harmonic trap and comparing the resulting total energy to the analytical solution for non-interacting quantum harmonic oscillators.

**Post Submission Checklist**


athomps (approved):
I approve.

akohlmey:
@yotamfe thanks for your submission. It looks like there should be some documentation added for the additional property that can be read from the fix (and specifically it should state that it is a scalar and not the more typical 3x3 tensor). 

This would normally go to a section titled:
`Restart, fix_modify, output, run start/stop, minimize info`

Which seems to be missing altogether. So we should take this opportunity to add one and fill in the respective information. You can compare to other fixes what is added there.

Thanks in advance.

yotamfe:
> 

@akohlmey Thanks for reviewing this! I'll look at the other fixes and add the relevant documentation.

akohlmey:
> > 
> 
> @akohlmey Thanks for reviewing this! I'll look at the other fixes and add the relevant documentation.

Great. When you are done and have updated your branch, please click on the "Ready for review" button, so we know you are done. Thanks.

BarakHirshberg:
Hi Axel and Yotam,
Just found two small bugs in the testing of fmass and sp variables in the constructor of FixPIMD.
Added them to the same pull request, I hope that's OK.
Barak

yotamfe:
> Hi Axel and Yotam, Just found two small bugs in the testing of fmass and sp variables in the constructor of FixPIMD. Added them to the same pull request, I hope that's OK. Barak

The fix for fmass is consistent with documentation that says that the parameter can be set to the number of quasi-beads P (<code>np</code> in the code).

yotamfe:
@akohlmey I added the documentation. There's one point I'm not sure about: From what I gather from the code, it seems that the binary restart files contain only the state of the Nose-Hoover thermostat, and this is what I wrote in the documentation. (We didn't write this feature, nor have experience with it.) Does it make sense that the x,y,z positions are not stored in that file?

akohlmey:
> it seems that the binary restart files contain only the state of the Nose-Hoover thermostat

A fix only needs to store information in a restart that it "owns". This is how restarting is orgnized in LAMMPS. That is, position data will be stored by the atom style, potential data by the pair style and so on. Because of that, each command documents which of its properties it saves in a restart file or not. That information is needed so that can know whether a) a clean restart is possible at all and b) which commands need to be included in the input when restarting (e.g. whether the pair_style and pair_coeff commands need to be repeated). For fixes generally the command needs to be repeated with the same fix ID.

doc/src/fix_pimd.rst
```diff
@@ -149,6 +149,31 @@ related tasks for each of the partitions, e.g.
    restart 1000 system_${ibead}.restart1 system_${ibead}.restart2
    read_restart system_${ibead}.restart2
 
+Restart, fix_modify, output, run start/stop, minimize info
+"""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
+
+This fix writes the state of the Nose/Hoover thermostat over all quasi-beads to
```
BarakHirshberg: Thanks, @akohlmey, for explaining how the restart files work on LAMMPS.
In that context, do you think this phrasing is adequate?
Is that how fixes usually describe the restart files? Just stating what they add to the binary?

akohlmey: I will look at it later today. If needed, I will make some final changes myself.
I appreciate you both taking the time to add the missing text, even for the parts you didn't write. It is a big help.

PR #3620, opened 2023-01-26

### LINKED ISSUE #3618: [BUG] LAMMPS incorrectly converts non-numerical strings to numbers for evaluating math functions
**Summary**

I encountered an undocumented feature when using index-style (or string-style) variables in math functions. LAMMPS will quietly convert pure strings to 0 in such cases. Additionally, if the string starts with a number, e.g. "5a", LAMMPS will ignore everything after the first non-digit character and will use the first part as a number. I think an error should be raised instead.

Current develop
Linux Mint 20.3

**Expected Behavior**

LAMMPS should raise an error about trying to use a non-numerical string in math functions.

**Actual Behavior**

LAMMPS assigns a string a value of 0 or use only the numerical part of the string in further calculations.

**Steps to Reproduce**

Use the following lines as an input script:

```
variable i index "a"
print "$(5+v_i)"
```

LAMMPS will output 5 as if variable `i` was equal to 0.

Second case:

```
variable i index "5a"
print "$(5+v_i)"
```

Here, LAMMPS will output 10 as if variable `i` was equal to 5.

akohlmey:
**Summary**

This pull request combines multiple small changes and bug fixes.

**Related Issue(s)**

Fixes #3618 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- Check that the value of an index or similar variable is a number when used in a numerical expression
- Update fix pimd input examples so they can run without creating folders, run a shorter time, don't write dump files by default, and output the fix pimd scalar global properties.
- Update ML-HDNNP package to use a newer release (2.2.0) of the n2p2 library.
- Apply (some) changes suggested by clang-tidy
- Remove redundant 'this->' from MEAM sources, enable and apply clang-format
- Variable "np" in fix pimd must be initialized in the constructor
- Initialize recently added pointer in fix reaxff/species to NULL
- Make bundled Kokkos lib compatible with musl-libc

**Post Submission Checklist**


akohlmey:
@stanmoore1 This pull request removes the excessive and redundant use of `this->` in the meam pair style. It makes the code much more readable, IMO. I haven't touched the KOKKOS package version, though, since there are cases where it is needed, and I don't understand Kokkos well enough to know when. If you have the time, I would appreciate if you could do a pass and see if any of the excess uses of the "this" pointer have carried over to the KOKKOS version and remove them. No big deal if not.

lib/kokkos/core/src/Kokkos_Macros.hpp
```diff
@@ -653,7 +653,9 @@ static constexpr bool kokkos_omp_on_host() { return false; }
 #if (defined(KOKKOS_COMPILER_GNU) || defined(KOKKOS_COMPILER_CLANG) ||  \
      defined(KOKKOS_COMPILER_INTEL) || defined(KOKKOS_COMPILER_PGI)) && \
     !defined(_WIN32)
+#if (defined(__linux__) && defined(__GLIBC_MINOR__))
```
stanmoore1, line 656: @akohlmey should this be ported to upstream Kokkos?

stanmoore1 (approved):

stanmoore1:
I approved. @akohlmey this PR fixes some issues with regression tests, I'd prefer to get it merged ASAP for testing purposes.

lib/kokkos/core/src/Kokkos_Macros.hpp
akohlmey: Good question. I don't have an answer for that, only some explanation of why I made this change.

The proper solution is already in their version 4.0 branch, but that requires C++17.
I read a comment from @crtrott that they want to support the current version with bugfixes for some time, so they may be interested to include this. But I am using some kind of "inofficial heuristic" here to tell GNU libc apart from musl-libc since the latter does not define a macro that identifies it, so there is a small chance that it would not be the right heuristic for all Kokkos use cases. The musl-libc is Linux-only and rarely used outside of embedded system environments. But you *can* configure Arch and Gentoo Linux to use it instead of GNU libc, hence the bug-reports (I found one on the Arch website) and the (proper) fix in Kokkos 4.

My use case is rather exotic: I am using a Linux-to-Linux cross-compiler that targets building fully binary portable and statically linked non-MPI executables for LAMMPS and the LAMMPS shell program (which is not really possible in the same way with GNU libc). The main intent is to offer people a download to quickly run a fully patched and full featured binary of LAMMPS (stable or develop branch) without having to compile it themselves for testing if a bug can be reproduced with the respective current version. But this can also be a more lightweight alternative to distributing virtual machine images for LAMMPS tutorials. Or useful for people that just want to try out a few things with LAMMPS.

stanmoore1:
> @stanmoore1 This pull request removes the excessive and redundant use of `this->` in the meam pair style. It makes the code much more readable, IMO. I haven't touched the KOKKOS package version, though, since there are cases where it is needed, and I don't understand Kokkos well enough to know when. If you have the time, I would appreciate if you could do a pass and see if any of the excess uses of the "this" pointer have carried over to the KOKKOS version and remove them. No big deal if not.

I'm looking into this.

stanmoore1:
I looked through the use of `this->` in Kokkos meam and only saw one instance that can be removed, but it looks like a bug anyway since it is in device code. 

This PR is ready to merge from my POV.

lib/kokkos/core/src/Kokkos_Macros.hpp
stanmoore1: @crtrott @dalg24 FYI, see change and discussion above for Kokkos that may need to be ported v3.7.2

PR #3621, opened 2023-01-27
Small bugfixes

stanmoore1:
**Summary**

- Fix reading uninitialized variable in Kokkos ReaxFF
- Fix out of bounds access in pair_vashishta_kokkos with skip list
- Fix memory leak in Kokkos ReaxFF
- Fix memory leak in PairLJCutKokkos
- Fix logic for neigh/trim yes and multiple runs
- Fix invalid memory read in ReaxFF
- Fix bug in ReaxFF with pair hybrid, where it was missing interactions and giving wrong numerics
- Fix bug in Kokkos neigh list build for half list when `newton_pair` was different than the neigh list request (only happens for ReaxFF).

**Related Issue(s)**

None

**Author(s)**

Stan Moore (SNL), Axel Kohlmeyer (Temple)

**Licensing**

**Backward Compatibility**

Yes

### DIFF
--- a/src/KOKKOS/pair_reaxff_kokkos.cpp
+++ b/src/KOKKOS/pair_reaxff_kokkos.cpp
@@ -184,6 +184,8 @@ void PairReaxFFKokkos<DeviceType>::init_style()
   if (neighflag == FULL)
     error->all(FLERR,"Must use half neighbor list with pair style reaxff/kk");
 
+  need_dup = lmp->kokkos->need_dup<DeviceType>();
+
   allocate();
   setup();
   init_md();
@@ -196,6 +198,8 @@ void PairReaxFFKokkos<DeviceType>::setup()
 {
   int i,j,k,m;
   int n = atom->ntypes;
+
+  setup_flag = 1;
 
   // general parameters
   for (i = 0; i < 39; i ++)
@@ -577,17 +581,17 @@ void PairReaxFFKokkos<DeviceType>::Deallocate_Lookup_Tables()
     for (j = i; j <= ntypes; ++j) {
       if (map[i] == -1) continue;
       if (LR[i][j].n) {
-        sfree(api->control->error_ptr, LR[i][j].y, "LR[i,j].y");
-        sfree(api->control->error_ptr, LR[i][j].H, "LR[i,j].H");
-        sfree(api->control->error_ptr, LR[i][j].vdW, "LR[i,j].vdW");
-        sfree(api->control->error_ptr, LR[i][j].CEvd, "LR[i,j].CEvd");
-        sfree(api->control->error_ptr, LR[i][j].ele, "LR[i,j].ele");
-        sfree(api->control->error_ptr, LR[i][j].CEclmb, "LR[i,j].CEclmb");
+        sfree(LR[i][j].y, "LR[i,j].y");
+        sfree(LR[i][j].H, "LR[i,j].H");
+        sfree(LR[i][j].vdW, "LR[i,j].vdW");
+        sfree(LR[i][j].CEvd, "LR[i,j].CEvd");
+        sfree(LR[i][j].ele, "LR[i,j].ele");
+        sfree(LR[i][j].CEclmb, "LR[i,j].CEclmb");
       }
     }
-    sfree(api->control->error_ptr, LR[i], "LR[i]");
+    sfree(LR[i], "LR[i]");
   }
-  sfree(api->control->error_ptr, LR, "LR");
+  sfree(LR, "LR");
 }
 
 /* ---------------------------------------------------------------------- */
@@ -737,8 +741,6 @@ void PairReaxFFKokkos<DeviceType>::compute(int eflag_in, int vflag_in)
       d_s = k_s.view<DeviceType>();
     }
   }
-
-  need_dup = lmp->kokkos->need_dup<DeviceType>();
 
   // allocate duplicated memory
   if (need_dup) {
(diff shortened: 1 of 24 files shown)


stanmoore1:
I'm still tracking down two bugs: one in Kokkos with trimmed lists, and another in ReaxFF when the system is rotated.

src/KOKKOS/pair_reaxff_kokkos.cpp
```diff
@@ -199,8 +199,6 @@ void PairReaxFFKokkos<DeviceType>::setup()
   int i,j,k,m;
   int n = atom->ntypes;
 
-  setup_flag = 1;
```
stanmoore1: @akohlmey setting `setup_flag = 1` fixed a memory leak when using Kokkos ReaxFF, since without that it won't deallocate some of the CPU arrays in the parent destructor. Not sure how to fix the memory leak without this.

stanmoore1: Here are the lines I was referring to: https://github.com/lammps/lammps/blob/develop/src/REAXFF/pair_reaxff.cpp#L128-L141

akohlmey: @stanmoore1 well, *with* it, the destructor is trying to deallocate memory that hasn't been allocated and that leads to crashes (e.g. in the unit tests). By setting the flag without calling the allocate function of the base class, you are creating inconsistent behavior. Perhaps the KOKKOS allocate() function just needs to call the base class allocate() function?

stanmoore1: Is it trying to deallocate wild pointers? Maybe setting all the pointers to `nullptr` in the base class would help?

akohlmey: Checked that, didn't help. Let me have a closer look again. Now with the additional input from you about why you want to deallocate stuff in the base class, there is a way.

stanmoore1: OK, we could call the base `allocate` as you suggested, or I can try to make it be more careful of only deallocating what is allocated.

akohlmey: That doesn't seem to be the issue. It looks more like the allocations that are leaking originate from calling init_md().

akohlmey: This is *really* tricky business: the sequence of what is allocated when is different in the KOKKOS version and the REAXFF version. I suspect we can only make this consistent by having different flags for each group of allocations.
ALso, for example, the KOKKOS allocate() does something different from the REAXFF allocate does (which is called in both cases from coeff(), while the former is only called from init_style()). Quite confusing...

stanmoore1: OK thanks for looking at this

akohlmey: @athomps @sjplimp do you have any suggestions for how to best consolidate this?

stanmoore1: I think I can fix this--give me a bit

stanmoore1: @akohlmey a92a3ec fixes the memory leak for my reproducer, but not sure if it will still segfault in the unit tests.

akohlmey: It looks like there is a SNAFU with the N2P2 repository. The downloaded archive with the given tag has suddenly a different MD5 sum. Thus a bunch of tests seem to be failing because they fail to compile when they download and check the library.

stanmoore1:
Looks like the CMake build is failing with:
```
CMake Error at Modules/Packages/ML-PACE.cmake:8 (file):
  file DOWNLOAD HASH mismatch

    for file: [/data/builder/workspace/dev/pull_requests/ubuntu/unit_tests/cmake_mpi_openmp_smallbig_shared/build/libpace.tar.gz]
      expected hash: [f418d32b60e531063ac4285bf702b468]
        actual hash: [5574b1d0dd7fad268be0cd54451e5b43]
             status: [0;"No error"]
```

src/KOKKOS/pair_reaxff_kokkos.cpp
stanmoore1: @akohlmey I think you mean PACE, not N2P2?

akohlmey: Hmmm... seems to affect both

akohlmey: ARRRGGGHHH!!!!

stanmoore1:
@akohlmey possibly related to https://github.blog/changelog/2023-01-30-git-archive-checksums-may-change/?

akohlmey:
> @akohlmey possibly related to https://github.blog/changelog/2023-01-30-git-archive-checksums-may-change/?

Most likely. But it seems they have underestimated what kind of chaos this will cause and are reverting the change (for now).

stanmoore1:
For 0870bc3, there was a bug in Kokkos when requesting a half list with newton off, when it didn't match the `newton_pair` flag. However, it doesn't show up with the regular regression tests (examples/reax). Also, when using a half-from-full neigh list, it had correct behavior.

stanmoore1:
I fixed all the bugs I was seeing. Running this through regression testing.

akohlmey:
This branch now passes all ReaxFF unit tests on my local machine with Kokkos/OpenMP enabled.

stanmoore1:
Regression tests failed from the github checksum issue, rerunning now.

athomps (approved):

akohlmey (approved):

stanmoore1:
With this PR + #3620, regression tests pass.

PR #3628, opened 2023-01-31
Add a fix mdi/qmmm command for running QMMM via MDI code coupling

sjplimp:
**Summary**

Add an MDI-based fix for performing QMMM when coupled to a quantum code.  Examples for doing this with
the PySCF and NWChem quantum codes are included. Two modes of QMMM coupling
are supported.  One called "potential", where LAMMPS provides a Coulomb potential to the QM code.  The other
called "direct", where the coords and charges on MM atoms are provided to the QM code.  Examples of coupling
LAMMPS to the LATTE tight-binding semi-empirical code in a similar manner for AIMD using fix mdi/qm are also provided.

Files and dirs:
src/MDI/fix_mdi_qmmm.cpp/h
examples/QUANTUM/PySCF
examples/QUANTUM/NWChem
examples/QUANTUM/LATTE

**Related Issue(s)**

N/A

**Author(s)**

Steve, with help from ...
Chenghan Li (Caltech) for PySCF
Eric Bylaska (PNNL) for NWChem
Chrisitan Negre (LANL) for LATTE 

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

**Post Submission Checklist**


sjplimp:
@akohlmey Axel - two things you may want to look at.  

(1) Three files now have a listing of chemical symbols for each element and input arg code to scan them and convert to an atomic number.  Look for "symbols" in mdi_engine.cpp, fix_mdi_qm.cpp, fix_mdi_qmmm.cpp.  Maybe a single copy of symbols (or chemical_symbols) should be put someplace else?  Maybe the lookup code should be a utils method?

(2) In mdi_plugin.cpp there is new code to do variable substitution in two command line args which in practice are often multiple words (in quotes).  Maybe there should instead be a utils method to pass it a string and return a new (longer) string as a char * which is the substituted version?  Then this would be one-line operation in this file, and maybe also useful in other commands which process mutli-word arguments.

akohlmey:
@sjplimp 
 
> (1) Three files now have a listing of chemical symbols for each element and input arg code to scan them and convert to an atomic number. Look for "symbols" in mdi_engine.cpp, fix_mdi_qm.cpp, fix_mdi_qmmm.cpp. Maybe a single copy of symbols (or chemical_symbols) should be put someplace else? Maybe the lookup code should be a utils method?

Long ago, I wrote something along those lines for VMD plugins: 
https://github.com/thatchristoph/vmd-cvs-github/blob/master/plugins/molfile_plugin/src/periodic_table.h
This may also be useful for some applications that use typelabels (or want to create labels from numbers or elements from labels or numbers from labels).

This is rather specific for "chemical" applications, so perhaps it may be better to start a new namespace for that.

> (2) In mdi_plugin.cpp there is new code to do variable substitution in two command line args which in practice are often multiple words (in quotes). Maybe there should instead be a utils method to pass it a string and return a new (longer) string as a char * which is the substituted version? Then this would be one-line operation in this file, and maybe also useful in other commands which process mutli-word arguments.

Would something like this be suitable?
https://github.com/lammps/lammps/blob/develop/src/LEPTON/lepton_utils.cpp#L116

It is perhaps more obvious how it works by looking at the corresponding unittest code: https://github.com/lammps/lammps/blob/develop/unittest/utils/test_lepton.cpp#L64
or the pair style docs (look for the paragraph that starts with "As an extension"):
https://docs.lammps.org/pair_lepton.html

sjplimp:
@akohlmey I removed the WIP tag from this PR.  It is ready for you to look-over and potentially merge.

I have not changed the use of symbols (3 places).  I like your idea of a new namespace for that tabulation, so if you want to do that, it would be great.

For the other code duplication in mdi_plugin.cpp, I think the LEPTON code is overkill.  B/c the Input::substitute() method
already does what is needed.  I was just suggesting a small wrapper method on that which would make it a one-line call from mdi_plugin.cpp or other classes, to manage the string allocations.  But wasn't sure if utils.cpp was the right place for it.  If is is, I can add the new method and you can look it over.

akohlmey:
> I have not changed the use of symbols (3 places). I like your idea of a new namespace for that tabulation, so if you want to do that, it would be great.

What would be a good name for the namespace?

"periodic_table" would be a bit long, "periodic" misleading, "pte" (for periodic table of elements) too short and difficult to recognize as what it stands for. We generally discourage initialisms unless they are widely used, anyway, and should follow our own advice. This is a bit of a corner case, though. "chemical" or "chemistry" would be too generic. For me personally, "pte", would do, but I don't want to go against the advice that I usually give others.

> For the other code duplication in mdi_plugin.cpp, I think the LEPTON code is overkill.

Are you sure we are talking about the same thing? I was referring to just a single function, that is standalone and mostly included in the LEPTON package, because it is not used elsewhere.

> B/c the Input::substitute() method already does what is needed.

`Input::substitute()` is one of the pieces of LAMMPS where we still heavily depend on C-style string handling and pointer magic and specifically that needs a scratch buffer. Also the Input class makes occasional assumptions about the state of the parsing, so I am rather opposed to exposing this to general use cases. If the input parsing would not be so complex and crucial to LAMMPS, I would have long ago tried to refactor it to get rid of the scratch buffer and make it a standalone function using mostly C++ string processing (and our own utility functions).

The `LeptonUtils::substitute()` function uses the `v_name` convention instead of `${name}` to avoid conflicts with the input file parsing (i.e. the *use* of Input::substitute()). I have not looked at your code yet to see whether the same kind of issue might apply in your case, too. 



> I was just suggesting a small wrapper method on that which would make it a one-line call from mdi_plugin.cpp or other classes, to manage the string allocations. But wasn't sure if utils.cpp was the right place for it. If is is, I can add the new method and you can look it over.

athomps:
I think the namespace could be called `Elements` or maybe `MDIElements`

sjplimp:
@akohlmey @athomps 

For the namespace (or variable) name, I think ChemicalSymbols or ElementNames is most descriptive.  It's not really MDI specific, thought that's all that is using it now.  MDI is only using it to convert a user-input element name to an atomic number,
based on its location in the list.  Dump/dump_modify element doesn't need to do that, and can be used for any label the user
chooses, not just element names.

For replacing variables in strings, a new (less C-like) function in utils and removing Input::substitute() is fine.
But note that the syntax here is not v_name, but $ n or ${n2} since it is coming from a line in the input script (same as Input::substitute).

I meant that the substitute() function in LEPTON was overkill since we already had almost what we needed.  But replacing
it with a more C++ function is fine.

akohlmey:
> MDI is only using it to convert a user-input element name to an atomic number,
> based on its location in the list.

There are additional applications for that kind of information. Apart from the element symbol to index in the periodic table mapping, there could be functions trying to guess the element from its mass (for dump output, e.g. as `dump_modify element guess`), or to provide a covalent radius for visualization with dump image.
Also, it could be used by the zbl pair style or similar to allow entering element symbols instead of atomic numbers.

sjplimp:
@akohlmey OK - I think is ready for you to look over and then will be ready to merge

akohlmey:
> @akohlmey OK - I think is ready for you to look over and then will be ready to merge

It will take a little while until I have the time to look into this one with the necessary care and detail (will need to build and set up the QM codes, too). I have a few other ongoing things that I want to finish first.

akohlmey:
@sjplimp fixed up some issues with the documentation (spelling, broken links, formatting).

What about the example in `examples/PACKAGES/mdi`? It was using an outdated LAMMPS syntax (fixed that, too) and didn't show how to use MPI (added that). This is minimal so it is useful, but it should probably be merged into the `examples/mdi` folder.

akohlmey:
@sjplimp FYI, I installed PySCF from the git repo in the README file and it has the same behavior than the version that can be downloaded via `python -m pip install pyscf`

BTW: the "python way" to install from source the cloned repo is to do the following in the main folder:
```
python -m pip install --user --upgrade pip
python -m pip install --user build wheel
python -m build -n --wheel -o .
python -m pip install pyscf-2.*.whl
```
And even cleaner would be to use a virtual environment.

sjplimp:
@akohlmey I think I made all the changes we talked about

akohlmey:
@sjplimp I'd like to make a few more changes, if you don't mind.
- when cloning the git repository, use a https://github.com URL instead of git@github.com. The former works for everybody, the latter only for people with a GitHub account that also uploaded an ssh key
- when cloning name the cloned directory like the software, not "git". This is the common convention and the default behavior when you do not provide a target directory. Personally, I also find it confusing to have all these directories called "git" that have no indication of what is in them.
- when installing the MDI python package, put "pip" first and "conda" second. Pip is a far more lightweight python package management tool and integrated into the default python for most distributions. Conda requires an additional setup first. Also, in my experience, conda packages are not always managed with as high a quality and consistency than the pip ones. At least for the cases that I have come across.
- when compiling/installing PySCF, I would like to add the building/installing of the wheel file using python as the first option. That is far simpler and less "hack-ish" than the "in place" usage with updating environment variables.

akohlmey (approved):

sjplimp:
@akohlmey All 4 of your suggested changes sound fine.  These would all just be changes to the 3 README files for each QM code?

akohlmey:
> @akohlmey All 4 of your suggested changes sound fine. These would all just be changes to the 3 README files for each QM code?

That is correct. Please check out: https://github.com/lammps/lammps/pull/3628/commits/667ba0febb74b724b75e2760d26e8025f3c6385d

PR #3631, opened 2023-02-01
Disable Kokkos GPU <--> CPU overlap when using pair hybrid with non-K…

stanmoore1:
…okkos styles

**Summary**

Fix a bug on GPUs when not using UVM when using non-Kokkos pair styles with pair hybrid.
Also fix a bug in Kokkos ML-IAP on GPUs without UVM

**Related Issue(s)**

None

**Author(s)**

Stan Moore (SNL)

**Licensing**

**Backward Compatibility**

Yes


stanmoore1:
I'm seeing one failing regression test with GPU

stanmoore1:
The failing regression test was a false positive. This is ready to merge from my POV.

athomps (approved):
I approve.

PR #3632, opened 2023-02-02
Implement a "caching in LDS" approach for EAM to improve performance in HIP

skyreflectedinmirrors:
- Use a TeamPolicy rather than a RangePolicy to expose scratch
	- Team cooperates to load row-zero of spline into LDS
	- At runtime, decide whether value can be 'cached' in LDS, and conditionally load from LDS based on this

Should be easily extendable / tunable for other architectures if ever required.

**Summary**

Implement a simple 'cache' in LDS for AMD GPUs using TeamPolicies.  This uplifts EAM performance by ~35% in some cases.

**Related Issue(s)**

N/A

**Author(s)**

Ian Bogle [AMD]
Nicholas Curtis [AMD]
Leopold Grinberg [AMD]

**Licensing**

**Backward Compatibility**

No impact

**Implementation Notes**

Verified correctness / performance uplift on a single GCD of an MI-200.
Correctness checked against unmodified develop results.

**Post Submission Checklist**

N/A


stanmoore1:
Thanks @arghdos, will take a look

src/KOKKOS/pair_eam_kokkos.cpp
```diff
@@ -31,9 +31,25 @@
...
 #include <cmath>
-
 using namespace LAMMPS_NS;
 
+#define MAX_CACHE_ROWS 500
+#define MAX_CACHE_COLS 7
+
+/* ---------------------------------------------------------------------- */
+template<class DeviceType, class tag>
+auto policyInstance(int inum){
```
weinbe2: This is a really nice design, I like it. I think it might make more sense as a functor member function so you can query info from the functor --- for example, instead of having `MAX_CACHE_ROWS` you can query the compile-time `7` in the size of the spline structure.

src/KOKKOS/pair_eam_kokkos.cpp
```diff
@@ -31,9 +31,25 @@
...
+  #ifdef KOKKOS_ENABLE_HIP
```
weinbe2: These optimizations also don't seem particularly special to one type of GPU :)

src/KOKKOS/pair_eam_kokkos.cpp
```diff
@@ -799,6 +843,232 @@ void PairEAMKokkos<DeviceType>::operator()(TagPairEAMKernelC<NEIGHFLAG,NEWTON_PA
...
+/* ---------------------------------------------------------------------- */
+
+////Specialisation for Neighborlist types Half, HalfThread, Full
+template<class DeviceType>
+template<int EFLAG>
+KOKKOS_INLINE_FUNCTION
+void PairEAMKokkos<DeviceType>::operator()(TagPairEAMKernelAB<EFLAG>,
+		                     const typename Kokkos::TeamPolicy<DeviceType>::member_type& team_member,
+				     EV_FLOAT& ev) const {
+  int ii = team_member.league_rank()*team_member.team_size() + team_member.team_rank();
```
weinbe2, line 842: There's never a check of if `ii < imax`, so if your number of atoms doesn't divide evenly into the `team_size()` (hard-coded as 1024 in the initial commit) you'll get some nasty out-of-bounds accesses. I see that you use all of the threads in a team as part of the load into scratch memory, so you could put the `ii < imax` check after that.

src/KOKKOS/pair_eam_kokkos.cpp
```diff
@@ -799,6 +843,232 @@ void PairEAMKokkos<DeviceType>::operator()(TagPairEAMKernelC<NEIGHFLAG,NEWTON_PA
...
+  // rho = density at each atom
+  // loop over neighbors of my atoms
+  const int m_max = d_rhor_spline.extent_int(1);
+  const int j_max = d_rhor_spline.extent_int(2);
```
weinbe2: iirc this is a compile-time 7

src/KOKKOS/pair_eam_kokkos.cpp
skyreflectedinmirrors: Our Ampere is dead for the moment, haven't been able to test if it matters for you guys :P

weinbe2:
This is a really nice PR, I'm grateful y'all did the lifting to put in scratch memory caching of some of the splines. I'm also a fan of the `policyInstance` abstraction and I'm sure I'll be using it elsewhere. I left a few comments (which it seems like y'all have already seen!).

src/KOKKOS/pair_eam_kokkos.cpp
skyreflectedinmirrors: IIRC, Ian did test this w/ this hard-coded and we didn't see much of a difference in performance, but I ... think we can change this to a [`static_extent` call](https://github.com/kokkos/kokkos/blob/develop/core/src/Kokkos_View.hpp#L652) such that if the spline degree ever changed (is this fixed for EAM?) we wouldn't miss it there.

src/KOKKOS/pair_eam_kokkos.cpp
skyreflectedinmirrors: This is a good idea.

src/KOKKOS/pair_eam_kokkos.cpp
weinbe2: I think it's fixed for this particular flavor of EAM, it's related to the spline order. You could probably do some type of `static_assert` in the functor to see if the `*7` in `tdual_ffloat_2d_n7`, etc, ever changes as a sanity check? Edit: or just what you did above in the function proper, don't mind me...

I'm not surprised it's not making a difference in performance, to a degree this is purely aesthetic... but also, may as well give the compiler a hard-coded number when you can.

src/KOKKOS/pair_eam_kokkos.cpp
skyreflectedinmirrors: If you get a chance @weinbe2 -- give this a test, and we can turn it on for y'all (or just LMP_ENABLE_GPU), our A100 box still isn't alive yet.

skyreflectedinmirrors:
I believe we got all of the comments except the "enable for other GPUs" which I'll leave for others to test :)

weinbe2:
**First off:** this is the leading-order problem on V100 and A100:

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Kokkos::Impl::ParallelReduce< Cuda > requested too large team size.
```

This was ultimately due to an "edge effect" --- the amount of shared memory needed for the spline lookup plus the reduction for a blocksize of 1024 threads exceeds 48KiB but is less than 64KiB (the LDS size on MI2***). By default on Volta/Ampere/Hopper a single block can only access a maximum of 48iB, so it barfs. (For posterity: you can opt into a single block being able to access more shared memory, up to the entire max shared memory carveout, via the CUDA API function `cudaFuncSetAttribute(my_kernel, cudaFuncAttributeMaxDynamicSharedMemorySize, 98304);`, where that number corresponds to the 96KiB max on Volta. Kokkos doesn't support that yet, so I'll get to that at some point.)

A quick workaround is testing a blocksize of 256 or 512, which leads to some redundant shared memory staging, but everything's already in L1 the first time around, so all's good.

**Performance after the workaround:**

It's ultimately still slower, by ~0 - 10% (with exceptions) on V100-16GB-SXM, and up to 20% on NVIDIA A100-80GB-SXM (see figures).

After thinking about it for a while, it does seem reasonable: EAM is very L2 limited on those architectures, and it's not lacking for L1 (there's a wee bit more than 16KB). It's plausible that the spline lookups are causing a negligible amount of L1 evictions. Further, it's reasonable that some amount of the lookup table is always going unused --- such as values corresponding to a very small distance between atoms, where there's a decently high potential barrier (the EAM potential drops exponentially) --- so just counting on the L1 avoids ever needing to load a subset of the data. tl;dr, the large L1 cache "doing its thing" is working more than well enough.

(The outlier is V100 at ~8 million atoms --- in this case, we're falling out of good L2 cache reuse, so _any_ cache management at any level is helping out. I don't really think it's worth losing our minds making an exception for V100 at ~> 4 million atoms.)

**tl;dr** it's slower on V100 and NVIDIA A100, the `#ifdef HIP` can indeed stay `HIP` as far as I'm concerned.

image
image

skyreflectedinmirrors:
Thanks Evan, that's really helpful. You bring up a few good points:

1. I didn't run a size-sweep, but that's absolutely worth doing -- so on one of our MI-210's shows this is a win for a broad set of sizes (i.e., the x/y/z size fed into the Cu_u3 bench: https://github.com/lammps/lammps/blob/develop/bench/in.eam):

image

To understand why, I'll just leave this [Chips & Cheese](https://chipsandcheese.com/2022/09/18/hot-chips-34-amds-instinct-mi200-architecture/) article here.  Ctrl+F for "L1" and imagine my envy :P.

Anyways, this brings me to the second point -- the reason we're not exceeding the LDS limit on the team size here is _likely_ because the ParReduce is taking the shuffle reduction in the HIP backend of Kokkos.  IIRC, this was because the compiler used to generate very 'interesting' code for the LDS-based reductions of large structs.  So, for the purposes of this PR, we might want to limit this LDS caching scheme to ParFors, such that any future change to the HIP backend doesn't suddenly cause the team-size failure @weinbe2 saw here.  Given that these happen relatively infrequently (usually: 1/50 - 1/100 steps), I doubt it will make of a perf difference.

weinbe2:
> To understand why, I'll just leave this [Chips & Cheese](https://chipsandcheese.com/2022/09/18/hot-chips-34-amds-instinct-mi200-architecture/) article here. Ctrl+F for "L1" and imagine my envy :P.

Haha, yeahhhhhh. But yes, the behavior in that figure definitely makes sense.

> Anyways, this brings me to the second point -- the reason we're not exceeding the LDS limit on the team size here is likely because the ParReduce is taking the shuffle reduction in the HIP backend of Kokkos. IIRC, this was because the compiler used to generate very 'interesting' code for the LDS-based reductions of large structs.

Ah, that's interesting to know, re: the first point. The second part's a nice fun fact... or I guess the opposite of a fun fact, an anti-easter egg?

> Anyways, this brings me to the second point -- the reason we're not exceeding the LDS limit on the team size here is likely because the ParReduce is taking the shuffle reduction in the HIP backend of Kokkos.

I should look and see why the CUDA backend isn't doing some type of smart shuffle for part of the reduction, to be honest, at least above some struct size --- on face value, the kernel should have only taken up `500 * 7 * sizeof(double)` == ~27.3KiB for the lookup table, plus `sizeof(EV_FLOAT) x (1024 / 32)` == 2 KiB for the reduction (baking in the block size of 1024), so you'll imagine my surprise when we were blowing out shared memory...

Anyway, yes, I agree with your idea of only using this optimized flavor for `parallel_for` and not `parallel_reduce`... for the HIP backend :)

skyreflectedinmirrors:
Stan, I think the latest commit should revert to just using the caching for `ParFor`, it does drop perf a tad (about 3.5%), probably because EAM does thermo steps every 50 timesteps instead of 100 like some of the others, but overall it's still a huge win over where we were w/o the caching (~31% in net)

stanmoore1:
@arghdos have you tried using `-pk kokkos neigh half newton off` for the EAM benchmark? That is fastest for V100, so maybe MI250X with FP64 hardware atomics as well. If it is faster, we may want to look at porting the `HALF` code paths to use a TeamPolicy as well, since currently it is only enabled `FULL`.

skyreflectedinmirrors:
@stanmoore1 -- it's been a few years since I've tried, but I imagine the same optimization will help there as well.

stanmoore1:
@arghdos I wanted to port this to `half` neighbor list, but haven't had time yet. Since it works correctly for `full` neigh list, I think we should merge this as is--can port to `half` list in the future.

stanmoore1:
I confirmed this gives correct numerics on MI250X and speedup as well.

stanmoore1 (approved):

stanmoore1:
@arghdos @IanBogle Using a "half" neighbor list is still faster than "full" for EAM on AMD MI250X GPU, so it could be beneficial to port KernelA and KernelB to use a team policy (this only ported the fused KernelAB for "full" list). However this also sped up "half" since it ported KernelC.

stanmoore1:
However, I think we should merge this PR now, can do more in the future.

skyreflectedinmirrors:
Oh boy, been awhile since I thought of this one -- definitely didn't realize half went down the un-fused kernel route :P.
I agree we can probably merge as is, I'll add porting to the unfused kernels to our list (it may be awhile though)

PR #3635, opened 2023-02-05
Review grammar in Developer guide and Howto pages

akohlmey:
**Summary**

This applies (some) grammar fixes suggested by languagetool.org for the first part of the Developer guide section of the manual and some of the Howto pages.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Post Submission Checklist**

### DIFF
--- a/doc/src/Developer_org.rst
+++ b/doc/src/Developer_org.rst
@@ -15,7 +15,7 @@ for more information about that part of the build process.  LAMMPS
 currently supports building with :doc:`conventional makefiles
 <Build_make>` and through :doc:`CMake <Build_cmake>`.  Those procedures
 differ in how packages are enabled or disabled for inclusion into a
-LAMMPS binary so they cannot be mixed.  The source files for each
+LAMMPS binary, so they cannot be mixed.  The source files for each
 package are in all-uppercase subdirectories of the ``src`` folder, for
 example ``src/MOLECULE`` or ``src/EXTRA-MOLECULE``.  The ``src/STUBS``
 subdirectory is not a package but contains a dummy MPI library, that is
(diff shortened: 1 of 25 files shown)


sjplimp (approved):

doc/src/Developer_org.rst
```diff
@@ -102,42 +103,44 @@ LAMMPS while passing it the command line flags and input script. It
...
-*hybrid* class that takes the place of the one allowed instance and then
-manages and forwards calls to the corresponding sub-styles for the
-designated subset of atoms or data.  The composite class may also have
-lists of class instances, e.g. Modify handles lists of compute and fix
-styles, while Output handles a list of dump class instances.
+*hybrid* class instance that acts as a proxy, and manages and forwards
+calls to the corresponding sub-style class instances for the designated
+subset of atoms or data.  The composite class may also have lists of
+class instances, e.g. ``Modify`` handles lists of compute and fix
+styles, while ``Output`` handles a list of dump class instances.
```
athomps, line 143: Should we make all LAMMPS class names ``inline literal``?

doc/src/Howto_bpm.rst
```diff
@@ -3,16 +3,16 @@ Bonded particle models
 
 The BPM package implements bonded particle models which can be used to
 simulate mesoscale solids.  Solids are constructed as a collection of
-particles which each represent a coarse-grained region of space much
-larger than the atomistic scale. Particles within a solid region are
+particles, which each represent a coarse-grained region of space much
+larger than the atomistic scale.  Particles within a solid region are
```
athomps, line 7: Oh-oh, are we going to have to put two spaces after *every* period?

stanmoore1: Please no...

athomps (approved):
I approve. See optional suggestion in comment on ``inline literal``

doc/src/Developer_org.rst
akohlmey: In an ideal world, those would be documented with Doxygen comments and their API would be imported into the Developer's Guide reference section and then those would be referred to with ":cpp:class:`Modify <LAMMPS_NS::Modify>`" . But we are still a long way away from that. When I see it, I use "``" markup for files, class names/functions, and defines or CMake variables. To make them stand out more. But I don't see this as a "must have" change.

doc/src/Howto_bpm.rst
akohlmey: No. Those are usually inserted automatically by emacs when I reindent or rewrap a block of text. Same as for the class name, this is a "can have" but not a "must have" change.

doc/src/Developer_org.rst
akohlmey: We have arrived at over 2800 pages (in PDF format) in the manual, so there is a lot of "legacy text" around that was added before we started to pay closer attention to consistency.  Keep in mind that most files were auto-converted to .rst markup from txt2html markup and that had its issues with some corner cases (even of we have by now eliminated 99.99% of those, at that ratio and with 2800 pages, there would still be 28 pages with a leftover issue)

PR #3636, opened 2023-02-06
Correct typos and clarify fix_controller.rst

srtee:
**Summary**

Fixes some typos and adds math expression c_0 for the initial control variable value; removes the "E-hat" term from equations (which doesn't appear elsewhere in the support doc or the source code)

**Author(s)**

Shern Tee, University of Queensland

**Licensing**


athomps (approved):
Yes, the \hat{E} term is inconsistent with the implementation and should be removed.

PR #3639, opened 2023-02-06
Heffte integration

mkstoyanov:
**Summary**

Replaces the builtin MPI communication algorithms for FFTs with heFFTe. The replacement is controlled via a CMake switch `-DFFT_HEFFTE` and the heFFTe library can be either independently installed or downloaded and compiled as part of this project.

Currently, heFFTe is simply a replacement inside the wrapper data-structures and no FFT calls have been modified.

**Related Issue(s)**

No current issue, although a PR was done in the past.

**Author(s)**

Miroslav Stoyanov (ORNL) and Stanimire Tomov (ICL)

**Licensing**

**Backward Compatibility**

Fully backwards compatible.

**Implementation Notes**

Changed the calls inside the FFT wrapper, all tests ran and passed.

**Post Submission Checklist**


stanmoore1 (commented):
This looks clean with minimal changes, nice job. @mkstoyanov is there a reason only FFTW and MKL are supported? How hard would it be to add support for cuFFT and hipFFT?

mkstoyanov:
cuFFT and hipFFT require synchronization between Kokkos and heFFTe backends. It's not much, but it is more and it does belong to a different set of files.

I don't want to overwhelm the PR with too many changes, I will add cuFFT and hipFFT in a separate PR.

sjplimp:
@mkstoyanov Thanks for submitting, as Stan said this does look nice and clean.  A few additional issues to address and/or clarify:

(1) Add support for traditional make, not just Cmake - all LAMMPS options support both.  For make, I think this would mean adding a lib/heffte dir and an Install.py script within it.  See lib/mdi for a good example.  The Install.py will allow for downloading and building heFFTe in that dir (via its own Cmake), or pointing to an existing installation of heFFTe on someone's system.

(2) Related to (1), the traditional make will need compile-time variables set and recognized by the build with heFFTe.  These are documented in doc/src/Build_settings.rst - see the FFT sub-section.  Note that -DFFT_HEFFTE would conflict with -DFFT_FFTW or -DFFT_MKL, so probably there needs to be a another variable to toggle between packages (heFFTe vs native) and the existing one for the 1d FFT specification (FFTW, MKL, KISS).  I.e. it looks like heFFTe needs both those flavors of variables.

(3) Related to Stan's Q, not clear if heFFTe in LAMMPS will run on GPUs (if available) on on the CPU.  Likewise is there overlap/collision between the support in LAMMPS via Kokkos for running FFTs on the GPUs with Cuda lib calls.  If heFFTe will be CPU-only in LAMMPS, then possibly something needs to be disabled when running the Kokkos version of PPPM with Kokkos FFTs.

(4) Somewhere in the doc pages there should be a paragraph or two added to explain to a LAMMPS user the motivation for trying or using heFFTe instead of the native LAMMPS FFTs.  E.g. what kind of hardware or what kind of use cases are likely to run faster with heFFTe, and by how much (percentage).  E.g. huge FFT grids, or large proc counts, or problems where KSpace dominates, or ??  For now, this motivation could go in 3.5.2 of the manual, but we could also think about putting it somewhere more visible.  E.g. in the kspace_style pppm section.

mkstoyanov:
I do not intend this PR to be the end-all of the Lammps-heFFTe integration. This integrates the CPU version of heFFTe with the CPU version of Lammps, it is self-contained, and it will be one of the building blocks towards the complete integration. Experience has taught me that several small and self-contained PRs are much better than a gigantic blob touching many aspects and issues.

Specifically to the questions:

(1) OK, I can do that.

(2) Right now, there is no conflict between the compile time options. The `-DFFT_HEFFTE` options toggles between native backend and heFFTe, while the `-DFFT_FFTW` and `-DFFT_MLK` select the backend. Thus `-DFFT_FFTW` meand native FFT with FFTW while `-DFFT_HEFFTE` with  `-DFFT_FFTW` means heFFTe + FFTW. I will do that with CUDA and HIP.

(3) The GPU support is intended for another PR.

(4) I can add the doc, but things will change. Right now, the main advantage comes from the pipelined version of the point-to-point communication, where we overlap packing, sending and receiving. The original Lammps code overlaps only packing and receiving (packing and sending are not overlapped). This leads to improvement for moderate to large size problems when using the point-2-point algorithm. 

An additional advantage can be exploited if we change the FFT wrapper class and allow for real inputs. Currently, Lammps converts read data to complex (by padding with zeros) then starts MPI operations, effectively sending a bunch of zeros on the first pass (in the brick to brick case). heFFTe can take the real data, communicate brick-2-pencil without padding, then pad and continue. The result is 10 - 15% less data being communicated. But I don't want to make all changes in a single PR.

sjplimp:
@mkstoyanov Sounds good - thanks for the clarifications.

Re (4) - I suggest just documenting motivation for the current PR (which could include promises of future features X,Y,Z), so users are educated.  Then future PRs can amend the doc/motivation.

akohlmey:
> The `-DFFT_HEFFTE` options toggles between native

This is a bit confusing within the context of how LAMMPS selects defines. -DFFT_XXX should be an indication of the backend FFT. Please also note that with traditional make it is possible to not set any -DFFT_XXX define and then it falls back to the internal KISS FFT routines. More consistent with other features from packages would be to have a define -DLMP_HEFFTE to enable the alternate code paths enabled and provided by the package. 

Another question is: does HEFFTE support -DFFT_SINGLE?
If not, there should be protection to error out at compile time if not.

mkstoyanov:
> This is a bit confusing within the context of how LAMMPS selects defines. -DFFT_XXX should be an indication of the backend FFT. Please also note that with traditional make it is possible to not set any -DFFT_XXX define and then it falls back to the internal KISS FFT routines. More consistent with other features from packages would be to have a define -DLMP_HEFFTE to enable the alternate code paths enabled and provided by the package.

Let me get this right, you want `-DFFT_HEFFTE` to enable heFFTe but also ignore/conflict with `FFT_FFTW` and `FFT_MKL`? 

Then have a separate variable `-DLMP_HEFFTE` that would be set `FFTW` or `MKL` or any future backend?

> 
> Another question is: does HEFFTE support -DFFT_SINGLE? If not, there should be protection to error out at compile time if not.

heFFTe uses C++ templates and overloads to automatically handle both single and double precision.

Related note: heFFTe doesn't support KISS

akohlmey:
> Let me get this right, you want `-DFFT_HEFFTE` to enable heFFTe but also ignore/conflict with `FFT_FFTW` and `FFT_MKL`?
> 
> Then have a separate variable `-DLMP_HEFFTE` that would be set `FFTW` or `MKL` or any future backend?

No. When the package is installed it sets -DLMP_HEFFTE **instead of** `-DFFT_HEFFTE` and that enables the code path using the HeFFTe library instead of the LAMMPS FFT wrappers.

Any choice of backend FFT that is not supported should lead to a compile time error.

> heFFTe uses C++ templates and overloads to automatically handle both single and double precision.

That is not what I am asking. Does it set/use/interface the FFT_SCALAR data type that is used elsewhere in LAMMPS to select between single and double precision FFTs and then compile, link, and run correctly when -DFFT_SINGLE is set.

LAMMPS has been around a long time and there are some quirky combination of settings that can happen. The FFT support affects multiple features, so we will have to go through a whole bunch of permutations of settings that will need to be either confirmed to compile/link/run or should be protected against with preprocessor checks leading to compile time errors.

As you plan to integrate your feature in stages, this will also need to be done in stages. I am just mentioning it now so it doesn't come as a surprise later.

cmake/Modules/Packages/KSPACE.cmake
```diff
@@ -46,6 +46,36 @@ else()
...
+if(FFT_HEFFTE)
+  # if FFT_HEFFTE is enabled, switch the builtin FFT engine with Heffte
+  if(FFT STREQUAL "FFTW3") # respect the backend choice, FFTW or MKL
+    set(HEFFTE_COMPONENTS "FFTW")
+    set(Heffte_ENABLE_FFTW "ON" CACHE BOOL "Enables FFTW backend for heFFTe")
+  elseif(FFT STREQUAL "MKL")
+    set(HEFFTE_COMPONENTS "MKL")
+    set(Heffte_ENABLE_MKL "ON" CACHE BOOL "Enables MKL backend for heFFTe")
+  else()
+    message(FATAL_ERROR "Using -DFFT_HEFFTE=ON, requires FFT either FFTW or MKL")
```
mkstoyanov, line 59: @akohlmey CMake will issue a configuration error if the wrong backend is selected for heFFTe.

I will adjust the variable names as you suggested.

src/KSPACE/fft3d_wrap.cpp
```diff
@@ -27,30 +27,66 @@ FFT3d::FFT3d(LAMMPS *lmp, MPI_Comm comm, int nfast, int nmid, int nslow,
...
 
 /* ---------------------------------------------------------------------- */
 
 void FFT3d::compute(FFT_SCALAR *in, FFT_SCALAR *out, int flag)
 {
+  #ifndef HEFFTE
   fft_3d((FFT_DATA *) in,(FFT_DATA *) out,flag,plan);
+  #else
+  if (flag == 1)
+      heffte_plan->forward(reinterpret_cast<std::complex<FFT_SCALAR>*>(in),
```
mkstoyanov, line 72: @akohlmey here `in` and `out` are declared as pointers to `FFT_SCALAR` which is float or double.

When I cast to `std::complex<FFT_SCALAR>` the precision is preserved, i.e., we will have single or double complex pointers. At this point, heFFTe will pick up the data and perform the appropriate operations in single or double precision.

cmake/CMakeLists.txt
```diff
@@ -1026,6 +1026,11 @@ if(PKG_KSPACE)
   else()
     message(STATUS "Using non-threaded FFTs")
   endif()
+  if (FFT_HEFFTE)
+    message(STATUS "Using distributed algorithms from heFTTe")
```
stanmoore1, line 1030: Using distributed algorithms from heFTTe for FFTs

cmake/CMakeLists.txt
```diff
@@ -1026,6 +1026,11 @@ if(PKG_KSPACE)
...
+  else()
+    message(STATUS "Using builtin distributed algorithms")
```
stanmoore1, line 1032: Using builtin distributed algorithms for FFTs

doc/src/Build_settings.rst
```diff
@@ -76,6 +82,8 @@ LAMMPS can use them if they are available on your system.
          -D MKL_INCLUDE_DIR=path     # ditto for Intel MKL library
          -D FFT_MKL_THREADS=on       # enable using threaded FFTs with MKL libraries
          -D MKL_LIBRARY=path         # path to MKL libraries
+         -D Heffte_ROOT=path         # path to an existing heFFTe installation
+
```
stanmoore1, line 86: Need to document how to use Heffte with Traditional Make as well, i.e. add the flag below.

stanmoore1 (changes requested):
@mkstoyanov this looks close, just a few small change requests.

akohlmey:
@mkstoyanov this has been sitting idle for over three months now. Will there be updates?

mkstoyanov:
> @mkstoyanov this has been sitting idle for over three months now. Will there be updates?

I'm trying, but you guys threw a pretty big curve ball at me with the support for GNU Make. I'll give it another shot by the end of the month.

mkstoyanov:
I think I managed to pull the GNU make dependencies from CMake, but will have to do another PR.

PR #3649, opened 2023-02-13
Collected small fixes

akohlmey:
**Summary**

This pull request collects multiple small fixes and updates

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

- define a new constant MAXBIGINT_DOUBLE in the Variable class that will not overflow like MAXBIGINT when assigned to a double and cast back to a bigint
- add `link_check` target to `doc/Makefile` to automate checking links
- fix outdated, broken links and URL related typos in the manual
- improve error message for variable evaluation with `dump_modify every`
- add missing fallback URL function to CMake module for plugins
- remove debug printf() from ML-IAP support in KOKKOS
- apply clang-format to beck pair styles

**Post Submission Checklist**

### DIFF
--- a/cmake/Modules/LAMMPSInterfacePlugin.cmake
+++ b/cmake/Modules/LAMMPSInterfacePlugin.cmake
@@ -88,6 +88,18 @@ function(get_lammps_version version_header variable)
     set(${variable} "${date}" PARENT_SCOPE)
 endfunction()
 
+# determine canonical URL for downloading backup copy from download.lammps.org/thirdparty
+function(GetFallbackURL input output)
+  string(REPLACE "_URL" "" _tmp ${input})
+  string(TOLOWER ${_tmp} libname)
+  string(REGEX REPLACE "^https://.*/([^/]+gz)" "${LAMMPS_THIRDPARTY_URL}/${libname}-\\1" newurl "${${input}}")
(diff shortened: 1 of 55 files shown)


akohlmey:
@sjplimp @athomps I am a bit conflicted about the change to `lmptype.h`. On one hand it seems to be natural, to make this constant (the largest value of a signed 64-bit integer that can be stored in a double with "full" precision (since the lower 9 bits are truncated to zero)) available globally. On the other hand, it only applies to the variable code. Elsewhere 64-bit ints are usually encoded/decodes into doubles with `ubuf` where the truncation does not happen. `lmptype.h` is a bit "messy" and confusing and I would like to keep it as "clean" as possible since it is included almost everywhere.

sjplimp:
@akohlmey I think the small changes you added to lmptype.h are fine.  However I am confused by what MAXBIGINT-512 is useful for.  I am thinking a double precision value has ~9 bits less precision than a full 64 bit int.  So wouldn't you have to shift MAXBIGINT by 9 bits, not just subtract 512 ?

sjplimp (approved):

PR #3650, opened 2023-02-14
Implementation of analytical expressions for Born matrix in bond_class2 & bond_gromos

evoyiatzis:
**Summary**

Implementation of analytical first and second order derivatives of potential energy with respect to distance  for bond gromos and bond class 2

**Related Issue(s)**

No related issue

**Author(s)**

E. Voyiatzis - email evoyiatzis@gmail.com

**Licensing**

**Backward Compatibility**

No issue with backward compatibility

**Implementation Notes**

The results with these expressions match the ones obtained with the numerical differentiation using compute born/matrix with numdiff 

**Post Submission Checklist**

### DIFF
--- a/src/CLASS2/bond_class2.cpp
+++ b/src/CLASS2/bond_class2.cpp
@@ -32,7 +32,10 @@ using namespace LAMMPS_NS;
 
 /* ---------------------------------------------------------------------- */
 
-BondClass2::BondClass2(LAMMPS *lmp) : Bond(lmp) {}
+BondClass2::BondClass2(LAMMPS *lmp) : Bond(lmp)
+{
+  born_matrix_enable = 1;
+}
 
 /* ---------------------------------------------------------------------- */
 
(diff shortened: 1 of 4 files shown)


evoyiatzis:
@akohlmey I am not sure which unit tests to add for this feature. can you provide some help?

akohlmey:
> @akohlmey I am not sure which unit tests to add for this feature. can you provide some help?

There is currently no provision to support testing this facility in the general force-style unit tests. That would require either adding some tag to the YAML file format to explicitly enable additional checks and store the properties and some introspection functionality to determine if the tests could and should be run and the data stored.

That would be a non-trivial effort.

You could add some explicit test program that tests only the born matrix component and then has individual tests for all supported styles. See unittest/force-styles/test_pair_list.cpp or unittest/force-styles/test_mliappy_unified.cpp. In this case there would be a "test fixture" that creates/deletes a LAMMPS instance with standard settings and then a set of commands changing the bond or pair style and tests/checks using it. More like in e.g. unittest/commands/test_regions.cpp

athomps (approved):

PR #3652, opened 2023-02-15
Moved region check from constructor to init()

### LINKED ISSUE #3642: [BUG] Fix gcmc region extends outside simulation box when using triclinic region
**Summary**

fix gcmc command throws "fix gcmc region extends outside simulation" when using a triclinic box and a prism region that has any dimension larger than their box dimension minus their tilt.

Operating System: Red Hat Enterprise Linux
CPE OS Name: cpe:/o:redhat:enterprise_linux:7.9:GA:server
Kernel: Linux 3.10.0-1160.36.2.el7.x86_64
Architecture: x86-64

LAMMPS (23 Jun 2022 - Update 2) issue happening both on intel and gnu compiled versions

**Expected Behavior**

fix gcmc should run normally with the selected prism region as it does not physically go beyond the boundary of the box

**Actual Behavior**

ERROR: Fix gcmc region extends outside simulation box (src/MC/fix_gcmc.cpp:144)

**Steps to Reproduce**

When the user defines a triclinic box such as:

0.0      22.85  xlo xhi
0.0      13.22  ylo yhi
0.0      40.0  zlo zhi
10.08 0  0  xy xz yz

Define a subregion as region mc prism 0 EDGE 0 EDGE 25 32 10.08 0 0 which should restrict where atoms can be acted upon only along the z direction. With this setup any fix gcmc command using this region will not run, throwing "ERROR: Fix gcmc region extends outside simulation box (src/MC/fix_gcmc.cpp:144)". The biggest possible region is then: mc prism 0 $(22.85 - 10.08) 0 EDGE 25 32 10.08 0 0

Axel Kohlmeyer gave me a quick fix that is available at https://matsci.org/t/gcmc-region-triclinical-box-error/46594

Unfortunately, this fix seems to only work when LAMMPS is ran in serial mode. Using MPI will end up in either ERROR: Fix gcmc region extends outside simulation box (src/MC/fix_gcmc.cpp:**150**) or LAMMPS being frozen.

athomps:
**Summary**

The region check is moved from the constructor to init(), so that current box settings are used.

**Related Issue(s)**

This fixes #3642

**Author(s)**

Aidan Thompson

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

See above

**Post Submission Checklist**


stanmoore1 (commented):
Does this still work if the box dimension change on the fly, like with `fix deform` or `fix npt`?

athomps:
Yes, if the box changes on-the-fly, the region might extend outside the box. Is there a way to detect this, maybe a flag in class Domain?

stanmoore1:
>Is there a way to detect this, maybe a flag in class Domain?

I don't think so. When the box changes, the KSpace coefficients need to be recomputed by calling `kspace->setup()`. Currently this is all done by manual calls, but a callback or flag in `Domain` seems like a better solution, which could be used by other classes like GCMC as well.

```
grep "kspace->setup" *

fix_box_relax.cpp:  if (kspace_flag) force->kspace->setup();
fix_deform.cpp:  if (kspace_flag) force->kspace->setup();
fix_deform_kokkos.cpp:  if (kspace_flag) force->kspace->setup();
fix_nh.cpp:    if (kspace_flag) force->kspace->setup();
fix_nh.cpp:    force->kspace->setup();
fix_nh_kokkos.cpp:    if (kspace_flag) force->kspace->setup();
fix_press_berendsen.cpp:  if (kspace_flag) force->kspace->setup();
```

akohlmey:
> Yes, if the box changes on-the-fly, the region might extend outside the box. Is there a way to detect this, maybe a flag in class Domain?

Wouldn't it be better to take the intersection of the region and the box?
This is what my expectation would be and also what people would want if they want to insert particles within a given range in one dimension, but across the entire box in the other two or similar.

Worst case, we would have to add an interface that would allow fix gcmc to adjust the region when the box dimensions change.

athomps:
Yes, intersection would be nice, if there is a good *general* way to implement it.  The requirements on the region are:

1. ability to sample uniformly
2. a measure of volume, currently done once using 10000000 random points in bounding box, in constructor.

athomps:
I decided to keep this bugfix simple, so just moved region bounds check to init(), nothing else. I also updated LJ reference log output, because the reported atom counts were mysteriously too low by about 3 atoms, even though all other data is correct.

stanmoore1 (approved):

PR #3657, opened 2023-02-23
Add fix wall/table and fix wall/lepton for customizable wall fixes

akohlmey:
**Summary**

This adds two new wall styles: wall/table and wall/lepton that use tabulation or scripting expressions to implement custom wall potentials. Both are fully integrated into the fix wall framework. This also fixes a bug in fix wall/morse.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

The implementation of the two new wall styles follows the conventions for the corresponding bond styles.

**Post Submission Checklist**


akohlmey:
@athomps @sjplimp I found that adding Lepton support to code that computes forces can be as useful a tool for validating an implementation as using numerical differentiation, since Lepton can do symbolic differentiation and thus the forces are "exact". In this case, it was easy to validate the implementation against the existing wall potentials and also pair or bond styles, where available.
However, I ran into a strange situation with fix wall/morse where the forces differ more than what can be explained by numerics on the *second* MD step, but not on the first. Any ideas where the issue could be? Or where (else) to start looking? The code is valgrind clean. Please see the inputs added to the examples/wall folder.

athomps:
I took a close look at this, but could not figure it out. It seems no matter what I did, the the force is *always* correct on the first invocation, but significantly wrong after that.  The new wrong force still depends on the new r, and will be correct if r is the same as first call. This suggests stale intermediate values, either in Lepton or in LAMMPS.

athomps:
I suspect that the new wrong force depends on both the new value of r and the also the initial value of r, not on any other condtions, but I have not confirmed that.

akohlmey:
> I took a close look at this, but could not figure it out. It seems no matter what I did, the the force is _always_ correct on the first invocation, but significantly wrong after that. The new wrong force still depends on the new r, and will be correct if r is the same as first call. This suggests stale intermediate values, either in Lepton or in LAMMPS.

I am quite confident that fix wall/table and fix wall/lepton are correct, because they a) reproduce the results for fix wall/harmonic and fix wall/lj126 (have not tried the others) and b) those also agree with the forces from bond style harmonic, and morse and pair style harmonic/cut, lj/cut and morse. It seems fix wall/morse is the odd one out. I suspect some uninitialized or not properly zeroed variable, but I cannot find it.

akohlmey:
> I suspect that the new wrong force depends on both the new value of r and the also the initial value of r, not on any other condtions, but I have not confirmed that.

Found the bug.
The wall force computation was incorrectly dividing by "delta". The reason it didn't give wrong forces in the first step originates in the geometry that I had set up where delta had the value of 1 at the first step and thus the forces was correct. But on the second step, after the atoms had moved, the delta was != 1 and then it led to wrong forces.

doc/src/fix_wall.rst
```diff
@@ -266,20 +360,19 @@ define the lattice spacings.
...
+manner.
+
+The *pbc* keyword can be used with a *yes* setting to allow walls to be
+specified in a periodic dimension.  See the :doc:`boundary <boundary>`
+command for options on simulation box boundaries.  The default for *pbc*
+is *no*, which means the system must be non-periodic when using a wall.
+But you may wish to use a periodic box.  E.g. to allow some particles to
+interact with the wall via the fix group-ID, and others to pass through
+it and wrap around a periodic box.  In this case you should ensure that
+the wall if sufficiently far enough away from the box boundary.  If you
```
stanmoore1, line 374: typo: if --> is

stanmoore1 (approved):
I approve, found a typo in the docs though

doc/src/fix_wall.rst
akohlmey: This must have been there for some time... fixing it now in #3663

PR #3662, opened 2023-02-25
New fix command for alchemical transformations using multiple partitions

akohlmey:
**Summary**

This pull request adds a `fix alchemy` command that allows to run an "alchemical transformation" simulation where the system changes during an MD run gradually from one topology to another. The forces used for propagating atoms are computed as a scaled mix from the two partitions.  Because of implementing this with a multi-partition approach, it can be applied to situations where commands like `fix adapt` cannot be used.  To support variable cell calculations a `compute pressure/alchemy` command is included that will return the "mixed system" pressure.

**Related Issue(s)**

N/A.

**Author(s)**

Axel Kohlmeyer, Temple U.

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

The implementation assumes that the initial systems are identical and thus have the same number and positions of atoms and the same box dimensions and domain decompositions.  During the run, the forces are mixed according to the lambda parameter thus trajectories for both partitions should be the same. To make certain, atom positions and box dimensions are synchronized.

**Post Submission Checklist**


stanmoore1:
Can this command turn lead into gold?

akohlmey:
> Can this command turn lead into gold?

Sure thing. You just need to feed it the corresponding potential files. :smiling_imp: 
There is already an example converting copper into bronze.

athomps:
This is a nice feature.  If I understand correctly, it provides functionality similar to `fix adapt[/fep]` but completely general, does not require access to energy model parameters. It also is similar to 'pair hybrid/scaled', but implementation is at the level of partitions, so much more general.  The hard-coding of `\lambda_p = ramp(0,1)` seems unnecessarily restrictive.  People might want to use this with constant `\lambda_p` or non-linear variation in time. Could this be supported in the same way as in `fix adapt` ?

akohlmey:
> Could this be supported in the same way as in `fix adapt` ?

Sure. I've been trying to keep it simple, but switching to require the lambda value provided through an equal or python style variable as an argument should be very straightforward to do. I'll have to look at fix adapt to see how similar this can be made.

akohlmey:
@athomps commit 446913f2f347083c4394bd613382dad883195784 implements the ramping through a variable.

Technically, this is could now also be extended to support more than 2 partitions and do an even more complex mixing of topology, e.g. like in EVB models.

athomps (approved):

PR #3665, opened 2023-02-28
Possibly fix a typo (or be shown a misunderstanding)

GenieTim:
Fix issue where brownian motion with uniform distribution had overpowered random noise

**Summary**

As I understand it, the Brownian motion with a uniform random generator had its random noise overpowered by a factor of sqrt(12):

https://github.com/lammps/lammps/blob/a67d82c183857531a70af3397848fddbcd4b8264/src/BROWNIAN/fix_brownian_base.cpp#L220-L223

This PR either fixes this probably-typo, or kindly requests an explanation of where that factor is coming from.

**Author(s)**

Tim Bernhard, tim@bernhard.site

**Licensing**

**Backward Compatibility**

Well, all the inputs will still run, but the resulting values definitely change.

**Post Submission Checklist**

### DIFF
--- a/doc/src/fix_press_berendsen.rst
+++ b/doc/src/fix_press_berendsen.rst
@@ -61,7 +61,7 @@ unchanged and controlling the pressure of a surrounding fluid.
    atoms.  This fix can be used in conjunction with thermostatting fixes
    to control the temperature, such as :doc:`fix nvt <fix_nh>` or :doc:`fix langevin <fix_langevin>` or :doc:`fix temp/berendsen <fix_temp_berendsen>`.
 
-See the :doc:`Howto baroostat <Howto_barostat>` page for a
+See the :doc:`Howto barostat <Howto_barostat>` page for a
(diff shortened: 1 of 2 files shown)


GenieTim:
Maybe @samueljmcameron, as this was originally your commit, would you mind explaining to me why there was a factor of 24 rather than 2 in the noise term of the uniform random distribution? Thanks!

GenieTim:
Ok, I think I finally understood: the 24 is composed in this way:
say we want each sample to be from a distribution with variance sigma. Therefore, we need to sample between ±sqrt(3) times sigma (where the sqrt(3) is coming from the need to get a unit variance from a uniform distribution). Mapping this to a random real between ±0.5, we have to multiply this by sigma*2*sqrt(3). Taking the 2 into the sqrt leads to sqrt(12), then adding the 2 that is promised in the equations, we are at the factor of 24. 
Sorry for having been confused.

PR #3666, opened 2023-02-28
Fixed GPU library builds with CUDA MPS

ndtrung81:
**Summary**

This PR updated GPU.cmake and Makefile.cuda_mps to not error out with exclusive process  mode with the CUDA MPS server in use. Fixed a typo in GPU.cmake: `$(CUDA_PROXY)` is replaced by `$(CUDA_MPS)`.

**Related Issue(s)**

N/A

**Author(s)**

Trung Nguyen (U Chicago)

**Licensing**

**Backward Compatibility**

Should maintain backward compability.

**Implementation Notes**

With the CUDA MPS server on, the GPU is switched to exclusive process mode, not allowing multiple processes sharing. The GPU lib source files (on the host side) then needs to be built with -DCUDA_PROXY to avoid generating errors. To be consistent with lib/gpu/README, this PR makes -DCUDA_MPS_SUPPORT and -DCUDA_PROXY both defined by CUDA_MPS (for traditional builds with Makefile.cuda_mps) and when CUDA_MPS_SUPPORT=on (for CMake builds).

Alternatively, a simpler approach may be to remove the check for DCUDA_MPS_SUPPORT in `lal_device.cpp:280`
```
#if !defined(CUDA_PROXY)
  if (_procs_per_gpu>1 && !gpu->sharing_supported(my_gpu))
    return -7;
#endif
```
and remove -DCUDA_MPS_SUPPORT at the above places in GPU.cmake and Makefile.cuda_mps. And update lib/gpu/README accordingly because CUDA_PROXY and CUDA_MPS_SUPPORT are always used together. (CUDA_PROXY is the ancestor of CUDA MPS, back to the time of Tesla Kepler GPUs, if I recall correctly.)

**Post Submission Checklist**

- [] The source code follows the LAMMPS formatting guidelines


stanmoore1 (approved):

akohlmey:
@ndtrung81 If CUDA_PROXY and CUDA_MPS_SUPPORT are both essentially the same thing, then why keep the outdated CUDA_PROXY and remove the newer CUDA_MPS_SUPPORT?

ndtrung81:
@akohlmey thanks for pointing this out. It makes more sense to drop the outdated CUDA_PROXY and keep the newer CUDA_MPS_SUPPORT. I will make the changes.

stanmoore1 (approved):

PR #3670, opened 2023-03-03
Update Kokkos library in LAMMPS to v4.1.0

stanmoore1:
**Summary**

Update Kokkos library in LAMMPS to v4.1.0
NOTE: Requires C++17

**Related Issue(s)**

None

**Author(s)**

Stan Moore (SNL), Kokkos developers

**Licensing**

**Backward Compatibility**

Yes


stanmoore1:
Note: here is a list of minimum compiler versions for Kokkos 4.0: https://github.com/kokkos/kokkos/issues/5285#issue-1320161352

stanmoore1:
Kokkos regression tests passed. Need to check performance.

rbberger (approved):
The CMake changes look fine to me.

cmake/CMakeLists.txt
```diff
@@ -136,15 +136,15 @@ if((PKG_KOKKOS) AND (Kokkos_ENABLE_CUDA) AND NOT (CMAKE_CXX_COMPILER_ID STREQUAL
...
 if(NOT CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 11)
 endif()
 if(CMAKE_CXX_STANDARD LESS 11)
   message(FATAL_ERROR "C++ standard must be set to at least 11")
 endif()
-if(PKG_KOKKOS AND (CMAKE_CXX_STANDARD LESS 14))
-  set(CMAKE_CXX_STANDARD 14)
+if(PKG_KOKKOS AND (CMAKE_CXX_STANDARD LESS 17))
+  set(CMAKE_CXX_STANDARD 17)
```
junghans, line 147: I think it would be better to do this on per-target basis with `set_target_properties()`, just in case somebody would like to compile lammps with C++20 or newer.

akohlmey: We've had something like that before, but that led to all kinds of unexpected inconsistencies.
If you want C++-20 you have to use it for "everything" by setting `-DCMAKE_CXX_STANDARD=20`.

junghans: see https://stackoverflow.com/questions/70566491/how-to-bump-the-c-standard-from-the-cmake-command-line

also Kokkos developer @dalg24 had some strong opinions on that.

dalg24: > If you want C++-20 you have to use it for "everything" by setting `-DCMAKE_CXX_STANDARD=20`.

On board with that approach.

Only unfortunate thing I see with the snippet above is if one explicitly configure with `-DCMAKE_CXX_STANDARD=11` or `14` it would silently be changed to `17` without even raising a warning.
You could do something like
```CMake
if(NOT CMAKE_CXX_STANDARD)
  if(PKG_KOKKOS)
    set(CMAKE_CXX_STANDARD 17)
  else()
    set(CMAKE_CXX_STANDARD 11)
  endif()
endif()

if(PKG_KOKKOS)
  if(CMAKE_CXX_STANDARD LESS 17)
    message(FATAL_ERROR "C++ standard must be set to at least 17")
  endif()
else()
  if(CMAKE_CXX_STANDARD LESS 11)
    message(FATAL_ERROR "C++ standard must be set to at least 11")
  endif()
endif()
```

akohlmey: That stackoverflow issue is setting the C++ standard unconditionally. The quoted cmake script code from LAMMPS does not. It will *raise* the C++ standard to 14 (old) or 17 (new) is it has not been set or set to a lower value.

akohlmey: @dalg24 True, but for now I am very happy with a solution that requires to write and maintain only 2 lines and wait until somebody runs into a situation where it creates a problem before considering a more complex approach.

junghans (approved):

opoplawski:
Any progress here?  Fedora Rawhide has updated kokkos to 4.0 and now the lammps builds are failing.  A new release would be very helpful.  Thanks.

stanmoore1:
@opoplawski I think #3701 will address the issue you are seeing, can you confirm?

stanmoore1:
@akohlmey it looks like the run test failure is unrelated to this PR?

stanmoore1:
Since this PR requires c++17, we are planning to wait until after the next stable release to merge this, to maximize backwards compatibility. Deprecated code was removed in #3701, which should allow LAMMPS to be compatible with an external Kokkos v4.0.

akohlmey:
> @akohlmey it looks like the run test failure is unrelated to this PR?

That is correct. It is a ripple from the removal of installing the python module from "make install". The run and regression tests depend on it, and the replacement has a race condition when running "make install-python" concurrently from the same source checkout. It took me a while to narrow down where the exact issue is since there were multiple problems that needed changes in multiple places.

skyreflectedinmirrors:
FWIW Stan, I still see a bunch of hash map errors:

```
lammps/src/KOKKOS/neigh_bond_kokkos.cpp:284:89: error: reference to __host__ function 'dual_hash_type' in __host__ __device__ function
    int atom1 = AtomKokkos::map_kokkos<DeviceType>(bond_atom(i,m),map_style,k_map_array,k_map_hash);
                                                                                        ^
lammps/src/KOKKOS/neigh_bond_kokkos.cpp:1230:16: note: in instantiation of member function 'LAMMPS_NS::NeighBondKokkos<Kokkos::HIP>::operator()' requested here
template class NeighBondKokkos<LMPDeviceType>;
               ^
lammps/src/KOKKOS/kokkos_type.h:511:8: note: 'dual_hash_type' declared here
struct dual_hash_type {
       ^
lammps/src/KOKKOS/neigh_bond_kokkos.cpp:372:89: error: reference to __host__ function 'dual_hash_type' in __host__ __device__ function
    int atom1 = AtomKokkos::map_kokkos<DeviceType>(bond_atom(i,m),map_style,k_map_array,k_map_hash);
```

when building w/ your branch + Kokkos develop.  I _think_ it's because there's still an implicit constructor in the dual_hash_type:

image

that isn't deleted (and therefore defaults to `__host__`)?

This might be something specific to develop however.

stanmoore1:
@arghdos thanks will take a look.

stanmoore1:
@arghdos I can reproduce the compile errors with Kokkos `develop`, does not seem to apply to 4.0.

rbberger:
@stanmoore1 created https://github.com/stanmoore1/lammps/pull/16 with some cleanups and a HIP bugfix, which would merge into this PR branch.

skyreflectedinmirrors:
Can confirm, this builds w/ the latest kokkos develop on MI-2XX

stanmoore1:
@akohlmey would like to get this merged ASAP now that the stable is out.

akohlmey:
> @akohlmey would like to get this merged ASAP now that the stable is out.

Yup. I am working through the back log right now.

PR #3671, opened 2023-03-03
Add new bond style harmonic/restrain

akohlmey:
**Summary**

This implements a harmonic bond potential that does not have a single equilibrium distance per bond type, but uses the initial distance of the atoms. Thus it is more like a restraining potential, but unlike fix restrain, it uses explicit bonds.

**Related Issue(s)**

This implementation was motivated by a [discussion on the MatSci forum](https://matsci.org/t/segmentation-fault-while-reading-atoms-x0-value/47240)

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

Rather than storing the initial bond lengths, the initial atom positions are stored via an internal instance of fix property atom. This has the advantage of requiring minimum effort to implement and avoiding to write a new atom style.

**Post Submission Checklist**


akohlmey:
@sjplimp This is now back to where it is supposed to be. I have made some more tests and found a few more kinks that needed straightening out. I'm attaching a test input. If you run with `-v bond fixed`  it will use bond style harmonic, with `-v bond restrain` it will use bond style harmonic/restrain and with `-v bond hybrid` it will use both in a hybrid configuration. Since all bonds are of length 1 \AA (within the precision of the molecule) the initial energies, forces, and pressure should be very close for all 3 cases.

[in.spce.txt](https://github.com/lammps/lammps/files/10915555/in.spce.txt)
[spce.txt](https://github.com/lammps/lammps/files/10915556/spce.txt)

akohlmey:
@sjplimp This is now waiting on you to update fix STORE/PERATOM as we had discussed elsewhere. I have thus changed the assignment. Please assign the PR it back to me when you are done and need me to complete it.

sjplimp:
@akohlmey thanks - should get to this, this afternoon

athomps (approved):

akohlmey:
@sjplimp fix STORE/PERATOM has been renamed to fix STORE/ATOM. I've updated the docs and added the notes we discussed. All tests pass on my machine.

sjplimp (approved):

PR #3674, opened 2023-03-06
GPU package bugfix / cmake improvement

rbberger:
**Summary**

* detect hipcub via CMake, and take advantage of the `hip::host` and `hip::hipcub` targets. we don't want to assume where header include path is.
* fixup broken HIP code path in GPU package. `num_atoms` was not defined in case of `USE_HIP_DEVICE_SORT`.
* drop `hcc` as HIP_PLATFORM, any modern ROCm install will use `amd` (only 5.1+ are worth looking at).

**Related Issue(s)**

**Author(s)**

@rbberger 

**Licensing**

**Backward Compatibility**

**Implementation Notes**

Bug was introduced by 4244d2e

**Post Submission Checklist**


cmake/Modules/Packages/GPU.cmake
```diff
@@ -461,30 +462,22 @@ elseif(GPU_API STREQUAL "HIP")
...
     target_include_directories(hip_get_devices PRIVATE ${CUDA_INCLUDE_DIRS})
     target_link_libraries(hip_get_devices PRIVATE ${CUDA_LIBRARIES} ${CUDA_CUDA_LIBRARY})
   elseif(HIP_PLATFORM STREQUAL "hcc")
     target_compile_definitions(gpu PRIVATE -D__HIP_PLATFORM_HCC__)
-    target_include_directories(gpu PRIVATE ${HIP_ROOT_DIR}/../include)
-
     target_compile_definitions(hip_get_devices PRIVATE -D__HIP_PLATFORM_HCC__)
-    target_include_directories(hip_get_devices PRIVATE ${HIP_ROOT_DIR}/../include)
   elseif(HIP_PLATFORM STREQUAL "amd")
     target_compile_definitions(gpu PRIVATE -D__HIP_PLATFORM_AMD__)
```
skyreflectedinmirrors: I believe these could be inherited from hip::host, and then we could get rid of the entire hcc vs amd branch (if we even care about hcc anymore? I think HIP_PLATFORM has been 'amd' since ~4.1ish?

skyreflectedinmirrors:
LGTM, might be able to simplify further by inheriting compile options from hip::host

cmake/Modules/Packages/GPU.cmake
rbberger: True, we should only care about 5.1+.

rbberger:
> LGTM, might be able to simplify further by inheriting compile options from hip::host

You're right. (Quite the challenge to find the right repo where this file comes from): https://github.com/ROCm-Developer-Tools/hipamd/blob/develop/hip-config.cmake.in#L180.

skyreflectedinmirrors:
Agreed -- quick reference is:

- hipamd -> the AMD backend of HIP, also has forwarding for NV backend
- HIP -> headers, that theoretically multiple backends could implement

HIP owns the HIP-Language cmake: https://github.com/ROCm-Developer-Tools/HIP/blob/develop/hip-lang-config.cmake.in#L81, which is what gets turned off if you used (e.g.,) `enable_language(HIP)`  for CMake >= 3.21(and/or setting the language)

stanmoore1 (approved):

PR #3677, opened 2023-03-06
Sort option for load balancing commands

sjplimp:
**Summary**

Add a sort yes/no option to the balance and fix balance commands.  For rcb balancing this controls whether the multiple messages per processor that exchange per-atom data are sorted or not.  Sorting is slower (than unsorted = random).
But sorting insures the final ordering of atoms on each processor is consistent when the same simulation is run multiple times.
This can be useful when debugging.  The default is to sort for the balance command (one-time operation), and to not sort for
the fix balance command, since it may be invoked frequently during a run.

**Related Issue(s)**

This will be a useful option to invoke for the new fix alchemy command, which requires consistent ordering of atoms on two different replicas.

**Author(s)**

Steve

**Licensing**

**Backward Compatibility**

Should be backward compatible, since the sorting default for the 2 commands is unchanged.

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/doc/src/balance.rst
+++ b/doc/src/balance.rst
@@ -53,6 +53,7 @@ Syntax
              name = name of the atom-style variable
            *store* name = store weight in custom atom property defined by :doc:`fix property/atom <fix_property_atom>` command
              name = atom property name (without d\_ prefix)
+       *sort* arg = *no* or *yes*
        *out* arg = filename
          filename = write each processor's subdomain to a file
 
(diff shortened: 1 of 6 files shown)


akohlmey:
@sjplimp The backward compatibility note says that options are unchanged, but the description of the new flag says that the balance command will default to sorting while fix balance will not. This seems to contradict the compatibility note. Can you please clarify?

stanmoore1 (approved):
I approve

Issue #3678, opened 2023-03-07
[Feature Request] Always use the -DLAMMPS_EXCEPTIONS branch and remove the non-exception code path


akohlmey:
**Summary**

Using the exception handling in LAMMPS has matured quite a bit over time. It is used by default by most precompiled binary packages and there have been no exception specific bug reports for quite some time.

While the major shortcoming - there is no mechanism to propagate exceptions across MPI - still exists, the number of benefits from enabling exceptions is growing.
The python module and LAMMPS shell benefit from it massively, also unit testing and auxiliary libraries are increasingly making use of exceptions, so adding a general catch-all handler is a good idea, too.

**Detailed Description**

This would mostly mean removal of the non-exception case of alternate code paths.
This would also allow to simplify some of the code.

dxhisboy:
I think another thing to consider is how it will affect offloaded computation workloads, for example, CUDA, OpenCL or KOKKOS kernels.

> While the major shortcoming - there is no mechanism to propagate exceptions across MPI - still exists, the number of benefits from enabling exceptions is growing.

I think this can be handled by a catch in the `main` function that calls `MPI_Abort` to propagate across MPI.

akohlmey:
> I think another thing to consider is how it will affect offloaded computation workloads, for example, CUDA, OpenCL or KOKKOS kernels.

It has been possible for a long time to compile with -DLAMMPS_EXCEPTIONS and check it out. I am not aware of any issues. If you have concerns, it is easy for you to test.

> I think this can be handled by a catch in the `main` function that calls `MPI_Abort` to propagate across MPI.

This is what is already done (and has been for quite a while).

dxhisboy:
> It has been possible for a long time to compile with -DLAMMPS_EXCEPTIONS and check it out. I am not aware of any issues. If you have concerns, it is easy for you to test.

AFAIK Sunway's computation cores cannot support exceptions mechanism. Although it is never supported officially, I'm working on porting specific potentials to Sunway's computation cores since 2017. Removing `-DLAMMPS_EXCEPTIONS` is affordable, that can be translated to a special function call via compiler plugins in my opinion.

> This is what is already done (and has been for quite a while).

Sorry for disturbing. But it has been a long time from last time I peek `main.cpp`.

akohlmey:
> AFAIK Sunway's computation cores cannot support exceptions mechanism. Although it is never supported officially. I'm working for a long time on it.

[...]

PR #3680, opened 2023-03-07
Address double linking issue with Kokkos libraries

akohlmey:
**Summary**

Address some issues caused by linking Kokkos libraries twice with the LAMMPS main program.

**Related Issue(s)**

https://github.com/kokkos/kokkos/issues/5596

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

This uses the library interface wrapper functions instead of the direct calls to finalize the Kokkos and Python environments. With that change we no longer need to pull in the corresponding headers and thus can drop the direct dependency of the LAMMPS executable on the Kokkos libraries. This also contains some changes to avoid linker errors due flags added to suppress warnings from nvcc during compilation of the LAMMPS library.

**Post Submission Checklist**


cmake/CMakeLists.txt
```diff
@@ -538,7 +538,10 @@ set(CMAKE_TUNE_FLAGS "${CMAKE_TUNE_DEFAULT}" CACHE STRING "Compiler and machine
 separate_arguments(CMAKE_TUNE_FLAGS)
 foreach(_FLAG ${CMAKE_TUNE_FLAGS})
   target_compile_options(lammps PRIVATE ${_FLAG})
-  target_compile_options(lmp PRIVATE ${_FLAG})
+  # skip these flags when linking the main executable
```
junghans, line 541: Can you comment why these flags need to be skipped?

cmake/Modules/Packages/KOKKOS.cmake
```diff
@@ -98,7 +96,6 @@ else()
                           ${LAMMPS_LIB_KOKKOS_BIN_DIR})
   target_include_directories(lammps PRIVATE ${Kokkos_INCLUDE_DIRS})
```
junghans, line 97: @dalg24, doesn't  the`kokkos` target export the above include directories already?

src/main.cpp
```diff
@@ -35,14 +34,20 @@
 
 using namespace LAMMPS_NS;
 
+// for convenience
+static void finalize()
```
junghans, line 38: Couldn't we make that part of the library interface?

src/main.cpp
```diff
@@ -99,14 +106,12 @@ int main(int argc, char **argv)
...
+    finalize();
     MPI_Abort(MPI_COMM_WORLD, 1);
     exit(1);
   }
 #endif
-  KokkosLMP::finalize();
-  Python::finalize();
+  finalize();
   MPI_Barrier(lammps_comm);
   MPI_Finalize();
```
junghans, line 116: Couldn't we use `lammps_mpi_finalize()` here?

cmake/CMakeLists.txt
akohlmey: Because they cause errors on multiple of my development machines and the flags were only added to suppress warnings when compiling non-kokkos non-GPU code through nvcc with nvcc_wrapper.

src/main.cpp
akohlmey: why? we already have the individual calls. I almost didn't do this one since the benefits are miniscule.

src/main.cpp
akohlmey: We have to include the MPI header anyway and do other explicit MPI calls so why use a wrapper?

Also the wrapper has MPI_COMM_WORLD hard coded, but with the main function, we may be only running on a subset in case LAMMPS is launched using the MDI coupling interface.

cmake/Modules/Packages/KOKKOS.cmake
dalg24: It does.  All is needed is  `target_link_libraries(... Kokkos::kokkos)`

akohlmey: I won't make changes here now. This has been working for a long time. Never change a winning team. :wink:

If all goes according to plan, we will raise the minimum CMake version required for LAMMPS after the next stable release to 3.16 and that will allow some significant refactor and simplification of CMake scripting. That would be a point to review this kind of change, too.

LAMMPS has to compile and work on a lot of quirky machines and Kokkos support is complex and rather fragile as it is.

dalg24: Won't argue with that :)

cmake/CMakeLists.txt
junghans: Makes sense, thanks for the explanation! It would be good to mention that in the comments as well then.

cmake/Modules/Packages/KOKKOS.cmake
junghans: Sounds good, let's save that for the CMake refactor then.

junghans: Sounds good, let's save that for the CMake refactor then.

src/main.cpp
junghans: Makes sense, `MPI_Init()` is explicit, too!

src/main.cpp
junghans: It would feel more natural to me if there would be a function that finalizes all the things that lammps initializes automatically.

src/main.cpp
```diff
@@ -99,14 +106,12 @@ int main(int argc, char **argv)
     delete lammps;
   } catch (fmt::format_error &fe) {
     fprintf(stderr, "fmt::format_error: %s\n", fe.what());
-    KokkosLMP::finalize();
-    Python::finalize();
+    finalize();
     MPI_Abort(MPI_COMM_WORLD, 1);
     exit(1);
   }
 #endif
```
junghans, line 113: Because there is now a try catch block even for the case without `LAMMPS_EXCEPTIONS` couldn't we write the whole block as:
```cpp
  try {
    auto lammps = new LAMMPS(argc, argv, lammps_comm);
    lammps->input->file();
    delete lammps;
#ifdef LAMMPS_EXCEPTIONS
  } catch (LAMMPSAbortException &ae) {
    finalize();
    MPI_Abort(ae.universe, 1);
  } catch (LAMMPSException &) {
    finalize();
    MPI_Barrier(lammps_comm);
    MPI_Finalize();
    exit(1);
#endif
  } catch (fmt::format_error &fe) {
    fprintf(stderr, "fmt::format_error: %s\n", fe.what());
    finalize();
    MPI_Abort(MPI_COMM_WORLD, 1);
    exit(1);
  } catch (std::exception &e) {
    fprintf(stderr, "Exception: %s\n", e.what());
    finalize();
    MPI_Abort(MPI_COMM_WORLD, 1);
    exit(1);
  }
```
feel a bit more natural, saves some code duplication and catches exception from any other third party library.

But feel free to ignore.

src/main.cpp
```diff
@@ -13,9 +13,8 @@
 
 #include "lammps.h"
 
-#include "accelerator_kokkos.h"
 #include "input.h"
```
junghans, line 16: Are we still using `input.h` explicitly?

src/main.cpp
akohlmey: These functions are primarily available for internal use so this is not much of a concern. Those have grown organically. As mentioned elsewhere in this PR, I don't want to rock the boat too much at the current state of development when we are slowly converging toward a stable release. These kinds of changes are better done at the beginning of a development cycle.

src/main.cpp
akohlmey: I have a more ambitious plan, please see issue #3678. But this is something for after the next stable release.

src/main.cpp
akohlmey: Yes, it is required. The functionality needed is not available from the library interface and also the behavior of the library interface with exceptions enabled is different and thus would require a more significant rewrite of main.cpp than I currently want to do.

src/main.cpp
junghans: Ok

junghans (approved):

stanmoore1 (approved):
I confirm this fixes the crash in https://github.com/kokkos/kokkos/issues/5596.

stanmoore1:
@bathmatt this fixes the issue you saw

PR #3684, opened 2023-03-10
Encode improper symmetries

jrgissing:
**Summary**

Encode 'symmetry atoms' of improper styles, as described by the docs. This feature will be used to infer improper types from a set of four atoms (example of such a function is attached). Submitting this PR first and separately because significant review of the various impropers will be needed to ensure symmetry atoms are identified correctly. Symmetry atoms are defined as those that cannot be interchanged with others, whereas non-symmetry atoms are interchangable. For example, for a simple out-of-plane improper where atom J is out of the IKL plane, the symmetry atom is the second atom J.

This PR is related to the type labels PR to enable some of the envisioned functionality of type labels.

**Related Issue(s)**

eventually to be utilized by 'fix bond/react' and OpenKIM, for example

**Author(s)**

JG

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

example of a function to infer improper type:

[infer_improper_example.txt](https://github.com/lammps/lammps/files/10938177/infer_improper_example.txt)

### DIFF
--- a/src/AMOEBA/improper_amoeba.cpp
+++ b/src/AMOEBA/improper_amoeba.cpp
@@ -36,6 +36,10 @@ using namespace MathConst;
 ImproperAmoeba::ImproperAmoeba(LAMMPS *lmp) : Improper(lmp)
 {
   writedata = 1;
+
+  // the second atom in the quadruplet is the atom of symmetry
+
+  symmatoms[1] = 1;
 }
 
 /* ---------------------------------------------------------------------- */
--- a/src/CLASS2/improper_class2.cpp
+++ b/src/CLASS2/improper_class2.cpp
@@ -39,6 +39,10 @@ using namespace MathConst;
 ImproperClass2::ImproperClass2(LAMMPS *lmp) : Improper(lmp)
 {
   writedata = 1;
+
+  // the second atom in the quadruplet is the atom of symmetry
+
+  symmatoms[1] = 1;
 }
 
 /* ---------------------------------------------------------------------- */
--- a/src/EXTRA-MOLECULE/improper_cossq.cpp
+++ b/src/EXTRA-MOLECULE/improper_cossq.cpp
@@ -37,7 +37,12 @@ using namespace MathConst;
 
 /* ---------------------------------------------------------------------- */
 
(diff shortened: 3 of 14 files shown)


jrgissing:
@tadmor @yafshar @ellio167 here is a PR encoding improper symmetries. please let me know if it looks okay, and if you have any other requests from an OpenKIM perspective

stanmoore1 (approved):

yafshar:
@jrgissing the PR sounds great to me. Thanks

sjplimp:
@jrgissing Jake - are you saying you are confident you have labeled the symmetry for each improper style correctly, or that you want others to check this?

sjplimp:
@jrgissing Also, this is only adding internal bookkeeping which does not affect the calculation of any of the improper interactions themselves?  Hence, a symmetry labeling error could be identified later?

jrgissing:
> @jrgissing Jake - are you saying you are confident you have labeled the symmetry for each improper style correctly, or that you want others to check this?

I have double-checked this PR myself, but of course any review would be welcome because impropers can be tricky (see below). It should be noted that I am basing my assignment of symmetry atoms (as defined in the PR summary above) on the docs and not consulting the source code.

> @jrgissing Also, this is only adding internal bookkeeping which does not affect the calculation of any of the improper interactions themselves? Hence, a symmetry labeling error could be identified later?
 
Yes, this PR is just bookkeeping, but it will affect the correct assignment of impropers for commands that utilize it. This correctness will also depend on the implementation of the 'infer the improper type' routine (i.e., given four type-labeled atoms and their connectivity, correctly assign an improper type). I should probably include that routine in this PR to be more self-contained. For the record, here are some potential pitfalls:

1) For impropers with multiple symmetry atoms (i.e., non-interchangeable atoms), it is likely necessary/safer to also define the central atom (the one bonded to the other three). The first symmetry atom happens to be the central atom in all currently affected impropers (improper_style fourier and umbrella). 

2) Several impropers are chiral (improper_style ring, distharm). For the relatively rare case where this could be relevant, such as nonzero out-of-plane angles, in general the original chirality of the newly assigned improper would not be checked for. However, it will be consistent with the parameters of the improper once assigned, because the order of atom types in the improper type label is not being changed.

sjplimp (approved):

PR #3685, opened 2023-03-12
Add TIP4P versions for compute dipole, compute dipole/chunk and fix efield

akohlmey:
**Summary**

This pull requests adds variants for TIP4P pair styles to compute dipole moments and to apply and external electric field.  The regular versions produce incorrect results because the charge of the virtual site M of the TIP4P geometry is assigned to (uncharged) oxygen atom.  The include styles compute the location of the virtual site and use that.

**Related Issue(s)**

This addresses issues discussed on the MatSci forum in:
https://matsci.org/t/compute-dipole-for-tip4p-water-models/47260
and:
https://matsci.org/t/tip4p-2005-under-electric-field/47349

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

The new styles use the `Pair::extract()` functionality to obtain the TIP4P parameters, so a TIP4P pair style must be used. Support for that needed to be added to a few pair styles with cutoff.

All functionality was tested against using equivalent geometries with explicit four site water geometries including cases where water molecules straddle subdomain boundaries to produce identical output (total dipole, per-molecule dipoles, forces, per-molecule torques) up to precision of the geometries in the molecule files used.

Since fix efield/tip4p is a derived class, some error messages in the base class had to be adjusted.

**Post Submission Checklist**


rbberger:
@akohlmey can you please mark these new TIP4P tests and all other existing ones as unstable, given that they usually break on ppc64le and arm64 on a regular basis? I was about to make a PR asking for this, but given that you're adding more here it would fit in.

akohlmey:
> @akohlmey can you please mark these new TIP4P tests and all other existing ones as unstable, given that they usually break on ppc64le and arm64 on a regular basis? I was about to make a PR asking for this, but given that you're adding more here it would fit in.

Do you want me to tag "everything TIP4P"? If I remember correctly, the tests that fail are pair styles only and and are failing only in multi-thread tests with OPENMP, or is my recollection wrong?

rbberger:
> Do you want me to tag "everything TIP4P"? If I remember correctly, the tests that fail are pair styles only and and are failing only in multi-thread tests with OPENMP, or is my recollection wrong?

I know it's a radical ask, but I've seen this show up more and more often. Here is my latest exclusion list (there's more, but these are the TIP4P ones), that would fit into the `unstable` category.
```
MolPairStyle:tip4p_long
MolPairStyle:tip4p_table
MolPairStyle:tip4p_cut
MolPairStyle:lj_cut_tip4p_table
MolPairStyle:lj_table_tip4p_long
MolPairStyle:lj_cut_tip4p_long
MolPairStyle:lj_cut_tip4p_cut
MolPairStyle:lj_cut_tip4p_long_soft
```

rbberger:
Thank you.

stanmoore1 (approved):
This looks OK to me. Is there any error check to prevent using TIP4P with the non-TIP4P versions of efield, etc., which would give the wrong numerics?

akohlmey:
> Is there any error check to prevent using TIP4P with the non-TIP4P versions of efield, etc., which would give the wrong numerics?

No. An error would be a bit drastic, since you can apply fix efield to a group that does not include TIP4P atoms while a TIP4P pair style is used and a detailed check for that quite convoluted.
A warning is more appropriate, e.g. what we do when we have explicit bonds with a many-body potential or similar.

PR #3687, opened 2023-03-14

### LINKED ISSUE #3690: [BUG] fix mscg example segfaults
**Summary**

The MSCG example input deck in `examples/mscg` produces a segmentation fault in the force matching part of the run.

8 Feb 2023

**Expected Behavior**

Run completes

**Actual Behavior**

Segmentation fault in second part of the run.

**Steps to Reproduce**

Compile LAMMPS with MSCG package included, run example in serial.

Example log was updated last in 2017. There is a difference in the force matching matrix dimensions. The range finder now finds 15 columns, the reference result 16.

akohlmey:
**Summary**

**Related Issue(s)**

Fixes bugs reported in https://matsci.org/t/ndx2group-does-not-create-a-group/47545
Fixes #3690 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- update Plumed2 support with the recently released 2.8.2 and 2.7.6 versions
- enable building GPU package library with OpenMP enabled, if `-DBUILD_OMP=on` is used. Mention OpenMP support in docs
- provide access to `Force::special_lj` and `Force::special_coul` through the library interface and python module
- fix two bugs in `ndx2group` command
- download more recent snapshot for MSCG package that fixes an internal bug. 
- avoid race condition in "make install-python" with cmake.
- print warning when fix efield (and not fix efield/tip4p) is used with a TIP4P pair style
- must have a pair style defined for fix efield/tip4p to avoid segfault

**Post Submission Checklist**


cmake/Modules/Packages/GPU.cmake
```diff
@@ -407,6 +417,10 @@ elseif(GPU_API STREQUAL "HIP")
     target_compile_definitions(gpu PRIVATE -DMPI_GERYON -DUCL_NO_EXIT)
   endif()
   target_link_libraries(gpu PRIVATE hip::host)
+  if(BUILD_OMP)
```
rbberger: Wouldn't it be enough to add this to the end of the file once?

akohlmey: I have thought about that, but didn't like the disconnect. It is more obvious, if this command is where the other targets are added.

rbberger: So instead we duplicate code that we now have to maintain in multiple places?

akohlmey: Most of this file is an exercise in duplication...

akohlmey: "triplication" to be more accurate. The way I see it is the following: I already *know* that I have to check everything 3 times (CUDA, OpenCL, HIP), by moving the statement outside the sections, I have to remember to look at a fourth place. From that perspective, the current choice made more sense to me.

rbberger: I don't see that as a reason to add more cruft.

rbberger (dismissed):
Please remove the redundancies. There is actually an opportunity here to further remove existing duplicated code.

akohlmey:
> Please remove the redundancies. There is actually an opportunity here to further remove existing duplicated code.

done

sjplimp (approved):

rbberger (approved):

PR #3667, opened 2023-03-02
Add detailed developer information about adding pair styles and use new born/gauss style as example

akohlmey:
**Summary**

This expands the documentation about adding new styles to LAMMPS with a detailed description about adding  pair styles to augment the existing example for writing fixes.

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

N/A

**Implementation Notes**

The pair style is unusual, since its application is for liquid mercury at different temperatures and the parameterization is tunable. So it will also give an example to use ``Pair::extract()`` and fix adapt for a simulation with a temperature ramp.

**Post Submission Checklist**

- [x] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [x] The source code follows the LAMMPS formatting guidelines
- [x] Suitable new documentation files and/or updates to the existing docs are included
- [x] The added/updated documentation is integrated and tested with the documentation build system
- [x] The feature has been verified to work with the conventional build system
- [x] The feature has been verified to work with the CMake based build system
- [x] Suitable tests have been added to the unittest tree.
- [ ] One or more example input decks are included

PR #3688, opened 2023-03-14
Add foreign pair infrastructure

vchuravy:
**Summary**

This adds infrastructure to register pairs at runtime with LAMMPS.
The primary motivation is to support potentials written in Julia and to be able to call them from LAMMPS.

**Related Issue(s)**

https://github.com/cesmix-mit/LAMMPS.jl/pull/27

**Author(s)**

Valentin Churavy, Massachussetts Institute of Technology

**Licensing**

**Implementation Notes**

In https://github.com/cesmix-mit/LAMMPS.jl/pull/27 I am working on a full-fledged example calling Julia from LAMMPS using this feature.
I choose this full callback infrastructure since Julia functions are compiled and as such implementing the neighbor list walk in Julia
is faster than calling the `compute_energy` function from C++ (like it is done in `pair_python`). Furthermore we do not need to add a
dependency on Julia to LAMMPS.

**Post Submission Checklist**

### DIFF
--- a/src/PLUGIN/plugin.cpp
+++ b/src/PLUGIN/plugin.cpp
@@ -158,7 +158,7 @@ void plugin_register(lammpsplugin_t *plugin, void *ptr)
       if (lmp->comm->me == 0)
         lmp->error->warning(FLERR, "Overriding built-in pair style {} from plugin", plugin->name);
     }
-    (*pair_map)[plugin->name] = (Force::PairCreator) plugin->creator.v1;
+    (*pair_map)[plugin->name] = [=](LAMMPS *lmp) -> Pair* { return (Pair*) plugin->creator.v1(lmp); };
 
   } else if (pstyle == "bond") {
     auto bond_map = lmp->force->bond_map;
--- a/src/force.h
+++ b/src/force.h
@@ -17,6 +17,7 @@
 #include "pointers.h"
 
 #include <map>
+#include <functional>
 
 namespace LAMMPS_NS {
 class Angle;
@@ -86,7 +87,7 @@ class Force : protected Pointers {
   KSpace *kspace;
   char *kspace_style;
 
-  typedef Pair *(*PairCreator)(LAMMPS *);
+  typedef std::function<Pair*(LAMMPS *)> PairCreator;
   typedef Bond *(*BondCreator)(LAMMPS *);
   typedef Angle *(*AngleCreator)(LAMMPS *);
   typedef Dihedral *(*DihedralCreator)(LAMMPS *);
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -5564,7 +5564,7 @@ int lammps_get_last_error_message(void *handle, char *buffer, int buf_size) {
   return 0;
 }
 
-int lammps_foreign_add_pair_style(void *handle, const char *_name) {
+int lammps_foreign_add_pair_style(void *handle, const char *_name, void *ctx, Compute compute_fptr) {
   LAMMPS *lmp = (LAMMPS *) handle;
   std::string name = _name;
   auto &pair_map = *lmp->force->pair_map;
@@ -5573,8 +5573,7 @@ int lammps_foreign_add_pair_style(void *handle, const char *_name) {
   if (it != pair_map.end()) {
     return 1;
   }
-  // Issue: Can't capture things
-  pair_map[name] = [](LAMMPS * lmp) -> Pair* { return new PairForeign(lmp); };
+  pair_map[name] = [=](LAMMPS * lmp) -> Pair* { return new PairForeign(lmp, ctx, compute_fptr); };
 }
 
 // Local Variables:
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -5564,7 +5564,7 @@ int lammps_get_last_error_message(void *handle, char *buffer, int buf_size) {
   return 0;
 }
 
-int lammps_foreign_add_pair_style(void *handle, const char *_name, void *ctx, Compute compute_fptr) {
+int lammps_foreign_add_pair_style(void *handle, const char *_name, void *ctx, ForeignCompute compute_fptr) {
   LAMMPS *lmp = (LAMMPS *) handle;
   std::string name = _name;
   auto &pair_map = *lmp->force->pair_map;
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -50,6 +50,7 @@
 #include "universe.h"
 #include "update.h"
 #include "variable.h"
+#include "pair_foreign.h"
 
 #include <cstring>
 
@@ -5561,6 +5562,19 @@ int lammps_get_last_error_message(void *handle, char *buffer, int buf_size) {
   }
 #endif
   return 0;
+}
+
+int lammps_foreign_add_pair_style(void *handle, const char *_name) {
+  LAMMPS *lmp = (LAMMPS *) handle;
+  std::string name = _name;
+  auto &pair_map = *lmp->force->pair_map;
+
+  auto it = pair_map.find(name);
+  if (it != pair_map.end()) {
+    return 1;
+  }
+  // Issue: Can't capture things
+  pair_map[name] = [](LAMMPS * lmp) -> Pair* { return new PairForeign(lmp); };
 }
 
 // Local Variables:
--- a/src/library.h
+++ b/src/library.h
@@ -254,7 +254,8 @@ void lammps_force_timeout(void *handle);
 int lammps_has_error(void *handle);
 int lammps_get_last_error_message(void *handle, char *buffer, int buf_size);
 
-int lammps_foreign_add_pair_style(void *handle, const char *name);
+typedef void (*ForeignCompute)(void*, int, int);
+int lammps_foreign_add_pair_style(void *handle, const char *name, void *ctx, ForeignCompute compute_fptr);
 
 #ifdef __cplusplus
 }
--- a/src/library.h
+++ b/src/library.h
@@ -254,6 +254,8 @@ void lammps_force_timeout(void *handle);
 int lammps_has_error(void *handle);
 int lammps_get_last_error_message(void *handle, char *buffer, int buf_size);
 
+int lammps_foreign_add_pair_style(void *handle, const char *name);
+
 #ifdef __cplusplus
 }
 #endif
--- a/src/pair_foreign.cpp
+++ b/src/pair_foreign.cpp
@@ -1,0 +1,130 @@
+/* -*- c++ -*- ----------------------------------------------------------
+   LAMMPS - Large-scale Atomic/Molecular Massively Parallel Simulator
+   https://www.lammps.org/, Sandia National Laboratories
+   Steve Plimpton, sjplimp@sandia.gov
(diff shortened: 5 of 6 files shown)


akohlmey:
@vchuravy This is a fairly invasive change, so I am not very comfortable with it.

Couldn't you do the same by creating a plugin for your specific pair style(s) using a C-style plugin loader as shown with the examples in https://github.com/lammps/lammps/tree/develop/examples/plugins ?

vchuravy:
For me an important feature is adding pair styles at runtime. I want my users to be able to have many Julia based pair styles and not just a generic "Julia" style. 

If the plugin interface can be used without a C compiler I would be happy to use that, but from looking at the interface that seems not in scope?

akohlmey:
> For me an important feature is adding pair styles at runtime. I want my users to be able to have many Julia based pair styles and not just a generic "Julia" style.
> 
> If the plugin interface can be used without a C compiler I would be happy to use that, but from looking at the interface that seems not in scope?

Plugins are compiled with the same (C++) compiler as LAMMPS itself only the loader API in in C. So, if people can compile LAMMPS, they can compile the plugins.

You can put as much "magic" into the lammpsplugin_init function as you want and load many factory functions for many pair styles in one go. The corresponding factory functions can be created from a template using CMake scripting or a shell or python script or otherwise (similar as the main LAMMPS code creates files with lists of include statements to include enabled styles).

It may be a bit more effort, but is strongly preferred over what is in your draft pull request.

stanmoore1:
I can confirm you do not need a C compiler to use the C-style plugin library, C++ compiler works fine.

akohlmey:
> In [cesmix-mit/LAMMPS.jl#27](https://github.com/cesmix-mit/LAMMPS.jl/pull/27) I am working on a full-fledged example calling Julia from LAMMPS using this feature.
> I choose this full callback infrastructure since Julia functions are compiled and as such implementing the neighbor list walk in Julia
> is faster than calling the `compute_energy` function from C++ (like it is done in `pair_python`). Furthermore we do not need to add a
> dependency on Julia to LAMMPS.

I've looked at your repository and am trying to make sense of it. Am I understanding correctly, that you have written and are using a LAMMPS interface wrapper that sits on top of the LAMMPS library interface but is written in Julia?

This makes a *big* difference. I will have to think about if and how we can accommodate your needs in a less disruptive way.

vchuravy:
> Plugins are compiled with the same (C++) compiler as LAMMPS itself only the loader API in in C. So, if people can compile LAMMPS, they can compile the plugins.

Ideally I want to use neither a C/C++ compiler **at runtime**. I can look into using a Plugin instead of expanding the API as I am doing here (I wasn't aware of them)

> I've looked at your repository and am trying to make sense of it. Am I understanding correctly, that you have written and are using a LAMMPS interface wrapper that sits on top of the LAMMPS library interface but is written in Julia?

You understand that correctly :) My end goal is to allow folks to extend LAMMPS with Julia functionality. We have a use-case for a pair style right now, but I could see other extension points in the future.

In particular I am making use of Julia ability to generate C callbacks:

https://github.com/cesmix-mit/LAMMPS.jl/pull/27/files#diff-61bb2a239e2c058bbce9be06e7de03d29f00bb28dec46e703e4635259480ab97R19

and then use the C-API to implement a loop over the neighbor list: https://github.com/cesmix-mit/LAMMPS.jl/pull/27/files#diff-61bb2a239e2c058bbce9be06e7de03d29f00bb28dec46e703e4635259480ab97R60

vchuravy:
So I think I could use the plugin interface (it effectively does the same). There are two considerations I see.

1. I ended up using `std::function` in this PR. `std::function<Pair*(LAMMPS *)> PairCreator;`
    This was crucial since I need to capture the `ctx` and `callback` pointer from Julia, and use that in the constructor.
    Looking at the Plugin definition, that seems like information I would still need to feed through.
    I could make a plugin v3 constructor that passes a ctx pointer, or we could switch to `std::function`.
2. I could package things as a plugin, but I effectively would just want to call `registerPlugin` through the `C-API`.

akohlmey:
Ok, now that I can see clearer, I am not so sure whether a plugin is a viable approach. Ultimately a pair plugin has to be a self-contained "binary blob" that has a function that instantiates a Pair style class that then can just be called like any other pair style. I don't know enough about Julia to know whether that can be done.

Do you absolutely need to register the callback with the pair style constructor?
What if we would add a `Pair::register_callback(void *, PairCallback)` function or something similar to pair styles and expose that function through the library interface somehow? Then you would have to have a PairJuliaWrap pair style as a placeholder, but could register whatever function you want to use with it at runtime.

vchuravy:
I am not too familiar with lammps and I think that shows here. What do I actually need to implement for a pair style?

From the Julia side I just need a `void*` context argument to pass the Julia closure through and a function pointer to call back into Julia: `(*callback(void* ctx, ...))`

This works for the case where I only implement `PairForeign::compute`, but what about the other methods of the `Pair` interface. 
In the "worst" case I would could provide a call pointer for every method. 

So my goal was to associate `pair name` with a factory that instantiates a pseudo pair `ForeignPair` that consists of a set of pointers to call back. We could make this a bit cleaner by defining a struct that is that pointer bundle + ctx. 

In some sense this is why I needed to use a `std::function` since I effectively have a factory of a factory :)

> Do you absolutely need to register the callback with the pair style constructor?

Not really as long as I can name the callback set that would be fine

```
Pair::register_callback("julia_lj", void *, PairCallback)
```

and then users of `LAMMPS.jl` could use:

```
command(lmp, "pair julia_lj ...")
```

akohlmey:
@vchuravy please have a look at https://docs.lammps.org/Modify_pair.html
Incidentally, over the last weeks I have been working on a far more detailed description, but that is not yet complete: https://github.com/lammps/lammps/pull/3667

The best way to learn about what pair styles do, is to read through some of them. Representative examples are pair_lj_cut.cpp/.h, pair_morse.cpp/.h, pair_sw.cpp/.h  pair_eam.cpp/.h

Which brings me to another question: how do you propose to handle the domain decomposition and access to neighbor lists and other details? How complex would the functionality programmed in Julia be? Some pair styles in LAMMPS are straightforward since they just walk the neighbor list for computing pairwise additive interactions, others have also terms where you loop over the list multiple times (or first create a "short list" to do so more efficiently), and then there are models where the calculations can be very complex like ReaxFF which includes implicit treatments of bonds, angles, dihedrals, hydrogen bonds, etc. Things get complex when you have cases like EAM, where you first need to compute a per-atom property from its neighbors, then collect this data from neighboring subdomains, distribute the results and then compute the actual interaction.

Yes, passing a struct with function pointers may be a more general approach. Ideally, the solution would not be limited to your Julia use case, but could be interfaced with other programming languages as well.

akohlmey:
Here is something else you may want to have a look at: https://docs.lammps.org/fix_external.html
This may be an alternate approach using an already existing infrastructure. While this is a fix instead of a pair style, it can still be used in the same way, i.e. to compute forces on atoms by calling into an external code and then have LAMMPS use those forces to do the propagation of particles. This is should be rather close to what you have implemented in your pull request.

While that was originally intended for calling out to C or Fortran codes, we have managed to make it work with python as well. While it is not the most readable code, but you might figure out a few things by looking at the unit test for the fix external python interface in `unittest/python/python-fix-external.py`.

vchuravy:
That looks very close to what I want will take a detailed look tomorrow and see if I can use this as is.

Can there be more than one `fix external`?

akohlmey:
> Can there be more than one `fix external`?

Yes. Fixes are identified by their "fix-ID" and the registration function for the callback uses that fix-ID, so you can connect different callbacks to different fixes.
https://docs.lammps.org/Library_utility.html#_CPPv432lammps_set_fix_external_callbackPvPKc16FixExternalFnPtrPv

vchuravy:
> Which brings me to another question: how do you propose to handle the domain decomposition and access to neighbor lists and other details? How complex would the functionality programmed in Julia be? 

In my prototype I was planning on doing two things.

1. Expose the full flexibility of a pair style. The developer would need only be limited through the functionality exposed in `library.h`
2. Use that to provide a default implementation that iterates over the neighbor list in Julia and calls user code to calculate force based on the distance (similar to python_pair)

vchuravy:
Fix external does indeed look promising https://github.com/cesmix-mit/LAMMPS.jl/pull/28

One question that arises was. Why does the callback receive `nlocal`, `ids`, `x`, `fexternal`?
In particular `x` and `f` I would have looked up with `lammps_extract_atom`

akohlmey:
> One question that arises was. Why does the callback receive `nlocal`, `ids`, `x`, `fexternal`? In particular `x` and `f` I would have looked up with `lammps_extract_atom`

This interface was designed so it can be used without the library interface. Nlocal is to know how many atoms a specific processor holds as local (not ghost) atoms (the forces are only used for those), the atom tags are so you can know the identity of atoms (not always needed). Fexternal is used as a buffer, since fix external can be used to compute/update the external force not in every timestep but always apply it. That makes sense in certain circumstances where the computing the external force is very time consuming, but the accuracy needed is not so high that you can skip computing it in every step.

akohlmey:
> Fix external does indeed look promising [cesmix-mit/LAMMPS.jl#28](https://github.com/cesmix-mit/LAMMPS.jl/pull/28)

I've had a quick look and a couple of thoughts that I would like to share.

- You have to keep in mind, that LAMMPS has to kinds of atom indexing: global atom-IDs (i.e. Atom::tag) and local index (for accessing the *distributed* per-atom arrays like Atom::x, Atom::f, Atom::q, Atom::tag etc.). Most operations are *distributed* so you access atoms in the "local" order. Most importantly this is for the force and position arrays and the neighbor lists. That means, when you run in parallel, you must not use any scatter/gather functions, since those collate all distributed data into a large local array. That is wasteful, performance limiting and prevents you from good parallel scaling. So the julia code would be run in parallel and the same code would be executed on all processors with mpirun. So can look at the examples for running the python module in parallel for inspiration.

- With fix external you don't have a neighbor list created for you, but you can still trigger the creation of one (without!! undesired overhead) by adding a "pair_style zero ${cutoff}" command plus "pair_coeff * *". That will trigger creation of the neighbor list (by default a "half" neighbor list, but you can also request a "full" neighbor list) and you can access/query/use it through the library interface.

akohlmey:
@vchuravy This discussion has now moved away from the original subject and it doesn't look like the proposed changes will be needed (or are likely to be included), so I suggest to close the pull request and move the discussion elsewhere.
There are two possible places.
1. If you want to keep discussing in public (so that others can benefit and learn from it), you can post in the sub-forum here: https://matsci.org/c/lammps/lammps-development/42
2. If you want to continue discussing among LAMMPS developers only, you can request access to the LAMMPS developer slack by sending an email to slack@lammps.org and we'll send you an invite.

vchuravy:
> With fix external you don't have a neighbor list created for you, but you can still trigger the creation of one (without!! undesired overhead) by adding a "pair_style zero ${cutoff}" command plus "pair_coeff * *". That will trigger creation of the neighbor list (by default a "half" neighbor list, but you can also request a "full" neighbor list) and you can access/query/use it through the library interface.

Yeah I just ran into that! Thanks :)

PR #3689, opened 2023-03-15
cmake: remove pip install from install target

rbberger:
**Summary**

The current `make install` uses pip to download dependencies and built the LAMMPS Python package, which is not something reasonable to do in certain packaging scenarios (what this part of our CMake was originally meant for) since you can't rely on internet access.

This drops the Python package install via pip from `make install`.

**Related Issue(s)**

**Author(s)**

@rbberger

**Licensing**

**Backward Compatibility**

To install the LAMMPS Python package you will now have to use `make install-python` or manually build it through some other means by using the package `python/setup.py` (pip wheel, pyproject, etc.).

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -856,35 +856,6 @@ else()
     ${CMAKE_COMMAND} -E echo "Must build LAMMPS as a shared library to use the Python module")
 endif()
 
-###############################################################################
-# Add LAMMPS python module to "install" target. This is taylored for building
-# LAMMPS for package managers and with different prefix settings.
-# This requires either a shared library or that the PYTHON package is included.
-###############################################################################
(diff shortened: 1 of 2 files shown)


akohlmey:
> I'm open to dropping it completely from `make install`.

I am in favor of that. There is always the "make install-python" target.

akohlmey (approved):

junghans (approved):

junghans:
Does this need to be documented somewhere?

akohlmey:
> Does this need to be documented somewhere?

Some documentation needed to be removed from the documentation about how to install the Python module.
This should make it less confusing to people compiling from source instead of using a packaged version. Now the preferred and less complex option for that class of users is shown first.

rbberger:
LGTM

PR #3691, opened 2023-03-18
Refactor per-chunk computes to use ComputeChunk class with shared functionality

akohlmey:
**Summary**

This change removes a significant amount of replicated code in per-chunk compute styles by deriving those classes from a common base class. In addition some code constructs are modernized and clang-format enabled and applied.

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U.

**Licensing**

**Backward Compatibility**

Yes. ComputeChunk unit tests pass.

**Implementation Notes**

By using a dynamic cast we no longer need to check, if the requested compute id does point to a chunk/atom compute. The cast will return a NULL pointer, if the type does not match.

**Post Submission Checklist**


stanmoore1 (approved):
Looks OK to me

sjplimp (approved):

PR #3694, opened 2023-03-20
Bugfix fep tools

ludwig-ahrens:
If the free energy is divided by the volume, this should be done outside the logarithm. `tok[2]` is an average over $exp(-k_b T \Delta U)$, i.e., division with the volume outside the logarithm is equivalent to averaging over $exp(-k_b T \Delta U / v)$.
In fdti.py the sum is divided by `i` instead of `i-1` because `i=0` after the first contribution to the sum.

**Summary**

Fixed two bugs in tools/fep/fdti.py and tools/fep/fep.py.
One regarding the averaging over the volume and one regarding the weights in the trapezoidal integration.

**Related Issue(s)**

None.

**Author(s)**

Ludwig Ahrens-Iwers (TUHH)
ludwig.ahiw@gmail.com

**Licensing**

**Backward Compatibility**

None.

**Implementation Notes**

See description of this PR.

**Post Submission Checklist**

### DIFF
--- a/tools/fep/fdti.py
+++ b/tools/fep/fdti.py
@@ -10,7 +10,7 @@ if len(sys.argv) < 3:
     print("usage: fdti.py temperature hderiv < out.fep")
     sys.exit()
 
-rt = 0.008314 / 4.184 * float(sys.argv[1])
+rt = 0.008314 / 4.184 * float(sys.argv[1]) # in kcal/mol
 hderiv = float(sys.argv[2])
 
 line = sys.stdin.readline()
@@ -21,7 +21,7 @@ v = 1.0
 tok = line.split()
 if len(tok) == 4:
     v = float(tok[3])
-lo = -rt * math.log(float(tok[2]) / v)
+lo = -rt * math.log(float(tok[2])) / v
     
 i = 0
 sum = 0.0
@@ -29,9 +29,9 @@ for line in sys.stdin:
     tok = line.split()
     if len(tok) == 4:
         v = float(tok[3])
-    hi = - rt * math.log(float(tok[2]) / v)
+    hi = - rt * math.log(float(tok[2])) / v
     sum += (hi + lo) / (2 * hderiv)
     lo = hi
     i += 1
 
-print(sum/(i - 1))    # int_0^1: divide by i - 1 == multiply by delta
+print(sum/i)    # int_0^1: divide by i == multiply by delta
--- a/tools/fep/fep.py
+++ b/tools/fep/fep.py
@@ -9,7 +9,7 @@ if len(sys.argv) < 2:
     print("usage: fep.py temperature < out.fep")
     sys.exit()
 
-rt = 0.008314 / 4.184 * float(sys.argv[1])
+rt = 0.008314 / 4.184 * float(sys.argv[1]) # in kcal/mol
 
 v = 1.0
 sum = 0.0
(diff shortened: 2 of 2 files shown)


agiliopadua:
No. When the volume fluctuates the expression for the excess chemical potential is:
mu^{ex} = - kT ln [ < V exp(- Delta U / kT) > / < V > ]        (Allen, Tildesley 1st ed., eq. 2.69b)
in that case ("volume yes" option in compute fep), tok[2] = < V exp(...) > and tok[3] = < V >.

And for the division by (i-1) in fdti.py also no: in the trapeze rule, with 2 points we have the area of one trapeze, so multiplying by 1/(i - 1) is equivalent to multiplying by delta x (the limits of integration are 0 to 1).

ludwig-ahrens:
You are correct about the excess chemical potential, sorry.

However, I still think 1/(i -1) should be changed to 1/i. If I use an input with just two lines, I get a ZeroDivisionError.

agiliopadua:
Good for the chemical potential!
You're right for the fdti script: the first line is read before the for loop, so the variable i should be initialised as 1, not 0.
I'll correct this, thanks.

tools/fep/fdti.py
```diff
@@ -10,7 +10,7 @@
     print("usage: fdti.py temperature hderiv < out.fep")
     sys.exit()
 
-rt = 0.008314 / 4.184 * float(sys.argv[1])
+rt = 0.008314 / 4.184 * float(sys.argv[1]) # in kcal/mol
```
agiliopadua, line 13: useful

tools/fep/fdti.py
```diff
@@ -21,17 +21,17 @@
 tok = line.split()
 if len(tok) == 4:
     v = float(tok[3])
-lo = -rt * math.log(float(tok[2]) / v)
```
agiliopadua, line 24: this is correct

tools/fep/fep.py
```diff
@@ -9,7 +9,7 @@
     print("usage: fep.py temperature < out.fep")
     sys.exit()
 
-rt = 0.008314 / 4.184 * float(sys.argv[1])
+rt = 0.008314 / 4.184 * float(sys.argv[1]) # in kcal/mol
```
agiliopadua, line 12: useful

tools/fep/fep.py
```diff
@@ -19,6 +19,6 @@
     tok = line.split()
     if len(tok) == 4:
         v = float(tok[3])
-    sum += math.log(float(tok[2]) / v)
```
agiliopadua, line 22: this is correct

agiliopadua (changes requested):

tools/fep/fdti.py
```diff
@@ -21,17 +21,17 @@
...
     if len(tok) == 4:
         v = float(tok[3])
-    hi = - rt * math.log(float(tok[2]) / v)
+    hi = - rt * math.log(float(tok[2])) / v
     sum += (hi + lo) / (2 * hderiv)
     lo = hi
     i += 1
 
-print(sum/(i - 1))    # int_0^1: divide by i - 1 == multiply by delta
+print(sum/i)    # int_0^1: divide by i == multiply by delta
```
agiliopadua, line 37: I'd rather change line 26 to i = 1. Will submit a pull request.

agiliopadua (changes requested):

PR #3696, opened 2023-03-20
Minor updates/patches for BPM, rigid, and multi

jtclemm:
**Summary**

This PR includes some minor updates based on discussions with various users over the past few months. Changes include:

1. Adding an error if users use pair hybrid with fix update/special/bonds to avoid a seg fault when the fix grabs the neighbor list from the pair style
2. Adding a _break/no_ option to BPM bond styles for situations where the user does not intend on modeling bond breakage (can be used to overcome the limitation above)
3. Clarifying some language in the doc pages on the _multi_ neighbor option and adding some information on the scaling of computational costs for benchmarking
4. Patching a bug in fix rigid where the calculated virial contributions would not be packed/unpacked when exchanging atoms for non-sphere atom styles.

**Related Issue(s)**

NA

**Author(s)**

Joel Clemmer (SNL)

**Licensing**

**Backward Compatibility**

No issues

**Post Submission Checklist**


doc/src/bond_bpm_spring.rst
```diff
@@ -45,9 +48,7 @@ Examples
 Description
 """""""""""
 
-.. versionadded:: 4May2022
```
akohlmey, line 48: Why are you deleting this?

src/RIGID/fix_rigid.cpp
```diff
@@ -2523,9 +2523,17 @@ int FixRigid::pack_exchange(int i, double *buf)
   buf[2] = displace[i][0];
   buf[3] = displace[i][1];
   buf[4] = displace[i][2];
-  if (!extended) return 5;
+
```
akohlmey, line 2526: Can you attach an input deck example to reproduce this?
We also need to check if the same issue applies to other fix rigid variants.

doc/src/bond_bpm_spring.rst
jtclemm: Oops, that was a mistake when I cleaned up a merge conflict. Will revert.

src/RIGID/fix_rigid.cpp
jtclemm: The example I had was quite large, but now that I narrowed down the origin of the bug I can try and put together a minimal script to reproduce the issue (weird stress values along MPI grid lines after exchanging atoms).

The issue does not affect fix rigid/small (it doesn't return if _extended_) and the other rigid methods should inherit the exchange methods. But I'll double check.

doc/src/bond_bpm_spring.rst
akohlmey: You probably want to add a `.. versionadded:: TBD` where you document the new flag(s) (TBD will be replaced with the version data when the release is made). Since people seem to be using old versions of LAMMPS, but reading the online docs, we have started adding these (also retroactively, where easily possible) so that people know whether they can expect a feature to be available for their version (or which version they will need to install at least to get the feature).

doc/src/neighbor.rst
```diff
@@ -59,9 +59,16 @@ long cutoff, but other type pairs have a much shorter cutoff. The
...
 or its cutoff distance for interacting with other particles. Different
 sets of bins are then used to construct the neighbor lists as as further
-described by Shire, Hanley, and Stratford :ref:`(Shire) <bytype-Shire>`.
-This imposes some extra setup overhead, but the searches themselves may
-be much faster. By default, each atom type defines a separate collection
+described by Shire, Hanley, and Stratford :ref:`(Shire) <multi-Shire>`
+and Monti et al. :ref:`(Monti) <multi-Monti>`. This imposes some extra
+setup overhead, but the searches themselves may be much faster. For
+instance in a dense binary system with a ratio of the size of the largest
+to smallest collection bin :math:`\lamda`, the computational costs of
```
akohlmey: This is why the doc build fails when creating the pdf. You probably want `\lambda` not `\lamda`.

src/RIGID/fix_rigid.cpp
jtclemm: Using the current develop branch, this script should reproduce the bug when run on 4 processors [in.melt.txt](https://github.com/lammps/lammps/files/11023140/in.melt.txt). It creates a small circular rigid body in one quadrant and slowly moves it through a LJ melt. The dump file includes a column for the XX component of the virial from fix rigid. On timestep 900, there should be two non-zero values inside the fluid (particle 329 and 222, rendered below). With this PR, only rigid particles have stress contributions from fix rigid.


image

doc/src/bond_bpm_spring.rst
akohlmey: Added those myself.

jtclemm: Thank you, I missed that comment. I never noticed the version tags before, so that's good to know for the future.

akohlmey: No problem. This is a rather recent thing. Bottom line, I got tired of having to tell people, that tried to use a new feature with an old version of LAMMPS, to look up when changes were made on the bugs and latest features pages in the LAMMPS homepage. I recalled that Python uses this all over the place and decided to start the same with LAMMPS. Mostly with new or changed features and then occasionally also with older features (when convenient).

doc/src/neighbor.rst
```diff
@@ -59,9 +59,16 @@ long cutoff, but other type pairs have a much shorter cutoff. The
...
 sets of bins are then used to construct the neighbor lists as as further
-described by Shire, Hanley, and Stratford :ref:`(Shire) <bytype-Shire>`.
-This imposes some extra setup overhead, but the searches themselves may
-be much faster. By default, each atom type defines a separate collection
+described by Shire, Hanley, and Stratford :ref:`(Shire) <multi-Shire>`
+and Monti et al. :ref:`(Monti) <multi-Monti>`. This imposes some extra
+setup overhead, but the searches themselves may be much faster. For
+instance in a dense binary system in d-dimensions with a ratio of the size
+of the largest to smallest collection bin :math:`\lambda`, the computational
+costs of building a default neighbor list grows as :math:`\lambda^d` while
```
sjplimp: if I'm reading this right, it's saying both the default and multi costs are lambda^d ??

sjplimp (changes requested):
wlll approve, but please clarify my Q on the improved multi neigh scaling

sjplimp:
@jtclemm good catch on the bug in exchange of vatom for fix rigid vs fix rigid/small

doc/src/neighbor.rst
jtclemm: Nope, that was a mistake. Default and multi/old go as lamda^2d. Fixed now. This was a benchmark performed in 3D. The solid line has an exponent of 3 (the other factor of 3 is removed by the pair force normalization).

image

sjplimp: That's a nice plot, though probably too complex to include in the doc page.

This is saying the neigh cost is 2-3x that of pairwise forces?  Possibly that includes the effect of reneighboring infrequently.
But I think the big-picture point is that the new multi cost scales the same as force computation, independent of particle size ratios?  Whereas the old method did not.  If that is correct, than adding that as a sentence at the end of your current updated paragraph might be useful context for users.

sjplimp:
@jtclemm @stanmoore1 Is the new multi reneighbor method (documented better in this PR), included in Kokkos?  If not, would that be hard to do?

doc/src/neighbor.rst
jtclemm: I think these simulations were run for a single timestep, so the constant prefactor will drop during actual simulations. But I'd have to go back and look at the original files to say for sure. I had no plans to include the plot in the documentation, but it is available (with more context) in the cited paper. 

Agreed with the big picture. I do have the statement: "the costs for *multi* grows as :math:`\lambda^d`, equivalent to the cost of force evaluations" but I can make this more explicit.

jtclemm:
> @jtclemm @stanmoore1 Is the new multi reneighbor method (documented better in this PR), included in Kokkos? If not, would that be hard to do?

My recollection is that Kokkos does not even have the original multi implementation (multi/old). Stan will certainly know better than I do, but I think porting the new multi method would be a substantial effort. It'd require new bin, stencil, and npair classes as well as all the infrastructure to create the per-atom collection array plus the new communication protocols. The fact that the bin/stencil/npair classes in Kokkos are already templated should be a big help though, greatly simplifying the combinatorics.

That being said, there is definitely interest in porting more tools - multi, BPM, GRANULAR, and maybe in the future RHEO - to Kokkos.

stanmoore1:
No Kokkos support for `multi` yet.

sjplimp (approved):

PR #3698, opened 2023-03-22
use of type labels in set commands

yafshar:
**Summary**

This PR enables the use of type labels in set commands.

```bash
labelmap atom 1 C 8 H

set type C charge 1.0
set type H charge -1.0
set region right type H
set group all type C
set group all type/fraction H 0.5 453246

labelmap bond 1 C-C
set group edge bond C-C
```

**Related Issue(s)**

**Author(s)**

Yaser Afshar

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/doc/src/set.rst
+++ b/doc/src/set.rst
@@ -250,8 +250,9 @@ from a file.
    Support for type labels was added for setting atom, bond, angle,
    dihedral, and improper types
 
-Keyword *type* sets the atom type or atom type label for all selected atoms.
-For atom type, specified value must be from 1 to ntypes, where ntypes was set
+Keyword *type* sets the atom type for all selected atoms.  A specified value
+can be a type id or a type label for atom type. When using a 
+type id, the specified value must be from 1 to ntypes, where ntypes was set
 by the :doc:`create_box <create_box>` command or the *atom types* field in the
 header of the data file read by the :doc:`read_data <read_data>`
 command. In the case of the specified type label, it must have been defined
@@ -259,16 +260,16 @@ before use via an alphanumeric type label. See the
 :doc:`Howto type labels <Howto_type_labels>` doc page for the allowed syntax
 of type labels and a general discussion of how type labels can be used.
 
-Keyword *type/fraction* sets the atom type or atom type label for a fraction
-of the selected atoms.  The actual number of atoms changed is not guaranteed
+Keyword *type/fraction* sets the atom type for a fraction of the selected
+atoms.  The actual number of atoms changed is not guaranteed
 to be exactly the specified fraction (0 <= *fraction* <= 1), but
 should be statistically close.  Random numbers are used in such a way
 that a particular atom is changed or not changed, regardless of how
 many processors are being used.  This keyword does not allow use of an
 atom-style variable.
 
-Keywords *type/ratio* and *type/subset* also set the atom type or atom
-type label for a fraction of the selected atoms.  The actual number of
+Keywords *type/ratio* and *type/subset* also set the atom type for
+a fraction of the selected atoms.  The actual number of
 atoms changed will be exactly the requested number.  For *type/ratio* the
 specified fraction (0 <= *fraction* <= 1) determines the number.  For
 *type/subset*, the specified *Nsubset* is the number.  An iterative
--- a/doc/src/set.rst
+++ b/doc/src/set.rst
@@ -28,18 +28,18 @@ Syntax
 
   .. parsed-literal::
 
-       *type* value = atom type
+       *type* value = atom type or type label
          value can be an atom-style variable (see below)
        *type/fraction* values = type fraction seed
-         type = new atom type
+         type = new atom type or type label
          fraction = approximate fraction of selected atoms to set to new atom type
          seed = random # seed (positive integer)
        *type/ratio* values = type fraction seed
-         type = new atom type
+         type = new atom type or type label
          fraction = exact fraction of selected atoms to set to new atom type
          seed = random # seed (positive integer)
        *type/subset* values = type Nsubset seed
-         type = new atom type
+         type = new atom type or type label
          Nsubset = exact number of selected atoms to set to new atom type
          seed = random # seed (positive integer)
        *mol* value = molecule ID
@@ -107,10 +107,10 @@ Syntax
        *image* nx ny nz
          nx,ny,nz = which periodic image of the simulation box the atom is in
          any of nx,ny,nz can be an atom-style variable (see below)
-       *bond* value = bond type for all bonds between selected atoms
-       *angle* value = angle type for all angles between selected atoms
-       *dihedral* value = dihedral type for all dihedrals between selected atoms
-       *improper* value = improper type for all impropers between selected atoms
+       *bond* value = bond type or bond type label for all bonds between selected atoms
+       *angle* value = angle type or angle type label for all angles between selected atoms
+       *dihedral* value = dihedral type or dihedral type label for all dihedrals between selected atoms
+       *improper* value = improper type or improper type label for all impropers between selected atoms
        *sph/e* value = energy of SPH particles (need units)
          value can be an atom-style variable (see below)
        *sph/cv* value = heat capacity of SPH particles (need units)
@@ -145,15 +145,19 @@ Examples
 .. code-block:: LAMMPS
 
    set group solvent type 2
+   set group solvent type C
    set group solvent type/fraction 2 0.5 12393
+   set group solvent type/fraction C 0.5 12393
    set group edge bond 4
    set region half charge 0.5
    set type 3 charge 0.5
+   set type H charge 0.5
    set type 1*3 charge 0.5
    set atom * charge v_atomfile
    set atom 100*200 x 0.5 y 1.0
    set atom 100 vx 0.0 vy 0.0 vz -1.0
    set atom 1492 type 3
+   set atom 1492 type H
    set atom * i_myVal 5
    set atom * d2_Sxyz[1] 6.4
 
@@ -184,8 +188,8 @@ This section describes how to select which atoms to change
 the properties of, via the *style* and *ID* arguments.
 
 The style *atom* selects all the atoms in a range of atom IDs.  The
-style *type* selects all the atoms in a range of types.  The style
-*mol* selects all the atoms in a range of molecule IDs.
+style *type* selects all the atoms in a range of types or type labels.
+The style *mol* selects all the atoms in a range of molecule IDs.
 
 In each of the range cases, the range can be specified as a single
 numeric value, or a wildcard asterisk can be used to specify a range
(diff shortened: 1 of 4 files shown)


sjplimp:
@yafshar Sorry, there are a few things about this PR I don't understand.

(1) I thought the only use for type labels in an input script was to replace an atom type (in this case for the set command) with a type label.  But you are proposing to replace atom IDs with a type label?  Shouldn't the supported command be:
set type C charge 1.0 ?  Assuming you want to change all C atoms.  Which would alter the TYPE_SELECT portion of selection(), not ATOM_SELECT.

(2) I don't understand how the code you've added to the ATOM_SELECT portion works.  Does utils::expand_type() return a numeric type as a string?  E.g. C --> 3 ?  You can't compare atom tags (IDs) to an atom type.  What does your test script do?

(3) If the functionality of the set command is enhanced, it has to be documented on the doc/src/set.rst file.  And probably other places where type labels are discussed in the doc pages. Those changes need to be in this PR.

Finally, you should add @jrgissing Jake Gissinger to the review list for this PR.

yafshar:
> (1) I thought the only use for type labels in an input script was to replace an atom type (in this case for the set command) with a type label. But you are proposing to replace atom IDs with a type label? Shouldn't the supported command be: set type C charge 1.0 ? Assuming you want to change all C atoms. Which would alter the TYPE_SELECT portion of selection(), not ATOM_SELECT.

No, it was not the only use case. When we did the type labels with Jake, we thought many places in LAMMPS could take advantage of type labels, but we should enhance the code step by step per users' requests or need. We already did the same for [mass command](https://docs.lammps.org/mass.html). Now I am enhancing the kim in LAMMPS for bonded force fields, and we need to use type labels for the charge there. 
In my view, ATOM_SELECT was the natural case to add this change. Because users naturally use labels for atoms, but I see your point, we can add the same enhancement to TYPE_SELECT  as well.
> 
> (2) I don't understand how the code you've added to the ATOM_SELECT portion works. Does utils::expand_type() return a numeric type as a string? E.g. C --> 3 ? You can't compare atom tags (IDs) to an atom type. What does your test script do?
> 

@akohlmey  wrote [expand_type](utils::expand_type()) utility to do extra checks on the label, which expand the type label string into its equivalent numeric type and returns a null pointer if the type label is not defined.

> (3) If the functionality of the set command is enhanced, it has to be documented on the doc/src/set.rst file. And probably other places where type labels are discussed in the doc pages. Those changes need to be in this PR.
> 
I will complete the command and add the proper document.

> Finally, you should add @jrgissing Jake Gissinger to the review list for this PR.
done!

sjplimp:
@yafshar I don't think you are understanding my (2) comment.  I do not see how comparing a atom type to atom IDs 
can possibly work to select atoms, which is what the logic in ATOM_SELECT is doing.  Am I mis-understanding something ?

Also, re (1), I am not saying type labels cannot be used internally within LAMMPS in creative ways (e.g. fix bond/react, KIM).
I am saying I don't think there is any use case for them in an input script (or data/molecule file) other than to replace a numeric type (atom, bond, etc).  That is why in the set command, you have to add support for a type label only to replace a numeric atom type.  Same as for the mass command.

yafshar:
> @yafshar I don't think you are understanding my (2) comment. I do not see how comparing a atom type to atom IDs can possibly work to select atoms, which is what the login in ATOM_SELECT is doing. Am I mis-understanding something ?

Thanks, @sjplimp, I got your point. You are right. I will modify the code to account for TYPE_SELECT, which makes more sense.

> Also, re (1), I am not saying type labels cannot be used internally within LAMMPS in creative ways (e.g. fix bond/react, KIM). I am saying I don't think there is any use case for them in an input script (or data./molecule file) other than to replace a numeric type (atom, bond, etc). That is why in the set command, you have to add support for a type label only to replace a numeric atom type. Same as for the mass command.

Agreed. I will modify the PR, later today.

yafshar:
@jrgissing, would you please check this PR, and update the doc if you see a need?

akohlmey:
@yafshar can you please complete and insert the following paragraph into the set command documentation:

```
.. versionchanged:: TBD

   Support for type labels was added for 

```

At line 186 before "The style *atom*" and the same at line 243 before "Keyword *type*".

TBD will be updated when the next release is made and then people reading the online documentation that use older versions of LAMMPS can see why their input fails if they try to use type labels.

akohlmey:
@yafshar You inserted the tags, but didn't complete them. So I did this now. Please check for correctness.
Also, please note that you introduced memory leaks, since the result for `utils::expand_type()` is a temporary string that must be freed with `delete[]`.

akohlmey:
@sjplimp This now looks OK to me.

yafshar:
> @yafshar You inserted the tags, but didn't complete them. So I did this now. Please check for correctness. Also, please note that you introduced memory leaks, since the result for `utils::expand_type()` is a temporary string that must be freed with `delete[]`.

Thanks, @akohlmey

doc/src/set.rst
```diff
@@ -237,24 +245,32 @@ from a file.
...
 header of the data file read by the :doc:`read_data <read_data>`
-command.
+command. In the case of the specified type label, it must have been defined
+before use via an alphanumeric type label. See the
+:doc:`Howto type labels <Howto_type_labels>` doc page for the allowed syntax
+of type labels and a general discussion of how type labels can be used.
 
-Keyword *type/fraction* sets the atom type for a fraction of the
-selected atoms.  The actual number of atoms changed is not guaranteed
+Keyword *type/fraction* sets the atom type or atom type label for a fraction
```
jrgissing: You don't need to add "or atom type label" here. The command does not set the atom type label, it still just sets the atom type. This PR simply allows type labels to set the atom type.

doc/src/set.rst
```diff
@@ -183,9 +187,13 @@ properties to reset and what the new values are.  Some strings like
...
 the properties of, via the *style* and *ID* arguments.
 
+.. versionchanged:: 28Mar2023
+
+   Support for type labels was added for selecting atoms by type
+
 The style *atom* selects all the atoms in a range of atom IDs.  The
-style *type* selects all the atoms in a range of types.  The style
-*mol* selects all the atoms in a range of molecule IDs.
+style *type* selects all the atoms in a range of types or type labels.
```
jrgissing: Adding "or type labels" here is somewhat redundant, as stated. Atoms are still just selected by type, this PR just allows type labels to do that (as is clarified above).

doc/src/set.rst
```diff
@@ -237,24 +245,32 @@ from a file.
...
 should be statistically close.  Random numbers are used in such a way
 that a particular atom is changed or not changed, regardless of how
 many processors are being used.  This keyword does not allow use of an
 atom-style variable.
 
-Keywords *type/ratio* and *type/subset* also set the atom type for a
-fraction of the selected atoms.  The actual number of atoms changed
-will be exactly the requested number.  For *type/ratio* the specified
-fraction (0 <= *fraction* <= 1) determines the number.  For
+Keywords *type/ratio* and *type/subset* also set the atom type or atom
```
jrgissing: See my earlier comment

doc/src/set.rst
```diff
@@ -465,7 +481,10 @@ molecule which straddles the periodic box.
 
 Keywords *bond*, *angle*, *dihedral*, and *improper*, set the bond type
 (angle type, etc) of all bonds (angles, etc) of selected atoms to the
-specified value from 1 to nbondtypes (nangletypes, etc).  All atoms in a
+specified value from 1 to nbondtypes (nangletypes, etc). One can also use
```
jrgissing: this looks good

jrgissing (changes requested):

doc/src/set.rst
```diff
@@ -237,24 +245,32 @@ from a file.
...
 
-Keyword *type* sets the atom type for all selected atoms.  The
-specified value must be from 1 to ntypes, where ntypes was set by the
-:doc:`create_box <create_box>` command or the *atom types* field in the
+.. versionchanged:: 28Mar2023
+
+   Support for type labels was added for setting atom, bond, angle,
+   dihedral, and improper types
+
+Keyword *type* sets the atom type or atom type label for all selected atoms.
```
jrgissing: here is another occurrence

doc/src/set.rst
yafshar: @jrgissing, thanks. I see your point, so do you suggest removing the "or atom type label" here?  So should it be as before "Keyword *type/fraction* sets the atom type for a fraction"

jrgissing: yes, i think that is more correct

jrgissing (approved):

sjplimp (changes requested):
Hi @yafshar - two things:

(a) I pushed a new set.rst file with suggested edits for the doc page to make things more clear.
(b) Several places in the new set.cpp and now once in atom.cpp, you are overwriting one of the vector values passed in by char **arg, with a new char * string created by utils::expand_type().  And then you are doing a delete [] of that char * string within set.cpp of atom.cpp.

This seems wrong to me, since input.cpp will later try to deallocate its own char **arg data ?
In a stylistic sense, I don't think set.cpp or atom.cpp methods should be changing the char **arg data passed to it from Input.

@akohlmey can comment on this better than me

yafshar:
> utils::expand_type(

@sjplimp, thanks for the edit. It makes this more clear. Regarding the `utils::expand_type`, I was thinking the same, but looking at the function it returns a new temporary array `new char` using 
https://github.com/lammps/lammps/blob/b8ff478d7874a2ae0ef9402fc4728a77e87214a4/src/utils.cpp#L862
https://github.com/lammps/lammps/blob/b8ff478d7874a2ae0ef9402fc4728a77e87214a4/src/utils.cpp#L999 and @akohlmey pointed out that if I do not delete these temporary arrays, I will introduce memory leaks.

jrgissing:
> In a stylistic sense, I don't think set.cpp or atom.cpp methods should be changing the char **arg data passed to it from Input.

I was originally okay with this idea (similar to an immediate variable evaluation using 'label2type') because all the internal lammps code assumes the argument is a numeric type

yafshar:
> > In a stylistic sense, I don't think set.cpp or atom.cpp methods should be changing the char **arg data passed to it from Input.

@sjplimp @akohlmey, if it is not OK to overwrite *arg, we can avoid that. It only duplicates the lines. For example,

```c++
char *typestr = utils::expand_type(FLERR,arg[iarg+1],Atom::ATOM,lmp);
if (typestr) ivalue = utils::inumeric(FLERR,typestr,false,lmp);
else ivalue = utils::inumeric(FLERR,arg[iarg+1],false,lmp);
delete[] typestr;
```

sjplimp:
@yafshar yes, the issue is not deleting the string which utils::expand_type() allocated - that is necessary

the issue is overwriting the pointer in args (allocated and managed by the input class)

akohlmey:
I agree that conceptually it is problematic to overwrite the pointers for the individual "words" in the `arg` list, because you don't know how they were allocated (e.g. there could be a mismatch between new/delete and malloc/free, newer code uses the former, older code the latter).

> the issue is overwriting the pointer in args (allocated and managed by the input class)

In this case, it works out by chance, because the "words" are not individually allocated, but they are pointers into a string buffer with the entire line of input after applying all substitutions. Thus the "words" are not individually deallocated, but just the entire buffer and *that* pointer cannot be modified by overwriting any of the "arg" list pointers.

akohlmey:
> It only duplicates the lines.

Not necessarily. In fact, you can get rid of the assignment and write this so only one line is added.
```
       char *typestr = utils::expand_type(FLERR, arg[iarg + 1], Atom::ATOM, lmp);
       ivalue = utils::inumeric(FLERR, typestr ? typestr : arg[iarg + 1], false, lmp);
```

yafshar:
> > It only duplicates the lines.
> 
> Not necessarily. In fact, you can get rid of the assignment and write this so only one line is added.
> 
> ```
>        char *typestr = utils::expand_type(FLERR, arg[iarg + 1], Atom::ATOM, lmp);
>        ivalue = utils::inumeric(FLERR, typestr ? typestr : arg[iarg + 1], false, lmp);
> ```

@akohlmey I fixed the copy to arg. If you rather have one line as you suggested, I can revert the change.

akohlmey:
> @akohlmey I fixed the copy to arg. If you rather have one line as you suggested, I can revert the change.

Yes, duplicating code should be avoided. Every line of code that does not need to be written is automatically bug free.
Remember that you can safely use `delete[]` on a null pointer (unlike `free()`).

akohlmey:
@yafshar please run "make fix-whitespace" in the src folder and commit any changes.
The SetProperty unit test was failing on MacOS, so please keep an eye on that.

sjplimp (approved):
@yafshar Thanks for making these changes

sjplimp:
@akohlmey Ready to merge from my perspective

PR #3705, opened 2023-03-25
Collected small fixes and updates

akohlmey:
**Summary**

This pull request combines multiple small bug fixes and updates.

**Related Issue(s)**

Addresses concerns from https://matsci.org/t/questions-and-improvements-about-the-pair-ylz-cpp-file/47676/
Fixes #3708 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

N/A

**Implementation Notes**

The following individual changes are included:
- fix typo in recently added ComputChunk class
- fix uninitialized data access bug in fix wall/gran/region
- plug memory leaks, improve error message, and remove redundant calls to fmt::format()  in fix electrode/conp
- remove redundant code and improve error message in EAM pair style variants
- move check for non-ellipsoid atoms in pair style ylz to compute() function so it is compatible with hybrid pair styles where not all pair styles use ellipsoids
- initialize *all* pointers of Pair class to null and move the initialization to the initializer list of the constructor
- enable and apply clang-format for a few more files
- apply logic simplifications and modernizations suggested by clang-tidy
- use MathSpecial::powint() instead of pow() for integer powers in wall fixes and Pair class
- tweak sphinx settings so the HTML and PDF manual have the same chapter numbers

**Post Submission Checklist**

- [x] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [x] The source code follows the LAMMPS formatting guidelines
- [x] The feature has been verified to work with the conventional build system
- [x] The feature has been verified to work with the CMake based build system

PR #3709, opened 2023-03-27
Minor fixes that might be of interest...

GenieTim:
**Summary**

This PR contains one bugfix and two documentation improvements.

I admit that bundling (foreign) changes is usually a bad idea, but given that these are so small, I would neither feel comfortable opening three PR for 1 line of code each. If you prefer the latter, i will split it up, of course.

Here a summary:
- fix a typo
- explain that the fix ave/correlate/long can also use vector type variables
- fix a bug where using wildcards in the fix ave/correlate/long lead to unjustified `utils::missing_cmd_args` because of the number of arguments not being properly updated from the expansion

**Related Issue(s)**

None.

**Author(s)**

**Licensing**

**Backward Compatibility**

Inputs that previously threw an error might suddendly work.

**Implementation Notes**

For the bug fix, rather than re-assigning `nargs`, I looked at similar code and therewith also just used the `nargnew` variable where needed.

**Post Submission Checklist**

None.

### DIFF
--- a/doc/src/fix_ave_correlate_long.rst
+++ b/doc/src/fix_ave_correlate_long.rst
@@ -15,7 +15,7 @@ Syntax
 * Nevery = use input values every this many time steps
 * Nfreq = save state of the time correlation functions every this many time steps
 * one or more input values can be listed
-* value = c_ID, c_ID[N], f_ID, f_ID[N], v_name
+* value = c_ID, c_ID[N], f_ID, f_ID[N], v_name, v_name[I]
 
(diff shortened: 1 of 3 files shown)


akohlmey:
@GenieTim Thanks for your submission. No problem on bundling minor changes. We do this all the time. I usually keep one pull request open to collect such changes in response to bug reports, comments from people, reports of documentation issues etc. See #3705 for example.

sjplimp (approved):

PR #3711, opened 2023-03-27
BUGFIX: pppm_kokkos.cpp: for triclinic case,  initialize h_inv on dev…

yury-lysogorskiy:
**Summary**

BUGFIX: pppm_kokkos.cpp: for triclinic case,  initialize h_inv on device to compute h_x, h_y, h_z

**Related Issue(s)**
pppm/kk  with triclinic cell  show
ERROR: Could not compute g_ewald (src/KOKKOS/pppm_kokkos.cpp:1103)

**Author(s)**

Yury Lysogorskiy

**Licensing**

**Backward Compatibility**

**Implementation Notes**

Move h_inv to device before calling x2lamdaT

```
h_inv = Few<double, 6>(domain->h_inv);
x2lamdaT(&tmp[0],&tmp[0]);
```

**Post Submission Checklist**

### DIFF
--- a/src/KOKKOS/pppm_kokkos.cpp
+++ b/src/KOKKOS/pppm_kokkos.cpp
@@ -1016,6 +1016,7 @@ void PPPMKokkos<DeviceType>::set_grid_global()
     tmp[0] = nx_pppm;
     tmp[1] = ny_pppm;
     tmp[2] = nz_pppm;
+    h_inv = Few<double, 6>(domain->h_inv);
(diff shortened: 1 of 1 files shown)


stanmoore1:
@yury-lysogorskiy this is indeed a bug, nice catch! The `set_grid_global` function (and a few others) are directly copied from the parent class, so I removed the duplicated code in #3712 which should fix the issue as well. The duplicated CPU code was calling a Kokkos function that got overidden with the same name, when it should have called the host function instead.

PR #3714, opened 2023-03-29
New pair styles lj/cut/sphere. lj/expand/sphere and lepton/sphere

akohlmey:
**Summary**

This adds three new pair styles lj/cut/sphere, lj/expand/sphere, and lepton/sphere that use the "Atom::radius" property in their force computation. OPENMP package versions are included.

**Related Issue(s)**

Inspired by this MatSci forum discussion: https://matsci.org/t/how-to-create-randomly-sized-spheres-within-lammps-by-command-lines-or-reading-a-data-file/47739

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- For lj/cut/sphere the radius is used to determine the sigma for the Lennard-Jones potential from mixing
- Fox lj/expand/sphere the radius is used to determine the delta parameter for in "inner" size of the potential
- For lepton/sphere two properties "radi" and "radj" are made available in the Lepton expression similar to how "qi" and "qj" are available with "lepton/coul".

**Post Submission Checklist**


doc/src/pair_lj_sphere.rst
```diff
@@ -0,0 +1,124 @@
...
+   E = 4 \epsilon \left[ \left(\frac{\sigma_{ij}}{r}\right)^{12} -
+       \left(\frac{\sigma_{ij}}{r}\right)^6 \right]
+                       \qquad r < r_c
+
+:math:`r_c` is the cutoff.
+
+This is the same potential function as used by the :doc:`lj/cut
+<pair_lj>` pair style, but the :math:`\sigma_{ij}` parameter is not set
+as a per-type parameter via the :doc:`pair_coeff command <pair_coeff>`,
+but taken from the per-atom radius attribute of :doc:`atom_style sphere
```
sjplimp: Please clarify if you mean radius or diameter is used in setting sigma.  I think you mean diameter?
E.g. if two particles with radius = 0.5 are interacting, then sigmaij = 1.0 = ave of two diameters?
I think from the user perspective per-atom radius is often an internal parameter and data files and the set command, e.g. for atom_style sphere, specify diameters.

src/EXTRA-PAIR/pair_lj_sphere.cpp
```diff
@@ -0,0 +1,354 @@
...
+      dely = ytmp - x[j][1];
+      delz = ztmp - x[j][2];
+      rsq = delx * delx + dely * dely + delz * delz;
+      jtype = type[j];
+
+      if (rsq < cutsq[itype][jtype]) {
+        r2inv = 1.0 / rsq;
+        r6inv = r2inv * r2inv * r2inv;
+
+        sigma6 = powint(mix_distance(rtmp, radius[j]), 6);
```
sjplimp: I think if these are not diameters, then 2 particles of radius 0.5, will want to be 0.5 apart (eng minimum) instead of 1.0 apart.

doc/src/pair_lj_sphere.rst
akohlmey: The code is using the radius, but if you put it this way, it seems indeed more natural to use the mixing result of 2x rad_i and 2x rad_j.

src/EXTRA-PAIR/pair_lj_sphere.cpp
sjplimp: Said another way, I think if you create a system of diam=1 spheroidal particles and use this new pair style, you want it to be identical to pair lj/cut with point particles and sigma = 1 ?

akohlmey: > you want it to be identical to pair lj/cut with point particles and sigma = 1 ?

My test had diameter 2.0 for that case, but I see now that I want diameter 1.0 -> sigma 1.0. So I have to apply a factor of 2.0 to the result of mix_distance() and update the documentation.

akohlmey: @sjplimp Please see the revised test inputs and their outputs attached here. This is what it should be.

[in.melt-lepton.txt](https://github.com/lammps/lammps/files/11114633/in.melt-lepton.txt)
[in.melt-sphere.txt](https://github.com/lammps/lammps/files/11114635/in.melt-sphere.txt)
[in.melt.txt](https://github.com/lammps/lammps/files/11114636/in.melt.txt)
[log.melt-lepton.txt](https://github.com/lammps/lammps/files/11114637/log.melt-lepton.txt)
[log.melt-sphere.txt](https://github.com/lammps/lammps/files/11114638/log.melt-sphere.txt)
[log.melt.txt](https://github.com/lammps/lammps/files/11114639/log.melt.txt)

sjplimp:
Conceptual Qs about defining cutoffs for a polydisperse finite-size LJ model and building neighbor lists.  
I'm including @jtclemm as he may have comments on this.  

a) For a highly polydisperse systems (large size ratios), how does a user choose a good cutoff(s) ?
     One thing they might do is have one atom type and a bunch of diameters, and define one
     (huge) cutoff something like 2.5 * max_diam.  Which could be a poor physical model and terrible
     computationally.
b) It seems like a consistent model might scale the cutoff by sigma for every pairwise sigma.
     So that every pair of particles start interacting when they are within a distance that is some
     multiple (e.g. 2.5) of their pairwise sigma.
c) The granular pair styles effectively do this, even for systems with very large particle size ratios
    to insure they find all contacts between any pair of particles.  They use a different Npair class
    for that (which lj/sphere could potentially invoke).  Essentially the granular styles do not
    define a per-type cutoff, that is internally automated based on the particle diameters.

Not saying the current PR needs to address this issue.  Did the MatSci folks provide any details
about what their use-case model is?

akohlmey:
> Not saying the current PR needs to address this issue. Did the MatSci folks provide any details
> about what their use-case model is?

There was only a mention of creating a porous layer. The indicated size range was small (1 to 3).
After I did the Lepton version, it occurred to me, that they might be better of with a harmonic potential.

akohlmey:
@sjplimp there is another aspect to this: you could also conceive a version that is more like lj/expand or colloid. I think when you have a spread of radii that is more than an order of magnitude, you probably want that kind of model and not what is done here.

jtclemm:
> The granular pair styles effectively do this, even for systems with very large particle size ratios
> to insure they find all contacts between any pair of particles.

Specifically, that would be like the cohesive DMT/JKR granular pair interactions which have an attractive force that extends beyond the sum of particle radii.

Looking at lj/sphere, since sigma depends on the particle radii but the cutoff does not, should there be a warning/error if the user chooses a cutoff < than the maximum atom diameter x 2^(1/6) which would create a discontinuous repulsive force?

akohlmey:
> Looking at lj/sphere, since sigma depends on the particle radii but the cutoff does not, should there be a warning/error if the user chooses a cutoff < than the maximum atom diameter x 2^(1/6) which would create a discontinuous repulsive force?

That is rather tricky business, since you have to walk the entire neighbor list. You can have per-type cutoffs (and per-type epsilons) and you'd have to check all "mixed" sigmas against the per-type cutoffs.

We don't have an equivalent check in lj/cut either, though, and I've seen recently seen people (on the MatSci forum) set cutoffs shorter than sigma without blinking an eye (because it was set in a publication, which was from a rather dubious journal).

jtclemm:
> We don't have an equivalent check in lj/cut either,

That's fair. I was thinking just to loop over local atoms and find the largest, no neighbor loop, but if other LJ pair styles don't have a similar restriction then it makes sense not to issue warnings/errors.

I didn't fully understand the application on MatSci, but similar to what Steve said, it seems like one might want one of three choices:

-    a per-type cutoff, defined with pair_coeff
-    a fixed additive distance beyond the sum of atom radii, cut = r_i + r_j + X
-    a fixed multiplicative factor beyond the sum of atom radii, cut = X * (r_i + r_j)

The latter two options using the same cutoff mechanics as granular pair styles. Maybe the others could be added as alternate options down the road if needed/requested?

sjplimp:
@akohlmey @jtclemm All good comments.  I think we should simply do the following for the current PR.

a) Check in init() for the issue Joel mentioned that a cutoff is smaller than 2^(1/6) sigma (discontinuous forces).
I think this could be done cheaply by looping over owned atoms once, find the maxdiam for each atom type.
Then Allreduce the per-type maxdiams.  Then a double loop over types to check the cutoff for that type-pair against
the sum of the two per-type maxdiams.  Issue a WARN if any of them are smaller.   I think this warning is important
b/c there is no required connection between atom types and particle sizes and it would be easy for the user to mess up.

b) Add some guidance to the doc page.  Explain the issue in (a) and emphasize the user needs to think carefully
about choice of cutoffs.  First in the context of their model.  Second for performance.   Suggest that for a system with large size ratios (even 3 is large I think), it would be wise to assign particle types based on a binning of the particle sizes and then
assign cutoffs accordingly.  I think if this were done, the script could also use the multi neighbor style
and get improved performance.  Is this correct @jtclemm ?  Maybe in that case they should only specify cutoffs
for type pairs with the same index?  E.g. 1/1, 2/2, 3/3, 4/4.

jtclemm:
> I think if this were done, the script could also use the multi neighbor style
> and get improved performance. Is this correct @jtclemm ?

Since cutoffs are solely determined by the pair coefficients and are independent of particle radii, this pair style is effectively the same as any other LJ pair style from the perspective of neighboring. So yes, multi will work regardless of how users decide to bin atoms by radius to assign to types. Although, as you mention, hopefully the users make reasonable choices with longer cutoffs for larger particles given the context of the model.

akohlmey:
Rather than adding that complex check in init() that could produce false positives, I would prefer to revisit Steve's idea of making the cutoff not a fixed distance, but rather a factor that is multiplied by sigma. Then the choices in pair_coeff would be independent from the particle diameter and mistakes much harder to make. We could have a symbolic value "min" or "minimum" to represent 2^(1/6) for convenience, to implement a repulsive-only interaction.
I would have to determine the per-type maximum diameter/radius to derive cutoff values for the neighborlist code.

jtclemm:
If the cutoff depends on the radius, one will need to find the global maximum radius (as opposed to the per-type maximum). For context, that is already calculated in pair gran init_style() starting here:
https://github.com/lammps/lammps/blob/e5a82bc4611a3a9a6dab10a554b792d406e079ce/src/GRANULAR/pair_gran_hooke_history.cpp#L485
It loops over local atoms and also checks for instances of fix pour/deposit in case larger atoms might be inserted.

For multi, the pair style would also need to define atom2cut() and radii2cut() methods so the neighbor class can identify how far a given particle can interact or how far the  maximum interaction for a collection bin size can be. E.g. here:
https://github.com/lammps/lammps/blob/e5a82bc4611a3a9a6dab10a554b792d406e079ce/src/GRANULAR/pair_gran_hooke_history.cpp#L815

sjplimp:
@akohlmey By this:
>> I would have to determine the per-type maximum diameter/radius to derive cutoff values for the neighborlist code.

do you mean follow the logic in pair_granular.cpp in init_one() for returning a cutoff,
and request an NPair size class that uses the radii of the 2 particles ?

akohlmey:
> @akohlmey By this:
> 
> > > I would have to determine the per-type maximum diameter/radius to derive cutoff values for the neighborlist code.
> 
> do you mean follow the logic in pair_granular.cpp in init_one() for returning a cutoff, and request an NPair size class that uses the radii of the 2 particles ?

Up until now, I have not yet considered to follow the logic in pair style granular, but my overall assumption so far has been that the diameter would not vary a lot. I believe for a system with a larger spread of diameters, one would rather use a pair style similar to lj/expand, where the diameter would represent the delta parameter and not sigma. I vaguely recall that somebody was asking for something like that a long time ago.

As a consequence of the discussion, I may need to revisit and review some of the choices I have made and also make some experiments with widely polydisperse test systems.

jtclemm:
> Up until now, I have not yet considered to follow the logic in pair style granular,

If interaction cutoffs depend on radii, an lj/sphere pair style could be implemented in the refactored pair granular in ~10 lines as a GranSubModNormal class. Pair granular would then handle all the issues with max radii/cutoffs. I could mock it up pretty quickly if that'd be helpful.

However, if this pair style is buried under the mountain of information in the pair granular doc page, it might never be found by any interested users. It'd also be slower.

sjplimp:
I like Joel's suggestion for adding an option to PairGranual for LJ for highly polydisperse systems.
But we could keep this PR for mildly polydisperse systems, where how the user handles cutoffs and the associated neighboring is not that significant an issue. The doc page for both commands could point to the other command for the different flavors of model.

akohlmey:
@sjplimp @jtclemm I have updated the code for lj/sphere for using a cutoff ratio instead of a cutoff. For my test example with reduced units, there is no difference in the input and the results match with previous runs. Only the on-the-fly (re-)computation of the cutoff makes it a tiny bit slower.

doc/src/pair_lj_sphere.rst
```diff
@@ -0,0 +1,130 @@
...
+
+This is the same potential function as used by the :doc:`lj/cut
+<pair_lj>` pair style, but the :math:`\sigma_{ij}` parameter is not set
+as a per-type parameter via the :doc:`pair_coeff command <pair_coeff>`,
+but taken from the per-atom diameter attribute of :doc:`atom_style
+sphere <atom_style>`.  The individual value of :math:`\sigma_{ij}` is
+computed from the diameters of the two atoms using the mixing rule for
+pair coefficients as set by the :doc:`pair_modify mix <pair_modify>`
+command (defaults to geometric mixing).  The cutoff is not specified as
+a distance, but as ratio that must be multiplied with
```
sjplimp: must be multiplied with --> is internally multiplied by

doc/src/pair_lj_sphere.rst
```diff
@@ -0,0 +1,130 @@
...
+Examples
+""""""""
+
+.. code-block:: LAMMPS
+
+   pair_style lj/sphere 2.5
+   pair_coeff * * 1.0
+   pair_coeff 1 1 1.1 2.8
+
+Description
```
sjplimp: This section needs some context like we discussed.  About using it only for mildly polydisperse systems for efficiency reasons. Also advice to bin particle sizes by type, if want to use it with neigh multi. 
Also, the dangers of using a cutoff scale factor less than 2*(1/6) which leads to discontinuous forces. Possibly that should be a warning or error in the code.

akohlmey: @sjplimp I've tried to write something to discuss the situation. Please have a look. I made some tests but getting a significant speedup from using `neighbor multi` is rather elusive, even for a diameter range from 0.01 to 1.5 sigma (in reduced units) and only grouping into two atom types. For most choices, `neighbor bin` is faster.

> Also, the dangers of using a cutoff scale factor less than 2*(1/6) which leads to discontinuous forces.

This is not really a specialty of this pair style, but applies to pretty much *all* pair styles that do not use a a force switching function. Furthermore, the force is discontinuous at the cutoff for *all* choices of the cutoff. So I don't see a particular reason to add this here and not elsewhere.

jtclemm: For atom sphere we added the `neigh_modify collection/interval` command to simplify the multi style. This option requires a pair style have the `atom2cut()` and `radii2cut()` so the neighbor class can sort atoms into collections and determine the stencil size for each collection bin. The flag `force->pair->finitecutflag` is also required.

In the granular package, these are defined here:
https://github.com/lammps/lammps/blob/cbf7f66a6f48080fe01ac2b919ed22474a2bdb08/src/GRANULAR/pair_gran_hooke_history.cpp#L815
https://github.com/lammps/lammps/blob/cbf7f66a6f48080fe01ac2b919ed22474a2bdb08/src/GRANULAR/pair_granular.cpp#L848

That being said, I think the regular per-type multi should still work without these methods. So it's odd a significant speedup wasn't attained...

akohlmey: > That being said, I think the regular per-type multi should still work without these methods. So it's odd a significant speedup wasn't attained...

Feel free to experiment. I am attaching my test example.
[in.melt-sphere-random.txt](https://github.com/lammps/lammps/files/11145084/in.melt-sphere-random.txt)


> For atom sphere we added the `neigh_modify collection/interval` command to simplify the multi style.

Before I can implement those correctly, I have to spend more time reading the pair styles' that use those.

akohlmey: I would propose to try and finish this PR "as is" and then try improvements together with the version for granular. For the granular style I can even see two options: one like lj/cut where sigma is taken from the atom diameter and a second like lj/expand, where sigma is specified (and the same for the same atom type) and instead the "delta" parameter is taken from the atom diameter. In the context of DEM models, the latter makes more sense to me.
This pair style is more like a shortcut to set up lj/cut style simulations with 100s or 1000s of atom types (as in the one example I provided on MatSci.org).

doc/src/pair_lj_sphere.rst
```diff
@@ -0,0 +1,158 @@
...
+
+Note that :math:`\sigma_{ij}` is defined in the LJ formula above as the
+zero-crossing distance for the potential, *not* as the energy minimum which
+is at :math:`2^{\frac{1}{6}} \sigma_{ij}`.
+
+.. admonition:: Notes on cutoffs, neighbor lists, and efficiency
+   :class: note
+
+   Because the cutoff in this pair style depends on the diameter of the
+   atoms, this influences which cutoff is used to build neighbor lists
```
sjplimp: which cutoff is used --> how cutoffs are used

doc/src/pair_lj_sphere.rst
```diff
@@ -0,0 +1,158 @@
...
+   and how effective those neighbor lists are in avoiding to compute
```
sjplimp: avoiding computation of pairwise distances

Also, is this referring to avoiding pairwise distance comp when the neighbor list is built,
or excluding pairs from the neigh list, to save pairwise distance comp in the pair style?

doc/src/pair_lj_sphere.rst
```diff
@@ -0,0 +1,158 @@
...
+   distances between non-interacting atoms.  This pair style uses a
+   conventional neighbor list construction like for :doc:`pair style
```
sjplimp: like for --> the same as (or "similar` to" if that is more accurate)

doc/src/pair_lj_sphere.rst
```diff
@@ -0,0 +1,158 @@
...
+   lj/cut <pair_lj>`, where the cutoffs are typically rather similar.
+   LAMMPS will determine the largest cutoff and use this value for
+   building the neighbor lists.  This can be inefficient, if the
+   difference between cutoffs is very large.  The command :doc:`neighbor
```
sjplimp: type pair cutoffs

Note that one possible confusion here for the user, is that their pair coeff commands could all have the same value for the cutoff, but b/c it is a ratio, the actual cutoffs are much different.  Maybe it would be better to explain this in terms of the degree of polydispersity, e.g. how big are the size ratios.

doc/src/pair_lj_sphere.rst
```diff
@@ -0,0 +1,158 @@
...
+   multi <neighbor>` can be used to enable a modified neighbor list
+   algorithm, that uses different size bins for atom types with
+   different cutoffs and constructs neighbor lists based on those
```
sjplimp: ditto on different cutoffs versus different particle sizes

doc/src/pair_lj_sphere.rst
```diff
@@ -0,0 +1,158 @@
...
+   cutoffs.
+
+   If atom diameters vary largely when using pair style *lj/sphere*,
```
sjplimp: I see that you are discussing size ratios here.  However I don't think anyone would likely use this pair style to have particles that are (nearly) the same size, and have a variety of different cutoffs (previous paragraph).  They would just use lj/cut for that.  So I suggest making this paragraph the first and primary focus of how the user needs to be careful with this pair style.

sjplimp:
Agree with @akohlmey that we don't need to warn about too short of cutoffs making discontinous forces - good point that same is true for other pointwise LJ models.  Also agree that once the doc info is complete, it's fine to merge this PR as-is, and worry about connections to granular pair styles later.  And I agree with @jtclemm that adding a simple LJ model to the granular options is a good idea.

doc/src/pair_lj_sphere.rst
akohlmey: We may not be communicating here.  The pair style is primarily not so much different cutoffs but about having particles with different *sigma*; like a different sigma for *every* atom. Using lj/cut for that would be a pain if you have a large number of particles. The different cutoffs are a consequence of the sigmas, since it makes more sense to have the cutoff depend on sigma rather than using a fixed cutoff.

akohlmey:
@sjplimp here is the next iteration of the docs. I've added parts of my test input as a demo. That may make it easier to understand. I am struggling to figure out a good way to explain this. Somehow the wires in my brain seem to be crossing all the time and what comes out is more convoluted than I want to.

doc/src/pair_lj_sphere.rst
jtclemm: > Feel free to experiment. I am attaching my test example.
> [in.melt-sphere-random.txt](https://github.com/lammps/lammps/files/11145084/in.melt-sphere-random.txt)

Thank you for sharing a script. I don't think multi will perform well for this example since radii are uniformly distributed. The primary savings is avoiding searching a large stencil for lots of little particles. But if you have the same number of small and large atoms, this isn't an obstacle and multi may not be necessary.

Alternatively, I grabbed the power law data file from one of the  multi examples which has far more small than large particles. I had to reduce the LJ cutoff to fit in the box size (maybe it could be increased again with a replicate command), but running multi with 3 types had a neighbor list build time of 0.75 seconds vs the default 280 seconds. So this pair style will work with multi if one uses atom types to create collections (to bin by radii will be a future task). 

Here's the input script for reference: 
[input.txt](https://github.com/lammps/lammps/files/11151220/input.txt)

doc/src/variable.rst
jtclemm, line 1: Not an important detail, but I think input commands generally use the diameter instead of the radius (e.g. set, read_data, fix adapt). However, other commands take both (e.g. dump, fix store/state, etc.) and personally I prefer working with the radius anyway.

akohlmey: I would have used diameter for consistency, but since the functionality uses pointers to internal data and internally the info is stored as radius, there was not much of a choice.

Modifying the Variable class was my only second choice, anyway, since I initially tried with compute property/atom, but that needs to be "consumed and stored" (e.g. with fix store/state 0) and for that I need to have the pair style already set up and do a "run 0" command.

doc/src/pair_lj_sphere.rst
akohlmey: > I don't think multi will perform well for this example since radii are uniformly distributed.

I've modified it and the corresponding example in the docs to create a bimodal distribution. Now "multi" is about 30% faster than "bin".
snap

doc/src/pair_lj_sphere.rst
```diff
@@ -0,0 +1,188 @@
...
+
+This is the same potential function as used by the :doc:`lj/cut
+<pair_lj>` pair style, but the :math:`\sigma_{ij}` parameter is not set
+as a per-type parameter via the :doc:`pair_coeff command <pair_coeff>`,
+but taken from the per-atom diameter attribute of :doc:`atom_style
+sphere <atom_style>`.  The individual value of :math:`\sigma_{ij}` is
+computed from the diameters of the two atoms using the mixing rule for
+pair coefficients as set by the :doc:`pair_modify mix <pair_modify>`
+command (defaults to geometric mixing).  The cutoff is not specified as
+a distance, but as ratio that is internally multiplied with
```
sjplimp: with --> by

sjplimp:
@akohlmey @jtclemm I simplified the pair lj/sphere doc page section that has cautionary performance notes for the user.  See if you like the new version or it misses anything important?

Also, I am thinking the mixing default for this pair style should be arithmetic instead of geometric.  So that sigma_ij = half of sigma_i + sigma_j.  That makes the most sense to me.  What do you think?

akohlmey:
> Also, I am thinking the mixing default for this pair style should be arithmetic instead of geometric. So that sigma_ij = half of sigma_i + sigma_j. That makes the most sense to me. What do you think?

Arithmetic mixing makes more sense for "hard" objects, geometric more for "soft" objects. A large "soft" object is "softer" than a small one (the potential less steep) and thus the small object can come closer. 

Also, it will be more predictable, when the pair style uses the same default rule as lj/cut.

jtclemm:
Thinking a bit more about pair styles that extend beyond a sum of radii, I realize we may need a new flavor of the npair class for this to work. The default npair class checks `rsq <= cutneighsq[itype][jtype]` while the size-variety of npair classes checks `rsq <= (radsum+skin)*(radsum+skin)`. Neither variety depends on both atom radii and atom type. LJ/sphere uses the size-variety so it'd require a skin larger than the maximum cohesive distance. Similar for the JKR option in pair granular.

This problem might not need to be resolved for this PR, but could explain why performance on multi was lacking. If the skin distance is large, small particles will search a large distance regardless of their actual interaction size so multi doesn't do much. If the skin distance is small, large particles will omit interactions beyond that skin distance.

I briefly emailed with Steve, and if this is indeed an issue, I think the only solution is to create a new series of npair classes, which would be much less tedious using PR https://github.com/lammps/lammps/pull/3430.

sjplimp (approved):
@akohlmey If you are fine with the new paragraphs in the lj/cut/sphere doc page, I am fine wtih merging this PR as-is.  Suggestions from @jtclemm for new neighbor list options can wait for a new PR.

akohlmey:
@sjplimp I've renamed lj/sphere to lj/cut/sphere and implemented lj/expand/sphere. There were some minor tweaks to the docs. I added a couple more details and some estimated performance improvements from a few test runs.

sjplimp (approved):

PR #3718, opened 2023-03-30
Cleanup internally created group in fix gcmc

sjplimp:
**Summary**

Add code in destructor to delete the exclusion group created by fix gcmc.  Also the rotation group.  Ditto for fix widom.

**Related Issue(s)**

N/A

**Author(s)**

Steve

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

**Post Submission Checklist**


sjplimp:
@akohlmey Realized this did not require a new method in group.cpp b/c the group has a custom name - it is not the group specified in the fix gcmc input script command - and thus not checked for by the Group class delete operation.

evoyiatzis:
@sjplimp fix widom is *heavily* based on fix gcmc. It worths checking if this modification should be propagated to fix widom as well.

sjplimp:
@evoyiatzis good catch - I added the same logic to fix widom

akohlmey (approved):

PR #1394, opened 2019-03-28
Kokkos exchange comm for fixes 

valleymouth:
**Summary**

This pull request implements exchange communication for fixes on Kokkos. This is a particular issue when using FixNeighHistory, which would fallback to classic communication. We expect to improve performance when using cuda-aware MPI implementations due to reduced number of copies to host memory.

This also adds support for exchange on device for `fix qeq/reaxff/kk` and `fix shake/kk`

**Related Issues**

No issue related to this pull request.

**Author(s)**

Denis Taniguchi (denis.taniguchi@gmail.com) - Newcastle University
Stan Moore (SNL)

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

No backward compatibility issues.

**Implementation Notes**

Major changes can be seen in CommKokkos::exchange_device. After all AtomVec exchanges are performed we loop over the fixes flagged as extra_grow. We call pack_exchange_kokkos/unpack_exchange_kokkos in each of them. Each fix has a separate call to MPI_Irecv/MPI_Send/MPI_Wait. Since there is no guarantee that the unpack in fixes will happen in the same order as in AtomVec (remember that we rely on atomic operations to define the index of the new incoming atom in AtomVec) we need a new interface that returns the indices assigned to each unpacked atom. That was included in AtomVecKokkos with empty default implementation (returning 0). Those indices are used while unpacking data for fixes ensuring that we correctly match the info in AtomVec with info in fixes.
This pull request include implementations for AtomVecSphereKokkos, FixNeighHistoryKokkos, and FixWallGranKokkos already compatible with the new feature in CommKokkos.
If you use AtomVecs not compatible with exchange comm for fixes we fallback to classic comm. This is verified using a new unpack_exchange_indices_flag member variable in AtomVecKokkos. If you wish to make your AtomVec* compatible with fixes exchange, set unpack_exchange_indices_flag to 1 and implement:
```
int unpack_exchange_kokkos(
  DAT::tdual_xfloat_2d &k_buf,
  DAT::tdual_int_1d &indices,
  int nrecv, int nlocal, int dim,
  X_FLOAT lo, X_FLOAT hi,
  ExecutionSpace `space);
```
Fixes must provide:
```
virtual int pack_exchange_kokkos(const int &nsend,DAT::tdual_xfloat_2d &buf,
[...]

PR #3719, opened 2023-03-30
Refactor fix shake to use `closest_image` instead of `minimum_image_once`

stanmoore1:
**Summary**

The order of ghost atoms in a single swap is random for Kokkos border comm. However the `atom_map` assumes that the closest ghost images have the lowest local atom id, which can cause issues with fix shake when it uses `domain->minimum_image_once`. This PR uses the more robust `domain->closest_image` function that is used elsewhere for this purpose. The closest images are also stored for each reneighbor, which saves a little computation since the minimum image calculation doesn't need to be done every step.

**Related Issue(s)**

#1394 

**Author(s)**

Stan Moore (SNL), discussed with @sjplimp 

**Licensing**

**Backward Compatibility**

Yes

### DIFF
--- a/src/KOKKOS/fix_shake_kokkos.cpp
+++ b/src/KOKKOS/fix_shake_kokkos.cpp
@@ -143,6 +143,8 @@ FixShakeKokkos<DeviceType>::~FixShakeKokkos()
   memoryKK->destroy_kokkos(k_shake_atom,shake_atom);
   memoryKK->destroy_kokkos(k_shake_type,shake_type);
   memoryKK->destroy_kokkos(k_xshake,xshake);
+  memoryKK->destroy_kokkos(k_list,list);
+  memoryKK->destroy_kokkos(k_closest_list,closest_list);
 
   memoryKK->destroy_kokkos(k_vatom,vatom);
 }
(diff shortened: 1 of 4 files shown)


akohlmey:
@sjplimp The only other code that uses `Domain::minimum_image_once()` is fix mdi/qmmm.
Could that also use `Domain::closest_image()`? Then we could get rid of `Domain::minimum_image_once()` which is a bit of an odd function to begin with.

stanmoore1:
Once I get the CPU version working I will port this to Kokkos

stanmoore1:
This appears to be working correctly now

stanmoore1:
I'm seeing small numerical differences with this PR over time comparing to the old version. They build up gradually over 100 timesteps and are on the same order as running on 1 vs 4 MPI ranks. The shake stats are the basically the same, so I'm very confident this is just roundoff error.

sjplimp:
@stanmoore1 - CPU version looks good to me - made a few cosmetic changes to fix_shake.cpp to clarify code and comments, and to avoid use of nlist,list as local vars in other methods

sjplimp:
@stanmoore1 fix rattle has same logic with minimum_image() - maybe no need to change it unless it is Kokkos-ized ?
@akohlmey fine to use minimum_image() in fix mdi/qmmm - it was sufficient to use minimum_image_once() since it existed

sjplimp:
@akohlmey To clarify fix mdi/qmmm should not use closest_image(), rather minimum_image() instead of minimum_image_once().
These are not ghost atoms.

stanmoore1:
This is all working correctly as far as I can tell. I tested vanilla CPU, Kokkos OpenMP, and Kokkos CUDA.

stanmoore1:
I confirm this fixes the issue I was seeing with PR #1394.

sjplimp (approved):

PR #3721, opened 2023-03-31
Alternative `minimum_image` code

stanmoore1:
**Summary**

The code for `minimum_image` seems a little complex with the while loop, and is only good for 16 box lengths. Here is alternative code for consideration, which is good for up to 2 billion box lengths in a single step 😄 Could go even further with a 64 bit integer but seems unnecessary.

**Related Issue(s)**

None

**Author(s)**

Stan Moore (SNL)

**Licensing**

**Backward Compatibility**

Yes

### DIFF
--- a/src/domain.cpp
+++ b/src/domain.cpp
@@ -971,7 +971,7 @@ void Domain::subbox_too_small_check(double thresh)
    minimum image convention in periodic dimensions
    use 1/2 of box size as test
    for triclinic, also add/subtract tilt factors in other dims as needed
-   changed "if" to "while" to enable distance to
+   allow multiple box lengths to enable distance to
      far-away ghost atom returned by atom->map() to be wrapped back into box
      could be problem for looking up atom IDs when cutoff > boxsize
    this should not be used if atom has moved infinitely far outside box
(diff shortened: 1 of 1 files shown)


stanmoore1:
@akohlmey @sjplimp I'm not saying we should necessarily merge this, but wanted to get your thoughts on this.

akohlmey:
@stanmoore1 you have to know the history of this function to understand why it is the way it has become.

The code was initially different (I think it used a modulus) and loops were used to make it faster and more reliable by avoiding divisions. For a well behaved simulation, there are only 1 or 2 trips of the loop.

But then we added the limit because there was an increasing number of reports of LAMMPS "getting stuck" without any indication of where and why. Those would have crazy coordinates (like > 1.0e100 or NaN) where the loop would not work due to the finite precision of floating point math. So the limit is a good thing because it reports bogus simulations sooner.

Bottom line: is using the loops in the current format (still) faster than the alternatives with a division or modulus?

stanmoore1:
The source of all absolute truth suggests this is the fastest way for x86-64: https://en.wikipedia.org/wiki/Periodic_boundary_conditions. But I haven't measured yet.

stanmoore1:
Though in the Wikipedia description they are storing `1/xprd`, etc. and only update it when the box length changes to avoid division.

stanmoore1:
Closing for now.

PR #3724, opened 2023-04-03
Enable fix rigid to support 2d enforcement internally

### LINKED ISSUE #2459: [BUG] Drift of z coordinate when using fix rigid + Langevin thermostat in 2d simulation
**Summary**

When using the fix rigid, rigid/small, rigid/nve and rigid/nve/small + Langevin thermostat, the z coordinate does not remain =0 in 2d simulations.

LAMMPS development version (master branch) patch_22Oct2020-50-g5872b05

Ubuntu 16.04

**Expected Behavior**

Running simulations of (pentagon-shaped) rigid particles in 2d. The particles interact via a cut-and-shifted Lennard-Jones potential with cutoff radius = 2.5.

I've run some tests using fix rigid, rigid/small, rigid/nve and rigid/nve/small + Langevin thermostat in conjunction with fix enforce2d using the latest stable version of LAMMPS (3 March 2020) and found that, as expected, the z coordinate of the particles, which is initially =0, remains =0 during the simulation.

**Actual Behavior**

When using the latest patch version (branch master) to perform the same simulations, I observe that the z coordinate does not remain =0 during the simulation. Qualitatively, one can see that if the simulation is run for for long enough, the particles will begin to rotate, with rotation axis parallel to the z=0 plane, and eventually some atoms will "rotate out" of the plane, usually causing a segmentation fault or similar problems. 

To be more quantitative, I've plotted the root-mean-squared z coordinate of the particles (z_rms) as a function of time, which is shown in the following plot:

z_rms

One can see that:

- For fix rigid and rigid/nve, z_rms initially increases and then seems to fluctuate around some mean value or to increase very slowly. This is probably not due only to rounding-off errors, since when performing the simulation with the stable LAMMPS version z remains exactly =0.

- For fix rigid/small, z_rms increases exponentially.

- For fix rigid/nve/small the simulation could not be terminated due to a segmentation fault (Signal code: Address not mapped).

**Steps to Reproduce**

I include below a folder containing the input script (run.in) and the starting configuration (data).

The starting configuration consists in 4 pentagon-shaped rigid particles arranged in regularly in the z=0 plane.

As suggested in the manual (https://lammps.sandia.gov/doc/Howto_2d.html), I chose a small but finite z dimension for the simulation box, zlo=0.0 zhi=0.1. I have also tried zlo=-0.05, zhi=0.05, with the same results. Decreasing zhi-zlo seems to mitigate the problem (discussion in the section below), but does not solve it.

Each particle is composed by 20 atoms, with 4 atoms per side of the pentagons. The distance between each atom and its neighbors is 0.5. A snapshot of the initial configuration is reported below:

snapshot

- I'll note again that this problem is _not_ present (the z coordinate of all the particles remains exactly =0 during the simulation) in the stable version of LAMMPS (3 Mar 2020), which however had other issues with the Langevin thermostat when used with rigid body particles (https://github.com/lammps/lammps/issues/2288).

- Reducing the thickness of the box in the z direction seems to somewhat mitigate the problem, but it does not solve it, as shown in the plot below:

z_rms_rigid_small

**_Folder with input script and starting configuration_**: [bug.zip](https://github.com/lammps/lammps/files/5451993/bug.zip)

**_Input script_**:

```
variable nsteps			equal 	1e7	
variable Tstart			equal   1.0		
variable Tend         	equal 	1.0	    	
variable thermodump     equal   $(v_nsteps/1e4)		
variable trajdump       equal   $(v_nsteps/100)    	
variable tstep        	equal   0.001   
variable damp   		equal   1.0    	
variable cutoff			equal	2.5		 
variable lseed			equal	2141	
variable vseed			equal	10188	
variable fname 			string 	data		
variable thermofile     string  thermo.dat	
variable simname 		string 	run_Tstart${Tstart}_Tend${Tend}	


units           lj          	
boundary        p p p       	
atom_style     	molecular
dimension		2 		

read_data       ${fname}

neighbor		0.8 bin 		
neigh_modify	every 10 delay 0 check yes
neigh_modify	exclude molecule/intra all 

pair_style	lj/cut ${cutoff}
pair_coeff	* * 1.0 1.0 ${cutoff}
pair_modify shift yes

timestep	${tstep}

variable step	equal step
variable temp	equal temp
variable etot	equal etotal
variable ke	equal ke
variable pe	equal pe
variable press	equal press

thermo_style	custom step etotal ke pe temp press
thermo		${thermodump}

fix 	fix_rigid	all rigid/small molecule langevin ${Tstart} ${Tend} ${damp} ${lseed}
fix		fix_2d 		all enforce2d 				

velocity 	all create ${Tstart} ${vseed} dist gaussian mom yes rot yes 
run 0
velocity	all scale ${Tstart}

fix		fix_print all print ${thermodump} "${step} ${etot} ${ke} ${pe} ${temp} ${press}" file ${thermofile} screen no title "#1:step 2:etot 3:ke 4:pe 5:temp 6:press"

dump		1 all custom ${trajdump} dump/dump.${simname}.*.lammpstrj mol xu yu zu
dump_modify	1 sort id
dump_modify	1 pad 10
dump_modify	1 format line "%d %.8e %.8e %.8e"

run		${nsteps}

unfix fix_print
unfix fix_rigid
unfix fix_2d

write_data      data.${simname}
```

sjplimp:
**Summary**

Add code to fix rigid and fix rigid/small to better enforce 2d motion of rigid bodies.  This involved invoking the enforce2d() method whenever forces or torques are computed.  And also zeroing the x and v of atoms in rigid bodies, since they are set by 3d matrix operations.

As a side issue, the 2d enforcement is not invoked natively within the two fix rigid variants, not by fix enforce2d.
Since fix rigid is the only fix that set a flag for fix enforce2d, the flag is no longer needed.  It was removed from
fix.cpp/h and fix_enforce2d.cpp/h.

**Related Issue(s)**

This fixes #2459

**Author(s)**

Steve

**Licensing**

**Backward Compatibility**

Previous rigid body simulation could produce epsilon size z coords or velocities.  That should no longer be the case.
Epsilon differences in trajectories may result.

**Implementation Notes**

**Post Submission Checklist**


sjplimp:
@akohlmey I tested the bug.zip input script in the issue and the dumpfile now has zero z values.  However that input script may be outdated since the issue discussion several other related topics.  If you have better/newer tests please try them

@stanmoore1 This has small deletions from fix.cpp/h and fix enforce2d.cpp/h.  I did not make the corresponding changes to KOKKOS versions, thinking you would be better to do that.  Any regression tests with rigid bodies should be tested with this new PR.

akohlmey:
@sjplimp Should this fix also be applied to `src/POEMS/fix_poems.cpp` and `src/DPD-SMOOTH/fix_rigid_meso.cpp`?

akohlmey:
I am also seeing issues with some of the OPENMP versions. Will be looking into it in a couple of days when I have more time.

akohlmey (approved):

PR #3729, opened 2023-04-12
Introduce compute reduce minabs and maxabs

GenieTim:
**Summary**

This PR introduces two new keywords for the `compute reduce` command, namely a method to reduce to the maximum absolute value, and one to reduce to the minimum absolute value. This can be used e.g. in conjunction with a `compute bond/local dx dy dz`, or similar, where the to-reduce property can be positive or negative, yet only the absolute maxima or minima are desired.

I needed this for a project of mine, and am happy to give back what I implemented, but if there is no general interest, I do not mind if this PR is closed unmerged.

**Author(s)**

**Licensing**

**Backward Compatibility**

No compatibility break. 

**Implementation Notes**

I have also changed the way the scalar reduction is called. Naïvely, this should lead to a slightly reduced cycle count as there is no more branching in the hot path, though I could not really measure it, I guess the compiler was smart enough. 

**Post Submission Checklist**

- [?] The source code follows the LAMMPS formatting guidelines

### DIFF
--- a/doc/src/compute_reduce.rst
+++ b/doc/src/compute_reduce.rst
@@ -71,12 +71,12 @@ vectors.
 The reduction operation is specified by the *mode* setting.  The *sum*
 option adds the values in the vector into a global total.  The *min*
 or *max* options find the minimum or maximum value across all vector
-values.  The *minabs* or *maxabs* options find the minimum or maximum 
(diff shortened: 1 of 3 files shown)


sjplimp (approved):
@GenieTim Thanks for contributing these new keywords

PR #3736, opened 2023-04-15
Always throw exceptions for errors

### LINKED ISSUE #3678: [Feature Request] Always use the -DLAMMPS_EXCEPTIONS branch and remove the non-exception code path

akohlmey:
**Summary**

This pull request removes processing the -DLAMMPS_EXCEPTIONS define and corresponding CMake variable and always chooses the code paths for the case where the define is enabled. This simplifies code that uses the library interface and makes LAMMPS better behaved if loaded into other, especially GUI programs.

**Related Issue(s)**

Implements and thus closes #3678 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

The exceptions code path has been in use a long time.

**Implementation Notes**

This mostly removes the define itself. Functions that query whether exceptions are enabled remain, but now will always report that it is. In addition, the unit test tree is purged of workarounds to skip certain tests that require exceptions enabled.

**Post Submission Checklist**


rbberger (approved):
`Library_config.html` should probably now mark `lammps_config_has_exceptions()` as deprecated, so it can be removed in the future. Otherwise, LGTM.

PR #3739, opened 2023-04-18
Fix compile error with latest Kokkos `develop` branch

stanmoore1:
**Summary**

Fix compile error with latest Kokkos lib, reported by Nick Curtis (AMD), @arghdos 

**Related Issue(s)**

#3670

**Author(s)**

Stan Moore (SNL)

**Licensing**

**Backward Compatibility**

Yes

### DIFF
--- a/src/KOKKOS/atom_kokkos.h
+++ b/src/KOKKOS/atom_kokkos.h
@@ -86,7 +86,7 @@ class AtomKokkos : public Atom {
 
   template<class DeviceType>
   KOKKOS_INLINE_FUNCTION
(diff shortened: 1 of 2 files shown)


stanmoore1:
According to Christian, this may be an unintended change in Kokkos, see https://github.com/kokkos/kokkos/issues/6066.

stanmoore1:
It sounds like this may not be fixed in Kokkos for a while (if ever).

sjplimp (approved):

akohlmey (approved):

akohlmey:
> It sounds like this may not be fixed in Kokkos for a while (if ever).

This change looks like a good change, regardless. If you can use a const reference it is always preferred over a non-const reference. This is one of the things that clang-tidy suggests, for example, if it sees an opportunity.

PR #3758, opened 2023-04-28
Fuse some Kokkos kernels to reduce launch latency for small systems

stanmoore1:
**Summary**

Fuse some Kokkos kernels to reduce launch latency for small systems. Gives ~20% speedup for 1k atom LJ benchmark on a single V100 GPU.

**Related Issue(s)**

None

**Author(s)**

Stan Moore (SNL)

**Licensing**

**Backward Compatibility**

Yes


stanmoore1:
Running this through Kokkos regression testing again.

stanmoore1:
@akohlmey all the Kokkos OpenMP regression tests passed, but the Kokkos OpenMP unit tests are failing. Is it possible that there is some assumption in the unit tests that would break if kernels are fused?

stanmoore1:
Kokkos regression tests all pass now.

akohlmey (approved):

PR #3761, opened 2023-05-01
New compute count/type command

sjplimp:
**Summary**

Add a compute count/type command which counts either all atoms by atom type or all bonds by bond type.  The latter
includes a count of broken bonds.  The output of the compute is a global vector of counts, and a global scalar of broken bond counts.  Useful for simulations where atoms are added/deleted or bonds are created/broken.  The doc page lists such commands.

**Related Issue(s)**

N/A

**Author(s)**

Steve

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

**Post Submission Checklist**


sjplimp:
@akohlmey Added a setting in lmptype.h for the largest integer which can be accurately represented in a double (along with all integers smaller than the largest).  I think it is useful for checking for overflow when integers are copied into doubles, e.g. by computes and fixes.  Maybe it will be useful other places in the code besides this compute.  This is distinct from the ubuf() method which does not have this problem.

sjplimp:
@jrgissing Hi Jake - if you have any feedback on this new compute, let me know.  I am thinking it is useful for fix bond/react models?  Am I correct than fix bond/react does not ever set a broken bond type to zero?

doc/src/compute_count_type.rst
```diff
@@ -0,0 +1,88 @@
...
+command.  They do not refer to bonds defined on-the-fly by bond-order
+or reactive pair styles.
+
+These commands can create and break toplogical bonds:
+
+* :doc:`fix bond/react <fix_bond_react>`
+* :doc:`fix bond/create <fix_bond_create>`
+* :doc:`fix bond/break <fix_bond_break>`
+* :doc:`bond_style quartic <bond_quartic>`
+* :doc:`BPM package <Howto_bpm>` bond styles
```
jtclemm, line 68: I'm pretty sure fix bond/break deletes bonds by shifting bond entries over instead of setting the bond type to zero, so computing the number of broken bonds won't work with that fix. I want to say fix bond/react works the same way, but I'd have to check. 

The text in Howto broken bonds might need to be clarified to emphasize this point too.

doc/src/compute_count_type.rst
```diff
@@ -0,0 +1,88 @@
...
+
+   compute 1 all count/type atom
+   compute 1 flowmols count/type bond
+
+Description
+"""""""""""
+
+Define a computation that counts the current number of atoms by atom
+type or the number of bonds by bond type.  The latter can be useful in
+reactive simulations where bonds are broken or created.
```
jtclemm: Are there any fixes or commands that could lead to an unknown number of atoms of a given type in LAMMPS? Maybe fix evaporation? I think fix pour/deposit could do this too if it fails insertions.

jrgissing:
> @jrgissing Hi Jake - if you have any feedback on this new compute, let me know. I am thinking it is useful for fix bond/react models? Am I correct than fix bond/react does not ever set a broken bond type to zero?

Thanks, yes I think it could be useful for bond/react. Is this functionality not already available with per-atom variables and the group count() function? You are right, typically bond/react deletes bonds rather than setting their types to zero. However, as a general tool, it can do either one.

sjplimp:
@jrgissing yes, if you put each type in its own group, you could use one variable per type to do a count(group) with a variable.
It would be more inefficient to calculuate that way.  The new compute is really just a shortcut to get all the counts at once
so a user could log them however they wished, e.g, to screen or file.

doc/src/compute_count_type.rst
sjplimp: @jtclemm Good catch on fix bond/break.  I updated both the Howto broken bond and compute count/type pages.  See if the changes satisfy your concerns?

jtclemm: I think the new edits to the broken bonds doc page look good. 

I guess if users want to track the number of bonds broken by fix bond/break, they can just rely on the output from its global vector. Maybe at some point in the future, if requested, this can be generalized to work across bond types.

~Any thoughts on analogously listing relevant fixes which change the number of atoms as possible applications for the atom mode? Evaporation, pour, etc.~ , just saw this was added.

akohlmey:
@sjplimp I've noticed that this command is not compatible with the `delete_bonds` command, since this "only" negates the bond type (same as fix shake) so it can be restored. This is a bit of an inconsistency. Should we perhaps differentiate between broken and disabled bonds by setting an additional bit (2^31)?
This may also be useful for other commands that break bonds, since this way, it can still be seen what type the bond was when it was broken and it could also be restored with the delete_bonds command (although I am not sure that that would be useful for anything).

akohlmey:
@sjplimp Why did you stop at "bond"? Shouldn't this be useful for angles, dihedrals, and impropers, too?

sjplimp:
Re: delete_bonds and negative types (an option, they can also be fully deleted). I think this compute should treat them the same as fix shake, i.e. as bonds that still exist. And just document more fully what it is doing.  If we want to change how bonds are flagged by fix shake vs delete_bonds, etc we could do that at some point, but I'd prefer that to be a separate PR.

Re: angles, etc - no one who gave me feedback seemed interested in those options, so I didn't add them.  Easy to do though,
either now or in the future.

sjplimp:
@akohlmey I think I have addressed both your comments with added doc page info and angle/etc support in the src code

jtclemm:
I made some changes to the doc pages clarifying that BPM bond styles operate like the quartic bond style. 

Should the count angle etc. methods either skip type zero or issue an error to avoid accessing count[-1]? Even if there are currently no commands that would set the angle type to zero, I could foresee the addition of breakable angle types in the future.

akohlmey:
@sjplimp Thanks for the changes, but please note that I had to fix compilation with -DLAMMPS_BIGBIG now for the second time. Perhaps we should generally change from `sometype *var = class->var;` to `auto var = class->var;` for those cached member variables (except for cases with type casts). Made a few more cosmetic edits and applied clang-format.

@jtclemm I believe in "better safe than sorry" and just added the required if statements so that itype==0 is skipped.

akohlmey (approved):

PR #3764, opened 2023-05-02
informational thermo column titles

jrgissing:
**Summary**

Add 'auto' keyword to 'thermo_modify colname'. This option replaces indexed computes and fixes in thermo output column titles with descriptions of what is being outputted, for more self-descriptive logfiles. For example, when using 'fix bond/react', instead of 'f_my_rxns[1]' as the thermo output column title, the column title is the name of the reaction, such as 'f_my_rxns:Diels-Alder'. 

Fixes with Global Output Values
- [x] Thermobarostats derived from FixNH
- [x] Fix bond/react

Computes with Global Output Values
- [x] Compute reduce

Additional fixes and computes with global output listed in the attached files.

[select_fixes_with_global_output.txt](https://github.com/lammps/lammps/files/12075651/select_fixes_with_global_output.txt)

**Related Issue(s)**

none

**Author(s)**

JG

**Licensing**

**Backward Compatibility**

only possible concern is hard-coded post-processing tools that assume previous non-descriptive titles

**Implementation Notes**

**Post Submission Checklist**


akohlmey:
I see a few cases where this could be useful, e.g. for compute reduce where it could say "sum(xxx)" or "min(xxx)", but for the most part, the manual option with fix_modify is fully sufficient and preferable. It would be preferable, if you could just call thermo_modify from your fix, if needed, but that would probably need adding some functions there.

What bothers me a lot is that this change modifies long established default behavior and that there is no option to recover that old behavior. Some post-processing tools people have written depend on the default behavior and I suspect it may also break PyLammps.

jrgissing:
I see your points. Another specific example I had in mind was fix nvt/npt, where the output can change based on user settings and is somewhat indecipherable when the variable names are not included in the column title. In cases like these, it seems preferable to make the self-describing version the default. If we proceed in this way for certain fixes/computes, I can include an option to recover old behavior.

jtclemm:
There was a short discussion about a related idea on slack a year or so ago revolving around ambiguous dump column names in compute X/local and property/X commands. I don't think an ideal solution was found, but Axel added the option to manually rename individual columns in thermo/dump files in #3197.

If this PR identifies a good way to automatically generate descriptive thermo column names, a similar option might be helpful for dump files.

stanmoore1:
@jrgissing do you have more changes for this PR, or is it ready for review?

jrgissing:
> @jrgissing do you have more changes for this PR, or is it ready for review?

This PR is still is progress, and will need significant additions based on requests from @sjplimp

akohlmey:
@jrgissing this is currently our oldest pending pull request. Do you have any comments on the status?
I am particularly concerned about these kinds of useful-but-potentially-disruptive changes and would like to have them merged early in the development cycle between stable releases.

jrgissing:
In my opinion, this PR could be merged as is (after I add docs and re-review it), with the idea that it establishes the framework for 'informational thermo column titles', and that support for specific fix/computes can be added as requested. Note that Steve requested a significant set of enhancements to this idea (I could clean up and post our offline discussion here, for reference), but those additional features will likely take me a while to get around to

doc/src/compute_reduce.rst
```diff
@@ -189,7 +189,7 @@ with maximum stretch, you can do it as follows:
 
    compute 1 all property/local batom1 batom2
    compute 2 all bond/local dist
-   compute 3 all reduce max c_1[1] c_1[2] c_2 replace 1 3 replace 2 3
+   compute 3 all reduce max c_1[1] c_1[2] c_2 replace 1 3 replace 2 3 inputs local
```
jtclemm: Admittedly, I struggled parsing how this example led to 'c_3:c_1[1]<-max(c_2)'. Since this example is supposed to help users follow the syntax, could compute 1 & 2 be renamed something more distinct like myProp and myBond? Or even A and B?

Between compute names, indices, and replacement vecs, there are a lot of 1, 2, and 3 integers. The above example with myPress was much clearer.

jrgissing: nice idea! done.

jtclemm (approved):

sjplimp:
I'll take a look at this to refresh myself on the current status

doc/src/thermo_modify.rst
```diff
@@ -158,14 +158,20 @@ block ("yaml").  This modify option overrides the *one*, *multi*, or
...
-string* replaces the default text with the provided string.  *ID* can be
-a positive integer when it represents the column number counting from
-the left, a negative integer when it represents the column number from
-the right (i.e., :math:`-1` is the last column/keyword), or a thermo keyword
-(or compute, fix, property, or variable reference) and then it replaces the
-string for that specific thermo keyword.
+The *colname* keyword can be used to change the default header keyword for
+a column or field of thermodynamic output.  The column names can either be
+manually set by the user, or automatically generated for certain fixes and
+computes.  The setting for *ID string* replaces the default text with the
```
sjplimp: make "The setting for ID ..." be a new paragraph

doc/src/thermo_modify.rst
```diff
@@ -158,14 +158,20 @@ block ("yaml").  This modify option overrides the *one*, *multi*, or
...
+provided string.  *ID* can be a positive integer when it represents the
+column number counting from the left, a negative integer when it represents
+the column number from the right (i.e., :math:`-1` is the last
+column/keyword), or a thermo keyword (or compute, fix, property, or
+variable reference) and then it replaces the string for that specific
+thermo keyword.  With a setting of *auto*, certain fixes or computes will
```
sjplimp: New paragraph for auto

doc/src/thermo_modify.rst
```diff
@@ -158,14 +158,20 @@ block ("yaml").  This modify option overrides the *one*, *multi*, or
...
+generate more descriptive strings as their thermo keywords, which are
+described in the 'output' section of their documentation. Current commands
+that automatically generate descriptive thermo output strings include 'fix
+nvt', 'fix npt', 'fix nph', 'compute reduce', and 'fix bond/react'.
```
sjplimp, line 179: New paragraph to describe what "default setting" does

doc/src/thermo_modify.rst
```diff
@@ -158,14 +158,20 @@ block ("yaml").  This modify option overrides the *one*, *multi*, or
...
```
sjplimp, line 178: Add a new paragraph to explain what "default" setting does

sjplimp, line 178: It is also not clear to me how the 3 options (ID, auto, default) interact with each other.  Does auto or default override all previous column-specific settings for ID ?  If so, that should be emphasized.  If not, then the details should be explained. Possibly auto/default only override column settings for fixes or computes ?

sjplimp (changes requested):
Made a few suggestions for the thermo_modify.rst file.

I'm willing to approve this for merging if this is a stepping-stone to some of the further enhancements we talked about earlier?  If these changes will be discarded (or completely refactored) for the bigger picture, then I'm not sure it makes sense to merge this now?

jrgissing:
Yes, I believe this PR is a stepping stone that will not need to be refactored. For reference, the further enhancements you requested include adding similar support to dump_modify, and to other fixes (e.g., ave/atom), computes, or variables that could benefit from this feature.

sjplimp (approved):

PR #3767, opened 2023-05-04
Vector-style variable enhancements

sjplimp:
**Summary**

Enable vector-style variables to be initialized with the following syntax:
      variable foo vector [1,4,6.5,13]

Also enable their output in the same syntax via the print and fix print commands.  E.g.
print "MyVec: ${foo}"
should give
MyVec: [1,4,6.5,13]

Vector variables initialized this way are "static".  Vector variables defined with a formula
(similar to equal- and atom-style variables) are "dynamic"; they are re-computed when referenced.

Static vector variables never need to be recomputed, but can be used in any context a dynamic vector variable can be.
Dynamic vector variables can also be output in the same syntax.
 
**Related Issue(s)**

N/A

**Author(s)**

Steve

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

**Post Submission Checklist**


sjplimp:
@akohlmey I am unclear if there are formatting options for individual vector elements (when output) which could be or should be supported and documented.

sjplimp:
@akohlmey Otherwise this is ready for you to look over.

akohlmey:
@sjplimp I significantly simplified the vector input parser using the Tokenizer class and added a few unit tests. Otherwise looking OK. 

P.S.: did the usual whitespace, and spelling fixes, too. Please consider making my life easier by doing that.
Also, you broke a couple of unit tests for variables.

akohlmey (approved):

PR #3770, opened 2023-05-08

akohlmey:
**Summary**

This pull request collects multiple small bug fixes and updates.

**Related Issue(s)**

Addresses issues mentioned in https://matsci.org/t/bug-in-commbrick-exchange-variable-function-causing-loss-of-intermediate-results-in-buf-recv/48419 and https://matsci.org/t/timer-timeout-command-in-lammps/48391

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues

**Implementation Notes**

- remove potentially buggy and long unused communication API
- fix timer timout output formatting issue outputting only a subset of the remaining time if it was over 24 hours
- small documentation fixes (spelling and clarifications)
- add note about using very small factor to include pairs in neighbor list that would be excluded with 0.0
- make LAMMPS Swig interface file compatible with Swig 4.1
- Bump python version requirement for building the manual to version 3.8
- When setting a custom python version via -D Python_EXECUTABLE also use it when looking for python3, if a python 3 version.

**Post Submission Checklist**

### DIFF
--- a/cmake/Modules/CodingStandard.cmake
+++ b/cmake/Modules/CodingStandard.cmake
@@ -5,6 +5,10 @@ if(CMAKE_VERSION VERSION_LESS 3.12)
         set(Python3_VERSION ${PYTHON_VERSION_STRING})
     endif()
 else()
+    # use default (or custom) Python executable, if version is sufficient
+    if(Python_VERSION VERSION_GREATER_EQUAL 3.5)
+      set(Python3_EXECUTABLE ${Python_EXECUTABLE})
+    endif()
     find_package(Python3 COMPONENTS Interpreter QUIET)
 endif()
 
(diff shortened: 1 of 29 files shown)


sjplimp (changes requested):
Re: exchange_variable() -- did some other method supercede this older one?  I don't see that it is called.  However it is related to Comm::exchange() and I can imagine it might be useful in some contexts for exchange of per-proc info with touching neighbor domains.

akohlmey:
> Re: exchange_variable() -- did some other method supercede this older one? I don't see that it is called. However it is related to Comm::exchange() and I can imagine it might be useful in some contexts for exchange of per-proc info with touching neighbor domains.

I haven't found any evidence for it being used for a very long time in the git history. There also was a report that it is likely buggy: https://matsci.org/t/bug-in-commbrick-exchange-variable-function-causing-loss-of-intermediate-results-in-buf-recv/48419

sjplimp:
ok - I'm fine with removing it - we would need a CommTiled version of it as well

sjplimp (approved):

PR #3771, opened 2023-05-08
Change how variables check if computes are current

sjplimp:
**Summary**

The aim of this PR is to eliminate error messages when variables are evaluated saying that a compute "is not current".  It does this by more aggressively evaluating computes when they are needed in between runs.  There are still two cases when a variable evaluation (which requires a compute either directly or indirectly via a thermo keyword) when errors may occur.

(a) evaluating such a variable before the first run - an error is generated
(b) evaluating such a variable between runs, but after also changing the system (after the previous run, but before the variable evaluation), e.g. deleting atoms - the evaluated variable value may be incorrect

Both of these special cases are explained on the variable doc page.  The solution to both is to perfrom a "run 0" command
before evaluating the variable.

**Related Issue(s)**

N/A

**Author(s)**

Steve

**Licensing**

**Backward Compatibility**

I think the only effect of this should be to not trigger error message that were triggered previously.  The two cases mentioned above (error and incorrect) were also the behavior previously.

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/doc/src/variable.rst
+++ b/doc/src/variable.rst
@@ -1285,18 +1285,19 @@ with a leading $ sign (e.g. $x or ${abc}) versus with a leading "v\_"
 (e.g. v_x or v_abc).  The former can be used in any input script
 command, including a variable command.  The input script parser
 evaluates the reference variable immediately and substitutes its value
-into the command.  As explained on the :doc:`Commands parse <Commands_parse>` doc page, you can also use un-named
-"immediate" variables for this purpose.  For example, a string like
-this $((xlo+xhi)/2+sqrt(v_area)) in an input script command evaluates
-the string between the parenthesis as an equal-style variable formula.
+into the command.  As explained on the :doc:`Commands parse
+<Commands_parse>` doc page, you can also use un-named "immediate"
+variables for this purpose.  For example, a string like this
+$((xlo+xhi)/2+sqrt(v_area)) in an input script command evaluates the
+string between the parenthesis as an equal-style variable formula.
 
 Referencing a variable with a leading "v\_" is an optional or required
-kind of argument for some commands (e.g. the :doc:`fix ave/chunk <fix_ave_chunk>` or :doc:`dump custom <dump>` or
-:doc:`thermo_style <thermo_style>` commands) if you wish it to evaluate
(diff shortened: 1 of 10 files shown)


sjplimp:
@akohlmey I simplified the variable doc page to explain what I think the remaining issues are.  I also noticed there are other good error checks already in Thermo and the energy/virial dependent computes to ensure needed values have been tallied when they are requested.  I am hoping this PR reduces the "compute is not current" errors uses see, without introducing any new problems.

akohlmey:
> I am hoping this PR reduces the "compute is not current" errors uses see, without introducing any new problems.

For now it broke some of the unit tests. Probably some of the "death tests". Will have to check, when I have some more time.

akohlmey:
@sjplimp the unit test failure was with the `reset_atoms image` command, which uses variables internally and does a manual init of the simulation. I have implemented a workaround in commit: https://github.com/lammps/lammps/pull/3771/commits/d2361ffe7d3460a010802a8fddbfe5c9f2d3e7f3

Please have a look and modify as needed, if you have a better idea. Here is a modified rhodo input that trigges the issue without my workaround.

[in.rhodo-reset-images.txt](https://github.com/lammps/lammps/files/11446193/in.rhodo-reset-images.txt)

```
[...]
  update: every = 1 steps, delay = 5 steps, check = yes
  max neighbors/atom: 2000, page size: 100000
  master list distance cutoff = 12
  ghost atom cutoff = 12
  binsize = 6, bins = 10 13 13
  2 neighbor lists, perpetual/occasional/extra = 1 1 0
  (1) pair lj/charmm/coul/long, perpetual
      attributes: half, newton on
      pair build: half/bin/newton
      stencil: half/bin/3d
      bin: standard
  (2) compute stress/cartesian, occasional, copy from (1)
      attributes: half, newton on
      pair build: copy
      stencil: none
      bin: none
ERROR: Variable ix_r_i_f: Variable formula compute cannot be invoked before first run (src/variable.cpp:1626)
Last command: reset_atoms image all
```

sjplimp:
@akohlmey altered reset_atoms_image.cpp as we discussed - the rhodo test script you posted works

akohlmey (approved):

PR #3776, opened 2023-05-10
Fixed read-in of initial electron temperature file

inflowencer:
**Summary**

Fixed read-in of initial electron temperature file for the `fix ttm/mod` command, which throws an error when providing the `infile` argument that otherwise runs when using the `fix ttm` command.

**Related Issue(s)**

No related issues.

**Author(s)**

Maximilian Maigler, Stanford SLAC (maximilian.maigler@unibw.de).

**Licensing**

**Backward Compatibility**

Backward compatibility is ensured.

**Implementation Notes**

Solution based on the code of `src/EXTRA-FIX/ttm.cpp`. It seems the `src/EXTRA-FIX/ttm_mod.cpp` file was outdated, and a `- 1`  was missing for correctness of the initialized integer counters.

**Post Submission Checklist**

### DIFF
--- a/src/EXTRA-FIX/fix_ttm_grid.cpp
+++ b/src/EXTRA-FIX/fix_ttm_grid.cpp
@@ -281,7 +281,7 @@ void FixTTMGrid::read_electron_temperatures(const std::string &filename)
   }
 
   // read the file
-  // Grid3d::read_file() calls back to read_grid_lines() with chunks of lines
+  // Grid3d::read_file() calls back to unpack_read_grid() with chunks of lines
 
   grid->read_file(Grid3d::FIX,this,fp,CHUNK,MAXLINE);
 
@@ -419,7 +419,7 @@ void FixTTMGrid::write_restart_file(const char *file)
   }
 
   // write file
-  // Grid3d::write_file() calls back to pack_write_file() and unpack_write_file()
+  // Grid3d::write_file() calls back to pack_write_grid() and unpack_write_grid()
 
   grid->write_file(Grid3d::FIX,this,0,1,sizeof(double), MPI_DOUBLE);
 
--- a/src/EXTRA-FIX/fix_ttm_mod.cpp
+++ b/src/EXTRA-FIX/fix_ttm_mod.cpp
@@ -658,7 +658,7 @@ void FixTTMMod::write_electron_temperatures(const std::string &filename)
       for (iz = 0; iz < nzgrid; iz++) {
         if (movsur == 1 && T_electron[ix][iy][iz] == 0.0)
           T_electron[ix][iy][iz] = electron_temperature_min;
-        fprintf(fp,"%d %d %d %20.16g\n",ix,iy,iz,T_electron[ix][iy][iz]);
+        fprintf(fp,"%d %d %d %20.16g\n",ix+1,iy+1,iz+1,T_electron[ix][iy][iz]);
       }
 
   fclose(fp);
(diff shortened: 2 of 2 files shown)


kawai125:
I think the "bug" exists in manual, not in code.

Currently, `read_electron_temperatures()` and `write_electron_temperatures()` in `fix_ttm.cpp` and `fix_ttm_mod.cpp` using grid index range as [0, n-1].
So the example of `infile` in [manual](https://docs.lammps.org/fix_ttm.html) must be written as follows,

The ix,iy,iz are grid point indices ranging from 0 to Nxyz-1 inclusive in each dimension. The lines can appear in any order. For example, the initial electronic temperatures on a 1 by 2 by 3 grid could be specified in the file as follows:
```
# UNITS: metal COMMENT: initial electron temperature
0 0 0 1.0
0 0 1 1.0
0 0 2 1.0
0 1 0 2.0
0 1 1 2.0
0 1 2 2.0
```

akohlmey:
> The ix,iy,iz are grid point indices ranging from 0 to Nxyz-1 inclusive in each dimension.

@kawai125 please note that LAMMPS generally uses Fortran style 1-based indexing in its input, even if the data is then converted to 0-based indexing internally (this is mostly for historical reasons. LAMMPS started as a Fortran program). So adjusting the code and not the documentation would be more consistent with the rest of LAMMPS.

inflowencer:
I would like to add 2 things:

1. The fix is based on the implementation of `fix ttm` (see [lines 480-483](https://github.com/lammps/lammps/blob/fceff845c7ae434ce8bfe1ca6adf1313e67af2dd/src/EXTRA-FIX/fix_ttm.cpp#LL480C9-L483C46), as I noticed that the same input file works when using `fix ttm`
2. The output file of the electron temperature $T_e$ written by LAMMPS actually starts with `0 0 0` and not `1 1 1` (the input file is supposed to be starting with `1 1 1`)

I propose to either accept the pull request and change the function for writing the output file or decline the request, change the code of `fix ttm` (i.e. removing the `-1`) and change the manual.

sjplimp:
Hi all - I agree with Axel that the LAMMPS standard is to read inputs and write outputs with indices for any vector or array from 1 to N, not 0 to N-1.  This is true for input scripts, dump files, and should also be for auxiliary files like TTM temperatures on a grid (inputs and outputs).  This is for convenience for users who create or read these files.  Internally, of course, the data structures (e.g. a 3d array for TTM)  are allocated as N in a dimension and indexexdby 0 to N-1.

For TTM, this may have become confused when fix ttm/grid was added (by me), which uses a distributed grid (fix ttm and ttm/mod use a global grid, stored by every proc).

Fix ttm/grid uses the new dump grid command for its output, which uses 1 to N indices.  It also reads a file with 1 to N indices, see
FixTTMGrid::unpack_read_grid().

Looking at the current fix ttm, in FixTTM::write_electron_temperatures() it is writing with 1 to N indices.  And ditto
for reading with FixTTM::read_electron_temperatures() with 1 to N indices.

However, for fix ttm/mod, the same two methods as in FixTTM, are reading and writing indices from 0 to N-1.
So that is the only code that I think should be changed.

The doc page for all 3 commands says that indices in the input and output files are 1 to N.  So no change to the doc page

kawai125:
Dear all,
Thank you very much for your comments.
My previous comment was based on my misunderstanding, and the I/O of `fix_ttm.cpp` was consistent with the discription in the manual.

I agree to modify the code accoding to this pull request.
(Also, please make the same modify to line 661 in `fix_ttm_mod.cpp`.)

Sorry about the fuss.

akohlmey:
@sjplimp this should be complete now. Please check it out.

sjplimp (approved):

PR #3778, opened 2023-05-15
Added mliap data structure for message passing models

danstine:
**Summary**
Modified unified ML-IAP and KOKKOS mliap to include a data structure that stores global tags of neighbors

**Related Issue(s)**
None

**Author(s)**
Dylan Anstine, Carnegie Mellon University

**Licensing**

**Backward Compatibility**

**Implementation Notes**

The included data structures "mapped_jatoms" and "k_mapped_jatoms" are similar to "jatoms" and "k_jatoms", but their contents point to the indices of the real atoms in the simulation cell and not the ghost atom indices.

**Post Submission Checklist**

### DIFF
--- a/src/KOKKOS/mliap_data_kokkos.h
+++ b/src/KOKKOS/mliap_data_kokkos.h
@@ -28,23 +28,25 @@
 namespace LAMMPS_NS {
 // clang-format off
 enum {
-  IATOMS_MASK      = 0x00000001,
-  IELEMS_MASK      = 0x00000002,
-  JATOMS_MASK      = 0x00000004,
-  JELEMS_MASK      = 0x00000008,
-  IJ_MASK          = 0x00000010,
-  BETAS_MASK       = 0x00000020,
-  DESCRIPTORS_MASK = 0x00000040,
-  EATOMS_MASK      = 0x00000080,
-  RIJ_MASK         = 0x00000100,
-  GRADFORCE_MASK   = 0x00000200,
-  GRADDESC_MASK    = 0x00000400,
-  NUMNEIGHS_MASK   = 0x00000800,
-  GAMMA_MASK_MASK  = 0x00001000,
-  GAMMA_ROW_MASK   = 0x00002000,
-  GAMMA_COL_MASK   = 0x00004000,
-  PAIR_I_MASK      = 0x00008000,
-  ELEMS_MASK       = 0x00010000,
+  IATOMS_MASK        = 0x00000001,
+  IELEMS_MASK        = 0x00000002,
+  JATOMS_MASK        = 0x00000004,
+  JELEMS_MASK        = 0x00000008,
+  IJ_MASK            = 0x00000010,
+  BETAS_MASK         = 0x00000020,
+  DESCRIPTORS_MASK   = 0x00000040,
+  EATOMS_MASK        = 0x00000080,
+  RIJ_MASK           = 0x00000100,
+  GRADFORCE_MASK     = 0x00000200,
+  GRADDESC_MASK      = 0x00000400,
+  NUMNEIGHS_MASK     = 0x00000800,
+  GAMMA_MASK_MASK    = 0x00001000,
+  GAMMA_ROW_MASK     = 0x00002000,
+  GAMMA_COL_MASK     = 0x00004000,
+  PAIR_I_MASK        = 0x00008000,
+  ELEMS_MASK         = 0x00010000,
+  MAPPED_JATOMS_MASK = 0x00010002,
+
 };
 // clang-format on
 
@@ -67,6 +69,7 @@ template <class DeviceType> class MLIAPDataKokkos : public MLIAPData {
   DAT::tdual_int_1d k_iatoms;           // index of each atom
   DAT::tdual_int_1d k_ielems;           // element of each atom
   DAT::tdual_int_1d k_jatoms;           // index of each neighbor
+  DAT::tdual_int_1d k_mapped_jatoms;    // index of each neighbor mapped to simulation cell
   DAT::tdual_int_1d k_elems;            // element of each atom in or not in the neighborlist
   DAT::tdual_int_1d k_pair_i;           // index of each i atom for each ij pair
   DAT::tdual_int_1d k_jelems;           // element of each neighbor
@@ -126,6 +129,7 @@ public:
     nneigh_max(base.nneigh_max),
     npairs(base.npairs),
     jatoms(base.k_jatoms.d_view.data()),
+    mapped_jatoms(base.k_mapped_jatoms.d_view.data()),
     jelems(base.k_jelems.d_view.data()),
     elems(base.k_elems.d_view.data()),
     rij(base.k_rij.d_view.data()),
@@ -179,6 +183,7 @@ public:
   const int nneigh_max;
   const int npairs;
   int *jatoms;
+  int *mapped_jatoms;
   int *jelems;
   int *elems;
   double *rij;
--- a/src/KOKKOS/mliap_data_kokkos.h
+++ b/src/KOKKOS/mliap_data_kokkos.h
@@ -45,7 +45,7 @@ enum {
   GAMMA_COL_MASK     = 0x00004000,
   PAIR_I_MASK        = 0x00008000,
   ELEMS_MASK         = 0x00010000,
-  MAPPED_JATOMS_MASK = 0x00010002,
+  MAPPED_JATOMS_MASK = 0x00020000,
 
 };
 // clang-format on
--- a/src/KOKKOS/mliap_data_kokkos.h
+++ b/src/KOKKOS/mliap_data_kokkos.h
@@ -45,7 +45,7 @@ enum {
   GAMMA_COL_MASK     = 0x00004000,
   PAIR_I_MASK        = 0x00008000,
   ELEMS_MASK         = 0x00010000,
-  MAPPED_JATOMS_MASK = 0x00020000,
+  JTAGS_MASK = 0x00020000,
 
 };
 // clang-format on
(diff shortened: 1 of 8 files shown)


rohskopf:
Currently we can simply use a larger ghost cutoff for communication with message passing models (thanks @lubbersnick for the tip):
https://github.com/lanl/hippynn/blob/main/examples/lammps/in.mliap.unified.hippynn.Al.ghostneigh#L45-L53

We already have examples that use this method, so it'd be nice to have examples that use these new mapped atoms as well. Do you have an example of a deployed model for testing purposes?

src/KOKKOS/mliap_data_kokkos.h
```diff
@@ -28,23 +28,25 @@
...
+  RIJ_MASK           = 0x00000100,
+  GRADFORCE_MASK     = 0x00000200,
+  GRADDESC_MASK      = 0x00000400,
+  NUMNEIGHS_MASK     = 0x00000800,
+  GAMMA_MASK_MASK    = 0x00001000,
+  GAMMA_ROW_MASK     = 0x00002000,
+  GAMMA_COL_MASK     = 0x00004000,
+  PAIR_I_MASK        = 0x00008000,
+  ELEMS_MASK         = 0x00010000,
+  MAPPED_JATOMS_MASK = 0x00010002,
```
akohlmey: Why is this explicitly defined as `MAPPED_JATOMS_MASK = 0x00010002,`? I would have either expected `MAPPED_JATOMS_MASK = 0x00020000,`? or `MAPPED_JATOMS_MASK = IELEMS_MASK|ELEMS_MASK,`.

athomps:
@danstine The use of the word map in the PR description and the name `mapped_jatoms` is too generic.  LAMMPS already contains at least two different `map` data structures for keeping track of fairly obscure things like chemical elements, `map` is even a member of MLIAPData.  In this case, what you are accessing is not at all obscure, it is `atom->tag[j]`, usually referred to as a global tag or global id.  I suggest expunging all words based on map from the code and the description, and instead use words based on tag.  The new array name should logically be `jtags`, following the pattern of `jelems` and `jatoms`.

athomps:
Also, it is possible that whatever your are using `jtags` for could be accomplished in a different way. Often, when LAMMPS code is accessing global tags, it is breaking the intrinsic non-global nature of LAMMPS.

lubbersnick:
I think I understand how this would work if there is no domain decomposition (serial lammps). But I don't know how it will work if there is domain decomposition. I also don't understand exaclty how this change would affect the partial forces f_ij required for the global stress computation.

athomps:
@danstine Could you comment on how `jtags` is used. Does it work in parallel i.e. comm->nprocs > 1 ?

src/KOKKOS/mliap_data_kokkos.h
danstine: @akohlmey I have updated the code to now be `MAPPED_JATOMS_MASK = 0x00020000`  I don't know what I was thinking when I originally wrote `MAPPED_JATOMS_MASK = 0x00010002`

danstine:
@athomps I have renamed everything to k_jtags and jtags for the kokkos and non-kokkos versions, respectively.

@rohskopf I have my own complex models I have been working on, but once I have a free minute I will look to write a "vanilla" example of a simple message passing network. Thanks for the link!

@lubbersnick and @athomps Currently implementing a message passing neural network in parallel with multiple GPUs is a nightmare because there is an issue with the information about the embedding of neighboring atoms being in different memory. The only conceivable solution to this, at least in my eyes, is writing synchronization after every message pass, which is nontrivial. The change to the kokkos mliap data and regular mliap data allows for message passing models to be used on a single GPU, which I believe Nick has keenly identified. The question about partial f_ij is likely a larger discussion, and one that I am happy to engage in. However, the current PR is simply giving access to additional data structures for LAMMPS users that are looking to use the unified interface(s).

stanmoore1:
>Currently implementing a message passing neural network in parallel with multiple GPUs is a nightmare because there is an issue with the information about the embedding of neighboring atoms being in different memory.

Hi, can you explain this more? I don't think there is any fundamental difference between writing a parallel implementation for multiple GPUs vs multiple CPUs. So what works for multiple CPUs should also work for multiple GPUs, unless I'm not understanding something?

lubbersnick:
@danstine If you can come up with an elegant solution for message passing that also allows these models to work nicely during the backwards pass for force computation, that would be really cool. My current understanding is that such a thing could be rather difficult, because for example it might need some custom backward passes for the message passing stages (for forces) that utilize something inside of MLIAPUnified. I don't see any way to automate that kind of thing, so in that case every implementation may need some 'surgery' to construct an MLIAPUnified interface... Do you have a solution to that problem?

For non-decomposed (like LAMMPS serial) simulations I recommend ASE - it native operates by index rather than constructing ghost atoms. However it does not do domain decomposition, and it does have a performance floor which for us is usually a few microseconds per atom step. 

On the other hand, I don't want to detract from your progress here - if it doesn't interfere with the speed or memory cost of the rest of the system then to me it makes sense to include it (although I do not have any formal authority here).  If it does have consequences for resources, perhaps the MLIAPUnified interface should contain a flag that turns this feature on.

danstine:
@stanmoore1 This is mainly an issue of how message passing is performed on the model side. To be more precise, you can consider an example where each atom maintains its own abstract representation (A). Every MD simulation timestep we would perform T message passing iterations (for instance, 3). Every message pass updates A for each atom (A_i) depending on each atom's A and the abstract states of neighbors (A_j), which are also being updated depending on their neighbors (and so on). If A_j is on a different device then we run into a problem. One solution is to update the A_j of the ghost atoms, which requires the ghost atoms to have a neighbor list (otherwise we have no way to update their A because we don't know the A of their neighbors). This solution is definitely not one that I would be pursuing in LAMMPS. This leads to the idea of the "synchronization" that I mentioned, i.e., transferring A's from the atoms in the simulation cell to an array that keeps track of A's for neighboring atoms, which unfortunately would be very slow. Message passing doesn't work well for any domain decomposition strategies, which is something that plagues the implementation of these models. I hope this helps and wasn't too confusing. I am happy to discuss further. 

@lubbersnick We have ASE calculators for our models as well! I just figured I would share the update with the LAMMPS community so those that are interested in using the other functionalities of LAMMPS with their MLIPs through the unified interface have that option. The unified interface is great and makes models very easy to implement after taking a little time to figure out how it works (unbiased endorsement). Unfortunately, despite being desperately needed, I think a domain decomposition implementation (multi-GPUs) for message passing models is going to require a much larger effort than I currently have bandwidth for. Happy to collaborate though.

athomps:
@danstine Thanks for the explanations. I am glad you have thought about these issues properly.  I will mention two things:
1. " One solution is to update the A_j of the ghost atoms, which requires the ghost atoms to have a neighbor list" This is not a bad solution, and has been used for other energy models in LAMMPS, including HIPNN via MLIAP Unified, which @rohskopf mentioned in his comment.
2. I am working with someone on a prototype of the message-passing scheme with MPI communication between layers. Not clear yet how it is going to go.

athomps:
There is a small overhead with allocating, assigning, and copying jtags, and even within MLIAPUnified, it is only used by very special models.  I think it would be a good idea to wrap it in an if-test with an optional flag e.g. `jtagsflag`, similar to `gradgradflag.`, but not set by the user, set by the MLIAP Unified derived class.

stanmoore1:
Yes as Aidan mentioned, LAMMPS supports building a neighbor list for ghost atoms, FWIW.

lubbersnick:
> Yes as Aidan mentioned, LAMMPS supports building a neighbor list for ghost atoms, FWIW.

Yes, and MLIAPUnified supports turning this on with a simple flag already.

> @danstine Could you comment on how jtags is used. Does it work in parallel i.e. comm->nprocs > 1 ?

I am also interested in this. And if it only works in serial, what happens if someone tries running it in parallel mode?

danstine:
Hi all, thanks for all the great feedback. 

@athomps do you mind if I ask you to point me to the locations where a jtagsflag should be implemented (specifically regarding "but not set by the user, set by the MLIAP Unified derived class."). I think I know what you are recommending, but I thought I would double check before making changes to this PR. Admittedly, I am still new to the MLIAP unfiied interface.

On a side note, I wonder if it is relevant to add these flags to the other MLIAP data structures. As an example, my models (with serial implementation) only need data.rij, data.pair_i, data.jtags, and data.elems. It would be nice to avoid the overhead with all of the other MLIAP data altogether. Maybe there is already a way to do this and I am just missing it.

Either way, thanks for all the help. I will wait for @athomps feedback before adding the flags.

athomps:
Yes, it is not obvious how to do that. I just pushed a commit that added the member variable jtagsflag to MLIAPData.  In the pair style, it defaults to 0, so you will have to add a mechanism somewhere to change that to 1 somewhere in the message-passing part that I do not have access to.  I did not test this heavily, nor did I look at Kokkos. Nor did I add an error check for nprocs >1 && jtagsflag, which I think is needed.

lubbersnick:
@danstine As a suggestion about where to go from Aidan's additions, I can see two strategies. I don't know which is best.

The "slick" one is to make sure the MLIAPUnified objects have whether they need this as an attribute, similar to how the radius is passed. (In this case, from the Unified to the DummyDescriptor, into the pair style) https://github.com/lammps/lammps/blob/816a64b50b1d0f6dc2e8dd3d5f0e89e974947285/src/ML-IAP/pair_mliap.cpp#L348 . To do this we might want to define the Descriptor itself to store whether it needs these flags; for existing descriptors this would default to zero, and for the DummyDescriptor, the result would be grabbed from the python MLIAPUnified object (where it would also default to zero).

The second way is to add it as an optional flag to the lammps script that the user has to be aware is needed, similar to the ghostneigh flag. https://github.com/lammps/lammps/blob/816a64b50b1d0f6dc2e8dd3d5f0e89e974947285/src/ML-IAP/pair_mliap.cpp#L192 . This is easier to implement but requires more from the user to get things right.

danstine:
Thank you all for the feedback. I am going close this PR for now. On my end, I made several modifications to the code and I need to take the time to package them well to ensure that they will integrate seamlessly without disrupting the current code base. 

@lubbersnick and @athomps and @stanmoore1  thank you for all the feedback. We can revisit this in the future.

PR #3779, opened 2023-05-16
Reorder fix ttm/mod 3d arrays to be consistent with fix ttm and fix ttm/grid

sjplimp:
**Summary**

The 3d arrays in fix ttm/mod are indexed as array[ix][iy][iz].  The 3d arrays in fix ttm and fix ttm/grid are indexed array[iz][iy][ix].
This PR inverts the 3 indices for all the 3d arrays in fix ttm/mod to be consistent with fix ttm and ttm/grid, so that the
x values in the grid data structure vary fastest, then y, and z slowest.

**Related Issue(s)**

N/A

**Author(s)**

Steve

**Licensing**

**Backward Compatibility**

There is a small difference in the last couple digits of log file outputs in examples/ttm for in.ttm.mod.  Not sure yet
whether this is to be expected or is an indication of a bug in the refactored code.

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/src/EXTRA-FIX/fix_ttm_mod.cpp
+++ b/src/EXTRA-FIX/fix_ttm_mod.cpp
@@ -894,9 +894,9 @@ double FixTTMMod::compute_vector(int n)
   double dz = domain->zprd/nzgrid;
   double del_vol = dx*dy*dz;
 
-  for (int ix = 0; ix < nxgrid; ix++)
+  for (int iz = 0; iz < nzgrid; iz++)
     for (int iy = 0; iy < nygrid; iy++)
-      for (int iz = 0; iz < nzgrid; iz++) {
+      for (int ix = 0; ix < nxgrid; ix++) {
         e_energy += el_sp_heat_integral(T_electron[iz][iy][ix])*del_vol;
         transfer_energy += net_energy_transfer_all[iz][iy][ix]*update->dt;
       }
--- a/src/EXTRA-FIX/fix_ttm_mod.cpp
+++ b/src/EXTRA-FIX/fix_ttm_mod.cpp
@@ -894,9 +894,9 @@ double FixTTMMod::compute_vector(int n)
   double dz = domain->zprd/nzgrid;
   double del_vol = dx*dy*dz;
 
-  for (int iz = 0; iz < nzgrid; iz++)
+  for (int ix = 0; ix < nxgrid; ix++)
     for (int iy = 0; iy < nygrid; iy++)
-      for (int ix = 0; ix < nxgrid; ix++) {
+      for (int iz = 0; iz < nzgrid; iz++) {
         e_energy += el_sp_heat_integral(T_electron[iz][iy][ix])*del_vol;
         transfer_energy += net_energy_transfer_all[iz][iy][ix]*update->dt;
       }
--- a/src/EXTRA-FIX/fix_ttm_mod.cpp
+++ b/src/EXTRA-FIX/fix_ttm_mod.cpp
@@ -659,6 +659,9 @@ el_heat_capacity_thermal_conductivity FixTTMMod::el_properties(double T_e)
   properties.el_thermal_conductivity = el_th_diff*properties.el_heat_capacity; // thermal conductivity
   return properties;
 }
+
+/* ---------------------------------------------------------------------- */
+
 double FixTTMMod::el_sp_heat_integral(double T_e)
 {
   double T_temp = T_e/1000.0, T_reduced = T_damp*T_temp;
@@ -751,6 +754,7 @@ void FixTTMMod::end_of_step()
     for (int iy = 0; iy < nygrid; iy++)
       for (int ix = 0; ix < nxgrid; ix++)
         T_electron_first[iz][iy][ix] = T_electron[iz][iy][ix];
+  
   do {
     for (int iz = 0; iz < nzgrid; iz++)
       for (int iy = 0; iy < nygrid; iy++)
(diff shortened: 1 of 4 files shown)


sjplimp:
@akohlmey As indicated in the description, there is a small (last couple digits) in the logfile for the examples/ttm/in.ttm.mod script (1 and 4 procs) due to this change.  I can't see why this should be, though possibly it is simply round-off.  But I've marked this work-in-process until we figure it out.

sjplimp:
It is more dramatically different when the grid is not 10x10x10 but 6x8x10 - so something is wrong I think.

src/EXTRA-FIX/fix_ttm_mod.cpp
```diff
@@ -348,14 +335,14 @@ void FixTTMMod::post_force(int /*vflag*/)
...
-          double T_i = T_electron[ix][iy][iz];
-          double T_ir = T_electron[right_x][iy][iz];
-          double T_iu = T_electron[ix][right_y][iz];
-          double T_if = T_electron[ix][iy][right_z];
-          double C_i = el_properties(T_electron[ix][iy][iz]).el_heat_capacity;
-          double C_ir = el_properties(T_electron[right_x][iy][iz]).el_heat_capacity;
-          double C_iu = el_properties(T_electron[ix][right_y][iz]).el_heat_capacity;
-          double C_if = el_properties(T_electron[ix][iy][right_z]).el_heat_capacity;
+          double T_i = T_electron[iz][iy][ix];
+          double T_ir = T_electron[right_z][iy][ix];
```
akohlmey: @sjplimp are you sure that these changes are correct? `T_ir` changes from "right_x" to "right_z" and `T_if` the other way around.

src/EXTRA-FIX/fix_ttm_mod.cpp
```diff
@@ -821,50 +804,50 @@ void FixTTMMod::end_of_step()
...
               if (ix >= t_surface_l) mult_factor = (intensity/(dx*skin_layer_d))*exp((-1.0)*(ix_d - surface_d)/skin_layer_d);
             }
-            if (ix < t_surface_l) net_energy_transfer_all[ix][iy][iz] = 0.0;
+            if (ix < t_surface_l) net_energy_transfer_all[iz][iy][ix] = 0.0;
             double cr_vac = 1;
-            if (T_electron_old[ix][iy][iz] == 0) cr_vac = 0;
+            if (T_electron_old[iz][iy][ix] == 0) cr_vac = 0;
             double cr_v_l_x = 1;
-            if (T_electron_old[left_x][iy][iz] == 0) cr_v_l_x = 0;
+            if (T_electron_old[left_z][iy][ix] == 0) cr_v_l_x = 0;
```
akohlmey: @sjplimp here is another similar case. `cr_v_l_x` is set based on `left_z` instead of `left_x` and so on.

sjplimp:
@akohlmey Good catch on the left/right indices.  I fixed that and now the old/new answers agree even with a non cubic grid.
However the issue with the last couple digits of thermo output being different between new and old versions is still there.
For cubic or non-cubic grids.  This is the last couple digits of 16-digit precision, for 1000 step runs, 1 proc or 4 procs.

Any ideas?  I'm not seeing how operations would be performed in different orders due to the changes I made.

akohlmey:
> Any ideas? I'm not seeing how operations would be performed in different orders due to the changes I made.

The changes could impact the loop unrolling done by the compiler at high optimization level.
You would have to compile the before and after versions with -O0 to check if that makes a difference.

sjplimp:
@akohlmey I tried the test you suggested with -O0, comparing old and new versions.  It didn't change anything.
O0 and O3 are the same for the old version.  Also the same for the new version.  But old vs new is slightly different in
the last one of two digits of the fix output (out of 16 printed digits).  Even on timesteps 1 or 2.  Same issue with a 10^3 grid
or 6x8x10 grid.

I tried running for 10,000 steps instead of 1000 steps.  Same issue, but the differences do not grow in size.  Also,
the dynamics of the system (temp, toteng) are identical on every step.  It's only the fix output (2 values) for electron
eng and temp, where the tiny differences are, even though it's coupled to the dynamics.

The code computing these two values is quite complex in ttm/mod (more so than the other two variants), so I'm
not optimistic I can find what is causing such a tiny difference.  I cannot see any mistakes I've made with swapping
the order of indices.

Any suggestions?

akohlmey:
> Any ideas? I'm not seeing how operations would be performed in different orders due to the changes I made.

There is at least one place, where you *did* change the order of operations:  `FixTTMMod::compute_vector()`.
I've switched back the order in which the two properties are summed and that recovers the exact output from before the other changes in this pull request.

sjplimp:
ah, good catch.  That's a summation over the grid only for purposes of output.  Which explains why it didn't affect the dynamics, even over long timesscales.  Just the last couple digits of hi-precision values printed on the thermo line.

The way I have it (z outer, x inner) is the way it should be in compute_vector() (and everywhere else), to access the 3d arrays with stride one.  So we should leave it that way and re-bless the output.  Do you want me to produce new log files for the PR ?

akohlmey:
I didn't see any significant performance difference either way. Just the fluctuations you see with any calculation these days in times of turbo boost and stack randomization.

sjplimp:
@akohlmey created new log files consistent with z,y,z loop ordering - this should be ready to merge
thanks for your help

akohlmey (approved):

PR #3782, opened 2023-05-17
Nonequilibrium Free-energy calculation of Phase-Boundaries using LAMMPS

samuelcajahuaringa:
**Summary**

This repository contains a set of source codes and input scripts that allows to perform nonequilibrium free-energy calculations of phase-boundaries of atomistic systems using LAMMPS code. Details about the code implementation, capabilities and obtained results with these methods will be found in the following github project: https://github.com/samuelcajahuaringa/dCCIforLAMMPS

The source codes are found in the directory /REPLICA/ and only need to copied in the /src directory. The files are:

- dcci.cpp / .h : dynamical integration of Clausius-Clapeyron equation.
- fix_adapt_dcci.cpp / .h : similar syntax of fix_adapt.cpp / . h but in this case the fscale is controlled by the dcci command.
- fix.h : existing file, we added function reset_target_pressure that control the pressure by dcci command.
- fix_nh.cpp / .h : existing file, controlling the external pressure on the barostat by dcci command. 

The dcci command allows phase boundary curve using nonequilibrium free-energy calculations, as example is calculated the phase boundary between the fcc and liquid phase of 3d LJ system, scripts found in the directory /examples/dcci/

From the example use the following command to run the dcci simulations:

mpirun -np 2 lmp_mpi -partition 2x1 -in in.dcci

**Author(s)**

Samuel Cajahuaringa
Postdoctoral researcher at Center for Computing in Engineering & Sciences, UNICAMP/Brazil
email: samuelcajahuaringa@gmail.com

**Licensing**

**Backward Compatibility**

The codes are compatible with latest lammps versions.

**Implementation Notes**

The dcci command don't need significant changes in the current version of lammps.

**Post Submission Checklist**

https://github.com/samuelcajahuaringa/dCCIforLAMMPS
https://doi.org/10.1016/j.commatsci.2022.111275

### DIFF
--- a/src/KOKKOS/angle_cosine_kokkos.cpp
+++ b/src/KOKKOS/angle_cosine_kokkos.cpp
@@ -150,20 +150,26 @@ void AngleCosineKokkos<DeviceType>::operator()(TagAngleCosineCompute<NEWTON_BOND
   const int i3 = anglelist(n,2);
   const int type = anglelist(n,3);
 
+  const F_FLOAT k = d_k[type];
+
+  const F_FLOAT x20 = x(i2,0);
+  const F_FLOAT x21 = x(i2,1);
+  const F_FLOAT x22 = x(i2,2);
+
   // 1st bond
 
-  const F_FLOAT delx1 = x(i1,0) - x(i2,0);
-  const F_FLOAT dely1 = x(i1,1) - x(i2,1);
-  const F_FLOAT delz1 = x(i1,2) - x(i2,2);
+  const F_FLOAT delx1 = x(i1,0) - x20;
+  const F_FLOAT dely1 = x(i1,1) - x21;
+  const F_FLOAT delz1 = x(i1,2) - x22;
 
   const F_FLOAT rsq1 = delx1*delx1 + dely1*dely1 + delz1*delz1;
   const F_FLOAT r1 = sqrt(rsq1);
 
   // 2nd bond
 
-  const F_FLOAT delx2 = x(i3,0) - x(i2,0);
-  const F_FLOAT dely2 = x(i3,1) - x(i2,1);
-  const F_FLOAT delz2 = x(i3,2) - x(i2,2);
+  const F_FLOAT delx2 = x(i3,0) - x20;
+  const F_FLOAT dely2 = x(i3,1) - x21;
+  const F_FLOAT delz2 = x(i3,2) - x22;
 
   const F_FLOAT rsq2 = delx2*delx2 + dely2*dely2 + delz2*delz2;
   const F_FLOAT r2 = sqrt(rsq2);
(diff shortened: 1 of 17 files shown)


athomps:
@samuelcajahuaringa There are some basic problems with the way you have added LAMMPS source files to this PR.  There are files from ./src that have been added to ./src/REPLICA ( I'm looking at you, pair_lj_cut.cpp / h). There are also new dCCI files that have been added to both ./src and ./src/REPLICA.

athomps:
@samuelcajahuaringa It is not obvious to me that you need to add function `reset_target_pressure()` to the Fix base class.  In general, you should try to arrange your new feature so as to minimize changes to the core of LAMMPS.  In most cases, this means zero changes.

akohlmey:
@samuelcajahuaringa In addition to the comments from @athomps, please note that submissions have to pass *all* automated tests. Also, for the latest information on formal and programming style requirements for submissions, please see sections 3.1 to 3.4 from here: https://docs.lammps.org/latest/Modify.html

samuelcajahuaringa:
> @samuelcajahuaringa There are some basic problems with the way you have added LAMMPS source files to this PR. There are files from ./src that have been added to ./src/REPLICA ( I'm looking at you, pair_lj_cut.cpp / h). There are also new dCCI files that have been added to both ./src and ./src/REPLICA.

@athomps It was a misunderstanding on my part. I thought I had to leave the files in the ./src folder. I move the files to the ./src/REPLICA folder

athomps:
In the same vein, if you *do* still need to change existing LAMMPS files like fix.h, you should modify the original in ./src, not a copy in ./src/REPLICA

samuelcajahuaringa:
> @samuelcajahuaringa It is not obvious to me that you need to add function `reset_target_pressure()` to the Fix base class. In general, you should try to arrange your new feature so as to minimize changes to the core of LAMMPS. In most cases, this means zero changes.

@athomps the dcci command is a multi-replica model, but in this case only two replica that represent each coexistence phase, that run simultaneously.  The dcci command works like parallel tempering commands such as temper, temper/npt, temper/grem where these command modified the target temperature of each replica through function reset_target that is define in the fix.h file. For the dcci command is necessary control the target pressure of each replica, due this reason I came the need to define reset_target_pressure to modified the target pressure of barostat implemented in fix_nh.h

akohlmey:
@samuelcajahuaringa to further elaborate on @athomps' comments. Your contribution is currently *very* broken. There are no duplicate/modified files allowed. Your copying of files that already exist in src/ from the REPLICA package is an absolute no-no. It is broken by design and leads to inconsistent and non-reproducible behavior. In addition, your contribution *must* support both the conventional build system and CMake builds and in CMake builds all files remain in place.

Moreover, your replicated files are not up-to-date and thus not compatible with the core LAMMPS code. They undo quite a few changes and are missing features and thus they will break existing functionality. This is not acceptable for a feature to be included into the distribution. Please carefully study the documents I have pointed out to you. They contain a lot of useful information.

I have not looked too closely, but from what I saw, there is going to be much more work needed until your contribution can be cleanly integrated. What you are doing is ok for an external package where breakage of unrelated functionality can be tolerated, but the bar for permanent contributions to the LAMMPS distribution is higher. However, it is pointless to update and modernize the code while it breaks the compilation of existing code. So that needs to be addressed first.

athomps:
@samuelcajahuaringa Everything @akohlmey is correct, and perhaps a bit discouraging. However, with a reasonable investment of effort on your part, it should be possible to get this worthy contribution added to the REPLICA package.

akohlmey:
And while working on getting things fixed, I would also suggest to try and find a more descriptive command name than "dcci". We try to avoid initialisms because they are similar to encryption: nobody except people who already know what it stands for know what they mean and represent. This is - while also not ideal - acceptable for some legacy commands like prd or tad, because their abbreviations are somewhat commonly used. But for new commands it means that rarely somebody will even look at them, and that cannot be in your interest. Why contribute something and than give it a name that hides it in plain sight?

PR #3789, opened 2023-05-20
Phase-Boundaries calculations using dynamical Clausius-Clapeyron integration 

samuelcajahuaringa:
**Summary**
 
Adds dynamical Claucius-Clapeyron integration (dcci) command that allow the calculation of coexistence curve. The advantage of the dcci method gives coexistence curves spanning a wide range of thermodynamic states from relatively short single non- equilibrium simulations. The dcci command is a multi-replica simulation, in this case with only two replica, one for each phase. 
The dcci command works in similar that temper/grem, but in this case the dcci command control the external pressure on each phase along the coexistence simulation.
List of source codes
dcci.cpp/.h : perform dynamical Claucius-Clapeyron integration to calculate the coexistence curve, save in log.lammps file.
fix_adapt_dcci.cpp /.h : scaling the force on atoms, where the scaling parameter is controlled by the dcci command.
fix.h : is modified with the function reset_target_pressure that allow dcci command modified the target pressure, function similar to reset_target that control the temperature use by the temper, temper/npt and temper/grem.
fix_nh.cpp / .h : adding the function reset_target_pressure to control the target pressure on barostat by dcci command.
pair_lj_cut.cpp / .h : pair style example, was added fscale parameter to scaling only the force on the atoms.

scripts found in /examples/dcci directory

test with at least 2 mpi process, one for each phase.

run: mpirun -np 16 lmp -partition 2x8 -in in.dcci  time execution less that 1 minute.

**Related Issue(s)**

**Author(s)**

Samuel Cajahuaringa, postdoc researcher at Center for Computing in Engineering and Sciences, UNICAMP/Brazil
email: samuelcajahuaringa@gmail.com

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

Further details and benchmark tests can be found the following github project: https://github.com/samuelcajahuaringa/dCCIforLAMMPS


athomps (commented):
@samuelcajahuaringa Unfortunately, this modification to pair_lj_cut.cpp to add the factor `fscale` is not acceptable.  The LAMMPS-ian way to achieve this is how it is done in `fix_dapt.cpp` i.e. `pair->extract(ad->pparam,ad->pdim)`.

samuelcajahuaringa:
> The LAMMPS-ian way to achieve this is how it is done in `fix_dapt.cpp` i.e. `pair->extract(ad->pparam,ad->pdim)`.
@athomps  you right, the fix_adapt.cpp change pair style parameters over time as a simulation runs, but in this case the dynamical Clausius Clapeyron integration only works with pair parameter that only change the force on the atoms. For this reason was  implemented the factor fscale on pair lj as example to other users implement this factor for other pair style. Also was defined a modified version of fix_adapt in the same fashion of fix_adapt_fep. The fix_adapt_dcci has the following differences:
* only scaling the force for each time step
* it is applied to pair style with the factor fscale
* the scaling factor is controlled by the dcci command

the modification on the pair style (fscale factor) only works with fix_adapt_dcci.cpp

athomps:
> > @athomps  you right, the fix_adapt.cpp change pair style parameters over time as a simulation runs, but in this case the dynamical Clausius Clapeyron integration only works with pair parameter that only change the force on the atoms. For this reason was  implemented the factor fscale on pair lj as example to other users implement this factor for other pair style. 

This is a showstopper.  We can't accept changes to the inner force loop of every pair style, it creates a performance loss and is large maintenance burden.  That is why we created the pair->extract() method. If you can't find a way to implement this dcci feature in that way, we can't accept it in to LAMMPS.

akohlmey:
> * it is applied to pair style with the factor fscale

Why use this factor? Why not adapt "epsilon"?

samuelcajahuaringa:
> This is a showstopper. We can't accept changes to the inner force loop of every pair style, it creates a performance loss and is large maintenance burden. That is why we created the pair->extract() method. If you can't find a way to implement this dcci feature in that way, we can't accept it in to LAMMPS

I used the pair->extract() method, review the lines 707-714 of pair_lj_cut.cpp

void *PairLJCut::extract(const char *str, int &dim)
{
  dim = 2;
  if (strcmp(str, "epsilon") == 0) return (void *) epsilon;
  if (strcmp(str, "sigma") == 0) return (void *) sigma;
  if (strcmp(str,"fscale") == 0) return (void *) fscale;  <-
  return nullptr;
}

the epsilon parameters not works in this case because scaling the force and energy, and the non-equilibrium method used to calculated the coexistence curve needs perform a dynamics with only scaled force and compute its potential energy of the system.

akohlmey:
@samuelcajahuaringa if you need scaled forces only, why can't you scale the forces *after the fact*?
Please note that your current implementation is broken as it has many failing tests that *must* pass for such a general change that is supposed to have no effect on other uses.

samuelcajahuaringa:
If was necessary only scaled the force, scaled after the calculation is a good option, but in this case the scaling force is necessary because is scaled the virial pressure that correspond only to pair terms excluded the bonds, angles, dihedral and constraints. The simulation of two phases correspond to perform NPT simulation on non-physical instantaneous pressure. I'm fairly new to using github, I test the code compile cmake and make, the codes works correctly.  What kind of conflicts are appearing?  Is it necessary some additional script to perform the tests?

athomps:
> the epsilon parameters not works in this case because scaling the force and energy, and the non-equilibrium method used to calculated the coexistence curve needs perform a dynamics with only scaled force and compute its potential energy of the system.

Thanks for the clarification! It is still very problematic that you have to add fscale to the inner force loop. I have not read the paper, nor do I know how the dcci method works in detail. However, it is very unusual to propose a method that calculates a "force" and an "energy" that are not consistent with each other. I say this in the hope that there is a way reformulate the dcci method in a way that "force" and "energY" can use a common "epsilon", as @akohlmey suggests. That would make it much more straightforward (i.e. feasible) to implement in LAMMPS.

akohlmey:
> I'm fairly new to using github, I test the code compile cmake and make, the codes works correctly.

Testing is independent from GitHub. Please see https://docs.lammps.org/Build_development.html#code-coverage-and-unit-testing-cmake-only and https://docs.lammps.org/Developer_unittest.html

The most common reasons for breaking unit tests for pair styles are using uninitialized data (check with memcheck tool from valgrind), incorrect/inconsistent restarting, or inconsistencies between compute() and single().

akohlmey:
@samuelcajahuaringa you have to understand the reason we are so concerned is that you want to make changes on very common code paths for an application that seems rather "special" and limited in its application. If your contribution could be implemented without the changes to the core code, we would have made a few suggestions to modernize it and follow some of our conventions more closely, and then it would have been good.

But for "disruptive" changes like you need to do, you need to give a very good justification beyond "it is needed for my feature", in that your feature is of significant use to many LAMMPS users and is of a generic nature and can be easily applied to other use cases beyond what you have tested so far.

For that it would probably be best, if you can give us an "elevator pitch version" of the method. There are a few things that sound odd from what you have stated so far:
- the fact that you need to scale forces but need unscaled energies
- you implement fscale as a 2d array so it can be applied to individual pairs of atom types, but your example and descriptions imply a single, global scaling factor
- you mentioned that bonded interactions should be unchanged/unscaled which seems odd as well. More importantly, how can this method then be applied for force fields where the bonded interactions are implicit like ReaxFF, AIREBO, BOP etc.?

At the moment, it seems much more straightforward, if you would maintain your feature as an external USER package or as a forked LAMMPS version with your changes added. Then most of our concerns do not apply, since only people that are interested in your feature will be downloading and using that specific version. We will be happy to add a link to it from the LAMMPS homepage and also give the notes on modernizing it.

samuelcajahuaringa:
Thank for the advices @akohlmey, my intention is not to modify the core of LAMMPS, but rather provide dcci command as an external sources in REPLICA package such as temper/grem fix_temper_grem, where users interested in using the dcci package could be installed it.

athomps:
@samuelcajahuaringa Do you see a way to get this implemented without requiring modifications to e.g. `pair_lj_cut.cpp`?  If not, then this is perhaps something that is best maintained and distributed separately by your research group. There are prominent precedents for this e.g. https://vcsgc-lammps.materialsmodeling.org. Interestingly, that functionality has recently been successfully merged back into LAMMPS MC package.

srtee:
@samuelcajahuaringa I've tried to read the paper and I'm not sure why this functionality couldn't be achieved with `pair_style hybrid/scaled`, which can apply a linear combination of potentials with the "mixing ratio" changing over time.

athomps:
I recommend this PR be closed due to inactivity.

akohlmey:
This PR has multiple unresolved issues: files added in the wrong location, modifications to core files without convincing justification, lack of generality, failing unit tests. Please have a close look at the discussions and suggestions, create a new branch that addresses the problems and resubmit, if you still think it is a valuable contribution. For the time being, we'll close this one.

PR #3790, opened 2023-05-20
is_typelabel function

jrgissing:
**Summary**

add is_typelabel function to check if a particular type label already exists in the simulation

**Related Issue(s)**

none

**Author(s)**

JG

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/doc/src/variable.rst
+++ b/doc/src/variable.rst
@@ -999,16 +999,14 @@ types, using label maps created by the :doc:`labelmap <labelmap>` or
 :doc:`read_data <read_data>` commands.  The first argument is the label
 map kind (atom, bond, angle, dihedral, or improper) and the second
 argument is the label.  The function returns the corresponding numeric
-type or -1 in case the specified label for the given kind of type does
-not exist.
+type or triggers an error if the queried label does not exist.
 
 .. versionadded:: TBD
 
 The is_typelabel(kind,label) function has the same arguments as
 label2type(), but returns 1 if the type label has been assigned,
 otherwise it returns 0.  This function can be used to check if a
-particular type label already exists in the simulation.  It is
-equivalent to `(label2type(kind,label)!=-1)`.
+particular type label already exists in the simulation.
 
 ----------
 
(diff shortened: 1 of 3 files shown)


akohlmey:
@jrgissing @sjplimp How about we make the `label2type()` function just return -1 for unknown labels (as the underlying function does) and not exit with an error? This would avoid having to have two nearly identical functions.

jrgissing:
that was my initial inclination as well. we'll just have to check if there is already usage that assumes label2type returns a valid label. also, in an input script setting, I was not happy that it would translate to 'True' when there is no matching label

akohlmey:
> I was not happy that it would translate to 'True' when there is no matching label

Since 0 is not a valid type, the variable function could return 0 instead of -1 if the label is not found.

jrgissing:
> Since 0 is not a valid type, the variable function could return 0 instead of -1 if the label is not found.

good point. I am still a little hesitant to 'overload' the label2type function in this way, due to clarity within input scripts, but I can update the code as you suggest

akohlmey:
> I can update the code as you suggest

I think both arguments (input file clarity versus avoiding redundant code) mostly cancel. I don't want to force my personal preference on this, so let's put it to a vote:

@stanmoore1 @athomps @sjplimp what is your preference?

athomps:
I don't like 0, as some users might think 0 is a valid numeric type. Better to use -1 or some other signal value.

stanmoore1:
> I don't like 0, as some users might think 0 is a valid numeric type. Better to use -1 or some other signal value.

I agree.

akohlmey:
That settles the matter. I've made some minor cleanups, fixed a sphinx formatting issue and added some unit tests. This is now ready for merging as far as I am concerned.

doc/src/variable.rst
```diff
@@ -996,10 +996,19 @@ via the link in this paragraph.
...
+map kind (atom, bond, angle, dihedral, or improper) and the second
+argument is the label.  The function returns the corresponding numeric
+type or -1 in case the specified label for the given kind of type does
+not exist.
+
+.. versionadded:: TBD
+
+The is_typelabel(kind,label) function has the same arguments as
+label2type(), but returns 1 if the type label has been assigned,
+otherwise it returns 0.  This function can be used to check if a
```
yafshar, line 1009: @akohlmey I thought the agreement was on -1. Should this be corrected to `otherwise it returns -1`

jrgissing:
I approve

stanmoore1 (approved):

athomps (approved):
Approved.

athomps (approved):
Approved.

PR #3791, opened 2023-05-20
Adding fix press/langevin

Bibobu:
**Summary**

This fix couples each of the box dimension with Langevin particles following the equation from Grønbech-Jensen and Farago:  https://doi.org/10.1063/1.4901303

The equation of motion initially coupled only to the volume can be easily extended to every terms of the strain tensor.  Several differences are to be mentioned compared to other pressure control fixes:
* The temperature used in the pressure computation is the ensemble temperature.
* As such, the fix uses no compute for the temperature and a pressure compute with the virial keyword.
* Following GJF equations, the volume change is done at the beginning of the step before integration
* The basis for this fix was press/berendsen file. But the addition of shear pressure control required the addition of flip keyword and tilt skew checks.

Comments on the fix modification, placement in the file tree or the addition of tests are welcomed.

**Related Issue(s)**

**Author(s)**

Germain Clavier (TUe)

**Licensing**

**Backward Compatibility**

No known backward compatibility issues is to be reported.

**Implementation Notes**

This is a straightforward fix similar to press/berendsen and fix langevin for some implementation details. The system was tested on lennard-jones crystal and fluid (reduced units) a gives correct pressure in the 0.3 to 0.7 temperature range. The shear values also appear to be correctly controlled after some shearing tests in the 0. 3. pressure range for the solid.

**Post Submission Checklist**

Original Niels Grønbech-Jensen and Farago paper: https://doi.org/10.1063/1.4901303


athomps:
@Bibobu This looks very nice.  I noticed one thing that would be highly desirable to change, if possible.  The units of Pdamp are not time, which deviates from the pattern followed by most other thermostats and barostats in LAMMPS. Can you redefine this to have time units, preferably related to some physical timescale of the barostat response, similar to e.g. fix press/berendsen?

Bibobu:
@athomps the equation of motion being similar to those implemented in `fix langevin` this seems feasible to me. But I guess I need a bit more time to figure out how the equations were transformed compared to GJF papers. My understanding at the moment  is that the 3rd parameter of temp in `fix langevin` refers to the friction coefficient whereas in this `fix press/langevin` the third parameter referred to the fictitious particle masses.

I agree that it is indeed more logical to swap and keep consistency between both `fix langevin`, `fix press/langevin` and `fix press/berendsen` by making the 3rd parameter the friction coefficient of the barostat and having another option to set the fictitious particles masses (and have it all the same for all fictitious particles as a start).

I'll make the change when I manage to find a bit of time and update the PR accordingly.

akohlmey:
@athomps @sjplimp Should this be added to the src folder or src/EXTRA-FIX?

athomps:
I predict this will be as widely used as the other popular Nose-Hoover/Langevin/Berendsen baro- and thermo-stats, so therefore in `src`.

stanmoore1 (approved):
The code looks good (though I haven't tested it with running).

athomps (changes requested):
In order to be consistent with the other widely used thermostats and barostats in LAMMPS (Nose-Hoover, Langevin, Berendsen), the definition of *Pdamp* needs to be changed i.e. it needs to be a quantity with units of time and roughly proportional to the actual response time of the barostat. For context, the LAMMPS Langevin thermostat takes a parameter *damp* that is described as follows "The damp parameter is specified in time units and determines how rapidly the temperature is relaxed. For example, a value of 100.0 means to relax the temperature in a timespan of (roughly) 100 time units ( or fs or ps - see the units command)"

charlessievers:
@Bibobu I can help with getting pdamp into the implementation, if you would like some help.

Bibobu:
@athomps @charlessievers Thanks for the comments and offer of help. I had some time to make the changes that makes the code more consistent with the GJF paper and the LAMMPS implementation of barostats as requested. From the format of the Langevin equation, I defined $\tau_{P}=\frac{Q}{\alpha}$ which allows to enter a damping value in time units. Similarly to the Berendsen barostat, I think it was more consistent to make the mass common to all barostat particles so I basically switched the *friction* and *mass* keywords, and changed friction to a damping time.

edit: Also corrected a bug in the GJFa and GJFb avlues where the alpha parameter was forgotten compared to the GJF paper. This might have to be compared to the langevin implementation but I do not whink this is a big deal.

This required several other minor changes in the code such as the computation of independent random forces. For consistency with the fix langevin command I also requested the random seed to be given with the *temp* keyword.

However, from a series of tests on a FCC LJ crystal (lj units), I see slow decorrelation of the pressure using low value like $\tau_{P}=10^{-3}$ (high friction, so overdamped oscillator) and oscillations with higher value (low friction so damped oscillator). I think this behaviour is different to what is expected from Berendsen and NH barostat and, if you can confirm this, I think the documentation might need correction to this regard. The value of $\tau_{P}=10^{-2}$ gave nice decorrelation with default $Q=10^{-3}$.

Would be very nice if this was confirmed on your side.
Press_corre_tauP

athomps:
@Bibobu Thanks for these changes, particularly redefining *Pdamp* to have units of time.  The dog page should be updated to reflect this, similar to other thermostats and barostats.  The behavior you show with different values of *Pdamp* is opposite to what is desired, suggesting that *Pdamp* now has units of 1/time.  Also, the fictitious mass should not be user-specified, since changing that will change the response time as well.  In other words, it should be possible to define *Pdamp* in such a way that all other parameters, such as Q and alpha, or combinations of them, are determined.

Bibobu:
After giving this some thought and trial and errors I am a bit puzzled at the moment. I understand the request to have a time constant as an input parameter but I do not see how to solve this in the implementation.

The relaxation time of a Langevin dynamics equation is given by the ratio $m/\gamma$. This is the 3rd argument of the temp keyword of the `fix langevin` where $m$ is taken as the individual mass of each particles. This relaxation time is roughly speaking the boundary between ballistic and diffusive regimes in Langevin dynamics.

In the case of the barostat I see two things to consider:
1. This time constant is apparently not equivalent to the time constant expected as an input parameter. The $\tau$ parameter I defined has units of time but has a different relationship with pressure correlation than what is expected. But it is equivalent to the `fix langevin` input parameter, so I think a choice has to be made here.

2. This time constant still requires the definition of a mass for the fictitious particles. This is why both are stated in the GJF papers and tested in all Langevin barostat papers as far as I could find. Yet, inspired by the NH barostat, I tried to use the same definition of the barostat mass with $Q=k_BT*(N_{atoms}+1)$. This gave me odd results with inconsistent behaviour of the correlation with time constants. I think this is because this value is derived from the extended NH-hamiltonian and so it makes no sense here. One could also define a mass from the expected oscillation of the non-Langevin dynamics of the fictitious particles since the force from the pressure acts as a harmonic oscillator. We could have similar to $\tau=\sqrt{Q/k}$ with $k=1$ as in the initial GJF paper. But this leaves the friction out of the definition and does not really solve the full problem.

So trying to follow LAMMPS guideline, an idea with which I agree here, and having a single input defining at least two required values seems like squaring the circle to me at the moment.

But maybe there is a solution and I can't find it yet. Any help and comment would be much appreciated.

akohlmey:
> But maybe there is a solution and I can't find it yet. Any help and comment would be much appreciated.

This makes me think of the "modulus" keyword in fix press/berendsen. There the time constant is not sufficient to fully determine the multiplier for the cell change and the modulus is included as a modifier and set to a default value for a given material. Please see the corresponding note in the documentation page. I am not sure, though, if there is a corresponding property that can relate a given time constant to the properties of the material in this case and thus achieve the desired behavior. But that may just be a direction worth looking into.

akohlmey:
@Bibobu @athomps @charlessievers Now that the stable release is out and the LAMMPS workshop done, this PR is eligible to be merged and an effort made to complete it. Thanks.

Bibobu:
> @Bibobu @athomps @charlessievers Now that the stable release is out and the LAMMPS workshop done, this PR is eligible to be merged and an effort made to complete it. Thanks.

I did not have time recently to put more work in this PR. However we planned some discussion about it with @athomps in the coming days. This should help setting up the last details about the implementation.

Bibobu:
I have worked on this last version of the PR. Several changes are to be noticed:
* First, a conversion of the kinetic part of the pressure was not done and the fix was not usable with non-LJ units. This is not the case anymore, and it worked with both argon and copper test case in real and metal units.
* The input parameters are now a Pdamp value that is similar to NH barostat and the decorrelation time (or Langevin time in the doc). A value of 0 for this friction parameter makes the barostat behave as an oscillator, with dissipation happening (very slowly) through kinetic equilibration of the system, if any. Else, a good combination of parameter has to be found for friction to be sufficient for pressure equilibration but not too high to allow convergence. Empirically a value half the Pdamp value gave good results.

athomps (approved):
After making minor edits, I am satisfied that this is ready for merge.  @Bibobu please look at my changes and make sure they are acceptable to you, and do not change behavior.

akohlmey:
@Bibobu @athomps I've made a few additional changes, mostly to modernize access to members of the Modify class. Also, the fix was missing to set the Fix::box_change flags. Finally, I also modernized the error messages to be consistent with the new guidelines. Since some of the same changes were also required in fix press/berendsen (that this fix seems to be based on) I also transferred the updates there.

We should probably have an example folder added with some representative examples for both fix press/berendsen and fix press/langevin (and add a couple fix npt examples for good measure) so that people can see that using these fixes correctly may require some additional adjustments.

Other than that, this seems good to go. @Bibobu I've just doubled the burden on you to check, since you need to check that neither @athomps nor me messed anything up...

Bibobu:
As far as I can tell, I see no problem with the changes made by @akohlmey and @athomps. The code gives similar results and the heuristic of setting $Pdamp=\tau_L$ is sound. As a quick test, a FCC LJ crystal converges to 0 pressure at $T=0.3$ with only $Pdamp=1$ set. The box change flags setting I forgot, new error format and clang format application are obviously sound.

Sounds good to me.

PR #3792, opened 2023-05-20
Inclusion of bond & angle contributions to "compute stress/mop"

### LINKED ISSUE #3683: [Feature Request] Inclusion of bond stretching contribution to "compute stress/mop"
**Summary**

Currently,  "compute stress/mop" cannot treat systems with intramolecular interactions. A first improvement is to include contributions from bond stretching terms.

**Detailed Description**

The current implementation derives the force of individual pair contributions from the pair->single method. Similarly, the force of individual bond contributions can be derived from the bond->single method or the bond->born_matrix.

 The geometric criterion for inclusion or exclusion of a contribution of bonded particles to the local pressure tensor can be the same as the current criterion for inclusion of non-bonded interactions.

A method similar to the born_matrix->compute_bonds() can perform the summation of bond contributions. 

evoyiatzis:
**Summary**

Inclusion of bond and angle contributions to "compute stress/mop". The expressions are taken from the appendix (A.6 & A.7) of "Stress tensor in model polymer systems with periodic boundaries" https://doi.org/10.1002/mats.1993.040020204

The validity of the bond contribution has been tested using the attached in.test_bonds_mop.txt script file which computes the stress profile on a plane between two atoms which are (i) either bonded using a harmonic spring with no non-bonded interactions or (ii) interacting exclusively via repulsive interactions using the harmonic/cut potential (thus the energy and the forces in these two systems should be identical). Different positions have been tried and the results between the two systems are identical.

The absolute value of the atomic forces as calculated by LAMMPS using a "compute 1 all property/atom fx fy fz" command in simple trimer systems with only angle potentials match the atomic forces calculated by stress/mop (confirmed via an additional line *std::cout << "The forces are " << duang*dcos_theta[0] << " " << duang*dcos_theta[1] << " " << duang*dcos_theta[2] << std::endl;* before line 737.

**Related Issue(s)**

closes #3683 

**Author(s)**

Evangelos Voyiatzis @ NovaMechanics Ltd (evoyiatzis@gmail.com)

**Licensing**

**Backward Compatibility**

Maintains backward compatibility

**Post Submission Checklist**

[in.test_bonds_mop.txt](https://github.com/lammps/lammps/files/11564437/in.test_bonds_mop.txt)


evoyiatzis:
The failing unit test for MacOS is computing the stress using the "stress/mop" command in a molecular system where there are both non-bonded Lennard-Jones and harmonic bond stretching contributions to the stress tensor. The current version of stress/mop neglects the contribution of the bond stretching terms while the new version does take into account this contribution. I think the expected values at lines 228-229-230 at test_compute_global.cpp should be corrected.

akohlmey:
@evoyiatzis thanks for your contribution. However, I have one concern: Your changes break backward compatibility. How about adding optional keywords "pair", "bond", "angle" similar to how they exist for "compute pressure" that select which contribution to include? The default setting would be the same as if only "pair" was specified. That would add some more flexibility *and* maintain backward compatibility. The warnings about using the compute with unsupported styles could be turned into hard errors instead. That would be more in line with LAMMPS behavior elsewhere.

@athomps Do you have any thoughts on this?

athomps:
@evoyiatzis I think making bonded contributions optional is a good idea, *unless* it makes no sense to use `fix stress/mop` without bonded contributions. Is there a valid use case of this method for bonded systems in which only non-bonded force contributions are tallied by `fix stress/mop`?

evoyiatzis:
> @evoyiatzis I think making bonded contributions optional is a good idea, _unless_ it makes no sense to use `fix stress/mop` without bonded contributions. Is there a valid use case of this method for bonded systems in which only non-bonded force contributions are tallied by `fix stress/mop`?

The only valid use case I can think of is a "droplet" of a molecular substance which in contact with a solid substrate and someone might be interested in the force/pressure between the droplet and the substrate. The surface separating the two entities might not be flat and using stress/mop would cut through the droplet or the substrate. Nevertheless, someone is better off using the group/group command than stress/mop. For the systems I am dealing with (pure polymers/blends/in contact with particles/...), omitting the bond contributions simply leads to wrong results.

evoyiatzis:
> @evoyiatzis thanks for your contribution. However, I have one concern: Your changes break backward compatibility. How about adding optional keywords "pair", "bond", "angle" similar to how they exist for "compute pressure" that select which contribution to include? The default setting would be the same as if only "pair" was specified. That would add some more flexibility _and_ maintain backward compatibility. The warnings about using the compute with unsupported styles could be turned into hard errors instead. That would be more in line with LAMMPS behavior elsewhere.

I can add the optional keywords - I would like to have a keyword like 'conf' that would include all bond/angle/pair contributions since this the option I am going to use.

In my view, it sounds inconsistent to have errors instead of warnings for unsupported styles. The default option is going to be 'pair' (so neglecting bond/angle/....) in any kind of system yet LAMMPS will complain for unsupported bond/angle/... styles although these contributions are going to be ignored by default. In any case, converting warnings to errors is easy and I can take care of it.

athomps:
> The only valid use case I can think of is a "droplet" of a molecular substance which in contact with a solid substrate...

For molecular systems, it is possible to define a valid statistical thermodynamic pressure based on a molecular virial expression written in terms of products of molecular positions and corresponding intermolecular forces.  This molecular virial is mathematical and numerically distinct from the atomic virial, and the corresponding kinetic contribution has 3N_mol degrees of freedom, rather than 3_Natom DOFs. Could MOP be used to localize this molecular virial?

evoyiatzis:
> For molecular systems, it is possible to define a valid statistical thermodynamic pressure based on a molecular virial expression written in terms of products of molecular positions and corresponding intermolecular forces. This molecular virial is mathematical and numerically distinct from the atomic virial, and the corresponding kinetic contribution has 3N_mol degrees of freedom, rather than 3_Natom DOFs. Could MOP be used to localize this molecular virial?

No idea !

athomps (approved):
I approve.

akohlmey:
Before I start the "technical" review, @evoyiatzis please confirm that this is your "final" version and no further updates to the pull request branch are pending. Thanks.

evoyiatzis:
@akohlmey no further updates on this PR are pending. It is the "final" version

evoyiatzis:
@akohlmey & @athomps  thanks for your time and effort!

Issue #3765, opened 2023-05-02
[BUG] stress/cartesian crashes with `free(): invalid next size (fast)`


Compizfox:
**Summary**

LAMMPS crashes when running a simulation with the following compute active:

```
compute p all stress/cartesian z 0.25
fix     p all ave/time 100 100 10000 c_p[*] file p_cartesian.dat mode vector
```

**LAMMPS Version and Platform**
```
Large-scale Atomic/Molecular Massively Parallel Simulator - 8 Feb 2023                                                                                                                                                                                                        
Git info (release / patch_8Feb2023)                                                                                                                                                                                                                                           

OS: Linux "Debian GNU/Linux 11 (bullseye)" 6.0.0-0.deb11.2-amd64 x86_64                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                              
Compiler: Clang C++ Debian Clang 11.0.1 with OpenMP 5.0                                                                                                                                                                                                                       
C++ standard: C++11                                                                                                                                                                                                                                                           
MPI v3.1: Open MPI v4.1.0, package: Debian OpenMPI, ident: 4.1.0, repo rev: v4.1.0, Dec 18, 2020                                                                                                                                                                              
```
**Expected Behavior**

LAMMPS not to crash.

**Actual Behavior**

LAMMPS crashes with the following error:

```
   Step          Temp          PotEng         TotEng         Press        c_Tmobile       CPULeft                                                                                                                                                                             
     21326   0.83354043     10.052026      11.302334      0.0033033425   0.83854251     0                                                                                                                                                                                     
     30000   0.8456891      10.107292      11.375823     -0.028525957    0.85081267     1315.3483                                                                                                                                                                             
     40000   0.84578774     10.107489      11.376167     -0.028547873    0.8509123      1193.2562                                                                                                                                                                             
     50000   0.84342135     10.110402      11.375531     -0.038234715    0.84852224     1054.1164                                                                                                                                                                             
     60000   0.84455696     10.111535      11.378368     -0.0076304926   0.8496692      909.15199                                                                                                                                                                             
     70000   0.84445594     10.108274      11.374954     -0.048069554    0.84956718     761.98723                                                                                                                                                                             
     80000   0.84509014     10.109867      11.3775       -0.048957145    0.85020772     614.02258                                                                                                                                                                             
     90000   0.8443078      10.106927      11.373385     -0.024884455    0.84941756     465.5284                                                                                                                                                                              
[...]

PR #3797, opened 2023-05-25
Include bond forces in `compute stress/cartesian`

Compizfox:
**Summary**

This PR modifies `compute stress/cartesian` to include bond forces.

**Related Issue(s)**

This was discussed at the end of https://github.com/lammps/lammps/issues/3765#issuecomment-1534959286 and in [this](https://matsci.org/t/extending-stress-cartesian-to-include-bond-forces-how-to-grow-num-bond-bond-atom-arrays-to-include-ghost-atoms/48736) MatSci forum thread.

**Author(s)**

Lars Veldscholte, University of Twente, l.b.veldscholte@utwente.nl

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

Below the loop over pairs in the neighbor list, I added a loop over bonds (`neighbor->bondlist`). Similar to the above loop for pair forces, I skip the iteration if one of both atoms are not in the current group, and for ghost atoms I make sure that the computation is only done on the appropriate processor when `newton_bond == 0`. Barring that, I compute the bond force using `force->bond->single()` and call `compute_pressure()` using that force.

I also refactored the `compute_array()` and `compute_pressure()` methods somewhat, mostly by removing unnecessary copies of variables (especially if they're only used once) and by declaring the variables locally (instead of at the top of the method) so that they are properly scoped.

I checked the correctness using a system of polymers. The averaged pressure profiles correspond to the global pressure as output by `compute pressure`.

**Post Submission Checklist**


src/EXTRA-COMPUTE/compute_stress_cartesian.cpp
```diff
@@ -229,35 +230,16 @@ void ComputeStressCartesian::init_list(int /* id */, NeighList *ptr)
...
-  int i, j, ii, jj, inum, jnum, itype, jtype;
-  int bin, bin1, bin2;
-  tagint itag, jtag;
-  double xtmp, ytmp, ztmp, delx, dely, delz;
-  double rsq, fpair, factor_coul, factor_lj;
-  int *ilist, *jlist, *numneigh, **firstneigh;
-
   double **x = atom->x;
   double **v = atom->v;
-  double *mass = atom->mass;
```
akohlmey, line 241: Please keep these cached values from Atom class members and other pointers and values. It makes for an easier read of the code, keeps it consistent with other parts of LAMMPS, and has a (small) performance benefit since it reduces the need to chase pointers in every loop iteration.

akohlmey (changes requested):
For backward compatibility, it would be preferred to have optional keywords (pair, bond, ...) that turn on individual contributions, similar to how it is done in compute pressure. Only the default setting would be "pair" to maintain backward compatibility. Please see the discussion of a very similar issue in PR #3792.
@athomps we should make this consistent across both pull requests.

athomps:
@Compizfox I think making bonded contributions optional is a good idea, unless it makes no sense to use fix stress/cartesian without bonded contributions. Is there a valid use case of this method for bonded systems in which only non-bonded force contributions are tallied by fix stress/cartesian?

src/EXTRA-COMPUTE/compute_stress_cartesian.cpp
Compizfox: I sincerely doubt it makes any difference for performance (with modern compilers, but I'll be happy to be proved wrong ;)), but I'll revert the copied pointers for consistency with the rest of LAMMPS if so desired. :)

Compizfox:
> @Compizfox I think making bonded contributions optional is a good idea, unless it makes no sense to use fix stress/cartesian without bonded contributions. Is there a valid use case of this method for bonded systems in which only non-bonded force contributions are tallied by fix stress/cartesian?

Hmm, I'm not sure if **not** taking into account bond forces for bonded systems ever makes sense. As far as I know, it simply results in a wrong force calculation, but maybe there are some edge cases that I don't know about where it does make sense. I can't think of any, though...

Maybe make it optional, but opt-out? [`compute stress/atom` also does this](https://docs.lammps.org/compute_stress_atom.html):

> If no extra keywords are listed, the kinetic contribution all of the virial contribution terms are included in the per-atom stress tensor. If any extra keywords are listed, only those terms are summed to compute the tensor.

athomps:
Yes, `stress/atom` is a good precedent to follow.  Can you also update doc page to mention that this compute only handles pair and bond contributions, so if contributions from other energy, such as kspace, angles etc. are used in the global virial calculation, these will not be accounted for by `stress/cartesian`

athomps (changes requested):
See comment.

Compizfox:
> Yes, `stress/atom` is a good precedent to follow. Can you also update doc page to mention that this compute only handles pair and bond contributions, so if contributions from other energy, such as kspace, angles etc. are used in the global virial calculation, these will not be accounted for by `stress/cartesian`

OK, working on it.

Would it make sense to split out the documentation for `compute stress/cartesian` from that of `compute stress/cylinder` and `compute stress/spherical`, since some of the stuff now only applies to the first and not to the other two? Otherwise, the combined documentation page might get a bit confusing maybe.

athomps:
> Would it make sense to split out the documentation for `compute stress/cartesian` from that of `compute stress/cylinder` and `compute stress/spherical`, since some of the stuff now only applies to the first and not to the other two? Otherwise, the combined documentation page might get a bit confusing maybe.

Ah, I forgot about those. Instead of breaking out the documents, it might not be much harder to add support for bonds to  `compute stress/cylinder` and `compute stress/spherical`

Compizfox:
> Ah, I forgot about those. Instead of breaking out the documents, it might not be much harder to add support for bonds to `compute stress/cylinder` and `compute stress/spherical`

I've looked into it, but `stress/cylinder` and `stress/spherical` don't exactly follow exactly the same structure as `stress/cartesian` (if they did, it would make sense for them to share significant code) and look considerably more complicated. Frankly, I do not immediately see how to also quickly add support for bonds for those two.

Compizfox:
@athomps @akohlmey is the PR alright like this?

doc/src/compute_stress_cartesian.rst
```diff
@@ -0,0 +1,118 @@
...
+"""""""""""
+
+Compute *stress/cartesian* defines computations that calculate profiles of the
+diagonal components of the local stress tensor over one or two Cartesian
+dimensions, as described in :ref:`(Ikeshoji)<Ikeshoji2>`. The stress tensor is
+split into a kinetic contribution :math:`P^k` and a virial contribution
+:math:`P^v`. The sum gives the total stress tensor :math:`P = P^k+P^v`.
+This compute obeys momentum balance through fluid interfaces. They use the
+Irving--Kirkwood contour, which is the straight line between particle pairs.
+
```
akohlmey, line 43:
```suggestion

.. versionadded:: TBD

   Added support for bond styles

```

akohlmey:
> @athomps @akohlmey is the PR alright like this?

Not entirely. You need to correct the links in the manual that link to the compute_stress_profile.rst file that now link to compute_stress_curvilinear.rst. That would be at the very least in `compute.rst`, and `Commands_compute.rst`.
In the "Related commands" section you add mutual references to the compute styles in the two files. I've also added a suggestion for adding the `.. versionadded` tag to indicate the (future) release date of this modification, so people with older LAMMPS versions that use the online manual can see why their version may be missing this feature. TBD will be replaced in the last step before a release when version.h is updated.

Compizfox:
> Not entirely. You need to correct the links in the manual that link to the compute_stress_profile.rst file that now link to compute_stress_curvilinear.rst. That would be at the very least in `compute.rst`, and `Commands_compute.rst`. In the "Related commands" section you add mutual references to the compute styles in the two files. I've also added a suggestion for adding the `.. versionadded` tag to indicate the (future) release date of this modification, so people with older LAMMPS versions that use the online manual can see why their version may be missing this feature. TBD will be replaced in the last step before a release when version.h is updated.

Ah, I missed that. Done.

akohlmey:
> Ah, I missed that. Done.

From here on the final word is with @athomps. From the formal/technical point of view this is ready.

athomps (approved):
I approve.

akohlmey (approved):

PR #3810, opened 2023-06-08
Create atoms: overlap keyword for atomic molecule

jrgissing:
**Summary**

when using the 'overlap' keyword of the create_atoms command, search for overlaps on an atom-by-atom basis when inserting molecules. previously, molecules were approximated as spheres, making use of the the overlap keyword less useful in many cases when molecule is not spherical

**Related Issue(s)**

none

**Author(s)**

JG

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**


doc/src/create_atoms.rst
```diff
@@ -401,11 +401,9 @@ to.
...
 newly created particles from being created closer than the specified
-*Doverlap* distance from any other particle.  When the particles being
-created are molecules, the radius of the molecule (from its geometric
-center) is added to *Doverlap*.  If particles have finite size (see
-:doc:`atom_style sphere <atom_style>` for example) *Doverlap* should
-be specified large enough to include the particle size in the
+*Doverlap* distance from any other particle.  If particles have finite
+size (see :doc:`atom_style sphere <atom_style>` for example) *Doverlap*
+should be specified large enough to include the particle size in the
 non-overlapping criterion.
```
sjplimp: Add a sentence:  If molecules are being randomly inserted, then an insertion is only accepted if each particle in the molecule meets the overlap criterion with respect to other particles (not including particles in the molecule itself).

src/create_atoms.cpp
```diff
@@ -800,25 +805,54 @@ void CreateAtoms::add_random()
...
+            domain->minimum_image(delx, dely, delz);
+            distsq = delx * delx + dely * dely + delz * delz;
+            if (distsq < odistsq) {
+              reject = 1;
+              break;
+            }
+          }
+        } else {
+          int incr;
+          if (comm->me == 0) {
```
sjplimp: I don't understand why one proc creates the molecule, then Bcasts its coords to the others.  Similar to the random generation of xone above, won't all procs create the same randomly oriented molecule?

src/create_atoms.cpp
```diff
@@ -845,7 +879,11 @@ void CreateAtoms::add_random()
       if (mode == ATOM) {
         atom->avec->create_atom(ntype, xone);
       } else {
-        add_molecule(xone);
+
+        // atomic coordinates calculated above for overlap check
+
+        if (!overlapflag) get_xmol(xone);
```
sjplimp, line 884: I don't understand why get_xmol() is called again?  At this point, isn't the previously generated molecule ready to insert (and known to all procs), and you just need to call add_molecule() ?

src/create_atoms.cpp
```diff
@@ -845,7 +879,11 @@ void CreateAtoms::add_random()
...
+        add_molecule();
```
sjplimp, line 885: Also, does this logic in this code section mean that the proc which owns the center pt of the molecule is adding all atoms in the molecule to it's list of atoms?  It seems like for a large molecule (which extended over multiple proc subdomains), that could lead to lost atoms when comm->exchange() is eventually called.  If that's the case, maybe it would be better for each proc to only add the atoms from the molecule which are in its subdomain?

sjplimp (changes requested):

src/create_atoms.cpp
jrgissing: I was probably worried that it is possible that different processors could call get_xmol a different number of times, resulting in ranmol->uniform getting out of sync.

src/create_atoms.cpp
jrgissing: i think get_xmol was only called if overlapflag == 1

src/create_atoms.cpp
jrgissing: good question. You are suggesting it is a bug prior to this PR, right? If so, I would be surprised that no one has tried to add a large molecule (I know I have, but maybe not in parallel).

src/create_atoms.cpp
sjplimp: ah, I see that get_xmol() uses ranmol and not random
Which means doing it on proc 0 followed by a Bcast is needed

I don't think xmolbuf and the double copy is needed. You can just do
MPI_Bcast(&xmol[0][0],onemol->natoms*3,...)

src/create_atoms.cpp
sjplimp: thanks, I see the logic now

src/create_atoms.cpp
sjplimp: yes, this is a prior bug
the same issue exists for adding molecules on a lattice via get_xmol() and add_molecule() in loop_lattice()
I don't see an easy way to solve the problem there since multiple procs are inserting different molecules simultaneously

However for random and single, the insertion is one molecule at a time.

So I think a good solution is to add a 2nd add_molecule(double **sublo, double **subhi)
method and have add_single() and add_random() call that one instead when mode == MOLECULE.
In the new method, loop over the atoms in the molecule and put the logic for testing against sublo/hi
inside the loop, so that each proc adds only the atoms it owns.  So the caller does not do the test for xone
against sublo/subhi when mode == MOLECULE.

And we should clarify on the doc page that if the lattice method is used to insert mols 2x larger (in any dim) than proc subdomain sizes, lost atoms could result.  But that should not be a problem for single and random mode.

sjplimp: wait, cancel that last comment.

I think we are OK, because after all molecules are created, the top-level method does an irregular comm to ensure
atoms go to owning procs.  That will not lose atoms, even if the molecule is big, b/c it does not do stencil comm with only neighboring procs.

src/create_atoms.cpp
jrgissing: I think I was worried that this array was not guaranteed to be contiguous. But, looking at memory.h it looks like we should be okay. Still, just want to confirm that you agree before making the change

sjplimp: if you mean is the 3*n data contiguous starting in xmol at location xmol[0][0], then yes it is
That is true for all multidimensional arrays allocated by the Memory class.
That was actually one of the main motivations for the Memory class.

sjplimp (approved):

sjplimp:
@jrgissing thanks for the changes
@stanmoore1 I think this is ready for you to look at and merge

PR #3811, opened 2023-06-09
PyLammps and Library interface changes

rbberger:
**Summary**

* adds new property `last_thermo_step` to Python interface to retrieve the last timestep where thermodynamic data was computed
* updates PyLammps to make use of `last_thermo` API and avoid output parsing
* add Fortran unit test for `last_thermo` Fortran API
* extend `extract_global` to return `atom_style`, `pair_style`, `bond_style`, `angle_style`, `improper_style`, and `kspace_style`
* Update PyLammps to make better use of newer library API

**Related Issue(s)**

**Author(s)**

@rbberger 

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/src/lmptype.h
+++ b/src/lmptype.h
@@ -276,8 +276,21 @@ union ubuf {
 \endverbatim
   */
 struct multitype {
-  // same values as LAMMPS_INT, LAMMPS_DOUBLE, and LAMMPS_INT64 in library.h
-  enum { NONE = -1, INT = 0, DOUBLE = 2, BIGINT = 4 };
+  /** Data type constants for extracting data from atoms, computes and fixes
+   *
+   * This enum must be kept in sync with the corresponding enum or constants
+   * in ``python/lammps/constants.py``, ``fortran/lammps.f90``, ``tools/swig/lammps.i``,
+   * ``src/library.h``, and ``examples/COUPLE/plugin/liblammpsplugin.h`` */
+  enum _LMP_DATATYPE_CONST {
+    LAMMPS_NONE = -1,     /*!< no data type assigned (yet) */
+    LAMMPS_INT = 0,       /*!< 32-bit integer (array) */
+    LAMMPS_INT_2D = 1,    /*!< two-dimensional 32-bit integer array */
+    LAMMPS_DOUBLE = 2,    /*!< 64-bit double (array) */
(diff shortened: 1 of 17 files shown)


src/lmptype.h
```diff
@@ -47,6 +47,8 @@
 #include <cstdint>      // IWYU pragma: export
 #include <cstdlib>      // IWYU pragma: export
 
+#include "library.h"
```
akohlmey: Including library.h into lmptype.h is not a good idea. lmptype.h is a "fundamental" header and its contents should be included before others.

Similarly, I would prefer that library.h does not include any LAMMPS headers, so that the library interface is self-contained. That would kill the possible solution to have a new header that can be used by both. It also needs to be usable from C.

I agree with the idea of using the constants from _LMP_DATATYPE throughout and thus avoiding the conversion. Unfortunately, taking all constraints into account would mean we would have to replicate the enum (I would just copy the entire enum and use the same constant names). But this is a section of code that is already replicated, so one more replication is acceptable in my book, especially since this is not a central feature and not likely to be used much beyond PyLammps and including thermo info into dump styles.

rbberger: Ok. Removed the include and duplicated the 3 constants.

rbberger:
@akohlmey Note, there might still be a few more changes coming in the coming days, so maybe not merge this right away.

rbberger:
@akohlmey I think this is as far as I can push it without adding a significant chunk of new introspection functionality to the library interface. There are a few holes in terms of retrieving metadata from computes, fixes, and dumps as well as some things that only the `info` command is currently providing. Not sure how much appetite there is to go in that direction.

In particular, 2f7c3bf959b8ebd11e9001b2e610f39acebbdc1b is a mixed bag. It kind of serves as a testament to the current state of affairs. If you're unhappy about this mixed state, I can go ahead and revert that commit, and keep it in a local branch for now.

akohlmey:
@rbberger I am OK with adding more introspection based on the assumption that if it is useful to PyLammps there will likely also be a use for some other scripting. However, that would be something to address in a couple of months or so, *after* the upcoming stable release.

My long term perspective for PyLammps is to either get rid of having to parse screen output or to retire PyLammps. The screen output parsing is predominantly what makes it fragile, slow, and sometimes difficult to maintain and to keep portable.

sjplimp (approved):

PR #3823, opened 2023-06-16
Collected small fixes

akohlmey:
**Summary**

This pull request collects multiple small fixes and improvements

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues

**Implementation Notes**

- flag LAMMPS version as development version
- improve error messages for dump styles derived from dump style custom
- better detection of unsupported properties in dump style image
- include git descriptor in development and maintenance version output
- fix segfault when attempting to delete Atom class callback in base class after error in constructor of derived class
- address issues flagged by static code analysis
- fix syntax issue in CMake script that prevented testing the Fortran coupling simple example

**Post Submission Checklist**


stanmoore1:
Could we print out the git sha (if available) in the log file? I think that would be useful for many reasons, especially when we have months in between feature releases.

stanmoore1:
And more convenient if you only have a log file and not the executable which can run `lmp -h`

akohlmey:
> Could we print out the git sha (if available) in the log file? I think that would be useful for many reasons, especially when we have months in between feature releases.

The current situation is at the request of @sjplimp. I had the git hash included originally, IIRC. I am happy to revert.

akohlmey:
@stanmoore1 what do you think of the change in commit a3de790a0da60c1ff13a39e0617bc75f45bdb9b1?

sjplimp (approved):

PR #3824, opened 2023-06-17
Fix reaxff/species fixes

jrgissing:
**Summary**

1) Automatically determine elements for the fix reaxff/species command from the reaxff pair_coeff command
2) Correctly report unique species when there are duplicate elements in the element-to-type mapping

**Related Issue(s)**

1) previously, the default elements were C H O N, and there was no check to see if the number of types was correct, resulting in species that could be wrong, garbled or nonsense. this fix also prevents requiring a redundant list of elements
    1.5) CHON elements are printed in that order, if they exist, for consistency, and to match common small organic molecule names
2) previously, reaxff/species reported duplicate species when there are duplicate elements in the element-to-type mapping. for example, H2 and HH and multiple other H2s and HHs could appear in the species.out file. These are now deduplicated to avoid needing to post-process (or pre-process the input). this also corrected the 'total number of distinct species' output value

**Author(s)**

JG

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**


stanmoore1 (commented):
Two concerns:
- why was `ntypes` renamed to `nutypes`? I know what `ntypes` means in the rest of LAMMPS, but have no idea what `nutypes` means, maybe a more self-explanatory name would be better?
- We normally don't use `std::vector` for arrays in LAMMPS, but I think in this case it might be alright--I defer to @akohlmey on this.

jrgissing:
> * why was `ntypes` renamed to `nutypes`? I know what `ntypes` means in the rest of LAMMPS, but have no idea what `nutypes` means, maybe a more self-explanatory name would be better?

'nutypes' refers to the number of unique types (elements, in this case). I was trying to keep it short due to appearing in many 'for' statements. if you wish, I can change to something like 'ntypes_unique'

akohlmey:
@jrgissing FYI, the changes in this pull request are related to some project that I have on my mind for the next development cycle. So that may require some changes, but also should make things simpler.

In the MDI coupling of codes, are already multiple lists of per-chemical-element properties and periodic tables used. The idea is to make this a more generic functionality (utilizing algorithms and data already available in VMD and TopoTools) to have some automatic translation. The idea is to have a namespace where related data is stored in static, read-only tables and a set of functions related to them. With that there could be an "element" command, that would work similar to the "mass" command to associate elements with types, but unlike typelabels, there would by no requirement for unique mappings. In addition, there could be something like `element * guess` where elements would be guessed based on mass. This information could then be used for the "element" property for dumps (and thus removing the requirement to set them with dump_modify. We could also seed the element map for manybody potentials from the type map defined in the pair_coeff command and thus making extracting the info from the pair style itself obsolete. There would also be the option to implement an element to core charge translation to be used by ZBL potentials.
... and probably some more.

jrgissing:
I like those ideas, sounds useful for bond order potentials. i wonder if it would be helpful to piggyback off of the type labels framework? not allowing repeat type labels was not a limitation of the framework, that restriction was just added to avoid ambiguity

jrgissing:
by the way, I have fielded multiple other cases of people forgetting the 'optional' *element* keyword when using 'fix reaxff/species', similar to this thread:

https://matsci.org/t/incorrect-chemical-formula-in-species-out-file/49623

so, I would suggest considering this PR for the upcoming stable release

akohlmey:
> so, I would suggest considering this PR for the upcoming stable release

Sorry, but it is too late in the process for that. We are in "no new features"-mode for a month already and very close to the stable release. There can be exceptions for small changes in peripheral and less used parts of LAMMPS, but ReaxFF is a very popular feature and thus the risk of breaking more than what we would fix is too large. We have survived without this for over 10 years; one more doesn't make that much of a difference. Besides, considering how popular ReaxFF in LAMMPS is, the number of people struggling with correct/incorrect elements is rather small. 

In comparison, a more comprehensive implementation that is also extended to other outputs like dump files and has features like guessing elements from atom masses or a generic interface for reading elements from pair styles would have a deeper impact and would make this more general.

stanmoore1:
> > * why was `ntypes` renamed to `nutypes`? I know what `ntypes` means in the rest of LAMMPS, but have no idea what `nutypes` means, maybe a more self-explanatory name would be better?
> 
> 'nutypes' refers to the number of unique types (elements, in this case). I was trying to keep it short due to appearing in many 'for' statements. if you wish, I can change to something like 'ntypes_unique'

I see, thanks for the explanation.

stanmoore1 (approved):

doc/src/fix_reaxff_species.rst
```diff
@@ -110,10 +110,10 @@ all types from 1 to :math:`N`.  A leading asterisk means all types from
 The optional keyword *element* can be used to specify the chemical
 symbol printed for each LAMMPS atom type. The number of symbols must
 match the number of LAMMPS atom types and each symbol must consist of
-1 or 2 alphanumeric characters. Normally, these symbols should be
-chosen to match the chemical identity of each LAMMPS atom type, as
-specified using the :doc:`reaxff pair_coeff <pair_reaxff>` command and
-the ReaxFF force field file.
+1 or 2 alphanumeric characters. By default, these symbols are the same
+as the chemical identity of each LAMMPS atom type, as specified by the
+:doc:`reaxff pair_coeff <pair_reaxff>` command and the ReaxFF force
```
sjplimp: ReaxFF, since it is the name of the FF, not the name of the LAMMPS command.

doc/src/fix_reaxff_species.rst
```diff
@@ -233,5 +233,5 @@ Default
 """""""
 
 The default values for bond-order cutoffs are 0.3 for all I-J pairs.
-The default element symbols are C, H, O, N.
+The default element symbols are taken from the pair_coeff command.
```
sjplimp: Change this to "the ReaxFF pair_coeff command"

src/REAXFF/fix_reaxff_species.h
```diff
@@ -62,7 +62,10 @@ class FixReaxFFSpecies : public Fix {
   int delete_Nsteps, *delete_Tcount;
   double massmin, massmax;
   int singlepos_opened, multipos_opened, del_opened;
-  char *ele, **eletype, *filepos, *filedel;
+  char *filepos, *filedel;
+  std::vector<int> ele2uele;
```
sjplimp: Can you add a trailing comment for this vector?
I.e. what is the Ith quantity stored in ele2uele[I] an index into?
In the cpp file it appears both of the 2 next vectors are ntypes in length, so it is not clear.

doc/src/fix_reaxff_species.rst
```diff
@@ -110,10 +110,10 @@ all types from 1 to :math:`N`.  A leading asterisk means all types from
 The optional keyword *element* can be used to specify the chemical
```
sjplimp, line 110: On the "keyword =" line at the top of this file can you add "... or delete or delete_rate_limit"

src/REAXFF/fix_reaxff_species.h
sjplimp: Trailing comments for the other 2 vectors would be good as well since it is not clear how they are different (as in previous comment).  Maybe ueletype is actually of only length nutype?

sjplimp:
@jrgissing This PR looks like it encodes changes and bug fixes to this comand.  I made a few minor comments/Qs.

@stanmoore1 Do these changes require the same changes in the Kokkos version?  Once Jake has addressed my comments, I will review it.  Can you also then review and merge it?

sjplimp:
@jrgissing and @stanmoore1 Does this new command syntax induce changes in any example scripts or regression tests ?
E.g. removal of the element keyword/values since they are not needed?

doc/src/fix_reaxff_species.rst
jrgissing: I am not sure what you want changed here, if anything

jrgissing:
> @jrgissing and @stanmoore1 Does this new command syntax induce changes in any example scripts or regression tests ? E.g. removal of the element keyword/values since they are not needed?

looks like all current examples happened to use the default 'C H O N' elements, so did not use an 'elements' keyword (perhaps part of what is causing confusion among users)

doc/src/fix_reaxff_species.rst
jrgissing: oh, I see your point. will correct

src/REAXFF/fix_reaxff_species.h
jrgissing: yes, ueletype should only have 'nutypes' values. I think I initialized it to length ntypes to do so at the same place as the other vectors, at which point nutypes is not known yet (noting that nutypes is always <= ntypes)

sjplimp (approved):

PR #3826, opened 2023-06-18
debug fix_modify press for press/berendsen

srtee:
**Summary**

 `fix press/berendsen` had a bug preventing it from accepting a valid pressure compute via `fix_modify press`. This PR fixes it. See https://matsci.org/t/a-disconnect-between-compute-pressure-and-press-berendsen-when-bpm-package-is-used-in-2023-feature-lammps/49034/26 for bug in action.

**Author(s)**

Shern Tee (Griffith University)

**Licensing**

**Post Submission Checklist**

### DIFF
--- a/src/compute_chunk_spread_atom.cpp
+++ b/src/compute_chunk_spread_atom.cpp
@@ -106,7 +106,7 @@ ComputeChunkSpreadAtom(LAMMPS *lmp, int narg, char **arg) :
 
     } else if (val.which == ArgInfo::FIX) {
       auto ifix = modify->get_fix_by_id(val.id);
-      if (ifix)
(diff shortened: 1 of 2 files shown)


akohlmey:
@srtee Good catch. I've checked the output of `grep -n -E -A2 -- '->get_(fix|compute)_by_id' src/*.cpp src/*/*.cpp` and found one more case of the same logic error.

Perhaps you can think of other search patterns or have a second look at that grep output, in case there is still some bug left. There also are the get_xxx_by_style() APIs that return a vector, so the vector length may be checked for incorrectly.

sjplimp (approved):

Issue #3140, opened 2022-02-18
[Feature Request] Use runtime detection for `have_gpu_aware`


ocaisa:
**Summary**

Currently `have_gpu_aware` is set at compile time, when this could actually be detected at runtime using `MPIX_Query_cuda_support()` which is supported by a number of MPI implementations (see https://github.com/pmodels/mpich/pull/4741).

**Detailed Description**

We are investigating a number of things right now related to MPI. One is using the dynamic addition of CUDA support to OpenMPI _after_ compilation (see https://github.com/easybuilders/easybuild-easyconfigs/pull/14919) and another is looking into leveraging [MPItrampoline ](https://github.com/eschnett/MPItrampoline) to facilitate dynamic switching of MPI runtimes. LAMMPS currently doesn't support this (without patching) as the detection of CUDA support in the MPI runtime is done at compile time, rather than dynamically at runtime (using `MPIX_Query_cuda_support()` which seems to be quite portable). 

**Further Information, Files, and Links**

Example of how this was implemented for GROMACS is at https://gitlab.com/gromacs/gromacs/-/commit/ae95fd7004aa5f4fbf6685a315f35b2dd58fe436

ocaisa:
I guess this is actually a Kokkos issue but I couldn't find the place to raise the issue in their repos.

akohlmey:
> I guess this is actually a Kokkos issue but I couldn't find the relevant LoCs in `kokkos.cpp` in their repos.

I have assigned this issue to the LAMMPS/Kokkos developer for further comments.
Please note that LAMMPS is trying to be out-of-the-box compatible to a wide range of platforms including some rather old ones like RHEL 7.x and Ubuntu 18.04LTS while at the same time accommodate modern programming paradigms (like KOKKOS) within reason. It mostly works for KOKKOS because it is optional and thus can be skipped on incompatible systems.

So whatever would be done in LAMMPS, there would have to remain a compile time choice, if only to decide whether to use runtime detection (after testing whether it is supported). This would also make it difficult to support outside of the CMake build system. It may be surprising to you, but there is still a very significant number of people compiling LAMMPS using the traditional makefile route. My impression is that the CMake route is most popular with people doing some form of packaging or working on automated deployments. Plenty of LAMMPS users also have been using and compiling LAMMPS before CMake was a viable option and thus have little incentive to switch.

ocaisa:
[...]

PR #3828, opened 2023-06-18
Dynamic check for CUDA and ROCm-aware OpenMPI support in KOKKOS package

vladgl:
**Summary**

This PR replaces the CUDA compile time check and adds a HIP runtime check for GPU-aware capability when compiling KOKKOS with CUDA and HIP backends. This feature for ROCm appeared in OpenMPI since version 5 [(see here)](https://docs.open-mpi.org/en/v5.0.x/man-openmpi/man3/MPIX_Query_rocm_support.3.html#mpix-query-rocm-support).

**Related Issue(s)**

**Author(s)**

Vladislav Galigerov @vladgl [vladgl3@yandex.ru](mailto:vladgl3@yandex.ru) (Higher School of Economics, Russia)

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**


vladgl:
@stanmoore1 Besides checking ROCm aware support, this PR may be partially related to #3140, perhaps we should extend it to a more general solution that includes CUDA?

stanmoore1:
@vladgl thank you for this PR, and sorry for the delay, I will take a look.

stanmoore1:
@vladgl thanks for this PR, the code looks good to me. I'm fine with either merging as-is, or extending this to CUDA. Do you have interest to work on that?

stanmoore1 (approved):

vladgl:
@stanmoore1 Thanks for the reply! Yes, I'll try to do that soon, it seems like it shouldn't be much different from the HIP option.

stanmoore1:
Great, thank you!

vladgl:
@stanmoore1 It seems better to use [MPIX_ROCM_AWARE_SUPPORT](https://github.com/open-mpi/ompi/blob/67026ef03fffdd49decb02aca038b6e7f6c16756/ompi/mpiext/rocm/configure.m4#L24) instead of `OMPI_HAVE_MPI_EXT_ROCM`, i.e. to use the same format as with CUDA. Also, I have removed `if (have_gpu_aware == 1)` as it is redundant.

It is worth noting that MPICH has similar functions [MPIX_Query_cuda_support](https://www.mpich.org/static/docs/v4.1.x/www3/MPIX_Query_cuda_support.html) and [MPIX_Query_hip_support](https://www.mpich.org/static/docs/latest/www3/MPIX_Query_hip_support.html), but unfortunately it has no macro to check for the presence of these functions. To overcome this, we can use `check_cxx_source_compiles` in cmake build system, like [gromacs ](https://gitlab.com/gromacs/gromacs/-/blob/release-2023/cmake/gmxManageGpuAwareMpi.cmake) does.

Also, MVAPICH has [MPIX_Query_cuda_support](https://mvapich.cse.ohio-state.edu/userguide/gdr/#_compile_time_and_run_time_check_for_cuda_aware_support) with macro, but there is no such function for ROCm.

vladgl:
@stanmoore1 At the moment, I would propose to include this pull request for openpmi. The search for a solution for other MPIs will take some effort.

stanmoore1:
@vladgl thank you, I will review and get this merged.

stanmoore1:
@vladgl @IanBogle do either of you have a build of OpenMPI 5.0 with ROCm that supports GPU-aware that could test this?

stanmoore1:
I tested this with CUDA and OpenMPI 4.0, it seems to be working for CUDA.

vladgl:
@stanmoore1 I just tried with OpenMPI 5.0 + ROCm with GPU aware, everything is fine, but I noticed that you have removed the "OMPI_HAVE_MPI_EXT_ROCM/OMPI_HAVE_MPI_EXT_CUDA" guards. I think they are important because someone might have OpenMPI without HIP or CUDA extensions, so they will get "MPIX_Query_rocm_support was not declared in this scope" error.

stanmoore1:
@vladgl thanks I added the guards back in.

stanmoore1:
This is working as far as I can tell, ready to merge from my POV.

PR #3830, opened 2023-06-20

### LINKED ISSUE #3835: [BUG] _Segmentation_Fault_from_npair_half_bin_newton.cpp_
**Summary**

When using a custom (fix) module that requests half pair list and sets a cutoff for the list using `set_cutoff()` function a segmentation fault occurs either every time or occasionally depending on the system. The segmentation fault is thrown on line 128 (`delx = xtmp - x[j][0];`), due to ill defined value of j that originates from assignment `j = binhead[ibin+stencil[k]]` in for command `for (j = binhead[ibin+stencil[k]]; j >= 0; j = bins[j])` on line 124. This is most likely due to memory corruption as far as I can tell.

I have reversed any changes and stripped my module of any code except for pair list request and call of `set_cutoff()` to make sure that the memory corruption does not occure from within the module.

It is also important to note that the latest stable version does not seem to have the issue.

LAMMPS version: LAMMPS (28 Mar 2023 - Update 1) (the release version that was cloned from github at the beginning of June)
The error always occurs on linux system (#40~20.04.1-Ubuntu SMP Mon Mar 7 09:18:32 UTC 2022) with 11th Gen Intel(R) Core(TM) i7-11700 @ 2.50GHz
It also occasionally manifests on MacOS Ventura with M2 processor (approximately once in 20 times)

**Steps to Reproduce**

To provide more information and reproduce the error I prepared a directory with most basic simulation setup (subfolder `simulation_setup`), additional code that requests the pair list and sets cutoff (using one of the values the gives error on my side, though it may occurs at smaller values as well) placed in subdirectory `src`, and log files from LAMMPS itself and Valgrind debug run (in `error_logs` subdirectory, also please note that line numbers after 128 may be shifted due to print that I added on my side to investigate the issue). The latter clearly reports the issues in `npair_half_bin_newton.cpp` that originate on line 124.

The archive of the directory described above is attached.
[bug_report_npair_half_bin_newton.zip](https://github.com/lammps/lammps/files/11904909/bug_report_npair_half_bin_newton.zip)

### LINKED ISSUE #3838: [BUG] Possible bug in install.py?
**Summary**

I think there is a small bug in the `install.py` or I am doing something completely wrong. 

Current LAMMPS develop head.

**Expected Behavior**

I try to install the LAMMPS python bindings via `python3 install.py -p lammps -l ../build/liblammps.so -v ../src/version.h`, which used to work.

**Actual Behavior**

I get the error "ERROR: Could not install packages due to an OSError: [Errno 2] No such file or directory: '/home/fiedlerl/tools/lammps_parallel/python/lammps-2023.3.28-cp310-cp310-linux_x86_64.whl' " 

**Steps to Reproduce**

(See expected behavior)

The wheel does briefly exist in `build-python` folder, which gets deleted in line 113 of the install script. The wheel is not moved to `olddir` - it is only moved when a `wheeldir` is specified, which I do not do. As long as one is specified (which is probably the case for many users), this bug would never occur. I fixed this locally by adding 
```  else:
    shutil.copy(wheel, olddir)
```
beginning in line 108. If that is a useful/correct addition, I can open a PR for this.

### LINKED ISSUE #3841: [BUG] Dividing 0 in agnle_dipole.cpp
**Summary**

In the angle_dipole.cpp code, line 114 " fmod_sqrtff = fmod / sqrt(fx * fx + fy * fy + fz * fz);", the denominator will be 0 if all the components (fx, fy, and fz) are equal to 0. They will be 0 if the dipole vector is intially parallel to the bond vector.

I used lammps-8Feb2023 version. In addition, the code has not been updated in the last 9 months. 

**Expected Behavior**

When all fx, fy, and fz are 0, fmod_sqrtff should be 0 as well. 

**Actual Behavior**

However, the current code gives "nan", leanding to error of simulations. 

**Steps to Reproduce**

This can be reproduced by creating a single angle and assign the dipole is parallel to the bond.

It should be not very difficult to correct this. I introudced a new variable f_sqrtff = sqrt(fx * fx + fy * fy + fz * fz) and set fmod_sqrtff=0 if f_sqrtff=0, without calling the line 114 command.

akohlmey:
**Summary**

This pull request combines multiple small changes and bug fixes

**Related Issue(s)**

fixes #3835 
fixes #3838 
fixes #3841

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues

**Implementation Notes**

- use a longer string variable in simple Fortran coupling unit test to be able to store longer paths and thus avoid incorrect test failures
- print an error message instead of segfaulting when trying to use 'run pre no' after creating a new dump
- put null pointer initialization into initializer lists for dump styles, add missing entries
- avoid segfaults in read_data by better detecting and flagging short reads in utils::read_lines_from_file()
- remove duplicate semicolons
- replace `for (;;)` with `while (true)`
- require *rmax* in *create_bonds many* to be smaller than the box length in periodic dimensions to avoid creating redundant bonds to periodic images of atoms in addition to their originals
- disallow neighbor list requests with a custom cutoff that exceeds the communication cutoff (thus situations like #3835 will stop with an error instead of a segmentation fault)
- fix issue in pair style kim where the neighbor list skin would be added twice
- use and document only fix_modify dynamic/dof and compute_modify dynamic/dof and extra/dof. The old syntax was deprecated six years ago. High time to remove it where it is visible (buy silently stay backward compatible).
- copy created wheel file to current folder is python/install.py is invoked without specifying a wheeldir
- avoid division by zero in angle style dipole

**Post Submission Checklist**


akohlmey:
@sjplimp added the check I talked about on the phone today in commit b7024288b66c3930d87b5cfc1dc825366e0a7390. Was simpler than expected, but exposed some issue with KIM.

akohlmey:
@ellio167 @yafshar  please have a look at the change in ab1e090e0e5aff5c943e25e64bc6e320d58c70a1. This was exposed by the check to disallow custom cutoffs that are larger than the communication cutoff in the commit before it.
This came about after the discussion in #3835 when trying to catch the undesired situation of a custom neighbor list cutoff request that is larger than the communication cutoff in order to avoid a segmentation fault when processing the data from binning the atoms during neighbor list builds.

However, there is a minor issue. The communication cutoff is set by default  to the largest pair cutoff plus neighbor skin  and this is also enforced as a minimum even if you would request a shorter cutoff via comm_modify. However, when setting the custom neighbor list cutoff, one *must not* add the neighbor list skin, since that is added automatically later for perpetual neighbor lists (but not for occasional neighbor lists). Thus it seems that for pair style kim the neighbor list skin was applied twice and after I added the additional check to the neighbor class, the kim pair style in the unit tests failed. Using too large a neighbor list cutoff can cause memory corruption or even segfaults as shown by the bug report in #3835. However, the example given there was very extreme.

Since we are very close to the next stable release of LAMMPS, this needs to be double checked very carefully. At my current understanding of the situation, this has been a long-standing bug that was exposed by the added check. But we need to make certain that there is no other issue that was previously hidden by the double adding of the neighborlist skin and now would be exposed. Can you please check and let me know ASAP?

ellio167:
@akohlmey I think it is OK.  @ilia-nikiforov-umn will run a few spot-checks to see if any issues show up and report back here.

ilia-nikiforov-umn:
> @akohlmey I think it is OK. @ilia-nikiforov-umn will run a few spot-checks to see if any issues show up and report back here.

@ellio167 With and without this PR are confirmed to have identical energies and forces on an Si cluster for the following models:
        EDIP_LAMMPS_JiangMorganSzlufarska_2012_SiC__MO_667792548433_000
        MEAM_LAMMPS_AslamBaskesDickel_2019_FeMnSiC__MO_427873955970_002
        SW_BalamaneHaliciogluTiller_1992_Si__MO_113686039439_005
        Tersoff_LAMMPS_ErhartAlbe_2005_SiC__MO_903987585848_005

Do you think this is sufficient testing, or are dynamics/relaxation required?

ilia-nikiforov-umn:
> > @akohlmey I think it is OK. @ilia-nikiforov-umn will run a few spot-checks to see if any issues show up and report back here.
> 
> @ellio167 With and without this PR are confirmed to have identical energies and forces on an Si cluster for the following models: EDIP_LAMMPS_JiangMorganSzlufarska_2012_SiC__MO_667792548433_000 MEAM_LAMMPS_AslamBaskesDickel_2019_FeMnSiC__MO_427873955970_002 SW_BalamaneHaliciogluTiller_1992_Si__MO_113686039439_005 Tersoff_LAMMPS_ErhartAlbe_2005_SiC__MO_903987585848_005
> 
> Do you think this is sufficient testing, or are dynamics/relaxation required?

I realized that the cluster may not actually be making any computation at the edge of the cutoffs, so I repeated the experiment with one of the KIM tests that use LAMMPS to compute the forces and energies of 64 Si atoms randomly distributed in a periodixc box and once again got identical energies and forces with and without the PR.

ellio167:
Great! I think this is approved from the KIM perspective.

sjplimp (approved):

PR #3846, opened 2023-07-07
Fix occasional bug in neighbor list build for triclinic geometries due to round off

### LINKED ISSUE #2759: [BUG] in rare cases roundoff makes neighbor list fail to find pairs

sjplimp:
**Summary**

As described in issue #2759 and [this matsci.org thread](https://matsci.org/t/apparently-inconsistent-forces-before-after-dump/36305/28), triclinic (non-orthogonal box) models can construct their neighbor list incorrectly in rare instances:

(a) triclinic geometry
(b) I,J atom pairs which straddle simulation box boundary
(c) systems with atoms on a perfect lattice (e.g. the initial config)
(d) binned half neighbor list with newton on

The problem is due to coordinate transformations for owned and ghost atoms from box coords to reduced (lambda) coords and back again, which occurs before a neighbor list is built.  This can (rarely) result in a coordinate of a ghost atom outside the simulation box being epsilon different than a coordinate of the same atom inside the box due to periodic box offsets being added, even when they would ideally be identical.  E.g. in an orthogonal box the Y coordinate will be identical for a ghost atom that straddles the periodic X boundary, but for triclinic its back-and-forth numerical transformation can make them different by epsilon due to round-off.  The neighbor list build method for triclinic geometries assumed binning and exact comparisons of coords would be consistent for an I,J and J,I pair to consistently exclude exactly one of them.

The signature of the bug is that the I.J pair is included in the neighbor list 0 or 2 times, rather than 1, leading to incorrect energy and forces.

The remedy in this PR is twofold.  For triclinic, change the stencil of bins to be a full stencil rather than a half stencil (in the last dimension).  And second, also for triclinic, change the logic in the neighbor list build (binned or Nsq) to only rely on the I,J atom IDs and in the special case (cutoff > box size) where the IDs of the I,J atoms are the same, to rely on large differences in coordinates rather than exact coordinates to make the decision to include/exclude the two I,J pairs consistently.

**Related Issue(s)**

Fixes #2759

**Author(s)**

Steve

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**


stanmoore1 (changes requested):
I need to port this to Kokkos too

sjplimp:
@stanmoore1 you should wait, b/c if may not yet be correct, and b/c once it is I will make the same change to several other NPair classes (those with tri or nsq in the name)

stanmoore1:
OK I will wait, I just wanted to make sure it doesn't get merged without Kokkos updates

akohlmey:
@sjplimp @stanmoore1 I have ported the existing changes to OPENMP, but it looks to me as if some more NPair styles need to be adjusted, e.g. the multi and respa variants, which have similar code patterns. I have not looked at INTEL, but it likely needs porting, too, same as KOKKOS. OPENMP was straightforward since the affected code is mostly identical to the non-OPENMP version.

sjplimp:
@akohlmey @stanmoore1 @jtclemm @wmbrownIntel Please take a look at this PR and its changes to a few NPair and NStencil classes to fix a (rare) issue with triclinic geometries.  I believe I have made all the changes needed for vanilla CPU classes, but not for accelerated classes.  At a code level, these are the rules-of-thumb for the code changes:

(a) for binning, the Nstencil class for triclinic needs to by "full" in all dims, because the code can't rely on an atom's bin being consistent for ghost atoms that straddle periodic boundaries - previous the triclinic stencil was half-size in z (or in y for 2d)
(b) for triclinic, the NPair class cannot compare two atoms (owned and ghost) expecting their coords (in some dim) to be identical, instead logic with the atom IDs (itag, jtag) needs to be used, as well as logic that reflects the full stencil
(c) this is not as simple as only checking NStencil and NPair classes with "tri" in their name, b/c some classes are invoked for both orthogonal and triclinic - e.g. NPair Nsq and HalfFull classes - in those cases exact coord compisons can still be used for orthogonal, but (b) needs to be used for triclinic

@akohlmey Leaving the OPENMP changes to you

@stanmoore1 Leaving the KOKKOS changes to you

@jtclemm I made the changes I thought were correct for the "multi" neighbor classes (and also "multi/old") but please check them over - this will also affect your pending PR that refactors these various classes

@wmbrownIntel I looked at the INTEL variants for triclinic but they invoke a tempted class in the parent which appears to be called for both orthogonal and triclinic - so I didn't want to make any changes in the templated method

sjplimp:
@akohlmey This version fixes all the problem cases Identified in the related issue, as well as the MatSci thread

akohlmey:
@stanmoore1 OPENMP seems to be done now.

stanmoore1:
Does this need a check like in ReaxFF to make sure tags are enabled?
```
  if (atom->tag_enable == 0)
    error->all(FLERR,"Pair style reaxff requires atom IDs");
```
i.e. if the user turns off atom ids with `atom_modify id no`?

sjplimp:
yes, good catch - using triclinic will now require tags - I will add that

src/nstencil_half_multi_3d_tri.cpp
```diff
@@ -81,7 +81,7 @@ void NStencilHalfMulti3dTri::create()
       cutsq = cutcollectionsq[icollection][jcollection];
 
       if (flag_half_multi[icollection][jcollection]) {
-        for (k = 0; k <= sz; k++)
+        for (k = -sz; k <= sz; k++)
```
jtclemm, line 84: While this works, a more consistent method would be to change the logic in `set_stencil_properties()` such that the "same collection" uses a full stencil instead of changing the definition of a half stencil. I.e.:

```
if (cutcollectionsq[i][i] == cutcollectionsq[j][j]) {
        flag_half_multi[i][j] = false; //true;
        bin_collection_multi[i][j] = i;
}
```

Practically, however, this makes no difference. So I can just make this alternate change in the other pull request.

sjplimp: agreed - just make the change in the later PR - thanks for checking

wmbrownIntel:
I don't think I can push here, but I put the changes for the intel package: [https://github.com/wmbrownIntel/lammps/tree/triclinic-neighbor-bug](https://github.com/wmbrownIntel/lammps/tree/triclinic-neighbor-bug )

Please let me know if this doesn't work.

sjplimp:
@wmbrownIntel - thanks Mike

stanmoore1:
@sjplimp @akohlmey do we have a canonical test for this issue? I'm having trouble reproducing with the input in https://github.com/lammps/lammps/issues/2759

stanmoore1:
When I run the latest `develop` branch on the CPU with no Kokkos (i.e. without the bugfix in this PR), I see `Total # of neighbors = 1` which is not expected according to https://github.com/lammps/lammps/issues/2759

stanmoore1:
The `two_atoms` test case gives expected behavior on one machine but not another, probably due to roundoff and a different OS/hardware.

stanmoore1:
Everything is passing, including the test cases in https://github.com/lammps/lammps/issues/2759, except the `nemd` example lost atoms with Kokkos CUDA, need to investigate. Also one of the unit tests failed but it looks unrelated to the changes in this PR.

akohlmey:
> Also one of the unit tests failed but it looks unrelated to the changes in this PR.

Yes, that test failure is due to a race condition in OpenMPI as bundled in the Ubuntu 20.04LTS container. It happens very rarely and it is usually sufficient to rerun the group of tests. I just have tried to trigger this.

stanmoore1:
The `nemd` lost atoms on GPU appears to be a false positive or at least I cannot reproduce it outside of Jenkins. I built on the same machine using the exact same commands and it runs fine. It also runs fine on another GPU testbed. Therefore I believe this PR is ready to merge from my POV.

stanmoore1:
Found the bug, testing again.

stanmoore1:
All regression tests pass now, but I'm seeing a slowdown for the chute benchmark on GPU, looking into it.

stanmoore1 (approved):

stanmoore1:
Performance is restored, ready to merge.

PR #3807, opened 2023-06-06
Misc minor patches/features in BPM/Granular packages

jtclemm:
**Summary**

This PR includes several minor patches or requested features to the BPM and GRANULAR packages including:

1. Adding option to normalize forces in BPM bond styles by reference lengths
2. Adding missing factor of 1/r in the BondBPMRotational->single() method (reported on Matsci)
3. Unifying the syntax for BPM bond styles
4. Adding flag to prevent 2x mixing of coefficients in pair granular (reported on Matsci)
5. Fixing various typos in doc pages

**Related Issue(s)**

Matsci forum topics
https://matsci.org/t/using-bpm-to-simulate-beam/48584/5
https://matsci.org/t/about-mixing-of-coefficients-of-pair-style-granular/48694/7

**Author(s)**

Joel Clemmer (SNL)

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

Input scripts for BPM bond styles and granular pair styles with mixing coefficients may need to be updated.

**Post Submission Checklist**

- [ ] The feature or features in this pull request is complete
- [ ] Licensing information is complete
- [ ] Corresponding author information is complete
- [ ] The source code follows the LAMMPS formatting guidelines
- [ ] Suitable new documentation files and/or updates to the existing docs are included
- [ ] The added/updated documentation is integrated and tested with the documentation build system
- [ ] The feature has been verified to work with the conventional build system
- [ ] The feature has been verified to work with the CMake based build system
- [ ] Suitable tests have been added to the unittest tree.
- [ ] A package specific README file has been included or updated
- [ ] One or more example input decks are included

**Further Information, Files, and Links**

NA

PR #3850, opened 2023-07-10
BPM package: allow bonded atoms to be excluded from neighbor lists if BPM cannot break + doc update

jibril-b-coulibaly:
**Summary**

1. Make error message consistent with new keyword values yes/no from #3807 
2. Allow special_bonds coul 0 1 1 when `break no`. There is no concern regarding neighbor list builds happening between bond breakage when there is no bond breakage. Eliminating bonded atoms from neighbor list can be useful for post-processing or to get per BPM pair quantities such as number of contacts.

**Related Issue(s)**

N/A

**Author(s)**

Jibril B. Coulibaly, Sandia National Laboratories (jibril.coulibaly@gmail.com)

**Licensing**

**Backward Compatibility**

Yes

**Implementation Notes**

N/A

**Post Submission Checklist**

N/A

### DIFF
--- a/doc/src/Howto_bpm.rst
+++ b/doc/src/Howto_bpm.rst
@@ -109,7 +109,7 @@ charges in BPM models, setting a nonzero coul weight for 1-2 bonds
 ensures all bonded neighbors are still included in the neighbor list
 in case bonds break between neighbor list builds. If bond breakage is
 disabled during a simulation run by setting the *break* keyword to *no*,
-a zero coul weight for 1-2 bonds can be used to exclude bonded neighbors
+a zero coul weight for 1-2 bonds can be used to exclude bonded atoms
 from the neighbor list builds. This can be useful for post-processing,
 or to determine pair interaction properties between distinct bonded particles.
 
--- a/doc/src/Howto_bpm.rst
+++ b/doc/src/Howto_bpm.rst
@@ -110,8 +110,14 @@ ensures all bonded neighbors are still included in the neighbor list
 in case bonds break between neighbor list builds. If bond breakage is
 disabled during a simulation run by setting the *break* keyword to *no*,
 a zero coul weight for 1-2 bonds can be used to exclude bonded atoms
-from the neighbor list builds. This can be useful for post-processing,
-or to determine pair interaction properties between distinct bonded particles.
+from the neighbor list builds
+
+   .. code-block:: LAMMPS
+
+      special_bonds lj 0 1 1 coul 0 1 1
+
+This can be useful for post-processing, or to determine pair interaction
+properties between distinct bonded particles.
 
 To monitor the fracture of bonds in the system, all BPM bond styles
 have the ability to record instances of bond breakage to output using
--- a/doc/src/Howto_bpm.rst
+++ b/doc/src/Howto_bpm.rst
@@ -107,7 +107,11 @@ bond lists is expensive.  By setting the lj weight for 1-2 bonds to
 zero, this turns off pairwise interactions.  Even though there are no
 charges in BPM models, setting a nonzero coul weight for 1-2 bonds
 ensures all bonded neighbors are still included in the neighbor list
-in case bonds break between neighbor list builds.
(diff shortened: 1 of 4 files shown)


doc/src/Howto_bpm.rst
```diff
@@ -107,7 +107,11 @@ bond lists is expensive.  By setting the lj weight for 1-2 bonds to
 zero, this turns off pairwise interactions.  Even though there are no
 charges in BPM models, setting a nonzero coul weight for 1-2 bonds
 ensures all bonded neighbors are still included in the neighbor list
-in case bonds break between neighbor list builds.
+in case bonds break between neighbor list builds. If bond breakage is
+disabled during a simulation run by setting the *break* keyword to *no*,
+a zero coul weight for 1-2 bonds can be used to exclude bonded atoms
```
sjplimp, line 112: What about the LJ weights ?  What should they be set to ?

jtclemm (approved):
This looks like a valid use for LJ[1] = Coul[1] = 0.0 and the PR looks good to me. In the future, it might be good to add a table listing the possible combinations of options, the required LJ/Coul weights, and what they do. We'll have to first think whether any other combinations might be useful for users.

doc/src/Howto_bpm.rst
jtclemm: From the code, the LJ weights still have to be either 0 1 1 or 1 1 1. Is this right Jibril?

I can't think of a use case for any other option...

akohlmey: Please note that the `special_bonds` command is a bit "special". If you specify only the `coul` keyword this will *reset* the `lj` settings. Thus it is recommended to always specify both or use `lj/coul`  (especially in examples or the documentation) so there is no misunderstanding what the settings should be.

jibril-b-coulibaly: - If `overlay/pair yes` is used, the lj weights **must be** 1 X Y, regardless of `break yes` or `break no`.

The code currently only checks for LJ[1]==1. It does not enforce X=1 or Y=1, nor coul weights to be 1 1 1.
The main downside of not enforcing LJ[2]=LJ[3]=Coul[1]=Coul[2]=Coul[3]=1 would be the cost of building the 1-3 and 1-4 special bonds if a user selects a value <1. We could enforce that, or leave it as is, please let me know your thoughts @jtclemm @sjplimp 

- If `overlay/pair no` is used, the lj weights **must be** 0 1 1, regardless of `break yes` or `break no`

Let me rephrase the doc to explicitly include `special_bonds lj 0 1 1 coul 0 1 1` per @akohlmey's comment about `coul` resetting `lj`.

jtclemm: > We could enforce that, or leave it as is, please let me know your thoughts @jtclemm @sjplimp

Oh right, no limits were placed on the `overlay yes` option. I don't have strong opinions, but I would lean on the side of simply always requiring the LJ[2] = LJ[3] = Coul[2] = Coul[3] = 1.0. The BPM package will already error if used with 3- or 4-body interactions, so this shouldn't add any practical restrictions.

I do have a version of 3-body BPM interactions which I may eventually update and add to the package (and could extend it to 4-body), but I think it's best to worry about those cases later.

sjplimp: I think @jtclemm 's comment is the right approach.  Have the user follow the simple rule now - if it needs to be relaxed for a later extension, do that later.

jibril-b-coulibaly: Sounds good to me. I updated the error check to enforce LJ[1] = LJ[2] = LJ[3] = Coul[1] = Coul[2] = Coul[3] = 1.0 when `overlay/pair yes` is used and updated the documentation accordingly.

jtclemm: Thanks for adding those checks Jibril. I think the PR should be good to go.

sjplimp (approved):

PR #3844, opened 2023-07-06
Collected small fixes

akohlmey:
**Summary**

This pull request collects multiple small changes and bug fixes

**Related Issue(s)**

addresses ellipsoid visualization issue discussed in: https://matsci.org/t/visualise-ellipsoid/49587

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

No known issues

**Implementation Notes**

- return quaternions as (1, 0, 0, 0) and shape as (1, 1, 1) for non-ellipsoid particles in compute property/atom
- update ASPHERE and ellipsoid examples for using compute property/atom
- update dump output in examples for ellipsoids using custom column names so OVITO will autodetect them
- bugfix for per-atom data access from fixes for fix ave/histo and fix ave/histo/weight (from Stan)
- fix bugs where Atom::tag values were stored in an `int` instead of a `tagint`.
- make use of blanks in conditions more consistent and balanced
- update list of known LAMMPS commands for syntax highlighting in the manual
- fix error message in compute property/atom where strings were inserted in the wrong order
- update URL for CasP website
- remove dead code and enforce some initializers

**Post Submission Checklist**

- [x] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [x] The source code follows the LAMMPS formatting guidelines
- [ ] Suitable new documentation files and/or updates to the existing docs are included
- [ ] The added/updated documentation is integrated and tested with the documentation build system
- [x] The feature has been verified to work with the conventional build system
- [x] The feature has been verified to work with the CMake based build system
- [ ] Suitable tests have been added to the unittest tree.

PR #3851, opened 2023-07-11
Documentation: swap "parsed-literal::" for "code-block:: LAMMPS" in Syntax sections

hammondkd:
**Summary**

This is a search-and-replace for

```
Syntax
""""""

.. parsed-literal::
```
that replaces the last line with
```
.. code-block:: LAMMPS
```

This should make the top code block demonstrating syntax look similar on all documentation pages.

Also included are some updates to the LAMMPS syntax lexer class to handle more keywords consistently.

**Author(s)**

Karl Hammond
University of Missouri
@hammondkd

Changes to `LAMMPSLexer.py` by @akohlmey 

**Licensing**

**Implementation Notes**

I don't know why I didn't think of this earlier, but this was a simple VIM-based search-and-replace across all .rst files:
```
for i in *.rst
do
  vim -c '%s/\(Syntax\n""""""\n\n\.\.\) parsed-literal::/\1 code-block:: LAMMPS' \
      -c ':wq' "$i"
done
```

**Post Submission Checklist**


sjplimp (approved):
I approve if Richard agrees ...

akohlmey:
@rbberger I've tried to update the lexer for pygments to recognize a few more keywords and arguments.
One thing is still puzzling, though. If you look at the last example on the documentation for the `jump`  command, there not all keywords are not properly recognized when indented (but some are). I could not figure out how which adjustments are needed.

rbberger:
@akohlmey I can push an update that fixes this, but at the same time causes a new issue. Words such `mass` will then get highlighted in places where they shouldn't. That was the original motivation of limiting it to line starts.

akohlmey:
> @akohlmey I can push an update that fixes this, but at the same time causes a new issue. Words such `mass` will then get highlighted in places where they shouldn't. That was the original motivation of limiting it to line starts.

Can you send me a diff offline and then I can experiment with it. In general, I am leaning toward having some parts not highlighted rather than too many and in the wrong places.

rbberger:
@akohlmey it's easier to just push it here. See b969b51

akohlmey:
> @akohlmey it's easier to just push it here. See [b969b51](https://github.com/lammps/lammps/commit/b969b51d37cd16ba083a0e77b41191a65f01e528)

Thanks. I have reverted it. It does indeed too much. But then I added a second word list that does have the "highlight everywhere" setting that only has `if, then, else, elif`. These seem to be the ones where this setting makes the most sense. We don't have many input examples with indentation anyway, so I think it is acceptable. I suspect, you came to a similar conclusion originally?

rbberger:
Yup.

rbberger:
@akohlmey the alternative is to do what I did, but the implement the full signature of the problematic commands, so that the later arguments are classified as just strings and don't fall back to root and therefore match with keywords.

akohlmey:
> @akohlmey the alternative is to do what I did, but the implement the full signature of the problematic commands, so that the later arguments are classified as just strings and don't fall back to root and therefore match with keywords.

Well, my feeling is that this is good enough for now. I went through the list of the commands, and  updated for recent changes in PR #3844, then checked the output for those and fixed a few issues either in the files or through modifications of the lexer. Together with @hammondkd's changes that started this PR, we have now quite consistent syntax highlighting. The return on invested time is starting to get too low and I am not bored enough to continue anyway ;-)

rbberger (approved):

PR #3858, opened 2023-07-17
Adding tools to identify rattlers and measure nonaffine displacement 

jtclemm:
**Summary**

This PR includes a new compute which allow users to identify under-coordinated particles (rattlers) in their system and a new fix to measure the nonaffine displacement of particles. 

The compute works through a loop: calculating a coordination number (using various criteria), masking under-coordinated particles, and then recalculating a coordination number until it finds a stable set of non-rattler particles. It returns a per-atom array which indicates which particles are rattlers. 

The fix calculates a per-atom array with various metrics of non-affine displacement which is calculated using two options. The first is simply to integrate the velocity of a particle which only works if deformation is applied by remapping particle positions. The second option is to calculate the D2_min of a particle every nevery timesteps. Both methods rely on labeling a reference state which the user can pick. A compute seems like it would be more appropriate, but I wasn't aware of method to regularly call methods of a compute (e.g. to refresh the reference state) when it isn't being directly invoked by other commands.

**Related Issue(s)**

NA

**Author(s)**

Joel Clemmer (SNL)
Ishan Srivastava (LBNL)

**Licensing**

**Backward Compatibility**

NA

**Implementation Notes**

NA

**Post Submission Checklist**

NA


akohlmey:
@jtclemm updated branch to the latest develop and addressed a few cosmetic/spelling issues

sjplimp (changes requested):
Not clear why fix nonaffine is using an instance of fix property/atom to store per-atom atom state?
If I understand the code, I think fix store/state (an internal fix) would be what should be used.
See fix adapt and its post_constructor() for an example of this.  Ditto for compute displace/atom
and its constructor.

There is a compute displace/atom, so I'm wondering if fix displace/nonaffine/atom is a better name for this fix.  I agree that I can't think of an easy way to do this as a compute, b/c of the unusual options you have implemented for updating the reference state.  The doc pages for both commands should reference each other.

doc/src/compute_rattlers.rst
```diff
@@ -0,0 +1,77 @@
+.. index:: compute rattlers
+
+compute rattlers command
```
sjplimp: Computes that produce per-atom data should have a name like
compute rattler/atom, i.e. suffixed by atom

doc/src/compute_rattlers.rst
```diff
@@ -0,0 +1,77 @@
...
+.. code-block:: LAMMPS
+
+   compute 1 all rattlers type 4 10
+
+Description
+"""""""""""
+
+.. versionadded:: TBD
+
+Define a compute that identifies rattlers in a system. Rattlers are
```
sjplimp: I think after the 1st sentence, you should defining what a rattler is and briefly explain scenarios where it is useful to be able to identify them.  Then a 2nd paragraph could explain how this compute operates to identify them.

jtclemm:
> Not clear why fix nonaffine is using an instance of fix property/atom to store per-atom atom state? If I understand the code, I think fix store/state (an internal fix) would be what should be used. See fix adapt and its post_constructor() for an example of this. Ditto for compute displace/atom and its constructor.
> 
> There is a compute displace/atom, so I'm wondering if fix displace/nonaffine/atom is a better name for this fix. I agree that I can't think of an easy way to do this as a compute, b/c of the unusual options you have implemented for updating the reference state. The doc pages for both commands should reference each other.

I believe I used fix property/atom because ghost atoms need to know their reference position to calculate a per-atom strain tensor. I recall fix store/state did not have an option for this. Is this still a shortcoming of fix store/atom, or am I mistaken?

doc/src/compute_rattlers.rst
```diff
@@ -0,0 +1,77 @@
...
+identified using an interactive approach. The coordination number of
+all atoms is first calculated.  The *type* and *radius* settings are
+used to select whether interaction cutoffs are determined by atom
+types or by the sum of atomic radii (atom style sphere), respectively.
+Rattlers are then identified as particles with a coordination number
+less than *zmin* and are removed from consideration. Atomic coordination
+numbers are then recalculated, excluding previously identified rattlers,
+to identify a new set of rattlers. This process is iterated, up to a maximum
+of *ntries*, until no new rattlers are identified and the remaining
```
sjplimp: I think a sentence or two of guidance about setting ntries would be useful.  Possibly it is different for 2d vs 3d systems?  Or for crystalline solids vs glasses vs granular, etc ?

sjplimp:
Possibly it was changed recently, but the top of the fix_store_state.cpp file describes a "gflag" option for ghost atom comm.

akohlmey:
> I believe I used fix property/atom because ghost atoms need to know their reference position to calculate a per-atom strain tensor. I recall fix store/state did not have an option for this. Is this still a shortcoming of fix store/atom, or am I mistaken?

Ghost atom support was added in commit 3815c0ef76b100bdee330a8ed64d2de76e879dae for use with fix alchemy, IIRC

doc/src/compute_rattlers.rst
sjplimp: should there be an "or" before until ?

doc/src/compute_rattlers.rst
jtclemm: Changed. 

Is there a preference for naming fixes in a similar manner? The fix I added, nonaffine displacement, is kind of an odd duck since it basically acts like a compute (doesn't affect dynamics, calculates per-atom quantities) but I had to construct it as a fix since it needs to operate every timestep.

doc/src/compute_rattlers.rst
jtclemm: I added some guidance but cautioned readers that it really depends on the system. E.g. a linear chain of N particles would take N/2 iterations to sequentially pop particles off the ends until it discovers that the entire chain is unstable (for a minimum coordination of 2).

jtclemm: Fixed

akohlmey:
@sjplimp this pull request has been waiting for you to check out the requested changes for quite some time now.

sjplimp (approved):
Thanks @jtclemm for addressing all the issues I flagged.

PR #3855, opened 2023-07-15
Collected small changes and fixes

akohlmey:
**Summary**

This pull request combines multiple small fixes and changes

**Related Issue(s)**

fixes #3854

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

No known issues.

**Implementation Notes**

The following individual changes are included:
- update VIM syntax highlighting for LAMMPS input files
- fix of an off-by-one bug when parsing the optional argument of the temper/npt command
- fixes to address static code analysis issues in selected files
- coding style improvement and code modernization in selected files
- small corrections and style improvements in the manual

**Post Submission Checklist**

- [x] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [x] The source code follows the LAMMPS formatting guidelines
- [ ] Suitable new documentation files and/or updates to the existing docs are included
- [ ] The added/updated documentation is integrated and tested with the documentation build system
- [x] The feature has been verified to work with the conventional build system
- [x] The feature has been verified to work with the CMake based build system
- [ ] Suitable tests have been added to the unittest tree.
- [ ] A package specific README file has been included or updated

PR #3860, opened 2023-07-18
Small refactor to better handle diamflag in fix adapt and fix adapt/fep

akohlmey:
**Summary**

Simplify the handling of the diamflag setting in fix adapt and fix adapt/fep by moving the flag to the base class.
This avoids some code complexity and complications from fix adapt/fep being optional.

**Related Issue(s)**

This fixes a bug introduced in PR #3855 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Post Submission Checklist**

### DIFF
--- a/src/BPM/atom_vec_bpm_sphere.cpp
+++ b/src/BPM/atom_vec_bpm_sphere.cpp
@@ -98,11 +98,12 @@ void AtomVecBPMSphere::init()
 
   // check if optional radvary setting should have been set to 1
 
-  for (auto ifix : modify->get_fix_by_style("^adapt")) {
-    if (ifix->diam_flag && (radvary == 0))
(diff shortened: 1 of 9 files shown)


jtclemm:
This change makes sense to me. Only one thought, would it be a little cleaner to swap the order of if statements and first check if `radvary == 0` and then loop through fixes?  Just to match the logic of the comment.

jtclemm (approved):

akohlmey:
> This change makes sense to me. Only one thought, would it be a little cleaner to swap the order of if statements and first check if `radvary == 0` and then loop through fixes? Just to match the logic of the comment.

done (plus a couple additional minor cleanups)

sjplimp (approved):

PR #3861, opened 2023-07-19

akohlmey:
**Summary**

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- building the LAMMPS shell requires pkg-config. Add check and print a better error message in case pkg-config is missing with LAMMPS shell enabled
- CMake preset and small changes to support building multi-arch executables on macOS
- include bench folder in "make install" target
- add option to select whether to build for multiple GPU architectures (on by default) with CUDA
- must lower epsilon for fix timestep tests when pppm is used with single precision FFTs
- add paragraph explaining that initializers for class members should not be placed in the header but in the constructor.
- add a string buffer for storing error messages when there is no LAMMPS instance, so that `lammps_has_error()` and `lammps_get_last_error_message()` can be calls with a NULL handle and then report errors from creating a LAMMPS instance. Useful when using the LAMMPS library interface from a GUI application.

**Post Submission Checklist**


akohlmey:
@sjplimp the last two commits change the CMake "make install" target so the `bench` folder is included and also a README file added to the top level directory with pointers where to find the executables, potential files and inputs.
This is what can be easily done and without creating too many disruptions.

I've updated the tar file at https://download.lammps.org/static/lammps-macOS-multiarch-develop.tar.gz
Please give it a try.

sjplimp (approved):

PR #3865, opened 2023-07-21
More fine grained monitoring if computes were initialized

akohlmey:
**Summary**

This updates and refactors the changes from PR #3771 to handle cases of computes being created *after* a run.

**Author(s)**

Axel Kohlmeyer and Steve Plimpton, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

This introduces a flag inside the Compute class that tracks whether the compute was initialized by a run or minimize or something else that calls `Modify::init()`. This flag can be queried with a new `Compute::is_initialized()` method.

**Post Submission Checklist**

### DIFF
--- a/src/reset_atoms_image.cpp
+++ b/src/reset_atoms_image.cpp
@@ -61,6 +61,22 @@ void ResetAtomsImage::command(int narg, char **arg)
   MPI_Barrier(world);
   double time1 = platform::walltime();
 
+  // create computes and variables
+  // must come before lmp->init so the computes are properly initialized
+
+  auto frags = modify->add_compute("frags_r_i_f all fragment/atom single yes");
+  auto chunk = modify->add_compute("chunk_r_i_f all chunk/atom c_frags_r_i_f compress yes");
+  auto flags = modify->add_compute("flags_r_i_f all property/atom ix iy iz");
+  input->variable->set("ix_r_i_f atom c_flags_r_i_f[1]");
+  input->variable->set("iy_r_i_f atom c_flags_r_i_f[2]");
+  input->variable->set("iz_r_i_f atom c_flags_r_i_f[3]");
+  auto ifmin = modify->add_compute("ifmin_r_i_f all reduce/chunk chunk_r_i_f min "
+                                   "v_ix_r_i_f v_iy_r_i_f v_iz_r_i_f");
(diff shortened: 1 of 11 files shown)


src/thermo.cpp
```diff
@@ -1134,8 +1134,8 @@ void Thermo::check_temp(const std::string &keyword)
   if (!temperature)
     error->all(FLERR, "Thermo keyword {} in variable requires thermo to use/init temperature",
                keyword);
-  if (update->first_update == 0)
-    error->all(FLERR,"Thermo keyword {} cannot be invoked before first run",keyword);
+  if (!temperature->is_initialized())
+    error->all(FLERR,"Thermo keyword {} cannot be invoked before initialized by run",keyword);
```
sjplimp: "by a run" to be consistent with variable.cpp
ditto for other changes in this file

src/variable.cpp
```diff
@@ -1505,8 +1505,8 @@ double Variable::evaluate(char *str, Tree **tree, int ivar)
 
         if (nbracket == 0 && compute->scalar_flag && lowercase) {
 
-          if (update->first_update == 0)
-            print_var_error(FLERR,"Variable formula compute cannot be invoked before first run",ivar);
+          if (!compute->is_initialized())
```
sjplimp, line 1508: "by a run" - same as dump and thermo error messages
better syntax

src/reset_atoms_image.cpp
```diff
@@ -93,24 +93,23 @@ void ResetAtomsImage::command(int narg, char **arg)
...
-  // need to ensure update->first_update = 1
-  //   to allow this input script command prior to first run/minimize
-  //   this is b/c internal variables are evaulated which invoke computes
-  //   that will trigger an error unless first_update = 1
-  // reset update->first_update when done
-
-  int first_update_saved = update->first_update;
-  update->first_update = 1;
-
+  // need to first initialize compute flags to allow this input script command prior
```
sjplimp: I think here what is needed is to insure the created computes actually invoke their init() method
before their compute_peratom() or array() method.  The simplest way would be to create them
before the lmp->init() call above.  Otherwise, use frags->init() rather than frags->init_flags(),
since the latter doesn't matter and the computes will be deleted below.

akohlmey: That does the trick. Done.

sjplimp (approved):

PR #3867, opened 2023-07-22
Add  nmax keyword to fix vector to restrict vector length

akohlmey:
**Summary**

This PR adds an optional keyword `nmax` to the fix vector command that will limit the number of entries
in the vector.  The vector will then become a sliding window, where only the last N entries are kept.

**Related Issue(s)**

https://matsci.org/t/terminate-lammps-simulation-once-potential-energy-fluctuation-is-small/49763/9

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

Yes.

**Implementation Notes**

It may be useful to promote the variable `ncount` (and more, if necessary to avoid truncation) to a bigint. Since with a sliding window (when using the `nmax` keyword) one could keep running over many run commands with many steps. An int will eventually overflow. With the full vector length, this would not matter, since the interface to the data only allows signed 32-bit integer indexing, so there is an intrinsic limit, but with nmax this is gone.

**Post Submission Checklist**


doc/src/fix_vector.rst
```diff
@@ -43,6 +45,10 @@ values, they are stored as rows in a global array, whose number of
 rows is growing.  The resulting vector or array can be used by other
 :doc:`output commands <Howto_output>`.
 
+The optional *nmax* keyword can be used to restrict the length of the
+vector to the given *length* value. Once the vector is filled, the
+oldest entries will be discarded when new entries are added.
```
jtclemm: Should it be mentioned either here or at the bottom of the page that the default behavior is *nmax* = the max integer  value so users won't expect it to store data indefinitely if *nmax* isn't specified?

Maybe you already described this somewhere in the doc page and I missed it.

jtclemm (approved):

stanmoore1 (approved):

akohlmey:
@sjplimp this should be (mostly) in line with your suggestions (and my preferences).

PR #3879, opened 2023-08-04
CMake build option for installing KOKKOS headers

ndtrung81:
**Summary**

This PR added an optional cmake variable (`Kokkos_INSTALL_HEADER`) that allows installing KOKKOS header and lib (off by default). By default, the change preserves the behavior made by commit  5878040f26b273c7ff9f56fe6c70c6d194b5d80c "skip undesired 'make install' targets from bundled Kokkos library". 

This feature, when turned on, enables users to build LAMMPS as a shared lib with KOKKOS support and then have direct access to LAMMPS KOKKOS API. An example use cases is the `lammps dlext` package (https://github.com/SSAGESLabs/lammps-dlext)

**Related Issue(s)**

N/A

**Author(s)**

Trung Nguyen (U Chicago)

**Licensing**

**Backward Compatibility**

Should maintain backward compatibility.

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/cmake/Modules/Packages/KOKKOS.cmake
+++ b/cmake/Modules/Packages/KOKKOS.cmake
@@ -26,6 +26,8 @@ endif()
 ########################################################################
 
 option(EXTERNAL_KOKKOS "Build against external kokkos library" OFF)
+option(Kokkos_INSTALL_HEADER "Intall kokkos header files" OFF)
+mark_as_advanced(Kokkos_INSTALL_HEADER)
 option(DOWNLOAD_KOKKOS "Download the KOKKOS library instead of using the bundled one" OFF)
 if(DOWNLOAD_KOKKOS)
   # extract Kokkos-related variables and values so we can forward them to the Kokkos library build
@@ -89,7 +91,11 @@ else()
   if(CMAKE_REQUEST_PIC)
     set(CMAKE_POSITION_INDEPENDENT_CODE ON)
   endif()
-  add_subdirectory(${LAMMPS_LIB_KOKKOS_SRC_DIR} ${LAMMPS_LIB_KOKKOS_BIN_DIR} EXCLUDE_FROM_ALL)
+  if(Kokkos_INSTALL_HEADER)
+    add_subdirectory(${LAMMPS_LIB_KOKKOS_SRC_DIR} ${LAMMPS_LIB_KOKKOS_BIN_DIR})
+  else()
+    add_subdirectory(${LAMMPS_LIB_KOKKOS_SRC_DIR} ${LAMMPS_LIB_KOKKOS_BIN_DIR} EXCLUDE_FROM_ALL)
+  endif()
 
   set(Kokkos_INCLUDE_DIRS ${LAMMPS_LIB_KOKKOS_SRC_DIR}/core/src
                           ${LAMMPS_LIB_KOKKOS_SRC_DIR}/containers/src


stanmoore1 (approved):

junghans:
Can somebody remind me why we need a shared kokkos library? Wouldn't it be enough to link the static kokkos library into the shared lammps library?

akohlmey:
@junghans I discussed this with @ndtrung81 privately. The motivation is rather unusual. The proper approach for the use case would be to compile and install Kokkos separately and then tell LAMMPS to use that external Kokkos library instead of the bundled one. The purpose of this PR to simplify the process (and avoid complications from compiling Kokkos in standalone form), by taking advantage of the fact that liblammps.so *contains* the kokkos libraries. Thus you can write new code that requires the Kokkos headers *and* the LAMMPS headers (from "make install") and then simply link to liblammps.so.

The static Kokkos libraries that would also be installed by the modification here are going to be ignored and them being copied was the main motivation for adding the "EXCLUDE_FROM_ALL" in the first place.

We do a few things in LAMMPS for the sake of simplicity (in maintenance) that are not "clean" from a more distant point of view that keeps in mind that the audience are not experienced programmers or package maintainers, but this is pushing it just one step too far, in my opinion.

ndtrung81:
The modification (more precisely, addition) made in the PR are proposed intentionally to keep the current setting intact (i.e. with `EXCLUDE_FROM_ALL` applied to the KOKKOS header files and generated static libs.), as simple as
```
option(Kokkos_INSTALL_HEADER "Intall kokkos header files" OFF)
mark_as_advanced(Kokkos_INSTALL_HEADER)
...
  if(Kokkos_INSTALL_HEADER)
    add_subdirectory(${LAMMPS_LIB_KOKKOS_SRC_DIR} ${LAMMPS_LIB_KOKKOS_BIN_DIR})
  else()
    add_subdirectory(${LAMMPS_LIB_KOKKOS_SRC_DIR} ${LAMMPS_LIB_KOKKOS_BIN_DIR} EXCLUDE_FROM_ALL)
  endif()
```
so that the audience don't need to make any changes for their regular builds.

One of the benefits with the KOKKOS header files installed along with LAMMPS built with the KOKKOS package is that developers can configure their builds with new KOKKOS-derived classes out of the LAMMPS KOKKOS source tree with a single cmake command.

The bundled KOKKOS static libs are not to be linked with, hence a untidy approach, and as @akohlmey has also warned me, they may cause subtle bugs if one unintentionally tries to do so (duplicated symbols during linking is the most trivial ones). But these cons aside, it would offer a convenient method for those who would like to build upon LAMMPS with KOKKOS support, for example, to have direct access device-side arrays from C++/Python levels,.

In the proper/cleaner approach mentioned, you need to require another step with building KOKKOS separately, before building LAMMPS. In my opinion, building KOKKOS separately already presumes experienced developers :-)

As I don't have a good sense of the implications from the maintenance point of view with this proposed added option, I am open to any suggestion, like only installing the KOKKOS header files, excluding the static libs when we do  `make install`.

cmake/Modules/Packages/KOKKOS.cmake
```diff
@@ -89,7 +91,11 @@ else()
   if(CMAKE_REQUEST_PIC)
     set(CMAKE_POSITION_INDEPENDENT_CODE ON)
   endif()
-  add_subdirectory(${LAMMPS_LIB_KOKKOS_SRC_DIR} ${LAMMPS_LIB_KOKKOS_BIN_DIR} EXCLUDE_FROM_ALL)
+  if(Kokkos_INSTALL_HEADER)
+    add_subdirectory(${LAMMPS_LIB_KOKKOS_SRC_DIR} ${LAMMPS_LIB_KOKKOS_BIN_DIR})
+  else()
+    add_subdirectory(${LAMMPS_LIB_KOKKOS_SRC_DIR} ${LAMMPS_LIB_KOKKOS_BIN_DIR} EXCLUDE_FROM_ALL)
+  endif()
```
junghans, line 98: I think you should be able to write it all as one line with something like:
```
add_subdirectory(${LAMMPS_LIB_KOKKOS_SRC_DIR} ${LAMMPS_LIB_KOKKOS_BIN_DIR} $<$<NOT:$<BOOL:${Kokkos_INSTALL_HEADER}>>:EXCLUDE_FROM_ALL>)
```

junghans:
I understand both sides, but I honestly think if a user needs a custom-built `libkokkos`, they should be able to build it outside of lammps (e.g. using spack). On the other hand the change is easy enough to not worry about it.

As far as I recall `make install` was never an _official_ feature, but merely a hack that we kept around to make the work of the packagers easier, so I am not even sure "normal" users should use it.

cmake/Modules/Packages/KOKKOS.cmake
junghans: Sorry, `add_subdirectory()` do not support generator expressions, so nothing to be done here.

akohlmey:
It is not the size of the change, it is the principle that worries me. We have the rule or "no feature without documentation", but I don't think we should advertise the use case that is enabled here. It just goes against my grain to have to explain that you need to link with LAMMPS to get Kokkos. If Kokkos matures to the point that there is little benefit in bundling it (I don't expect that to happen soon, but who knows?) then we would be forced to keep it because of this hack. If people absolutely want to do this, they may just as well edit the CMakeLists.txt file to remove the EXCLUDE_FROM_ALL. That can be automated with, e.g., sed, too. So this brings this down to the difference between adding one more option to the cmake command line vs. running a small sed command beforehand and I suspect both can be automated with a script wrapper.

ndtrung81:
Since this is the final verdict, I will close the PR.

PR #3889, opened 2023-08-17
Implement anti-aliasing for dump image (and dump movie)

akohlmey:
**Summary**

This implements a simple full scene anti-aliasing (FSAA) scheme in dump image for smoother edges and higher quality pictures.

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

With this simple FSAA scheme, the width and height of the image are doubled internally and then 2x2 blocks of pixels averaged to produce a single pixel of the picture at the original size. This produces visibly smoother edges.

**Post Submission Checklist**

Image created with default setting, i.e. `fsaa off`.
image 300

Image created with anti-aliasing, i.e. `fsaa on`.
image-aa 300


akohlmey:
@sjplimp we could do more complex scaling algorithms or use a 3x3 or 4x4 sample for more improvements, but those are rather small gains compared to the added computational cost. So I think this is the best compromise and also requires rather few changes. This scheme also has the advantage that the scaling down can be done in-place.

doc/src/dump_image.rst
```diff
@@ -957,6 +959,17 @@ frequently.
 
 ----------
 
+.. versionadded:: TBD
+
+The *fsaa* keyword can be used with the dump image command to improve
```
sjplimp: This paragraph needs a line like the following (see SSAO paragraph for similar):

This option will increase the cost of computing the image by 4x, or somewhat more for large images
on large processor counts.

sjplimp (changes requested):
see doc page line

doc/src/dump_image.rst
akohlmey: done

akohlmey:
> see doc page line

Please note that also from the doc page but also from the technical point of view it would be more consistent and less programming effort to have the `fsaa` keyword alongside `ssao` with the main dump command line instead of in dump_modify(). The two options are somewhat related since they both aim to improve image quality.

doc/src/dump_image.rst
```diff
@@ -957,6 +959,20 @@ frequently.
...
+the image quality by enabling full scene anti-aliasing.  Internally the
+image is rendered at twice the width and height and then scaled down by
+computing the average of each 2x2 block of pixels to produce a single
+pixel in the final image at the original size. This produces images with
+smoother, less ragged edges.  The calculation of this effect can
+increase the cost of computing the image by roughly 4x or more,
+especially for large images and with large processor counts due to
```
sjplimp: "increased communication cost" is not actually the issue here - it is that the last stage (shrink image with aliasing) is done serially on a single proc - so no parallelism and no communication.

So I think my sentence is better, but add a phrase to clarify:

This option will increase the cost of computing the image by 4x, or somewhat more for large images
on large processor counts due to the final anti-aliasing operation being performed on a single processor.

sjplimp (changes requested):
if you mean you want to make "fsaa" an optional keyword in the dump image command like "ssao" that is fine by me - I guess it would need no args?  Also, see my additonal comment on the 4x factor doc page sentence.

doc/src/dump_image.rst
akohlmey: I think you are overestimating the cost of the downscaling. The way it is done is very cache friendly and the really slow step is the part were the data is actually written to a file (and for JPEG and PNG there is also a compression step, but I don't recall any discussion of that). The communication cost is something that is significant because the Image::merge() function is collecting the image data into one buffer on rank 0. With a 4x larger image, that is more expensive especially with many MPI ranks. A more sophisticated implementation would do the downscaling **before** the "merge" step. But I have not looked into it. Also the downscaling could be easily made thread-parallel, but that would require creating a new buffer and it would be more memory bandwidth limited rather than CPU time limited, since the cost of computing the average of 4 numbers is negligible (the bicubic algorithm would be *much* more expensive since up to 16 pixels will have to be looked at and cubic polynomials evaluated.

sjplimp: I'm thinking about it differently.  The cost of creating an image in parallel on P procs (separate from writing it out)
has calculation and communication costs.  To create an image 4x larger on the same P procs costs 4x more b/c the calculation and communication costs both go up by 4x.  4x more pixels to calculate and 4x more to communicate in each message.  The more-than-4x cost for ffsa is extra serial computation at the end.  But maybe you are right, that extra operation will never be that significant.

That said, I'm fine with the doc page saying:  "This option will increase the cost of computing the image by 4x or more."
I don't think SSAO is precisely 2x more expensive.

sjplimp: I forgot that after each comm stage, the receiving procs need to merge 2 full-size images.  Which requires them to non-scalably loop over all the pixels.  So that is extra computation cost which happens log2(P) times when the image size is 4x larger.  It is also likely similar cost to the one-time anti-alias loop over all pixels.  So I think saying "4x or more" is sufficient info for the user.

akohlmey:
Let's have a look at some real numbers. Thanks to your implementation of "timer" style variables and the "write_dump" command, it is not so hard to do. I am taking the in.rhodo.scaled input with a 2x2x2 replicate for a total of 256000 atoms.
Then I write the image to a ramdisk to minimize that cost and write out a 1024x1024 size image (instead of the default of 512x512) to have some more time spent on this. with serial execution I get (in seconds):

```
Regular: 0.365  
JPEG: 0.3654  Ratio: 1.001
PNG: 0.6264   Ratio: 1.716
FSAA: 0.9748  Ratio: 2.671
SSAO: 1.924   Ratio: 5.271
SSAO + FSAA: 10.53  Ratio: 28.849
```

With the same input but running of 4 MPI ranks I get:

```
Regular: 0.1233            Peff: 0.74
JPEG: 0.124  Ratio: 1.006  Peff: 0.737
PNG: 0.3959  Ratio: 3.211  Peff: 0.396
FSAA: 0.36   Ratio: 2.920  Peff: 0.677
SSAO: 0.8563 Ratio: 6.945  Peff: 0.562
SSAO + FSAA: 5.007  Ratio: 40.608  Peff: 0.526
```

Since this is on a local multi-core machine, the communication cost is minimal, but some trends are visible.
There is some overhead from communication and writing out the file that applies to all cases. When using JPEG you can effectively compensate the cost of compression by the fact that the amount of data shrinks and thus file writing is faster (and by writing to a ramdisk, writing is as fast as it gets). With PNG the cost of compression is significant, about 50% of the serial time and thus writing to PNG in parallel has a lot of serial overhead added and the parallel efficiency is the worst. For FSAA there is a non-negligible overhead from scaling down in serial, but overall the cost is *less* than 4x more like about 3x (but with more processors the efficiency obviously will go down until the downscaling would be done in parallel before the merge step).
The cost of SSAO is far more than 2x (more like ~6x) and that includes some serial overhead.
The combination of SSAO and FSAA is a particular performance killer.

sjplimp:
Interesting - but I'm not fully clear on what these numbers are comparing.
Are you saying that on 1 or more procsesors, the cost of creating (but not writing to disk) an image (JPG or PNG) which is 4x larger than another one is less than 4x more?  This is w/out FSAA or SSAO.  I'm not sure how that could be, maybe cache effects?  Or possibly if the image is mostly blank that would influence the ratio.

akohlmey:
> Interesting - but I'm not fully clear on what these numbers are comparing. Are you saying that on 1 or more procsesors, the cost of creating (but not writing to disk) an image (JPG or PNG) which is 4x larger than another one is less than 4x more?

Yes.

> This is w/out FSAA or SSAO. I'm not sure how that could be, maybe cache effects? Or possibly if the image is mostly blank that would influence the ratio.

The image is full of objects (rhodo system, dump group all). *But* the number of objects that need to be realized is the *same* for the different image *sizes* and that has a significant impact. If you create the same size image, but use a different group, e.g. only the Rhodopsins, i.e. about 52000 atoms out of the 256000 (about 1/5th) , then the cost of creating the same size image drops to about 1/4th.

Just take the `in.rhodo.scaled` input and append the following lines:
```
group rhodopsin type 1:3 5:32 35:36

variable start timer
write_dump  all image /tmp/image.ppm type type view 60 -30 bond atom 0.5 size 1024 1024
variable stop timer
variable regular equal $(v_stop-v_start:%.4g)
print "Regular: ${regular}  "

variable start timer
write_dump  all image /tmp/image.ppm type type view 60 -30 bond atom 0.5 size 2048 2048
variable stop timer
variable fourtimes equal $(v_stop-v_start:%.4g)
print "2x2 Regular: ${fourtimes}  Ratio: $(v_fourtimes/v_regular) "

variable start timer
write_dump  rhodopsin image /tmp/image.ppm type type view 60 -30 bond atom 0.5 size 1024 1024
variable stop timer
variable rhodopsin equal $(v_stop-v_start:%.4g)
print "Rhodopsin: ${rhodopsin}  Ratio: $(v_rhodopsin/v_regular) "

variable start timer
write_dump  all image /tmp/image.jpg type type view 60 -30 bond atom 0.5 size 1024 1024
variable stop timer
variable jpeg equal $(v_stop-v_start:%.4g)
print "JPEG: $(v_jpeg:%.4g)  Ratio: $(v_jpeg/v_regular:%.3f)"

variable start timer
write_dump  all image /tmp/image.png type type view 60 -30 bond atom 0.5 size 1024 1024
variable stop timer
variable png equal $(v_stop-v_start:%.4g)
print "PNG: $(v_png:%.4g)   Ratio: $(v_png/v_regular:%.3f)"

variable start timer
write_dump  all image /tmp/image.ppm type type view 60 -30 bond atom 0.5 size 1024 1024 modify fsaa yes
variable stop timer
variable fsaa equal $(v_stop-v_start:%.4g)
print "FSAA: $(v_fsaa:%.4g)  Ratio: $(v_fsaa/v_regular:%.3f)"

variable start timer
write_dump  all image /tmp/image.ppm type type ssao yes 4539 0.6 view 60 -30 bond atom 0.5 size 1024 1024 modify fsaa no
variable stop timer
variable ssao equal $(v_stop-v_start:%.4g)
print "SSAO: $(v_ssao:%.4g)  Ratio: $(v_ssao/v_regular:%.3f)"

variable start timer
write_dump  all image /tmp/image.ppm type type ssao yes 4539 0.6 view 60 -30 bond atom 0.5 size 1024 1024 modify fsaa yes
variable stop timer
variable both equal $(v_stop-v_start:%.4g)
print "SSAO + FSAA: $(v_both:%.4g)  Ratio: $(v_both/v_regular:%.3f)"
```

akohlmey:
BTW: I just experimented with changing the compression level for PNG images and will change it to the setting with fastest compression (instead of best). The compression rate appears to be almost the same, but the overhead is substantially less, so a significant gain with parallel runs and large images at very little change in storage space use.
People concerned about storage should use JPEG (size drops to half of PNG, but compression is lossy).

sjplimp:
>> the number of objects that need to be realized is the same for the different image sizes and that has a significant impact.

Still doesn't make sense to me.  A proc renders the same # of atoms, independent of the image size.  In a 4x larger image
the circle for each atom is twice the diameter = 4x as many pixels.  So the number of pixels each proc calculates goes up by 4x.

But regardless of understanding this fully, I'm fine with just putting approximate cost factors you think are valid for SSAO and FSAA in the doc page and merging the PR.  Having AA available for users is a nice option.   Thanks for working on this.

akohlmey:
> Still doesn't make sense to me. A proc renders the same # of atoms, independent of the image size. In a 4x larger image the circle for each atom is twice the diameter = 4x as many pixels. So the number of pixels each proc calculates goes up by 4x.

The cost for creating the pixels goes up, but the cost for computations that are per-atom or per-bond (like location, orientation, etc.) in the image coordinate system remains constant. If you look at the `Image::draw_XXX()` functions, there are a whole bunch of vector operations *outside* the loops that place the pixels. From the numbers it looks the ratio of time spent outside and inside the loops is about 1:3 for the given example.

sjplimp:
I see - makes more sense now - thanks for explaining

sjplimp (approved):

akohlmey:
@sjplimp do think it would be helpful if I create 4 small images (by cutting out the same area from larger ones and zooming in for better effect, e.g. of the peptide example showing just part of the peptide) with the different quality settings: default, fsaa, ssao, ssao+fsaa and add them with a description as an "Image quality" paragraph to the dump image documentation?

sjplimp:
yes, I do think that's a good idea

akohlmey:
> yes, I do think that's a good idea

done.

PR #3892, opened 2023-08-20

### LINKED ISSUE #3901: The build with MESONT package enabled fails
Hi, I'm preparing the LAMMPS slackbuild package for slackbuilds.org. The script builds the installation package from the source. It works with several packages but the compilation fails if the MESONT package is enabled

LAMMPS version: lammps-stable-2Aug2023
platform: Slackware Current

Details
Adding to cmake the flag -D PKG_MESONT=yes

the output stream returns this

```
[ 59%] Building CXX object CMakeFiles/lammps.dir/tmp/SBo/lammps-stable_2Aug2023/src/lattice.cpp.o
In file included from /tmp/SBo/lammps-stable_2Aug2023/cmake/build/styles/style_bond.h:3,
                 from /tmp/SBo/lammps-stable_2Aug2023/src/force.cpp:17:
/tmp/SBo/lammps-stable_2Aug2023/src/MESONT/bond_mesocnt.h:28:10: fatal error: bond_harmonic.h: File o directory non esistente
   28 | #include "bond_harmonic.h"
      |          ^~~~~~~~~~~~~~~~~
compilation terminated.
[ 60%] Building CXX object CMakeFiles/lammps.dir/tmp/SBo/lammps-stable_2Aug2023/src/math_eigen.cpp.o
[ 60%] Building CXX object CMakeFiles/lammps.dir/tmp/SBo/lammps-stable_2Aug2023/src/math_extra.cpp.o
[ 60%] Building CXX object CMakeFiles/lammps.dir/tmp/SBo/lammps-stable_2Aug2023/src/math_special.cpp.o
make[2]: *** [CMakeFiles/lammps.dir/build.make:2932: CMakeFiles/lammps.dir/tmp/SBo/lammps-stable_2Aug2023/src/force.cpp.o] Error 1
make[2]: *** Attesa per i processi non terminati....
In file included from /tmp/SBo/lammps-stable_2Aug2023/cmake/build/styles/style_bond.h:3,
                 from /tmp/SBo/lammps-stable_2Aug2023/src/lammps.cpp:19:
/tmp/SBo/lammps-stable_2Aug2023/src/MESONT/bond_mesocnt.h:28:10: fatal error: bond_harmonic.h: File o directory non esistente
   28 | #include "bond_harmonic.h"
      |          ^~~~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [CMakeFiles/lammps.dir/build.make:3254: CMakeFiles/lammps.dir/tmp/SBo/lammps-stable_2Aug2023/src/lammps.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:172: CMakeFiles/lammps.dir/all] Error 2
```
The header should be included as it happens for other packages (for example INTEL), the code appears similar. What would be wrong? I think this is not a bug but could depend on something in my system, but I have no ideas
Thanks for your reply

PS: the build of version patch-24Mar2022 with MESONT enabled works fine, but several things has changed since this version

### LINKED ISSUE #3902: [BUG] numpy.extract_atom() ignores ghost atoms
**Summary**

It looks to me that the Python wrapper `lammps.numpy.extract_atom()` does not extract the ghost atom information. See [line 91 in `lammps/python/lammps/numpy_wrapper.py`](https://github.com/lammps/lammps/blob/c072719a44ee03df80b1c23ed0d40c34d3c36b22/python/lammps/numpy_wrapper.py#L91).

Is this intentional (perhaps to avoid memory issues)?

`develop` branch

**Expected Behavior**

If my understanding is correct, this line should be:
```python
nelem = self.lmp.extract_global("nmax")
```

in order to ensure that the full array of length `nmax` is parsed, as is indicated should happen in the docstring of the function.

However, given the possibly large size of this array, it would seem sufficient to at least do:
```python
nelem = self.lmp.extract_global("nlocal") + self.lmp.extract_global("nghost")
```

**Actual Behavior**

The line currently is:
```python
nelem = self.lmp.extract_global("nlocal")
```

which means that the returned array is never larger than `nlocal`.

**Steps to Reproduce**

Example script (run with >1 procs):
```python
from lammps import lammps
from mpi4py import MPI

rank = MPI.COMM_WORLD.Get_rank()

def main():
    
    lmp = lammps()

    # build the cell
    lmp.command('units           lj')
    lmp.command('atom_style      atomic')
    lmp.command('lattice         fcc 0.8442')
    lmp.command('region          box block 0 20 0 20 0 20')
    lmp.command('create_box      1 box')
    lmp.command('create_atoms    1 box')

    # make sure there are ghost atoms
    lmp.command('comm_modify cutoff 5.0')

    # for neighbor list rebuild
    lmp.command('mass 1 1.0')
    lmp.command('pair_style zero 1.0 full')
    lmp.command('pair_coeff * *')
    lmp.command('run 0 post no')

    # check expected vs. actual array size
    nlocal = lmp.extract_global('nlocal')
    nghost = lmp.extract_global('nghost')

    xyz = lmp.numpy.extract_atom('x')

    print(f'Rank {rank}: expected = at least {nlocal+nghost}; actual = {xyz.shape[0]} rows')


if __name__ == '__main__':
    main()
```

The output of the above script (with 2 procs) will be:
```bash
Rank 0: expected = at least 40315; actual = 16000 rows
Rank 1: expected = at least 40315; actual = 16000 rows
```

None

akohlmey:
**Summary**

This pull request combines multiple small changes and fixes that do not warrant a pull request on their own.

**Related Issue(s)**

Fixes #3901 
Fixes #3902 
Addresses the issue reported here: https://matsci.org/t/pair-style-hybrid-overlay-hdnnp/50362

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

- fix element mapping bug in pair style hdnnp when used as a hybrid substyle
- silence some bogus warnings with MSVC++ compilers
- explicitly signal that destructors derived from `Pointers` may throw exceptions and thus make them catchable, e.g. from the library interface, instead of calling `std::terminate()`
- parallelize SSAO calculation in dump image with with OpenMP
- make inline functions and constants in EFF package inline header static so they won't pollute the global namespace and create linker problems when being used multiple times
- initialize all class members in `compute composition/atom`
- the MESONT package depends on the MOLECULE package. reflect that in the build system.
- clarify documentation about dimensions of arrays returned by NumPy wrapper of extract_atom()
- re-implement lammps_commands_string() library function using C++ idioms and add support for """ ... """ heredoc strings

**Post Submission Checklist**


sjplimp (approved):
thanks Axel for the changes to compute composition/atom

src/lmppython.h
```diff
@@ -20,7 +20,7 @@ namespace LAMMPS_NS {
 
 class PythonInterface {
  public:
-  virtual ~PythonInterface() = default;
+  virtual ~PythonInterface() noexcept(false) {}
```
rbberger, line 23: Why can these throw?

akohlmey: It cannot, but if you have multiple inheritance, all base classes have to have the same setting. `PythonImpl` inherits from `Pointers` and `PythonInterface` and thus both need to have the same noexcept setting. At least, this is what I figured out about it, after I changed Pointers and then got multiple compilation errors and googled around.

rbberger: I only commented on one of these, but the change in Pointers better have a **very** good justification. Otherwise you're just inviting resource leaks and forced termination.

PR #3899, opened 2023-08-23
Implement reversed harmonic walls

Eddy-Barraud:
**Summary**

< New feature >

This code is doing the opposite of the current `fix wall/harmonic`, as it is recalling atoms if they go over the walls instead of repelling them if they go too close of the walls.

The need of this fix comes from the calculation of osmotic pressures using virtual walls, as described by (Luo and Roux, [2010](https://doi.org/10.1021/jz900079w)). I could not reach satisfactory results using any of the currently available wall fix. 

In fact, I tried to reproduce the osmotic pressure of NaCl solutions modeled with DPD using electrostatics and PPPM solver. The walls were either too soft (harmonics), and leaded to simulation crashes when an atom crossed a wall, or it was too hard, leading to PPPM solver crashing because of atoms moving too fast.
 
Using this reversed harmonic wall fix, I successfully created a control volume between the walls, permeable only to water, reproducing correctly the osmotic pressure values at different concentrations of NaCl.

I also added some comments to the code.

This code was used to produce results that will be published in a scientific paper, that I will give the reference afterwards.

**Related Issue(s)**

None to my knowledge

**Author(s)**

Eddy BARRAUD, PhD. student, eddy.barraud@outlook.fr

**Licensing**

**Backward Compatibility**

It is not touching any existent fix.

**Implementation Notes**

The new fix name I suggested is: **wall/harmonic/reversed**

The usage of the wall is the same as fix wall/harmonic, here is an example creating a control volume between the z coordinates 5.0 and 15.0, with a cutoff of 2.0, and a force constant of 10.0.

`fix         walls ionsGrp wall/harmonic/returned zlo 5.0 10.0 0.0 2.0 zhi 15.0 10.0 0.0 2.0 pbc yes`

**Post Submission Checklist**


akohlmey:
@Eddy-Barraud thanks for your submission. Can you please attach an input deck for a small test case that demonstrates what this wall fix can do where the regular fix wall/harmonic fails?

Eddy-Barraud:
> @Eddy-Barraud thanks for your submission. Can you please attach an input deck for a small test case that demonstrates what this wall fix can do where the regular fix wall/harmonic fails?

Hello @akohlmey, so the first use case that can handle this new fix and not the regular is to equilibrate a solution of NaCl that was created by placing atoms randomly. Below is an example where I create a box of NaCl 3M in water, then the walls will force the ions to go inside the control volume that is between z coordinates 5 and 15, in a box of dimensions 10x10x20.

[make-fail-1-random-positions.tar.gz](https://github.com/lammps/lammps/files/12420336/make-fail-1-random-positions.tar.gz)

The other use case is with little values of the force constant, where it's not enough for the regular fix to repeal the atoms and leading to a crash. Example 2 below is the same aqueous solution, but the ions were placed randomly inside the control volume region. 

[make-fail-2-low-repulsion.tar.gz](https://github.com/lammps/lammps/files/12420450/make-fail-2-low-repulsion.tar.gz)

The third case is about the PPPM error, it was caused only for big systems and long runs, where I needed to use a quite high repulsion to prevent the error of atoms inside/over the walls. I can't reproduce easily this problem with a little system, and it is solved by the usage of a reasonably small force constant with the new fix.

evoyiatzis:
Hello @Eddy-Barraud, I have also performed similar osmotic pressure calculations using the method of Luo and Roux,  so this PR got my attention. Back in the days, I had focused on atomistic and not on DPD force fields. Out of curiosity, could an initial configuration with confined ions be obtained using an external field via the fix addforce or fix efield? Or even by not placing randomly the ions in the whole simulation box but just in the region of interest (middle section of the simulation box)? 
It is also of interest the combination of the soft DPD pair style (no excluded volume present in the pairwise interactions) with the coul/long style (which leads to infinitely strong attraction of two ions at infinitesimal distances). Some researchers have proposed smearing of partial charges to avoid such a strong ion-pairing (for instance [R. D. Groot J. Chem. Phys. 118, 11265–11277 (2003) https://doi.org/10.1063/1.1574800] and [Minerva González-Melchor J. Chem. Phys. 125, 224107 (2006)  https://doi.org/10.1063/1.2400223]). Could this be one reason for observing PPPM errors?

Eddy-Barraud:
Hello and thank you for your message @evoyiatzis 
I wasn't aware of the fixes addforce/efield but they seem to be appropriated to equilibrate a random solution of ions to a biphasic system.
You are right, we can place the ions in the confined region, it's what I do in the example [make-fail-2-low-repulsion.tar.gz](https://github.com/lammps/lammps/files/12420450/make-fail-2-low-repulsion.tar.gz). I was just suggesting that it's easier to just generate global random positions and use this fix straight forward with equilibration, then production.
Concerning the charges, you are also right 😉 I am actually using smeared charges like [Minerva González-Melchor J. Chem. Phys. 125, 224107 (2006) https://doi.org/10.1063/1.2400223]) for my paper, it's really better representing ions in solutions for DPD simulations, and it prevents ion pairing and infinite electrostatic potential. So, the reason of using coul/long by the time of my testing cases, was just to accelerate calculations using the GPU, because the force style coul/slater/long is not yet compatible with GPU. 😅 
I investigated a bit more and the PPPM error seems to be the result of the neighbor list update frequency `neigh_modify every 1 delay 5 check yes`. What I suppose is that the particle was moving too fast, due to the wall repulsion, and the processors coudn't have the neighbor list at the next time step. In DPD and using this fix I am now always using `neigh_modify every 2 delay 0`.

evoyiatzis:
@Eddy-Barraud I don't think that GPU support for coul/slater/long is high in the developers priority list (definitely not in my list). Maybe you can submit a feature request providing some arguments how lammps users performing DPD simulations with charged beads will benefit and someone picks it up.

stanmoore1:
@sjplimp @akohlmey I think this should probably go into the `src/EXTRA-FIX` folder, do you agree?

@Eddy-Barraud is this PR ready for review?

Eddy-Barraud:
> @sjplimp @akohlmey I think this should probably go into the `src/EXTRA-FIX` folder, do you agree?
> 
> @Eddy-Barraud is this PR ready for review?

Hello @stanmoore1 

We expect to submit a paper during the next two weeks which will mention this modified fix by its name wall/harmonic/returned, so I just need to be sure it keeps its name.

Do you think we could wait a DOI to mention it in the documentation of LAMMPS ?

In the meantime, you could probably proceed to the review.

Thank you

stanmoore1:
>Do you think we could wait a DOI to mention it in the documentation of LAMMPS ?

@Eddy-Barraud yes please let us know when the DOI is available.

Eddy-Barraud:
Hello @stanmoore1, 
Let me finally dig up this pull request with our DOI 😁
doi:10.1021/acs.langmuir.4c04293
Barraud, Dalmazzone, Moureta, De Bruin, Creton, Pasquier, Lachet and Nieto-Draghi, _Langmuir_, 41 (11), 7272-7282 (2025) 

The work in that paper was actively using the reversed harmonic wall. 
I also took the opportunity to mention the GPU pair style dpd/coul/slater/long/gpu, but I will better thank @ndtrung81 in another paper on polymers.

stanmoore1:
@Eddy-Barraud is this ready to merge?

Eddy-Barraud:
@stanmoore1 Hello, yes it's ready now, i've put a short citation inside the doc but I didn't check the rendered HTML.

sjplimp:
I will try to do my review soon.

sjplimp (changes requested):
Please see my recent comment.

sjplimp:
Thanks @Eddy-Barraud for your further work on this and congrats on getting your paper published.
Just a few conceptual issues to address, then this new feature can be merged.

(1) You can add a "citation" to your paper in the source code so that when the feature is used, the citation will be printed out and the user prompted to cite it.

(2) The doc page calls this harmonic/reversed and the file name is harmonic/returned.  They need to agree.  I actually don't like either name.  Maybe harmonic/exterior or harmonic/outside is better?  Meaning the harmonic influence acts on particles outside the box or wall, but not inside ? 

(3) The fix_wall.rst doc page has a NOTE explaining all particles must be "inside" the wall (r > 0) for all the variants, else an error is generated.  Did you disable that error check for this new variant?  Or possibly just don't check?  That is all fine, but the doc page needs to clarify that this variant is different than all the others.

(4) If you have time, it would be nice to add 2 simple (2d) example scripts to examples/wall for harmonic and harmonic/outside.  Which if visualized with a dump image command would show the difference.  And also how to use harmonic/outside with appropriate boundary conditions if particles are allowed to go outside the simulation box.

Eddy-Barraud:
Hello @sjplimp ,
Thanks for taking the time to check this out. 

1) Thanks for the tip, I registered a citeme const in the constructor, it's now showing credits.

2) I am sorry that the name doesn't suit you. I do like the name wall/outside; however, I cited in my paper the exact name of the fix page 7274. I would be a little embarrassed to change the name now and lose people coming from the paper to LAMMPS. I changed the doc to wall/harmonic/returned

3) I removed the check for particles overlapping or being outside the walls, as it's a wanted behavior. Initially, it was a simple check of distances with respect to the walls and a boolean (onflag) triggering the error. I updated the note consequently: "For all of the styles (except wall/harmonic/returned), you must ensure that r is always > 0 for..."

4) I created two small examples so that we can see the difference between the two methods.

wall/harmonic is creating a slight band of "void" reducing the effective controlled volume:
image wall harmonic 2d 15000


Meanwhile, wall/harmonic/returned is allowing particles to slightly overpass the controlled volume, but in a reduced proportion:
image wall harmonic returned 2d 20000

The photos were taken as the last image from dump image of example scripts.

akohlmey:
> 2\. I would be a little embarrassed to change the name now and lose people coming from the paper to LAMMPS.

Please see the discussion and changes in https://github.com/lammps/lammps/pull/4795 for an example of how to change the name of a fix while retaining backward compatibility (for a while).

The final approach is to add the alias to fix deprecated and explain in the error message that change of the fix style name and also document this change in the manual here: https://docs.lammps.org/Commands_removed.html

As you can see from the way that renaming and removing is documented, that we prefer to favor the long-term perspective and thus are usually going for the more consistent naming and then documenting the step, even if it breaks backward compatibility eventually.

There are far more embarrassing things than having a command renamed. :wink:

sjplimp:
I agree with Axel that changing the name of the fix in the code versus the paper should not be an issue.
As developers, our goal is to make it as easy as possible for users to understand what a style command (e.g. fix, compute)
does by its name.  When they look at the options for the fix wall command, I think harmonic/outside is more clear than
harmonic/returned -- since all of the walls "return" an escaping particle to the proper side of the wall.  In the doc page you
can explain the mathematical difference between harmonic and harmonic/outside.

Thanks for the new example scripts.  However the 2 pics are not very distinguishable.  I suggest putting the walls at the
simulation box boundaries, so that for harmonic, all particles are inside the box, and for harmonic/outside, some particles are visibly outside the box (by a small amount).

Eddy-Barraud:
Alright @akohlmey  and @sjplimp 
I changed the name of the fix as requested and added backward compatibility with warning signaling.
The doc is also updated.

For the example, I can't put wall/harmonic/outside at cell boundaries without periodic conditions; otherwise, it fails the simulation. So I don't see the point here; feel free to change it.

Below are two dump images using square cells and greater controlled volumes with the same repulsion force constant.

This is wall/harmonic:
image wall harmonic 2d 20000

This is wall/harmonic/oustide:
image wall harmonic outside 2d 20000

sjplimp (approved):
thanks @Eddy-Barraud for addressing my comments.

Re: images - what you have are fine now that I understand them better - I think the 2 component solvent/solute model is complicating what I was imagining the inside vs outside images could look like, but what you have is a good use case for your new fix wall

sjplimp:
@Eddy-Barraud It's also nice that these images use the new "dump image region" capability which @akohlmey recently added

And thanks again for agreeing to the name change for the fix

Eddy-Barraud:
@sjplimp Thank you for your time on this. It was a pleasure to use freely LAMMPS during my PhD thesis while adding some functionalities. I agree with you on the dump image from @akohlmey. I was curious to use it as soon as it was implemented, and I'm not disappointed!

PR #3912, opened 2023-09-01
Add reformatting of line and command completion to LAMMPS GUI

akohlmey:
**Summary**

This adds the option to reformat the current line when the TAB key is hit. The amount of padding added can be configured in the preferences dialog. In the preferences there is also a checkbox to enable/disable reformatting the line on hitting the return/enter (enabled by default).

Also added is a completion popup which is triggered when Shift+TAB is hit. It is also triggered when 3 or more characters of a word have been typed. This automatic completion popup can be enabled/disabled in the preferences dialog (enabled by default).

Some tweaks for more consistent syntax highlighting are included, too.

Finally, a few bugs from earlier pull requests with enhancements for LAMMPS GUI are fixed.

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

For reformatting, the current line is split into "words" with support single/double quotes (using the same logic that LAMMPS uses) and then reassembled with the configured padding added. Padding is only for the first few words. For the rest a single blank is used.

For command completion a list of commands is taken from an embedded file (for commands built into the Input class) and from querying LAMMPS for included command styles. This will then create a popup when Shift-TAB is hit on the first word in a line in case there are possible completions, or automatically if auto-completion is enabled.

**Post Submission Checklist**


sjplimp:
@akohlmey Made some (mostly cosmetic) edits to the Howto_lammps_gui.rst page

sjplimp (approved):

ndtrung81 (approved):

PR #3917, opened 2023-09-07
Implementation of Slip-Spring Dynamics

GenieTim:
**Summary**

This is a very rough implementation of what could finally be an implementation of slip-spring dynamics as described by e.g. Langleloth et al., https://aip.scitation.org/doi/10.1063/1.4794156

I submit this now in order to get some early feedback regarding my overall ideas (not the code yet, thanks).

**Author(s)**

Tim Bernhard, ETHZ, tim@bernhard.dev

**Licensing**

**Backward Compatibility**

While there are some adjustments to existing commands (particularly `bond create`), no inputs should be broken.

**Implementation Notes**

&lt;rant&gt;
At the time of writing, things do not yet work as they should. A particular difficulty for me is about how to modify the bonds and angles; there always seem to be so many things to think of, things not to forget. Would be nice if there were some abstractions (e.g., `add_bond(atomId1, atomId2, addAngles:bool, addDihedrals:bool)`, `remove_bond(bondId)`) that take care of all the dihedrals, communication, data-structure, neighbours etc.
&lt;/rant&gt;

That said, my current idea is to split the procedure into a series of different commands:

- `fix bond/mcmove`: shift a mounting point of a bond by 1 bead, direction random, equal probability. Accepted if Metropolis criterion is fulfilled, rejected otherwise. Repeated for both mounting points of the bond.
- `fix bond/break/self`: the previous command can lead to bonds being from one atom to itself. I don't know enough about LAMMPS' internals to know whether the previous command should take care of those, or if I can create a new command, this one here. The output of this command, the nr of bonds broken, can be used as argument to my changes introduced to `fix bond/create`.
- `fix bond/relocatemc`: relocate bonds that have reached at least one end. Also only if fulfills Metropolis criterion!

The actual commands' arguments are yet to change I guess.

The changes to the `fix bond/create` are introduced to enable the creation of the slip-springs (bonds) in the manner described in the paper listed above:

- add possibility to also add a minimum distance for bonds to form
- add possibility to restrict the total number of bonds to form

**Post Submission Checklist**


src/arg_info.cpp
```diff
@@ -84,6 +85,7 @@ ArgInfo::ArgInfo(const std::string &arg, int allowed) : type(NONE), dim(0), inde
       name = arg.substr(offset);
     }
   } else {
+    if (!arg.empty() && std::all_of(arg.begin(), arg.end(), ::isdigit)) { type = NUMCONSTANT; }
```
akohlmey, line 88: Please use `utils::is_integer()` here (or `utils::is_double()`)

akohlmey:
> <rant>
> At the time of writing, things do not yet work as they should. A particular difficulty for me is about how to modify the bonds and angles; there always seem to be so many things to think of, things not to forget. Would be nice if there were some abstractions (e.g., `add_bond(atomId1, atomId2, addAngles:bool, addDihedrals:bool)`, `remove_bond(bondId)`) that take care of all the dihedrals, communication, data-structure, neighbours etc.
> </rant>

You are implicitly making the assumption that these operations are common things. The opposite is the case. LAMMPS is a classical MD code with distributed data. You are trying to make it do things that are very complex to do with distributed data. LAMMPS has not been designed to do this well, because it is a code for modular dynamics and not for Monte Carlo simulations. This is why fix styles like fix gcmc have become such convoluted and hard to maintain "monsters". As with most things in LAMMPS: if you want something done, your best chances to get it done is to do it yourself. But you also have to be aware that making abstractions that work well, is hard work, especially in parallel, and especially with a software like LAMMPS that has so much flexibility and thus people would do very diverse things and make things work consistently and reliably under such circumstances can be very difficult.

akohlmey:
Your pull request introduces many whitespace-only changes or changes formatting relative to the existing LAMMPS conventions, please revert those changes, so that the diff will only show functional changes and not hide them within cosmetic changes (which would have to be undone later, anyway).

GenieTim:
Thank you for your insightful responses, @akohlmey. 
Yes, I am aware of the difficulties of introduction of the named abstractions, it was only wishful thinking remaining after some failed attempts, leading to the "rant" above. I understand your reasons for not introducing this in LAMMPS in the first place.

> Your pull request introduces many whitespace-only changes or changes formatting relative to the existing LAMMPS conventions

Sure, I will try to revert them, I guess they were introduced by the auto-formatting of the IDE. 
Do you happen to have an astyle or clang-format configuration that could be used to automatically reset these formatting issues?

akohlmey:
> Do you happen to have an astyle or clang-format configuration that could be used to automatically reset these formatting issues?

There are .clang-format files bundled with LAMMPS, but they are not generally applied. We have found that clang-format can make some files very hard to read. So we turned it off by default and are enabling it on a file-by-file basis, where it produces acceptable output. 
Beyond that, there is a summary here: https://docs.lammps.org/Modify_style.html

evoyiatzis:
@GenieTim Do you have access to IBISCO or you are implementing everything from scratch?

GenieTim:
What is IBISCO, @evoyiatzis ? (no, I am doing this on my own; I have a standalone single-core implementation of the DPD code that works, though, as you can see with the timestamps of my commits, this version here was actually a first attempt until I admitted to myself that I need to start simpler)

evoyiatzis:
@GenieTim it is the name of the software package developed by Florian & Michael at Darmstadt for mesoscopic simulations. It runs in parallel with mpi and having access would make things easier (i guess).

GenieTim:
Ah, I see, @evoyiatzis, thanks. Yes, I guess that would make things easier, if not turn even this PR obsolete 😬

jrgissing:
fix bond/react may be able to do what you want

GenieTim:
No, @jrgissing , fix bond/react unfortunately does not do what I need, but thanks

jrgissing:
> No, @jrgissing , fix bond/react unfortunately does not do what I need, but thanks

Note that fix bond/react has already successfully been used for slip-spring type dynamics. E.g. here:
https://doi.org/10.1016/j.compscitech.2021.108790
I would be happy to talk about ways to incorporate MC-like criteria as well. At least, it may enable a working model to get started.

akohlmey:
Closed for inactivity. Please resubmit, when updated and test failures addressed or otherwise resolved.

PR #3919, opened 2023-09-10

### LINKED ISSUE #3920: [BUG] Fix_Plumed documentation example contains typo
**Summary**

In the documentation for the plumed fix, the example in https://docs.lammps.org/fix_plumed.html#examples seems to be slightly incorrect. Instead of `fix pl all plumed all plumed plumedfile plumed.dat outfile p.log`, the example should instead be `fix pl all plumed plumedfile plumed.dat outfile p.log`
My guess is that `all plumed` was accidentally copy-pasted twice when the documentation was being written.

The source code for this documentation is located at: https://github.com/lammps/lammps/blame/develop/doc/src/fix_plumed.rst

This bug is affecting the most recent version of the LAMMPS online manual, https://docs.lammps.org/fix_plumed.html
Based on the commit history for fix_plumed.rst, it looks like this typo in the example has existed for quite a while.

**Expected Behavior**

As described in the manual, the syntax should be `fix ID group-ID plumed keyword value ...`
https://docs.lammps.org/fix_plumed.html

Thank you for looking into this small issue.

akohlmey:
**Summary**

This pull request combines multiple small fixes and updates that don't warrant pull requests on their own.

**Related Issue(s)**

See this discussion for the REBO neighbor list change: https://matsci.org/t/lammps-exiting-without-error-and-without-results/50684
Fixes #3920 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

- workaround for an issue with neighbor list trimming when using REBO pair styles as a hybrid sub-style. With trimming, the neighbor list entries for ghost atoms are not initialized.
- updates for building tools and removal of references to tools that have been removed
- updates for LAMMPS shell to adjust lists of possible completions to the current state of LAMMPS
- ignore reading `atom->map_user` from binary restart files so atom_modify map will be honored when reading restarts.
- various updates and corrections for the documentation

**Post Submission Checklist**


akohlmey:
@stanmoore1 please have a look at the change to pair_airebo.cpp. It seems the neighbor list trimming is removing neighbors too aggressively for building the REBO lists. Those seem to be missing the data for neighbors of ghost atoms.
Changing the default works around the issue, but does not really resolve it.

stanmoore1:
@akohlmey do you have a reproducer for the REBO case?

akohlmey:
> @akohlmey do you have a reproducer for the REBO case?

Please check out the post in the MatSci forum that is mentioned above.

stanmoore1:
@akohlmey can you update this compute to the new name:
https://github.com/lammps/lammps/blob/develop/src/KOKKOS/Install.sh#L99-L100

akohlmey:
> @akohlmey can you update this compute to the new name: https://github.com/lammps/lammps/blob/develop/src/KOKKOS/Install.sh#L99-L100

done

stanmoore1:
Thank you!

sjplimp (approved):

akohlmey:
> @akohlmey do you have a reproducer for the REBO case?

@stanmoore1 looks like the same issue also happens to pair style extep as reported in issue #3940. The common denominator seems to be the request of a neighbor list with neighbors for ghost atoms.

stanmoore1:
@akohlmey the trim list issues should be fixed by #3941, can you revert the workaround you put in?

akohlmey:
> @akohlmey the trim list issues should be fixed by #3941, can you revert the workaround you put in?

Will check it out. ;-)

PR #3921, opened 2023-09-14
Adding intel variant of snap pair style.

wmbrownIntel:
**Summary**

This adds an intel package variant of the snap pair style that is optimized for AVX-512. In my tests, it can be over 4X faster than the existing variants on CPUs.

**Related Issue(s)**

**Author(s)**

Mike Brown, Intel

**Licensing**

**Backward Compatibility**

**Implementation Notes**

This is implemented somewhat differently than other intel variants due to the unique bottlenecks and loop structure. First, it is only available if it is built with an Intel compiler for a CPU supporting AVX-512 (due to use of math library vector intrinsics). Second, it is currently only double precision, regardless of the Intel package precision mode selected. These notes are added to the snap doc page. While there is still performance on the table here, I need to set this aside for a bit. But as is, there are great speed ups and I believe it to be "production ready".

**Post Submission Checklist**

### DIFF
--- a/doc/src/Commands_pair.rst
+++ b/doc/src/Commands_pair.rst
@@ -265,7 +265,7 @@ OPT.
    * :doc:`smd/tri_surface <pair_smd_triangulated_surface>`
    * :doc:`smd/ulsph <pair_smd_ulsph>`
    * :doc:`smtbq <pair_smtbq>`
-   * :doc:`snap (k) <pair_snap>`
+   * :doc:`snap (ik) <pair_snap>`
    * :doc:`soft (go) <pair_soft>`
    * :doc:`sph/heatconduction <pair_sph_heatconduction>`
    * :doc:`sph/idealgas <pair_sph_idealgas>`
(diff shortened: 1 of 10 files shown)


stanmoore1:
@wmbrownIntel it looks like the changes to snap doc page didn't get included? Also it looks like this doesn't support OpenMP threading yet either, just SIMD?

wmbrownIntel:
> @wmbrownIntel it looks like the changes to snap doc page didn't get included? Also it looks like this doesn't support OpenMP threading yet either, just SIMD?

Thanks. I just committed the missing doc page changes and added that only a single OpenMP thread will be used. It is relatively straight forward to add OpenMP support here, but I don't believe it will currently be advantageous since we mostly use that for SMT (that will not benefit the current bottlenecks).

athomps:
This is a nice addition that will provide speed up for people running MD with SNAP potentials on CPUs.  It would be nice if these optimizations could also be ported to `compute_snap.cpp`, so that the same speedup can be obtained in training SNAP potentials.

sjplimp (approved):
@wmbrownIntel Very nice Mike.  And impressive speed-up!

PR #3923, opened 2023-09-18
LAMMPS GUI improvements, fixes, refactoring, and backports for Ubuntu 20.04LTS

akohlmey:
**Summary**

This pull request implements some improvements to the LAMMPS GUI based on user feedback, adds some backporting to Qt v5.12 as shipped with Ubuntu 20.04LTS which is missing a few features from Qt v5.15. In the process also some functionality is made more consistent.

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- fix bug where the incorrect graph was saved to image files
- fix bug where reducing the number of threads would cause crashes
- enforce proper threads initialization with OPENMP package
- avoid segfaults when polling for thermo data before the Thermo class is fully instantiated
- Qt 5.12 does not support `QString.find()` with `QRegularExpression`: reformulate to use QRegExp then
- update current working directory also when saving files
- change completion popup mode to always display the full completer list with the most likely choice highlighted
- hide popup when changing completer class instance
- add support for completion on file names for selected commands, but only for files in the current folder
- improve insertion of selected text from completer by always replacing the current word with the selected text
- move helper functions into a separate file, add overloads to "mystrdup()" function
- merge periodic table data into imageviewer.cpp file
- remove unused class member from ChartWindow class
- move icon files to icons folder and thus reduce clutter in the main source folder
- add pushbutton to slideshow viewer that offers option to export animation to movie file with ffmpeg
- check version of LAMMPS library to be compatible with GUI requirements
- change default of creating new log/chart/image windows to replace them
- add run counter and display run number in log/chart/slideshow windows
- automatically define variable "gui_run" to the current value of the run counter
- add "Save Log to File" option to the context menu of the log window and to "CTRL-S" hotkey
- refactor output creation for .dat and .csv files from chart window

**Post Submission Checklist**


akohlmey:
FYI, binaries for testing, compiled from *exactly* this branch are on https://download.lammps.org/static/

akohlmey:
Pre-compiled binaries for testing have been updated with the latest changes from this branch.  https://download.lammps.org/static/

cmake/presets/macos-multiarch.cmake
```diff
@@ -10,5 +10,3 @@ set(CMAKE_CXX_FLAGS_RELEASE "-O3 -DNDEBUG" CACHE STRING "" FORCE)
 set(CMAKE_C_FLAGS_RELEASE "-O3 -DNDEBUG" CACHE STRING "" FORCE)
 
 set(BUILD_MPI FALSE CACHE BOOL "" FORCE)
-set(BUILD_SHARED_LIBS FALSE CACHE BOOL "" FORCE)
```
stanmoore1, line 13: Just curious, what is the reason behind the changes here?

stanmoore1 (approved):
Looks OK to me, see one comment.

cmake/presets/macos-multiarch.cmake
akohlmey: Shared libraries work just as well as static libraries when compiling for macOS with apple and intel CPUs at the same time. So there is no need to forcibly turn this off. This preset was originally used for the LAMMPS .dmg files where the binaries were supposed to be linked to a static LAMMPS library.

PR #3930, opened 2023-10-03
Small Kokkos changes

stanmoore1:
**Summary**

- Add support for FLUX scheduler/runtime
- Fix harmless compiler warnings
- Add missing code for dynamic groups
- Fix issue with Kokkos `fix property/atom`
- Add error check in Kokkos PACE for unsupported radial basis style
- Fix bug with Kokkos `pair hybrid/overlay` + full neighbor list where forces were zeroed out erroneously

**Related Issue(s)**

Fixes #3927
Fixes #3929 
Fixes #3935
Fixes #3936

**Author(s)**

Stan Moore

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

Yes

PR #3933, opened 2023-10-05
Second Set of Collected Bug Fixes and Maintenance Updates for 2 August 2023 Stable release

akohlmey:
**Summary**

This pull request collects bugfixes backported from the develop branch to ultimately become the second update to the latest stable release.

**Author(s)**

Backporting by Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

- enforce consistent OpenMP threads settings for /omp styles (avoids crashes and memory corruption)
- accept very old style ReaxFF force field that are missing the ovc property for bonds and set it to 0.0 (it is 0.0 most of the time anyway) 
- avoid segfaults when querying LAMMPS for thermo data before the Thermo class has been instantiated
- identify Windows 11 23H2
- bugfix to ellipsoid region for correct forces with fix wall/region
- bugfix and updates to computes cluster/atom, aggregate/atom, and fragment/atom to add missing communication of ghost atoms and to avoid deadlocks
- fix cut-n-paste type in minimize style error message
- update CWD in LAMMPS GUI also when saving a buffer
- fix bug in LAMMPS GUI where the number of threads could not be increased
- fix delete/delete[] mismatch in LAMMPS GUI
- check LAMMPS library loaded as plugin to be compatible with LAMMPS GUI
- bugfix for neighbor list trimming on requests that request neighbors of ghosts 
- properly quote keywords containing special characters in dump yaml output
- detect more cases of inconsistent KOKKOS style usage and error out gracefully
- add sanity check to `python/install.py` file and improve readability of error messages
- if `tools/phonon/phana` uses `linalg`, make sure it is built before linking
- nullify deleted pointers when creating styles in Force or Update since the creation may fail and trigger an exception and the we may run into a double free error crashing the application
- avoid "invalid escape" warning with python 3.12
- fix bug in fix pimd/langevin
- make certain that generated argv lists are NULL terminated to avoid segfaults with OpenMPI 5.x and later
- fix bug in pace/extrapolate
- fix bug in the rigid fixes that was causing a drift in the z-coordinate with 2d systems
- fix bugs in fix mvv/* compatibility checks
- restore using nvcc_wrapper with kokkos_cuda.cmake preset
- fix memory leak in dump style image and remove redundant code
- the scale yes option is not supported with fix adapt/fep for atom charges and diameters, update list of supported pair styles, fix bug with global scale factor
- Avoid segfault in fix atom/swap for systems without per-type masses. Warn if both per-type and per-atom masses are set that per-type masses will be used
- correct DPI settings in mesh images for compute saed and compute xrd so the images display correctly in PDF files
- avoid copying over terminating null byte in balance and fix balance shift keyword
- may only check for region inside box in fix gcmc and fix widom if the region keyword was used
- avoid deprecated python initialization for python 3.12 and later
- avoid warnings from including multiple cython generated include files
- add valgrind suppresions for MPICH

**Post Submission Checklist**


akohlmey:
@stanmoore1 you may eventually want to check out commit 78adc1727a702a000aec9136820d1e2d11447734 which is my attempt to backport PR #3930 to the stable version (almost all changes went it in cleanly).

stanmoore1:
@akohlmey it looks good to me. I could put it through regression testing when it is ready

akohlmey:
> @akohlmey it looks good to me. I could put it through regression testing when it is ready

I don't expect there to be another update for the stable version until around Christmas or the new year.
Unless there would be some significant and urgent bugfix, that shows up in between.

stanmoore1:
@akohlmey the Kokkos regression tests passed on this branch too

akohlmey:
> @akohlmey the Kokkos regression tests passed on this branch too

Great! Thank you again.

sjplimp (approved):

stanmoore1 (approved):

PR #3937, opened 2023-10-11

akohlmey:
**Summary**

This pull request combines multiple small changes and bug fixes that do not warrant individual pull requests

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- avoid segfault in multi-threaded neighbor lists, if the number of threads has been reduced during the executable run (e.g. when using LAMMPS GUI)
- small doc updates for fix srd, pair style reaxff, and pair style tmd/ilp
- port of LAMMPS GUI to Qt6
- replace entire word under cursor with auto-completion
- avoid infinite loop in StdCapture and instead time out after one second
- avoid multiple out of bounds memory accesses in LAMMPS GUI detected by compiling to Qt with assertions enabled
- auto detect OpenMP availability when compiling LAMMPS GUI as standalone executable

**Post Submission Checklist**

### DIFF
--- a/lib/pace/Makefile
+++ b/lib/pace/Makefile
@@ -1,5 +1,4 @@
 SHELL = /bin/sh
-CMAKE= cmake
 
 # ------ FILES ------
 
@@ -36,7 +35,7 @@ lib: $(OBJ) lib-yaml-cpp
 
 
 lib-yaml-cpp:
-	cd $(YAML_CPP_PATH) && $(CMAKE) . && $(MAKE) all
+	cd $(YAML_CPP_PATH) && $(MAKE) lib
 
 # ------ COMPILE RULES ------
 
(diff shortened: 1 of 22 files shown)


lib/pace/Makefile
```diff
@@ -1,4 +1,5 @@
 SHELL = /bin/sh
+CMAKE= cmake
```
stanmoore1: @akohlmey can you wait on this change? I submitted a PR that adds the makefile back in here: https://github.com/ICAMS/lammps-user-pace/pull/19 which doesn't require CMake.

stanmoore1 (changes requested):

lib/pace/Makefile
akohlmey: Sure. You need the access permissions changed, too, right? So we would need another PR from Yuri before we can finish the release anyway.

stanmoore1: Right

oywg11 (approved):

lib/pace/Makefile
akohlmey: The lib/pace/Makefile changes have been reverted.

stanmoore1: Thanks!

stanmoore1 (approved):
I approve

PR #3950, opened 2023-10-23

### LINKED ISSUE #3956: [BUG] _fourier dihedral style not found, make solution clear in error message if possible
Not sure whether this should be called a bug. 
Normally when a feature is not present for running a particular example, the error message clearly indicates which package needs to have been compiled. 

When I tried to run any of the input scripts from examples/relres, I got an error message indicating that the `dihedral_style fourier` does not exist.
```
$~/lammps/examples/relres$ ../../build/lmp -in in.22DMH.relres
LAMMPS (2 Aug 2023)
OMP_NUM_THREADS environment is not set. Defaulting to 1 thread. (src/comm.cpp:98)
ERROR: Unrecognized dihedral style 'fourier' (src/force.cpp:510)
Last command: dihedral_style  fourier
``` 

After I rebuilt with -D PKG_EXTRA-MOLECULE (in addition to the other required packages), the problem was fixed. What I'm requesting is that, in a future release, the error message be modified to make it clear that rebuilding with this extra package (or whatever solution is better) will fix this problem.

### LINKED ISSUE #3961: [BUG] KOKKOS version of fix dt/reset sets timestep incorrectly in most cases
**Summary**

The KOKKOS version of `fix dt/reset` sets a larger timestep than the standard implementation in most cases.

Current develop, Linux Mint 21.3

**Expected Behavior**

The timestep should be the same in both versions.

**Actual Behavior**

The timestep is higher than it should be in almost all cases. It only works if the fastest atom has the very last id. The input file below shows the issue. If you change `id 5` to `id 10`, the code correctly sets the timestep.

**Steps to Reproduce**

Compile LAMMPS with KOKKOS and run the following input in serial mode:

```
units real
region test sphere 0 0 0 30
create_box 1 test
mass 1 1
create_atoms 1 random 10 235 test
group move id 5
velocity move set 100 0 0 units box
fix dt all dt/reset 1 0.0 0.1 0.0000001 units box
thermo_style custom step dt
run 0
```

This is a little bit embarrassing, as I've coded the KOKKOS implementation of `fix dt/reset`. After some debugging I found that the problem lies in the `parallel_reduce` function, which calculates the timestep for each atom, but only uses its value for the last atom when looking for the smallest timestep.
This issue went unnoticed through my testing, because the fastest atom had always the largest id.

Had I misunderstood how `parallel_reduce` works, or is there bug somewhere there? I have a working bugfix, but it's less elegant than just using `parallel_reduce`, so I'd like to ask first if the current code can be corrected.

akohlmey:
**Summary**

This pull request combines multiple small fixes and updates.

**Related Issue(s)**

Fixes #3956
Fixes https://github.com/lammps/lammps/issues/3961

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues

**Implementation Notes**

- refactor `read_dump` and various dump reader styles to use common enumerators for indicating atomic data fields
- nullify force and minimize/integrate style related pointers after deleting them and *before* attempting to create new style instances to avoid double free errors and crashes of LAMMPS GUI when the creation failed due to requesting a not available style
- improve error messages in `read_dump`; print informational message when atom types are required due to `add` keyword
- add sanity check on python module path in `python/install.py`, improve readability of error messages
- add python code example for parsing a dump style yaml file
- avoid invalid escape sequence warnings when building the LAMMPS documentation with python 3.12
- some programming style updates
- address minor issues like dead code or unchecked dynamic casts as detected by static code analysis
- update "CONTRIBUTING" document available on GitHub
- add check for commonly misformatted roles in ReST source files
- fix harmless compiler warnings from `kokkos.cpp`, replace deprecated Kokkos code
- fix bugs where available but non-included styles were not reported as expected (e.g. for dihedrals see issue #3956)
- Fix bug in Kokkos `fix dt/reset`

**Post Submission Checklist**


src/read_dump.cpp
```diff
@@ -276,7 +259,7 @@ bigint ReadDump::seek(bigint nrequest, int exact)
 
   // proc 0 finds the timestep in its first reader
 
-  if (me == 0 || parallel) {
+  if (comm->me == 0 || parallel) {
```
sjplimp, line 265: What is the motivation for this change for "me" and "nprocs" throughout this file.  I think using local copies of these Comm variables is done in several other files as well, for convenience.  Ditto for changing to domain->triclinic instead of triclinic.

akohlmey: The motivation is to get rid of redundant calls to MPI_Comm_rank()/size(), redundant class members (and thus wasted RAM), and indicate that the canonical location for this kind of information is the Comm class instance (or Domain).

> I think using local copies of these Comm variables is done in several other files as well, for convenience.

If you look carefully, you should notice that it is used in fewer files than where it used to be used a few years ago. ;-)

sjplimp (approved):

PR #3958, opened 2023-10-29
Fix deposit var keyword

jrgissing:
**Summary**

add var and set keywords to fix deposit to allow deposition within an isosurface, same as create_atoms

**Related Issue(s)**

builds on #3957

**Author(s)**

JG

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/src/fix_deposit.cpp
+++ b/src/fix_deposit.cpp
@@ -20,6 +20,7 @@
 #include "domain.h"
 #include "error.h"
 #include "fix.h"
+#include "input.h"
 #include "lattice.h"
 #include "math_const.h"
 #include "math_extra.h"
@@ -29,6 +30,7 @@
 #include "random_park.h"
 #include "region.h"
 #include "update.h"
+#include "variable.h"
 
 #include <cmath>
 #include <cstring>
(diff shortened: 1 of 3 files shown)


src/fix_deposit.cpp
```diff
@@ -583,8 +591,10 @@ void FixDeposit::pre_exchange()
 
   // warn if not successful b/c too many attempts
 
-  if (!success && comm->me == 0)
-    error->warning(FLERR,"Particle deposition was unsuccessful");
+  if (warnflag && !success && comm->me == 0) {
+    error->warning(FLERR,"One or more particle depositions were unsuccessful");
+    warnflag = 0;
```
jtclemm, line 598: I do find the nonstop warnings tiring, but would it make sense to reset `warnflag` to 1 during the setup phase so you get one warning per run command? Might be handy to know whether changes between run commands improved the success.

jtclemm (approved):

src/fix_deposit.cpp
jrgissing: i agree, thanks for catching that

sjplimp (approved):

sjplimp:
@jrgissing Nice addition - what is the envisioned use case ?

PR #3960, opened 2023-10-30
New pair_style nb3b/screened

akohlmey:
**Summary**

This pull request adds a new pair style nb3b/screened which is a variant of nb3b/harmonic

**Related Issue(s)**

N/A

**Author(s)**

Federica Lodesani, University of Modena, fede.lode@gmail.com

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

To reduce replicated code, the pair style is implemented as a derived class with a few related modifications to the nb3b/harmonic base class.

**Post Submission Checklist**


examples/nb3b/README.nb3b.screened
```diff
@@ -0,0 +1,41 @@
...
+type 2: Na, not involved in the three body so indicated as NULL;
+type 3: O.
+
+In the potential file PSiO.nb3b.screened there are the parameters of
+all permutations of the required atom types. In the example, also
+parameters for Si permutations appear but they are not employed in
+the current case.
+
+# i  j  k   K     theta0   rho  cutoff
+O   P   P    65.0  109.47   1.0   3.3
```
flodesani: Can we change the K values with Kvalues divided by 2 also here in the README?

akohlmey: done

src/MANYBODY/pair_nb3b_harmonic.cpp
flodesani, line 1: Can we add at line 345? 
params[m].rho = 1.0 / params[m].rho;
The inverse of rho is needed by the threebody function.
Otherwise it can be directly calculated in the threebody function.

akohlmey: If you want to go that route, then why not go all the way and just call the parameter invrho, document it, and read it directly from the potential file?

flodesani: I think it is more confusing to ask for the inverse since the value is always reported as denominator.

akohlmey: Ok. But at least internally it should be using "invrho" as storing "invrho" in a variable called "rho" is confusing, too. Please see my updated version in commit ac527e3d3ff823d561bc75eb3b9fcf9880d8181a

BTW: for a moment I was confused why there was no difference in the forces and energies without the missing 1.0/rho computation, but then I saw that all "rho" parameters in the potential file are set to 1.0 so for that special case, there is no difference.

flodesani: It's perfect! Thank you. 
Yes, I should have probably told you. Sorry

sjplimp (approved):

PR #3963, opened 2023-10-31
Heffte integration

mkstoyanov:
**Summary**

Enable the use of the heFFTe library as the FFT engine.

**Related Issue(s)**

N/A

**Author(s)**

Miroslav Stoyanov (ORNL)

**Licensing**

**Backward Compatibility**

The PR is backwards compatible.

**Implementation Notes**

The code was linked and tested against the heFFTe library.

**Post Submission Checklist**


cmake/CMakeLists.txt
```diff
@@ -960,33 +960,53 @@ if(PKG_KOKKOS)
...
-  endif()
-  if(FFT_FFTW_THREADS OR FFT_MKL_THREADS)
-    message(STATUS "Using threaded FFTs")
+  if (LMP_HEFFTE)
+    message(STATUS "<<< FFT settings >>>
+-- Primary FFT lib:  heFFTe")
+    if (HEFFTE_BACKEND)
+      message(STATUS "heFFTe backend:  ${HEFFTE_BACKEND}")
+    else()
+      message(STATUS "heFFTe backend:  stock (not intended for production)")
```
stanmoore1: What does this message mean? Not sure what "stock" and "production" are referring to?

cmake/Modules/Packages/KSPACE.cmake
```diff
@@ -46,6 +46,40 @@ else()
   target_compile_definitions(lammps PRIVATE -DFFT_KISS)
 endif()
 
+option(LMP_HEFFTE  "Use heFFTe as the distributed FFT engine, supersedes the FFT option."  OFF)
```
stanmoore1: Do you mean "overrides" instead of "supersedes"?

doc/src/Build_settings.rst
```diff
@@ -43,6 +43,12 @@ When the KSPACE package is included in a LAMMPS build, the
 require use of an FFT library to compute 1d FFTs.  The KISS FFT
 library is included with LAMMPS, but other libraries can be faster.
 LAMMPS can use them if they are available on your system.
+Alternatively, LAMMPS can use the
+`heFFTe <https://icl-utk-edu.github.io/heffte/>`_
+library for the MPI communication algorithms,
+which comes with many optimizations for special cases,
+e.g., leveraging 2D or 3D backend transforms or
```
stanmoore1: or --> and

cmake/Modules/Packages/KSPACE.cmake
```diff
@@ -46,6 +46,40 @@ else()
...
+  set_property(CACHE HEFFTE_BACKEND PROPERTY STRINGS ${HEFFTE_BACKEND_VALUES})
+
+  if(HEFFTE_BACKEND STREQUAL "FFTW") # respect the backend choice, FFTW or MKL
+    set(HEFFTE_COMPONENTS "FFTW")
+    set(Heffte_ENABLE_FFTW "ON" CACHE BOOL "Enables FFTW backend for heFFTe")
+  elseif(HEFFTE_BACKEND STREQUAL "MKL")
+    set(HEFFTE_COMPONENTS "MKL")
+    set(Heffte_ENABLE_MKL "ON" CACHE BOOL "Enables MKL backend for heFFTe")
+  else()
+    message(WARNING "HEFFTE_BACKEND not selected, defaulting to 'stock' backend, which is not intended for production")
```
stanmoore1: Same as above, not sure what "stock" and "not intended for production" means.

stanmoore1 (commented):
@mkstoyanov thank you for working on this. Overall this looks like a clean integration, I had a few minor questions/comments.

cmake/CMakeLists.txt
mkstoyanov: The `stock` backend is what heFFTe calls the builtin FFT implementation. It's very useful for testing, since we have checked it for correctness and we can use it to verify that interfaces to the other backends work correctly. Similarly, our users can utilize the `stock` backend and check their interface to heFFTe without bothering to install third party libraries, such as FFTW or cuFFT.

However, we don't optimize the `stock` backend for performance across relevant platforms. It behaves well but won't match FFTW or MKL.

I have updated the documentation.

cmake/Modules/Packages/KSPACE.cmake
mkstoyanov: See the comments above.

cmake/Modules/Packages/KSPACE.cmake
```diff
@@ -46,6 +46,42 @@ else()
   target_compile_definitions(lammps PRIVATE -DFFT_KISS)
 endif()
 
+option(LMP_HEFFTE  "Use heFFTe as the distributed FFT engine, overrides the FFT option."  OFF)
```
rbberger: Maybe not call this `LMP_HEFFTE` , since we don't use `LMP_` for any other CMake options.

mkstoyanov: It was a suggestion from here:

https://github.com/lammps/lammps/pull/3639#issuecomment-1430103475

What is the lammps naming convention?

rbberger: CMake options are usually grouped by their purpose. In this case, since it only affects the FFT, I would suggest something like  `FFT_USE_HEFFTE`.

rbberger: Also, if I read this right the suggestion was for the name of define, not the option itself.

mkstoyanov: I see your point. In all my project, I always keep the same names for the defines and the CMake options, didn't expect lammps would do it differently.

I updated the name.

sjplimp:
Similar to Richard's comments for CMake, for traditional make the convention is that all -D defines for FFT options start with FFT.
E.g. FFT_INC = -DFFT_FFTW -DFFT_FFTW_THREADS

So I think the analog for HEFFTE should be
FFT_INC = -DFFT_HEFFTE -DFFT_HEFFTE_MKL
i.e. the first means use HEFFTE (instead of KISS, FFTW, etc), the second setting is for the internal HEFFTE option

See this doc page for all the FFT options: https://docs.lammps.org/Build_settings.html under the "Traditional make" tab in 3.5.2

mkstoyanov:
> Similar to Richard's comments for CMake, for traditional make the convention is that all -D defines for FFT options start with FFT. E.g. FFT_INC = -DFFT_FFTW -DFFT_FFTW_THREADS
> 
> So I think the analog for HEFFTE should be FFT_INC = -DFFT_HEFFTE -DFFT_HEFFTE_MKL i.e. the first means use HEFFTE (instead of KISS, FFTW, etc), the second setting is for the internal HEFFTE option
> 
> See this doc page for all the FFT options: https://docs.lammps.org/Build_settings.html under the "Traditional make" tab in 3.5.2

Well, the previous PR followed this naming convention ...

I've update the current PR with the `FFT_HEFFTE` and `FFT_HEFFTE_FFTW` and `FFT_HEFFTE_MKL`

sjplimp:
Thanks @mkstoyanov - sorry, the only thing I remembered about the previous PR was that it was lacking traditional make support, so I thought the defs of -D names for this PR were new choices (in the traditional make doc section).

doc/src/Build_settings.rst
```diff
@@ -111,6 +121,22 @@ LAMMPS can use them if they are available on your system.
...
+         FFT_INC = -DFFT_HEFFTE -DFFT_HEFFTE_FFTW $(heffte_include)
+         FFT_PATH =
+         FFT_LIB = $(heffte_link) $(heffte_libs)
+
+      The heFFTe install path will contain `HeffteMakefile.in`.
+      which will define the `heffte_` include variables needed to link to heFFTe from
+      an external project using traditional make.
+      The `-DFFT_HEFFTE` is required to switch to using heFFTe, while the optional `-DFFT_HEFFTE_FFTW`
+      selects the desired heFFTe backend, e.g., `-DFFT_HEFFTE_FFTW` or `-DFFT_HEFFTE_MKL`,
+      omitting the variable will default to the `stock` backend.
```
sjplimp, line 144: Please add a sentence to explain "stock" backend after this sentence.

sjplimp:
@mkstoyanov One more small request.  Per Stan's earlier comment, the build doc page also says "stock" with no explanation.  I don't think readers will know what that means.  So please add a few words similar to what is in the CMake status message.  Thanks, then I will approve the PR.

mkstoyanov:
I expanded the documentation.

sjplimp (approved):
Perfect - thanks!

stanmoore1 (approved):

stanmoore1:
@mkstoyanov I think it would be good to also add support for HeFFTe + KOKKOS package so a user can use HeFFTe on the GPU. Is this something you can help with or at least give us some pointers on how to do this?

mkstoyanov:
> @mkstoyanov I think it would be good to also add support for HeFFTe + KOKKOS package so a user can use HeFFTe on the GPU. Is this something you can help with or at least give us some pointers on how to do this?

@stanmoore1 after the end of the ECP, I have no funding for work on heFFTe. I'm trying to start a project with Kokkos kernels where heFFTe will become their default FFT ending, in which case connecting to lammps will be trivial, but such work will be some time into the future.

Using the current build, the main trick is to enable the correct backend in heFFTe based on the backend enabled in Kokkos. This has to be done in both the CMake build system and C++ code. We can setup a Zoom/Teams call to go over some of the details.

PR #3968, opened 2023-11-07
Bugfix: Correctly build argv with nullptr at the end

rbberger:
**Summary**

- both C and C++ standards guarantee that `argv[argc]` is a null pointer.
- OpenMPI 5 has [logic](https://github.com/open-mpi/ompi/blob/main/opal/util/argv.c#L300) that depends on this, so not adhering to this convention causes segfaults.
- also fixes another bug due to argv[0] not being copied and therefore could become invalid.

**Related Issue(s)**

This fixes https://bugzilla.redhat.com/show_bug.cgi?id=2246679

**Author(s)**

@rbberger 

**Licensing**

**Backward Compatibility**

Unchanged.

**Implementation Notes**

Added a new `LAMMPS::LAMMPS` constructor to take a `LAMMPS::argv` (`std::vector<std::string>`) for the common case and does the correct conversion to pass it as `argc` and `argv` to the existing constructor.

This is now used in most tests.

**Post Submission Checklist**

### DIFF
--- a/python/lammps/core.py
+++ b/python/lammps/core.py
@@ -379,8 +379,9 @@ class lammps(object):
           for i in range(narg):
             if type(cmdargs[i]) is str:
               cmdargs[i] = cmdargs[i].encode()
-          cargs = (c_char_p*narg)(*cmdargs)
-          self.lib.lammps_open.argtypes = [c_int, c_char_p*narg, MPI_Comm, c_void_p]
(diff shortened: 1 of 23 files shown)


stanmoore1 (approved):
I approve

rbberger:
Just pushed one more cleanup, since I had an unneeded include from an earlier version of this. This is ready for merge.

rbberger:
Found one more thing when using the Python interface.

PR #3969, opened 2023-11-09
Brownian package new integrator

samueljmcameron:
**Summary**

Added a new integrator to the BROWNIAN package, which allows for rotating dipoles with fluctuating magnitudes. Also allowed for dynamic groups to be simulated with all of the integrators (having this turned off previously was an oversight).

**Author(s)**

Sam Cameron samuel.j.m.cameron@gmail.com

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

Tested output of integrator and verified that dipole time correlations were consistent with regular rotational diffusion ( e^{-(d-1)*D_r*t}) in both two and three dimensions. Otherwise the integrator is the same as brownian/sphere.

**Post Submission Checklist**

### DIFF
--- a/doc/src/fix_brownian.rst
+++ b/doc/src/fix_brownian.rst
@@ -2,7 +2,7 @@
 .. index:: fix brownian/sphere
 .. index:: fix brownian/asphere
 .. index:: fix brownian/dipole
-	   
+
 fix brownian command
 ===========================
 
--- a/doc/src/fix_brownian.rst
+++ b/doc/src/fix_brownian.rst
@@ -92,7 +92,7 @@ For the style *brownian*, only the positions of the particles are
 updated. This is therefore suitable for point particle simulations.
 
 For the style *brownian/sphere*, the positions of the particles are
-updated, and a dipole slaved to the spherical orientation is also
+updated, and a dipole following the spherical orientation is also
 updated. This style therefore requires the hybrid atom style
 :doc:`atom_style dipole <atom_style>` and :doc:`atom_style sphere
 <atom_style>`. The equation of motion for the dipole is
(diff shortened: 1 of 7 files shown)


sjplimp (approved):
Thanks for adding new example scripts as well

jtclemm:
Since the new fix also requires atom style sphere, does it make sense for the name to somehow reflect this? Is it fair to think of this as an alternative take on fix brownian/sphere with a different dipole evolution? Not sure whether a name like brownian/sphere/dipole would make sense? @sjplimp 

I also noticed that the other subclass-specific variables like gamma_r and gamma_t will issue an error if defined in an incompatible subclass (e.g. fix brownian/asphere errors if gamma_r is defined). Would it make sense for lamba to behave similarly?

sjplimp:
@jtclemm Good point.  Since there is already a fix brownian/asphere with a dipole option, I think the name of this fix should be fix brownian/sphere/dipole and the doc page should make clear that this is a related but alternative integrator which allow for fluctuating magnitudes (which fix brownian/asphere dipole does not.  I am unclear why the fix brownian/sphere command does not have a dipole option for integrating dipoles with fixed magnitude?  The doc page indicates it is just for brownian/asphere?  The spherical case would be similar to fix nve/sphere with its update dipole option.

akohlmey:
@samueljmcameron can you please address the concerns stated by @jtclemm and @sjplimp

akohlmey:
Closed for inactivity. Please resubmit while addressing the mentioned concerns, if possible.

samueljmcameron:
Hello, sorry I have not answered this early (turning github notifications on from this point forward - didn't realise I had them completely switched off and have missed lots). I have implemented the changes as suggested above. @sjplimp the brownian/sphere command only has the option for integrating dipoles with fixed magnitude (this distinguishes it from the basic brownian fix), whereas brownian/asphere gives option to either slave dipole to the quaternion update or not have a dipole at all.

Should I reopen this or create a new pull request?

PR #3972, opened 2023-11-13
getters for fix property/atom

srtee:
**Summary**

Adds public functions to `fix property/atom` that get the total number of variables managed, their styles, corresponding index numbers, and whether it communicates properties to ghosts (`get_border`).

This lets other code check what properties are managed by a particular `fix property/atom`. As a use case, a fix may rely on a custom atom property being communicated for ghost atoms. Currently it is not obvious how to do this: we can `find_custom` to locate the variable-name's index, but we cannot know for sure if the `fix property/atom` managing it has `ghost yes`. With this PR, a fix can loop through all `fix property/atom` instances, determine which one manages the property of interest, and determine whether the property is communicated for ghost atoms using `get_border()`.

**Related Issue(s)**

**Author(s)**

Shern Tee, Griffith University (Brisbane)

**Licensing**

**Backward Compatibility**

No changes to results from user scripts and no documentation changes needed

**Post Submission Checklist**

### DIFF
--- a/src/fix_property_atom.cpp
+++ b/src/fix_property_atom.cpp
@@ -948,3 +948,21 @@ int FixPropertyAtom::size_restart(int /*nlocal*/)
 {
   return values_peratom + 1;
 }
+
+/* ----------------------------------------------------------------------
+   utility getters
+------------------------------------------------------------------------- */
+
+int FixPropertyAtom::get_nvalue() { return nvalue; }
+
+int FixPropertyAtom::get_style(int n) {
+  if (n >= nvalue)
+    error->all(FLERR, "Invalid fix property/atom style request");
+  return styles[n];
+}
+
+int FixPropertyAtom::get_index(int n) {
+  if (n >= nvalue)
+    error->all(FLERR, "Invalid fix property/atom index request");
+  return index[n];
+}
--- a/src/fix_property_atom.cpp
+++ b/src/fix_property_atom.cpp
@@ -966,3 +966,5 @@ int FixPropertyAtom::get_index(int n) {
     error->all(FLERR, "Invalid fix property/atom index request");
   return index[n];
 }
+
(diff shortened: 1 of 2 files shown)


akohlmey:
FYI, the commonly used way to get internal information from class instances in LAMMPS is via "extract()" member function.

srtee:
> FYI, the commonly used way to get internal information from class instances in LAMMPS is via "extract()" member function.

Ah yes! I can see that `fix.h` doesn't have an `extract()` member function yet. Should I add that and then make an override for `fix_property_atom`?

akohlmey:
> > FYI, the commonly used way to get internal information from class instances in LAMMPS is via "extract()" member function.
> 
> Ah yes! I can see that `fix.h` doesn't have an `extract()` member function yet. Should I add that and then make an override for `fix_property_atom`?

What about this?  https://github.com/lammps/lammps/blob/781eadc9c3206d210565b9cf9ae3b0a31471beb7/src/fix.h#L264

akohlmey:
> Should I add that and then make an override for `fix_property_atom`?

You also need to check if extract() can work for all your use cases.
In fact, it would be good to provide a real use case, before the code is added.

srtee:
> What about this? 
> https://github.com/lammps/lammps/blob/781eadc9c3206d210565b9cf9ae3b0a31471beb7/src/fix.h#L264

Argh, missed that.

> > Should I add that and then make an override for `fix_property_atom`?
> 
> You also need to check if extract() can work for all your use cases.
> In fact, it would be good to provide a real use case, before the code is added.

I'll check back with the other ELECTRODE people. I didn't want to roll a small, discrete change in with our next block of updates -- but if it's preferable to include this with our ELECTRODE update then I'm happy to do that instead.

akohlmey:
@srtee please see the LAMMPS developer slack.

PR #3976, opened 2023-11-16

akohlmey:
**Summary**

This pull request collects multiple small updates and fixes

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- unconditionally disable `if constexpr` in fmtlib for Intel compilers
- catch error exceptions in `command()` lambda functions in unittests
- fix bug introduced by PR #3896 where immediate variables were not treated like equal style variables when evaluating references to computes or fixes
- add check for failed unix socket creation to fix ipi
- add mode=pimd example for fix pimd/langevin, avoid per-bead repeated output, align output better with LAMMPS

**Post Submission Checklist**


akohlmey:
@athomps I noticed some part of fmtlib that was not properly protected for C++17 compliance but instead depends on some internal compiler macro that may not be properly set by some compiler (including some Intel version on our CI servers that was warning about using a C++17 feature with C++11. Can you check out this branch against the Intel compiler failures you were seeing?

stanmoore1:
> @athomps I noticed some part of fmtlib that was not properly protected for C++17 compliance but instead depends on some internal compiler macro that may not be properly set by some compiler (including some Intel version on our CI servers that was warning about using a C++17 feature with C++11. Can you check out this branch against the Intel compiler failures you were seeing?

@akohlmey I checked and unfortunately this doesn't fix the KOKKOS package + Intel compiler issue we discussed on Slack:
```
ERROR: Cannot open file data.1.00000000000001.tpm: No such file or directory (../read_data.cpp:367)
```

akohlmey:
> @akohlmey I checked and unfortunately this doesn't fix the KOKKOS package + Intel compiler issue we discussed on Slack:
> 
> ```
> ERROR: Cannot open file data.1.00000000000001.tpm: No such file or directory (../read_data.cpp:367)
> ```

It won't work, since the check is true for C++17 and later. To disable this for your case, we'd have to disable this for all Intel compilers (or possibly do a check on the Intel compiler version, in case we find one that is not affected. I've modified the code accordingly in the latest commit, but I don't have much hope that this makes a difference. Since the issue you report would also happen with `sprintf()`, I suspect something more fundamental, e.g. some macro or define in fmtlib conflicting with some internal define or macros in the Intel compiler headers.

stanmoore1 (approved):
I approve

PR #3979, opened 2023-11-18
Workaround for bug in C++17 support in Intel Classic compilers with fmtlib

akohlmey:
**Summary**

This lowers the C++ standard to C++14 for a few selected source files when compiling with the Intel Classic compilers older than version 21.10 to avoid a bug in the C++17 support leading to misformatted floating point numbers.

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

N/A

**Post Submission Checklist**


stanmoore1:
@akohlmey this does fix the issue when using an Intel Makefile, e.g. `Makefile.intel_cpu_openmpi`, but does not work when using a Kokkos Makefile, e.g. `src/MAKE/OPTIONS/Makefile.kokkos_omp`. Would it be possible to move the workaround into the master Makefile, instead of adding it to every Makefile individually? Or add it to `/lib/kokkos/Makefile.kokkos`? Or `Makefile.package.settings` when Kokkos is enabled?

akohlmey:
It is compiler specific, but I can put the rules and the variable in, just add a comment to the variable and leave it empty.
The latest Intel classic (21.10 from the 23.6 package) is fixed and I just asked my colleague about the LLVM based compilers which will soon be the only option from Intel.
Unfortunately, you will need the 2023 IntelLLVM compiler to be compatible with Kokkos when I read the error message from when I was trying version 18.1 of the classic compiler correctly.

akohlmey:
@stanmoore1 check out commit 0487dc5331504ddef7168f0cc3495d6410d86874 
Does this what you are looking for?

stanmoore1 (approved):

stanmoore1:
@akohlmey yes that works, thanks. Hopefully we can get rid of this workaround in the future.

rbberger:
I question what is the point of continuing support for Intel Classic. It's been deprecated and will be removed very soon.
> Intel® C++ Compiler Classic has been deprecated and will be removed starting from version 2024.0.0. Users should use Intel® DPC++/C++ Compiler which is included in both Intel® oneAPI Base Toolkit and Intel® oneAPI HPC Toolkit.

akohlmey:
> I question what is the point of continuing support for Intel Classic. It's been deprecated and will be removed very soon.

The same reason we keep the minimum C++ standard support for the LAMMPS core at C++11, keep supporting the GNU make based build and so on: these are (still) in use on a significant number of machines, there are LAMMPS users with a preference to use them, and not all of those machines are always updated to the latest version of "everything". We want to keep frustrations when compiling and using LAMMPS small.

While the classic Intel compiler had some issues in recent years (often due to too aggressive optimization heuristics), the Intel LLVM compiler has been worse when compiling LAMMPS. So the classic compiler is often the only viable option.

cmake/CMakeLists.txt
```diff
@@ -428,6 +428,18 @@ if(BUILD_OMP)
   target_link_libraries(lmp PRIVATE OpenMP::OpenMP_CXX)
 endif()
 
+# lower C++ standard for fmtlib sources when using Intel classic compiler
+if((CMAKE_CXX_COMPILER_ID STREQUAL "Intel") AND (CMAKE_CXX_STANDARD GREATER_EQUAL 17)
+    AND (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 2021.10))
+  message(STATUS "Lowering C++ standard for compiling fmtlib sources with Intel Classic compiler")
+  get_filename_component(LMP_UTILS_SRC "${LAMMPS_SOURCE_DIR}/utils.cpp" ABSOLUTE)
```
rbberger, line 435: `LAMMPS_SOURCE_DIR` is already absolute. You could make these just `set` or inline them.

src/MAKE/OPTIONS/Makefile.intel_cpu_openmpi
```diff
@@ -7,13 +7,14 @@ SHELL = /bin/sh
...
 export OMPI_CXX = icc
-CC =		mpicxx -std=c++11 -diag-disable=10441 -diag-disable=2196
+CC =		mpicxx -std=c++11 -diag-disable=10441 -diag-disable=2196 -diag-disable=11074 -diag-disable=11076
 OPTFLAGS =      -xHost -O2 -fp-model fast=2 -no-prec-div -qoverride-limits \
                 -qopt-zmm-usage=high
 CCFLAGS =	-qopenmp -qno-offload -ansi-alias -restrict \
                 -DLMP_INTEL_USELRT -DLMP_USE_MKL_RNG $(OPTFLAGS) \
                 -I$(MKLROOT)/include
 SHFLAGS =	-fPIC
+FMTFLAGS = 	-std=c++11
```
rbberger, line 17: Why is it c++14 in CMake and C++11 in the Makefiles?

akohlmey: Because in CMake we can add this hack selectively to affected compiler versions and when the required standard has been raised to C++17 by inclusion of the KOKKOS package. Lowering to C++14 is sufficient.
In the make based build this flag is *always* applied when using the modified makefiles, even when KOKKOS is not installed, So it has to be compatible with the C++11 requirement for core LAMMPS. Installing the KOKKOS package raises the requirement to C++17 through modifying `Makefile.package.settings` and that needs to be cancelled.

PR #3984, opened 2023-11-21
Add suppport for general triclinic simulation boxes and systems

sjplimp:
**Summary**

Allow users to input general triclinic boxes and atom configurations to LAMMPS, and likewise produce general triclinic outputs.  All the new enhancements are optional new features, no change to current features.

Background: 

A general triclinic (tri) simulation box has arbitrary edge vectors A,B,C.  LAMMPS currently only allows use of a restricted tri box where A is along the x-axis, B is in the xy plane with a +y-coord, and C has a +z-coord.  This is for consistency with neighbor list building, ghost communication, and other orthogonal simulation box operations, as well as for efficient parallelization.

Any general tri box can be rotated to become a restricted tri box, so conceptually this is not a restriction.  However, it can be more convenient for users to specify general tri boxes and their atoms, and also to generate output for the same.

Key idea: 

Allow input and output of general tri boxes and corresponding atom properties (coordinates, per-atom vectors).  Do this by performing transformations from general <--> restricted at the time of input and output.  Internally, LAMMPS continues to store and perform simulations with restricted tri boxes.  All other commands (e.g. velocity, region) still operate on the restricted tri simulation box and its atoms.

Restriction:

LAMMPS requires that A,B,C for a general tri box define a right-handed coordinate system.  I.e. A x B is in the direction of C.  This allows simple rotation matrices to convert between general and restricted triclinic representations.

Details:

(1) Domain class
    add triclinic_general variable to indicate general tri box is defined
    add variables for A,B,C edge vectors of box, boxlo is still used as the origin of the A,B,C vectors
    store rotation matrices used for general <--> restricted rotations
    add method to define the genearal triclinic box
    add methods which perform general <--> restricted rotations for coords and vectors
    when the restricted tri box changes (NPT, fix deform, etc) general A,B,C edge vectors are changed as well

(2) read_data command
  add header keywords for avec,bvec,cvec and "abc origin"
  these replace "xlo xhi", etc for orthogonal and "xy xz yz" for restricted tri
  rotated quantities are the following (hence data file values are affected)
    Atoms section: x, mu (DIPOLE and DIELECTRIC package), x0 (MACHDYN package), sp (SPIN package)
    Velocities section: v, omega, angmom
    Lines, Triangles, Ellipsoids, Bodies sections

(3) write_data command
  add optional triclinic/general keyword
  header keywords for avec,bvec,cvec and "abc origin"
    these replace "xlo xhi", etc (for orthogonal) and "xy xz yz" for restricted tri
  same quantities from read_data command are rotated for output as general tri values

(4) lattice command
  add optional triclinic/general keyword
  a1,a2,a3 are treated as edge vectors of a general tri unit cell
  used by create_box and create_atoms command

(5) create_box command
  for general tri box, region = NULL, add 6 params
  6 params are multiples of general tri unit cell defined by lattice command

(6) create_atoms command
  box and region styles can use a general tri lattice
  this can fill the general tri box or a subset region of it
  the region refers to the restricted tri box

(7) read_restart and write_restart commands
  write/read triclinic_general variable and rotation_g2r 3x3 matrix

(8) dump atom and dump custom commands
  new dump_modify triclinic/general keyword
  affected keywords: x, xu, vx, fx, mux, omegax, angmomx, tqx
  per-snapshot header binary writes avec, bvec, cvec, boxlo
  per-snapshot header text is of form:
    ITEM: BOX BOUNDS abc origin boundary-string
      avec[0] avec[1] avec[2] boxlo[0]
      bvec[0] bvec[1] bvec[2] boxlo[1]
      cvec[0] cvec[1] cvec[2] boxlo[2]

(9) thermo_style custom command
  new thermo_modify triclinic/general keyword
  affected keywords = 6 pressure tensor values = pxx, pyy, pzz, pxy, pxz, pyz

Still to do (before no longer work-in-progress):

(a) more testing
(b) add an examples dir
(c) add thermo keywords for output of avec, bvec, cvec, origin

Ideas for future work:  (probably not in this PR)

(a) read_dump support for general tri
(b) dump image support for general tri
(c) C-style lib interface support for general tri
(d) ditto for Python wrapper on lib interface
(e) add a dump_modify option to enable a compute/fix/variable which produces a per-atom vector to be output with a general tri rotation, similar to velocity
      

**Related Issue(s)**

N/A

**Author(s)**

Steve

**Licensing**

**Backward Compatibility**

Most of the code changes are new optional features which should not break existing inputs.
A few are new constraints on old inputs which may generate error messages for existing inputs.
An example of the latter is that for 2d systems, z coords of atoms are required to be zero, and
the simulation box extent in z needs to straddle zero.  I.e. zlo < 0.0 and zhi > 0.0.

**Implementation Notes**

**Post Submission Checklist**


stanmoore1:
@sjplimp I took a quick look and I don't see any changes needed for the KOKKOS package yet. There are two unit tests failing:
```
[==========] 20 tests from 1 test suite ran. (52 ms total)
[  PASSED  ] 18 tests.
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] LatticeRegionTest.lattice_sq
[  FAILED  ] LatticeRegionTest.lattice_custom
```

sjplimp:
@akohlmey Can you provide me with the test scripts for those 2 failed tests ?

akohlmey:
> @akohlmey Can you provide me with the test scripts for those 2 failed tests ?

I cannot. Those are failures in "death tests" because you changed error messages.
```
37: [ RUN      ] LatticeRegionTest.lattice_sq
37: /home/akohlmey/compile/lammps/unittest/commands/test_lattice_region.cpp:325: Failure
37: Value of: mesg
37: Expected: contains regular expression ".*ERROR: Lattice settings are not compatible with 2d simulation.*"
37:   Actual: "lattice sq 1.0 orient x 1 1 2 orient y -1 1 0 orient z -1 -1 1\nERROR: Lattice orient vectors are not compatible with 2d simulation (src/lattice.cpp:256)\nLast command: lattice sq 1.0 orient x 1 1 2 orient y -1 1 0 orient z -1 -1 1\n"
37: [  FAILED  ] LatticeRegionTest.lattice_sq (1 ms)
```

```
37: [ RUN      ] LatticeRegionTest.lattice_custom
37: /home/akohlmey/compile/lammps/unittest/commands/test_lattice_region.cpp:466: Failure
37: Value of: mesg
37: Expected: contains regular expression ".*ERROR: Lattice settings are not compatible with 2d simulation.*"
37:   Actual: "lattice custom 1.0 origin 0.5 0.5 0.5 basis 0.0 0.0 0.0\nERROR: Lattice origin z coord must be 0.0 for 2d simulation (src/lattice.cpp:249)\nLast command: lattice custom 1.0 origin 0.5 0.5 0.5 basis 0.0 0.0 0.0\n"
37: [  FAILED  ] LatticeRegionTest.lattice_custom (3 ms)
```

I updated the tests in commit fb243eba9b0ca50efb02e9fa4df05614ab7c8dcc

sjplimp:
>> Those are failures in "death tests" because you changed error messages.

I think death tests were what Fox News said Obamacare was going to implement.

akohlmey:
@sjplimp after merging with upstream, there were more unit test failures, that were due to uninitialized data. Should be fixed in fbb9c697717a158fa0e43658e8fafab587bb151e

akohlmey:
@sjplimp here is another test failure (from the run tests):
```
lmp -in in.read_data-005 
LAMMPS (21 Nov 2023)
OMP_NUM_THREADS environment is not set. Defaulting to 1 thread. (src/comm.cpp:98)
New timer settings: style=off  mode=nosync  timeout=off
Reading data file ...
  orthogonal box = (-1 -1 -1) to (1 1 1)
  1 by 1 by 1 MPI processor grid
  read_data CPU = 0.000 seconds
Reading data file ...
ERROR: Read_data subsequent file cannot switch to restricted triclinic (src/read_data.cpp:639)
Last command: read_data few-tric.data add append
```
This uses the input:
```
read_data empty-orth.data
read_data few-tric.data add append
```
with the data files:
```
Empty Orth Data file

0 atoms

1 atom types

-1.0 1.0 xlo xhi
-1.0 1.0 ylo yhi
-1.0 1.0 zlo zhi

```
and
```
Triclinic Data file

1 atom types
3 atoms

-1.0 1.0 xlo xhi
-1.0 1.0 ylo yhi
-1.0 1.0 zlo zhi
 0.3 0.2 0.1 xy xz yz

Masses

 1 1.0


Atoms # atomic

1 1 0.5 0.0 0.0
2 1 0.0 0.0 0.0
3 1 0.0 0.0 0.5
```

So it looks like the new code disallows an input that used to be accepted.

sjplimp:
@akohlmey yes, that is intentionally now an error - didn't know it was allowed before for this special case of empty orthog box
followed by a triclinic box.

A new note on the read_data doc page says this in the Reading multiple data files section:

If the first read_data command defined a restricted or general
   triclinic simulation box (see the sub-section below on header
   keywords), then subsequent data files have restrictions.  For a
   restricted triclinic box, the 3 tilt factors ("xy xz yz" keyword)
   must have the same values in subsequent data files.  For a general
   triclinic box, the avec, bvec, cvec, and "abc origin" keywords must
   have the same values in subsequent data files.  Also the *shift*
   keyword cannot be used in subsequent read_data commands.

I will enhance it to say orthogonal -> triclinic is also not allowed (code checks for this).
I think it is fine to allow orthog box sizes to be changed by subsequent data files
since it is conceptually simple.

I don't think it's a good idea to do that for restricted triclinic (i.e. tilt factors) or the
new general triclinic.  Two reasons, it's conceptually confusing and easy to screw up.
And the change box command already enables this more explicitly, with options like remap.

E.g. adding this line between the 2 read_data commands in the test script enables it to work:
change_box all triclinic xy final 0.3 xz final 0.2 yz final 0.1

sjplimp:
Modulo this last test, which as stated above I think should either be amended with a change_box command or just dropped,
does this mean the current PR now causes no issues with any test we have ?

akohlmey:
> does this mean the current PR now causes no issues with any test we have ?

No, it does not. There are several inputs in the examples tree that fail. They are 2d systems that fail with:
```
lmp -in in.voronoi.2d 
LAMMPS (21 Nov 2023)
OMP_NUM_THREADS environment is not set. Defaulting to 1 thread. (src/comm.cpp:98)
Lattice spacing in x,y,z = 1 1.7320508 1
ERROR: Create_box region zlo/zhi for 2d simulation must straddle 0.0 (src/create_box.cpp:100)
Last command: create_box      1 box
```

akohlmey:
@sjplimp I've updated example inputs, logs, and run and regression test inputs and references.

All tests we run at Temple and through GitHub pass. We need to document that 2d systems may not use 0.0 as box boundary anymore.

sjplimp:
@akohlmey thanks for the help

>> We need to document that 2d systems may not use 0.0 as box boundary anymore.

That should already be in the new doc pages.
I think previously the user could create a 2d system with zlo/zhi from 5 to10 and the z coords of atoms = 8.0.
I no longer think that is a good idea.  I think we should enforce the z coords to always be zero (e.g. fix rigid for 2d)
and thus the box to straddle zero (asymmetric -1 to 10 would be OK).  The new PR enforces created and read_data
atoms for 2d to be z = 0.0.

But if zlo (or zhi) = exactly zero, that doesn't allow for round-off errors.  Not sure we have caught all of those yet.
Hence the new error checks and messages for straddling zero.

akohlmey:
@sjplimp I was primarily thinking about the release notes. Under "Backward Compatibility" is still says:

All the code changes should be new optional features and not break existing inputs.

This is clearly not the case.

sjplimp:
@akohlmey Just edited the Backward Compatible section in the PR header.

stanmoore1:
@sjplimp I think `examples/PACKAGES/phonon/2-1D-diatomic/in.Ana` needs to be updated to straddle 0.0 in the z-direction, otherwise internal CPU regression tests passed.

sjplimp:
@akohlmey For this phonon example, I tried to build with the PHONON package via traditional make.
It built, but gave error at run time that fix phonon was missing due to dependency?  Is that supposed to
be a build-time warning or error that gives info as to the dependency?

akohlmey:
> It built, but gave error at run time that fix phonon was missing due to dependency?

Yes, because fix phonon also needs the KSPACE package to be installed. The other parts of the PHONON package don't, so - consistent with other packages - the parts of the package that depend on other packages are skipped.
The documentation is a bit vague about this, but PR #4004 contains some clarifications.

LAMMPS cannot know which package (or style) is missing since there can be indirect dependencies, too. It only knows that the PHONON package is installed and fix phonon is part of it, but missing. So the logical conclusion is that it was skipped because of a missing prerequisite (the FFT wrappers from KSPACE in this case).

stanmoore1:
@sjplimp one failing regression test:
```
The following tests FAILED:
	270 - AtomicPairStyle:eam_alloy_real (Failed)
```

akohlmey:
> @sjplimp one failing regression test:
> 
> ```
> The following tests FAILED:
> 	270 - AtomicPairStyle:eam_alloy_real (Failed)
> ```

That one is bogus. That is due to a race condition in OpenMPI on Ubuntu 20.04LTS that I have not figured out how to avoid. Right now, there occasionally leads to a test failure, but I can usually just retrigger the failing test and it will pass.

sjplimp:
>> LAMMPS cannot know which package (or style) is missing since there can be indirect dependencies, too.

Could a new make option list these?  Or possibly added info from make ps or make pi ?
Installed YES: package PHONON
  src/fix_phonon.cpp does not exist, depends on KSPACE

akohlmey:
> Could a new make option list these? Or possibly added info from make ps or make pi ?

I don't see how.

The information is in the manual (now that #4004 is merged it is more obvious).

```
This fix is part of the PHONON package.  It is only enabled if LAMMPS
was built with that package.  This fix also requires LAMMPS to be built
with 3d-FFT support which is included in the KSPACE package.  See the
:doc:`Build package <Build_package>` page for more info.
```

stanmoore1:
@sjplimp @akohlmey Is this ready to merge??

akohlmey:
@sjplimp FYI, I've updated the branch to the latest upstream code, fixed the merge conflict and also made a few corrections flagged by compiler warnings:
- in create_atoms there was an assignment where a comparison used to be
- you added new polymorph methods, but only flagged them as virtual in the base class, yet forgot the "override" keyword in the derived classes. I've corrected those that I've found. This is documented in our coding style guidelines: https://docs.lammps.org/Modify_style.html#miscellaneous-standards-varied

sjplimp:
@akohlmey I am starting to work on this PR again.  I pulled and merged with current master.  I don't understand why in my list of changed files for this PR, various files from 3 sub-dirs in the examples dir are showing up?

akohlmey:
> @akohlmey I am starting to work on this PR again. I pulled and merged with current master. I don't understand why in my list of changed files for this PR, various files from 3 sub-dirs in the examples dir are showing up?

Those examples had to be changed since they would produce errors with your changes in this PR that do no longer allow a 2d system to have one of the box boundaries be 0.0.

sjplimp:
@akohlmey - this is now ready to be reviewed for merging

I think all the reviewers on the list should take a look before merging

For @akohlmey, there are 2 NOTE comments with Axel in create_atoms.cpp where I think the definition and scaling of readthresh and mesh_density need adjustments.

With a clean traditional make I get this odd warning which seems ominous - don't know what is causing it:

mpicxx -g -O3 -std=c++11  -DLAMMPS_GZIP -DLAMMPS_MEMALIGN=64    -DMPICH_SKIP_MPICXX -DOMPI_SKIP_MPICXX=1    -c ../atom_vec_hybrid.cpp
In file included from /usr/include/string.h:535,
                 from /usr/include/c++/11/cstring:42,
                 from ../fmt/format.h:38,
                 from ../pointers.h:33,
                 from ../atom_vec.h:17,
                 from ../atom_vec.cpp:14:
In function ‘void* memcpy(void*, const void*, size_t)’,
    inlined from ‘virtual void LAMMPS_NS::AtomVec::write_data_restricted_to_general()’ at ../atom_vec.cpp:2272:21:
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:29:33: warning: ‘void* __builtin_memcpy(void*, const void*, long unsigned int)’ specified bound between 18446744056529682432 and 18446744073709551592 exceeds maximum object size 9223372036854775807 [-Wstringop-overflow=]
   29 |   return __builtin___memcpy_chk (__dest, __src, __len,
      |          ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
   30 |                                  __glibc_objsize0 (__dest));
      |                                  ~~~~~~~~~~~~~~~~~~~~~~~~~~

Similar warning if I build with the machdyn package installed.

sjplimp:
@ndtrung81 When I was going thru the package per-atom vectors that should have added support for general triclinic boxes, the normx, normy, normz vector for the DIELECTRIC package was on the list.  But looking at the code it appears to use mu internally,
i.e. nothing called norm in atom.cpp/h.  So my Q is why the read_data.rst doc page names it norm xyz instead of just mu?  They both appear to be dipole moments?  It would simplify the doc page if we just called it mu ?

sjplimp:
@akohlmey Possibly the input scripts in examples/triclinic are too simple, though I think they will be useful for users to examine.  Possibly these could be unit tests for this new functionality, and we should add (or replace) them in examples/triclinic with scripts which actually perform dynamics on bigger systems.  If you agree, I suggest we postpone that until a later PR.

ndtrung81:
@sjplimp right, norm[xyz] is essentially the direction of the induced dipole moment at each boundary elements (interface particles in the example scripts) caused by surface polarization effects. Thanks for pointing this out. Please help me rename them in the read_data doc page. 

In the atom_style page, I mentioned normxyz in the phrase "..each particle also should be assigned a normal unit vector (defined by normx, normy, normz),..", as illustrated in the example data files. Does it make sense to leave it as is, or should I mention the connection between these normxyz to mu?

sjplimp:
Thanks @ndtrung81 - I made small edits to the read_data and atom_style doc pages, removing references to norm xyz.  See if it reads OK?

ndtrung81:
@sjplimp it is good for me. Thanks!

akohlmey:
> With a clean traditional make I get this odd warning which seems ominous - don't know what is causing it:

This looks like it is bug in gcc where it miscomputes pointer offsets. Sadly, turning off this one type of warning triggers a different one and turning that off triggers yet another one. At that point, i gave up digging deeper. It looks gcc gets confused handling pointers where it internally applies the "restrict" property, i.e. they must not overlap.

It doesn't print a warning with clang, even with maximum warnings enabled, and clang is usually better at spotting and flagging broken code and prints more warnings. So I would consider it a false positive.

akohlmey:
> For @akohlmey, there are 2 NOTE comments with Axel in create_atoms.cpp where I think the definition and scaling of readthresh and mesh_density need adjustments.

@sjplimp I've changed the value for consistency, but it has no impact. There is no code path where this variable is used *without* it being set from the LAMMPS input as a keyword argument. So the initial value is simply set to silence compiler warnings.

akohlmey (approved):

akohlmey:
@sjplimp did notify Alex Stukowski about the changes in the LAMMPS native dump file format so that he or his staff can adapt OVITO as needed?

You won't need to notify the VMD developers to update the LAMMPS reader :wink:

doc/src/Howto_body.rst
```diff
@@ -154,12 +154,22 @@ values consistent with the current orientation of the rigid body
...
+
 The coordinates of each sub-particle are specified as its x,y,z
 displacement from the center-of-mass of the body particle.  The
 center-of-mass position of the particle is specified by the x,y,z
 values in the *Atoms* section of the data file, as is the total mass
 of the body particle.
 
+If the data file defines a general triclinic box, then the orientation
+of the body particle and its corresponding 6 moments of inertia and
+sub-particle displacements should reflect the fact the body is defined
```
ndtrung81: "..the fact that the body.."

doc/src/Howto_body.rst
```diff
@@ -154,12 +154,22 @@ values consistent with the current orientation of the rigid body
...
+withing a general triclinic box with edge vectors **A**,**B**,**C**.
```
ndtrung81: typo "withing" --> "within"

doc/src/Howto_body.rst
```diff
@@ -270,6 +281,15 @@ A disk, whose diameter is 3.0, mass 1.0, is specified as follows:
    0 0 0
    3.0
 
+If the data file defines a general triclinic box, then the orientation
```
ndtrung81: This section (lines 284-292) duplicates the above (lines 164-172 for nparticle), same as lines 459-467 for body rounded/polyhedron. Three duplicated paragraphs are not specific to individual body styles.

ndtrung81 (approved):
@sjplimp I found a couple of typos and have a comment on duplicated text in the doc page Howto_body.rst.

sjplimp:
Thanks @ndtrung81 - I made the changes.  I kept the duplicated paragraph in one place (before the 3 body styles) and just left one sentence in each style to clarify which values were orientation dependent.

athomps (approved):
I approve

stanmoore1 (approved):
I approve

PR #3997, opened 2023-11-27
adding the scale feature of quip calculator, allows fix adapt functio…

ConnorSA:
…nallity

**Summary**

ML-QUIP calculators now support the scale feature with fix adapt. This is useful for thermodynamic integration and reversible scaling.

**Related Issue(s)**

**Author(s)**

Connor Allen
University of Warwick, UK

**Licensing**

**Backward Compatibility**

Not expected

**Implementation Notes**

Verified by running MD and monitoring the outputs.

**Post Submission Checklist**


akohlmey:
@ConnorSA thanks for your submission. However, your implementation allocates a 2d scale factor arrays by only uses the [1][1] value (and in a hardcoded fashion). In that case it would be better and cleaner to implement this as a global scale factor and make it a scalar instead of a 2d array.

ConnorSA:
Hi @akohlmey. Thank you for the feedback, I agree it doesn't look nice seeing the hard coded array indexing, however this was out of necessity. I am trying to get `scale` from the extract method. I have tried (and failed) to get the extract method to simply pull a scalar value (c++ is not my forte), so opted for this as it at least gave me the functionality I was after and though others would find useful. I couldn't find any examples in the source of the extract method being using to get something of `dim=0` (all implementations I was following, use `scale` as a 2D array). 

a key difference as far as I can tell from QUIP and the other potentials is that QUIP calculates does some of the potential energy surface (PES) calculations globally, so I don't see how to follow similar conventions of using different atom types when accessing `scale` to calculate the PES (hence the hardcoded [1][1]). 

Any further advice is appreciated!

Many thanks,

Connor

P.S noticed a bug in my commit, hence the new commit, despite no real changes to the structure.

akohlmey:
Please see pair style kim for an example that uses a global scalar scale factor (ignore the "scale_extracted" property). The kim pair style is rather special, but it is somewhat similar to pair style quip in that it also is an interface to a library with multiple different potentials.

akohlmey:
> I agree it doesn't look nice seeing the hard coded array indexing, however this was out of necessity.

Rather than implementing such an ugly hack, you could have asked a question about how to extract a scalar in the "LAMMPS Development" category in the MatSci.org forum:  https://matsci.org/c/lammps/lammps-development/42
That is what the forum is meant for. :wink:

akohlmey:
Please also don't forget to add the quip pair style with the scale global property to the list of supported pair styles to the fix adapt documentation.

ConnorSA:
Thank you for signposting me to the KIM potential, this is helped a lot! I've updated the pair_quip.cpp and .h to reflect the desired implementation!

Also, I have updated the docs, I hope I have done this correctly!  

Thanks again!

Connor

src/ML-QUIP/pair_quip.cpp
```diff
@@ -314,8 +318,16 @@ void PairQUIP::init_style()
 /* ----------------------------------------------------------------------
    init for one type pair i,j and corresponding j,i
 ------------------------------------------------------------------------- */
+double PairQUIP::init_one(int i, int j)
```
akohlmey: The init_one function can be restored to its original since the arguments are ignored.

src/ML-QUIP/pair_quip.cpp
```diff
@@ -65,7 +66,7 @@ PairQUIP::~PairQUIP()
 
 void PairQUIP::compute(int eflag, int vflag)
 {
-  int inum, jnum, sum_num_neigh, ii, jj, i, iquip;
+  int inum, jnum, sum_num_neigh, ii, jj, i, iquip, itype;
```
akohlmey: itype doesn't seem to be used anywhere.

akohlmey (changes requested):
Some cosmetic changes should be applied. See my comments in the source.

akohlmey:
... and please don't use tabs. https://ci.lammps.org/job/dev/job/pull_requests/job/checkstyle/8877/whitespace/

You can address whitespace issues from the src folder by typing: make fix-whitespace
and commit the changes. We like to have a consistent programming style...

ConnorSA:
I have implemented the above changes :)

stanmoore1 (approved):

akohlmey (approved):

PR #3998, opened 2023-11-28
Adding copy of label maps while using replicate command

Bibobu:
**Summary**

The current use of the replicate command does not copy already defined label maps when copying atom classes. This modification should correct the issue.

**Author(s)**

Germain Clavier (Unicaen, germain.clavier@unicaen.fr)

**Licensing**

**Backward Compatibility**

This should not cause backward compatibility issues.

**Implementation Notes**

This is an edit in the replicate.cpp file checking labelmapflag and copying labelmap accordingly.

**Post Submission Checklist**


src/replicate.cpp
```diff
@@ -228,6 +229,11 @@ void Replicate::command(int narg, char **arg)
   atom->extra_improper_per_atom = old->extra_improper_per_atom;
   atom->maxspecial = old->maxspecial;
 
+  if (old->labelmapflag) {
+    atom->add_label_map();
+    for (int i = 0; i<5; i++)  atom->lmap->merge_lmap(old->lmap, i);
```
stanmoore1: It's not wrong, but it seems a little strange to me to hard code a `5` here, rather than referring to a variable in the label_map class.

stanmoore1 (approved):

src/replicate.cpp
Bibobu: I agree that the hard coding of an integer value is clearly not elegant. As such I referenced the enum in the `Atom` class which is used to loop over interaction types. I don't think the label_map class has a public equivalent over which we could iterate.

sjplimp (approved):
I suggest we wait to merge until Jake looks at this.  Perhaps he will have a suggestion on Stan's comment.

src/replicate.cpp
jrgissing: @Bibobu thanks for this addition. looks good, but you could just write out each item in the list explicitly to avoid the technically dangerous iteration over enum. it would add several lines, but is probably most in line with how we designed the type labels framework. 

@sjplimp if this type of operation shows up again, we could add a better 'merge_all_lmaps' function

sjplimp:
I like Jake's suggestion.  There are only 5 possible maps I think, so just list them explicitly.

akohlmey:
> I like Jake's suggestion. There are only 5 possible maps I think, so just list them explicitly.

Made the change

PR #4003, opened 2023-12-01
Thread over neighbors in addition to atoms when using a half neighlist

stanmoore1:
**Summary**

Previously the Kokkos package `neigh/thread` option only supported a full list. This PR extends the option to include a half list as well. Also tune team threading params, and change default for `neigh/thread` to only be on if using GPUs.

**Related Issue(s)**

None

**Author(s)**

Stan Moore

**Licensing**

**Backward Compatibility**

yes


stanmoore1:
Thanks to @weinbe2 for pointing out an inefficiency in the threading code (now fixed).

stanmoore1:
I confirmed correctness on GPUs with half list and full list. The tuning gives a marginal speedup when using a larger cutoff for LJ 500 atoms (5.0 sigma instead of 2.5 sigma).

sjplimp (approved):

PR #4007, opened 2023-12-06
KSPACE: decouple KOKKOS and non-KOKKOS FFT

### LINKED ISSUE #3775: [BUG] Unable to enable hipFFT-backend with KSPACE + Kokkos with cmake
**Summary**

We noticed that our build was not utilizing the hipFFT backend for KSPACE.

```
-- Kokkos Architecture: VEGA90A
-- <<< FFT settings >>>
-- Primary FFT lib:  KISS
-- Using double precision FFTs
```

Develop from ~ a month ago: 56b5291529f197dbabec397dbaa66939be6f562a
MI-200 system w/ ROCm 5.4.3

**Expected Behavior**

We want to be able to enable the hipFFT backend for KSPACE.

**Actual Behavior**

The issue (seems) to stem from a lack of the `KSPACE`'s package awareness of `Kokkos`, and vice-versa.
Namely:

https://github.com/lammps/lammps/blob/fceff845c7ae434ce8bfe1ca6adf1313e67af2dd/cmake/Modules/Packages/KSPACE.cmake#L13

Sets the available choices for the "FFT" variable to be FFTW, KISS, and MKL.

Meanwhile:

https://github.com/lammps/lammps/blob/fceff845c7ae434ce8bfe1ca6adf1313e67af2dd/cmake/Modules/Packages/KSPACE.cmake#L11

sets the default FFT backend to "KISS" in the case that FFTW is not found, while

https://github.com/lammps/lammps/blob/fceff845c7ae434ce8bfe1ca6adf1313e67af2dd/cmake/Modules/Packages/KSPACE.cmake#L15

enforces the FFT selection to be in the FFTW/KISS/MKL list.

Finally:

https://github.com/lammps/lammps/blob/fceff845c7ae434ce8bfe1ca6adf1313e67af2dd/cmake/Modules/Packages/KOKKOS.cmake#LL139C13-L139C16

Selects `hipFFT` IFF the FFT choice is not KISS.  It appears to me that the same problem would prevent selection of the cuFFT backend as well, but I haven't dug into it as far.

**Steps to Reproduce**

from within lammps/build:

```
cmake -DPKG_KOKKOS=on    \
      -DPKG_KSPACE=on   \
      -DKokkos_ENABLE_HIP=on \
      -DKokkos_ENABLE_SERIAL=on \
      -DCMAKE_CXX_COMPILER=$(which hipcc) \
      -DKokkos_ARCH_VEGA90A=ON  ../lammps/cmake

...
-- <<< FFT settings >>>
-- Primary FFT lib:  KISS
-- Using double precision FFTs
-- Using non-threaded FFTs
-- Kokkos FFT: KISS
```

Or

```
cmake -DPKG_KOKKOS=on    \
      -DPKG_KSPACE=on   \
      -DKokkos_ENABLE_HIP=on \
      -DKokkos_ENABLE_SERIAL=on \
      -DCMAKE_CXX_COMPILER=$(which hipcc) \
      -DFFT=HIPFFT \
      -DKokkos_ARCH_VEGA90A=ON  ../lammps/cmake
...

CMake Error at Modules/LAMMPSUtils.cmake:18 (message):
  

  ########################################################################

  Invalid value 'HIP' for option FFT

  

  Possible values are:

  * KISS

  * FFTW3

  * MKL

  ########################################################################
Call Stack (most recent call first):
  Modules/Packages/KSPACE.cmake:15 (validate_option)
  CMakeLists.txt:525 (include)

```

I think this is a bit of a design flaw in the current cmake.  Basically, KSpace is included here: https://github.com/lammps/lammps/blob/fceff845c7ae434ce8bfe1ca6adf1313e67af2dd/cmake/CMakeLists.txt#L525

While, Kokkos gets included about a 100 lines later:

https://github.com/lammps/lammps/blob/fceff845c7ae434ce8bfe1ca6adf1313e67af2dd/cmake/CMakeLists.txt#L674

This means that there's no way for us to find whether hipfft/cufft are valid targets (on older ROCm installs at least, this was a problem), which is done by Kokkos:

https://github.com/lammps/lammps/blob/fceff845c7ae434ce8bfe1ca6adf1313e67af2dd/cmake/Modules/Packages/KOKKOS.cmake#L139

The easiest solution I can see is before including KSpace, we check if Kokkos is enabled.
If so, we set (e.g.,) a KOKKOS_FFT_VALUES variable (which defaults to "") to include cuFFT / hipFFT.
Then the KSpace validation check checks whether the FFT value provided is in either KOKKOS_FFT_VALUES _or_ FFT_VALUES.
Finally, in the Kokkos package we validate that the selected FFT value works for the backend selected (i.e., can't select cuFFT for HIP, and vice versa, and that hipFFT exists if selected).

Or, it's entirely possible I've missed something obvious here!

hagertnl:
**Summary**

This PR decouples the FFT implementation in KOKKOS from non-KOKKOS styles. The non-KOKKOS FFTs can be defined with `-DFFT_KISSFFT`, as normal, and the KOKKOS FFT can be selected by `-DFFT_KOKKOS_KISSFFT`. Then, unique FFT implementations may be used depending on whether KOKKOS is used or not.

All FFT settings also apply to this except FFT_PRECISION which is still unified between vanilla and Kokkos.

The KOKKOS FFT settings currently do not inherit from the non-KOKKOS FFT settings.

**Related Issue(s)**

Fixes #3775 

**Author(s)**

Nick Hagerty, Oak Ridge National Laboratory, nhagerty1@gmail.com

**Licensing**

**Backward Compatibility**

This PR will change the behavior of the `-DFFT_<x>` compiler flags. `-DFFT_<x>` itself will no longer select the FFT to use for Kokkos pair styles. We may want to consider having the KOKKOS FFT inherit from the standard `-DFFT_<x>` flag if no `-DFFT_KOKKOS_<x>` flag is provided. But allowing the KOKKOS FFT to default to KISSFFT when not specified would be the safest option.

**Implementation Notes**

Verified correctness using an SPC/e water box of 4 million water molecules on ORNL's Frontier using two runs -- one utilized 8x MI25X with hipFFT, one was CPU-only with the default KISSFFT. Tested using the Make-based build system to specify the exact compiler flags desired.

**Post Submission Checklist**


stanmoore1:
>All FFT settings also apply to this -- ie, FFT_PRECISION was changed to FFT_KOKKOS_PRECISION, and can be different from the non-KOKKOS FFT_PRECISION.

@hagertnl do you think it is useful to have Kokkos FFT precision different than non-Kokkos? I'm inclined to keep it simple and require these to be the same.

hagertnl:
> @hagertnl do you think it is useful to have Kokkos FFT precision different than non-Kokkos? I'm inclined to keep it simple and require these to be the same.

@stanmoore1 I can't say I have a strong preference. I decoupled everything to start with. Thinking about it, I would probably be in favor of using the same precision for Kokkos and non-Kokkos the vast majority of the time. If you use different, may be confusing to a less-experienced user.

stanmoore1:
@hagertnl I'm happy with the code now. I believe this would work for GNU Makefile but we still need to update CMake and the documentation.

hagertnl:
@stanmoore1 I can work on drafting for the CMake build -- do you want the FFT_KOKKOS variable to default to the same setting as FFT, or should it default to KISSFFT? I would also think sanity checks like
`
if(Kokkos_ENABLE_CUDA and (FFT_KOKKOS STREQUAL "HIPFFT"))
`
would probably be good. I think the only file that should require modification is the `KOKKOS.cmake` file.

hagertnl:
I think initially, I'm in favor of changing this logic
https://github.com/lammps/lammps/blob/00c3f14a6362405d94a6e7f1f1d3a4f5dcd048d8/cmake/Modules/Packages/KOKKOS.cmake#L134-L147

To something more like:
```
if (Kokkos_ENABLE_CUDA and NOT ((FFT_KOKKOS STREQUAL "KISS") or (FFT_KOKKOS STREQUAL "CUFFT"))
  message(FATAL_ERROR "The CUDA backend of Kokkos requires either KISS FFT or CUFFT.")
elseif(Kokkos_ENABLE_HIP and NOT ((FFT_KOKKOS STREQUAL "KISS") or (FFT_KOKKOS STREQUAL "HIPFFT")))
  message(FATAL_ERROR "The CUDA backend of Kokkos requires either KISS FFT or CUFFT.")
endif
```
There's a few benefits and side effects of both ways, so I am good with whatever you want, but the main benefit I see here is easily explained behavior -- if you don't set FFT_KOKKOS, you're going to get KISS no matter what backend you have. And, if you set an incompatible option, you're going to get an error for it, instead of the FFT_KOKKOS just being switched to something compatible.

stanmoore1:
@hagertnl thanks, that would be great. I think it should default to KISS FFT if it isn't set, just like the non-Kokkos option. Perhaps a warning would be good though, that performance may be sub-optimal. The error check snippet looks good too, we definitely want to prevent incompatible choices.

hagertnl:
@stanmoore1 I drafted up what we described. I ran a build on Frontier that uses KISS on CPU and HIPFFT on the GPU and ran both CPU & GPU test cases. Both passed and gave what appear to be correct answers. I'm not aware of any edge cases I may have missed. I am OoO tomorrow, so I can work on the docs on Monday. Thanks!

stanmoore1:
@hagertnl thanks! I will take a look

hagertnl:
Explaining the changes I've made to the docs here:
Build_setting.rst:
- Removing the following line, as it's not true -CMake does a TOUPPER call on FFT values in both KSPACE.cmake and KOKKOS.cmake: https://github.com/lammps/lammps/blob/df7f3b8dea934a62399937b459c2d271fbfd25b5/doc/src/Build_settings.rst?plain=1#L66
- Added FFT_KOKKOS doc to the FFT library section of the previously linked file. Added short snippet about CUFFT and HIPFFT to the paragraphs below. **Edge case to highlight: does HIPFFT/CUFFT work with OpenMP tgt offload backend? I haven't tried yet, this is more or less reminding myself to try.**
- I also used shorthand notation of `{option1,option2,option3}` when describing the `FFT_KOKKOS_` flag. I can change that, but it gets quite verbose and there's an example right next to the comment, so hopefully it's apparent how it works.

Howto_cmake.rst:
- Added FFT_KOKKOS as a common option. I think this will be a flag that a large portion of HPC users will need to be aware of. It's relevant to both CPU & GPU users.

kspace_style.rst:
- provided link to optional build settings page for information on selecting a 3rd-party FFT library.

stanmoore1:
@hagertnl this is looking good. One complication is that support for Heffte (https://github.com/lammps/lammps/pull/3963) was recently added to LAMMPS, which caused a merge conflict in `cmake/CMakeLists.txt`. Since Heffte is totally independent of the FFT library in LAMMPS, I think it should be fairly straightforward to fix the conflict?

hagertnl:
@stanmoore1 Ah. Interesting. I did see we'd have a merge conflict pop up.

I think the Heffte conditionals should all be separate from what we've changed. It looks like everything we've done just needs to be moved into an `else` branch in the `if (LMP_HEFFTE)` and indented, if I'm understanding correctly that Heffte is mutually exclusive to all 3rd party FFT options. The GitHub UI definitely doesn't show it well. I may manually pull in the modifications of CMakeLists.txt from Heffte to my local copy and see if that resolves the merge conflict.

Other than that, I think I've got the docs updated and CMake files up-to-date. Only other thing I might want to do is modify the example Makefiles provided for Summit & Frontier in src/MAKE/MACHINES.

hagertnl:
I think I have it simplified now with that latest commit -- now we can just select what we have in this PR as what to keep. I removed the conditionals that printed KISS or HIPFFT, since the FFT_KOKKOS variable holds that value anyway, so we don't need those conditionals.

stanmoore1:
@hagertnl great thanks, I'll test this out. We are getting an issue with some of the tests but I think they may need to be updated to match the changes in this PR.

```
-- Kokkos Devices: OPENMP;SERIAL, Kokkos Backends: OPENMP;SERIAL
CMake Error at Modules/Packages/KOKKOS.cmake:135 (set_property):
  set_property could not find CACHE variable FFT_KOKKOS.  Perhaps it has not
  yet been created.
Call Stack (most recent call first):
  CMakeLists.txt:661 (include)


CMake Error at Modules/LAMMPSUtils.cmake:12 (string):
  string no output variable specified
Call Stack (most recent call first):
  Modules/Packages/KOKKOS.cmake:136 (validate_option)
  CMakeLists.txt:661 (include)


CMake Error at Modules/Packages/KOKKOS.cmake:137 (string):
  string no output variable specified
Call Stack (most recent call first):
  CMakeLists.txt:661 (include)
```

hagertnl:
Hah -- no, I forgot to declare the FFT_KOKKOS variable anywhere. I'll push a commit momentarily with the fix.

hagertnl:
@stanmoore1 anything more you need from me for this PR? I'm not aware of anything from my end at the moment -- tested locally and things seem to work as expected.

stanmoore1:
@hagertnl sorry for the delay. I just need to test this and then it should be good to merge.

alphataubio:
im getting "CUDA driver is a stub library " but this might be a local problem with my build environment, LMOD modules are extra flimsy here on the clusters. 

> module load StdEnv/2020 nvhpc/23.7 openmpi/4.1.4

> CONFIG=issue3775_fft_kokkos-pascal60-cufft; mkdir $CONFIG; cd $CONFIG;
    cmake ../cmake -D LAMMPS_MACHINE=$CONFIG \
    -D PKG_KOKKOS=yes -D Kokkos_ENABLE_CUDA=yes -D FFT_KOKKOS=CUFFT \
    -D Kokkos_ARCH_HOSTARCH=NATIVE -D Kokkos_ARCH_GPUARCH=PASCAL60 \
    -D Kokkos_ENABLE_OPENMP=yes -D BUILD_OMP=yes \
    -D PKG_KSPACE=yes -D PKG_MOLECULE=yes -D PKG_RIGID=yes |tee $CONFIG;
    cmake --build . -j 24 |tee -a $CONFIG

[issue3775_fft_kokkos-pascal60-cufft.txt](https://github.com/lammps/lammps/files/13760042/issue3775_fft_kokkos-pascal60-cufft.txt)




> $ mpirun -np 1 ../lammps-fork/$CONFIG/lmp_$CONFIG -k on g 1 -sf kk -var x 2 -var y 2 -var z 2 -in in.rhodo.scaled -l log.$CONFIG.rhodo.scaled.1_1
LAMMPS (21 Nov 2023 - Development - dd1ac640ae)
KOKKOS mode with Kokkos version 4.2.0 is enabled (src/KOKKOS/kokkos.cpp:72)
  will use up to 1 GPU(s) per node
WARNING: When using a single thread, the Kokkos Serial backend (i.e. Makefile.kokkos_mpi_only) gives better performance than the OpenMP backend (src/KOKKOS/kokkos.cpp:202)
Exception: (CudaInternal::singleton().cuda_get_device_count_wrapper<false>( &m_cudaDevCount)) error( cudaErrorStubLibrary): CUDA driver is a stub library /home/~~~/scratch/pull4007-test/lammps-fork/lib/kokkos/core/src/Cuda/Kokkos_Cuda_Instance.cpp:278

>$ echo $LIBRARY_PATH |tr ':' '\n'|xargs ls -1
[ld_library_path.txt](https://github.com/lammps/lammps/files/13796711/ld_library_path.txt)

>$ nvidia-smi 

```
Sat Dec 23 14:26:40 2023       
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 535.129.03             Driver Version: 535.129.03   CUDA Version: 12.2     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  Tesla P100-PCIE-12GB           Off | 00000000:04:00.0 Off |                    0 |
| N/A   36C    P0              26W / 250W |      0MiB / 12288MiB |      0%      Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+
                                                                                        
> +---------------------------------------------------------------------------------------+
| Processes:                                                                            |
|  GPU   GI   CI        PID   Type   Process name                            GPU Memory |
|        ID   ID                                                             Usage      |
|=======================================================================================|
|  No running processes found                                                           |
+---------------------------------------------------------------------------------------+
```

alphataubio:
>$ nvidia-smi 

```
Sat Dec 23 14:26:40 2023       
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 535.129.03             Driver Version: 535.129.03   CUDA Version: 12.2     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  Tesla P100-PCIE-12GB           Off | 00000000:04:00.0 Off |                    0 |
| N/A   36C    P0              26W / 250W |      0MiB / 12288MiB |      0%      Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+
                                                                                        
> +---------------------------------------------------------------------------------------+
| Processes:                                                                            |
|  GPU   GI   CI        PID   Type   Process name                            GPU Memory |
|        ID   ID                                                             Usage      |
|=======================================================================================|
|  No running processes found                                                           |
+---------------------------------------------------------------------------------------+
```

alphataubio:
i've been trying many different combinations, most of them failed. my build command is of the form:

> CONFIG=develop-power9-volta70-kokkos-kiss; mkdir build-$CONFIG; cd build-$CONFIG;
    cmake ../cmake -D BUILD_SHARED_LIBS=yes -D LAMMPS_MACHINE=$CONFIG \
    -D PKG_KOKKOS=yes -D Kokkos_ENABLE_CUDA=yes \
    -D Kokkos_ARCH_HOSTARCH=POWER9 -D Kokkos_ARCH_GPUARCH=VOLTA70 \
    -D Kokkos_ENABLE_OPENMP=yes -D BUILD_OMP=yes \
    -D Kokkos_ENABLE_SERIAL=no \
    -D PKG_EXTRA-DUMP=yes -D PKG_KSPACE=yes -D PKG_COLVARS=yes \
    -D PKG_EXTRA-PAIR=yes -D PKG_MOLECULE=yes -D PKG_RIGID=yes ; \
    cmake --build . -j 48
    
where $CONFIG =
develop
develop-power9-volta70-gpu (my "lab control" of building cuda without kokkos)
develop-power9-volta70-kokkos-kiss
issue3775_fft_kokkos-power9-volta70-kiss
issue3775_fft_kokkos-power9-volta70-cufft

i'm using an oddball power9/v100 cluster because the login node has 4 gpus attached, more convenient than requesting time-limited debug session with slurm scheduler on intel/p100/v100 clusters to a compute node for building with a gpu attached. the failures are mostly my responsibility for having such a crummy build environment, but i've run into some edge cases which i document in the following test cases A, B, ...

### *** **in particular, GPU and KOKKOS packages do not build with latest version of cmake 3.28.1 for different reasons ( see below D and E )**. maybe this is known already *** 

please let me know if you want the CMakeOutput.log or CMakeError.log for any of these test cases.

how can i find out on ci.lammps.org what the exact versions of cmake / gcc / cuda / nvhpc / openmpi / ... used in builds so i can replicate a working combination for myself. i've looked around, for example https://ci.lammps.org/job/dev/job/develop/job/ibmxl_ppc64le/904/execution/node/3/ws/build/ but i cant find specific versions of libraries sorry

my primary test input is the rhodo benchmark:

> lmp_$CONFIG -sf gpu -pk gpu 1 -var x 2 -var y 2 -var z 2 -in in.rhodo.scaled -l log.$CONFIG.rhodo.scaled.1


> mpirun -np 1 lmp_$CONFIG -sf gpu -pk gpu 1 -var x 2 -var y 2 -var z 2 -in in.rhodo.scaled -l log.$CONFIG.rhodo.scaled.1_1


> mpirun -np 4 lmp_$CONFIG -sf gpu -pk gpu 4 -var x 2 -var y 2 -var z 2 -in in.rhodo.scaled -l log.$CONFIG.rhodo.scaled.4_4


>lmp_$CONFIG -k on g 1 -sf kk -pk kokkos neigh half -var x 2 -var y 2 -var z 2 -in in.rhodo.scaled -l log.$CONFIG.rhodo.scaled.1


> mpirun -np 1 lmp_$CONFIG -k on g 1 -sf kk -pk kokkos neigh half -var x 2 -var y 2 -var z 2 -in in.rhodo.scaled -l log.$CONFIG.rhodo.scaled.1_1


> mpirun -np 4 lmp_$CONFIG -k on g 4 -sf kk -pk kokkos neigh half -var x 2 -var y 2 -var z 2 -in in.rhodo.scaled -l log.$CONFIG.rhodo.scaled.4_4


My other test input is a protein-dna simulation (1HVN), the smallest number of atoms available on PDB with at least one protein, at least one dna, and no modified engineered residues. 1HVN lammps input scripts created by charmm-gui.org (now working thanks to @akohlmey cmap fixes):

[1hvn-lammps.zip](https://github.com/lammps/lammps/files/13796625/1hvn-lammps.zip)

alphataubio:
==== (TEST A) ==== pkg kokkos cufft / cmake 3.21.4 / nvhpc 23.11 ====

```
nvcc-Error-A CUDA toolkit matching the current driver version (11.7) or a supported older version (11.0 or 11.8) was not installed with this HPC SDK
CMake Error at /home/~~~/scratch/lammps/lib/kokkos/cmake/kokkos_compiler_id.cmake:12 (STRING):
  STRING sub-command REPLACE requires at least four arguments.
Call Stack (most recent call first):
  /home/~~~/scratch/lammps/lib/kokkos/cmake/kokkos_compiler_id.cmake:56 (kokkos_internal_have_compiler_nvcc)
  /home/~~~/scratch/lammps/lib/kokkos/cmake/kokkos_tribits.cmake:175 (INCLUDE)
  /home/~~~/scratch/lammps/lib/kokkos/CMakeLists.txt:226 (KOKKOS_SETUP_BUILD_ENVIRONMENT)


-- Using --c++17 for C++17 standard as feature
CMake Error at /home/~~~/scratch/lammps/lib/kokkos/cmake/kokkos_test_cxx_std.cmake:125 (MESSAGE):
  Invalid compiler for CUDA.  To allow nvc++ as Cuda compiler,
  Kokkos_ENABLE_IMPL_NVHPC_AS_DEVICE_COMPILER=ON must be set!
Call Stack (most recent call first):
  /home/~~~/scratch/lammps/lib/kokkos/cmake/kokkos_tribits.cmake:203 (INCLUDE)
  /home/~~~/scratch/lammps/lib/kokkos/CMakeLists.txt:226 (KOKKOS_SETUP_BUILD_ENVIRONMENT)
```

alphataubio:
==== (TEST B) ==== pkg kokkos cufft / cmake 3.21.4 / nvhpc 22.11 / openmp no / serial yes ====

```
/usr/bin/ld: cannot find -lcufft
make[2]: *** [CMakeFiles/lmp.dir/build.make:109: lmp_issue3775_fft_kokkos-power9-volta70-cufft] Error 2
make[1]: *** [CMakeFiles/Makefile2:997: CMakeFiles/lmp.dir/all] Error 2
make: *** [Makefile:136: all] Error 2

/home/~~~/local/nvhpc_22.11/Linux_ppc64le/22.11/math_libs/11.8/targets/ppc64le-linux/lib/libcufft_static.a
/home/~~~/local/nvhpc_22.11/Linux_ppc64le/22.11/math_libs/11.8/targets/ppc64le-linux/lib/libcufft.so
```

SOLUTION: only -D BUILD_SHARED_LIBS=yes works because otherwise cufft static library is called libcufft_static.a not libcufft.a

alphataubio:
==== (TEST C) ==== pkg kokkos cufft / cmake 3.21.4 / nvhpc 22.11 / openmp no / serial yes ====

SUCCESS (almost) !!!
 
### (C.1) serial version freezes in libpthread:

```
gdb --args lmp_$CONFIG -k on g 1 -sf kk -pk kokkos neigh half -var x 2 -var y 2 -var z 2 -in in.rhodo.scaled -l log.$CONFIG.rhodo.scaled.1
^C
Program received signal SIGINT, Interrupt.
0x00007fffec377f44 in read () from /lib64/glibc-hwcaps/power9/libpthread-2.28.so
(gdb) bt
#0  0x00007fffec377f44 in read () from /lib64/glibc-hwcaps/power9/libpthread-2.28.so
#1  0x00007fffd9560f68 in fork_hnp () at ../../../../../orte/mca/ess/singleton/ess_singleton_module.c:695
#2  0x00007fffd955f63c in rte_init () at ../../../../../orte/mca/ess/singleton/ess_singleton_module.c:171
#3  0x00007fffd9621fc0 in orte_init () at ../../orte/runtime/orte_init.c:271
#4  0x00007fffed12ff3c in ompi_mpi_init () at ../../ompi/runtime/ompi_mpi_init.c:518
#5  0x00007fffecedab44 in PMPI_Init () at pinit.c:69
#6  0x000000001000b374 in main (argc=24, argv=0x7fffffffbcf8) at /home/~~~/scratch/lammps-fork/src/main.cpp:47
```

### (C.2) OK with 1 MPI rank, 1 gpu:

> mpirun -np 1 lmp_$CONFIG -k on g 1 -sf kk -pk kokkos neigh half -var x 2 -var y 2 -var z 2 -in in.rhodo.scaled -l log.$CONFIG.rhodo.scaled.1_1

```
------------ Step            100 ----- CPU =      3.75987 (sec) -------------
TotEng   =   -202314.9834 KinEng   =    172724.3971 Temp     =       301.0607
PotEng   =   -375039.3806 E_bond   =     20545.2148 E_angle  =     86257.3238
E_dihed  =     41587.6716 E_impro  =      1733.5573 E_vdwl   =    -15187.7391
E_coul   =   1627614.2832 E_long   =  -2137589.6922 Press    =        10.9237
Volume   =   2465078.7364
Loop time of 3.75996 on 1 procs for 100 steps with 256000 atoms

Performance: 4.596 ns/day, 5.222 hours/ns, 26.596 timesteps/s, 6.809 Matom-step/s
71.8% CPU use with 1 MPI tasks x 1 OpenMP threads

MPI task timing breakdown:
Section |  min time  |  avg time  |  max time  |%varavg| %total
---------------------------------------------------------------
Pair    | 2.0872     | 2.0872     | 2.0872     |   0.0 | 55.51
Bond    | 0.10963    | 0.10963    | 0.10963    |   0.0 |  2.92
Kspace  | 0.51199    | 0.51199    | 0.51199    |   0.0 | 13.62
Neigh   | 0.5077     | 0.5077     | 0.5077     |   0.0 | 13.50
Comm    | 0.054609   | 0.054609   | 0.054609   |   0.0 |  1.45
Output  | 0.0003697  | 0.0003697  | 0.0003697  |   0.0 |  0.01
Modify  | 0.38513    | 0.38513    | 0.38513    |   0.0 | 10.24
Other   |            | 0.1033     |            |       |  2.75

Nlocal:         256000 ave      256000 max      256000 min
Histogram: 1 0 0 0 0 0 0 0 0 0
Nghost:         163372 ave      163372 max      163372 min
Histogram: 1 0 0 0 0 0 0 0 0 0
Neighs:    1.05605e+08 ave 1.05605e+08 max 1.05605e+08 min
Histogram: 1 0 0 0 0 0 0 0 0 0

Total # of neighbors = 1.0560531e+08
Ave neighs/atom = 412.52076
Ave special neighs/atom = 7.431875
Neighbor list builds = 11
Dangerous builds = 0
Total wall time: 0:00:17
```

### (C.3) ucx errors with 4 MPI ranks, 4 GPUs:

> mpirun -np 4 lmp_$CONFIG -k on g 4 -sf kk -pk kokkos neigh half -var x 2 -var y 2 -var z 2 -in in.rhodo.scaled -l log.$CONFIG.rhodo.scaled.4_4

```
[1703838002.785735] [mist-login01:4154733:0]         cma_ep.c:87   UCX  ERROR process_vm_readv(pid=4154731 length=524288) returned -1: Bad address
[1703838002.785735] [mist-login01:4154731:0]         cma_ep.c:87   UCX  ERROR process_vm_readv(pid=4154733 length=524288) returned -1: Bad address
[1703838002.787852] [mist-login01:4154732:0]         cma_ep.c:87   UCX  ERROR process_vm_readv(pid=4154734 length=524288) returned -1: Bad address
[1703838002.787851] [mist-login01:4154734:0]         cma_ep.c:87   UCX  ERROR process_vm_readv(pid=4154732 length=524288) returned -1: Bad address
```

https://github.com/openucx/ucx/issues/6058:
>So we upgraded our Cmake to 3.19.2 to be able to set dynamic linking of CUDA libraries.
See CMake Documentation. The option CMAKE_CUDA_RUNTIME_LIBRARY is supported since Cmake 3.17 while previously our Cmake version is 3.16. By adding set(CMAKE_CUDA_RUNTIME_LIBRARY SHARED) to the CMake configuration file, we are able to dynamically linking the CUDA libraries to our application. Now, our application can run using mpirun with UCX_MEMTYPE_CACHE=n or UCX_MEMTYPE_CACHE=y.


but then after "export UCX_MEMTYPE_CACHE=n" i get:

```
[1703838225.133562] [mist-login01:4155176:0]    cuda_ipc_md.c:234  UCX  ERROR cuIpcGetMemHandle(&key->ph, (CUdeviceptr)addr)() failed: invalid argument
[1703838225.133558] [mist-login01:4155178:0]    cuda_ipc_md.c:234  UCX  ERROR cuIpcGetMemHandle(&key->ph, (CUdeviceptr)addr)() failed: invalid argument
[1703838225.133559] [mist-login01:4155177:0]    cuda_ipc_md.c:234  UCX  ERROR cuIpcGetMemHandle(&key->ph, (CUdeviceptr)addr)() failed: invalid argument
[1703838225.133559] [mist-login01:4155174:0]    cuda_ipc_md.c:234  UCX  ERROR cuIpcGetMemHandle(&key->ph, (CUdeviceptr)addr)() failed: invalid argument
[1703838225.134336] [mist-login01:4155176:0]    cuda_ipc_md.c:234  UCX  ERROR cuIpcGetMemHandle(&key->ph, (CUdeviceptr)addr)() failed: invalid argument
[1703838225.134325] [mist-login01:4155178:0]    cuda_ipc_md.c:234  UCX  ERROR cuIpcGetMemHandle(&key->ph, (CUdeviceptr)addr)() failed: invalid argument
```

alphataubio:
==== (TEST D) ==== pkg kokkos cufft / cmake 3.28.1 / nvhpc 22.11 / openmp no / serial yes ====

so i built from scratch the latest cmake (3.28.1):

CMake Error at /home/~~~/scratch/lammps-fork/lib/kokkos/cmake/build_env_info.cmake:113 (target_compile_features):
  target_compile_features specified unknown feature "" for target
  "impl_git_version".
Call Stack (most recent call first):
  /home/~~~/scratch/lammps-fork/lib/kokkos/CMakeLists.txt:217 (check_git_setup)

alphataubio:
==== (TEST E) ==== pkg gpu / cmake 3.28.1 ====

so i tried going back to basics with develop branch, -D PKG_KOKKOS=no -D PKG_GPU=yes, then i got:

CMake Warning (dev) at Modules/Packages/GPU.cmake:46 (find_package):
  Policy CMP0146 is not set: The FindCUDA module is removed.  Run "cmake
  --help-policy CMP0146" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

Call Stack (most recent call first):
  CMakeLists.txt:661 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.

https://cmake.org/cmake/help/latest/policy/CMP0146.html:
> New in version 3.27. 
The FindCUDA module is removed.

>The FindCUDA module has been deprecated since CMake 3.10. CMake 3.27 and above prefer to not provide the module. This policy provides compatibility for projects that have not been ported away from it.

>Projects using the FindCUDA module should be updated to use CMake's first-class CUDA language support. List CUDA among the languages named in the top-level call to the project() command, or call the enable_language() command with CUDA. Then one can add CUDA (.cu) sources directly to targets, similar to other languages.

alphataubio:
==== (TEST F) ==== cmake/3.21.4 cuda/10.2.2 gcc/8.5.0 openmpi/4.0.5 ====

CMake Error at /home/~~~/scratch/lammps/lib/kokkos/cmake/kokkos_compiler_id.cmake:192 (MESSAGE):
  Compiler not supported by Kokkos.  Required compiler versions:

      Clang(CPU)         8.0.0 or higher
      Clang(CUDA)       10.0.0 or higher
      GCC                8.2.0 or higher
      Intel             19.0.5 or higher
      IntelLLVM(CPU)  2021.1.1 or higher
      IntelLLVM(SYCL) 2023.0.0 or higher
      NVCC              11.0.0 or higher
      HIPCC              5.2.0 or higher
      NVHPC/PGI           22.3 or higher
      MSVC               19.29 or higher
      XL/XLClang           not supported

  Compiler: NVIDIA 10.2.89

Call Stack (most recent call first):
  /home/~~~/scratch/lammps/lib/kokkos/cmake/kokkos_tribits.cmake:175 (INCLUDE)
  /home/~~~/scratch/lammps/lib/kokkos/CMakeLists.txt:226 (KOKKOS_SETUP_BUILD_ENVIRONMENT)

CONCLUSION: going back too far to cuda 10 and openmpi 4.0.5 to get away from ucx is not supported by kokkos

alphataubio:
==== TEST G ==== cmake/3.21.4 cuda/11.2.2 gcc/8.5.0 openmpi/4.1.1+ucx-1.10.1 ====

so now im trying to triangulate this UCX issue:

### develop-power9-volta70-gpu

> lmp_$CONFIG -sf gpu -pk gpu 1 -var x 2 -var y 2 -var z 2 -in in.rhodo.scaled -l log.$CONFIG.rhodo.scaled.1

> mpirun -np 1 lmp_$CONFIG -sf gpu -pk gpu 1 -var x 2 -var y 2 -var z 2 -in in.rhodo.scaled -l log.$CONFIG.rhodo.scaled.1_1

> mpirun -np 4 lmp_$CONFIG -sf gpu -pk gpu 4 -var x 2 -var y 2 -var z 2 -in in.rhodo.scaled -l log.$CONFIG.rhodo.scaled.4_4

ALL OK

### develop-power9-volta70-kokkos-kiss
### issue3775_fft_kokkos-power9-volta70-cufft

> mpirun -np 4 lmp_$CONFIG -k on g 4 -sf kk -pk kokkos neigh half -var x 2 -var y 2 -var z 2 -in in.rhodo.scaled -l log.$CONFIG.rhodo.scaled.4_4

```
[1703896150.031182] [mist-login01:1972523:0]         cma_ep.c:87   UCX  ERROR process_vm_readv(pid=1972525 length=524288) returned -1: Bad address
[1703896150.031182] [mist-login01:1972525:0]         cma_ep.c:87   UCX  ERROR process_vm_readv(pid=1972523 length=524288) returned -1: Bad address
[1703896150.031567] [mist-login01:1972524:0]         cma_ep.c:87   UCX  ERROR process_vm_readv(pid=1972522 length=524288) returned -1: Bad address
[1703896150.031567] [mist-login01:1972522:0]         cma_ep.c:87   UCX  ERROR process_vm_readv(pid=1972524 length=524288) returned -1: Bad address
```

CONCLUSION: this ucx issue is specific to package KOKKOS only, package GPU ok, and not specific to issue3775_fft_kokkos branch because it also happens with develop-power9-volta70-kokkos-kiss. is it to power9 or volta70 ?

maybe this is a problem for ucx to resolve not lammps ??

alphataubio:
==== TEST H ==== cmake/3.21.4 cuda/11.7.1 gcc/11.4.0 ucx/1.15.1 openmpi/5.0.1 ====

to see if "UCX ERROR process_vm_readv" is only specific to openmpi/4.1.1+ucx-1.10.1 up to openmpi/4.1.5+ucx-1.14.1, download latest and build ucx/1.15.1 and openmpi/5.0.1:

https://docs.open-mpi.org/en/v5.0.x/tuning-apps/networking/cuda.html
https://openucx.readthedocs.io/en/master/running.html#ucx-build-and-install

> module purge; module load MistEnv/2021a cmake/3.21.4 cuda/11.7.1 gcc/11.4.0 autotools/2021b
git clone https://github.com/openucx/ucx.git ucx
cd ucx; ./autogen.sh; mkdir build; cd build
../configure --prefix=$HOME/scratch/local --with-cuda=$CUDA_HOME
make -j 64 install

```
In file included from ../../../src/ucm/bistro/bistro_ppc64.c:22:
/home/~~~/scratch/ucx/build/../src/ucm/bistro/bistro_int.h:36:46: error: unknown type name ‘ucm_bistro_lock_t’
   36 | void ucm_bistro_modify_code(void *dst, const ucm_bistro_lock_t *bytes);
      |                                              ^~~~~~~~~~~~~~~~~
make[2]: *** [Makefile:874: bistro/libucm_la-bistro_ppc64.lo] Error 1
make[2]: *** Waiting for unfinished jobs....
In file included from ../../../src/ucm/bistro/bistro.c:18:
/home/~~~/scratch/ucx/build/../src/ucm/bistro/bistro_int.h:36:46: error: unknown type name ‘ucm_bistro_lock_t’
   36 | void ucm_bistro_modify_code(void *dst, const ucm_bistro_lock_t *bytes);
      |                                              ^~~~~~~~~~~~~~~~~
../../../src/ucm/bistro/bistro.c:52:46: error: unknown type name ‘ucm_bistro_lock_t’
   52 | void ucm_bistro_modify_code(void *dst, const ucm_bistro_lock_t *bytes)
      |                                              ^~~~~~~~~~~~~~~~~
../../../src/ucm/bistro/bistro.c: In function ‘ucm_bistro_apply_patch_atomic’:
../../../src/ucm/bistro/bistro.c:72:36: error: ‘ucm_bistro_lock_t’ undeclared (first use in this function); did you mean ‘ucm_bistro_protect’?
   72 |     size_t skip           = sizeof(ucm_bistro_lock_t);
      |                                    ^~~~~~~~~~~~~~~~~
      |                                    ucm_bistro_protect
../../../src/ucm/bistro/bistro.c:72:36: note: each undeclared identifier is reported only once for each function it appears in
../../../src/ucm/bistro/bistro.c:83:5: error: implicit declaration of function ‘ucm_bistro_patch_lock’; did you mean ‘ucm_bistro_patch_toc’? [-Werror=implicit-function-declaration]
   83 |     ucm_bistro_patch_lock(dst);
      |     ^~~~~~~~~~~~~~~~~~~~~
      |     ucm_bistro_patch_toc
../../../src/ucm/bistro/bistro.c:83:5: error: nested extern declaration of ‘ucm_bistro_patch_lock’ [-Werror=nested-externs]
cc1: all warnings being treated as errors
make[2]: *** [Makefile:853: bistro/libucm_la-bistro.lo] Error 1
make[2]: Leaving directory '/gpfs/fs0/scratch/~~~/ucx/build/src/ucm'
make[1]: *** [Makefile:937: install-recursive] Error 1
make[1]: Leaving directory '/gpfs/fs0/scratch/~~~/ucx/build/src/ucm'
make: *** [Makefile:783: install-recursive] Error 1
```

CONCLUSION: oops i forgot im actually trying to get away from ucx, so let's try next to build openmpi WITHOUT ucx

alphataubio:
==== TEST I ==== cmake/3.21.4 cuda/11.7.1 gcc/11.4.0 openmpi/5.0.1 ====

> wget https://download.open-mpi.org/release/open-mpi/v5.0/openmpi-5.0.1.tar.gz
tar xzvf openmpi-5.0.1.tar.gz; cd openmpi-5.0.1
./configure --prefix=$HOME/scratch/local --with-cuda=$CUDA_HOME --with-cuda-libdir=$CUDA_HOME/lib64/stubs/
make -j 64 all; make install

> lmp_$CONFIG -k on g 1 -sf kk -pk kokkos neigh half -var x 2 -var y 2 -var z 2 -in in.rhodo.scaled -l log.$CONFIG.rhodo.scaled.1


> mpirun -np 1 lmp_$CONFIG -k on g 1 -sf kk -pk kokkos neigh half -var x 2 -var y 2 -var z 2 -in in.rhodo.scaled -l log.$CONFIG.rhodo.scaled.1_1


> mpirun -np 4 lmp_$CONFIG -k on g 4 -sf kk -pk kokkos neigh half -var x 2 -var y 2 -var z 2 -in in.rhodo.scaled -l log.$CONFIG.rhodo.scaled.4_4

ALL OK

openmpi/5.0.1 built cuda aware so let's push it to full node 32 mpi ranks > 4 gpus. MPS is not available yet, i have to wait until cluster support staff back from holidays but it should run correctly even if there might be performance issues.

"If you use multiple MPI ranks per GPU, then you MUST enable CUDA MPS (Multi-Process Service :: GPU Deployment and Management Documentation ) to get good performance." - @stamoor on matsci

> mpirun -np 32 lmp_$CONFIG -k on g 4 -sf kk -pk kokkos neigh half -var x 8 -var y 8 -var z 8 -in in.rhodo.scaled -l log.$CONFIG.rhodo.scaled.32_4

FAILED with "Shake atoms missing". this is the standard rhodo scaled benchmark, unlikely related to model correctness or "unphysical behavior", plus it runs fine with mpi ranks <= gpus.

```
Per MPI rank memory allocation (min/avg/max) = 1332 | 1333 | 1334 Mbytes
------------ Step              0 ----- CPU =            0 (sec) -------------
TotEng   = -12982067.7154 KinEng   =  10979753.6175 Temp     =       299.0273
PotEng   = -23961821.3329 E_bond   =   1299452.9334 E_angle  =   5591743.5948
E_dihed  =   2668434.6878 E_impro  =    109317.9317 E_vdwl   =  -1181626.0687
E_coul   = 103991193.7841 E_long   = -136440338.1960 Press    =      -148.6868
Volume   = 157693457.1520
ERROR on proc 26: Shake atoms missing on proc 26 at step 10 (src/KOKKOS/fix_shake_kokkos.cpp:275)
Last command: run             100
ERROR on proc 28: Shake atoms missing on proc 28 at step 10 (src/KOKKOS/fix_shake_kokkos.cpp:275)
Last command: run             100
ERROR on proc 31: Shake atoms missing on proc 31 at step 10 (src/KOKKOS/fix_shake_kokkos.cpp:275)
Last command: run             100
ERROR on proc 25: Shake atoms missing on proc 25 at step 10 (src/KOKKOS/fix_shake_kokkos.cpp:275)
Last command: run             100
MPI_ABORT was invoked on rank 28 in communicator MPI_COMM_WORLD
```

ok let's back up and find the breaking point of mpi rank to gpu oversubscription. watching nvidia-smi, it looks like unpredictable behavior once enough mpi ranks to sustain gpus at 100% continuously. this might not happen with MPS enabled but MPS not always possible to enable unless user has root access or support willing to enable it. kokkos should run correctly if MPS not available, even you have to tolerate performance issues (better slower than broken...)

> mpirun -np 8 lmp_$CONFIG -k on g 4 -sf kk -pk kokkos neigh half -var x 8 -var y 8 -var z 8 -in in.rhodo.scaled -l log.$CONFIG.rhodo.scaled.8_4

OK (gpus operating in ~60-100% range, but never all 4 at 100% at same time)

> mpirun -np 16 lmp_$CONFIG -k on g 4 -sf kk -pk kokkos neigh half -var x 8 -var y 8 -var z 8 -in in.rhodo.scaled -l log.$CONFIG.rhodo.scaled.16_4

FAILED with cudaErrorIllegalAddress in Kokkos_Cuda_Instance.cpp:166

```
Per MPI rank memory allocation (min/avg/max) = 2779 | 2780 | 2781 Mbytes
------------ Step              0 ----- CPU =            0 (sec) -------------
TotEng   = -14604810.7581 KinEng   =  12352222.8197 Temp     =       299.0273
PotEng   = -26957033.5778 E_bond   =   1461884.5501 E_angle  =   6290711.5441
E_dihed  =   3001989.0237 E_impro  =    122982.6732 E_vdwl   =  -1329329.3273
E_coul   = 116924984.0366 E_long   = -153430256.0783 Press    =      -148.6717
Volume   = 177405139.2960
(ptr->cuda_stream_synchronize_wrapper(stream)) error( cudaErrorIllegalAddress): an illegal memory access was encountered /home/~~~/scratch/lammps-fork/lib/kokkos/core/src/Cuda/Kokkos_Cuda_Instance.cpp:166
Backtrace:
[0x7fffa5397f24] Kokkos::Impl::save_stacktrace()
[0x7fffa53682c4] Kokkos::Impl::host_abort(char const*)
[0x7fffa53a1918] Kokkos::Impl::cuda_internal_error_abort(cudaError, char const*, char const*, int)
[0x7fffa53a1be8] Kokkos::Impl::cuda_stream_synchronize(CUstream_st*, Kokkos::Impl::CudaInternal const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
[0x7fffa53a1c90] Kokkos::Impl::CudaInternal::fence(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) const
[0x7fffa53a1e5c] Kokkos::Cuda::fence(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) const
[0x7fffa3e05960] void LAMMPS_NS::CommKokkos::exchange_device<Kokkos::Cuda>()
[0x7fffa3de9478] LAMMPS_NS::CommKokkos::exchange()
[0x7fffa44166e8] LAMMPS_NS::VerletKokkos::run(int)
[0x7fffa3a9a298] LAMMPS_NS::Run::command(int, char**)
[0x7fffa389da48] LAMMPS_NS::Input::execute_command()
[0x7fffa389dedc] LAMMPS_NS::Input::file()
[0x100011d4]
[0x7fffa1f29de8]
[0x7fffa1f29fd4] __libc_start_main
[mist-login01:3598728] *** Process received signal ***
[mist-login01:3598728] Signal: Aborted (6)
[mist-login01:3598728] Signal code:  (-6)
[mist-login01:3598728] [ 0] linux-vdso64.so.1(__kernel_sigtramp_rt64+0x0)[0x7fffacbb04d8]
[mist-login01:3598728] [ 1] /lib64/glibc-hwcaps/power9/libc-2.28.so(gsignal+0xd8)[0x7fffa1f49628]
[mist-login01:3598728] [ 2] /lib64/glibc-hwcaps/power9/libc-2.28.so(abort+0x164)[0x7fffa1f23eb4]
[mist-login01:3598728] [ 3] /home/~~~/scratch/local/lib64/liblammps_issue3775_fft_kokkos-power9-volta70-cufft.so.0(_ZN6Kokkos4Impl10host_abortEPKc+0x78)[0x7fffa53682d8]
[mist-login01:3598728] [ 4] /home/~~~/scratch/local/lib64/liblammps_issue3775_fft_kokkos-power9-volta70-cufft.so.0(_ZN6Kokkos4Impl25cuda_internal_error_abortE9cudaErrorPKcS3_i+0x128)[0x7fffa53a1918]
[mist-login01:3598728] [ 5] (ptr->cuda_stream_synchronize_wrapper(stream)) error( cudaErrorIllegalAddress): an illegal memory access was encountered /home/~~~/scratch/lammps-fork/lib/kokkos/core/src/Cuda/Kokkos_Cuda_Instance.cpp:166
Backtrace:
/home/~~~/scratch/local/lib64/liblammps_issue3775_fft_kokkos-power9-volta70-cufft.so.0(_ZN6Kokkos4Impl23cuda_stream_synchronizeEP11CUstream_stPKNS0_12CudaInternalERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE+0x1b8)[0x7fffa53a1be8]
[mist-login01:3598728] [ 6] /home/~~~/scratch/local/lib64/liblammps_issue3775_fft_kokkos-power9-volta70-cufft.so.0(_ZNK6Kokkos4Impl12CudaInternal5fenceERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE+0x60)[0x7fffa53a1c90]
[mist-login01:3598728] [ 7] /home/~~~/scratch/local/lib64/liblammps_issue3775_fft_kokkos-power9-volta70-cufft.so.0(_ZNK6Kokkos4Cuda5fenceERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE+0x1c)[0x7fffa53a1e5c]
[mist-login01:3598728] [ 8] /home/~~~/scratch/local/lib64/liblammps_issue3775_fft_kokkos-power9-volta70-cufft.so.0(_ZN9LAMMPS_NS10CommKokkos15exchange_deviceIN6Kokkos4CudaEEEvv+0x1720)[0x7fffa3e05960]
[mist-login01:3598728] [ 9] /home/~~~/scratch/local/lib64/liblammps_issue3775_fft_kokkos-power9-volta70-cufft.so.0(_ZN9LAMMPS_NS10CommKokkos8exchangeEv+0x1a8)[0x7fffa3de9478]
[mist-login01:3598728] [10] /home/~~~/scratch/local/lib64/liblammps_issue3775_fft_kokkos-power9-volta70-cufft.so.0(_ZN9LAMMPS_NS12VerletKokkos3runEi+0x10b8)[0x7fffa44166e8]
[mist-login01:3598728] [11] /home/~~~/scratch/local/lib64/liblammps_issue3775_fft_kokkos-power9-volta70-cufft.so.0(_ZN9LAMMPS_NS3Run7commandEiPPc+0x1228)[0x7fffa3a9a298]
[mist-login01:3598728] [12] /home/~~~/scratch/local/lib64/liblammps_issue3775_fft_kokkos-power9-volta70-cufft.so.0(_ZN9LAMMPS_NS5Input15execute_commandEv+0xdb8)[0x7fffa389da48]
[mist-login01:3598728] [13] /home/~~~/scratch/local/lib64/liblammps_issue3775_fft_kokkos-power9-volta70-cufft.so.0(_ZN9LAMMPS_NS5Input4fileEv+0x22c)[0x7fffa389dedc]
[mist-login01:3598728] [14] lmp_issue3775_fft_kokkos-power9-volta70-cufft[0x100011d4]
[mist-login01:3598728] [15] /lib64/glibc-hwcaps/power9/libc-2.28.so(+0x29de8)[0x7fffa1f29de8]
[mist-login01:3598728] [16] /lib64/glibc-hwcaps/power9/libc-2.28.so(__libc_start_main+0xb4)[0x7fffa1f29fd4]
[mist-login01:3598728] *** End of error message ***
[0x7fff92c87f24] Kokkos::Impl::save_stacktrace()
[0x7fff92c582c4] Kokkos::Impl::host_abort(char const*)
[0x7fff92c91918] Kokkos::Impl::cuda_internal_error_abort(cudaError, char const*, char const*, int)
[0x7fff92c91be8] Kokkos::Impl::cuda_stream_synchronize(CUstream_st*, Kokkos::Impl::CudaInternal const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
[0x7fff92c91c90] Kokkos::Impl::CudaInternal::fence(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) const
[0x7fff92c91e5c] Kokkos::Cuda::fence(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) const
[0x7fff916f5960] void LAMMPS_NS::CommKokkos::exchange_device<Kokkos::Cuda>()
[0x7fff916d9478] LAMMPS_NS::CommKokkos::exchange()
[0x7fff91d066e8] LAMMPS_NS::VerletKokkos::run(int)
[0x7fff9138a298] LAMMPS_NS::Run::command(int, char**)
[0x7fff9118da48] LAMMPS_NS::Input::execute_command()
[0x7fff9118dedc] LAMMPS_NS::Input::file()
```

alphataubio:
==== TEST J ==== openmpi/4.1.6 ====

regression test to openmpi/4.1.6 (i) cuda-aware and (ii) without ucx.

is kokkos mpi ranks > gpu unpredictable behavior specific to openmpi/5.0.1 ? or does it happen also in previous "still supported / current stable" version of openmpi (4.1.6) ?? openmpi3 now retired by open-mpi.org so no need to also test openmpi3.

> wget https://download.open-mpi.org/release/open-mpi/v4.1/openmpi-4.1.6.tar.gz
tar xzvf openmpi-4.1.6.tar.gz; cd openmpi-4.1.6
module purge; module load MistEnv/2021a cmake/3.21.4 cuda/11.7.1 gcc/11.4.0
./configure --prefix=$HOME/scratch/local/openmpi-4.1.6 --with-cuda=$CUDA_HOME --with-cuda-libdir=$CUDA_HOME/lib64/stubs
make -j 64 all; make install

> module use $HOME/local/modules
module purge; module load MistEnv/2021a cuda/11.7.1 gcc/11.4.0 openmpi/4.1.6
CONFIG=issue3775_fft_kokkos-power9-volta70-cufft-openmpi4
mkdir build-$CONFIG; cd build-$CONFIG
cmake ../cmake -D BUILD_SHARED_LIBS=yes -D LAMMPS_MACHINE=$CONFIG \
    -D CMAKE_INSTALL_PREFIX=~/scratch/local \
    -D PKG_KOKKOS=yes -D Kokkos_ENABLE_CUDA=yes -D FFT_KOKKOS=CUFFT \
    -D Kokkos_ARCH_HOSTARCH=POWER9 -D Kokkos_ARCH_GPUARCH=VOLTA70 \
    -D Kokkos_ENABLE_OPENMP=yes -D BUILD_OMP=yes \
    -D Kokkos_ENABLE_SERIAL=yes -DKokkos_ARCH_VOLTA70=ON  \
    -D PKG_EXTRA-DUMP=yes -D PKG_KSPACE=yes -D PKG_COLVARS=yes \
    -D PKG_ASPHERE=yes -D PKG_CG_DNA=yes \
    -D PKG_EXTRA-PAIR=yes -D PKG_MOLECULE=yes -D PKG_RIGID=yes ; \
    cmake --build . -j 64; make install

> mpirun -np 4 lmp_$CONFIG -k on g 4 -sf kk -pk kokkos neigh half -var x 2 -var y 2 -var z 2 -in in.rhodo.scaled


```
By default, for Open MPI 4.0 and later, infiniband ports on a device
are not used by default.  The intent is to use UCX for these devices.
You can override this policy by setting the btl_openib_allow_ib MCA parameter
to true.

  Local host:              mist-login01
  Local adapter:           mlx5_0
  Local port:              1

WARNING: There was an error initializing an OpenFabrics device.

  Local host:   mist-login01
  Local device: mlx5_0

libgomp: Thread creation failed: Resource temporarily unavailable
```

> export OMPI_MCA_btl_openib_allow_ib=1
export OMPI_MCA_btl_openib_warn_default_gid_prefix=0
export OMP_NUM_THREADS=1
mpirun -np 4 lmp_$CONFIG -k on g 4 -sf kk -pk kokkos neigh half -var x 2 -var y 2 -var z 2 -in in.rhodo.scaled

```
Setting up Verlet run ...
  Unit style    : real
  Current step  : 0
  Time step     : 2
The call to cuIpcGetMemHandle failed. This means the GPU RDMA protocol
cannot be used.
  cuIpcGetMemHandle return value:   1
  address: 0x628d3ea80
Check the cuda.h file for what the return value means. Perhaps a reboot
of the node will clear the problem.
```

various opemnpi network transports (omnipath, rdma, OpenFabrics, ucx, ... ) on my local clusters are a nightmare to get working, and not lammps responsibility.

so im giving up on openmpi-4.1.6 regression test to focus on my priority task of charmmfsw-kk update (see next test K). i have one environment that works (cuda/11.7.1 gcc/11.4.0 openmpi/5.0.1) so that's what i'll be using. lammps devs have to be right for ALL (current) possible combinations of hardware/os/libraries, i only need one that works <grin> ...

maybe openmpi5 is not supported by kokkos library and lammps kokkos pkg, only openmpi4 ?

alphataubio:
==== TEST K ====


now im finally getting back to work on my charmmfsw-kk update ( [PR 4030](https://github.com/lammps/lammps/pull/4030) ). im using the 1hvn charmm-gui scripts from above but charmmfsw needs to be changed to charmm temporarily in 1hvn input files (these scripts use kspace pppm/kk with cufft):

[1hvn-lammps-charmm.zip](https://github.com/lammps/lammps/files/13808240/1hvn-lammps-charmm.zip)

even though nvidia docs says MPS daemon needs root, i tried it with my user and it worked. so maybe cluster admins allow whatever special root permission is needed for starting MPS server to all users.

> export CUDA_VISIBLE_DEVICES=0,1,2,3
export CUDA_MPS_LOG_DIRECTORY=~/scratch/log
export CUDA_MPS_PIPE_DIRECTORY=/tmp/nvidia-mps
nvidia-cuda-mps-control -d

### lmp_issue3775_fft_kokkos-power9-volta70-cufft (serial and mpi with gpu) hangs at 100% cpu on host, 0% on device in LAMMPS_NS::Domain::closest_image at lammps/src/domain.cpp:1075 (maybe this is related to @akohlmey recent cmap changes in MOLECULE/fix_cmap.cpp, or maybe cmap never worked in kokkos before ?):

>module use $HOME/local/modules
module purge; module load MistEnv/2021a cuda/11.7.1 gcc/11.4.0 openmpi/5.0.1
CONFIG=issue3775_fft_kokkos-power9-volta70-cufft
lmp_$CONFIG -k on g 1 -sf kk -pk kokkos neigh half -in step4.0_minimization.inp

    
```
Setting up cg/kk style minimization ...
  Unit style    : real
  Current step  : 0
WARNING: Fix with atom-based arrays not compatible with sending data in Kokkos communication, switching to classic exchange/border communication (src/KOKKOS/comm_kokkos.cpp:666)
WARNING: Fix with atom-based arrays not compatible with Kokkos sorting on device, switching to classic host sorting (src/KOKKOS/atom_kokkos.cpp:215)
^C
Thread 1 "lmp_issue3775_f" received signal SIGINT, Interrupt.
0x00007fffee205cf8 in LAMMPS_NS::Domain::closest_image (this=0x1bc0f3e0, i=<optimized out>, j=0)
    at /home/t/tperkins/mmurp027/scratch/lammps/src/domain.cpp:1075
1075      if (rsq < rsqmin) {
(gdb) bt
#0  0x00007fffee205cf8 in LAMMPS_NS::Domain::closest_image (this=0x1bc0f3e0, i=<optimized out>, j=0)
    at /home/t/tperkins/mmurp027/scratch/lammps/src/domain.cpp:1075
#1  0x00007fffee8ec980 in LAMMPS_NS::FixCMAP::pre_neighbor (this=0x1bc47970) at /home/t/tperkins/mmurp027/scratch/lammps/src/MOLECULE/fix_cmap.cpp:267
#2  0x00007fffee8e9fc8 in LAMMPS_NS::FixCMAP::min_setup (this=<optimized out>, vflag=<optimized out>)
    at /home/t/tperkins/mmurp027/scratch/lammps/src/MOLECULE/fix_cmap.cpp:224
#3  0x00007fffeeb2f100 in LAMMPS_NS::ModifyKokkos::setup (this=0x1bc24c70, vflag=<optimized out>)
    at /home/t/tperkins/mmurp027/scratch/lammps/src/KOKKOS/modify_kokkos.cpp:72
#4  0x00007fffeea0c35c in LAMMPS_NS::MinKokkos::setup (this=0x10fc67d0, flag=<optimized out>)
    at /home/t/tperkins/mmurp027/scratch/lammps/src/KOKKOS/min_kokkos.cpp:218
#5  0x00007fffee537cfc in LAMMPS_NS::Minimize::command (this=0x1bc380b0, narg=<optimized out>, arg=<optimized out>)
    at /home/t/tperkins/mmurp027/scratch/lammps/src/minimize.cpp:59
#6  0x00007fffee3f8f38 in LAMMPS_NS::Input::execute_command (this=0x1b30ac50) at /home/t/tperkins/mmurp027/scratch/lammps/src/input.cpp:868
#7  0x00007fffee3f93cc in LAMMPS_NS::Input::file (this=0x1b30ac50) at /home/t/tperkins/mmurp027/scratch/lammps/src/input.cpp:313
#8  0x00000000100011b4 in main (argc=<optimized out>, argv=<optimized out>) at /home/t/tperkins/mmurp027/scratch/lammps/src/main.cpp:77
(gdb)
```

### however this runs fine with (i) lmp_issue3775_fft_kokkos-power9-volta70-cufft (mpi ranks only no gpu):

> mpirun -np 32 lmp_$CONFIG -in step4.0_minimization.inp

```
98.5% CPU use with 32 MPI tasks x 1 OpenMP threads

Minimization stats:
  Stopping criterion = linesearch alpha is zero
  Energy initial, next-to-last, final =
     -40209.7946227601  -44187.4202160619  -44181.5485423886
  Force two-norm initial, final = 839.77625 81.291537
  Force max component initial, final = 211.61099 16.774262
  Final line search alpha, max atom move = 1.220508e-12 2.0473121e-11
  Iterations, force evaluations = 204 467

MPI task timing breakdown:
Section |  min time  |  avg time  |  max time  |%varavg| %total
---------------------------------------------------------------
Pair    | 3.8515     | 4.3607     | 6.2207     |  19.4 | 52.00
Bond    | 0.0089264  | 0.01475    | 0.034281   |   7.4 |  0.18
Kspace  | 1.2931     | 3.1587     | 3.6549     |  22.9 | 37.66
Neigh   | 0.13553    | 0.13586    | 0.13596    |   0.0 |  1.62
Comm    | 0.04528    | 0.047985   | 0.055784   |   1.3 |  0.57
Output  | 0.0011617  | 0.0016163  | 0.0024809  |   0.8 |  0.02
Modify  | 0.65113    | 0.65202    | 0.65308    |   0.1 |  7.77
Other   |            | 0.0147     |            |       |  0.18

Nlocal:        320.156 ave         345 max         305 min
Histogram: 5 3 6 3 5 4 1 2 2 1
Nghost:        7407.28 ave        7532 max        7316 min
Histogram: 4 4 5 4 5 2 1 2 3 2
Neighs:         302288 ave      331051 max      286922 min
Histogram: 5 6 7 1 3 4 3 2 0 1

Total # of neighbors = 9673205
Ave neighs/atom = 944.1879
Ave special neighs/atom = 2.3664226
Neighbor list builds = 9
Dangerous builds = 0
```

### and (ii) package GPU instead of kokkos:

> CONFIG=develop-power9-volta70-gpu
mpirun -np 32 lmp_$CONFIG -sf gpu -pk gpu 4 -in step4.0_minimization.inp

```
90.0% CPU use with 32 MPI tasks x 1 OpenMP threads

Minimization stats:
  Stopping criterion = linesearch alpha is zero
  Energy initial, next-to-last, final =
     -40209.7642377623  -44267.3238550334  -44260.6392704077
  Force two-norm initial, final = 839.77628 102.84307
  Force max component initial, final = 211.61096 35.255666
  Final line search alpha, max atom move = 1.4843068e-12 5.2330227e-11
  Iterations, force evaluations = 217 493

MPI task timing breakdown:
Section |  min time  |  avg time  |  max time  |%varavg| %total
---------------------------------------------------------------
Pair    | 0.17514    | 0.31824    | 0.44341    |  10.7 |  9.85
Bond    | 0.0092351  | 0.014932   | 0.034654   |   7.2 |  0.46
Kspace  | 1.9177     | 1.9412     | 1.9511     |   0.7 | 60.08
Neigh   | 0.00070074 | 0.0011064  | 0.0012073  |   0.4 |  0.03
Comm    | 0.051038   | 0.054623   | 0.062868   |   1.3 |  1.69
Output  | 0.0013147  | 0.0018611  | 0.0028205  |   0.8 |  0.06
Modify  | 0.75188    | 0.87734    | 1.0191     |   6.3 | 27.15
Other   |            | 0.02171    |            |       |  0.67

Nlocal:        320.156 ave         344 max         303 min
Histogram: 2 5 6 4 3 6 2 0 1 3
Nghost:        7405.78 ave        7530 max        7309 min
Histogram: 4 4 3 4 5 4 1 2 3 2
Neighs:              0 ave           0 max           0 min
Histogram: 32 0 0 0 0 0 0 0 0 0

Total # of neighbors = 0
Ave neighs/atom = 0
Ave special neighs/atom = 2.3664226
Neighbor list builds = 10
Dangerous builds = 0
```

alphataubio:
==== TEST I (cont.) ====

> FAILED with "Shake atoms missing".

UPDATE: i tried fix wall/reflect to keep water molecules from flying out the simulation box. maybe the larger the system, the higher charges involved and randomly a few water molecules shoot off with too much energy to be properly handled by default boundary p p p ?

`boundary f f f`
`fix 3 all wall/reflect xlo EDGE xhi EDGE ylo EDGE yhi EDGE zlo EDGE zhi EDGE`

[in.rhodo-wall.scaled.txt](https://github.com/lammps/lammps/files/13808683/in.rhodo-wall.scaled.txt)


but i get many error messages:

```
WARNING: Non-zero imageflag(s) in x direction for non-periodic boundary reset to zero (src/read_data.cpp:1421)
WARNING: Non-zero imageflag(s) in y direction for non-periodic boundary reset to zero (src/read_data.cpp:1425)
WARNING: Non-zero imageflag(s) in z direction for non-periodic boundary reset to zero (src/read_data.cpp:1429)
[...]
WARNING: Replicating in a non-periodic dimension (src/replicate.cpp:79)
[...]
ERROR: Cannot use fix npt on a non-periodic z dimension (src/fix_nh.cpp:394)
```

based on my reading of a few matsci topics, this change from periodic to fixed boundary conditions is:

(i) not something i should be doing in the first place.

(ii) i dont have the original rhodo data to re-equilibrate from scratch, just a restart file after 5000 timesteps.

(iii) the resulting system would be too much different from the original rhodo benchmark anyways

(iv) im a lammps beginner, i dont even know how to do a "dissipative thermostat"

stanmoore1:
@hagertnl I tested this PR and it appears to be giving the expected behavior. One limitation that remains (that I would not expect this PR to fix) is that we cannot specify a different library for Kokkos Host CPU vs Kokkos GPU. For example, if one tries to run on the host CPU with Kokkos and cuFFT (i.e. using `pppm/kk/host` in the input) then it errors out. This would require a much more non-trivial refactor to fix. That aside, I think this is a useful change and should be merged.

I talked to @akohlmey and we want to add some output to PPPM that shows which FFT library is selected. Otherwise this should be good to go from my POV.

akohlmey:
@hagertnl @stanmoore1 I have done a first pass over this PR to include support for heFFTe. Unfortunately, I came across some inconsistencies and even compilation problems of heFFTe when testing. This requires a bit more work from me to implement the introspection that I had talked with @stanmoore1. I will have to see if I can make selecting heFFTe a bit more user friendly with little effort and include this here or start with a new PR with some more fundamental changes that will also allow to incorporate heFFTe at some later point in time for KOKKOS.

stanmoore1:
Thanks @akohlmey!

akohlmey:
@stanmoore1 this now includes displaying the FFT configuration with the "info fft" command. (I have not yet worked out a good way to do introspection.). Output looks like this:
```
info fft

Printed on Mon Feb 05 16:47:42 2024

FFT information:
FFT engine  = HeFFTe
FFT library = (builtin)
KOKKOS FFT engine  = mpiFFT
KOKKOS FFT library = KISS

```
Or this:
```
Printed on Mon Feb 05 16:50:01 2024

FFT information:
FFT engine  = mpiFFT
FFT library = FFTW3

```

stanmoore1:
Very nice! Thanks for working on this.

akohlmey:
@stanmoore1 I am satisfied with this (for now). Do you want to give it a test (to see if it still works as documented and expected on the machines you have access to but I don't? Especially those with GPUs?

stanmoore1:
Yes will do.

akohlmey (approved):

stanmoore1:
I tested this on V100 with cuFFT using the Makefile build and it works as expected.

cmake/Modules/Packages/KOKKOS.cmake
```diff
@@ -126,16 +126,32 @@ if(PKG_KSPACE)
...
+  validate_option(FFT_KOKKOS FFT_KOKKOS_VALUES)
+  string(TOUPPER ${FFT_KOKKOS} FFT_KOKKOS)
+
   if(Kokkos_ENABLE_CUDA)
-    if(NOT (FFT STREQUAL "KISS"))
-      target_compile_definitions(lammps PRIVATE -DFFT_CUFFT)
+    if(NOT ((FFT_KOKKOS STREQUAL "KISS") OR (FFT_KOKKOS STREQUAL "CUFFT")))
+      message(FATAL_ERROR "The CUDA backend of Kokkos requires either KISS FFT or CUFFT.")
+    elseif(FFT_KOKKOS STREQUAL "KISS")
+      message(WARNING "Using KISS FFT with the CUDA backend of Kokkos may be sub-optimal.")
```
stanmoore1, line 139: Do we want to add a runtime warning as well?

stanmoore1 (approved):
Looks good. Had one question in the comments.

src/KSPACE/msm.cpp
```diff
@@ -140,10 +140,6 @@ void MSM::init()
   if ((order < 4) || (order > 10) || (order%2 != 0))
     error->all(FLERR,"MSM order must be 4, 6, 8, or 10");
 
-  if (sizeof(FFT_SCALAR) != 8)
```
stanmoore1, line 142: Is this safe to remove? I thought there was some reason for the error check, but can't remember, would have to look at the code.

EDIT: looks like it is never used.

sjplimp (approved):

src/KSPACE/msm.cpp
akohlmey: Yes. Nothing inside the msm code *uses* FFT_SCALAR except this line so there is no connection to FFT single.

Perhaps initially the thought was that single precision could be used as well to speed up communication, but the MSM code has been all but abandoned. If somebody wants to work on this, it would first need to be updated to the latest state. Dave Hardy told me at a meeting a few years ago, that there are a couple new papers that describe how it can be made faster and more accurate.

cmake/Modules/Packages/KOKKOS.cmake
akohlmey: Be my guest. ;-)

src/KOKKOS/pppm_kokkos.h
```diff
@@ -36,30 +36,30 @@ KSpaceStyle(pppm/kk/host,PPPMKokkos<LMPHostType>);
 // fix up FFT defines for KOKKOS with CUDA and HIP
 
 #ifdef KOKKOS_ENABLE_CUDA
```
stanmoore1: @akohlmey I think this code should be deleted now, do you agree?

akohlmey: No. Only the CMake build is set up to have only one FFT choice for either KOKKOS and non-KOKKOS. For the GNU make build we have to assume any mix of "overlapping" FFT defines.

With the changes in this PR, it may be better to move those to `lmpfftsettings.h` which should be the canonical file to guarantee consistent FFT settings and defines.

akohlmey: This is a bit like:
```
// if a user sets FFTW, it means FFTW3

#ifdef FFT_FFTW
#ifndef FFT_FFTW3
#define FFT_FFTW3
#endif
#endif
```

To this block which we should probably now also add an `#undef FFT_FFTW` for increased consistency.
I have been pondering to add statements to error out for incompatible choices.
Ultimately, I've chosen to wait a bit longer until FFT support for Intel GPUs in KOKKOS is sorted out and possibly HeFFTe support extended to support Kokkos and then integrated into LAMMPS.
After that would be a good point to look at all possible combination of options and map out what should be allowed, preferred, or disallowed.

stanmoore1: OK thanks for the explanation

stanmoore1:
I tested the warning with KISS FFT on GPUs and I ran cuFFT on V100 again, and it still works. This PR is ready to merge from my POV.

cmake/Modules/Packages/KOKKOS.cmake
stanmoore1: Done.

src/KOKKOS/fftdata_kokkos.h
```diff
@@ -12,117 +12,108 @@
...
-#error "FFT_PRECISION needs to be either 1 (=single) or 2 (=double)"
-#endif
-
-
 // with KOKKOS in CUDA or HIP mode we can only have
 // CUFFT/HIPFFT or KISSFFT, thus undefine all other
 // FFTs here, since they may be valid in fft3d.cpp
 
 #ifdef KOKKOS_ENABLE_CUDA
-# if defined(FFT_FFTW)
```
stanmoore1, line 44: @akohlmey I just noticed more redundant checks here, I think they should be merged into `lmpfftsettings`

PR #4015, opened 2023-12-14
add border information to Atom::add_custom and Atom::find_custom

srtee:
**Summary**

Add `border` info to `find_custom`

**Author(s)**

Shern Tee, Griffith University, Brisbane, Australia

**Licensing**

**Backward Compatibility**

No input-affecting changes.

**Implementation Notes**

Allows other classes to determine if a custom field managed by a FixPropertyAtom instance has border communication enabled (`fix property/atom ghost yes`) by supplying an additional int reference when calling `find_custom`. This information is set by `fix property/atom` during the fix's constructor. 

**Post Submission Checklist**


akohlmey:
Not so sure about the utility of this function. The reason to use fix property/atom is when there is a specific name for a property to be used in the input that is being referred to, e.g. as property in set or other commands (or as of recent in (atom style) variables). Otherwise an (internal) fix STORE/ATOM instance should be used.

 Thus access to "nvalue", "index" and "styles" is irrelevant. The "ncols" information is provided when you use `Atom::find_custom()` to look up the custom property. So only "border" is an information that is not available or relevant otherwise.

srtee:
Hi Axel, I've now taken a different and more systematic route:

- `Atom::add_custom` now stores border information supplied by `FixPropertyAtom`, in `Atom::custom_border`
- `Atom::find_custom` returns border information if supplied three `int` references.

Using default arguments and function overloading, previous uses of the `Atom::add_custom` and `Atom::find_custom` functions should not be affected.

I have also given an example usage in the AMOEBA package, where `fix property/atom` is used to read custom fields from a data file and maintain custom fields for `pair amoeba`. The package requires that the integer vector `i_amtype` has ghost communication, as well as that the double array `d2_xyzaxis` has 3 columns (I assume it is not meant to be used in 2D simulations). Previously neither of these conditions were checked. This can be seen by either commenting out `ghost yes` or changing the dimensionality of `d2_xyzaxis` in the following lines in any of the `examples/amoeba in.*` input files:

```
fix             amtype all property/atom i_amtype ghost yes
fix             extra all property/atom i_amgroup d_redID d_pval ghost yes
fix             extra2 all property/atom i_polaxe d2_xyzaxis 3
```

This change allows much earlier error checking and (to me) clearer error messages.

akohlmey:
@srtee the final word for changes like this lies with @sjplimp, who is the architect of fix property/atom and wrote most of the AMOEBA package. I have assigned him to this PR and he should iterate with you all the details of the implementation. Once he is satisfied, he should assign the PR back to me and I will look over formal and style issues and check out tests. There should be tests added for this to the unittest tree once it is in its final form.

sjplimp:
@srtee Hi Shern - I looked over the PR and discussion.  I can see it is a
good idea to have a ghost flag for custom properties so that code can
verify the user defined them as expected in the input script.  The
only thing I would like to change is that the implementation be
similar to the existing custom property code for simplicity and
consistency.  Specifically, see the next 4 comments.

sjplimp:
(1) fix property/atom.cpp changes are fine
    except this comment:
    old: get ghost first for settings
    new: check for ghost keyword to use as add_custom() arg

sjplimp:
(2) atom.h
    use 4 integer vectors:
      int *ivghost, *iaghost, *dvghost, *daghost;
    instead of
      std::array<std::vector<int>, 4> custom_border;

sjplimp:
(3) atom.cpp

    init the 4 ghost vectors to NULL in constructor and free them in the destructor

    rename find_custom() with extra border arg to find_custom_ghost()
      use "ghost" as arg name instead of "border", add 4 lines of logic to return ghost for 4 cases

    could you add a doc header (same one for both find_custom() 
      and find_custom_ghost()) which docs the args, similar to
      the one for add_custom()

    it should emphasize the only reason to use find_custom_ghost()
      is if the code wants  to check the ghost status (not typically the case)

    add_custom() with extra optional arg:
      doc border arg in function header, rename arg to ghost
      add memory->srealloc() calls for ghost vectors to each of 4 cases

sjplimp:
(4) pair_amoeba.cpp changes are fine
    however, note that it is not really an "error" if ghost = 1 and the code doesn't use ghost value
    it's only an error if the code depends on ghost values and ghost = 0
    this will be the case for any client of fix property/atom which is why I think explicitly invoking a separate find_custom_ghost() method is a good idea

You are correct that AMOEBA is not usable for 2d systems, an error is generated.

srtee:
Hi Steve,

I agree with the reason for developing a separate `find_custom_ghost` method, it is quite a niche case and it's useful to not overload the much more commonly used `find_custom` method. I have made the suggested changes.

src/AMOEBA/pair_amoeba.cpp
```diff
@@ -827,28 +827,35 @@ void PairAmoeba::init_style()
...
-  if (index_polaxe < 0 || flag || cols)
-    error->all(FLERR,"Pair {} polaxe is not defined", mystyle);
-  index_pval = atom->find_custom("pval",flag,cols);
-  if (index_pval < 0 || !flag || cols)
-    error->all(FLERR,"Pair {} pval is not defined", mystyle);
+  char const * names[6] = {"amtype", "amgroup", "redID",
+    "xyzaxis", "polaxe", "pval"};
+  int const flag_check[6] = {0, 0, 1, 1, 0, 1}; // correct type (0 int, 1 dbl)
+  int const cols_check[6] = {0, 0, 0, 3, 0, 0}; // xyzaxis 3 cols, all others 0
+  int const ghost_check[6] = {1, 0, 0, 0, 0, 0}; // which types need ghost
```
sjplimp: In examples/amoeba/in.ubiquitin, 4 of the 6 property/atom vecs have ghost set.  Are you thinking  only one of them needs it?  I just spot-checked amgroup and I think it needs the ghost setting.

sjplimp (changes requested):
Thanks for the changes Shern - just want to double-check ghost settings to AMOEBA, then I will approve the merge.

src/AMOEBA/pair_amoeba.cpp
srtee: I've struggled to check this myself -- I can only get the water dimer and hexamer to run on my laptop, not the water box or ubiquitin. I've turned on the ghost check for `amgroup` and `pval` on the heuristic that, like `amtype`, there are settings in `pack_forward_comm` and `unpack_forward_comm` for those property/atom vecs.

sjplimp: Looks like a bug was introduced some time earlier.  This line in pair_amoeba.cpp
if (first_flag_compute && (comm->me == 0)) print_settings();
should be
if (first_flag_compute) print_settings();

Within print_settings() there is a call to final_accuracy_mpole() which has MPI_Allreduce() calls.
You can make the change in this PR.
@akohlmey - please verify that change is compatible with the intent of the refactoring of print_settings() ?

@srtee - I think the ghost error checks in AMOEBA should be for all 4 fix property/atom vecs/arrays which the input script defines with ghost yes.  I believe the options in (un)pack_forward_comm are only used in the setup of the system.  But AMOEBA requires those properties are assigned to ghost atoms every time atoms migrate and neighbor and ghost atoms are redefined.  That operation happens in fix property/atom, so AMOEBA needs to ensure fix property/atom has the ghost flag set for all 4 of them.

akohlmey: > Looks like a bug was introduced some time earlier. This line in pair_amoeba.cpp
> if (first_flag_compute && (comm->me == 0)) print_settings();
> should be
> if (first_flag_compute) print_settings();

This kind of construct has been in place from the very beginning. The inconsistent MPI behavior is *caused* by *calling* `final_accuracy_mpole()` which was added in commit 288629bd139 recently.

> @akohlmey - please verify that change is compatible with the intent of the refactoring of print_settings() ?

It needs additional changes to `print_settings()`. I will add them.

Please also note that compilation currently fails to complete due to missing changes to the KOKKOS package code.

src/AMOEBA/pair_amoeba.cpp
```diff
@@ -827,28 +827,35 @@ void PairAmoeba::init_style()
...
-  if (index_polaxe < 0 || flag || cols)
-    error->all(FLERR,"Pair {} polaxe is not defined", mystyle);
-  index_pval = atom->find_custom("pval",flag,cols);
-  if (index_pval < 0 || !flag || cols)
-    error->all(FLERR,"Pair {} pval is not defined", mystyle);
+  char const * names[6] = {"amtype", "amgroup", "redID",
+    "xyzaxis", "polaxe", "pval"};
+  int const flag_check[6] = {0, 0, 1, 1, 0, 1}; // correct type (0 int, 1 dbl)
+  int const cols_check[6] = {0, 0, 0, 3, 0, 0}; // xyzaxis 3 cols, all others 0
+  int const ghost_check[6] = {1, 1, 0, 0, 0, 1}; // which types need ghost; TO-DO: check
```
sjplimp: Unless @srtee has a reason he thinks the ghost check for redID is not needed, I think this line should be {1,1,1,0,0,1} - which matches the settings in examples/amoeba/in.ubiquitin

srtee: Happy to make the change -- I just haven't had time to do more testing myself this week.

sjplimp (approved):

Issue #2944, opened 2021-09-13
Fix deform produce incorrect restarts


antonpeshkov:
This is a continuation of the bug report described in the following MATSCI thread https://matsci.org/t/fix-deform-screw-restarts-and-what-to-do-with-it/38239 . 

They are two bugs related to fix deform and restarts. While preparing the report for this bug, I discovered two more problems, one related to the Windows build of lammps and another to the virial computation for spherical particles. As these two bugs are not related to fix deform, I will open separate issues for them. 

First, because box size/shape change happens after force computation, then if a run is restarted, the newly computed forces will be different from those computed by lammps before the restart.

Here is the archive, with files needed to reproduce this error:

[Lammps bug report.zip](https://github.com/lammps/lammps/files/7157474/Lammps.bug.report.zip)

The “*_random_positons” files contain the initial positions of particles. There are two lammps scripts that are identical, except that one use the lj/cut pair style, as asked by Alex, and the other the granular pair style. The script performs an isotropic compression with a true strain rate for N steps. It then makes a “run 0” to recompute the pressure as if we had restarted the simulation. At the end it writes to the screen the pressure after the N steps and the pressure after the “0” step restart. You can see than in both of the cases of lj and granular potentials these pressures are close but not equal (N.B. do not use Windows builds to explore these bugs, as mentioned previously, they are broken). However, in some cases, they can be wildly different.

To illustrate what is happening I include a Julia script that computes the pressure from the dump file output by the granular script. Initially the granular lammps script use no velocity-dependent forces (gamman=0). If you run with Julia the “collisional_damping_pressure_test.jl” file, it will compute you the pressure from the dump file that is equal to the one computed by lammps after the restart. Now, if you uncomment lines 45-47 in the Julia script, it will reincrease the box size one step back before computing the pressure. In that case, it will find a pressure equivalent to the one computed by lammps at the end of N steps. Note, that in order to normalize the pressure by the volume, I still use the volume after the final box compaction. This makes the behavior of pressure computation and fix deform even more nonsensical, as the forces are computed at one volume, but then normalized at another volume to compute the pressure.

[...]

PR #4017, opened 2023-12-14
Expanding deformation controls

jtclemm:
**Summary**
 
Many meso- and macroscopic problems using the GRANULAR or BPM packages require loading conditions which are currently not supported in LAMMPS. In particular, the ability to mix strain and stress controls is limited. For instance, one cannot apply a fixed mean pressure (isotropically dilating/contracting the box) with a fixed deviatoric strain rate (distorting the box shape).
 
Since these loading conditions cannot always decouple the deformation in different dimensions and only one fix can deform a dimension, I added more options to fix deform instead of creating a new fix. These include:
 
•	X, y, z, xy, xz, yz `pressure`: a standard linear controller with a gain constant. Accepts variable target pressures so it can oscillate, ramp, or be a function of other current stress components
•	X, y, z `pressure/mean`: analogous to the `volume` option, the dimension will deform to maintain a mean pressure
•	An `iso` option that isotropically scales the box to maintain either a constant volume or pressure. This option can be used in conjunction with separate x, y, and z options allowing one to simultaneously control the shear state/shape of the box (using x, y, z options) and its mean pressure/volume (using iso). In other words, one can separately control the 1st invariant of the stress tensor and the 2nd/3rd invariants of the strain tensor (or vice versa). This should be renamed to avoid confusion with fix nh `iso`, but I haven't thought of a better alternative.
•	Several pressure-relevant keywords like `couple` (analogous to fix nph), `max/rate` (capping the maximum strain rate on pressure-controlled dimensions), `normalize/pressure` (whether gains are normalized by the target pressure), and `vol/balance/p` (whether constant volume dimensions should try and balance their pressures)
 
With these options, fix deform should be able to support almost all loading conditions commonly used in the literature. The downside, of course, is that fix deform is even more complex. There might be a better way of organizing these controls or separating functionality across fixes, so if anyone has ideas I’d be happy to reorganize this PR. 
 
These changes should not affect any of the preexisting behavior of fix deform EXCEPT for the fact that the h_rate matrix is no longer zeroed between run statements (only in the constructor) and is now stored in restart files to resolve issue #2944. 
 
**Author(s)**
 
Joel Clemmer (SNL)
Kevin Hanley (Univ. Edinburgh)
Jose Salomon (Imperial College)
 
**Licensing**
 

jtclemm:
@sjplimp @akohlmey unless there are other ideas for parsing the child arguments, this PR is now ready for review. 

Besides changes to accommodate fix deform/pressure, there was one other important change to fix deform: the h matrix is now zeroed in the constructor instead of init() and it is saved to restart files. While this would resolve issue #2944, I'm not certain it's a necessary change.

sjplimp:
>> the h matrix is now zeroed in the constructor instead of init() and it is saved to restart files

Not sure that is a good idea.  LAMMPS does not know if the new fix deform command in the restart script is identical to the original one.  That is not required, since a user may want to apply a new kind of deformation in the 2nd stage of a run and could easily use the same fix ID.  In which case the h matrix will become the one in the restart file and not the one the new fix deform command defines (when set in init) ?  Previously the restart file only stored the initial box size, which is only needed if the user specifically wants the run to continue using that info, via the run start/stop keywords.  In which case they would presumably also use the same fix deform command.  If they don't do run start/stop, the fix deform restart info doesn't affect anything.

jtclemm:
> LAMMPS does not know if the new fix deform command in the restart script is identical to the original one.

Yes, that is a risk. Restarting does require that the styles match, i.e.
https://github.com/lammps/lammps/blob/3358fe36d1ae7c8723dd454d00e00ff3d200ca13/src/fix_deform.cpp#L999-L1000
but there's no guarantee the rates etc. will be equivalent. 

I don't think there is a correct solution, is it better to assume the old hmatrix still applies on the first timestep after reading a restart file or that the hmaxtrix is zero on that first timestep? Then again, it is a minor numerical effect so maybe the best answer is not to modify behavior.

sjplimp:
@jtclemm I looked at the old issue #2944 which also discussed the ramifications of changing when the hmatrix is initialized.  I think not modifying the current behavior is best for now (as in your last comment).  We could discuss that in the future if there are other problems it is causing.  Aside from that change is this PR ready to merge from your perspective?  If so, I will take a final look.

jtclemm:
> @jtclemm I looked at the old issue #2944 which also discussed the ramifications of changing when the hmatrix is initialized. I think not modifying the current behavior is best for now (as in your last comment). We could discuss that in the future if there are other problems it is causing. Aside from that change is this PR ready to merge from your perspective? If so, I will take a final look.

That's fair, we'll hold off on any such changes to the hmatrix. Reverted in the most recent commit.

This pull request should be good to go then.

sjplimp:
@jtclemm - all looks good - I made a few small changes
(1) Code - just added a couple explanatory comments for future development (e.g. new child classes)
(2) Doc pages - wanted to clarify how a user should think about these 2 commands, not so much the code structure (inheritance).
So I put the full explanation of the command syntax for both on the fix deform page (with some ONLY fix deform/pressure lines).
And the syntax for just the fix deform/pressure additions on its page.  And added a few sentences at the start of each page about how the 2 commands are used independently.   See if you like this OK, or think anything is now unclear ?

jtclemm:
I checked the commits and everything looks good to me, thanks for making those edits.

sjplimp (approved):

PR #4021, opened 2023-12-15
Bug fixes for BPM package

jtclemm:
**Summary**

This PR fixes a few misc bugs in the BPM package. 

This includes modifying displace_atoms and fix move so they  now rotate quaterions for the bpm/sphere atom style. Thanks to @jibril-b-coulibaly  for spotting the oversight.

**Author(s)**

Joel Clemmer (SNL)

**Licensing**

**Backward Compatibility**

NA


src/fix_move.cpp
```diff
@@ -1354,6 +1371,11 @@ void FixMove::set_arrays(int i)
         if (quat) {
           // qoriginal = f(quat,-delta);   // NOTE: edit this line
```
jibril-b-coulibaly, line 1372: Any insight on the original note to edit this line ?

jibril-b-coulibaly (approved):
Everything looks good, this is more rigorous to track atoms orientation for `BPM/sphere`.

Thinking again about the consequences, I am not sure leaving the quaternion unrotated as it was would generate artificial torques, since torque is based on rolling velocity, which does not depend on `quat` for spheres. Would you agree?

src/fix_move.cpp
jtclemm: Good point, I meant to check on this. I have no insight whatsoever, maybe @sjplimp would know? Looks like this line goes back quite a few years.

jtclemm:
> Everything looks good, this is more rigorous to track atoms orientation for `BPM/sphere`.
> 
> Thinking again about the consequences, I am not sure leaving the quaternion unrotated as it was would generate artificial torques, since torque is based on rolling velocity, which does not depend on `quat` for spheres. Would you agree?

Unlike the granular pair styles, `bond_style bpm/rotational` actually uses the quaternion to calculate forces. It doesn't accumulate a relative rotational history. 

I ran a test where I generated a BPM molecule with `bond_style bpm/rotational` and tried rotating it with both fix move and displace_atoms. Before this patch, rotating the grain introduced large stresses. No stresses are introduced when the quaternions are rotating with this patch.

However, come to think of it, wouldn't pair granular need to rotate all of its history elements to maintain frictional contacts?

jibril-b-coulibaly:
> > Everything looks good, this is more rigorous to track atoms orientation for `BPM/sphere`.
> > Thinking again about the consequences, I am not sure leaving the quaternion unrotated as it was would generate artificial torques, since torque is based on rolling velocity, which does not depend on `quat` for spheres. Would you agree?
> 
> Unlike the granular pair styles, `bond_style bpm/rotational` actually uses the quaternion to calculate forces. It doesn't accumulate a relative rotational history.
> 
> I ran a test where I generated a BPM molecule with `bond_style bpm/rotational` and tried rotating it with both fix move and displace_atoms. Before this patch, rotating the grain introduced large stresses. No stresses are introduced when the quaternions are rotating with this patch.
> 
> However, come to think of it, wouldn't pair granular need to rotate all of its history elements to maintain frictional contacts?

Thank you for the clarification regarding the role of quaternions in bpm/sphere: glad that this is fixing the artificial high stress issue!

If I recall correctly, at the beginning of each force calculation, the granular tangent history has its component in the normal contact direction removed (because the normal changed from t-1 to t and the t-1 history could now have non-zero normal component at t), and is re-scaled to its t-1 length to ensure it always points in a tangent direction with the correct magnitude.

There are discussions in the Luding 2008 paper whether this is good enough or if proper rotation should be carried out (the paper says Ideally, rotation). I think the current method would work well enough for small incremental changes, i.e., those expected from one timestep to the next.

For a large rotation from `displace_atom` / `fix move` however, this is a good point. Intuitively, I'd say that the difference between rotation and projection would become larger for larger changes in the normal direction.  There might also be some cancellation if the history vector is almost colinear with the new contact normal (~90 deg rotation).

This could be a safe change to make. Rotating granular history would only be valid if both atoms in contact are displaced/moved together though. If only one atom in the interaction is moved and they remain in contact, history would likely lose physical meaning. This is already the case, so if we make that change we'd be just as wrong as it already is now in that scenario: the user would have to make sure all interacting atoms are moved consistently.

jtclemm:
> For a large rotation from `displace_atom` / `fix move` however, this is a good point. Intuitively, I'd say that the difference between rotation and projection would become larger for larger changes in the normal direction. There might also be some cancellation if the history vector is almost colinear with the new contact normal (~90 deg rotation).
> 
> This could be a safe change to make. Rotating granular history would only be valid if both atoms in contact are displaced/moved together though. If only one atom in the interaction is moved and they remain in contact, history would likely lose physical meaning. This is already the case, so if we make that change we'd be just as wrong as it already is now in that scenario: the user would have to make sure all interacting atoms are moved consistently.

Yeah I think this is right. I don't see a usecase for rigidly rotating a packing of granular particles which are already in frictional contact. So probably nothing to worry about.

akohlmey:
@jtclemm please make sure that you assign this PR back to me when all open questions are resolved. @sjplimp is not available for a few days, so if you need his input, this will have to wait a bit.

jtclemm:
Hi @sjplimp, this should be ready to review when you have the time. Two main notes:

1) I ran into an odd comment in fix move: 
https://github.com/lammps/lammps/blob/1e9eb933beae8583f2e3e1b0caa73b3570720796/src/fix_move.cpp#L1412
Do you happen to recall what this referred to? I mirrored the comment in my changes.

2) To handle BPM simulations with multiple neighbor lists (e.g. pair hybrid or the addition of various computes), I updated fix update/special/bonds to loop through all `neighbor->nlist` lists in the neighbor class when a bond breaks to update its special bond bits. Previously, the fix only updated `force->pair->list`. Since this was one of the more complicated parts of the BPM package, I wanted to check whether you see any issues with this change.

sjplimp:
@jtclemm One general comment before answering your 2 Qs.

Re: this comment by @jibril-b-coulibaly -

>> Rotating granular history would only be valid if both atoms in contact are displaced/moved together though. If only one atom in
>> the interaction is moved and they remain in contact, history would likely lose physical meaning.

I think the common use of fix move with granular (or BPM?) would be to use it to move/rotate a set of boundary particles in a prescribed manner, e.g. a container, by small increments each timestep.  Does this mean that if the granular model has pairwise history, the history term between the a container and flow particle in contact will become invalid if fix move only moves the container particle?  I'm not clear how that is any different than the contact between 2 flow particles.  Fix move is just moving one of the particles instead of fix nve/sphere.

sjplimp:
Re: the 4 NOTE commented lines in fix_move.cpp.  These are in the set_arrays() method which is only called when a new particle is created, e.g. fix pour.  If that particle will be moved in the future by fix move, then it needs to have an xoriginal consistent with the time that fix move was specified.  Likewise if it has a quaternion, it needs a qoriginal consistent with the original time.  So those NOTEs are where that computation should take place.  Delta is the current elapsed time since the original time.  So current theta and quat would be reverse rotated by some function of Delta to produce toriginal and qoriginal.

It looks like the code for that has not been written.  But it's only for the use case of creating lines, tris, bodies, ellipsoids on-the-fly. It would be fine to flag that case as an error and say it is not yet supported.  So nothing to do with granular.  Are you saying for BPM you need to have that feature for its stored quat ?  If so, we could discuss further.

jtclemm:
> I think the common use of fix move with granular (or BPM?) would be to use it to move/rotate a set of boundary particles in a prescribed manner, e.g. a container, by small increments each timestep. Does this mean that if the granular model has pairwise history, the history term between the a container and flow particle in contact will become invalid if fix move only moves the container particle? I'm not clear how that is any different than the contact between 2 flow particles. Fix move is just moving one of the particles instead of fix nve/sphere.

If you incrementally move/rotate a rigid container, then yes, my understanding is that granular pair styles are not guaranteed to correctly update histories. There are two potential issues:
1) For interactions between a container+ flow particle, sliding/twisting/rolling histories are updated by integrating relative velocities. So the user would have to set consistent translational/angular velocities for each container particle (based on the global container motion) to ensure the history between the container + flow particle update correctly that timestep. This isn't an issue for the BPM package since it doesn't integrate relative velocities to figure out relative orientations.
2) For interactions between a container + container particle, the direction of the history will not rotate with the particles. This currently cannot be avoided. However, the container is likely a rigid body so the intra-container pair forces are irrelevant.

sjplimp:
Re: fix update/special/bonds - I don't recall the details of why this is needed or how it works.  Is there a reason that class is in src and not in the BPM package?  I don't see that it is used elsewhere.
Can you be more specific about your question on this class?

sjplimp:
>> So the user would have to set consistent translational/angular velocities for each 
>> container particle (based on the global container motion) 

If fix move is used to rotate the container particle(s), each is assigned a new position and velocity each timestep, consistent with the rotation:

        x[i][0] = point[0] + c[0] + disp[0];
        x[i][1] = point[1] + c[1] + disp[1];
        x[i][2] = point[2] + c[2] + disp[2];
        v[i][0] = omega_rotate * (runit[1] * disp[2] - runit[2] * disp[1]);
        v[i][1] = omega_rotate * (runit[2] * disp[0] - runit[0] * disp[2]);
        v[i][2] = omega_rotate * (runit[0] * disp[1] - runit[1] * disp[0]);

Are you saying that is not sufficient?

jtclemm:
> It looks like the code for that has not been written. But it's only for the use case of creating lines, tris, bodies, ellipsoids on-the-fly. It would be fine to flag that case as an error and say it is not yet supported. So nothing to do with granular. Are you saying for BPM you need to have that feature for its stored quat ? If so, we could discuss further.

I don't think this presents any additional complications for the BPM package vs. lines, tris, etc. So if this code is ever written, it can just be copied for the BPM atom style too.

jtclemm:
> If fix move is used to rotate the container particle(s), each is assigned a new position and velocity each timestep, consistent with the rotation:
> 
> 
> Are you saying that is not sufficient?

Ah, I missed that detail of fix move. So then issue 1 (container+fill interactions) is not a problem. Issue 2 (container+container interactions) would still be a problem though. But I can't see a reason why someone would rotate a non-rigid collection of granular particles.

sjplimp:
>> I can't see a reason why someone would rotate a non-rigid collection of granular particles.

ok good - I agree.  Fix move is meant to replace any other time integration fix for its group, so any forces on its particles are ignored.  Even if another integration fix did track container/container pairwise forces and attempt an additional update on the particles, fix move would just overwrite them with new positions and velocities.

jtclemm:
> Re: fix update/special/bonds - I don't recall the details of why this is needed or how it works. Is there a reason that class is in src and not in the BPM package? I don't see that it is used elsewhere. Can you be more specific about your question on this class?

There currently isn't a strict reason fix_update_special_bonds.cpp is in src instead of src/BPM. It was placed there since the fix could theoretically be used by other code in the future (e.g. one could update the peridynamics package to work with bonds, rewrite bond quartic to remove calls to pair->single(), or use it to offload the handling of special bonds in fixes that create/break bonds).

My only question was confirming the loop across neighbor lists was tolerable:
https://github.com/lammps/lammps/blob/ded160cd41653fb9a6d4835045d46279d1245124/src/fix_update_special_bonds.cpp#L167
I'm not aware of any other instances in LAMMPS where a fix/compute performs such an operation so I'm being cautious.

sjplimp:
@jtclemm I don't see an issue with that loop over neighbor lists.  If that is the list over all neighbor lists then
there are some flavors it probably doesn't make sense to loop over.  E.g. copy lists or skip lists.  Because other neigh lists
in the list already contain that info.  @akohlmey may have additional thoughts.

I think moving this fix into the BPM package would be better, to keep src uncluttered.  If another use case arises in the future, we can revisit.

jtclemm:
> @jtclemm I don't see an issue with that loop over neighbor lists. If that is the list over all neighbor lists then there are some flavors it probably doesn't make sense to loop over. E.g. copy lists or skip lists. Because other neigh lists in the list already contain that info. @akohlmey may have additional thoughts.

`NP_COPY` lists could definitely be skipped since updating the same list twice would do nothing (I think `if (list->copy) continue` would do the trick?). I think `NP_SKIP` lists still need to be updated since I understand they replicate data. However, selectively skipping unnecessary lists would only be a small performance gain (proportional to the number of broken bonds per timestep) so may not be worth the added complication. 

> I think moving this fix into the BPM package would be better, to keep src uncluttered. If another use case arises in the future, we can revisit.

Works for me. Should we also move fix_bond_history.cpp to src/BPM?

sjplimp:
I agree on NP_SKIP lists - I thought you were just scanning the lists and changing masks, not changing the lists.
Yes on fix_bond_history -> BPM.

Steve

sjplimp (approved):

PR #4023, opened 2023-12-17
Updated the Developer unittest section for debugging failed unit tests individually

ndtrung81:
**Summary**

Added a couple of paragraphs to the Developer guide for debugging failed unit tests based on Axel's instructions over Slack.

**Related Issue(s)**

N/A

**Author(s)**

Trung Nguyen (U Chicago)

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

N/A

### DIFF
--- a/doc/src/Build_development.rst
+++ b/doc/src/Build_development.rst
@@ -255,16 +255,18 @@ A test run is then a a collection multiple individual test runs each
 with many comparisons to reference results based on template input
 files, individual command settings, relative error margins, and
 reference data stored in a YAML format file with ``.yaml``
-suffix. Currently the programs ``test_pair_style``, ``test_bond_style``, and
-``test_angle_style`` are implemented.  They will compare forces, energies and
-(global) stress for all atoms after a ``run 0`` calculation and after a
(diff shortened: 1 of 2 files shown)


akohlmey:
Thanks @ndtrung81
I've added some corrections and some clarifications. There is additional information and explanations in the Build_development.rst file, so I added a link to that.

One could probably also point out that the failed assertions provide line numbers in the test source and that it can sometimes help to read the test source to understand what specific assertion failed. When using the -V flag to ctest, you can also see the exact command line of the test and that can be used, if prefixed with `gdb --args `, for debugging and catching exceptions. It used to be particularly difficult to figure out errors, if they happened during the LAMMPS commands issues by the tester program. But thanks to a recent change to the tester programs those exceptions are now caught (this change was possible since we are now *always* have the error classes throw exceptions) and then errors printed to stderr.

ndtrung81:
@akohlmey thanks very much for the corrections and clarification. I added the exact command line of the test and how to use gdb with the test command. Now it is clear how a test binary is invoked with what yaml file.

akohlmey (approved):

stanmoore1 (approved):
Looks good, thank you for adding this

PR #4018, opened 2023-12-14
Collected small fixes and updates

akohlmey:
**Summary**

This pull request combines multiple small fixes and updates that do not warrant a pull request of their own.

**Related Issue(s)**

Fixes #4016 
Closes #4024 
Fixes #4033 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

No known issues.

**Implementation Notes**

The following individual changes are included:
- move of "disk_free()" function from fix halt to platform namespace and port to Windows
- correct documentation of fix halt match the code
- update and clarify discussion of LAMMPS versions and releases
- modernize fix halt sources
- fix small omission in msi2lmp in warning message
- address some minor issues reported by coverity scan for compute slcsa/atom, compute pace, compute sna/atom
- remove unused class members and commented out debug code from compute pace
- silence some compiler warnings
- small enhancements signaled by clang-tidy
- let ReaxFF force field parser cleanly handle files without hydrogen bond parameters
- fix uninitialized data access bug in Grid2d and Grid3d classes
- skip python tests requiring numpy on macOS as those seem to fail randomly with GitHub actions with numpy installed (but not detected)
- check for using INTEL package kspace with run style `verlet/split`. If yes, error out with message that they are incompatible
- improve compatibility of openapi.cmake preset

**Post Submission Checklist**

- [x] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [x] The source code follows the LAMMPS formatting guidelines
- [x] Suitable new documentation files and/or updates to the existing docs are included
- [x] The added/updated documentation is integrated and tested with the documentation build system
- [ ] The feature has been verified to work with the conventional build system
- [x] The feature has been verified to work with the CMake based build system
- [ ] Suitable tests have been added to the unittest tree.

PR #4024, opened 2023-12-18
Doc: Explain better the release model

doronbehar:
**Summary**

As a distributor of LAMMPS, I had difficulty and confusion in understanding the releases' titles, Git tags' naming, and the Git branches. Even after reading the file changed in this PR it wasn't completely clear the reason behind the Git tags' naming. The main contribution of this PR IMO is the table added to the docs.

**Related Issue(s)**

None.

**Author(s)**

Me - email is available in the commit's patch.

**Licensing**

**Backward Compatibility**

Yes - it's documenting the release model.

**Implementation Notes**

**Post Submission Checklist**

None.

### DIFF
--- a/doc/src/Manual_version.rst
+++ b/doc/src/Manual_version.rst
@@ -10,34 +10,47 @@ Modifications of the LAMMPS source code (like bug fixes, code refactors,
 updates to existing features, or addition of new features) are organized
 into pull requests.  Pull requests will be merged into the *develop*
 branch of the git repository after they pass automated testing and code
-review by the LAMMPS developers.  When a sufficient number of changes
-have accumulated *and* the *develop* branch version passes an extended
-set of automated tests, we release it as a *feature release*, which are
-currently made every 4 to 8 weeks.  The *release* branch of the git
-repository is updated with every such release.  A summary of the most
-important changes of the patch releases are on `this website page
-<https://www.lammps.org/bug.html>`_.  More detailed release notes are
-`available on GitHub <https://github.com/lammps/lammps/releases/>`_.
+review by the LAMMPS developers.
 
-Once or twice a year, we have a "stabilization period" where we apply
-only bug fixes and small, non-intrusive changes to the *develop*
-branch.  At the same time, the code is subjected to more detailed and
-thorough manual testing than the default automated testing.  Also,
-several variants of static code analysis are run to improve the overall
-code quality, consistency, and compliance with programming standards,
-best practices and style conventions.
+_Feature_ V.S _Stable_ release channels
+'''''''''''''''''''''''''''''''''''''''
 
-The release after such a stabilization period is called a *stable*
-version and both, the *release* and the *stable* branches are updated
-with it.  Between stable releases, we collect back-ported bug fixes and
-updates from the *develop* branch in the *maintenance* branch.  From the
-*maintenance* branch we make occasional update releases and update the
-*stable* branch accordingly.
+LAMMPS has 2 release channels - _Feature_ and _Stable_. By default commits
+reach first the `develop` branch, and every certain amount of time we
+create a GitHub release, along with a Git tag for either of the channels.
+In parallel to the Git tags, we also maintain a Git branch that is
+identical of the latest Git tag of the release (either a _Feature_ or
+_Stable_). The table below lists the names
+
+Particularly for the _Stable_ channel, we use a branch named `maintenance`,
+to backport commits from the `develop` branch, that we want the stable
+channel to include. Hence the `maintenance` branch is a mediator between
+the main `develop` branch and the `stable` branch that is parallel to the
+latest tag of the _Stable_ channel.
+
++--------------+----------------+-----------------------------------+----------------------------------------------+------------------+
+| Release name | Git tag prefix | Git Branch parallel to latest tag | Git branches workflow                        | Release schedule |
++--------------+----------------+-----------------------------------+----------------------------------------------+------------------+
+| Feature      | ``patch_``     | ``release``                       | ``develop`` -> ``release``                   | Every 4-8 weeks  |
++--------------+----------------+-----------------------------------+----------------------------------------------+------------------+
+| Stable       | ``stable_``    | ``stable``                        | ``develop`` -> ``maintenance`` -> ``stable`` | 1-2 times a year |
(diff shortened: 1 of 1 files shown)


akohlmey:
> As a distributor of LAMMPS

What do you mean by "distributor". LAMMPS is *distributed* by the LAMMPS developers.

> I had difficulty and confusion in understanding the releases' titles, Git tags' naming, and the Git branches. Even after reading the file changed in this PR it wasn't completely clear the reason behind the Git tags' naming.

Before investing the effort of writing a pull request, it would have been beneficial to post an issue or a topic in the MatSci forum to explain *what* is the confusion and where you have problems.  Mostly the reasons are historic since the way releases were created has changed over the last decade and there are some inconsistencies due to the fact that not everything could be changed fully consistently in order to maintain backward compatibility with some services that depend on "historic behavior":

- The difficulties start with the (historic) choice to not use "semantic versioning" but indicate each version with a date.
- Initially, there were *only* patches and patch releases and these were frequent, sometimes even more than one per day (thus where the second would be using the next day's date as version)
- At some point in time, the decision was made to declare one of those patch releases (one without significant known bugs) to be a *stable* version (often about a month or so *after* its release) and this was done 3-4 times per year.
- At some point we changed the source code management to use git and have changes integrated via pull requests instead of a LAMMPS developer making the changes manually, commit them to an internal source code management system and then make patches and publish them
- There were 3 branches: master, unstable, stable. All PRs were merged with "master" and with patch releases the "unstable" branch was updated, and with stable release the "stable" branch (both with a fast-forward merge) so that people could choose to follow either branch and update their LAMMPS version with git pull instead of downloading a new tarball (the historic/traditional and previously *only* way of obtaining an updated LAMMPS version, short of downloading and applying the published patches).
- Using git and GitHub changed the workflow and patch releases were made less frequently but they were still called patch releases. Originally for some people doing packaging of LAMMPS we added tags to releases, so they could use automated tools to tell them when a new release was done (that predates using the "Releases" feature of GitHub).
- At some point GitHub made an effort to have branches called "master" renamed to something more politically correct and the LAMMPS developers realized at that point that also the naming conventions were no longer applicable: what used to be LAMMPS patches are now the "PR merges" and the "master" branch effectively had become the sequence of continuous release patches of the old times. So we decided to rename it "develop".
- Also around that time, we developed a large number of tests and set up automated and integration testing (on ci.lammps.org and through GitHub actions). This way, a large number of issues were resolved *before* even merging pull requests and thus the "unstable" branch name had become misnomer. If the patch releases had become the "develop" branch, the "unstable" branch was then renamed to "release" and we changed the name from "patch releases" to "feature releases". Thanks to the automated testing, those releases were at a higher quality level than previously the stable releases (there are additional, time consuming tests run *after* PR merges and in the last few days before a feature release only bugfix PRs were merged).
- The stable release were now done less frequent (currently the goal is once per year) and we now have a longer (2-3 months) of a stabilization period before such a stable release where "risky" changes are postponed and additional manual testing (e.g. running example inputs with valgrind's memcheck) is included. They needed to be distinguished with tags as well, so we have "stable" tags (the corresponding "patch" tags are for the systems that detect LAMMPS updates through them).
- At some point we became aware that *binary packaging* of LAMMPS for distributions and installing LAMMPS through systems like Homebrew, Spack, Gentoo, Arch, BSD ports had become significantly more important than distributing source code, but that people doing the packaging were hesitant to follow the "release" branch as the releases were too frequent for most of them and the required effort in testing. So the LAMMPS developers experimented with backporting bug fixes to the "stable" branch *without* adding new packages or new features. Previously the sentiment was: if somebody wants the bug fixed, just use the latest release version where the bugfix is included (or cherry-pick it from the develop branch). This lead to the need of a branch equivalent to "develop" but for "stable" and that explains the "maintenance" branch. This backporting is still a bit experimental, but it looks like it is a success and thus will stay for the foreseeable future.

This is where we currently are and we are aware that due to the history and changes in the release engineering things may look a bit confusing or inconsistent to a new person. For the most part, things remain the same since nobody has come up with a significant better way to describe it. Also, some choices cannot be changed because of the services that depend on them.

I have looked at your proposed changes and there is not a lot that can be improved because of the history I just outlined above. Some of your text is factually incorrect and needs to be corrected.

Please let me know how you want to proceed?
- you can try to rewrite your text based on review comments from me (and others)
- you can let me do a rewrite of the corresponding segment in the manual that tries to incorporate improvements from your PR (e.g. the term "release channels", although that is something that may be more familiar to people using Windows than other platforms) and then you provide your comments to me

doronbehar:
Thanks for the thorough historical overview!

> What do you mean by "distributor". LAMMPS is _distributed_ by the LAMMPS developers.

I consider myself a distributor because I am involved in the maintenance of [Nixpkgs](https://github.com/NixOS/nixpkgs) distribution, which distributes the Nix package of LAMMPS. I presume you refer to this role as "Packager".

> Before investing the effort of writing a pull request, it would have been beneficial to post an issue or a topic in the MatSci forum to explain _what_ is the confusion and where you have problems.

As the maintainer of LAMMPS in the Nix distribution, I wasn't sure what Git tags to follow - those with the `stable_` or `patch_` prefix. I didn't understand the meaning of the `patch_` prefix in the Git tags of the "Feature release", and I also didn't understand what does "feature" mean in that context. I had to go as far as to read the output of this command to get a sense of the rules of this release model:

```
curl --silent https://api.github.com/repos/lammps/lammps/releases | jq --raw-output '.[] | [.tag_name, .name] | @tsv' | awk -F$'\t' '$2 ~ /Feature/ {print $0}'
```

I was thinking of asking about it here or on the forum but then I encountered the documentation, and I reached the conclusion that it is _possible_ to understand the release model from it, but it is still hard, and the situation could be improved.

> I have looked at your proposed changes and there is not a lot that can be improved because of the history I just outlined above. Some of your text is factually incorrect and needs to be corrected.
> 
> Please let me know how you want to proceed?

I'm not surprised that some of the details I mentioned could be improved. I'm open to reviews :) The only thing I will probably insist upon sort-of, is the table - which explains the important things in a concise manner.

akohlmey:
I have done a rewrite and update of the file in question: https://github.com/lammps/lammps/pull/4018/files#diff-107fbd60208757c42806ad940bb714be1cf1d4b74af8a348362496d293676f25L1
After the reorganization adding the table seems largely redundant.

PR #4037, opened 2024-01-10
Make computation of PPPM grid counts and DOFs removed by fixes compatible with large systems

akohlmey:
**Summary**

The `Fix::dof()` function currently returns a signed 32-bit integer. This causes overflows for large systems with many small molecules (e.g. large bulk water systems) when LAMMPS is compiled with `-D LAMMPS_BIGBIG`. This pull request changes the return type and the individual dof computations to use a `bigint` instead of an `int`.

Furthermore, in a few places in the various PPPM variants, the number of total grid points (nx_pppm * ny_pppm * nz_pppm) is used. This can also overflow for very large systems and the correspondingly large grids. This is converted to use bigint instead of int (explicitly and implicitly).

A check was added to the replicate command to avoid overflowing the total number of replica. The command needs to be used multiple times smaller numbers in that case.

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U and Stan Moore, SNL

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

This only updates the (removed) DOF computation. There may be other limits in the affected code that will still limit the use for larger systems (e.g. `nbody` being an `int` in fix rigid).
Also the computation and output of SHAKE cluster stats is updated to use bigint instead of int.

**Post Submission Checklist**


src/KSPACE/pppm.cpp
```diff
@@ -1944,7 +1944,8 @@ void PPPM::poisson_ik()
 
   // global energy and virial contribution
 
-  double scaleinv = 1.0/(nx_pppm*ny_pppm*nz_pppm);
+  bigint ngridtotal = (bigint) nx_pppm * ny_pppm * nz_pppm;
```
hagertnl, line 1947: Is it possible that nx_pppm, ny_pppm, and nz_pppm may need to also be bigint's, or do they have a limit within the bounds of 32-bit ints? If you cast those variables as bigint's in `kspace.h`, then the multiplication is implicitly type `bigint`, which also avoids the overflow.

akohlmey: The grid counts per direction must be within the range of 32-bit signed ints.

stanmoore1: Even making the product of 2 dimensions a 64-bit integer, e.g. `bigint nxy_pppm = (bigint) nx_pppm * ny_pppm;` may be excessive, since `sqrt(2^31) = 46,341`, which is an absolutely huge grid for PPPM: 99 *trillion* grid points in 3d). But it doesn't really hurt anything and makes the code more future proof.

hagertnl:
After the `develop` branch was merged into this branch, I get a compiler error from `delete_bonds.cpp` because the `fix_bond_history.h` file is no longer in `src`, it's in `src/BPM`. The commit is here: https://github.com/akohlmey/lammps/commit/f6649762e2d2707575f6ac0cff9a0b31f6ac90aa#diff-59169b28dc5ecb09e009254dbde85a22810e1e2860f4694ad4d9415a9ba253c9

akohlmey:
> After the `develop` branch was merged into this branch, I get a compiler error from `delete_bonds.cpp` because the `fix_bond_history.h` file is no longer in `src`, it's in `src/BPM`. The commit is here: [akohlmey@f664976#diff-59169b28dc5ecb09e009254dbde85a22810e1e2860f4694ad4d9415a9ba253c9](https://github.com/akohlmey/lammps/commit/f6649762e2d2707575f6ac0cff9a0b31f6ac90aa#diff-59169b28dc5ecb09e009254dbde85a22810e1e2860f4694ad4d9415a9ba253c9)

Ok. Will adjust for this in PR #4039. Workaround is to install the BPM package.

akohlmey:
@hagertnl for your convenience, I also cherry picked the delete_bonds/BPM issue commit into this branch.

hagertnl:
Thank you! I tested our 3.2b atom system with `fix shake` turned on and was able to verify that the fix shake stats output overflow is fixed:
```
Finding SHAKE clusters ...
       0 = # of size 2 clusters
       0 = # of size 3 clusters
       0 = # of size 4 clusters
1073741824 = # of frozen angles
```
And the temperature at time 0 looks perfect -- 297.77242 K.
Unfortunately, shortly after step 30 (if we extend from the 20 steps we were doing for demonstration purposes), we're now getting a repeatable memory fault (AMD GPU reports this as `HSA_STATUS_ERROR_MEMORY_APERTURE_VIOLATION: The agent attempted to access memory beyond the largest legal address.`). My core files are being dumped incompletely, for some reason, so I haven't been able to get a good backtrace on where this fault is happening yet.

stanmoore1:
@hagertnl can you try running with Kokkos bounds checking? How many nodes are you using?

hagertnl:
@stanmoore1 I'm on 64 nodes, 8 ranks per node.
I am heading out for the weekend now, but I will try adding `-DKOKKOS_ENABLE_DEBUG_BOUNDS_CHECK` to the compilation flags on Tuesday and seeing what comes of it, thank you!

I had a CPU build ready to go, so I sent that in the queue to run over the weekend. That would possibly help narrow down Kokkos as the problem, or hopefully give us a better core file to investigate.

hagertnl:
Checked in on the CPU run briefly and that job completed with no issues. Will follow up on the GPU side on Tuesday. Thanks!

jtclemm (approved):

hagertnl:
@stanmoore1 I added`-DKOKKOS_ENABLE_DEBUG_BOUNDS_CHECK ` and re-built and got the following message shortly after step 35 was printed in a 3.2b atom system at 64 nodes:
```
:0: : Device-side assertion `View bounds error' failed.
```

stanmoore1:
@hagertnl thanks, the next step is to compile with `-G` (not `-lineinfo`) to get a core dump in device code. Then it should hopefully be straightforward to debug with the exact line number. Alternatively you could compile with `-ftrapv` and see if any integer overflow is causing the out of bounds.

stanmoore1:
Actually I have NVIDIA on my mind, but you are on AMD, sorry... I'm less familiar debugging AMD GPUs, @IanBogle any suggestions?

hagertnl:
Happened to already have `-g` enabled. The source code is the latest in `develop` branch. Here's the backtrace. Will also try `-ftrapv` to see if it catches something.
```
#0  0x00007fe0c27d2460 in ?? () from /opt/rocm-5.5.1/hsa/lib/libhsa-runtime64.so.1
#1  0x00007fe0c27d232a in ?? () from /opt/rocm-5.5.1/hsa/lib/libhsa-runtime64.so.1
#2  0x00007fe0c27c60a9 in ?? () from /opt/rocm-5.5.1/hsa/lib/libhsa-runtime64.so.1
#3  0x00007fe0c534976b in ?? () from /opt/rocm-5.5.1/lib/libamdhip64.so.5
#4  0x00007fe0c5338dd2 in ?? () from /opt/rocm-5.5.1/lib/libamdhip64.so.5
#5  0x00007fe0c5294613 in ?? () from /opt/rocm-5.5.1/lib/libamdhip64.so.5
#6  0x00007fe0c52949c2 in hipStreamSynchronize () from /opt/rocm-5.5.1/lib/libamdhip64.so.5
#7  0x0000000018cc741f in Kokkos::Impl::HIPInternal::fence(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) const::{lambda()#1}::operator()() const (this=<optimized out>) at ../../lib/kokkos/core/src/HIP/Kokkos_HIP_Instance.cpp:159
#8  Kokkos::Tools::Experimental::Impl::profile_fence_event<Kokkos::HIP, Kokkos::Impl::HIPInternal::fence(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) const::{lambda()#1}>(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, Kokkos::Tools::Experimental::Impl::DirectFenceIDHandle, Kokkos::Impl::HIPInternal::fence(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) const::{lambda()#1} const&) (devIDTag=..., name=...,
    func=...) at ../../lib/kokkos/core/src/impl/Kokkos_Profiling.hpp:186
#9  Kokkos::Impl::HIPInternal::fence (this=0x197c8300, name=...) at ../../lib/kokkos/core/src/HIP/Kokkos_HIP_Instance.cpp:155
#10 0x00000000184976f2 in Kokkos::Impl::HIPParallelLaunch<Kokkos::Impl::ParallelFor<LAMMPS_NS::FixShakeKokkos<Kokkos::HIP>, Kokkos::RangePolicy<Kokkos::HIP, LAMMPS_NS::TagFixShakeUnpackExchange>, Kokkos::HIP>, Kokkos::LaunchBounds<256u, 1u>, (Kokkos::Impl::HIPLaunchMechanism)1>::HIPParallelLaunch (this=this@entry=0x7ffe9d25ed40, driver=..., grid=...,
    block=..., shmem=shmem@entry=0, hip_instance=0x197c8300) at ../../lib/kokkos/core/src/HIP/Kokkos_HIP_KernelLaunch.hpp:570
#11 0x0000000018497321 in Kokkos::Impl::hip_parallel_launch<Kokkos::Impl::ParallelFor<LAMMPS_NS::FixShakeKokkos<Kokkos::HIP>, Kokkos::RangePolicy<Kokkos::HIP, LAMMPS_NS::TagFixShakeUnpackExchange>, Kokkos::HIP>, Kokkos::LaunchBounds<0u, 0u>, (Kokkos::Impl::HIPLaunchMechanism)1, false> (driver=..., grid=..., block=..., shmem=0,
    hip_instance=0x7fe0c28ccfa0, prefer_shmem=false) at ../../lib/kokkos/core/src/HIP/Kokkos_HIP_KernelLaunch.hpp:613
#12 Kokkos::Impl::ParallelFor<LAMMPS_NS::FixShakeKokkos<Kokkos::HIP>, Kokkos::RangePolicy<Kokkos::HIP, LAMMPS_NS::TagFixShakeUnpackExchange>, Kokkos::HIP>::execute (
    this=this@entry=0x7ffe9d25ede0) at ../../lib/kokkos/core/src/HIP/Kokkos_HIP_Parallel_Range.hpp:93
#13 0x000000001849704e in Kokkos::parallel_for<Kokkos::RangePolicy<Kokkos::HIP, LAMMPS_NS::TagFixShakeUnpackExchange>, LAMMPS_NS::FixShakeKokkos<Kokkos::HIP>, void> (str=...,
    policy=..., functor=...) at ../../lib/kokkos/core/src/Kokkos_Parallel.hpp:144
#14 0x0000000018471f75 in Kokkos::parallel_for<Kokkos::RangePolicy<Kokkos::HIP, LAMMPS_NS::TagFixShakeUnpackExchange>, LAMMPS_NS::FixShakeKokkos<Kokkos::HIP> > (policy=...,
    functor=...) at ../../lib/kokkos/core/src/Kokkos_Parallel.hpp:153
#15 LAMMPS_NS::FixShakeKokkos<Kokkos::HIP>::unpack_exchange_kokkos (this=0x199e3530, k_buf=..., k_indices=..., nrecv=<optimized out>) at ../fix_shake_kokkos.cpp:1758
#16 0x0000000018a11173 in LAMMPS_NS::CommKokkos::exchange_device<Kokkos::HIP> (this=0x197bd2d0) at ../comm_kokkos.cpp:932
#17 0x000000001740852c in LAMMPS_NS::VerletKokkos::run (this=0x19e8f9c0, n=100) at ../verlet_kokkos.cpp:341
#18 0x000000001757ec77 in LAMMPS_NS::Run::command (this=0x19b24ff0, narg=<optimized out>, arg=<optimized out>) at ../run.cpp:176
#19 0x000000001750ebf4 in LAMMPS_NS::Input::execute_command (this=this@entry=0x197f72e0) at ../input.cpp:868
#20 0x000000001750d38b in LAMMPS_NS::Input::file (this=0x197f72e0) at ../input.cpp:313
#21 0x000000001723dd9d in main (argc=<optimized out>, argv=<optimized out>) at ../main.cpp:77
```

stanmoore1:
Ah that is helpful, I will look at the code

IanBogle:
`-g` will get you a meaningful backtrace, but if you want to run with `rocgdb`, you will also need `-ggdb` for the most detailed information. Also, for debugging make sure the build type is RelWithDebInfo or similar. That's my more general advice, but if specific questions come up here I can do my best to track down an answer.

hagertnl:
`-ftrapv` did not catch any overflows. I'll try `-ggdb` to add some more detailed info for `rocgdb`. @IanBogle I did have one specific question -- do you know if `-ftrapv` also applies to integer overflows on the device, or does it only apply to host memory?

IanBogle:
`-ftrapv` only applies to host memory. Something that might get even better info in this instance: in `rocgdb`, before starting the run, try doing `set amdgpu precise-memory on`. That will make sure `rocgdb` points you to exactly where a memory fault occurs.

stanmoore1:
@IanBogle the stack trace above seems to be from the CPU--the last line in LAMMPS code is `../fix_shake_kokkos.cpp:1758` which is the kernel launch point. Is there a way to get a core dump from inside device code? On NVIDIA this requires `-G`...

hagertnl:
There are other threads in the corefile that were on the device, but since I didn't have `-ggdb`, they were just `??`. I'm re-running now with the better debugging enabled.

IanBogle:
It could be that `-ggdb` allows the default backtrace to show device code, but I don't know that offhand. I know that running a binary built with `-g` and `-ggdb` through `rocgdb` should give you a device backtrace as well.

stanmoore1:
@hagertnl another thing you could try is running the Kokkos `Serial` backend on the host CPU, similar to your (non-Kokkos) CPU only test case that ran successfully. In that case `-ftrapv` should be much more useful.

stanmoore1:
Actually that was without `fix shake`, rerunning.

hagertnl:
The CPU run of the problem is going to take quite a while, so I'm going to send a longer job overnight, I'll report a stack trace later tonight/tomorrow. Thanks for the help on this!

stanmoore1:
This doesn't crash if I use `-pk kokkos comm/exchange no`. This is consistent with the stack trace too, so it seems to narrow down the bug.

hagertnl:
As you noted, appears to be related to the Kokkos comms, so it may come as little surprise that the Serial Kokkos run was successful.

Related to comms, it may be relevant that even entirely on the host, `-ftrapv` does not catch overflows in third party library calls like `MPI_Reduce`, so there's a possibly common source of integer overflow that may not be caught there.

hagertnl:
I ran a Serial backend (compile-time KOKKOS debugging not enabled, since it wasn't giving a core file when it crashed) with `comm host`, which can reproduce the crash. I got the following stack trace from the segmentation fault that resulted:
```
(gdb) bt
#0  LAMMPS_NS::FixShakeKokkos<Kokkos::Serial>::operator() (this=this@entry=0x7fffffff3f70, i=@0x7fffffff3f30: 483) at ../fix_shake_kokkos.cpp:1707
#1  0x0000000000f96ac2 in Kokkos::Impl::ParallelFor<LAMMPS_NS::FixShakeKokkos<Kokkos::Serial>, Kokkos::RangePolicy<Kokkos::Serial, LAMMPS_NS::TagFixShakeUnpackExchange>, Kokkos::Serial>::exec<LAMMPS_NS::TagFixShakeUnpackExchange> (
    this=0x7fffffff3f70) at ../../lib/kokkos/core/src/Serial/Kokkos_Serial_Parallel_Range.hpp:46
#2  Kokkos::Impl::ParallelFor<LAMMPS_NS::FixShakeKokkos<Kokkos::Serial>, Kokkos::RangePolicy<Kokkos::Serial, LAMMPS_NS::TagFixShakeUnpackExchange>, Kokkos::Serial>::execute (this=0x7fffffff3f70)
    at ../../lib/kokkos/core/src/Serial/Kokkos_Serial_Parallel_Range.hpp:52
#3  Kokkos::parallel_for<Kokkos::RangePolicy<Kokkos::Serial, LAMMPS_NS::TagFixShakeUnpackExchange>, LAMMPS_NS::FixShakeKokkos<Kokkos::Serial>, void> (str=..., policy=..., functor=...) at ../../lib/kokkos/core/src/Kokkos_Parallel.hpp:144
#4  0x0000000000f7715d in Kokkos::parallel_for<Kokkos::RangePolicy<Kokkos::Serial, LAMMPS_NS::TagFixShakeUnpackExchange>, LAMMPS_NS::FixShakeKokkos<Kokkos::Serial> > (policy=..., functor=...)
    at ../../lib/kokkos/core/src/Kokkos_Parallel.hpp:153
#5  LAMMPS_NS::FixShakeKokkos<Kokkos::Serial>::unpack_exchange_kokkos (this=0x196d2d0, k_buf=..., k_indices=..., nrecv=960) at ../fix_shake_kokkos.cpp:1758
#6  0x000000000143abed in LAMMPS_NS::CommKokkos::exchange_device<Kokkos::Serial> (this=0x183aa50) at ../comm_kokkos.cpp:932
#7  0x000000000064b362 in LAMMPS_NS::VerletKokkos::run (this=0x17b28a0, n=100) at ../verlet_kokkos.cpp:341
#8  0x00000000007cbf7a in LAMMPS_NS::Run::command (this=0x18649b0, narg=<optimized out>, arg=<optimized out>) at ../run.cpp:176
#9  0x0000000000748abc in LAMMPS_NS::Input::execute_command (this=this@entry=0x18397d0) at ../input.cpp:868
#10 0x0000000000746fdb in LAMMPS_NS::Input::file (this=0x18397d0) at ../input.cpp:313
#11 0x00000000004d637d in main (argc=<optimized out>, argv=<optimized out>) at ../main.cpp:77
```
This is very close to the stack trace we had from the GPU anyway, so I'm not sure if this is helpful or not.

stanmoore1:
@hagertnl thanks. I'm seeing a crash with `/bench/in.rhodo` on 4 ranks with `processors 1 1 4` and 32k atoms (reported by @weinbe2). I'm pretty sure this is the same bug, debugging now. Should be a lot easier than 3.2 billion atoms on 512 ranks!

stanmoore1:
@hagertnl can you try out #4047, I think it will fix your 3.2 billion atom crash too.

PR #4039, opened 2024-01-12

akohlmey:
**Summary**

This pull request contains multiple small changes and fixes that do not warrant a pull request of their own.

**Related Issue(s)**

Fixes the bug reported here: https://matsci.org/t/bug-in-angle-cosine-periodic-from-extra-molecule/52963
Fixes the issue reported here: https://github.com/lammps/lammps/pull/4037#issuecomment-1889939953

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- bugfix for angle style cosine/periodic with multiplicity 1
- update unit test for angle style cosine/periodic to include multiplicity 1 in angle_coeff settings
- BPM package must set a define so that `delete_bonds` only accesses fix bond/history data when BPM is included
- similar changes are required for fix `bond/break`, `fix bond/swap`, and `fix bond/react`
- fix bug in fix update/special/bonds
- fix bug in pair style slater/coul/long/gpu (from @ndtrung81)
- fix typo in OpenMP neighbor lists class for r-RESPA
- improved error messages
- remove a few obsolete machine makefiles
- whitespace fixes

**Post Submission Checklist**


src/delete_bonds.cpp
jtclemm, line 1: Wow, totally overlooked this. Similar changes will be needed to various fixes in the MC package...

akohlmey: ... and to fix bond/react.

jrgissing (approved):

src/delete_bonds.cpp
jtclemm: Thank you for fixing those errors.

jtclemm (approved):

src/delete_bonds.cpp
akohlmey: No worries. This is what I do and these kind of straightforward fixes are welcome after struggling with a bunch or hard to pin down issue.
Also, it is not just your fault. This issue should have been detected by integration testing, but only the GPU unit tests (which are only run *after* a merge or when requested explicitly) did not include the BPM package and thus failed.

sjplimp (approved):

sjplimp:
@akohlmey Do you like the solution of adding #ifdefs to the MC fixes better than just moving fix_bond_history.cpp back into src, rather than the BPM package, where it was before ?

akohlmey:
> @akohlmey Do you like the solution of adding #ifdefs to the MC fixes better than just moving fix_bond_history.cpp back into src, rather than the BPM package, where it was before ?

I am a bit conflicted about this one. Moving the fix back would make things simpler, but the ifdefs indicate that this is a feature related to BPM (only). That is why I was waiting for your review since there was a possibility that you would object.

akohlmey:
@jtclemm please have a look at commits e3f81a25130d80fcf436f25be3d46519ec2809ce e82ba61183d7f6b8faa63fc52cfc04675fe4c1bd and let me know if I have guessed the correct changes. These were spotted through checking out compiler warnings issued by Clang 17 and static code analysis reports by Coverity Scan (https://scan.coverity.com/).

sjplimp:
@akohlmey I think moving fix_bond_history back into src is probably best, treating it as a utility class which can be used by multiple commands, similar to fix_neigh_history.  Joel and I thought it was only used by BPM but the delete_bonds command is another user.  Basically any model that wants to store time history values for a pairwise or bond interaction could use these 2 classes.

akohlmey:
> I think moving fix_bond_history back into src is probably best

Like this?

sjplimp:
yes - thanks

jtclemm:
> @jtclemm please have a look at commits [e3f81a2](https://github.com/lammps/lammps/commit/e3f81a25130d80fcf436f25be3d46519ec2809ce) [e82ba61](https://github.com/lammps/lammps/commit/e82ba61183d7f6b8faa63fc52cfc04675fe4c1bd) and let me know if I have guessed the correct changes. These were spotted through checking out compiler warnings issued by Clang 17 and static code analysis reports by Coverity Scan

These changes look correct. Thanks again for fixing the mistakes!

PR #4047, opened 2024-01-17
Fix bug in some Kokkos fixes' unpack exchange on device

stanmoore1:
**Summary**

Fix bug in some Kokkos fixes' unpack exchange on device when a proc received data from more than 1 other proc in a given direction. Kokkos versions of `fix shake`, `fix neigh/history`, and `fix spring/self` were affected, while Kokkos versions of `fix wall/gran` and `fix qeq/reaxff` were correct and not affected.

The code in comm exchange concatenated buffers from 2 neighbor procs into a single buffer, see https://github.com/lammps/lammps/blob/develop/src/KOKKOS/comm_kokkos.cpp#L923-L925. However the code in `fix shake` and others appended data to front of the buffer in the pack and assumed that the buffer was only from a single proc when unpacking the buffer. Adding an additional offset in the unpack for the data from the second proc fixes the issue.

**Related Issue(s)**

#4037 

**Author(s)**

Stan Moore (SNL), reported by Evan Weinberg (NVIDIA) and Nick Hagerty (ORNL)

**Licensing**

**Backward Compatibility**

Yes

### DIFF
--- a/src/KOKKOS/comm_kokkos.cpp
+++ b/src/KOKKOS/comm_kokkos.cpp
@@ -931,6 +931,7 @@ void CommKokkos::exchange_device()
             if (nextrarecv) {
               kkbase->unpack_exchange_kokkos(
                 k_buf_recv,k_indices,nrecv/data_size,
+                nrecv1/data_size,nextrarecv1,
                 ExecutionSpaceFromDevice<DeviceType>::space);
               DeviceType().fence();
             }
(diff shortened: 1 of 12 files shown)


stanmoore1:

akohlmey:
FYI: backported to maintenance branch in commit 20dae33563cb2b94e0af81b15e6c101e0537d6a9

hagertnl:
@stanmoore1 This appears to have fixed my crash as well.

akohlmey (approved):

PR #4048, opened 2024-01-18
Update LEPTON styles for more flexibility

akohlmey:
**Summary**

This adds support to handle Lepton expressions resulting in constant forces and constant potential functions.
For bond/angle styles lepton an optional keyword is added to select whether the potential is shifted to be zero at r_0 / theta_0 or taken as is. Previously the potential energy was always shifted.

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

Binary restart files are not compatible with those for older versions of LAMMPS due to storing the state of the "auto_offset" vs "no_offset" keyword. Regular inputs remain backward compatible since the keyword is optional and the default setting remains unchanged.

**Implementation Notes**

Previously LAMMPS would throw an error when using a potential function that is missing the variable designated to be varied for the given potential due to the Lepton library throwing an error when trying to get a reference to that variable. This also applies to potential functions where the derivative is constant. This specific kind of exception is now caught or the expression evaluated without setting the variable reference.

**Post Submission Checklist**


sjplimp (approved):

doc/src/angle_lepton.rst
```diff
@@ -41,6 +51,13 @@ angle coefficient.  For example `"200.0*theta^2"` represents a
 
    U_{angle,i} = K (\theta_i - \theta_0)^2 = K \theta^2 \qquad \theta = \theta_i - \theta_0
 
+.. versionchanged:: TBD
+
+By default the potential energy U is shifted so that he value U is 0.0
```
jtclemm: Typo: the value U

jtclemm (approved):

doc/src/angle_lepton.rst
jtclemm: Also, should this be lepton/sphere?
https://github.com/lammps/lammps/blob/2d8f622c6af9aa3ab5733fdfb44e852898631e8d/doc/src/pair_lepton.rst?plain=1#L75-L76

akohlmey: Thanks for spotting these.

PR #4054, opened 2024-01-19
Remove sphere_flag and replace with radius_flag and omega_flag as needed

akohlmey:
**Summary**

The test for `Atom::sphere_flag` and the corresponding error messages are in some cases quite confusing, especially since sphere_flag is set for atom styles that may not be spherical at all. Rather it is more consistent to check for either `Atom::radius_flag` or `Atom::omega_flag` or both.

This pull request removes sphere_flag and makes the corresponding changes, also for more consistent error messages and are more along the lines of other messages listing the missing atom attributes instead of a specific pair style.
In case of the atom style bpm/sphere actually the omega attribute was not checked for and that is now added, too.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

All unit tests pass.

**Implementation Notes**

There may other flags requiring similar scrutiny.

**Post Submission Checklist**


jtclemm:
Many of the modified fix/pair/bond styles require the torque attribute but don't check that it exists. This isn't a problem for the majority of them since they check the `atom->omega_flag` which (currently) implies torques are defined. 

I believe the exceptions (cases that don't check `atom->omega_flag` or `atom->torque_flag` but require torques) are:

- src/BROWNIAN/fix_brownian_sphere.cpp
- src/COLLOID/pair_brownian.cpp
- src/COLLOID/pair_brownian_poly.cpp

However, maybe all should check `atom->torque_flag` to be thorough?

akohlmey:
> However, maybe all should check `atom->torque_flag` to be thorough?

I am learning, again, the deeper truth behind the saying that my phd adviser would use a lot: the closer you look, the worse it gets.

To make this all consistent, quite a bit of effort would be required. In particular:
- update the table of per-atom attributes in https://docs.lammps.org/atom_style.html to list attribute names that are consistent with the error messages and complete
- decide on consistent attribute names also for the checks and error messages for example
   - charge or q?
   - dipole or mu?
   - angular velocity or omega?
   - mass or rmass?
   - quaternion or quat?
   - diameter or radius?
   - others?
   What is rather confusing about these are that the names of the internal flags are usually different from the name used for the properties.
- decide on whether to check for all required attributes in a single statement with a single error message or test for each attribute independently.

Since this kind of change has significant impact on external LAMMPS packages it is probably not a good idea to change the internal names (unless we want to use these kinds of changes as leverage to annoy people into contributing their code to the LAMMPS distribution. however, the people suffering the most are usually *other* users that would be forced into using an outdated LAMMPS version because of the incompatibility and their inability or unwillingness to make the necessary updates ). Perhaps we can share the workload by distributing which attributes to make consistent or which subset of files to look at.

@sjplimp what do you think about this whole topic as it is mushrooming much faster than I anticipated?

sjplimp:
3 comments:

(1) Perspective - I don't think this topic is important to get "perfect".  The goal is simply to flag when a class uses a per-atom property not defined by the current atom style(s).  Else the code will crash.  I can't recall a recent time that kind of bug has been reported.  So we're probably flagging 99% of the cases already.  And if such a bug is reported, it's easy to fix.

(2) I agree that making these checks for individual atom properties (omega) rather than categories of properties (sphere) is the right approach.  Over time as atom styles have been added the distinction between categories has blurred.  I also agree that making the atom_style page table consistent with the error messages is good.  I suggest using the code variable (omega) in the table, with a footnote that gives a fuller explanation for the user (angular velocity).

(3) There are probably error checks in the code where properties fooA and fooB are used by a class and only fooA is error-checked for.  B/c fooA and fooB were originally always in the same atom style.  That could be faulty logic if fooA is now
used by a 2nd atom style which does not define fooB. 

Probably easier to discuss this further in a phone call.

akohlmey:
@sjplimp this has now the updated atom_style documentation.

sjplimp:
@akohlmey I like the new table in the atom_style doc page.  I made some minor edits to that page for clarity.  The biggest change was to alphabetize the section with Additional Info on individual atom styles.

sjplimp:
@akohlmey I noticed there are some flagged spelling errors on the atom_style doc page - don't know how to turn them off.

sjplimp (approved):

akohlmey:
> @akohlmey I noticed there are some flagged spelling errors on the atom_style doc page - don't know how to turn them off.

These are false positives and those need to be added (in alphabetical order) to `doc/utils/sphinx-config/false_positives.txt`.
I have just done that.

PR #4058, opened 2024-01-24
Add function to library interfaces to set internal style variable values

akohlmey:
**Summary**

This adds a function `lammps_set_internal_variable()` to the LAMMPS C-library interface to give access to the `Variable::internal_set()` function. That way it becomes possible to change the value of a LAMMPS variable directly. For consistency, the `lammps_set_variable()` function was renamed to `lammps_set_string_variable()` while retaining a backward compatibility wrapper that prints a one time deprecation warning.

The changes are propagated to all functionality that sits on top of the C-library interface: Fortran module, Python module, Plugin library loader, Swig wrappers.

Also unit tests are included.

**Related Issue(s)**

Implements a feature asked for on MatSci: https://matsci.org/t/fortran-constants-variables-as-lammps-variables/53165

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issue. Calling the function `lammps_set_variable()` will print a deprecation warning as it will be removed in the future.

**Implementation Notes**

N/A

**Post Submission Checklist**

### DIFF
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -2460,8 +2460,13 @@ static int set_variable_deprecated_flag = 1;
 .. deprecated:: TBD
 
 This function assigns a new value from the string str to the
-string-style variable *name*. Returns -1 if a variable of that
-name does not exist or is not a string-style variable, otherwise 0.
+string-style variable *name*.  This is a way to directly change the
+string value of a LAMMPS variable that was previous defined with a
+:doc:`variable name string <variable>` command without using any
+LAMMPS commands to delete and redefine the variable.
+
+Returns -1 if a variable of that name does not exist or if it is not
+a string-style variable, otherwise 0.
 
 .. warning::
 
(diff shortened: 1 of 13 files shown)


src/library.cpp
```diff
@@ -2477,6 +2517,35 @@ int lammps_set_variable(void *handle, char *name, char *str)
   return err;
 }
 
+/* ---------------------------------------------------------------------- */
+
+/** Set the value of an internal-style variable.
+ *
+ * This function assigns a new value value to an internal-style variable.
```
sjplimp: Please add a line of context.  E;g. This is the way to set a variable in LAMMPS which stores a numeric value by passing it a numeric value.

src/library.cpp
```diff
@@ -2451,19 +2451,59 @@ int lammps_extract_variable_datatype(void *handle, const char *name)
...
+}
+
+/* ---------------------------------------------------------------------- */
+
+/** Set the value of a string-style variable.
+\verbatim embed:rst
+
+.. versionadded:: TBD
+
+This function assigns a new value from the string str to the
```
sjplimp, line 2500: Please add a line of context. E;g. This is the way to set a variable in LAMMPS which stores a string by passing it a string.

sjplimp (changes requested):
Just want to add some clarification for lib users.

akohlmey:
> Just want to add some clarification for lib users.

done

sjplimp (approved):

Issue #3999, opened 2023-11-29
Refactor bonus struct for ellipsoid atom style


jibril-b-coulibaly:
**Summary**

I was looking into the ellipsoid atom style with the goal to implement granular contact between ellipsoids. I would like to better understand the rationale for the current `bonus` structure, which constains `shape` and `quat`, and discuss possible refactoring.

**Detailed Description**

The [documentation](https://docs.lammps.org/atom_style.html#) mentions:

> For the ellipsoid style, the particles are ellipsoids and each stores a flag which indicates whether it is a finite-size ellipsoid or a point particle. If it is an ellipsoid, it also stores a shape vector with the 3 diameters of the ellipsoid and a quaternion 4-vector with its orientation.

What are applications in which there is only a small number of ellipsoids and a large number of point particles, and for which defining the shape and quaternion only for actual ellipsoids saves up significant memory?

Atom style `sphere` does a similar thing where
>If the diameter > 0.0, the particle is a finite-size sphere. If the diameter = 0.0, it is a point particle. 

However, per-atom data that is not valid for point particles, e.g., `omega` is readily accessible and not located inside a bonus structure. What explains such difference?

Given that a `quat` pointer already exists in the `Atom` class, a `shape` pointer could similarly be defined for finite-sized particles and used for ellipsoids, thereby getting rid of the bonus struct and the complexity associated with handling the information (e.g., special checks and communication).

I would be happy to discuss more about the pros and cons of the current bonus structure vs having quaternion and shape for all ellipsoid atoms.

Thank you!

**Further Information, Files, and Links**

N/A

stanmoore1:
>What are applications in which there is only a small number of ellipsoids and a large number of point particles, and for which defining the shape and quaternion only for actual ellipsoids saves up significant memory?

A typical example would be ellipsoids in a stochastic rotation dynamics (SRD) solvent, see https://github.com/lammps/lammps/blob/develop/examples/ASPHERE/ellipsoid/in.ellipsoid and https://docs.lammps.org/fix_srd.html

akohlmey:
The documentation is a bit misleading, since the ASPHERE package pair style for Gay-Berne describes an ellipsoidal generalization of the LJ potential. Also, there are multi-point "tri" and "line" particles that interact with LJ like groups of LJ point particles.

Please note that the bonus struct pointer is also used for atom style body and the pair styles in the BODY package.

jibril-b-coulibaly:
[...]

PR #4060, opened 2024-01-24
Add npair classes for a general pairwise cutoff

jtclemm:
**Summary**

Currently, LAMMPS can only build neighbor lists with either
1) type-based constant cutoffs
2) a cutoff set by the sum of radii (`NP_SIZE`)

In some pair styles, an ideal cutoff is more generally a function of both radius and type (or maybe even other properties like `atom->bonus`, https://github.com/lammps/lammps/issues/3999#issuecomment-1905269035).  In pair/lj/cut/sphere, this limitation is overcome by calculating a max `atom->radius` for each atom type to define a type-based cutoff (which won't scale well with polydispersity). In pair granular, JKR is buggy and prone to lose cohesive contacts depending on the skin distance.

In this PR, I created a new npair variant that calls a pair-defined method `pair->pair2cut()` to calculate an arbitrary cutoff length. Currently, I only tested the new class with the JKR granular pair style to fix its inconsistent contact detection. Support could be added to pair lj/cut/sphere (and maybe lepton/sphere?) to solve scaling issues with large polydispersity. However, constantly recalculating cutoffs would lead to worse performance for monodisperse systems so it isn't obviously beneficial. 

I still need to run some benchmarks/tests before this PR is ready for review, but any criticism/comments on the current approach are welcome.

**Related Issue(s)**

NA

**Author(s)**

Joel Clemmer (SNL)

**Licensing**

**Backward Compatibility**

NA

**Implementation Notes**

The old method `pair->radii2cut` was generalized to `pair->pair2cut`. Previously this method was only used to calculate cutoffs for the multi neighbor option.

**Post Submission Checklist**

Attached is an example input scripr that demonstrates the incorrect behavior of the JKR granular pair style. Two particles with radii 0.5 interact with a JKR potential with a maximum cohesive distance of 1.5. They are first pushed into contact, then pulled apart. Without this patch, the maximum interaction distance depends on the neighbor list skin.

[in.jkr.txt](https://github.com/lammps/lammps/files/14042599/in.jkr.txt)


akohlmey:
> Support could be added to pair lj/cut/sphere (and maybe lepton/sphere?) to solve scaling issues with large polydispersity. However, constantly recalculating cutoffs would lead to worse performance for monodisperse systems so it isn't obviously beneficial.

For the case where there are two possible choices, the more common one could be the default and the secondary choice could be selected via pair_modify. It would be important that the flag triggering the choice is stored in binary restarts.

jtclemm:
> For the case where there are two possible choices, the more common one could be the default and the secondary choice could be selected via pair_modify. It would be important that the flag triggering the choice is stored in binary restarts.

What if the secondary choice was automatically chosen if neighbor multi was active (instead of having a separate pair_modify command)?

jtclemm:
@akohlmey I prototyped an optional setting for pair style lj/cut/sphere to toggle how the neighbor list is built and ran some benchmarks. Since it is a very niche feature affecting ~3 pair styles, I added a flag to the pair style settings instead of pair_modify.

If one appropriately defines atom types to capture polydispersity (e.g. 2 types for a binary system), the option in this PR is always slower, as expected, due to the function call in the inner loop. Building a neighbor list seems to be around 20%-200% longer depending on the system. However, the new option does do away with the need for defining atom types to overcoming scaling issues with multi. One can just define a set of collection size intervals. This may be useful for v. broad distributions or for trading off user-effort vs CPU-effort.

Do you have any thoughts on whether this option is worthwhile or on the implementation details?

jtclemm:
I further generalized the pair methods so they flag whether particles are neighbors instead of just providing a custom cutoff. Also cut down some more code duplication in the npair classes.

jtclemm:
Can be closed if/when #4685 is merged.

Issue #3025, opened 2021-11-15
[Feature Request] GRANULAR package (pair_style granular): Polydisperse system simulations using Linear Spring Dashpot (LSD) model.


vikaskok:
**Summary**

pair_style granular cannot simulate poly-disperse system of particles with Linear Spring Dashpot (LSD) model  (with 'damping velocity' and 'damping mass_velocity')

**Detailed Description**
In 'pair_style granular', we have 4 types of damping: velocity, mass_velocity, viscoelastic, tsuji. (https://docs.lammps.org/pair_granular.html)  
I will concentrate on the first two (velocity, mass_velocity) in this query.  As per documentation, we (the user) has to specify $\eta_{n0}$. In case of damping velocity, $\eta_{n0}$ specified by the user should be in terms of mass/time units, and damping force is evaluated as: $F_{damp} = \eta_{n0} \times v$  (v=normal or tangential velocity, as the case may be.)  In case of damping mass_velocity, $\eta_{n0}$ specified by the user should be in terms of 1/time units, and damping force is evaluated as: $F_{damp} = \eta_{n0} \times meff \times v $ where meff = effective mass of colliding particles meff=(mi*mj/(mi+mj))

Evaluation of $\eta_{n0}$ in both of the above cases is as follows:

Following Silbert(2001), 
if we are to use damping velocity, then  
$\eta_{n0}=\frac{2 \ln(e_n) \sqrt{m_{eff}k}}{\sqrt{\pi^2 + \ln^2(e_n)}}$ (units: mass/time) [Eqn. 1]
and, 
if we are using damping mass_velocity, then 
$\eta_{n0}=\frac{2 \ln(e_n) \sqrt{k/m_{eff}}}{\sqrt{\pi^2 + \ln^2(e_n)}}$ (units: 1/time) [Eqn. 2]

(Some more details about [Eqn. 1, 2] are given towards the end in the footnote.) 

With the above expressions of $\eta_{n0}$, we shall have $\eta_{n0}$ in appropriate units required/expected by LAMMPS as given in the documentation.
(see also: https://github.com/lammps/lammps/issues/3016)

My concern is that the term meff appears in the expression used to evaluate $\eta_{n0}$.
Hence, we have an issue that these damping styles will work only for a mono-disperse system.

In a poly-disperse system (For eg. 10000 particle system whose dia is uniformly distributed in a range, say 0.8 to 1.2), the meff of the colliding particles is not known apriori before the simulation starts.

[...]

PR #4068, opened 2024-02-06
Coefficient of restitution based damping in granular models

dhairyaiitb:
**Summary**

Hi,

I am a postdoctoral researcher at Northwestern University and am working with Prof. Richard Lueptow, Prof. Paul Umbanhowar, and Prof. Julio Ottino. I have introduced two damping approaches making use of the implementation of Brilliantov et al., 1996: "enhooke" (for Hooke elastic force) and "enhertz" (for Hertz and Hertz/material elastic models). The aim is to input the coefficient of restitution 'en' and let the model autonomously calculate damping coefficients for each pair. This ensures a constant 'en' across all binary collisions of particle pairs, even in polydisperse media with different sizes. This resolves the issues highlighted in https://github.com/lammps/lammps/issues/3025 for linear models and provides a solution for Hertzian models as well. I've conducted validation studies for velocity evolution in two-particle collisions, but there are still some issues that I need help with from the LAMMPS granular team. These are divided into two sections:

RELATED TO THE UPDATED VERSION OF THE GRANULAR PACKAGE

1. When using the pair/local command to write forces, the solver currently only writes the elastic force, not the total force on a granule. This issue isn't exclusive to our models; it's observed in other existing models, too. These issues were not there in the older 2021 version of LAMMPS but were introduced in the updated pair granular files with sub-classes.

2. We need to verify how the modification of normal damping influences tangential damping models. I have checked this for the older 2021 version of LAMMPS, but I am not sure how to do it for this develop version without writing accurate pair/local forces. 

GENERAL

3. Assuming that these additions are accepted, I am unclear on how to incorporate the same in the documentation.

This is my first pull request, so in case I did something in an incorrect manner, I apologize for it.

**Related Issue(s)**

Fixes issue #3025 and also provides a model for Hertzian contacts. 

**Author(s)**

Dhairya Vyas
Postdoctoral Researcher, 
Department of Mechanical Engineering, 
Northwestern University Technological Institute 
2145 Sheridan Rd, Evanston, IL 60208

**Licensing**

**Backward Compatibility**

None

**Implementation Notes**

The post-collision velocities for different test cases were verified. It was ensured that for both models, the input coefficient of restitution is reproduced. 

**Post Submission Checklist**


src/GRANULAR/granular_model.cpp
```diff
@@ -10,7 +10,7 @@
 
    See the README file in the top-level LAMMPS directory.
 ------------------------------------------------------------------------- */
-
+
```
akohlmey: There seems to be only a whitespace change here.  Please run "make fix-whitespace" in the "src" folder and commit the resulting changes.

akohlmey (changes requested):
Please also update the relevant documentation.

src/GRANULAR/granular_model.cpp
dhairyaiitb: Thank you. I have removed whitespace now.

akohlmey: Looks like you missed a spot elsewhere: https://ci.lammps.org/job/dev/job/pull_requests/job/checkstyle/9310/whitespace/

dhairyaiitb: Sorry about that. I have removed the white spaces and have also added an example file to show how the new models are used. 

I am having some trouble in compiling the documentation. I am getting an error saying "/bin/bash: line 4: deactivate: command not found make: *** [Makefile:271: /mnt/d/gitlammps/lammps/doc/docenv/bin/rst_anchor_check] Error 127"  

I am trying to understand and resolve it. Sorry for the inconvenience.

akohlmey: What platform are you on? Is it some kind of Linux system?
You need python3 with a working venv module, doxygen, and internet access, so the missing python modules can be installed into a virtual environment.
Specifically, you have to be able to do (on Linux or MacOS):
```
python -m venv newenv
source newenv/bin/activate
pip install --upgrade pip
deactivate
rm -r newenv
```

If you are on Windows, things can get complicated. It is probably better to use a Linux machine (real or virtual).

sjplimp:
@akohlmey I will review after Joel/Dan have a chance to look at it.

jtclemm:
Thanks Dhairya for the contribution. It looks like a helpful addition to the granular package. 

Am I correct to assume there would be no reason to use the enhooke damping model with any normal force model other than hooke? And similar for enhertz and hertz(/material)? Would it ever make sense to pair one with a JKR or DMT cohesive model?

I'm wondering if they could be combined into one class with a more descriptive name. Then the init() method could calculate the correct prefactor (or error) depending on the normal model.

src/GRANULAR/granular_model.cpp
dhairyaiitb: Thank you. I am using a Linux system. I had everything installed but was still getting an error. I just had to do a make clean-all once before running make again. I am currently working on the documentation and will update it in a few days after finalizing it with my university collaborators.

dhairyaiitb:
Yes, @jtclemm, you are right. We will use *enhooke* only with the *hooke* model and *enhertz* with *hertz* and *hertz/material* models. I haven't studied the cohesive models in-depth, so I'm not sure how useful this will be for JKR or DMT. However, I have verified that in the absence of cohesive forces, *enhertz* reproduces the input coefficient of restitution for the JKR model.

Having a single class called *en* will be useful. When I started working on this problem using the 2021 version of the LAMMPS code, I was using the *en* damping model, which calculated the dissipation term based on the chosen normal model. However, I couldn't figure out how to do that with the new subclass system, so I ended up implementing two different damping models.

jtclemm:
The normal model currently indicates whether the contact model is cohesive, `cohesive_flag`. Conflicts with cohesive contacts (e.g. in the Tsuji damping model) can then be defined using an `allow_cohesion` flag. This could be expanded to designate whether a contact model is Hookean vs. Hertzian. Let me talk to @dsbolin and think about it some more.

Does _en_ stand for $\epsilon^N$? Maybe I'm not familiar with standard DEM notation, but this wasn't obvious to me. Is there a more descriptive name that would be appropriate? Brilliantov, CoeffRestitution, ...?

dhairyaiitb:
Thank you. I will also have another look at it and think more. Sorry about the confusion in the name. I am using en as the normal coefficient of restitution (so e is for coefficient of restitution and n for normal). We can use any other suitable name too.

jtclemm:
I think the best solution at the moment might be to add a user specified argument which controls the form of the prefactor (e.g. _hertz_ or _hooke_). We may then add an enum which lists possible (idealized) normal models (hooke, hertz, cohesive, other). If there's a clear contradiction (e.g. the user requests a Hookean prefactor but uses a Hertz normal contact), it can error.

Does this sound reasonable to you?

dhairyaiitb:
Thank you. Sorry for the late response; I have been caught up with a few things and will get back to this soon. Mainly after the 10th of March as I will be traveling for the next few days. 

I have updated the documentation and added the formulae. However, this is just the first draft; we will update it further as we make more changes. 

The prefactor idea sounds really good, and the enum error-based approach sounds good.  I will go through it and get back in some time.

akohlmey:
@dhairyaiitb Any updates on this pull request?

dhairyaiitb:
Hello, I am currently working on addressing the prefactor approach and errors. I will update both the code and documentation by the end of the week. My apologies for the delay.

dhairyaiitb:
Hi,

I have now changed the command to hertz/en and hooke/en and have added checks which result in error messages when hooke normal model is used with hertz/en and vice versa. Should we add checks for cohesive models too?

jtclemm:
Thanks for making the changes. I have a few more suggestions for the organization of the classes/flags (which I can just make myself if you don't mind). However, I first have a few questions: 

1. Is there a particular citation for the derivations of eta as a function of the coeff. of restitution? I didn't see it in the cited Brilliantov paper (maybe I missed it). Or are these expressions you derived and verified?
2. I may be missing something, but there seems to be an extra negative sign in the expressions for `damp` in the sub models. There also seems to be an extra sign in the expression in the documentation for $\eta_n$ for Hertz (but not in Hooke). Can you confirm?
3. Have you have any ideas of a more descriptive name for the damping model than en? Would CoeffRestitution be appropriate in your opinion?

Regarding the other questions in your original post:

1. Can you elaborate more on what is different in the new vs. old output of compute pair/local? Is it a particular component that does not include damping forces? Checking the code, the values returned should be unchanged from the old pair granular.

2. The tangential dissipative forces are all constructed as described in the documentation, $F_{T,D} = - x \eta_{normal} \vec{v}_{t,rel}$. This can be most easily seen in the `LinearNoHistory` model: https://github.com/lammps/lammps/blob/9ca47a36d2cf8f6df15424bb09b2d958cf070063/src/GRANULAR/gran_sub_mod_tangential.cpp#L71.

dhairyaiitb:
Sure, @jtclemm, feel free to make any changes you deem necessary. Here are the responses to your questions:

1. To my understanding, the formula for the Hertzian case is derived by solving Eqs. 19 and 20 from Brilliantov's paper. Although I haven't re-derived or verified it myself, I've observed that it accurately reproduces the coefficient of restitution for various sizes and material properties. This same formula is also utilized in several other open-source codes like LIGGGHTS (refer to https://www.cfdem.com/media/DEM/docu/gran_model_hertz.html) and Lethe-DEM (see https://arxiv.org/pdf/2106.09576.pdf Eq. 13).

2. Regarding the $\eta_n$ term, it must be positive to ensure the correct sign of the damping term, especially considering that log(en) is negative. Therefore, I've introduced an additional negative sign in the Hertz en model. In Hooke's model, since the log(en) term is within the square root, it's assumed that we'll use the positive root, hence the absence of the negative sign. Alternatively, we could consider adding an absolute value sign in the code's documentation to prevent confusion.

3. Certainly, names like "CoeffRestitution" or "cor" sound suitable.

Additional responses:

1. I am using the following commands to obtain contact forces:

```
compute nl all property/local patom1 patom2 cutoff radius # neighbor list
compute nforces all pair/local force dist p1 p2 p3
```

The results in the output file differ between the master LAMMPS script dated February 8, 2023, and the current version we are working on. I haven't verified the individual codes, but it seems like the newer version only writes the elastic component. Here's a comparison of the force with the analytical solution for both scripts. Please note that only the dumped forces are incorrect; the actual forces used in both simulations are identical.

Old:
image


New:
image


2. Thank you for your input. In that case, it might be beneficial to allow users to determine the value of x they wish to use. We can suggest that for the Hertzian model with Poisson's ratio $v$, $x = \sqrt{2(1-v)/(2-v)}$ would be a preferable choice, based on recommendations from LIGGGHTS and Lethe documentation.

I'd also like to extend my gratitude for your patience and tolerance throughout this process. This marks my initial endeavor in modifying LAMMPS and committing code, so I'm still in the learning phase. Your guidance and support have been invaluable, and I'm gaining valuable insights from this experience. Thank you for your assistance.

jtclemm:
Great, thanks for the explanations Dhairya. I made some minor organization changes, primarily I groupingthe two damping models into one class. Please let me know if these changes look good. I didn't add a recommendation for x_t, but feel free to add it if you think it's appropriate.

Good catch on the force output, I think you are right that there was an undocumented change in behavior between those two versions in LAMMPS (probably my fault). Thanks for catching that. I made an update to pair_granular.cpp which should include the damping contribution in the `single()` method. ~@dsbolin , can you confirm that this is the intended output?~ this is stated in the documentation.

jtclemm (approved):

stanmoore1:
>Great, thanks for the explanations Dhairya. I made some minor organization changes, primarily I groupingthe two damping models into one class. Please let me know if these changes look good.

@dhairyaiitb does this look good to you?

akohlmey (approved):

dhairyaiitb:
Thanks, @jtclemm, this looks great. I ran some verification studies, and now the forces with the coeff_restitution model also match well with the analytical solution.

dhairyaiitb:
> > Great, thanks for the explanations Dhairya. I made some minor organization changes, primarily I groupingthe two damping models into one class. Please let me know if these changes look good.
> 
> @dhairyaiitb does this look good to you?

Yes thank you @stanmoore1

PR #4069, opened 2024-02-07
New fix implementing flow boundary conditions

vladgl:
**Summary**

This fix implements the flow boundary condition (FBC) method for a stationary flow simulation. The method, some implementation details, and performance measurements are described [here](https://journals.sagepub.com/doi/10.1177/10943420231213013). 

**Related Issue(s)**

**Author(s)**

Vladislav Galigerov @vladgl vladgl3@yandex.ru (Higher School of Economics)
Daniil Pavlov @Dann239 pavlov.dg@phystech.edu (The Moscow Institute of Physics and Technology)

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

Here are simulation animations for ~230 million atoms with 5 fbc planes, produced with our fix (graphs show density and vorticity aggregated on a grid).

https://github.com/lammps/lammps/assets/18269157/6bb278cb-4a3f-4ccc-bebd-48bb98362f18
https://github.com/lammps/lammps/assets/18269157/2b13fb1b-305e-46d2-aba0-977c1b981934


akohlmey:
@lammps/core do you consider this fix sufficiently important and useful to stay in `src` or should it be moved to the `EXTRA-FIX` package?

akohlmey:
@vladgl please go to the `src` folder, type `make fix-whitespace` and commit the resulting changes.  Please see: https://docs.lammps.org/Modify_style.html#whitespace-preferred

sjplimp:
@akohlmey move to src/EXTRA-FIX

src/EXTRA-FIX/fix_wall_flow.cpp
```diff
@@ -0,0 +1,294 @@
...
+#include "modify.h"
+#include "update.h"
+#include "variable.h"
+#include "random_mars.h"
+#include "memory.h"
+#include "force.h"
+#include "math_const.h"
+
+#include <cstring>
+#include <iostream>
```
akohlmey: You should not be using iostreams from STL. LAMMPS uses stdio from the c-library and those two generally should not be mixed. Furthermore there are a bunch of convenience functions for output (which should be usually directed to the "screen" and "logfile" FILE pointers via the `utils::logmesg()` convenience function.
Formatting generally uses fmtlib in anticipation of the C++20 standard which implements the same functionality as part of the standard.

src/EXTRA-FIX/fix_wall_flow.cpp
```diff
@@ -0,0 +1,294 @@
...
+#include <algorithm>
+#include <functional>
+#include <fstream>
```
akohlmey: Same as above.

src/EXTRA-FIX/fix_wall_flow.cpp
```diff
@@ -0,0 +1,294 @@
...
+  " author = {Daniil Pavlov and Vladislav Galigerov and Daniil Kolotinskii and Vsevolod Nikolskiy and Vladimir Stegailov},\n"
+  " title = {GPU-based molecular dynamics of fluid flows: Reaching for turbulence},\n"
+  " journal = {The International Journal of High Performance Computing Applications},\n"
+  " year =    2024,\n"
+  " volume =  38,\n"
+  " number =  1,\n"
+  " pages =   34-49\n"
+  "}\n\n";
+
+FixWallFlow::FixWallFlow(LAMMPS *lmp, int narg, char **arg) :
```
akohlmey, line 58: In general, we prefer contributed code to follow the LAMMPS programming style for formatting, indentation, so on. This is closely reflected in the provided configuration for the "clang-format" tool. By removing the comment at the top and then running clang-format on the source, it would be reformatted to follow the expected style.
https://docs.lammps.org/Modify_style.html

src/EXTRA-FIX/fix_wall_flow.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+  ++iarg;
+  // parsing wall count
+  int wallcount = utils::inumeric(FLERR,arg[iarg],do_abort,lmp);
+  if(wallcount <= 0) error->all(FLERR,"Illegal fix wall/flow argument: wall count must be positive integer");
+
+  ++iarg;
+  // parsing walls
+  if(narg - iarg != wallcount && narg - iarg != wallcount + 2) error->all(FLERR, "Wrong fix wall/flow wall count");
+  auto getscale = [&]() -> double {
```
akohlmey: Why use a lambda function for such a trivial purpose? It just makes the code harder to read for people that are not C++ gurus.
Why not simply use simple if/else if/else if/else construct? The setting of scale in line 127 seems redundant.

src/EXTRA-FIX/fix_wall_flow.cpp
```diff
@@ -0,0 +1,294 @@
...
+  for (size_t w = 1; w <= wallcount; ++w, ++iarg)
+  {
+    walls[w] = utils::numeric(FLERR,arg[iarg],do_abort,lmp) * scale;
+  }
+  walls.back() = domain->boxhi[flowax];
+  if (!std::is_sorted(walls.begin(), walls.end(), std::less_equal<double>()))
+  {
+    error->all(FLERR, "Wrong fix wall/flow wall ordering or some walls are outside simulation domain");
+  }
+  std::cout << "Walls:\n";
```
akohlmey: This has two problems:
- output should only be created by MPI rank 0, i.e. for `comm->me == 0` so that it is not replicated for parallel runs
- informational output should use the `utils::logmesg()` function
- i/o stream formatting is clumsy and slow.

src/EXTRA-FIX/fix_wall_flow.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+  return mask;
+}
+
+/* ---------------------------------------------------------------------- */
+
+void FixWallFlow::init()
+{
+  int nrigid = 0;
+  for (int i = 0; i < modify->nfix; i++)
```
akohlmey: This kind of construct is strongly discouraged since it requires accessing internal data of the Modify class.
Instead you can use:
```
for (auto ifix : modify->get_fix_list()) if (ifix->rigid_flag) nrigid++;
```

src/EXTRA-FIX/fix_wall_flow.h
```diff
@@ -0,0 +1,60 @@
...
+// clang-format off
+FixStyle(wall/flow,FixWallFlow);
+// clang-format on
+#else
+
+#ifndef LMP_FIX_WALL_FLOW_H
+#define LMP_FIX_WALL_FLOW_H
+
+#include "fix.h"
+#include <vector>
```
akohlmey: This include is not needed. See: https://docs.lammps.org/Modify_style.html#include-files-varied

src/EXTRA-FIX/fix_wall_flow.h
```diff
@@ -0,0 +1,60 @@
...
+  void grow_arrays(int) override;
+  void copy_arrays(int, int, int) override;
+
+  int pack_exchange(int, double *) override;
+  int unpack_exchange(int, double *) override;
+
+ protected:
+  FlowAxis flowax;
+  double flowvel;
+  double kT;
```
akohlmey, line 46: This is not needed. If anything should be pre-computed it would be "gamma" in the `generate_velocity()` member function and for that it would be best to pre-compute it *outside* the loop in `end_of_step()` and then pass it as a const reference.

src/EXTRA-FIX/fix_wall_flow.cpp
```diff
@@ -0,0 +1,294 @@
...
+       }
+    }
+  }
+}
+
+void FixWallFlow::generate_velocity(int atom_i)
+{
+  const int newton_iteration_count = 10;
+  double *vel = atom->v[atom_i];
+  double mass = atom->mass[atom->type[atom_i]];
```
akohlmey: This assumes that masses are per-type, but that may not be true. For systems with per-atom masses, you must use `atom->rmass[atom_i]`. You would get an incorrect mass or a crash due to a segmentation fault otherwise. See, for instance, fix nve.

akohlmey (changes requested):
I've flagged several programming style issues and provided suggestions to make this code better aligned with how the rest of LAMMPS is written. I understand that it can be annoying to have to program according to somebody else's preferences, but keep in mind that sticking to a similar style makes it easier for maintainers and others looking to improve, modify or extend your code.

akohlmey:
> @akohlmey move to src/EXTRA-FIX

done

src/EXTRA-FIX/fix_wall_flow.cpp
vladgl: I used it for debugging, so I deleted it. I don't think there is a need to output wall coordinates, since the user sets them in the input file.

src/EXTRA-FIX/fix_wall_flow.h
vladgl: We can't pre-compute "gamma" outside the loop, because it depends on the atom mass

src/EXTRA-FIX/fix_wall_flow.cpp
vladgl: Done

src/EXTRA-FIX/fix_wall_flow.cpp
vladgl: Done

src/EXTRA-FIX/fix_wall_flow.cpp
vladgl: I took this into account in the kokkos version, but somehow forgot it here, fixed

src/EXTRA-FIX/fix_wall_flow.h
vladgl: Done

akohlmey:
@athomps can you have a closer look at this? I recall you did something that sounds similar some time ago.

stanmoore1 (approved):
I approve

sjplimp:
Hi - thanks for submitting this new fix flow/wall command.  I pushed a few changes to the doc page file and the source code file.
The changes to the source code were just cosmetic formatting.  I reworded portions of the doc file to make it more clear to users how this works.  You should read the new version carefully to ensure I didn't make any mistakes.  In the math formulas I changed v_z to v_a, since I thought it was confusing to have z stand for any of x,y,z.  So make sure the math formulas are also still correct.

Here are a few other conceptual Qs:

(1) Is this only meant to be used with a system periodic in the flow direction?  So that the impliclty added wall at the boundary is effectively both the lo and hi boundary?  If so, it should be checked for in the code.
(2) The doc page says N+1 walls, but the code allocates walls for N+2 ?  Possibly this is related to the periodic effect in (1) of having an extra wall at both the lo/hi boundaries?
(3) The code only alters velocities of particles in the group, which I now mention in the doc page.  If applying this fix to a subset of particles is a legitimate use case, it would be good to explain the use case in the doc page.
(4) Does the is_sorted() error check on wall coords include checks that no specified wall is "on" the boundary, or that 2 or more specified walls are not identical.  Also what happens if the box size in the flow direction varies in time, e.g. a constant pressure condition or shrink-wrapped boundary condition.  Does that mess up the implicit wall position, or in the extreme case the box boundary could cross one of the implicit walls.  You could do checks on settings in domain.h if you need to prohibit this.

vladgl:
@sjplimp Hi, thank you for your review and comments. I've checked your changes in the documentation and everything seems fine to me. Regarding your questions: 
(1) Yes, this should only work with periodic boundary conditions along the flow, I have added the check. 
(2) The first and the last elements in the walls array are the same wall on a periodic boundary. This just simplifies the loop in the compute_current_segment function and nothing more.
(3) I've added these examples to the document, as well as the periodic boundary details.
(4) Yes, you're right! I added a check for duplicates. About changing size along the flow direction: there is no flag in domain.h to check this, but there is a box_change flag that is 1 when any of the box sizes change. I can do this check myself in FixWallFlow::init() just like it is done in Domain::init(), or maybe we can add necessary flags as members of the Domain class?

sjplimp:
Thanks @vladgl - Re (4): For a flow in the x direction are you saying you want to prohibit any other fix from changing the box size in x ?  Or are you saying that is OK, so long as the specified walls remain "inside" the box, and only need to check that?  For the latter, that could be checked each timestep in fix wall/flow.  I.e. check that the first wall > boxlo, and last wall < boxhi.

For the former, it is more tricky, since there is no specific setting in domain.h for changes in just the x size.  I presume you want to allow variations in y and z dimensions ?  Possibly fix deform and fix npt (possibly other fixes as well), could create more details settings in domain.h for what they do.  However note, they also do things for non-triclinic boxes to change the box shape.  Perhaps your fix should check that the simulation box is not triclinic, as your walls assume an orthogonal box?

sjplimp:
@vladgl - Maybe a simpler thing to do is just check in fix wall/flow each timestep if the boxlo/boxhi in the chosen dimension have changed, and throw an error at run time rather than setup time.

akohlmey:
@sjplimp FYI, fixes report in the `Fix::box_change` bitmap, whether they change individual box dimensions. We use this to detect conflicts between, e.g. fix deform and fix npt, yet allow to use them when they don't operate both on the same dimension.

vladgl:
@sjplimp I can't imagine scenarios where someone would need a feature with box resizing along the flow direction, so it's better to prohibit it. 
I already added a check for triclinic box in the last commit via the domain->triclinic flag, but I'm not sure whether this can be changed by other fixes?
@akohlmey Here we have some local variables that mark whether or not we are resizing the box in different directions. Can we just make them members of the Domain class itself?
https://github.com/lammps/lammps/blob/81609d04e3fc863eac9596fa475726018761242d/src/domain.cpp#L144

sjplimp:
Thanks @akohlmey -- @vladgl it looks like you could check that for just X, Y, or Z and throw an error in fix wall/flow.  You should
also check if domain->triclinic == 0 for an orthonormal box.  If that is the case, there are no shape issues.

akohlmey:
> Can we just make them members of the Domain class itself?

I don't think that is a good idea.

If your fix changes box dimensions, however, you *must* indicate that by setting the proper box_change bitmap flags, so that the check in the Domain class will throw an error, if there is a conflict. Even if you are not changing the box, but don't want any other fix to change it you can pretend to change it yourself and set the BOX_CHANGE_<DIM> flag to trigger that general error check.

sjplimp:
I agree with @akohlmey.  I think fix wall/flow simply needs to check that domain->triclinic = 0 and do the bit-wise mask check @akohlmey pointed to in fix.h to ensure that the flow dimension box size is not being changed by some other fix.  Doing both of those in init() is sufficient to protect against a new definition of a fix (e.g. between runs) which messes things up for fix wall/flow.

vladgl:
@akohlmey @sjplimp Added a check for varying box size in the init function, and checks for a triclinic box and periodic dimension in the constructor.

akohlmey:
> @akohlmey @sjplimp Added a check for varying box size in the init function, and checks for a triclinic box and periodic dimension in the constructor.

Why only use `error->warning()`, why not `error->all()`?

vladgl:
@akohlmey Fixed it, forgot that error->warning doesn't interrupt execution

src/EXTRA-FIX/fix_wall_flow.cpp
```diff
@@ -0,0 +1,328 @@
...
+void FixWallFlow::init()
+{
+  int nrigid = 0;
+  int box_change_flowax = 0;
+  for (const auto &ifix : modify->get_fix_list()) {
+    if (ifix->rigid_flag) nrigid++;
+    switch (flowax) {
+      case FlowAxis::AX_X:
+        if (ifix->box_change & Fix::BOX_CHANGE_X) box_change_flowax++;
+        if (ifix->box_change & Fix::BOX_CHANGE_XY) box_change_flowax++;
```
sjplimp: Since the box cannot be triclinic, there is no need to check in init() for XY, XZ, YZ

sjplimp (changes requested):
Thanks for the changes - if you address my one additional comment, I am ready to approve.

akohlmey:
> Thanks for the changes - if you address my one additional comment, I am ready to approve.

Just made the change and added a check for triclinic to init() since somebody may have used change_box after the definition of the fix.

akohlmey (approved):

sjplimp (approved):

PR #4072, opened 2024-02-07
Fix issue from #4007

stanmoore1:
**Summary**

Fix issue from #4007 that causes
```
In file included from ../pppm.h:24,
                 from ../pppm.cpp:22:
../lmpfftsettings.h:96:4: error: #error "Must enable CUDA with KOKKOS to use -DFFT_KOKKOS_CUFFT"
 #  error "Must enable CUDA with KOKKOS to use -DFFT_KOKKOS_CUFFT"
    ^~~~~
make[1]: *** [pppm.o] Error 1
```
even when CUDA is enabled. This is due to `KOKKOS_ENABLE_CUDA` and similar not being defined in `lmpfftsettings.h`. Including `accelerator_kokkos.h` would fix the issue but we don't want to include headers in other "clean" headers when possible. So the alternative is to move this into `src/KOKKOS/fftdata_kokkos.h` where `KOKKOS_ENABLE_CUDA` is already defined.

**Related Issue(s)**

#4007 

**Author(s)**

Stan Moore (SNL), plus discussion with Axel Kohlmeyer (Temple U)

**Licensing**

**Backward Compatibility**

Yes


stanmoore1:
Confirmed working with `cuFFT` on V100.

sjplimp (approved):

akohlmey:
@stanmoore1 had to make some small changes to one of the two cases. It compiles for me with KOKKOS/OpenMP. But I have not tested beyond that.

stanmoore1:
I will test again, give me a bit.

stanmoore1:
Still compiles for me and works as expected for cuFFT on V100

stanmoore1:
Still works as expected, ready to merge.

PR #4094, opened 2024-03-01

### LINKED ISSUE #4093: [BUG] Compute RDF unphysical with neighbor multi
**Summary**

`compute rdf` produces incorrect results with `neighbor multi`, the curve decays to zero. My scenario is a WCA liquid for which I want an RDF for comparably long distances. The issue does not occur with `neighbor bin`.

7 Feb 2024 - Development - patch_7Feb2024_update1-147-g4d89741d8c

**Expected Behavior**

`neighbor multi` should lead to the same RDF as `neighbor bin`.

**Actual Behavior**

The RDF produced with `neighbor multi` decays to zero as shown in the plot.

**Steps to Reproduce**

A simplified input script is attached ([in.spheres.txt](https://github.com/lammps/lammps/files/14438358/in.spheres.txt)). Equilibrate a system of WCA spheres. Next, make sure to use `neighbor bin` and measure the RDF with `compute rdf`. Subsequently, turn to `neighbor multi` and measure it again. The discrepancies can be spotted directly in the output files without plotting if desired.

Longer force cutoffs can solve the issue. Note that if I calculate the RDF in post-processing, it’s fine in both cases.

akohlmey:
**Summary**

This pull request combines multiple small fixes and changes.

**Related Issue(s)**

Fixes #4093 by disallowing the combination of options
Fixes issue mentioned here: https://matsci.org/t/invalid-dump-frequency-error-when-exceeding-32-bit-integer-timestep/53947

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- when searching for fix drude, more stringent search patterns are required to avoid false positives
- bug fixes for meam/ms pair style from @athomps 
- try alternate approach to avoid triggering the "fix not computed at compatible time" error with write_dump. This should also work with timestep numbers that require 64-bit integers.
- flag incompatibilities of some computes and fixes with "multi" or "multi/old" style neighbor lists
- add a "report" style to fix balance to enable monitoring imbalance without rebalancing the subdomain. may be used with fix halt to stop a job when the system has changed too much.
- more notes for porting old code to current LAMMPS

**Post Submission Checklist**


sjplimp:
@akohlmey I want to talk to Joel about neigh multi and cutoffs before this PR is merged

sjplimp:
@akohlmey all clear on neigh multi error messages

For the write_dump.cpp change, I don't understand what you mean by work-around.  If the fix is not current (or compatible) with the current timestep, won't it still print bad values into the dump snapshot ?

akohlmey:
> If the fix is not current (or compatible) with the current timestep, won't it still print bad values into the dump snapshot ?

When you have a write_dump you know you get whatever is currently stored. The workaround to outsmart a check the Output class does on *future* outputs. This happens because of how write_dump works, i.e. set up a regular dump, but then tell the Output class to write this dump's output just once.

Below is a simple test input. Try running this without the workaround, i.e. use the line
```
dumpargs[3] = utils::strdup("1");    // dump frequency
```
Then try with the current code and finally with the change in this pull request.
```
region box block 0 1 0 1 0 1
create_box 1 box
create_atoms 1 single 0.5 0.5 0.5
pair_style zero 2.5
pair_coeff 1 1
mass 1 1.0
reset_timestep 200
fix 1 all ave/atom 10 2 20 x y z
run 100 post no
write_dump all custom 0.lammpstrj id type x y z f_1[1] f_1[2] f_1[3]
unfix 1
reset_timestep 2147483647001
fix 1 all ave/atom 10 2 20 x y z
run 40 post no
write_dump all custom 1.lammpstrj id type x y z f_1[1] f_1[2] f_1[3]
```

rbberger:
@akohlmey something small for this PR: https://github.com/lammps/lammps/blob/develop/cmake/presets/kokkos-hip.cmake#L16 should be:
```cmake
set(FFT_KOKKOS "HIPFFT" CACHE STRING "" FORCE)
```
This might be in other places too.

akohlmey:
> This might be in other places too.

I recall that it *was* in the KOKKOS/CUDA preset, but that was fixed about a couple of weeks ago...

stanmoore1 (approved):
I approve

sjplimp (approved):

PR #4098, opened 2024-03-10
Fix variables compatibility with chunk arrays

GenieTim:
**Summary**

When using variables with chunk computes that produce arrays (such as `compute com/chunk`) the compute will not have set `size_array_rows` to the appropriate value before it has ever been called and will therefore incorrectly have thrown the error "Variable formula compute array is zero length".

**Related Issue(s)**

None, as far as I see.

**Author(s)**

**Licensing**

**Backward Compatibility**

Less errors are thrown :)

**Implementation Notes**

Maybe this should be fixed on the side of the computes instead, i.e., to have consistent behaviour?
In that case, this is just an issue.

**Post Submission Checklist**

None, thanks.


sjplimp:
@GenieTim - thanks for submitting this PR.  I'm not clear on how you are generating the error (before your code change).
In current master, the test script below runs fine.  If you uncomment the print line it throws an appropriate error, which is different than the one you mention: 
 
ERROR: Variable one: Variable formula compute cannot be invoked before initialization by a run (../variable.cpp:1566)

Possibly you are using an older version of LAMMPS?
Or can you post a simple test script which illustrates the issue this PR is fixing?

Steve

------------------------

units           lj
atom_style      atomic

lattice         sc 0.8442
create_box      1 box
create_atoms    1 box
mass            1 1.0

velocity        all create 1.44 87287 loop geom

pair_style      lj/cut 2.5

neighbor        0.3 bin
neigh_modify    delay 0 every 20 check no

fix             1 all nve

compute         1 all chunk/atom bin/1d x lower 0.25 units reduced

compute         2 all com/chunk 1

variable        one equal c_2[2][1]

# comment this line in or out
#print           "VAR = ${one}"

thermo_style    custom step temp pe press v_one

run             100

GenieTim:
Dear @sjplimp 

The script I used was something along the lines of

```
# (box & atom setup here)

# (force-field here)

# now, the relevant code:


compute moleculeChunks all chunk/atom molecule
compute moleculeCom all com/chunk moleculeChunks
variable globalComX vector c_moleculeCom[1]
fix comPosCorr all ave/correlate/long 1 1000000 v_globalComX[*100] type auto nlen 16 ncount 2 ncorr 30 file ./output/some-output-file.out.correlate.txt
```

I can still reproduce this with a newly compiled version of LAMMPS, fresh from the develop branch: it results in a `ERROR: Variable globalComX: Variable formula compute array is zero length (src/variable.cpp:1661)`.

sjplimp:
Thanks - is this error happening when you define the correlate fix?  Or when you start a run?

GenieTim:
When I start a run

sjplimp:
Also, how many molecules (chunks) in your system?

GenieTim:
> Also, how many molecules (chunks) in your system?

100. But I suppose my change proposed in this PR would not work if this was the issue.

sjplimp:
@GenieTim - Now that I understand it, I agree with your change in this PR.  Two requests:

(1) Just move the one test on size_array_rows below the compute invocation.  The test on size_array_cols is fine to do before since the column size is already set.  Only the row count is still pending until the initial invocation.

(2) Add a comment before the test on size_array_rows:
// wait until after compute invocation to check size_array_rows
// b/c may be zero until after initial invocation

Also, I think using v_globalComX[*100] in your script is fine if the mol count is not 100.  If it is less, then those
entries should be returned as zeroes by the variable evaluation.

GenieTim:
I have made the requested adjustments.

> Also, I think using v_globalComX[*100] in your script is fine if the mol count is not 100. If it is less, then those
entries should be returned as zeroes by the variable evaluation.

Indeed. 
Me using `*100` instead of simply `*` is due to another possible bug: just using `*` results in an error `ERROR: Unknown fix ave/correlate/long data type: v_globalComX[*] (src/EXTRA-FIX/fix_ave_correlate_long.cpp:98)`.

sjplimp:
No, the error message is admittedly opaque, but is correct.  See the NOTE on the fix ave/correlate doc page about needing to use an upper bound with a wildcard for a vector-style variable.  In a separate pending PR of Axel's, I edited the fix ave/correlate/long doc page to tell the user to read about the wildcard rules in the other doc page where that NOTE appears. That error message encompasses a variety of syntax errors with the c,f,v prefixed values.

sjplimp (approved):

PR #4102, opened 2024-03-15
Fix Kokkos "team size too large" issue

stanmoore1:
**Summary**

When running Kokkos on GPUs with neighbor threading (default for 16k atoms or less) the code sometimes gave an error that the requested team size was too large. The issue was that both a `parallel_for` and `parallel_reduce` could be called between reneighboring which have a different max team size, but the code assumed the team size was fixed between reneighboring.

**Related Issue(s)**

#4003 

**Author(s)**

Stan Moore (SNL)

**Licensing**

**Backward Compatibility**

Yes


stanmoore1:
@akohlmey this is small enough, could adsorb into `collected-small-changes` (I would have pushed to https://github.com/lammps/lammps/pull/4103  originally but  wasn't open at the time).

sjplimp (approved):

PR #4110, opened 2024-03-25
Implementation of pair_style uf3 and uf3/kk

monk-04:
**Summary**

This pull request implements the [ultra-fast forcefield (uf3) machine learned interatomic potential](https://www.nature.com/articles/s41524-023-01092-7). Both the CPU and KOKKOS versions of uf3 are implemented.

**Related Issue(s)**

None

**Author(s)**

1. Ajinkya C Hire (University of Florida), email- ajinkya.hire@ufl.edu, ajinkyahire62@gmail.com
2. Hendrik Kraß (University of Constance)
3. Matthias Rupp (Luxembourg Institute of Science and Technology), email- mrupp@mrupp.info
4. Richard Hennig (University of Florida), email- rhennig@ufl.edu

**Licensing**

**Backward Compatibility**

Will not break backward compatibility

**Implementation Notes**

The correctness of the implementation was verified by comparing the lammps computed energy and forces with that calculated from the python version of uf3 found [here](https://github.com/uf3/uf3). The current version computes two-body and three-body interactions, contributions from one-body are ignored.

**Post Submission Checklist**


src/ML-UF3/pair_uf3.cpp
```diff
@@ -0,0 +1,1313 @@
...
+        fjk[1] = fkj[1] = 0;
+        fjk[2] = fkj[2] = 0;
+
+        k = neighshort[kk];
+        ktype = type[k];
+        del_rki[0] = x[k][0] - xtmp;
+        del_rki[1] = x[k][1] - ytmp;
+        del_rki[2] = x[k][2] - ztmp;
+        rik = sqrt(
+            ((del_rki[0] * del_rki[0]) + (del_rki[1] * del_rki[1]) + (del_rki[2] * del_rki[2])));
```
srtee: Lots of unnecessary rij recalculations here -- see comment at line 1015

src/ML-UF3/pair_uf3.cpp
```diff
@@ -0,0 +1,1313 @@
...
+        if (pot_3b) {
+          if (rij <= cut_3b_list[itype][jtype]) {
+            neighshort[numshort] = j;
+            if (numshort >= maxshort - 1) {
+              maxshort += maxshort / 2;
+              memory->grow(neighshort, maxshort, "pair:neighshort");
+            }
+            numshort = numshort + 1;
+          }
+        }
```
srtee, line 1451: from what I understand, this creates an internal "shorter-neighbor list" within which to evaluate 3body interactions. To me a simpler approach would be:
1. something like `double * rij_list = new double[inum]` (but create and manage the memory in a broader context for better predictability and performance)
2. after calculating each `rij`, store it: `rij_list[jj] = rij`
3. you now have the distances on hand for 3-body interactions later on

An extra `inum`-worth of doubles isn't a prohibitive amount of memory -- 1000 doubles is 8 KB.

src/ML-UF3/pair_uf3.cpp
```diff
@@ -0,0 +1,1313 @@
...
+        for (int k = j; k < num_of_elements + 1; k++) {
+          bytes += (double)2*n3b_knot_matrix[i][j][k][0].size()*sizeof(double);
+          bytes += (double)2*n3b_knot_matrix[i][j][k][1].size()*sizeof(double);
+          bytes += (double)2*n3b_knot_matrix[i][j][k][2].size()*sizeof(double);
+
+          std::string key = std::to_string(i) + std::to_string(j) + std::to_string(k);
+
+          for (int l=0; l < n3b_coeff_matrix[key].size(); l++){
+            for (int m=0; m < n3b_coeff_matrix[key][l].size(); m++){
+              bytes += (double)2*n3b_coeff_matrix[key][l][m].size()*sizeof(double);
```
srtee: I don't see that n3b_coeff_matrix is used outside the initialization functions -- clearing it after initialization would save some memory (probably not much?).

src/ML-UF3/uf3_pair_bspline.cpp
```diff
@@ -0,0 +1,157 @@
...
+  ret_val[0] += bspline_bases[knot_affect_start].eval3(rth, rsq, r);
+
+  // Calculate force
+
+  ret_val[1] = dnbspline_bases[knot_affect_start + 2].eval0(rsq, r);
+  ret_val[1] += dnbspline_bases[knot_affect_start + 1].eval1(rsq, r);
+  ret_val[1] += dnbspline_bases[knot_affect_start].eval2(rsq, r);
+
+  return ret_val;
+}
```
srtee: Instead of returning a pointer-to-double you can return a force double and modify the energy by reference. So the function signature would be

`double uf3_pair_bspline::eval(double r, double &eng)`

and your evaluations would proceed as

```
... eng = 0
    eng += \* first energy component *\
    eng += \* second energy component *\
...
    f = 0
    f += \* first force component *\
    f += \* second force component *\
... return f
```

I am not sure how much this changes performance but it will make code more readable for future maintainers.

src/ML-UF3/pair_uf3.cpp
```diff
@@ -0,0 +1,1313 @@
...
+        fz = delz * fpair;
+
+        f[i][0] += fx;
+        f[i][1] += fy;
+        f[i][2] += fz;
+        f[j][0] -= fx;
+        f[j][1] -= fy;
+        f[j][2] -= fz;
+
+        if (eflag) evdwl = pair_eval[0];
```
srtee: See comment on the `eval` function -- it may be more readable to modify energy by reference. So the function signature would be

`double CLASSNAME::eval(double r, double &eng)`

and it could be called as

`fpair = -1 * UFBS2b[itype][jtype].eval(rij, evdwl) / rij;`

This has minor readability impact here -- later on in the 3-body functions, it is extremely useful to have the variable names in both the function definition in the spline CPPs and in the function call here, instead of having to separately document what each index in a returned double pointer means.

src/ML-UF3/pair_uf3.cpp
```diff
@@ -0,0 +1,1313 @@
...
+            f[j][2] += Fj[2];
+
+            Fk[0] = fki[0] + fkj[0];
+            Fk[1] = fki[1] + fkj[1];
+            Fk[2] = fki[2] + fkj[2];
+            f[k][0] += Fk[0];
+            f[k][1] += Fk[1];
+            f[k][2] += Fk[2];
+
+            if (eflag) evdwl = *triangle_eval;
```
srtee: This is where a call-by-reference method can be much clearer. For consistency with the 2body call I'd make `eval` still return `evdwl` by reference, so this `eval` has to be `void` (there isn't a natural particular double to return). But you could call it as:

```
UFBS3b[itype][jtype][ktype].eval(rij, rik, rjk,
Fmag_ij, Fmag_ik, Fmag_jk, evdwl)
```

and voila! Each of your return variables are now named, not numbered!

src/ML-UF3/pair_uf3.cpp
srtee: You could even afford to store the delx, dely, and delz doubles in their own arrays as well.

monk-04: I agree storing an array of rij will reduce the number of computations later on.
But I also need to know the type of atoms in the 3b interaction evaluation. So I need to keep the current `neighshort`.

Also, in practice, the 3b cutoffs are always shorter than 2-body cutoffs (usually by a couple of angstroms). Do you still think I need to make the `rij_list` of the size of `inum`?

src/ML-UF3/pair_uf3.cpp
monk-04: Should I also pass eflag to the eval function? Something like `UFBS2b[itype][jtype].eval(rij, evdwl, eflag)`? This will save some computations as energy evaluation is not needed on every timestep?

src/ML-UF3/pair_uf3.cpp
monk-04: I am passing n3b_coeff_matrix[key] by reference to 3b bspline object (triplet_bspline UFBS3b), which then uses it to calculate energies and forces

src/ML-UF3/pair_uf3.cpp
srtee: If the energy calculation is costly enough, then yes, you can pass eflag and let `eval` save some time. But you would know better whether that's worth it.

As a style thing I'd prefer to keep the reference doubles together after the value doubles:

```
double CLASSNAME::eval(
  double rij,
  int eflag,
  double& evdwl // return energy by ref
){ ...
```

akohlmey: > Should I also pass eflag to the eval function? Something like `UFBS2b[itype][jtype].eval(rij, evdwl, eflag)`? This will save some computations as energy evaluation is not needed on every timestep?

since eflag doesn't change, the most efficient way to handle this would be to make eval a templated function and eflag a template argument. That could hoist the if statement to the outermost level and the if statement will be optimized away since you either call `eval<0>()` or `eval<1>()`. The compiler will then optimize the "if" away since the condition is a compile time constant. Check out examples in OPT or OPENMP.

cmake/CMakeLists.txt
akohlmey, line 1: There should be updates for the preset files. Included in my PR to your branch.

src/ML-UF3/pair_uf3.cpp
srtee: My main worry is a bunch of unpredictable memory realllocations happening inside the inner loop of a force calculation. The `if(pot3b) ...` code is being triggered `nlocal * jnum` times and all it takes is one processor needing to do a `grow` to imbalance that timestep's force calculation.

Types are straightforward to track: if `jj` runs from 0 to `jnum`, the distance `rij_list[jj]` belongs to particle `jlist[jj]` which has particle type `type[jlist[jj]]`.

potentials/A_A_A.uf3_pot
akohlmey, line 1: What element does A stand for?

doc/src/pair_uf3.rst
```diff
@@ -0,0 +1,184 @@
...
+   pair_coeff 1 2 A_B
+   pair_coeff 2 2 B_B
+   pair_coeff 3b 1 1 1 A_A_A
+   pair_coeff 3b 1 1 2 A_A_B
+   pair_coeff 3b 1 2 2 A_B_B
+   pair_coeff 3b 2 1 1 B_A_A
+   pair_coeff 3b 2 1 2 B_A_B
+   pair_coeff 3b 2 2 2 B_B_B
+
+If a value of "2" is specified in the :code:`pair_style uf3` command, only the two-body potential files are needed. For 3-body interaction the first atom type is the central atom. We recommend using the :code:`generate_uf3_lammps_pots.py` script (found `here <https://github.com/uf3/uf3/tree/master/lammps_plugin/scripts>`_) for generating the UF3 LAMMPS potential files from the UF3 JSON potentials.
```
akohlmey: If possible this script should be added to the `tools` tree in `tools/uf3`.

doc/src/pair_uf3.rst
```diff
@@ -0,0 +1,184 @@
...
+   COEFF
+   #
+
+The second line indicates whether the potential file contains data for 2-body (:code:`2B`) or 3-body (:code:`3B`) interaction. This is followed by :code:`LEADING_TRIM` and :code:`TRAILING_TRIM` number on the same line. The current implementation is only tested for :code:`LEADING_TRIM=0` and :code:`TRAILING_TRIM=3`. If other values are used LAMMPS is terminated after issuing an error message. The :code:`Rij_CUTOFF` sets the 2-body cutoff for the interaction described by the potential file. :code:`NUM_OF_KNOTS` is the number of knots (or the length of the knot vector) present on the very next line. The :code:`BSPLINE_KNOTS` line should contain all the knots in ascending order. :code:`NUM_OF_COEFF` is the number of coefficients in the :code:`COEFF` line. All the numbers in the BSPLINE_KNOTS and COEFF line should be space-separated.
+
+The format of a generic 3-body UF3 LAMMPS potential file is as follow-
+
+.. code-block:: LAMMPS
+
+   #UF3 POT
```
akohlmey: The potential files should have metadata tags that indicate the source, a reference, the data, and in particular can indicate the units, so that LAMMPS will refused to read potential files created for incompatible units.
You are already using `utils::open_potential()` so this check will be automatic, provided the metadata tags are present.

src/ML-UF3/pair_uf3.cpp
```diff
@@ -0,0 +1,1313 @@
...
+        UFBS3b[i][j].resize(num_of_elements + 1);
+      }
+    }
+    memory->create(neighshort, maxshort, "pair:neighshort");
+  }
+}
+
+void PairUF3::uf3_read_pot_file(int itype, int jtype, char *potf_name)
+{
+  utils::logmesg(lmp, "UF3: {} file should contain UF3 potential for {} {}\n", \
```
akohlmey: This will print one line per MPI ranks. Not a good idea. Since the number of pair_coeff lines for systems with multiple elements can be large, it is probably best to drop this output altogether.

src/ML-UF3/pair_uf3.cpp
```diff
@@ -0,0 +1,1313 @@
...
+          potf_name, itype, jtype);
+
+  if (!platform::file_is_readable(potf_name))
```
akohlmey: This is redundant you can check for the file pointer being NULL below.

src/ML-UF3/pair_uf3.cpp
```diff
@@ -0,0 +1,1313 @@
...
+    error->all(FLERR, "UF3: {} file is not readable", potf_name);
+
+  FILE *fp;
+  fp = utils::open_potential(potf_name, lmp, nullptr);
```
akohlmey: these two lines could be merged and the fest for failure should be added below.

src/ML-UF3/pair_uf3.cpp
```diff
@@ -0,0 +1,1313 @@
...
+
+  TextFileReader txtfilereader(fp, "UF3:POTFP");
+  txtfilereader.ignore_comments = false;
+
+  std::string temp_line = txtfilereader.next_line(1);
+  Tokenizer file_header(temp_line);
+
+  if (file_header.count() != 2)
+    error->all(FLERR, "UF3: Expected only two words on 1st line of {} but found \n\
```
akohlmey: The trailing backslash is rather unusual. Is fixed in my PR branch. also, there don't need to be so many newlines.

akohlmey (changes requested):
This is it for a first round of review. I didn't change most of the unprotected utils::logmesg() calls. In my opinion most should just be removed. They look like debugging output.

One more general comment. It is generally considered a bad idea to read the same file from all the MPI ranks. This can create huge problems with jobs using many MPI processes. The more common procedure is to read everything on rank 0 only and then use MPI communication to broadcast the information to all other ranks.

src/ML-UF3/pair_uf3.cpp
monk-04: > Types are straightforward to track: if `jj` runs from 0 to `jnum`, the distance `rij_list[jj]` belongs to particle `jlist[jj]` which has particle `type type[jlist[jj]]`.

I am bit confused.
`jnum` is decided by the cutoff of 3body interaction and, in practice, is always less than 2body.

In `neighshort`, I am storing `j` ie `jlist[jj]` during the 2body loop. Not every `j` in 2body loop forms a valid 3body triangle. 

So, if my understanding of your suggestion is correct, you are suggesting to set `jnum` of 3body loop equal to `jnum` of 3body loop(?). If that is done, a lot of iterations in the 3body loop will do nothing because of the distance  cutoff checks. 

I apologize if I am not understanding your suggestion correctly.

monk-04:
@akohlmey Thank you for your comments and edits.

> It is generally considered a bad idea to read the same file from all the MPI ranks

I will modify the code so that only rank 0 will read the potential file.


>utils::logmesg() 

Many of the logmesg were put in the code at the request of early users of the `pair_style`. I will think about this point a bit more.

- [ ] Pimplify pair_uf3.cpp
- [ ] Read UF3 potential files on rank 0 and use MPI communicate to other processors

Currently the unittest is failing because of the following error-

`LAMMPS Error: ERROR on proc 0: Cannot open UF3 potential file A_A.uf3 (src/ML-UF3/pair_uf3.cpp:215)
Last command: pair_coeff * * A_A.uf3`

I think this is because the `A_A.uf3` and `A_A_A.uf3` potential files were moved to the `potentials` directory.

akohlmey:
> I think this is because the `A_A.uf3` and `A_A_A.uf3` potential files were moved to the `potentials` directory.

No. It is because you didn't close the files. The unit test runs will execute multiple short runs and after several of them, you run out of available file descriptors. That is when the open failed. I'll improve the error message.

akohlmey:
I've submitted https://github.com/uf3/lammps/pull/2 with more changes including the file handle leak fix.

akohlmey:
> * Read UF3 potential files on rank 0 and use MPI communicate to other processors

Here are more things to consider:
- add metadata to UF3 potential files so that units can be checked
- use a map so that unique potential files and its parameters need to be read only once. For example in the unit test input you need to read the same A_A_A.uf3 potential file 512 times and store 512 copies of the same parameters. with a map only one read and one copy would be needed. Check out other manybody potentials like Tersoff where they have a elem3param list
- consider moving away from the vector of vector (of vector) data storage toward multi-dimensional arrays created with the memory class. These are much easier to handle for broadcasting data. You just need to broadcast the dimensions (if they are not already known), destroy then create the arrays and broadcast the data in fortran style with a single  MPI_Bcast() of &array[0][0] or similar since the data holding block is consecutive like in fortran. There are examples for that all over the LAMMPS code.

akohlmey:
> I've submitted [uf3#2](https://github.com/uf3/lammps/pull/2) with more changes including the file handle leak fix.

You need to merge this pull request to have it pass the tests here.

monk-04:
I am running some numerical accuracy (comparison with Python) and speed tests on our systems. They should be over in some time.

monk-04:
I have merged the PR.

Next I will work on the following things-

- [x] Pimplify pair_uf3.cpp
- [x] Add `UNITS`, `AUTHOR`, `DATE` and `CITATION` metadata
- [x] Read all UF3 potentials from a single file
- [x] Use arrays from memory class instead of std::vectors
- [x] Read UF3 potential files on rank 0 and use MPI to communicate to other processors
- [ ] ~~Modify `eval` of 2body bspline classes to take  `rij`, `evdwl`, and `eflag` as input and return only `fpair`~~
- [ ] ~~Modify `eval` of 3body bspline classes to also take `Fmag_ij`, `Fmag_ik`, `Fmag_jk`, `evdwl`, and `eflag` as input~~

potentials/A_A_A.uf3_pot
monk-04: The potential files were modeled after Niobium UF3 potential files. I will replace them with the real files.

doc/src/pair_uf3.rst
monk-04: The python script depends on the `uf3` python package and is not made to work out of the box. Should I still add it?

doc/src/pair_uf3.rst
monk-04: Yup, working on it

doc/src/pair_uf3.rst
akohlmey: No that would be overkill.

monk-04:
@akohlmey Thank you for your suggestion about moving away from std::vectors. It forced me to re-write certain functions. Now arrays are used to store constants of bspline basis set instead of storing uf3_pair_bspline and uf3_triplet_bspline objects (at UFBS2b std::vector\<std::vector\> and UFBS3b std::vector\<std::vector\<std::vector\>\>). Also, I evaluate the energy contribution only when 'eflag' is true instead of every time step.

All these changes led to about 30%-35% speed up for the CPU version.

stanmoore1:
@monk-04 is this PR ready for review or are more changes coming?

monk-04:
@stanmoore1 I was waiting for PR #4146 so that the MacOS unit test could pass.
I see that it has been merged and will merge lammps:develop into ml-uf3.

Other than that, I don't have any more changes in mind.

The PR is ready for review.

akohlmey:
@monk-04 I've submitted a pull request with a few minor changes and replaced the example.

akohlmey (approved):

stanmoore1:
@monk-04 I took a look at the Kokkos code (thanks for adding that btw), it looks pretty good. I have a few cleanup changes that I will submit as a PR to your fork. The Kokkos version does give correct numerics, but doesn't saturate the GPU with only a couple hundred atoms in the UF3 example. If you plan to run with small atom counts/GPU, it could be beneficial to thread over both atoms and neighbors as done here to expose more parallelism: https://github.com/lammps/lammps/blob/develop/src/KOKKOS/pair_kokkos.h#L604-L780. Otherwise you will need ~500k or more atoms/GPU to saturate, based on other potentials I've run.

monk-04:
@stanmoore1 Thank you for looking through the code.

The current code needs a very large system size to saturate the GPU-
image
(Run on 1 A100)


I will look in to the code you suggested, and will try to implement it.

akohlmey:
@monk-04 we're merging this pull request now, so it will be in the next feature release and thus also the stable release. You can create a new pull request with optimizations later when you have them ready. If we would wait for them, chances are that your contribution would have to be postponed until after the stable release.

monk-04:
@akohlmey That works for me. I am working on my thesis, so any more optimizations will take a bit more time from my side. Thank You.

akohlmey:
@monk-04 FYI, we may have to pull the KOKKOS version of your pair style from the distribution until it is corrected. It is missing the "host" version of the implementation and fails compilation under certain circumstances.  E.g. I just ran into issues trying to compile with OpenMP and HIP (for an AMD GPU).

Ping me on the LAMMPS developer slack for more details.

monk-04:
Ok. I will message you on Slack.

PR #4111, opened 2024-03-27
Collected small fixes

akohlmey:
**Summary**

This pull request contains multiple small changes and fixes

**Related Issue(s)**

Fixes #4114

**Author(s)**

Axel Kohlmeyer, Temple U and Trung Nguyen, U Chicago

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

Output for some computes are changing due to reordering of eigenvalues and eigenvectors in jacobi3 function

**Implementation Notes**

The following individual changes are implemented:
- correct formatting of tables in documentation of "variable" command
- includes fix to fix ave/histo/weight and thus closes #4116 
- pimplify xtc dump style to not require xdr_compat.h include in header file
- update XDR code with BSD licensed version and remove parts not used by dump xtc
- add optional argument to `jacobi3()` function so one can override the default choice for ordering eigenvalues and eigenvectors
- simplify output handling when Windows build is not recognized
- some updates to the Errors_details documentation and use second link for format errors
- Fix bug in Kokkos when shrink-wrapping with no atoms leading to a bogus error

**Post Submission Checklist**

- [ ] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [x] The source code follows the LAMMPS formatting guidelines
- [x] Suitable new documentation files and/or updates to the existing docs are included
- [x] The added/updated documentation is integrated and tested with the documentation build system
- [x] The feature has been verified to work with the conventional build system
- [x] The feature has been verified to work with the CMake based build system
- [x] Suitable tests have been added to the unittest tree.

PR #4112, opened 2024-03-28
GPU package has a hard cufft dependency when GPU_API=CUDA

rbberger:
**Summary**

CMake change was missing to add linking to cuFFT in the GPU package when using CUDA.

**Related Issue(s)**

**Author(s)**

@rbberger 

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/cmake/Modules/Packages/GPU.cmake
+++ b/cmake/Modules/Packages/GPU.cmake
@@ -188,6 +188,9 @@ if(GPU_API STREQUAL "CUDA")
     target_compile_definitions(gpu PRIVATE -DUSE_CUDPP)
   endif()
 
+  find_package(CUDAToolkit REQUIRED)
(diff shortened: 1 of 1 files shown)


akohlmey:
This must be bogus.
If you look through the code in lib/gpu, only the amoeba support files reference cuFFT and in the .cpp file all code related to cufft is disabled. So how should there be a dependency?

There must be some part where this was done incompletely, possibly in the header.

stanmoore1:
Right, the GPU package does not support FFTs on the GPU, it is all done on the host CPU.

rbberger:
The header does the include and fails compilation. If the cpp parts are disabled, probably the header should also follow suit.

akohlmey:
should be fixed in https://github.com/lammps/lammps/pull/4111/commits/a13d53dd65a435df205a727f77efe752c1f8c986

PR #4116, opened 2024-03-30
fix_ave_histo_weight bugfix

jrgissing:
**Summary**

tricky reference bug. pointer to second value (the weights for fix/ave/histo/weight) was getting overwritten after 'step 0' in the simulation

caused incorrect plots, for example, for compute xrd, after the first 'step 0' plot

Thanks to Joshua Kemppainen and Ivan Gallegos (Michigan Tech) for reporting and testing

**Related Issue(s)**

notably, compute xrd could appear to produce incorrect results since the introduction of this bug, when using this fix to output xrd values

**Author(s)**

Jake

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

- [x] Suitable tests have been completed

### DIFF
--- a/src/fix_ave_histo_weight.cpp
+++ b/src/fix_ave_histo_weight.cpp
@@ -113,7 +113,7 @@ void FixAveHistoWeight::end_of_step()
   double weight = 0.0;
   double *weights = nullptr;
   int stride = 0;
-  auto &val = values[1];
+  auto val = values[1];
   int j = val.argindex;
 
   // atom attributes
--- a/src/fix_ave_histo_weight.cpp
+++ b/src/fix_ave_histo_weight.cpp
@@ -38,7 +38,7 @@ enum { DEFAULT, GLOBAL, PERATOM, LOCAL };
 enum { IGNORE, END, EXTRA };
 enum { SINGLE, VALUE };
 
-#define BIG 1.0e20
+static constexpr double BIG = 1.0e20;
 
 /* ---------------------------------------------------------------------- */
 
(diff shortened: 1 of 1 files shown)


akohlmey:
@jrgissing I am not sure that making a copy instead of using the reference is a good solution since `value_t` is a struct without an explicit copy constructor and assignment operator that doesn't just use primitive data types.
Can you please try out the alternate approach in commit f6bdc05c242c39adf35e6acfa08674a6e27383f3?

jrgissing:
this update looks good to me, thanks!

jrgissing:
@akohlmey to give a little credit to those who reported, tested, tracked down and corrected bugs like these, I think we should try to 'merge' rather than 'close' such contributions whenever possible

akohlmey:
Please see the summary of #4111  
This is exactly the same thing we do for any bug report issue. Only that this bug report came in the form of a pull request.

Without this statement, GitHub would have noticed that there were no more changes relative to develop and would have closed it automatically without any reference.

jrgissing:
This was not a bug report issue, it was also a bugfix. What I am saying is, I am concerned that contributors who take the time to help out the core developers by tracking down (!) bugs, such as this subtle memory bug, will be less motivated to do so without getting some small credit in the form of their PR being merged into LAMMPS

stanmoore1:
>This is exactly the same thing we do for any bug report issue. Only that this bug report came in the form of a pull request.

I think @jrgissing's PR here is a little different than a regular bug report since he provided a working bugfix in his commit [1c5a105](https://github.com/lammps/lammps/pull/4116/commits/1c5a1053610897383a02c155520600d8772f2e13). Then @akohlmey improved Jake's solution, but it was a relatively small jump to [f6bdc05](https://github.com/akohlmey/lammps/commit/f6bdc05c242c39adf35e6acfa08674a6e27383f3) using Jake's idea. When [f6bdc05](https://github.com/akohlmey/lammps/commit/f6bdc05c242c39adf35e6acfa08674a6e27383f3) was merged in #4111, there is now zero record of @jrgissing's contribution in the Git history, so I can understand him feeling shortchanged...

@akohlmey I think a better workflow would be to either push changes to this PR instead of #4111, or do a `co-authored` commit from both @jrgissing and @akohlmey: https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors. Then at least Jake is recognized in the Git history. The disadvantage of a co-authored commit instead of a PR merge is that it is less likely to get recognized in the release notes of a feature release. However this is a pretty minor change.

@jrgissing I agree that contributors need to get credit, otherwise they won't be as motivated. However, I don't think that a PR necessarily needs to be merged instead of closed for a single commit if a co-authored commit is used. I think the reason that @akohlmey bundles small PRs into a single PR is that it can reduce the work of reviewing many small, separate PRs.

@akohlmey I think it may be good to write down a policy for this situation in our developer documentation, so that contributors know what to expect.

akohlmey:
@stanmoore1 this would be a good topic for the developer zoom. Perhaps Jake can join, if he has time. Just let Aidan know.

PR #4121, opened 2024-04-04
Fix issue with nearly linear 2d rigid bodies

sjplimp:
**Summary**

When a 2d rigid body is nearly linear it has 1 eigen vector (with a large evalue) along the z-axis and two in the xy plane.
One of those two will be along the rigid body and its eigenvalue can be tiny.  Unless this eigenvector is last, the quaternion math can exhibit round-off issues which accumulate over timestepping and induce incorrect dynamics for the body.

The solution is to add an option to the jacobi3() method to request eigenvalues be returned in ascending order (not the default).
If the z-axis eigenvector is last, then the resulting quaternion represents a simple 2d rotation (in the xy plane) and has no 
round-off issues.  If there are 2 large and equal eigenvalues, then the z-axis one may not be last.  We then reorder the eigen values/vectors to force the z-axis eigenvector to be last.

**Related Issue(s)**

**Author(s)**

Trung Nguyen (U Chicago) and Steve Plimpton

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**


sjplimp:
@ndtrung81 Please take a look and see if this final version is what you expect.

src/RIGID/fix_rigid.cpp
```diff
@@ -1986,6 +1983,22 @@ void FixRigid::setup_bodies_static()
     ez_space[ibody][1] = evectors[1][2];
     ez_space[ibody][2] = evectors[2][2];
 
+    // for 2d, ensure that evector along z axis is last
+    // necessary so that quaternion is a simple rotation around +z axis
+    //   or a 180 degree rotation for a -z axis
+    // otherwise richardson() method for a body with a tiny evalue (near-linear)
+    //  may not preserve the correct z-aligned quat and assocated evectors
```
ndtrung81: typo: `assocated` --> `associated`

src/RIGID/fix_rigid_small.cpp
```diff
@@ -2134,6 +2131,22 @@ void FixRigidSmall::setup_bodies_static()
     ez[1] = evectors[1][2];
     ez[2] = evectors[2][2];
 
+    // for 2d, ensure that evector along z axis is last
+    // necessary so that quaternion is a simple rotation around +z axis
+    //   or a 180 degree rotation for a -z axis
+    // otherwise richardson() method for a body with a tiny evalue (near-linear)
+    //  may not preserve the correct z-aligned quat and assocated evectors
```
ndtrung81: a typo: `assocated` --> `associated`

src/RIGID/fix_rigid_small.cpp
```diff
@@ -1110,11 +1110,6 @@ void FixRigidSmall::enforce2d()
     b->angmom[1] = 0.0;
     b->omega[0] = 0.0;
     b->omega[1] = 0.0;
-    if (langflag && langextra) {
```
ndtrung81, line 1113: I think we still need this part because FixRigidSmall doesn't support tflag and fflag (whereas FixRigid does). In the example in.micelle-rigid, the fix rigid/small command does not have the langevin option.

ndtrung81:
@sjplimp thanks! I only have a question on the langextra components for enforce2d in FixRigidSmall and find a typo.

sjplimp:
made the spelling change

I removed the clearing of langextra in enforce2d, b/c I think it is not needed in either rigid or rigid/small.  The Langevin forces/torques have already been used in compute_force_torque(), so zeroing the forces and torque components in enforce2d() is sufficient I think?

ndtrung81:
yes, you are right, `enforce2d()` is always invoked after `compute_forces_and_torques()` so no need to zero `langextra` in `enforce2d()`.

ndtrung81 (approved):

PR #4123, opened 2024-04-05
Address minimum image issue when restarting fix rigid/small

### LINKED ISSUE #4109: [BUG] _Unable to restart from restart files with rigid molecules.
**Summary**

When Molecules with rigid bonds are in the simulation it is not possible to restart the calculation
ERROR on proc 0: Atoms have moved too far apart (1699.9999999999998) for minimum image

I suspect that this has to do with the rigid molecule crossing the periodic boundary. 

21 Nov 2023

**Expected Behavior**
I expect the simualtion to continue from where it stopped.

**Actual Behavior**

I get the following error message:
ERROR on proc 0: Atoms have moved too far apart (1699.9999999999998) for minimum image

**Steps to Reproduce**

The files required to generate this error are on the matsci forum:
See post https://matsci.org/t/fix-rigid-atoms-have-moved-too-far-apart/53926/3

akohlmey:
**Summary**

When restarting fix rigid/small it uses domain->minimum_image() in a way where they may be a large number of box crossings when the simulation was long and the rigid objects mobile or subject to a flow. This now calls an alternate implementation suitable for this specific case

**Related Issue(s)**

Fixes #4109 
Further discussions are here: https://matsci.org/t/fix-rigid-atoms-have-moved-too-far-apart/53926

**Author(s)**

Axel Kohlmeyer, Temple U, Stan Moore SNL, Steve Plimpton

**Licensing**

**Backward Compatibility**

Yes.

**Implementation Notes**

This uses the code from PR #3721 but adds it as a new method `minimum_image_big()`.

**Post Submission Checklist**

### DIFF
--- a/src/RIGID/fix_rigid_small.cpp
+++ b/src/RIGID/fix_rigid_small.cpp
@@ -2160,7 +2160,7 @@ void FixRigidSmall::setup_bodies_static()
     xgc = body[ibody].xgc;
     double delta[3];
     MathExtra::sub3(xgc,xcm,delta);
-    domain->minimum_image(delta);
+    domain->minimum_image_big(delta);
     MathExtra::transpose_matvec(ex,ey,ez,delta,body[ibody].xgc_body);
     MathExtra::add3(xcm,delta,xgc);
   }
--- a/src/domain.cpp
+++ b/src/domain.cpp
@@ -973,6 +973,9 @@ void Domain::subbox_too_small_check(double thresh)
    changed "if" to "while" to enable distance to
      far-away ghost atom returned by atom->map() to be wrapped back into box
      could be problem for looking up atom IDs when cutoff > boxsize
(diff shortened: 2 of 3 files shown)


sjplimp (approved):
I checked that this solves the problem with the input and restart file posted on MatSci

stanmoore1:
One optimization could be to store `1.0/xprd` as `xprd_inv` when the box changes, then we don't have to use division which is more expensive than multiplication.

akohlmey:
> One optimization could be to store `1.0/xprd` as `xprd_inv` when the box changes, then we don't have to use division which is more expensive than multiplication.

The Domain class is a candidate for some major cleanup and refactoring. It stores data in redundant ways. There should be just the minimal amount and then suitable accessors. Adding data like pre-computed inverses could be very helpful for that, too. At that time we could also do some thorough benchmarking, add more unit tests and so on.

PR #4125, opened 2024-04-05
Fix bug in MDI time integration logic

taylor-a-barnes:
**Summary**

This is a very small bugfix for an error in the LAMMPS MDI code.  Normally, when LAMMPS runs as an MDI engine and initiates a molecular dynamics simulation, it should not terminate the molecular dynamics simulation until the driver instructs it to do so.  There is an oversight that can cause LAMMPS to terminate the simulation when it receives the `@` command, which should not cause this behavior.

**Related Issue(s)**

**Author(s)**

Taylor Barnes, MolSSI

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/src/MDI/mdi_engine.cpp
+++ b/src/MDI/mdi_engine.cpp
@@ -695,7 +695,7 @@ void MDIEngine::mdi_md()
   if (strcmp(mdicmd, "EXIT") == 0) return;
 
   // run one step at a time forever
-  // driver triggers exit with @ command other than @COORDS,@FORCES,@ENDSTEP
+  // driver triggers exit with @ command other than @COORDS,@FORCES,@ENDSTEP,@
 
   update->integrate->setup(1);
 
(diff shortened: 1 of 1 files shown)


sjplimp (approved):
thanks @taylor-a-barnes - didn't realize this attribute of the MDI @ command

taylor-a-barnes:
@sjplimp The general expectation for the `@` command is that it should be possible to progress through an entire simulation by sending it many times in sequence, without necessarily needing to know the individual names of the engine's nodes _a priori_.  That makes it easier for a driver to be written in a way that is agnostic to the engine and its specific sequence of nodes.

To be more accurate, what LAMMPS was doing with `@` before isn't strictly a violation of the MDI Standard, but it is less ideal than what is implemented in this PR.  With the PR, a driver can send the `@` as many times as it wishes, to run an arbitrarily long simulation.  Without it, LAMMPS will exit the simulation (and return to the `@DEFAULT` node) after reaching the end of a single timestep.

Issue #4132, opened 2024-04-11
[BUG] Fortran scatter_atoms() does not update atom positions when using atom_style atomic


jwrm2:
**Summary**

I am attempting to use the Fortran interface. While the coordinates are read from the initial file correctly, any attempt to update the coordinates with scatter_atoms does nothing when using atom_style atomic. 

**LAMMPS Version and Platform**

LAMMPS (7 Feb 2024 - Development - patch_7Feb2024_update1-384-g9ca47a36d2)

develop branch commit 9ca47a36d2cf8f6df15424bb09b2d958cf070063 10th April

Debian Linux Bookworm, Intel i9-13900K

Compiler version: GNU Fortran (Debian 12.2.0-14) 12.2.0

LAMMPS built in path-to-lammps/build/ configured with

`cmake ../cmake/  -DBUILD_MPI=no -DBUILD_OMP=no -DPKG_MOLECULE=yes`

Fortran programme compiled with

`gfortran path-to-lammps/fortran/lammps.f90 test.f90 path-to-lammps/build/liblammps.a -lstdc++`

**Expected Behavior**

I would expect that calling scatter_atoms() to update the atom positions should update the atom positions. When atom_style is full, the positions are updated as expected.

**Actual Behavior**

When atom_style is atomic, the atom positions are not updated by scatter_atoms().

**Steps to Reproduce**

Compile LAMMPS with above settings and Fortran file test.f90 as above. Have attached input files in working directory. Run Fortran executable.

**Further Information, Files, and Links**

test.f90 - short Fortran programme to reproduce the bug
```
PROGRAM TEST

    USE LIBLAMMPS
    USE, INTRINSIC :: ISO_C_BINDING, ONLY : C_DOUBLE

    IMPLICIT NONE

    TYPE(LAMMPS) :: LMP
    REAL(KIND=C_DOUBLE), ALLOCATABLE, DIMENSION(:) :: COORDS

    ! Read original coordinates from start_atomic
    LMP = LAMMPS([CHARACTER(LEN=10) :: 'liblammps', '-log', 'lammps.log', '-screen', 'none'])
    CALL LMP%FILE('in_atomic.txt')
    CALL LMP%GATHER_ATOMS('x', 3, COORDS)
    ! Prints 1.0 1.0 1.0 as expected
    PRINT *, COORDS(:)

    ! Try to update coordinates
    COORDS = (/2.0, 2.0, 2.0/)
[...]

PR #4134, opened 2024-04-13

### LINKED ISSUE #3982: documentation and/or error handling for eam/alloy
**Summary**

The documentation on pair_style eam/alloy does not specify how LAMMPS handles situations where the electron density at an atom site exceeds the maximum value specified by the potential file. The code doesn't crash or give an error. I suspect it applies the value of embedding energy at the maximum value of electron density specified in the potential file for all higher values of electron density. I suggest a) changing the code to give an error and exit or b) updating the documentation to make your convention explicit.

23June2022

**Details**

I encountered this issue while computing binding energy curves (cold curves) for FCC Cu using the EAM potential of Mishin et al. (PRB 63, 224106 [2001]). The potential file that comes with LAMMPS (Cu_mishin1.eam.alloy) has a max electron density of 1.6402, which is exceeded at lattice parameters of about 3.21A and below. However, LAMMPS gives values of the cold curve for this potential down to lattice parameters as low as 2.5A (I didn't try any lower). When I substituted the embedding energy data in Cu_mishin1.eam.alloy with the analytical expression in PRB 63, 224106 [2001], I found that the cold curve differs from the one obtained using the potential file for atomic electron densities above 1.6402 (i.e., lattice parameters of about 3.21A and below).

### LINKED ISSUE #4106: [BUG] Small error in in.ttm.mod example (Si.ttm_mod)
**Summary**

γ_p and y_s values are switched with each other.

Ideally, users should be referencing papers when stating values, but I thought I'd report this to clear up confusion. 

Latest/Windows 

**Expected Behavior**

gamma_p, mass/time units
24.443
gamma_s, mass/time units
39.235

**Actual Behavior**
gamma_p, mass/time units
39.235
gamma_s, mass/time units
24.443

T. Jarrin, A. Jay, M. Raine, N. Mousseau, A. Hémeryck and N. Richard, "Simulation of Single Particle Displacement Damage in Si₁₋ₓGeₓ Alloys—Interaction of Primary Particles With the Material and Generation of the Damage Structure," in IEEE Transactions on Nuclear Science, vol. 67, no. 7, pp. 1273-1283, July 2020, doi:

Antoine Jay, Melanie Raine, Nicolas Richard, Normand Mousseau, Vincent Goiffon, et al.. Simulation of Single Particle Displacement Damage in Silicon – Part II: Generation and Long-Time Relaxation of Damage Structure. IEEE Transactions on Nuclear Science, 2017, 64 (1), pp.141-148.
ff10.1109/TNS.2016.2628089ff. ffhal-01407740f

akohlmey:
**Summary**

This pull request combines multiple small changes and bugfixes that do not warrant a pull request of their own

**Related Issue(s)**

Consolidates the library interface to flag errors better to avoid issues like #4132 
Also "Restrictions" boxes have been added to the scatter/gather functions to indicated the problems.
Improves on the bugfix from PR #4135.
Fixes #4106 
Fixes #3982

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

Since `-DLAMMPS_EXCEPTIONS` is now the only option the library functions will work the same, only that the previous warning is stored as error message and can be checked for with `lammps_has_error()`.

**Implementation Notes**

The following individual changes are includes:
- initialize all pointer to null in pair style pedone
- add references to the new typelabel paper in the typelabel howto and the labelmap command
- print warning when embedding energy exceeds the tabulated rho range in EAM potentials. add corresponding note to documentation.
- make `Atom.mass` property in PyLammps correctly return the mass of this atom, either via the per-type mass through the type value or from the per-atom mass using the index.
- turn calls to `Error:warning()` in the scatter/gather functions of the library interface into calls to `Error::all()`. This allows to check for them with `lammps_has_error()`. improve error messages and consolidate them.
- improve documentation for scatter/gather when using python
- update and correct `fix ttm/mod` example
- limit Sphinx version to a version before 7.3.x since those currently break the breathe extension

**Post Submission Checklist**


src/MANYBODY/pair_eam.cpp
```diff
@@ -237,7 +239,9 @@ void PairEAM::compute(int eflag, int vflag)
     fp[i] = (coeff[0]*p + coeff[1])*p + coeff[2];
     if (eflag) {
       phi = ((coeff[3]*p + coeff[4])*p + coeff[5])*p + coeff[6];
-      if (rho[i] > rhomax) phi += fp[i] * (rho[i]-rhomax);
+      if (rho[i] > rhomax) {
```
sjplimp, line 242: did you mean to set or increment beyond_rhomax here ?
Seems like setting it to one would be sufficient.

stanmoore1: Does this need to be ported to Kokkos?

akohlmey: > did you mean to set or increment beyond_rhomax here ?

Yes. No idea how it vanished.

src/MANYBODY/pair_eam.cpp
```diff
@@ -145,6 +146,7 @@ void PairEAM::compute(int eflag, int vflag)
   double *coeff;
   int *ilist,*jlist,*numneigh,**firstneigh;
 
+  int beyond_rhomax = 0;
```
sjplimp: see comment on this line below

src/MANYBODY/pair_eam.cpp
```diff
@@ -424,6 +438,8 @@ void PairEAM::init_style()
 
   neighbor->add_request(this);
   embedstep = -1;
+
+  exceeded_rhomax = 0;
```
sjplimp, line 443: Also initialize beyond_rhomax = 0 here ?  See comment above

src/MANYBODY/pair_eam.cpp
```diff
@@ -322,6 +326,16 @@ void PairEAM::compute(int eflag, int vflag)
     }
   }
 
+  if (eflag && (exceeded_rhomax >= 0)) {
```
sjplimp: Is the intent here to only check for a possible warning on a step where energy is needed (e.g. a thermo output step)?  And then disable the check and the warning message thereafter?  I agree with that strategy
so that Allreduce is not called often. 

However, it then seems like beyond_rhomax should not be zeroed at the top of compute() every step?  Rather it should be initialized to zero in init_style()

src/MANYBODY/pair_eam.cpp
akohlmey: > Does this need to be ported to Kokkos?

Yes, it could be. I have not looked too deeply at implementing it myself (it was too late for KOKKOS).
It is only an additional diagnostic for some rare and obscure use cases, so I am not sure whether it would be needed in KOKKOS. I would expect that people paying attention to the warning would be running without KOKKOS first. I mostly put the code there so that the others that were discussing it in the issue page would see what it could look like. I already set it up so the check is only on energy output and thus the collective communication is not in every step by default.
I would be happy to revert it and only keep the warning in the manual.

src/MANYBODY/pair_eam.cpp
akohlmey: `beyond_rhomax` is a local variable in the compute method and that is either 0 or 1 depending on whether `rho[i]` has exceeded rhomax.

`exceeded_rhomax` is a per-class instance variable that is set to 0 at the beginning of each run and then becomes 1 during the MPI_Allreduce() when `beyond_rhomax` is 1 for at least one of the MPI ranks. If `exceeded_rhomax` is 1 after the reduction, the warning is printed and then `exceeded_rhomax` is set to -1 so it won't flag the same warning during the run again.

akohlmey: Yes, the intent is to only check on steps with energy output (which is often the first step) to reduce the frequency of the MPI_Allreduce().

sjplimp:
@akohlmey So say thermo energy is output every 1000 steps.  And rhomax is not exceeded on step 0, but is on step 452 and step 500.  Then I think we want the next check on step 1000 to warn that rhomax was exceeded (sometime since the last check).  So beyond_rhomax needs to be initialized to 0 in init_style, but not in compute().  Otherwise you only get a warning if rhomax happened to be exceeded on a thermo output step.

akohlmey:
> Otherwise you only get a warning if rhomax happened to be exceeded on a thermo output step.

That was my intention. I didn't want to code up an elaborate check that requires me to define even more variables and write more code to take care of every eventuality. 
If rhomax is exceeded only a few times, it is not much of a problem. Only if it happens all the time, e.g. for a high-pressure system.

sjplimp:
@akohlmey I see now that the check for rho[i] > rhomax is only done on energy steps, so I left beyond_rhomax as a non-class variable.  Just adjusted the error message and logic with exceeded_rhomax.  

I also asked @demkowicz about not checking for this condition on non-energy steps.  Either way, I think it is fine to merge this PR as-is.

doc/src/pair_eam.rst
```diff
@@ -140,6 +140,20 @@ The OpenKIM Project at
...
 
+.. warning::
+
+   The EAM potential files tabulate the embedding energy as a function
+   of the local electron density :math:`\rho`.  When atoms get too
+   close, this electron density may exceed the range for which the
+   embedding energy was tabulated for.  For simplicity and to avoid
+   errors during equilibration of randomized geometries, LAMMPS will
+   assume a linearly increasing embedding energy for electron densities
+   beyond the maximum tabulated value.  This usually means that the EAM
```
jrgissing: Suggest rewording last three sentences to "LAMMPS will print a single warning if the maximum tabulated value is exceeded, which usually indicates that the EAM model is not a good model for the kind of system under investigation.  It may be harmless at the beginning of an equilibration but is a big concern for accuracy if it happens during production runs."

src/MANYBODY/pair_eam.cpp
stanmoore1: I don't think it is worth it to port this to Kokkos.

stanmoore1 (approved):

PR #4142, opened 2024-04-18
flag output for compute count/type as intensive vs extensive

jrgissing:
**Summary**

flag output for compute count/type as intensive so that non-negative integers are printed by the 'thermo' command when counting atom, bond, etc. types when using LJ units

**Related Issue(s)**

none

**Author(s)**

Jake

**Licensing**

**Backward Compatibility**

No. Previously, atom, bond, angles, dihedral and improper counts were normalized by natoms when using LJ units.

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/doc/src/compute_count_type.rst
+++ b/doc/src/compute_count_type.rst
@@ -12,7 +12,7 @@ Syntax
 
 * ID, group-ID are documented in :doc:`compute <compute>` command
 * count/type = style name of this compute command
-* mode = {atom} or {bond} or {angle} or {dihedral} or {improper}
+* mode = *atom* or *bond* or *angle* or *dihedral* or *improper*
 
 Examples
 """"""""
@@ -69,29 +69,29 @@ for each type:
 
 ----------
 
-If the {mode} setting is {atom} then the count of atoms for each atom
+If the *mode* setting is *atom* then the count of atoms for each atom
 type is tallied.  Only atoms in the specified group are counted.
 
-If the {mode} setting is {bond} then the count of bonds for each bond
+If the *mode* setting is *bond* then the count of bonds for each bond
 type is tallied.  Only bonds with both atoms in the specified group
 are counted.
 
-For {mode} = {bond}, broken bonds with a bond type of zero are also
+For *mode* = *bond*, broken bonds with a bond type of zero are also
 counted.  The :doc:`bond_style quartic <bond_quartic>` and :doc:`BPM
 bond styles <Howto_bpm>` break bonds by doing this.  See the :doc:`
 Howto broken bonds <Howto_broken_bonds>` doc page for more details.
 Note that the group setting is ignored for broken bonds; all broken
 bonds in the system are counted.
 
-If the {mode} setting is {angle} then the count of angles for each
+If the *mode* setting is *angle* then the count of angles for each
 angle type is tallied.  Only angles with all 3 atoms in the specified
 group are counted.
 
-If the {mode} setting is {dihedral} then the count of dihedrals for
+If the *mode* setting is *dihedral* then the count of dihedrals for
 each dihedral type is tallied.  Only dihedrals with all 4 atoms in the
 specified group are counted.
 
-If the {mode} setting is {improper} then the count of impropers for
+If the *mode* setting is *improper* then the count of impropers for
 each improper type is tallied.  Only impropers with all 4 atoms in the
 specified group are counted.
 
@@ -101,11 +101,11 @@ Output info
 """""""""""
 
 This compute calculates a global vector of counts.  If the mode is
-{atom} or {bond} or {angle} or {dihedral} or {improper}, then the
+*atom* or *bond* or *angle* or *dihedral* or *improper*, then the
 vector length is the number of atom types or bond types or angle types
 or dihedral types or improper types, respectively.
 
-If the mode is {bond} this compute also calculates a global scalar
+If the mode is *bond* this compute also calculates a global scalar
 which is the number of broken bonds with type = 0, as explained above.
 
 These values can be used by any command that uses global scalar or
--- a/doc/src/compute_count_type.rst
+++ b/doc/src/compute_count_type.rst
@@ -78,8 +78,8 @@ are counted.
 
 For *mode* = *bond*, broken bonds with a bond type of zero are also
 counted.  The :doc:`bond_style quartic <bond_quartic>` and :doc:`BPM
-bond styles <Howto_bpm>` break bonds by doing this.  See the :doc:`
-Howto broken bonds <Howto_broken_bonds>` doc page for more details.
+bond styles <Howto_bpm>` break bonds by doing this.  See the
+:doc:`Howto broken bonds <Howto_broken_bonds>` doc page for more details.
 Note that the group setting is ignored for broken bonds; all broken
 bonds in the system are counted.
 
--- a/doc/src/compute_count_type.rst
+++ b/doc/src/compute_count_type.rst
@@ -80,7 +80,7 @@ atoms like so:
    compute typevec all count/type atom # number of atoms of each type
    variable normtypes vector c_typevec/atoms # divide by total number of atoms
    variable ntypes equal extract_setting(ntypes) # number of atom types
-   thermo_style custom step v_normtypes[*$(v_ntypes)] # vector variable needs upper limit
+   thermo_style custom step v_normtypes[*${ntypes}] # vector variable needs upper limit
 
 Similarly, bond counts can be normalized by the total number of bonds.
 The same goes for angles, dihedrals, and impropers (see below).
--- a/doc/src/compute_count_type.rst
+++ b/doc/src/compute_count_type.rst
@@ -112,7 +112,8 @@ These values can be used by any command that uses global scalar or
 vector values from a compute as input.  See the :doc:`Howto output
 <Howto_output>` page for an overview of LAMMPS output options.
 
-The scalar and vector values calculated by this compute are "extensive".
+The scalar and vector values returned by this compute are non-negative
+integers.
 
 Restrictions
 """"""""""""
--- a/doc/src/compute_count_type.rst
+++ b/doc/src/compute_count_type.rst
@@ -72,6 +72,19 @@ for each type:
 If the *mode* setting is *atom* then the count of atoms for each atom
 type is tallied.  Only atoms in the specified group are counted.
 
+The atom count for each type can be normalized by the total number of
+atoms like so:
+
+.. code-block:: LAMMPS
+
+   compute typevec all count/type atom # number of atoms of each type
+   variable normtypes vector c_typevec/atoms # divide by total number of atoms
+   variable ntypes equal extract_setting(ntypes) # number of atom types
+   thermo_style custom step v_normtypes[*$(v_ntypes)] # vector variable needs upper limit
+
+Similarly, bond counts can be normalized by the total number of bonds.
+The same goes for angles, dihedrals, and impropers (see below).
+
 If the *mode* setting is *bond* then the count of bonds for each bond
 type is tallied.  Only bonds with both atoms in the specified group
 are counted.
@@ -112,8 +125,8 @@ These values can be used by any command that uses global scalar or
 vector values from a compute as input.  See the :doc:`Howto output
 <Howto_output>` page for an overview of LAMMPS output options.
 
-The scalar and vector values returned by this compute are non-negative
-integers.
+The scalar and vector values calculated by this compute are both
+"intensive".
 
 Restrictions
 """"""""""""
--- a/.github/workflows/unittest-macos.yml
+++ b/.github/workflows/unittest-macos.yml
@@ -15,7 +15,7 @@ jobs:
   build:
     name: MacOS Unit Test
     if: ${{ github.repository == 'lammps/lammps' }}
-    runs-on: macos-latest
+    runs-on: macos-13
     env:
       CCACHE_DIR: ${{ github.workspace }}/.ccache
 
--- a/.github/workflows/unittest-macos.yml
+++ b/.github/workflows/unittest-macos.yml
@@ -43,8 +43,10 @@ jobs:
       working-directory: build
       run: |
         ccache -z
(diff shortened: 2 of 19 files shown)


akohlmey (commented):
This change makes this compute behave inconsistently with other computes. The property it computes *is* extensive and thus with "thermo_modify norm yes" its results *should* be divided by the number of atoms.

If this is not desired then "thermo_modify norm no" should be used (or required?). With reduced units "thermo_modify norm yes" is the default otherwise not.

jrgissing:
'compute count/type' feels like a natural exception. for example, other fixes that change types and count events (fix bond_break/create/react, atom/swap, etc.) report values as intensive, even though these values are technically extensive

sjplimp:
First, the title and summary of this PR is misleading.  It is not about integers vs floating point.  It is about flagging the compute's output as extensive versus intensive.  The values the compute outputs are the same regardless, just counts of atoms (or bonds), etc.  And the only thing that pays attention to extensive vs intensive is thermo output (and fix ave/time for labeling its output as ext vs int).

I think for non-LJ units, the thermo output is what the user expects - the raw counts from compute count/type.

For LJ units, the outputs will be fractions (divided by natoms), which is likely not what the user expects.
It's worse if bond (angle, etc) types are being counted, b/c the normalization is by natoms, not total bond count.

The user can set thermo_modify norm no for LJ units, but then many of the other LJ unitless quantities (e.g. energies)
will be output as extensive as well, which I don't think is desirable.

So I think it is best to output raw counts for LJ units as well.  The easy way to do that is change compute count/type to intensive as in this PR.  

And explain in its doc page that it is technically an extensive quantity, but setting it intensive allows thermo output of the raw counts with all units, including LJ.  For any unit system, if the user wants fractions, they can define a vector-style variable which divides all the counts by natoms or nbonds or whatever they desire.

I don't know if we do consistently this for all fixes and computes.  But we do it in the MC fixes (counts of events) which Jake is correctly changing the doc pages for.  We also do it in fix hyper/local (sum of bias potential over biased atoms) which is technically extensive, but we flag it intensive.  Its doc page is wrong, similar to the changed doc pages in this PR.

sjplimp:
Q for @akohlmey - I was looking to see how fix shake flagged its scalar output.  I don't see that it sets extscalar = 0 or 1.  I'm thinking this is fine when the output energy is just added to total energy.  But if a user explicitly output the SHAKE energy as a thermo field, would it potentially be a memory access bug in Thermo::compute_fix() ?

jrgissing:
was it on purpose that the 'extensive flags' were not set to defaults in compute.cpp and fix.cpp?

sjplimp:
@akohlmey Should have added that I also don't see that extscalar is set to a default value in fix.cpp

jrgissing:
I guess we can set as default intensive?

sjplimp:
oops @jrgissing - my comment just crossed with yours - I think it is intentional that no default is set b/c any fix that enables scalar_flag = 1 should set it.  But I want to see what @akohlmey says about this

doc/src/compute_count_type.rst
```diff
@@ -101,18 +101,19 @@ Output info
...
-If the mode is {bond} this compute also calculates a global scalar
+If the mode is *bond* this compute also calculates a global scalar
 which is the number of broken bonds with type = 0, as explained above.
 
 These values can be used by any command that uses global scalar or
 vector values from a compute as input.  See the :doc:`Howto output
 <Howto_output>` page for an overview of LAMMPS output options.
 
-The scalar and vector values calculated by this compute are "extensive".
+The scalar and vector values returned by this compute are non-negative
```
sjplimp: Just say they are both intensive.  I suggest also adding a couple lines up above which illustrate what to do if if type fractions are desired.  I.e. show 3 input script lines which define the compute and a vector-style variable (dividing by natoms or nbonds) and a line for thermo output (with a wildcard).

akohlmey:
> Q for @akohlmey - I was looking to see how fix shake flagged its scalar output. I don't see that it sets extscalar = 0 or 1. I'm thinking this is fine when the output energy is just added to total energy. But if a user explicitly output the SHAKE energy as a thermo field, would it potentially be a memory access bug in Thermo::compute_fix() ?

Yes, it is. You'll read from uninitialized memory and then extscalar would be 0 the first time, but likely non-zero after a clear.

akohlmey:
> @akohlmey Should have added that I also don't see that extscalar is set to a default value in fix.cpp

Neither is extvector or extarray. Only extlist is initialized to a null pointer. And that goes not only for fix.cpp but also for compute.cpp

sjplimp:
ok - thanks for confirming - so the bug is in fix_shake.cpp (and rattle, etc).  My view is any fix or compute that sets scalar_flag or vector_flag should be setting extscalar/extvector.  We should not initialize those values in fix.cpp or compute.cpp.

akohlmey:
> oops @jrgissing - my comment just crossed with yours - I think it is intentional that no default is set b/c any fix that enables scalar_flag = 1 should set it. But I want to see what @akohlmey says about this

I'd rather have it set to "invalid", e.g. -1, and then have a check if the scalar_flag is set to either 0 or 1 or else stop with an error stating that the developer for the fix/compute needs to be contacted to set "extscalar". Same for the other settings.

In the Compute class there is a natural candidate for this in the `Compute::init_flags()` function which is called after `Compute::init()` in modify.cpp. I suggest we add the same function to the `Fix` class and call it the same way with the same added checks including for extvector, extarray, and extlist.

akohlmey:
> ok - thanks for confirming - so the bug is in fix_shake.cpp (and rattle, etc). My view is any fix or compute that sets scalar_flag or vector_flag should be setting extscalar/extvector. We should not initialize those values in fix.cpp or compute.cpp.

Leaving them uninitialized is about as bad as having a default setting. This can lead to undefined or unexpected behavior, since we have no automatic mechanism to check whether these settings are consistent and there is no uninitialized data access. Tracking down uninitialized data access specifically in thermo output with valgrind is very difficult. So I would suggest to initialize to an "invalid" flag and then check for it.

sjplimp:
@akohlmey That would be fine.  What I really meant was when extscalar/extvector were added to the code (at the very beginning), the intent was that any fix/compute which output a global scalar/vector would also set those flags.  Not sure how that didn't happen for fix shake.  Either my mistake, or possibly it was accidentally removed when it was refactored by others.

akohlmey:
> the intent was that any fix/compute which output a global scalar/vector would also set those flags.

@sjplimp 
Expecting the programmer to know this and follow it, specifically when this requirement is not documented or commented in a way that it is easily found, is not something that is good for a large software package with many contributors. 

Would you prefer the check to be added to this pull request or submitted in a separate pull request?
Adding the test may flag some more omissions during the test runs. We could also write a small python program that tries to analyze the fix/compute source if the settings are consistent, but those would fail for derived classes where the setting would be inherited, so it would need a false positives list.

akohlmey:
> Not sure how that didn't happen for fix shake. Either my mistake, or possibly it was accidentally removed when it was refactored by others.

My guess is that it wasn't added. The computed scalar is the bond energy from the "stiff" bonds during minimization. This is a feature that  I added (and you reviewed).

Here is what it looks like with the check added:
```
Generated 2278 of 2278 mixed pair_coeff terms from arithmetic mixing rule
ERROR: Must set 'extscalar' when setting 'scalar_flag' for fix shake.  Contact the developer. (src/fix.cpp:135)
Last command: run             100
```
It is rather a small amount of code changes:
```
commit a88e8757e3cedfa0c13f5fec85fd0ab318ba2e2e
gpg: Signature made Mon 22 Apr 2024 07:55:52 PM EDT
gpg:                using RSA key EEA103764C6C633EDC8AC428D9B44E93BF0C375A
gpg:                issuer "akohlmey@gmail.com"
gpg: Good signature from "Axel Kohlmeyer <akohlmey@gmail.com>" [ultimate]
Author: Axel Kohlmeyer <akohlmey@gmail.com>
Date:   Mon Apr 22 19:55:52 2024 -0400

    guard against not setting extscalar, extvector, or extarray when required

diff --git a/src/compute.cpp b/src/compute.cpp
index a12373fd51..5d5a4d5246 100644
--- a/src/compute.cpp
+++ b/src/compute.cpp
@@ -59,6 +59,7 @@ Compute::Compute(LAMMPS *lmp, int narg, char **arg) :
   // set child class defaults
 
   scalar_flag = vector_flag = array_flag = 0;
+  extscalar = extvector = extarray = -1;
   peratom_flag = local_flag = pergrid_flag = 0;
   size_vector_variable = size_array_rows_variable = 0;
 
@@ -116,6 +117,16 @@ void Compute::init_flags()
   initialized_flag = 1;
   invoked_scalar = invoked_vector = invoked_array = -1;
   invoked_peratom = invoked_local = -1;
+
+  if (scalar_flag && (extscalar < 0))
+    error->all(FLERR, "Must set 'extscalar' when setting 'scalar_flag' for compute {}.  "
+               "Contact the developer.", style);
+  if (vector_flag && (extvector < 0))
+    error->all(FLERR, "Must set 'extvector' when setting 'vector_flag' for compute {}.  "
+               "Contact the developer.", style);
+  if (array_flag && (extarray < 0))
+    error->all(FLERR, "Must set 'extarray' when setting 'array_flag' for compute {}.  "
+               "Contact the developer.", style);
 }
 
diff --git a/src/fix.cpp b/src/fix.cpp
index 754948fdd1..1e1ab8c928 100644
--- a/src/fix.cpp
+++ b/src/fix.cpp
@@ -81,6 +81,7 @@ Fix::Fix(LAMMPS *lmp, int /*narg*/, char **arg) :
   diam_flag = 0;
 
   scalar_flag = vector_flag = array_flag = 0;
+  extscalar = extvector = extarray = -1;
   peratom_flag = local_flag = pergrid_flag = 0;
   global_freq = local_freq = peratom_freq = pergrid_freq = -1;
   size_vector_variable = size_array_rows_variable = 0;
@@ -119,13 +120,28 @@ Fix::~Fix()
 {
   if (copymode) return;
 
-  delete [] id;
-  delete [] style;
+  delete[] id;
+  delete[] style;
   memory->destroy(eatom);
   memory->destroy(vatom);
   memory->destroy(cvatom);
 }
 
+
+void Fix::init_flags()
+{
+   if (scalar_flag && (extscalar < 0))
+    error->all(FLERR, "Must set 'extscalar' when setting 'scalar_flag' for fix {}.  "
+               "Contact the developer.", style);
+  if (vector_flag && (extvector < 0))
+    error->all(FLERR, "Must set 'extvector' when setting 'vector_flag' for fix {}.  "
+               "Contact the developer.", style);
+  if (array_flag && (extarray < 0))
+    error->all(FLERR, "Must set 'extarray' when setting 'array_flag' for fix {}.  "
+               "Contact the developer.", style);
+}
+
    process params common to all fixes here
    if unknown param, call modify_param specific to the fix
diff --git a/src/fix.h b/src/fix.h
index ca0a1ef84b..594fbb51bf 100644
--- a/src/fix.h
+++ b/src/fix.h
@@ -145,6 +145,7 @@ class Fix : protected Pointers {
 
   virtual void post_constructor() {}
   virtual void init() {}
+  void init_flags();
   virtual void init_list(int, class NeighList *) {}
   virtual void setup(int) {}
   virtual void setup_pre_exchange() {}
diff --git a/src/modify.cpp b/src/modify.cpp
index ba04c5969a..1a9c056027 100644
--- a/src/modify.cpp
+++ b/src/modify.cpp
@@ -188,6 +188,8 @@ void Modify::init()
   //   since any of them may be invoked by initial thermo
   // do not clear out invocation times stored within a compute,
   //   b/c some may be holdovers from previous run, like for ave fixes
+  // perform check whether extscalar, extvector, and extarray have been
+  //   set when scalar_flag, vector_flag, or array_flag are true.
 
   for (i = 0; i < ncompute; i++) {
     compute[i]->init();
@@ -200,8 +202,13 @@ void Modify::init()
   //   used to b/c temperature computes called fix->dof() in their init,
   //   and fix rigid required its own init before its dof() could be called,
   //   but computes now do their DOF in setup()
+  // perform check whether extscalar, extvector, and extarray have been
+  //   set when scalar_flag, vector_flag, or array_flag are true.
 
-  for (i = 0; i < nfix; i++) fix[i]->init();
+  for (i = 0; i < nfix; i++) {
+    fix[i]->init();
+    fix[i]->init_flags();
+  }
 
   // set global flag if any fix has its restart_pbc flag set
 
```
```

**Comment by akohlmey** [comment]
@jrgissing @sjplimp I have pushed my changes with the extra tests (had to refine them to account for extlist) and updated fix shake and fix spring/rg which were missing settings. I'll now turn on run and regression tests to check for more missing settings. Found issues with fix pour, fix deposit, and compute voronoi/atom this way.

If it is undesired to have these changes here, you or I can just revert them.

File: doc/src/compute_count_type.rst (hunk 1 above)
Comment (jrgissing) [reply doc/src/compute_count_type.rst]: By the way, vector variables need an upper limit when using a wildcard. As far as I know, there is no way to get number of atom types in an input script. If that's right, do you think its worth it to add thermo keywords for number of atom, bond, etc. types?

Comment (jrgissing) [reply doc/src/compute_count_type.rst]: wait, 'extract_setting' might do it...

File: doc/src/compute_count_type.rst [hunk 2]
```diff
@@ -69,29 +69,42 @@ for each type:
...
-If the {mode} setting is {bond} then the count of bonds for each bond
+The atom count for each type can be normalized by the total number of
+atoms like so:
+
+.. code-block:: LAMMPS
+
+   compute typevec all count/type atom # number of atoms of each type
+   variable normtypes vector c_typevec/atoms # divide by total number of atoms
+   variable ntypes equal extract_setting(ntypes) # number of atom types
+   thermo_style custom step v_normtypes[*$(v_ntypes)] # vector variable needs upper limit
```
akohlmey:
```suggestion
   thermo_style custom step v_normtypes[*${ntypes}] # vector variable needs upper limit
```
Why not this?

stanmoore1:
Is this now ready for review/merge?

akohlmey (approved):

PR #4144, opened 2024-04-20
Fix for fix_ave_histo bug with vector style variables

Bibobu:
**Summary**

Following this discussion: https://matsci.org/t/problem-with-fix-ave-histo/54902

I think I noted a discrepancy between documentation and actual behavior of fix ave/histo.
The documentation states that the command can take variables input in the form
`v_name = value(s) calculated by an equal-style or vector-style or atom-style variable with name`
However, only equal style and atom style variables are tested for. This PR should add a test for vector style variables.

**Related Issue(s)**

None that I know about.

**Author(s)**

Germain Clavier - Unicaen (germain.clavier@unicaen.fr)

**Licensing**

**Backward Compatibility**

Keeps backward compatibility as far as I can tell.

**Implementation Notes**

The lines from the linked discussion were tested with the current develop version of LAMMPS. The behavior from the 2019 version used is still present.

**Post Submission Checklist**


akohlmey (approved):
Looks reasonable to me

PR #4150, opened 2024-04-26
Avoiding forcezero infinite loop with zero energy.

Bibobu:
**Summary**

Following [this discussion](https://matsci.org/t/minimisation-iteration-gets-stuck-once-the-energy-is-zero/54981/10), I happened to notice that ``line forcezero`` line search algorithm displacement variables was initialized using the absolute value of the current energy. This can cause an infinite loop in the algorithm if the energy is exactly equal to 0 or falls below machine precision and gets LAMMPS stuck.

**Related Issue(s)**

None that I know about.

**Author(s)**

Germain Clavier -- germain.clavier@unicaen.fr

**Licensing**

**Backward Compatibility**

Should be good.

**Implementation Notes**

The `alpha_init` value is initialized either to its original value or to 1e-28 (EPS_QUAD). The implementation was tested and solved the original issue while allowing for a continuous decrease in forces magnitudes. I think a WARNING message could be issued to signal the unphysical situation the simulated system is likely to be in.

**Post Submission Checklist**


sjplimp:
I think @athomps is the best developer to review this change

athomps (approved):
I approve.

athomps:
The initialization used for ``line forcezero`` is clearly not very general. This is consistent with the implicit recommendation not to use ``line forcezero`` unless you know what you are doing. For example, what happens if PE < 0?

PR #4154, opened 2024-04-29
new comm_style brick/direct option

sjplimp:
**Summary**

Implement a new comm_style brick/direct, derived from brick, which performs forward/reverse communication in a different manner.  Instead of using 6-way data exchanges (brick) with only neighboring procs that share faces of a proc's subdomain,
it performs direct communication with all procs that need ghost atoms.  For cutoff < subdomain size this would be 26 procs in 3d or 8 in 2d.  This leads to more MPI messages, but they can all be done simultaneously.  We anticipate this will be slower on CPUs, but may be faster on GPUs, where the pack/unpack operations for message data can be performed with more GPU parallelism and fewer kernel launches.

**Related Issue(s)**

N/A

**Author(s)**

Stan and Steve

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

**Post Submission Checklist**


sjplimp:
@stanmoore1 Here is an initial implementation I've tested on the LJ benchmark.  Still needs some additional work, but it is giving correct answers on small problems and handfuls of procs.  Maybe you can test at larger scales.  Bonded systems do not yet work b/c an additional variant of forward comm is needed.  Looking at forward/reverse/borders you should be able to see what changed and how it could be Kokkos-ized.

sjplimp:
@akohlmey and @stanmoore1 On the 2 LJ test problems I will post in a moment, this version gives a valgrind error in the 2 Comm destructors, which I don't understand.  I think it is something to do with the copy constructor memory.  Ditto for leaked memory.  This is even on a single proc.

sjplimp:
Here are 2 good test scripts for LJ

[in.lj.3d.txt](https://github.com/lammps/lammps/files/15150519/in.lj.3d.txt)
[in.lj.2d.txt](https://github.com/lammps/lammps/files/15150520/in.lj.2d.txt)

stanmoore1:
@sjplimp many thanks! I will work on this.

stanmoore1:
@sjplimp I fixed the Valgrind issues, but when I try to run on more than 1 MPI rank, I'm getting an MPI error for both the 2D and 3D. I checked the commit before the Valgrind fix and it gives the same crash, any ideas?

```
Setting up Verlet run ...
  Unit style    : lj
  Current step  : 0
  Time step     : 0.005
NDIRECT 34 ijk lo/hi -3 3: -2 2: 0 0 proc 0
NDIRECT 34 ijk lo/hi -3 3: -2 2: 0 0 proc 1
Fatal error in MPI_Irecv: Invalid tag, error stack:
MPI_Irecv(153): MPI_Irecv(buf=0x2091204, count=1, MPI_INT, src=0, tag=538445311, MPI_COMM_WORLD, request=0x2099400) failed
MPI_Irecv(99).: Invalid tag, value is 538445311
Fatal error in MPI_Irecv: Invalid tag, error stack:
MPI_Irecv(153): MPI_Irecv(buf=0x22c0384, count=1, MPI_INT, src=1, tag=538445311, MPI_COMM_WORLD, request=0x22c1d80) failed
MPI_Irecv(99).: Invalid tag, value is 538445311
```

sjplimp:
The 2 scripts I sent work fine on my box for up to 12 procs (valgrind clean and correct answers).
How many procs are you running on, and did you edit the scripts?

stanmoore1:
I'm running on 2 procs, and I did not edit the scripts. I'm using `mpich/3.2.1`, will try OpenMPI instead

stanmoore1:
@sjplimp it works fine with OpenMPI (4.0.5) but not with MPICH (3.2.1).

stanmoore1:
@sjplimp I think I found the issue. The maximum upper bound for MPI tags using MPICH is `268,435,455`, but the tags in the code are larger than that. The maximum upper bound for tags using OpenMPI is `2,147,483,647` which is why it works correctly. This can be checked with the `MPI_TAG_UB` variable. So we may need to rethink "MPI tags based on 3d offset between 2 procs from receiver's perspective".

stanmoore1:
@sjplimp can you check my changes? If we keep them, we'll need to port `MPI_Comm_get_attr(world,MPI_TAG_UB,&maxtag_mpi_ptr,&tmp)` to work with MPI STUBS library to get the serial version to compile.

I'm going to port this to Kokkos now so we can get an initial performance eval on GPUs.

sjplimp:
@stanmoore1 I adjusted the MPI tag calculation and check logic a bit to simplify it and keep the tags smaller.
For STUBS/mpi.cpp, MPI_Comm_get_attr() and MPI_TAG_UB there are 2 options.

(1) Could just return the largest possible 32-bit positive int.  It's a bit odd b/c in serial, LAMMPS should never make an MPI_Send or MPI_Recv or MPI_IRecv call.  So the maxtag value is irrelevant b/c memory copies are performed instead of MPI Send/Recv.
But the code that checks for maxtag doesn't know that.

(2) Or return 0 for a maxtag from STUBS, but only do the maxtag check if nprocs > 1.

stanmoore1:
@akohlmey it looks like the CentOS compilation test doesn't support C++11 when compiling `STUBS/mpi.cpp`, it is giving:

```
make[1]: Entering directory `/data/workspace/dev/pull_requests/centos7/legacy_serial_openmp_smallsmall_static/build/lammps/src/STUBS'
g++ -O2 -g -fopenmp -c mpi.cpp
mpi.cpp: In function 'int MPI_Comm_get_attr(int, int, void**, int*)':
mpi.cpp:717:25: error: 'nullptr' was not declared in this scope
     attribute_val_ptr = nullptr;
                         ^
make[1]: [mpi.o] Error 1 (ignored)
```

stanmoore1:
> I adjusted the MPI tag calculation and check logic a bit to simplify it and keep the tags smaller.

@sjplimp thanks, looks good.

akohlmey:
> @akohlmey it looks like the CentOS compilation test doesn't support C++11 when compiling `STUBS/mpi.cpp`, 

STUBS is C code, not C++. We only compile it with g++ to work around namespace clashes when using the LAMMPS library interface from an MPI code when LAMMPS is compiled without MPI.

stanmoore1:
> STUBS is C code, not C++. We only compile it with g++ to work around namespace clashes when using the LAMMPS library interface from an MPI code when LAMMPS is compiled without MPI.

OK I'll use `NULL` instead.

sjplimp:
@stanmoore1 Added support for other flavors of forward/reverse comm -- from pair, fix, compute styles.  Was pretty easy since they are similar to standard forward/reverse and to each other.  This means all the benchmark problems now run (chain, eam, rhodo).

On issue: for eam and chain under valgrind with --leak-check=full, the derived class errors for constructor/destructor leaked memory are still there.  Not for LJ though.

Other (non?) issue:  After 100 steps rhodo gives epsilon-different answers with brick versus brick/direct.  Last digit of a few thermo values.  Looks like this also happens with varying proc count for brick itself.  I think this is likely just due to different orderings of ghost atoms in brick vs brick/direct.  I.e. round-off diffs for force summations.

stanmoore1:
@sjplimp thanks! I'll look at the memory leak issue. The slowly diverging trajectories over time seems expected.

stanmoore1:
We could never get this new comm method to be faster than the original 6 passes method. Therefore I'm closing this PR for now. Can revisit in the future if there is more interest.

PR #4156, opened 2024-05-01
Fix CUDA block count limitation in GPU package

tpcbtw:
**Summary**

When running large-scale simulations with GPU acceleration in LAMMPS, a "Cuda driver error 1" sometimes occurs. This error is caused by the number of CUDA blocks (GX) exceeding the limitation of 65535. The current block size (BX) is set by block_size() for optimal performance, which is significantly smaller than the upper limit of 1024.
To avoid this error, the patch modifies the parts of the GPU library where BX and GX are specified for each force field. If GX exceeds 65535, the block size is dynamically adjusted. If the new block size is within the limit of 1024, it updates the block size and block count accordingly, allowing the program to run within the block count and block size limitations.
The same modification has been applied to 53 files, as similar BX and GX specifications exist for each force field. However, some files use unique methods for specifying GX and BX and have not been modified.

**Related Issue(s)**

None to my knowledge

**Author(s)**

nei, nei@tpcbtw.com <Corresponding author>

**Licensing**

**Backward Compatibility**

We don't think that's a possibility.

**Implementation Notes**

The operation was verified by modifying some input files in benchmark to a large system and by GV100 and CUDA 12.0.
We have also verified that the problem of the number of blocks exceeding 65535 and terminating the computation on certain large systems has been corrected.

While systems that could not be computed can now be computed, performance on those systems will be slightly worse because optimal block placement is no longer possible.
In a normal system, this does not affect calculation speed.

**Post Submission Checklist**

This is my first pull request in my life.
Therefore, I apologize in advance for any inconvenience caused.

### DIFF
--- a/lib/gpu/geryon/nvd_kernel.h
+++ b/lib/gpu/geryon/nvd_kernel.h
@@ -335,6 +335,46 @@ class UCL_Kernel {
 
   /// Run the kernel in the default command queue
   inline void run() {
+    const unsigned int maxBlockSize = 65535;
+    unsigned long totalThreads = static_cast<unsigned long>(_num_blocks[0]) * _num_blocks[1] * _num_blocks[2];
+    if (_num_blocks[0] > maxBlockSize) {
+        _num_blocks[1] = static_cast<unsigned int>((totalThreads / maxBlockSize) / _num_blocks[2]);
+        _num_blocks[0] = maxBlockSize;
+    }
+    if (_num_blocks[1] > maxBlockSize) {
+        _num_blocks[2] = static_cast<unsigned int>((totalThreads / maxBlockSize) / _num_blocks[0]);
+        _num_blocks[1] = maxBlockSize;
+    }
+    if (_num_blocks[2] > maxBlockSize) {
+        _num_blocks[0] = static_cast<unsigned int>((totalThreads / maxBlockSize) / _num_blocks[1]);
+        _num_blocks[2] = maxBlockSize;
+    }
+    unsigned long adjustedTotalThreads = static_cast<unsigned long>(_num_blocks[0]) * _num_blocks[1] * _num_blocks[2];
+    if (adjustedTotalThreads != totalThreads) {
+        unsigned long difference = totalThreads - adjustedTotalThreads;
+        unsigned long increment = difference / (_num_blocks[0] * _num_blocks[1] * _num_blocks[2]);
+
+        while (adjustedTotalThreads < totalThreads) {
+            if (_num_blocks[0] < maxBlockSize) {
+                _num_blocks[0]++;
+                adjustedTotalThreads += (_num_blocks[1] * _num_blocks[2]);
+            } else if (_num_blocks[1] < maxBlockSize) {
+                _num_blocks[1]++;
+                adjustedTotalThreads += (_num_blocks[0] * _num_blocks[2]);
+            } else if (_num_blocks[2] < maxBlockSize) {
+                _num_blocks[2]++;
+                adjustedTotalThreads += (_num_blocks[0] * _num_blocks[1]);
+            } else {
+                break;
+            }
+        }
+    }
+    adjustedTotalThreads = static_cast<unsigned long>(_num_blocks[0]) * _num_blocks[1] * _num_blocks[2];
+    if (adjustedTotalThreads != totalThreads) {
+        fprintf(stderr, "Warning: Adjusted block size changes total thread count from %lu to %lu\n",
+                totalThreads, adjustedTotalThreads);
+    }
+
     CU_SAFE_CALL(cuLaunchKernel(_kernel,_num_blocks[0],_num_blocks[1],
                                 _num_blocks[2],_block_size[0],_block_size[1],
                                 _block_size[2],0,_cq,_kernel_args,nullptr));
--- a/lib/gpu/lal_amoeba.cpp
+++ b/lib/gpu/lal_amoeba.cpp
@@ -257,8 +257,17 @@ int AmoebaT::umutual2b(const int /*eflag*/, const int /*vflag*/) {
 
   // Compute the block size and grid size to keep all cores busy
   int BX=this->block_size();
-  int GX=static_cast<int>(ceil(static_cast<double>(this->ans->inum())/
-                               (BX/this->_threads_per_atom)));
+  int GX=static_cast<int>(ceil(static_cast<double>(this->ans->inum())/(BX/this->_threads_per_atom)));
+  // Increase block size to reduce the block count
+  if (GX > 65535) {
+    int newBX = static_cast<int>(ceil(static_cast<double>(this->ans->inum()) / 65535.0));
+    newBX = ((newBX + this->_threads_per_atom - 1) / this->_threads_per_atom) * this->_threads_per_atom;
+    if (newBX <= 1024) {
+        BX = newBX;
+        GX = static_cast<int>(ceil(static_cast<double>(this->ans->inum()) / (BX / this->_threads_per_atom)));
+    }
+  };
+
   this->time_pair.start();
 
   // Build the short neighbor list if not done yet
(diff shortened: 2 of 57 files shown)


ndtrung81:
@tpcbtw It seems that the block size exceeds the maximum block count on the device (65,535) when the number of particles (atoms) per GPU exceeds (65,535 * 128) = 8,388,480 (for neigh yes), or (65,535 * 256) = 16,776,960 (for neigh no), correct? If you reduce the number of particles per GPU by increasing the number of MPI procs per GPU, does the CUDA driver error 1 issue persist without the changes in this PR?

tpcbtw:
@ndtrung81 Thank you for your comment.
Your understanding is largely correct.
If atom_style is set to 'atomic', and the number of atoms allocated per GPU exceeds 8,388,480, the block size limit will be exceeded.
However, when using atom_style such as 'bond' or 'full', the limits on the number of CUDA blocks might be reached with fewer atoms because not only atoms but also bonds, angles, dihedrals, and other degrees of freedom are allocated to the GPU. When the combined total of these elements exceeds 8,388,480, it similarly surpasses the maximum block count. For example, a polymer system with approximately one million atoms on a single GPU could potentially fail to compute.

> If you reduce the number of particles per GPU by increasing the number of MPI procs per GPU, does the CUDA driver error 1 issue persist without the changes in this PR?

Your point is well-taken.
This approach could potentially avoid the CUDA driver error 1. (In fact, in the first system where I encountered this error, I was able to resolve it by increasing the number of processes.)
However, it might not be feasible for all scenarios due to the diverse configurations and requirements of different simulations.

ndtrung81:
@tpcbtw for the GPU package none of the data structures for bonds, angles, and dihedrals are allocated to the GPU. Not sure where in the code leads you to this statement. The big size of the neighbor list could be the cause of failure for your polymer system.

One of the advantages of using the GPU package is that multiple MPI procs can share the GPU(s) on a node. If you keep increasing the number of atoms per proc, sooner or later either the execution configuration of the kernels will become invalid (as you reported), or memory allocation will fail for certain GPUs. Rather than adjusting the number of blocks in all the pair styles, I suggest you add an assert for GX < max grid size of the GPU in use, specifically in Device::init() in lal_device.cpp.

tpcbtw:
@ndtrung81 First of all, I apologize for my misunderstanding and the delay in replying for confirmation.
In conclusion, as you pointed out, due to some of my errors and misunderstandings, this is not the correct PR.

1. Thank you for correcting my mistake regarding memory allocation and for identifying the big size of the neighbor list as the cause. However, considering this, even a model with only a few hundred thousand atoms allocated to one process will fail to run. Given the recent increase in VRAM, a fundamental solution is needed for the problem related to block_size per process.
2. I believed this PR was the fundamental solution to the problem, but I was mistaken. To verify this PR, I had modified parts related to cuLaunchKernel in nvd_kernel.h (not included in the PR), which caused the NVD_GERYON_EXIT to be skipped in CU_SAFE_CALL.
3. As you pointed out, it is inappropriate to adjust BX and GX in each pair style. In fact, even though I attempted to adjust GX within the pair_style in this PR, I found that cuLaunchKernel still attempted to get more than 65535 blocks and did not function at all.I agree with your suggestion to modify lal_device.cpp or nvd_kernel.h instead.
4. I haven't fully understood the structure of lal_device.cpp yet, so I couldn't implement your suggested fixes. Instead, I added a simple code snippet in nvd_kernel.h, which seems to have resolved the issue for now. I would like your opinion on whether I should proceed with improving this snippet or adjust my approach to fix lal_device.cpp as you suggested.
5. As I'm not familiar with GitHub or LAMMPS culture, I'm unsure whether to close this PR or implement the fixes mentioned in point 4 and ask for your review. I would like your opinion.


Details for points 1 to 4 are as follows:

 1.
The attached in.lj has only 530,604 atoms, but _num_blocks={1,66326,1}, which exceeds the _num_blocks limit and makes the calculation impossible.
As you first pointed out, increasing the number of processes would be a good solution, but this system is only 500k atoms and about 3GB of VRAM.
Considering the recent increase in memory per GPU, it is not a good choice to divide this into sequential processes.
This is because the large systems we are interested in, with more than 10 million atoms, often allocate more than 500k atoms per process, no matter how much the process is divided.
[in.lj.txt](https://github.com/lammps/lammps/files/15447094/in.lj.txt)

  2. and 3.
To verify this PR, I had modified parts related to cuLaunchKernel in nvd_kernel.h, but these changes were not reflected in the PR.
```
*** nvd_kernel.h        2024-05-26 15:38:49.451988489 +0900
--- nvd_kernel.h.back   2024-05-25 21:30:21.549249366 +0900
*************** class UCL_Kernel {
*** 335,343 ****

    /// Run the kernel in the default command queue
    inline void run() {
!     CUresult result = cuLaunchKernel(_kernel,_num_blocks[0],_num_blocks[1],_num_blocks[2],
!                                      _block_size[0],_block_size[1],_block_size[2],
!                                      0,_cq,_kernel_args,nullptr);
    }

    /// Clear any arguments associated with the kernel
--- 335,343 ----

    /// Run the kernel in the default command queue
    inline void run() {
!     CU_SAFE_CALL(cuLaunchKernel(_kernel,_num_blocks[0],_num_blocks[1],
!                                 _num_blocks[2],_block_size[0],_block_size[1],
!                                 _block_size[2],0,_cq,_kernel_args,nullptr));
    }

    /// Clear any arguments associated with the kernel
```
I had identified that the error messages I was encountering were due to CU_SAFE_CALL and _num_blocks, and to debug this, I was executing cuLaunchKernel without passing through CU_SAFE_CALL. As a result, NVD_GERYON_EXIT was skipped, and the calculation continued even though cuLaunchKernel failed.

While I had not modified nvd_kernel.h in this PR, CU_SAFE_CALL in its current state returns a Cuda driver error 1. This indicates that despite attempts to adjust GX within the pair style, cuLaunchKernel was trying to get more than 65535 blocks, showing that my PR was not functioning at all.

4.
You suggested to include an assertion in Device::init() in lal_device.cpp to make GX<65535. However,I have not yet deciphered the structure of lal_device.cpp, so I could not make your intended modification. However, instead I have added a simple code snippet in nvd_kernel.h in run() just before CU_SAFE_CALL(cuLaunchKernel) that fixes _num_blocks not to exceed 65535. With this modification, we have confirmed that it is possible to analyze systems like the input file shown in point 1 and larger systems in a single process.
This snippet has not been fully validated, so it is a work in progress, but  added to the PR. I would like to hear your opinion whether I should proceed with improving this snippet or if I should change course and fix lal_device.cpp as you intend.
And in any case, I think the pair-style PR should be reverted.

ndtrung81:
@tpcbtw thanks for looking closely at the code. The input file in.lj.txt you shared using `atom_style full` which then triggers the launch of the kernel `k_transpose` inside the neighbor build (`Neighbor::build_nbor_list()`, lal_neighbor.cpp). The block sizes for this kernel are rather small (8), which then give rise to the high block count as you saw for 530,604 atoms. One way to workaround the current issue with 500K atoms per proc is to adjust `_block_cell_2d`, as defined by `BLOCK_CELL_2D`, until `g2y` is below the max block count of the current device. So the fail is due not to the pair compute kernel but to this particular transpose kernel. You can switch to  `neigh no` to see that the run doesn't involve this kernel.

The UCL_Device class (see nvd_device.h for the CUDA backend) provides several member functions to get the max block count of the current device. You can see how in the LAL_Device class `compile_kernels()` checks for the valid values for its current device `gpu`.

If you increase the atoms per proc, you may decrease the number of threads per atom (`tpa`), from 8 (default) to 4, to 2 and then 1, to keep the block count below the max value.

LAMMPS is designed for running multiple procs in parallel, they are not sequential in any way. For higher atom counts like this case, multiple procs sharing the GPU(s) is actually recommended.  If you wish to increase the number of atoms per proc, sooner or later the device you are using is not capable of doing your calculation. It's like I was trying to run 100,000 atoms on a very old card.

I would suggest you close this PR, and create a new one if you would like to 1) implement the adjustment to `b2y` and `b2x` used in the `k_transpose` kernel launch, and maybe 2) add a check in `LAL_Device::init()` for the max block count and error out with some error code (say -17) to be handled by `check_flag()` in `gpu_extra.h`.

akohlmey:
@tpcbtw are you going to follow the advice of @ndtrung81 or do you have some argument why this variant is superior?
Please let us know because in the current state this pull request will not be merged into LAMMPS.

tpcbtw:
@akohlmey 
I sincerely apologize for my delayed reaction. Recently, I have not had the time to take action on this issue.

In conclusion, it is clear that this PR needs to be closed. During our discussions, I have demonstrated that this PR no longer functions, and @ndtrung81 has pointed out that the core issue lies with the transpose kernel, rather than what I had identified. He also recommended creating a new PR.

I still believe that this issue requires a fundamental solution. However, I have not been able to take concrete action on the implementation suggested by @ndtrung81. Whether I work on it or someone else does, I believe that solving this will take time. Therefore, I intend to close this PR and create a new Issue instead of a new PR.

Finally, I would like to express my heartfelt gratitude to @ndtrung81 for his courteous response and advice.

PR #4163, opened 2024-05-16
Add an interface to machine learning models through metatensor (`pair_style metatensor`)

Luthaf:
**Summary**

This adds the ability to use [metatensor atomistic models](https://docs.metatensor.org/latest/atomistic/index.html) as machine learning (ML) potentials in LAMMPS. Metatensor is not related to any specific ML architecture, but rather define an interface that arbitrary models can follow. This interface then allows to use the exact same model with a variety of simulation engines: currently ASE and i-PI are supported, this PR adds the ability to use such models with LAMMPS.

**Related Issue(s)**

None

**Author(s)**

- Guillaume Fraux (guillaume.fraux@epfl.ch), École Polytechnique Fédérale de Lausanne (EPFL), Switzerland

**Licensing**

**Backward Compatibility**

This will not break backward compatibility

**Implementation Notes**

Reviewers can check the steps to use a metatensor model here: https://docs.metatensor.org/latest/atomistic/overview.html#data-flow-between-the-model-and-engine. The gist of it is as follow

- load the model
- get information about what the model can do and what it requires as input
- compute neighbor lists requested by the model
- execute the model and get the output in torch tensor
- use torch autograd framework to compute forces/virial as the gradient of the energy

The neighbor list format of metatensor differs a bit from the one of LAMMPS, and that's where most of the complexity of this PR is. For example, message passing models might want a NL with a cutoff of 3A, but need to know about neighbors up to 15A away. For proper integration with domain decomposition, we ask LAMMPS to compute a NL with 15A cutoff, and filter it ourself to reconstruct smaller NL down the line. This has been tested to produce the same result regardless of the number of domains with the help of @abmazitov.

**Post Submission Checklist**

This work will be extended in follow-up PR to add support for kokkos, and the ability to use metatensor models that compute something else than the energy (using a custom `compute` command).

This PR also contains a small fix for the general CMake build system in 67fc41b793b182168d408d031333f9ca5b13adc4. I can submit this fix in a seperate PR is you prefer!


src/ML-METATENSOR/pair_metatensor.h
```diff
@@ -0,0 +1,73 @@
...
+#else
+
+#ifndef LMP_PAIR_METATENSOR_H
+#define LMP_PAIR_METATENSOR_H
+
+#include <memory>
+
+#include "pair.h"
+
+#include <metatensor/torch/atomistic.hpp>
```
akohlmey: This include should not be in this header as it can create conflicts due to having to include all pair_<style>.h header files at the same time. The details of the implementation should be hidden from general LAMMPS. See comment below.

src/ML-METATENSOR/pair_metatensor.h
```diff
@@ -0,0 +1,73 @@
...
+    void coeff(int, char **) override;
+    void init_style() override;
+    double init_one(int, int) override;
+    void init_list(int id, NeighList *ptr) override;
+
+    void allocate();
+
+private:
+    void load_torch_model(const char* path, const char* extensions_directory);
+
```
akohlmey, line 41: The details of the implementation, if they need classes and data types outside of the ones used by LAMMPS itself, should be hidden and the header "PIMPL"-ified, i.e. the header should include just a pointer to a struct and then the struct itself is defined in the corresponding .cpp file. Please see pair_pace.h and pair_pace.cpp in the ML-PACE package for an example where it uses `struct ACEImpl *aceimpl;` in the header.

src/ML-METATENSOR/pair_metatensor.h
```diff
@@ -0,0 +1,73 @@
...
+    torch::Device device;
+    // model capabilities, declared by the model
+    metatensor_torch::ModelCapabilities capabilities;
+    // run-time evaluation options, decided by this class
+    metatensor_torch::ModelEvaluationOptions evaluation_options;
+    // should metatensor check the data LAMMPS send to the model
+    // and the data the model returns?
+    bool check_consistency;
+    // how far away the model needs to know about neighbors
+    double interaction_range = -1;
```
akohlmey: No initializers in the header, please. For consistency with the rest of LAMMPS, put them into the constructor.

src/ML-METATENSOR/pair_metatensor.h
```diff
@@ -0,0 +1,73 @@
+/* -*- c++ -*- ----------------------------------------------------------
+   LAMMPS - Large-scale Atomic/Molecular Massively Parallel Simulator
+   https://www.lammps.org/, Sandia National Laboratories
+   Steve Plimpton, sjplimp@sandia.gov
```
akohlmey:
```suggestion
   LAMMPS Development team: developers@lammps.org
```
Same for the other files.

Luthaf: Sorry, that was a copy/paste leftover. I'll change it!

akohlmey:
The ML-METATENSOR package should be added to the following cmake preset files:
- all_on.cmake
- all_off.cmake
- nolib.cmake

Luthaf:
@athomps and @sjplimp, when do you think you'll have some time to look into this?

akohlmey:
@Luthaf This weekend I have spent considerable time trying to compile LAMMPS with the ML-METATENSOR package and have failed to do so. Here are some observations
- You have made things needlessly complex by using Rust instead of C++ and thus adding new dependencies to LAMMPS for compiling your package. Generally, we have been systematically trying to *reduce* such external dependencies (e.g. by replacing Fortran code with equivalent C++ code, be it through alternate implementations (ex. ReaxFF), be it trough re-implementation in C++ following the Fortran version (ex. MEAM), be it through custom translations via f2c (ex. linalg lib which is a subset of BLAS and LAPACK) and thus increasing the portability and ease of installation). LAMMPS is supposed to be portable and easy to install. Using a programming language different from C++ or writing in a complex C++ style depending on features from the latest C++ standards goes contrary to that.
- I have tried to install the torch library from PyPi with pip (in a virtual environment), but it is not usable since CMake configuration fails due to lack of CUDA (I don't have an Nvidia GPU).
- I have tried to install the torch library from the pytorch.org website as suggested via pip but that failed due to an apparent dependency on MKL
- I have tried to install the torch library from the pytorch.org website as a C++ library zip file but the same thing happened
- I noticed the instructions for patching the libtorch installation are incorrect and I adjusted the patch accordingly.

Here is the output from CMake:
```
CMake Error in /home/akohlmey/compile/lammps/build-test/_deps/metatensor-torch-src/CMakeLists.txt:
  Imported target "torch" includes non-existent path

    "/home/akohlmey/compile/lammps/build-test/_deps/metatensor-torch-src/MKL_INCLUDE_DIR-NOTFOUND"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not
  provide.

```

Since it is not possible to use your package with a minimal CPU-only setup and because it needs manual patching and a custom compiler, I personally think that is pushing the limits as to what can be accepted into the LAMMPS distribution a bit too far. Please keep in mind that - regardless of how much we point out that this is contributed code - it is going to be the LAMMPS developers that will be blamed for distributing a software that is difficult to install with all features. I can already imagine the nightmares of having to answer questions about installing and compiling your package. Most LAMMPS users are not very familiar with compilation, let alone such complex ones with many dependencies and manual steps. Very few will read the instructions in the Build_extra.rst file and be confused that things just don't work automatically.

Thus my recommendation is to convert your pair style and thus your package into a plugin. That can be distributed and compiled separately (so it would not be part of the LAMMPS distribution, but compatible with it) and then loaded at runtime using the plugin command. Please see https://docs.lammps.org/Developer_plugins.html
This would have the benefit that you have complete control over the distribution and the benefit for the LAMMPS developers that we do not have to support a piece of software that is complex to install.
There is precedent for that, since also the DeepMD folks are using that mechanism, since they are facing similar problems with the complexity of compiling their package due to the dependency on pytorch.

Another possible option could be to add an interface to the MDI library so that it can be run as a separate software using the MDI protocol to exchange information with LAMMPS and be used as a force engine for LAMMPS. I have limited experience with that, but perhaps @sjplimp can give you some pointers https://docs.lammps.org/Howto_mdi.html

What makes matters worse is that LAMMPS already *has* a kind of "meta-ML" package with the ML-IAP package and that one also has support for pytorch based models. Although the latter also requires an external tool (cython), but that one is far easier to manage and mostly optional and the complexities from the pytorch dependencies are all runtime issues, not compile time issues and there are no further requirements to compile than having a working cython installation.

Having said all that, my personal opinion is not the only opinion that matters, and we generally try to handle decisions of what *not* to include, which dependencies are *not* acceptable, or what to *remove* from the distribution in a democratic fashion. Thus I suggest that other developers compiling installing packages like @sjplimp, @athomps, @stanmoore1, @ndtrung81, or @rbberger also make an attempt to compile/install LAMMPS with your package and offer their assessment of the situation. @rbberger can also offer the insight from the perspective of a person that is packaging LAMMPS. If there is no simple consensus we would then have a discussion about this PR at one of the monthly LAMMPS developer video meetings and - if possible - would invite you to join so questions can be asked and concerns addressed.

akohlmey:
P.S.: Since you have an "institutional" GitHub account, I cannot simply push updates into your branch. So please pull the branch https://github.com/akohlmey/lammps/tree/ml-metatensor which contains updates to the latest upstream development code, fixes the merge conflict, and addresses some minor issue.

rbberger:
I'll try to look at this in the coming days.

Luthaf:
> This weekend I have spent considerable time trying to compile LAMMPS with the ML-METATENSOR package and have failed to do so.

First of all, I need to apologize for this. I should have foreseen these difficulties since most of them are known issues with using Torch from CMake. We have some workaround documented here: https://docs.metatensor.org/latest/installation.html#install-torch-cxx, but I should either link to them in the docs I've added; or maybe even duplicate the instructions in LAMMPS docs.

> I have tried to install the torch library from PyPi with pip (in a virtual environment), but it is not usable since CMake configuration fails due to lack of CUDA (I don't have an Nvidia GPU).

So we don't actually need a CUDA compiler for this code, but Torch's CMake files try to find one if Torch itself was configured with CUDA. The workaround is to install the CPU-only version of pytorch with `pip install torch --extra-index-url https://download.pytorch.org/whl/cpu` / using the CPU-only tar archive on PyTorch's website.

> I have tried to install the torch library from the pytorch.org website as suggested via pip but that failed due to an apparent dependency on MKL

This is also a known issue with Torch's CMake, the usual workaround for this is to set `-DMKL_INCLUDE_DIRECTORY=/usr/include`. I am now wondering if I can do this automatically without introducing issues in the case where MKL is present, I'll check! EDIT: I found a workaround, see the latest commit. I'll see if I can incorporate this in metatensor-torch itself.

> Since it is not possible to use your package with a minimal CPU-only setup and because it needs manual patching and a custom compiler, I personally think that is pushing the limits as to what can be accepted into the LAMMPS distribution a bit too far.

It should 100% be possible to use the code in a CPU-only setup, we are already using the code this way internally. I agree that manual patching should be removed as much as possible. If "manual patching" refers to our torch patch, this is bringing the changes from https://github.com/pytorch/pytorch/pull/119945 to older torch PR. Now that there is a released PyTorch version with this patch, we could also require users to use at least this version and get rid of the patch.

-------

> You have made things needlessly complex by using Rust

We are using Rust in the core implementation of our data structure, and I am afraid that switching this to C++ will be a huge amount of work (the code uses some very specific, highly optimized algorithms for hash map, small stack-allocated vectors and a couple of others optimization tools that we can pull in as dependencies in Rust but we would have to re-write to use in C++).

However, I understand that needing another compiler makes building the code quite a bit harder for end users. Our strategy to mitigate this is (a) to provide code that calls the compiler for them in CMake, so they only need to install such compiler if they want to build from source (`apt install cargo` or using https://rustup.rs/), but don't need to call the compiler themself; and (b) we are providing pre-built version of the code in many places. 

At the time of writing the code in this PR, the code only existed as a pre-built in PyPI, but since then we've also added a conda distribution: https://github.com/conda-forge/libmetatensor-feedstock and https://github.com/conda-forge/libmetatensor-torch-feedstock. The Make build system in this PR can actually use the PyPI version and removes the need to build metatensor/metatensor-torch from sources. I could add something similar in the CMake build system, that would mainly add more steps to the build instructions (download, install, configure CMake with the right flags). 

If you prefer a solution like this, I could also explore packaging the code in [conan](https://conan.io/) or [vcpkg](https://vcpkg.io/), which should enable downloading pre-built versions from CMake, and remove the need for manual configuration from the users (i.e. `-DPKG_ML-METATENSOR=ON` is all they need). I might also be able to write something similar to this by hand:

```cmake
# would have to write this function
detect_cpu_arch(CPU_ARCH)

if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux" AND ${CPU_ARCH} STREQUAL "x86_64")
   # download the pre-built & use it in `target_link_libraries`
elseif(...)
    # ...
endif()
```

------

Finally, while I would highly prefer adding the code to upstream LAMMPS, writing the code as a plugin would also work as an alternative. 

> Please keep in mind that - regardless of how much we point out that this is contributed code - it is going to be the LAMMPS developers that will be blamed for distributing a software that is difficult to install with all features. I can already imagine the nightmares of having to answer questions about installing and compiling your package.

I am very aware of this and the work that goes into answering all these questions. I am also very happy to help answering such questions, and if I miss one somewhere (here, on the discourse forum, …) core developers should feel free to ping/CC me and I'll do my best to help.

> If there is no simple consensus we would then have a discussion about this PR at one of the monthly LAMMPS developer video meetings and - if possible - would invite you to join so questions can be asked and concerns addressed.

I can join such a meeting to answer all questions and try to find the best solution for everyone!

> P.S.: Since you have an "institutional" GitHub account, I cannot simply push updates into your branch. So please pull the branch https://github.com/akohlmey/lammps/tree/ml-metatensor which contains updates to the latest upstream development code, fixes the merge conflict, and addresses some minor issue.

I'll update the branch, and I'll make sure to submit future PR from my personal GitHub account.

akohlmey:
> > This weekend I have spent considerable time trying to compile LAMMPS with the ML-METATENSOR package and have failed to do so.
> 
> First of all, I need to apologize for this. I should have foreseen these difficulties since most of them are known issues with using Torch from CMake. We have some workaround documented here: https://docs.metatensor.org/latest/installation.html#install-torch-cxx, but I should either link to them in the docs I've added; or maybe even duplicate the instructions in LAMMPS docs.

Yes. For anything more complex the LAMMPS manual has installation instructions. Even for installing MPICH from source as an example.

> It should 100% be possible to use the code in a CPU-only setup, we are already using the code this way internally. I agree that manual patching should be removed as much as possible.

There is a workaround for some of these cases in the "ExternalCMakeProject" function provided with LAMMPS. This allows to replace the CMakeLists.txt file of an external package with a customized copy (usually to disable or augment default settings). One of the use cases for this is to change the compilation settings to always and only produce static libraries (but with -fpic enabled), even if a shared libraries build is requested, so that all the libraries are included in the resulting liblammps.so and thus this is the only library that needs to be installed (e.g. for the python module) and no dependencies on shared libraries built as part of the LAMMPS build procedure are produced.

>  If "manual patching" refers to our torch patch, this is bringing the changes from [pytorch/pytorch#119945](https://github.com/pytorch/pytorch/pull/119945) to older torch PR. Now that there is a released PyTorch version with this patch, we could also require users to use at least this version and get rid of the patch.

That is certainly a better option.

> > You have made things needlessly complex by using Rust
> 
> We are using Rust in the core implementation of our data structure, and I am afraid that switching this to C++ will be a huge amount of work (the code uses some very specific, highly optimized algorithms for hash map, small stack-allocated vectors and a couple of others optimization tools that we can pull in as dependencies in Rust but we would have to re-write to use in C++).

I am not saying that you should change since I am aware of the amount of work, but when writing software that you want to be used by a wide audience, you need to consider the complexity of the build process from the very beginning. I am pointing this out simply because this rather technical detail (which on top of that is irrelevant for your personal projects) has impact on usability and desire of adaption of your work. This requires some code duplication (e.g. we have our own mini-regex library in LAMMPS or have our own numerical constants definition, to make certain they are available and consistent across all platforms). This gets even more complex when these choices concern performance critical code paths. Your project is not the only one that creates headaches, CUDA and ROCm are another example. We used to have many questions on the LAMMPS mailing list (before we switched to a forum) about compiling LAMMPS. We have worked hard over many years to make this process simpler and less error prone, especially considering that the average knowledge of a LAMMPS user about compiling from source these days is a fraction of what LAMMPS users knew when it was in its infancy. This is one of the reasons why so much effort went into supporting building LAMMPS with CMake, and if it wasn't for a large number of conservative LAMMPS users (and developers) and the fact that there are some corner cases where the GNU build system is superior over the CMake one, we would have switched to CMake-only. Coming back to GPU support, most of the problems people have these days are with CUDA and ROCm. This is a pity since OpenCL is from the software maintainer standpoint a much more viable option. I can set up binaries for *all* platforms using all C and C++ code and still have access to GPU acceleration, and that in a platform neutral fashion as well. Sadly, the powers that be have decided against giving OpenCL the support that it would need to supersede all platform specific approaches (for understanable reasons).

[...]

> At the time of writing the code in this PR, the code only existed as a pre-built in PyPI, but since then we've also added a conda distribution: https://github.com/conda-forge/libmetatensor-feedstock and https://github.com/conda-forge/libmetatensor-torch-feedstock.

I am not touching anything conda related with a 10 foot pole. Since conda has the ability (and ambition??) to supersede the system provided package management for pre-compiled software, it sooner or later leads to a mess when people use both. I've also found that through conda software is sometimes installed with incompatible binaries. The PyPI system may be less convinient but it is more thourough and their "manylinux" runtime environments are a superior system over dependency processing and overriding of software in conda (also conda seems to be less thorough in checking for correctness and portability of contributions, so there are far more low-quality packages available).

> The Make build system in this PR can actually use the PyPI version and removes the need to build metatensor/metatensor-torch from sources. I could add something similar in the CMake build system, that would mainly add more steps to the build instructions (download, install, configure CMake with the right flags).
 
> If you prefer a solution like this, I could also explore packaging the code in [conan](https://conan.io/) or [vcpkg](https://vcpkg.io/), which should enable downloading pre-built versions from CMake, and remove the need for manual configuration from the users (i.e. `-DPKG_ML-METATENSOR=ON` is all they need). I might also be able to write something similar to this by hand:
> 
> ```cmake
> # would have to write this function
> detect_cpu_arch(CPU_ARCH)
> 
> if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux" AND ${CPU_ARCH} STREQUAL "x86_64")
>    # download the pre-built & use it in `target_link_libraries`
> elseif(...)
>     # ...
> endif()
> ```

That would be an insular solution and also add dependencies on new software and has the added complexity of being OS and hardware specific. This creates a significant maintenance overhead and dependencies on external sources and the need to validate their binaries. LAMMPS currently is tested on x86, ARM, Power CPU platforms and Linux, Windows, macOS, FreeBSD, Solaris operating systems in various versions, often with support for multiple compilers like GNU, CLang, Intel, PGI/Nvidia, and MSVC++.

It *is* possible to include creating a virtual environment and installing software via pip into it during the build process with CMake, but it requires some significant trickery, as can be seen from the "install-python" target in cmake/CMakeLists.txt and the python script files it calls.

Also, the (conventional) makefile in the doc folder uses a custom virtual environment to manage packages installed via pip (including a custom package provided in the doc tree and some packages installed from custom git reposities/branches due to lack of cooperation from upstream of the corresponding packages (or abandonment?)).

> Finally, while I would highly prefer adding the code to upstream LAMMPS, writing the code as a plugin would also work as an alternative.

This would be my personal preference, but let's wait and see what other LAMMPS developers think.

> > Please keep in mind that - regardless of how much we point out that this is contributed code - it is going to be the LAMMPS developers that will be blamed for distributing a software that is difficult to install with all features. I can already imagine the nightmares of having to answer questions about installing and compiling your package.
> 
> I am very aware of this and the work that goes into answering all these questions. I am also very happy to help answering such questions, and if I miss one somewhere (here, on the discourse forum, …) core developers should feel free to ping/CC me and I'll do my best to help.

What is your timeline on this? LAMMPS has significant staying power, I have contributed software in LAMMPS that I have been supporting for more than 15 years now.  Given that machine learning is still in the "gold rush" phase, it is hard to see how long things will be developing fast and APIs and software have changes or problems.  Unfortunately, the situation of most contributions to LAMMPS is that they come when people are near the end of their grad student or postdoc term and then they often move on and have to (mostly) abandon their code or the response times are unbearably long. That is regardless of how competent and responsive a person was at the time of the submission. That is just a fact of (scientific) life. But as a maintainer I do have to worry about this, because - by default - this falls into my lap eventually. For a small, self-contained, C++-only contribution there is usually little worry, but for something more complex the worry grows with the complexity.

Luthaf:
Adding a TODO list for myself regarding points above (and below)

- [ ] Require torch > 2.3 and remove the mkl patch
- [ ] Add remaining build workaround (mainly the CPU/CUDA version of torch) to the documentation
- [x] Check for C++17 standard everywhere

--- 

> There is a workaround for some of these cases in the "ExternalCMakeProject"

This unfortunately does not help in my case, since the issue is not with a CMakeLists.txt, but with a `TorchConfig.cmake` file (and in particular config files for the Caffe2 library which are used by Torch). I am not trying to build torch from source, but rather to consume an existing pre-built library. Changing these files is harder, since I need to know about build options that where used to compiler the code: for example on Linux, is the code using the pre-cxx11 ABI, or the post-cxx11 ABI? Since torch is relatively modular, which ones of the 2-4 possible libraries with different backends (CUDA, HIP, XLA, XPU, …) should be linked?

> This creates a significant maintenance overhead and dependencies on external sources and the need to validate their binaries. LAMMPS currently is tested on x86, ARM, Power CPU platforms and Linux, Windows, macOS, FreeBSD, Solaris operating systems in various versions, often with support for multiple compilers like GNU, CLang, Intel, PGI/Nvidia, and MSVC++.

Would your own preference be to build as much as possible from source, to enable the use of the code on as many systems as possible?

Any solution that relies on pre-built libraries would be limited to whatever PyTorch provides themself anyway, which as far as I know are (these also match the configurations we test on CI ourself):
- x86_64, Linux, GNU compilers (should also work with clang/Intel/PGI compilers)
- x86_64, Windows, MSVC
- aarch64, Linux, GNU compilers (should also work with clang/Intel/PGI compilers)
- aarch64, macOS, Clang (should also work with GNU compilers)
- (x86_64 macOS was dropped recently, the latest version with a pre-built Torch is 2.2.2.)

There is no pre-built torch for ppcle64 CPU, or for FreeBSD/Solaris OS. These users would have to compile all of Torch from source first (which does not seems to be well supported, cf https://github.com/pytorch/cpuinfo/issues/171), then compile metatensor and finally LAMMPS. While this is very unfortunate, users of such system should have access to an IT team to help them compile code (more so than individual users trying to install on their laptops).

Additionally, any strategy to use pre-built code would have a fallback to build from source anyway, if it is required for whatever reason.

> What is your timeline on this? [...] 

Nothing is set in stone yet, but I have a good path toward getting a permanent research software engineering position at EPFL. If this happen, I'll be able to dedicate work time to the maintenance of this code. Hopefully, this should be settled before the end the year.

akohlmey:
> Changing these files is harder, since I need to know about build options that where used to compiler the code: for example on Linux, is the code using the pre-cxx11 ABI, or the post-cxx11 ABI? Since torch is relatively modular, which ones of the 2-4 possible libraries with different backends (CUDA, HIP, XLA, XPU, …) should be linked?

C++11 is the minimum C++ standard required by LAMMPS (we switched from C++98 in early 2020) and turned on by default. Packages may raise the required standard, but currently only KOKKOS does that (C++17).

> Would your own preference be to build as much as possible from source, to enable the use of the code on as many systems as possible?

I am between a rock and a hard place here. Generally, I prefer building as much as possible from source and consistently so, but when building from source is notoriously difficult (as is the case with PyTorch) then there simply is no good solution. The best solution for me is to offload the difficulties to the run time, that would be either to access the torch library through python as it is done with mliappy in `python/lammps/mliap` and the code in `src/ML-IAP/mliap_model_python_couple.pyx` or as a plugin. Then the onus of building and configuring is mostly outside of the LAMMPS build. That is also why I prefer OpenCL over CUDA/ROCm, but - as mentioned before - this is an option with a limited future and current support from vendors.

> While this is very unfortunate, users of such system should have access to an IT team to help them compile code (more so than individual users trying to install on their laptops).

You seem to be working in a very privileged environment. Even at well funded institutions there is usually very little technical IT support beyond Windows and macOS and even if the IT departments hire people, they are often not very competent (old style computer gurus are in short supply) in compiling software from source and my group for example has no funding or staffing to provide that kind of training and support. Since I am operating a large HPC resource at our institution, I get to see quite a bit how people are struggling or following bad advice due to lack of residual knowledge in the research groups. It is quite embarrassing for a research group that specializes in computational research and method development to have nobody in-house that is sufficiently competent in compiling the main application software packages they use outside of their desktops. This is why systems like conda are so popular since they promise to solve that problem, but in the hand of untrained users, the damage they do is often larger than the benefit.

If you look internationally (and I have been visiting researchers in quite a few different regions on all continents outside the antarctic and australia) my experience is that support for compiling scientific software is often between poor to non-existing (not to mention any availability of properly maintained HPC equipment). There is a *very* large LAMMPS user community outside of the US and Europe. Check out the discussions on MatSci.org.

Luthaf:
> C++11 is the minimum C++ standard required by LAMMPS (we switched from C++98 in early 2020) and turned on by default. Packages may raise the required standard, but currently only KOKKOS does that (C++17).

C++11 required a new breaking change in the ABI, which is called the cxx11-abi. On Linux with GNU compilers and libstdc++, they allowed to keep using the old ABI with the new standards for backward compatibility sakes. 

The code in the PR uses C++17 standard (since that's the one Torch uses), but can use either ABI. This is controlled by  the `_GLIBCXX_USE_CXX11_ABI` flag, and we must match whatever was used to compile libtorch. See also https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html for more information on this mess.

> my experience is that support for compiling scientific software is often between poor to non-existing 

I was only talking about PPCLE/FreeBSD/Solaris OS here, not about more mainstream CPU/OS combinations. My understanding is that these CPU/OS combinations are only used by a handful of HPC centers where the would be such staff available (but I might be wrong here!); while most users inside and outside of Europe/USA are using more mainstream hardware, either on personal laptops/workstations or x66_64/Linux HPC clusters.

This means that setting up a build system which uses pre-compiled libraries **only** for such mainstream hardware would already address the needs of **most** users, and might be worth it even if this is not possible to do in all cases. Basically, I was advocating for a solution using pre-compiled libraries as much as possible, even if this means leaving users of less common CPU/OS a bit behind and force them to compile from sources.

akohlmey:
> The code in the PR uses C++17 standard (since that's the one Torch uses),

If your code requires C++17, it should check for that and require that *all* of LAMMPS is compiled with C++17 to avoid subtle inconsistencies and hard to debug issues resulting from multiple compiler settings. We've learned that with KOKKOS. For example with something like this:
```
  diff --git a/cmake/CMakeLists.txt b/cmake/CMakeLists.txt
  index ada38138f6..fc85b89ca6 100644
  --- a/cmake/CMakeLists.txt
  +++ b/cmake/CMakeLists.txt
  @@ -145,7 +145,7 @@ endif()
   if(CMAKE_CXX_STANDARD LESS 11)
     message(FATAL_ERROR "C++ standard must be set to at least 11")
   endif()
  -if(PKG_KOKKOS AND (CMAKE_CXX_STANDARD LESS 17))
  +if((PKG_KOKKOS OR PKG_ML-METATENSOR) AND (CMAKE_CXX_STANDARD LESS 17))
     set(CMAKE_CXX_STANDARD 17)
   endif()
   set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
and a check and proper error message in `cmake/Modules/Packages/ML-METATENSOR.cmake`
as it is done in `cmake/Modules/Packages/KOKKOS.cmake`

akohlmey:
> while most users inside and outside of Europe/USA are using more mainstream hardware, either on personal laptops/workstations or x66_64/Linux HPC clusters.

Yes, but *they* do need *that* kind of IT support just as well. When I ventured into the HPC business some 25 years ago, people *using* parallel software for simulations were also people capable of *programming* and *compiling* it and very much able to help themselves after been given some pointers. However, this has changed massively and most of the people *using* software and not *programmers* and often have no concept of programming (and often only limited understanding of the underlying physics of the software as well). In LAMMPS we have noticed this trend in an accelerated way. Particularly people with an engineering background are very eager but have a tendency to skip over details and expect software (or the people supporting it) to tell them exactly what to do (which may work for commercial software, but not elsewhere). The advent of ChatGPT has made things even worse, since I have seen especially inexperienced people putting undeserved trust into ChatGPT (I presume because its answers *sound* confident even if they are complete garbage). Check out this recent post: https://matsci.org/t/lammps-installation/56044

I suspect, if the LAMMPS developers decide to include ML-METATENSOR into the distribution, there needs to be a companion "Howto" document that describes the complete installation process step by step on a number of specifically named platforms and environments.

akohlmey:
@Luthaf CMake configure could be completed on Fedora 40 with the following "native" packages and their dependencies added:
- cargo
- python3-torch-devel 
- python3-torchvision 
- protobuf-devel

```
# dnf install cargo.x86_64 python3-torch-devel.x86_64 python3-torchvision.x86_64 protobuf-devel.x86_64
Last metadata expiration check: 2:05:27 ago on Mon 24 Jun 2024 12:07:52 PM EDT.
Dependencies resolved.
===========================================================================================================================================================================================
 Package                                           Architecture                      Version                                                      Repository                          Size
===========================================================================================================================================================================================
Installing:
 cargo                                             x86_64                            1.79.0-3.fc40                                                updates                            5.9 M
 python3-torch-devel                               x86_64                            2.1.2-11.fc40                                                updates                            4.5 M
 python3-torchvision                               x86_64                            0.16.0-7.fc40                                                fedora                             1.0 M
 protobuf-devel                                    x86_64                            3.19.6-8.fc40                                                fedora                             391 k
Installing dependencies:
 rust                                              x86_64                            1.79.0-3.fc40                                                updates                             25 M
 rust-std-static                                   x86_64                            1.79.0-3.fc40                                                updates                             34 M
 cpuinfo                                           x86_64                            23.2.14-3.giteb4a667.fc40.2                                  fedora                              42 k
 onnx-libs                                         x86_64                            1.14.1-2.fc40                                                fedora                             842 k
 pthreadpool                                       x86_64                            0.0^git20230829.4fe0e1e-4.fc40                               fedora                              45 k
 python3-mpmath                                    noarch                            1.3.0-7.fc40                                                 fedora                             1.1 M
 python3-sympy                                     x86_64                            1.12-4.fc40                                                  fedora                              14 M
 python3-torch                                     x86_64                            2.1.2-11.fc40                                                updates                             41 M
 sleef                                             x86_64                            3.5.1-31.fc40                                                fedora                             858 k
 xnnpack                                           x86_64                            0.0^git20221221.51a9875-4.fc40                               fedora                             424 k
Installing weak dependencies:
 python3-antlr4-runtime                            noarch                            4.10.1-14.fc40                                               fedora                             289 k
 python3-gmpy2                                     x86_64                            2.1.5-6.fc40                                                 fedora                             203 k
 python3-pyglet                                    noarch                            1.5.27-4.fc40                                                fedora                             1.7 M
 texlive-eulervm                                   noarch                            11:svn15878.4.0-71.fc40                                      fedora                              39 k
```

akohlmey:
I pushed a few more small changes to https://github.com/akohlmey/lammps/tree/ml-metatensor 
This reduces compiler warnings and implements the C++17 CMake guards I mentioned.
I also moved the examples to the correct folder and added logs.

The biggest problem now is that libmetatensor.so and libmetatenso_torch.so are built as shared libraries. They should be build as static libraries for LAMMPS. This would normally work since packages follow the "BUILD_SHARED_LIBS" variable. However in your package it is set to ON by default. Please check out the code in the KOKKOS package CMake support:
For downloaded KOKKOS (not the copy included in LAMMPS)
```
  # build KOKKOS downloaded libraries as static libraries but with PIC, if needed
  list(APPEND KOKKOS_LIB_BUILD_ARGS "-DBUILD_SHARED_LIBS=OFF")
  if(CMAKE_REQUEST_PIC)
    list(APPEND KOKKOS_LIB_BUILD_ARGS ${CMAKE_REQUEST_PIC})
  endif()
```

Or when using the internal KOKKOS sources:
```  # build KOKKOS internal libraries as static libraries but with PIC, if needed
  if(BUILD_SHARED_LIBS)
    set(BUILD_SHARED_LIBS_WAS_ON YES)
    set(BUILD_SHARED_LIBS OFF)
  endif()
  if(CMAKE_REQUEST_PIC)
    set(CMAKE_POSITION_INDEPENDENT_CODE ON)
  endif()
```
and then to reset to the original settings:
```
  target_link_libraries(lammps PRIVATE kokkos)
  if(BUILD_SHARED_LIBS_WAS_ON)
    set(BUILD_SHARED_LIBS ON)
  endif()
```
```

**Comment by Luthaf** [comment]
> The biggest problem now is that libmetatensor.so and libmetatenso_torch.so are built as shared libraries. They should be build as static libraries for LAMMPS. 

This is (mostly) the intended behavior. The idea is that ML models are exported (serialized) from Python to TorchScript, and some models might use custom TorchScript operations. These operations are implemented using native code and/or custom CUDA kernel, and distributed as shared libraries that must be loaded in the process before they can be used. Some of these extensions are relying on metatensor data structures, meaning their shared libraries need to also load metatensor/metatensor-torch. One example of such extension is https://github.com/Luthaf/rascaline/, which computes representations of atomistic systems that are then used to build the models. 

During model training, we load the shared libraries for the extension from Python, which then load/resolve symbols from metatensor/metatensor-torch. During MD simulations, we have code inside metatensor-torch to load extensions, but they still need to resolve the same symbols that come from the metatensor libraries. I'll have a look if I can statically link metatensor inside LAMMPS and use RTLD_GLOBAL shenanigans to resolve the symbols in dynamically loaded extensions, but this might be hard to do in a cross-platform manner.

I see why you would want most libraries to be statically linked inside LAMMPS, since dealing with shared libraries can be a hassle. Since libtorch itself will very likely be dynamically linked anyway (there is no static pre-built version AFAIK), I feel like the price has already been paid and it might not be worth the non-standard approach of resolving shared library symbols from the LAMMPS executable when dynamically loading extensions.

**Comment by akohlmey** [comment]
> I see why you would want most libraries to be statically linked inside LAMMPS, since dealing with shared libraries can be a hassle. Since libtorch itself will very likely be dynamically linked anyway (there is no static pre-built version AFAIK), I feel like the price has already been paid and it might not be worth the non-standard approach of resolving shared library symbols from the LAMMPS executable when dynamically loading extensions.

There is a big difference between libtorch and the metatensor libraries from the perspective of building/using/packaging LAMMPS. Libtorch is an *external* dependency. That is, it has to be made available in a global way, e.g. as a pre-packaged software or as an environment module or similar. That means that it doesn't matter how and where LAMMPS is installed or run, the library will be available.

In contrast to that, the metatensor libraries are built as part of LAMMPS and thus have to be bundled with it and thus any installation/packaging has to know about them and set them up and so on. This applies to the "make install" when building with CMake, but also to the installation of the LAMMPS python module ("make install-python") which includes the LAMMPS shared library in the wheel that is built and installed. It will just magically work with all enabled features if the shared library contains all dependent libraries or they are automatically found because they are in a "system location". This is not the case with ML-METATENSOR.

The more I explore this and the more I learn about the complications of building and installing, the more I am leaning towards suggesting to not include this PR into LAMMPS itself but build and distribute this as a plugin. Having to build the plugin is only a minor additional complication to the build process, while adding ML-METATENSOR is a major complication for building/installing LAMMPS for most people.

**Comment by Luthaf** [comment]
Two small points, and then I'll wait to see what other LAMMPS developers have to say about this!

> This applies to the "make install" when building with CMake, but also to the installation of the LAMMPS python module ("make install-python") which includes the LAMMPS shared library in the wheel that is built and installed

This should already work for the CMake build system, and I could make it work as well for the traditional Make build system (although in this case metatensor & metatensor-torch are treated closer to external libraries, installed in `lammps/lib/metatensor/usr/`) and the Python wheel build.

> Having to build the plugin is only a minor additional complication to the build process, while adding ML-METATENSOR is a major complication for building/installing LAMMPS for most people.

This additional complication will only impact users who want to use ML-METATENSOR, not all LAMMPS users. For these people, having to use a plugin is actually a small increase in complexity, since they now have to deal with two different repositories, with separate build instructions, documentation, etc. The main advantage of a plugin is that it would redirect user support requests from the LAMMPS forums to our own, which is fair enough!

-------

In general, I am willing to spend time to fix the build/install procedure and write documentation and step by step instructions for it. I would like to hear what other LAMMPS developers think about this, and I'd be happy to join a LAMMPS meeting to discuss this directly.

**Comment by akohlmey** [comment]
> This additional complication will only impact users who want to use ML-METATENSOR, not all LAMMPS users.

That is not entirely true. You are discarding people that try to package LAMMPS and they want to have as many features included in their packaging of LAMMPS as possible, this includes people that pre-compile LAMMPS for environment modules on supercomputing machines and HPC clusters. There is some contention with the existing LAMMPS distribution and the choices made in it, that complicate packaging, as it is. Also, there is a significant number of LAMMPS users that want to compile LAMMPS with "everything", just in case they need it (they usually don't).

> For these people, having to use a plugin is actually a small increase in complexity, since they now have to deal with two different repositories, with separate build instructions, documentation, etc.

Yes, but those would *only* be those that want to use ML-METATENSOR. The group of people that I mentioned above would generally not include external LAMMPS packages unless they are explicitly asked about it. ... and if those have managed to compile the metatensort libraries and thus navigated around the complexity of its built, they will likely consider the LAMMPS compilation / installation rather straightforward. :wink:

**Comment by athomps** [comment]
@Luthaf and @akohlmey Thanks for the detailed dialogue. I learned quite a bit from reading this.  ML-METATENSOR is a really nice addition to LAMMPS that enables a lot of new functionality.  However, because of its intrinsic complexity, it does put some strain on LAMMPS maintenance and testing, because we are responsible for all packages in LAMMPS. Specifically, the build-time PyTorch dependency, the Rust source code, and the additional MetaTensor shared libraries require special attention from the person building the code.   None of these are fatal flaws, but they do create a burden.  After reading the full discussion, I agree with the suggestion of @akohlmey (and kindly accepted by @Luthaf ) to switch from building ML-METATENSOR directly with LAMMPS to building it separately as a PLUGIN.

**Comment by rbberger** [comment]
First of all, thanks for all the effort for trying to make it into a regular LAMMPS package and adding the necessary CMake trickery. I've just installed it myself and it's not that problematic after all the work that went already in. 

Some general notes:
- I would object to any MKL related patching, that sounds like an upstream problem
- the `find_package(Torch REQUIRED)` seems redundant after `FetchContent_MakeAvailable`, since the inclusion of `metatensor-torch` should already force that.

That said, I do agree with @akohlmey that this will be better off as a plugin. My general sense is that LAMMPS is already doing too much in CMake at this point and this would go one step too far towards LAMMPS-as-package-manager territory. Requiring Rust doesn't help, but isn't a blocker. What concerns me more is that once it would become part of LAMMPS, we'd become responsible for making it work on all the different platforms/combos that people care about. Even if you already do that in some capacity, it would create an additional maintenance burden on LAMMPS itself to stay in sync and to do testing. 

As a side note: There are other tools for building complex software stacks in HPC. This reminded me that it might be worth looking into streamlining LAMMPS plugins in [Spack](https://spack.io). I'll look into packaging our [example](https://github.com/lammps/lammps/tree/develop/examples/plugins), maybe other projects like this one could then follow that model. Update: see https://github.com/spack/spack/pull/44977

**Comment by ceriottm** [comment]
Hi everyone, I had followed the technical discussion and discussed this much with @Luthaf. I just want to chime in to make  a couple of more "sociological" points. This is part of an effort we are making to have a more modular ecosystem for ML models and I thought that having a deeper level of integration with LAMMPS would have been good for adoption, and possibly in the long run reduce duplication of effort. 

That said I do acknowledge that at present this is still yet another ML interface, with more complexity because it is aimed at being more widely applicable. On my side, I can promise that there will be someone responsible for developing and maintaining this in the foreseeable future, and that we will be happy to step in to help with any direct or indirect effect of having this additional complexity. However if these assurances are not enough to convince you, of course no hard feelings and we will do otherwise. 

In view of the mid-term objectives of this project, I wonder if it would make more sense for us to maintain a fork of LAMMPS incorporating this code, which would be easier to merge back once the approach has proven itself by having real-life examples of multiple applications and ML architectures being deployed with it. Let me know what you think.

**Comment by akohlmey** [comment]
> reduce duplication of effort.

It is too late for that, because the ML-IAP package already offers a similar, modular integration and mix-and-matching of different components of machine learning potentials, including integration of pytorch based models and KOKKOS acceleration. It is not as sophisticated as your implementation, but that is besides the point. You *will* be duplicating efforts and there will likely be others trying the same at different levels of sophistication. That is just in the nature of these kinds of models and attractive to do, since it is rather easy to get support from funding organizations around the world for such efforts. 

> In view of the mid-term objectives of this project, I wonder if it would make more sense for us to maintain a fork of LAMMPS incorporating this code, which would be easier to merge back once the approach has proven itself by having real-life examples of multiple applications and ML architectures being deployed with it. Let me know what you think.

Please note that what you have is - as far as LAMMPS is concerned - a wrapper around your libraries. The complications are not really complications of the code itself, but having a seamless integration into the LAMMPS build process that would make it look like *all* of the code is integrated within LAMMPS. So the main integration effort is in having the build system work so that your libraries can be build automatically alongside LAMMPS and linked with it.

As far as building a plugin is concerned, the code will have to remain a valid pair style in the process, only a few lines of code (that you can borrow from `examples/plugins/morse2plugin.cpp`) are required to load the plugin and then it should behave just as any other pair style. There are already multiple styles where we integrate the building as a plugin into the distribution (KIM and ML-PACE) to avoid licensing issues when distributing *binaries* (as we do for Windows). Another package with PLUMED is in preparation for the same reasons (see PR #4209, the Plumed library didn't allow cross-compiling to Windows until recent). Those use the same CMake script code from `cmake/Modules/Packages`. In fact, it should be possible to compile the pair style at *all times* within LAMMPS by building/installing your libraries separately, then copy the two pair style source files to `src`, and augment the preprocessor/linker flags for the chosen build system to point to the folders with required header files and link with the required libraries.

We also have the lammps-package repository for code that is missing requirements for being included in the LAMMPS tree or was "abandoned" by its developers. All styles in that package are compiled as plugins which proves that even the CMake build system support can be maintained outside the LAMMPS tree. Also, it is easy to bring code into the fold as has happened to two styles that were in lammps-packages, but then the missing pieces have been added or unsupportable code removed and they were then integrated into the main LAMMPS distribution.

Since this is all open source, there is nothing stopping you from maintaining a fork, but my expectation would be that that is more effort than maintaining a plugin. But don't just take my word for it. You could reach out to the DeepMD developers and ask for their experience, since they are doing exactly that, maintaining a fork of LAMMPS for their machine learning potential.

**Comment by athomps** [comment]
@ceriottm You are free to maintain your fork separately.  However, as @akohlmey indicated, it will probably be less work for you and users if you go the plugin route, because you will be more insulated from changes in the rest of LAMMPS.  For young, rapidly evolving areas like this, I view any redundancy between ML-METATENSOR and ML-IAP and as not bad and maybe a good thing.

**Comment by Luthaf** [comment]
I looked a bit deeper into how to transform the code into a plugin, and there is one point where I'm unclear on what to do. The documentation mentions that 

> Plugins need to be compiled with the same compilers and libraries (e.g. MPI) and compilation settings (MPI on/off, OpenMP, integer sizes) as the LAMMPS executable and library. 

But I could not find a way to find this information when using a pre-compiled version of LAMMPS. I could not find a pkg-config/CMake config file created during installation, which could contain such information. The example in https://github.com/lammps/lammps/blob/develop/examples/plugins/LAMMPSInterfaceCXX.cmake seems to re-determine the set of options and compiler instead of re-using the ones from an existing build. 

Is the official solution here "document to users that they should be careful to use the same compilers and settings"?

----

Overall, even if we move this code in a plugin for now, we would still be interested to integrate it in LAMMPS later down the line. Would you be open to such a possibility? I can see a couple of scenario here:

1. we work to provide our code in apt/dnf/homebrew/..., and link the libraries as "system" shared libraries instead of having them built as part of LAMMPS build system. This would make them identical to how libpython.so or libtorch.so can be loaded now;
2. we manage to statically link our code with the LAMMPS binary, removing the need for shared libraries (see below);
3. we find a way to better integrate our code within LAMMPS build system, ensuring it works flawlessly for all users

(3) is how I started this PR, but seems to be hard and would increase the maintenance load on LAMMPS. Would (1) or (2) be acceptable solutions for you?

----

In particular, I looked into the technical side of (2) a bit more; i.e. what it would take to statically link metatensor_torch & metatensor in the main LAMMPS binary and/or library, while making sure these are still available to load for TorchScript extensions (i.e. Torch plugins).

 - on macOS, every seems to work fine, and the same extension plugin can be used both when it's dependency is statically linked and when it isn't
 - on Linux, this is much harder. One needs to compile the code with `-Wl,--export-dynamic`, which exports the binary symbols to be used with future DSO, and remove any DT_NEEDED tags in the extension plugin referencing the dependencies. I can think of a couple of ways to remove the DT_NEEDED tags on extensions, but would linking with `--export-dynamic` be a possibility on the LAMMPS side?
 - I haven't yet looked into Windows/BSD, this is on my TODO list if this is an acceptable solution.
 
If anyone reading this knows more about loading symbols in a DSO from either a shared library or an executable, please let me know!

**Comment by akohlmey** [comment]
> Is the official solution here "document to users that they should be careful to use the same compilers and settings"?

Historically, LAMMPS didn't have a "make install" (it is only available in CMake and not fully supported but rather added by people maintaining LAMMPS packages). Make install does create some .cmake files and a pkg-config liblammps.pc file, but they don't store any information about the configuration of LAMMPS.

Generally speaking, the expectation is that people build LAMMPS from source according to their own preferences and thus know the settings they use and can apply them during plugin compilation. 

> we work to provide our code in apt/dnf/homebrew/..., and link the libraries as "system" shared libraries instead of having them built as part of LAMMPS build system. This would make them identical to how libpython.so or libtorch.so can be loaded now

We don't do any packaging except for Windows (with a cross-compiler, and customized build scripts), so the decision to include your package and libraries lies entirely within the discretion of the people doing the packaging.  We don't interfere with that. 

> we would still be interested to integrate it in LAMMPS later down the line. Would you be open to such a possibility?

We are certainly open to changing our minds. But it may take a loooong time.

> on Linux, this is much harder. One needs to compile the code with -Wl,--export-dynamic, which exports the binary symbols to be used with future DSO, and remove any DT_NEEDED tags in the extension plugin referencing the dependencies. I can think of a couple of ways to remove the DT_NEEDED tags on extensions, but would linking with --export-dynamic be a possibility on the LAMMPS side?

That should happen automatically for *any* platform except on Windows when compiling with CMake because of these lines in CMakeLists.txt
```cmake
# re-export all symbols for plugins
if(PKG_PLUGIN AND (NOT ((CMAKE_SYSTEM_NAME STREQUAL "Windows"))))
  set_target_properties(lmp PROPERTIES ENABLE_EXPORTS TRUE)
endif()
```
https://cmake.org/cmake/help/v3.16/prop_tgt/ENABLE_EXPORTS.html
This may need to be extended to the "lammps" target for `-D BUILD_SHARED_LIBS=on` but I have not come across the need so far (and I build LAMMPS predominantly this way).

Luthaf:
> We are certainly open to changing our minds. But it may take a loooong time.

What are your requirements here? My understanding was that the main blocker was our use of shared libraries. If I can get the workaround with `ENABLE_EXPORTS` and `DT_NEEDED` manipulation working, we would be able to link both metatensor and metatensor-torch statically with both the `lmp` and `lammps` targets. 

What other things should we improve (including non technical points obviously!) before inclusion? 

Re-reading the discussion above, I can find the following:

- requirement for a Rust compiler: can be improved with docs (including platform-specific step by step guides), or by requiring the user to build the code outside of LAMMPS (moving this responsibility outside of the LAMMPS build system)
- picking between CUDA/CPU versions of libtorch: can be improved with docs/platform-specific guides
- MKL issue in libtorch: the missing `include` should be fixed now, the cmake patch can be removed by requiring torch >=2.3

Other points that we can't do much about, and thus might prevent inclusion forever:

- there is a dependency on libtorch, which needs to be provided by the user one way or another

Anything else?

akohlmey:
> What are your requirements here?

The main concern is the added complexity (for maintenance and to users that are compiling LAMMPS from source, which is the majority) versus the tangible gain for LAMMPS users. 

There currently is too much of the former (and some that you cannot easily remove, e.g. the fact that you need to download external source code and compile part of it with a different compiler) and very little of the latter (only the potential of benefits in the future assuming your project is actually successful). You are now arguing individual technical issues, but we are mainly looking at the "big picture". 

Your contribution is not the only project that has this kind of problem to get their code included and some of the currently included code is under consideration to be dropped for similar reasons (in most cases compounded by the lack of active development and support by the contributors, which does not apply here) and some contributions only made it into the package after heavy discussions and modifications (same as here) and despite the complexity because of the tangible benefits.

Luthaf:
Ok, so we need to show actual uses of this code to solve scientific problems in ways that where not possible before. Before the next submission, I'll make sure to compile a list of these & explain how they where enabled by metatensor!

Thank you for your time and feedback on the code and project, this is always useful.

PR #4169, opened 2024-05-22
CG-DNA: Real units and potential reader

ohenrich:
**Summary**

This PR implements the use of real units within the oxDNA, oxDNA2 and oxRNA2 bond and pair styles. The tool lj2real.py is included, which converts data and input files of similar format to those in the examples directory to their real unit equivalents, using analytically obtained conversion factors.

The oxDNA, oxDNA2 and oxRNA2 parameters can now be also set via the LAMMPS potential file reader class. The supplied potential files detail non-modifiable arguments for the bond and pair styles within the different models. As such, this gives the option to neaten the input script.
Modifiable values like the temperature, average sequence / sequence-specific stacking and hydrogen bonding switch, along with the salt concentration are not included in the potential files in order to allow easier access.

For consistency with previous results, this PR contains also now `trim_flag = 0` in the constructors of all pair classes to avoid default trimming of neighbour lists and restore the old (pre-stable_2Aug2023) behaviour.

**Author(s)**

Kierran Falloon & Oliver Henrich, Department of Physics, University of Strathclyde, Glasgow, UK

**Licensing**

**Backward Compatibility**

Backward compatibility is fully retained.

**Implementation Notes**

The implementation of real units was verified against the existing tests with LJ units.

**Post Submission Checklist**


sjplimp:
@ohenrich @akohlmey This all looks good with respect to the CGDNA packages and its examples.  I am unclear if this has any overlap with the examples/UNITS dir and its README about how to write an equivalent script in lj vs metal vs real units.  In particular, if the python scripts you include here would be useful for non-CGDNA scripts, like the one in examples/UNITS or an arbitrary input script that a LAMMPS user creates.

If the python scripts are meant to be useful beyond the CGDNA package, then they should probably be in the tools dir as opposed to within examples/PACKAGES/cgdna which I think is where they are now ?

ohenrich:
@sjplimp @stanmoore1 Thanks for the comments. I don't think the python script `lj2real.py` is meant to be used beyond the CG-DNA package. The script is  designed to handle the many bespoke interaction styles of the CG-DNA package and the fixes we typically use for simulations.

In particular, we have interaction styles that contain mixes of parameters with time, length and/or mass scale dependence which need converting, and parameters that express e.g. zero angles in radians and do not need changing. This makes the script barely generalisable.

akohlmey:
@ohenrich I am a bit unhappy about the choice of names for the potential files. Generally, the file suffix should indicate which pair style (or package) the potential file is for. See the README file in the potentials folder. So .lj and .real are not in line with this convention. How about renaming `oxdna.lj` to `oxdna_lj.cgdna` and equivalent for the other files?

ohenrich:
@akohlmey Totally fine with that. Would you be able to do this? It might take me a while as I'm quite absorbed with a proposal and a viva and Kierran is on jury duty.

sjplimp:
@akohlmey @ohenrich good catch - I am in favor of that name change as well

akohlmey:
> @akohlmey Totally fine with that. Would you be able to do this? It might take me a while as I'm quite absorbed with a proposal and a viva and Kierran is on jury duty.

Done. Please check it out.

sjplimp (approved):

PR #4177, opened 2024-05-28
Fourth Set of Collected Bug Fixes and Maintenance Updates for 2 August 2023 Stable release 

akohlmey:
**Summary**

This pull request collects bugfixes backported from the develop branch to ultimately become the fourth update to the latest stable release.

**Related Issue(s)**

N/A

**Author(s)**

Backporting done by Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

The following individual fixes and changes are included:
- fix a thread initialization issue in LAMMPS-GUI
- add workaround for fix not fomputed at compatible time with write_dump for timestep counts > 2 billion
- fix off-by-one bug in f2c string conversion in Fortran interface module
- fix bug in electron radius update for fix eff Nose-Hoover thermostat variants
- fix memory leaks in gather*concat() functions of the library interface
- avoid segfaults when changing timestep before run but after fix definition for Nose-Hoover fixes and fix shake/rattle
- update XDR support in dump xtc with code version that is GPL compatible
- Simplify detection and output of Windows versions
- fix bug in GPU/CPU overlap in KOKKOS
- Must always return nfaces == 0 when there are fewer than 3 vertices for rounded polyhedrons in 3d
- fix atom map bug in fix pour and fix deposit
- fix bugs in argument parsing for fix ave/correlate and fix ave/correlate/long
- fix multiple bugs in fix ipi
- fix logic bug in verbose output for NEB
- fix bug when requesting occasional neighbor lists where the list would not always be rebuilt properly
- Do not reference cuFFT when compiling the GPU package with CUDA
- fix Kokkos shrinkwrap bug when having groups without atoms
- fix logic bug in compute spin
- fix bug in converting strings to 64-bit integers on Windows
- fix hbondchk bug in ReaxFF/OpenMP
- fix bug in the SPH package where velocities are only partially remapped by a deformation fix (see PR #4243 for details)
- fix bug in reading incomplete label maps from restart files. Cherry-picked from PR #4252 
- allow using compute spin for groups other than "all"
- make vector style variables compatible with chunks
- symmetrize cutoff for pair style ylz
- make "mass" property in PyLammps class behave similar to "rmass"
- force using sphinx 7.2 and older so we remain compatible with breathe
- add vectorstyle variable check to fix ave/histo
- do not output data from fix reaxff/bonds during setup() except for the very first time
- avoid infinite loop with cg minimizer and small forces
- support that cmdargs argument is not modified when creating LAMMPS object and thus can be reused
- fix logic bug with newton bond off for compute stress/cartesian
- may install runtime dll dependencies for LAMMPS library only when building with -DBUILD_SHARED_LIBS=on
- remove spurious dependency of GPU package on cuFFT (which is currently not used)
- fix compilation issue with recent versions of libAtoms/QUIP
- lock GitHub actions on macOS to macOS 13 as macOS 14 images are incompatible and fail.
- set up and install python packages into a virtual environment for GitHub macOS runner to avoid issues with modifying system files
- add compatibility with plumed 2.9.x
- add support to compile PLUMED as plugin to enable distribution of Windows binary packages
- make LAMMPS python module compatible with mpi4py version 4.x.x
- avoid issues with write_dump re-using and existing dump ID when a previous command crashed
- improve error handling when parsing the MEAM user parameter file: catch value tokenizer exceptions record line numbers and report filename and line number with the error

**Post Submission Checklist**


src/EXTRA-DUMP/xdr_compat.h
```diff
@@ -1,100 +1,70 @@
...
- * in their system libraries. It was written for windows, but will
- * most probably work on other platforms too. better make sure you
- * test that the xtc files produced are ok before using it.
+ * This file contains the definitions for Sun External Data Representation (XDR).
+ * They have been adapted specifically for the use with the LAMMPS xtc dump style
+ * to produce compressed trajectory files in the Gromacs XTC format.
  *
- * It is also needed on BG/L, BG/P and Cray XT3/XT4/XT5 as we don't
- * have XDR support in the lightweight kernel runtimes either.
+ * The XDR sources are avaiable under the BSD 3-clause license for example in
```
jtclemm: Found a typo: "avaiable" 

Ditto in xdr_compat.cpp

akohlmey: Thanks. Fixed it. Also for "develop"

akohlmey:
@stanmoore1 I've tried to backport commit 8bba2d12ec12dbb065a3db9a029088b10e033c65 from upstream. Please have a look.

sjplimp (approved):

stanmoore1:
> @stanmoore1 I've tried to backport commit [8bba2d1](https://github.com/lammps/lammps/commit/8bba2d12ec12dbb065a3db9a029088b10e033c65) from upstream. Please have a look.

Looks good, thanks Axel.

PR #4181, opened 2024-05-31
Charge Transfer Interatomic Potential (CTIP) pair style and qeq fix

gplummer317:
**Summary**

Implementation of CTIP, which is a variable charge potential and an improved version of the existing coul/streitz potential (see Zhou et al., Phys Rev B, 69, 035402 (2004)). CTIP implements a charge bounding scheme which provides greater numerical stability to the potential and allows for more physical charge transfer in systems with >2 components.

There are also several efficiency improvements to the original CTIP model which have been incorporated here:
-Charge bounds enforced with a piecewise quadratic function which is compatible with the existing qeq matrix inversion algorithm.
-Long range electrostatic interactions are calculated with a damped, shifted potential rather than Ewald summation (see pair_style coul/dsf).
-Coulomb integrals are replaced with an empirical charge shielding function as in the reaxff potentials.

**Related Issue(s)**

**Author(s)**

Gabriel Plummer - NASA Ames Research Center
gabriel.w.plummer@nasa.gov
gplummer317@gmail.com

**Licensing**

**Backward Compatibility**

**Implementation Notes**

3 new parameters have been added to the qfile which is read in as part of the fix qeq commands to enforce charge bounding - qmin, qmax, omega. This should not affect the other fix qeq styles which will not utilize these parameters.

2 new keywords have been added for fix qeq/ctip - cdamp and maxrepeat which control the damping of the coulomb potential and the self consistency of qeq calculations with the piecewise quadratic function, respectively.

An Ni-O CTIP parametrization has been developed to perform tests on the new features - it utilizes the hybrid/overlay pair style in combination with an eam/fs potential. This parametrization can be made available by request if any testing is needed.

**Post Submission Checklist**


src/QEQ/fix_qeq_ctip.cpp
```diff
@@ -0,0 +1,394 @@
...
+  if (nlocal > n_cap*DANGER_ZONE || m_fill > m_cap*DANGER_ZONE)
+    reallocate_matrix();
+
+  for (i=1; i <= maxrepeat; i++) {
+       init_matvec();
+       matvecs = CG(b_s, s);         // CG on s - parallel
+       matvecs += CG(b_t, t);        // CG on t - parallel
+       matvecs /= 2;
+       n=calculate_check_Q();
+       MPI_Allreduce(&n, &nout, 1, MPI_INTEGER, MPI_SUM, world);
```
akohlmey: MPI_INTEGER is a Fortran MPI interface constant. In C/C++ this would be MPI_INT.

```suggestion
       MPI_Allreduce(&n, &nout, 1, MPI_INT, MPI_SUM, world);
```

src/QEQ/fix_qeq.cpp
```diff
@@ -53,12 +55,12 @@ namespace {
 
 FixQEq::FixQEq(LAMMPS *lmp, int narg, char **arg) :
   Fix(lmp, narg, arg), list(nullptr), chi(nullptr), eta(nullptr),
-  gamma(nullptr), zeta(nullptr), zcore(nullptr), chizj(nullptr), shld(nullptr),
+  gamma(nullptr), zeta(nullptr), zcore(nullptr), qmin(nullptr), qmax(nullptr), omega(nullptr), chizj(nullptr), shld(nullptr),
```
stanmoore1, line 58: These variables are not used by the parent `FixQEq` class and therefore should be moved to the child `FixQEqCTIP` class instead.

src/QEQ/fix_qeq.cpp
```diff
@@ -53,12 +55,12 @@ namespace {
...
   s(nullptr), t(nullptr), s_hist(nullptr), t_hist(nullptr), Hdia_inv(nullptr), b_s(nullptr),
   b_t(nullptr), p(nullptr), q(nullptr), r(nullptr), d(nullptr),
   qf(nullptr), q1(nullptr), q2(nullptr), qv(nullptr)
 {
-  if (narg < 8) utils::missing_cmd_args(FLERR, "fix " + std::string(style), error);
```
stanmoore1: This PR should not revert a previous change.

src/QEQ/fix_qeq.cpp
```diff
@@ -339,6 +343,12 @@ void FixQEq::setup_pre_force(int vflag)
   if (force->newton_pair == 0)
     error->all(FLERR,"QEQ with 'newton pair off' not supported");
 
+  if (force->pair) {
```
stanmoore1: Not sure if this check is necessary--the suffix styles should be able to fall back to the non-accelerated styles in most cases.

src/QEQ/fix_qeq.cpp
```diff
@@ -777,7 +790,7 @@ void FixQEq::read_file(char *file)
         auto values = reader.next_values(0);
 
         if (values.count() == 0) continue;
-        if (values.count() < 6)
+        if (values.count() < 9)
```
stanmoore1: We want to maintain backwards compatibility, so these changes to the `read_file()` function need to be moved to the child class so the parent class is unchanged.

src/QEQ/fix_qeq.h
```diff
@@ -100,6 +100,11 @@ class FixQEq : public Fix {
 
   double alpha;
 
+  // ctip
+
+  double cdamp;
```
stanmoore1: Move to child class.

src/QEQ/fix_qeq.h
```diff
@@ -70,10 +70,10 @@ class FixQEq : public Fix {
   int maxwarn;                 // print warning when max iterations was reached
   double cutoff, cutoff_sq;    // neighbor cutoff
 
-  double *chi, *eta, *gamma, *zeta, *zcore;    // qeq parameters
+  double *chi, *eta, *gamma, *zeta, *zcore, *qmin, *qmax, *omega;
   double *chizj;
   double **shld;
-  int streitz_flag, reax_flag;
+  int streitz_flag, reax_flag, ctip_flag;
```
stanmoore1, line 76: I think this `ctip_flag` flag is OK to keep in the parent class so the parsing doesn't need to be duplicated in the child class.

stanmoore1 (changes requested):
@gplummer317 thank you for the contribution. I have a few requests for changes. Also, can you please add documentation for the new styles, and add an example that we can use for testing? Thank you

gplummer317:
@stanmoore1 I'm sorry for the delayed response on this as I've recently been busy with other projects. I've now added documentation for the CTIP pair style and fix qeq commands along with an example. The documentation is included on the general fix qeq and pair_style coul pages as it seemed to fit in best there rather than creating entirely new pages. The new commands are closely related to the existing pair_style coul/streitz and fix qeq/slater commands as described below.

Regarding your requested changes, I modeled these new commands after the existing pair_style coul/streitz and fix qeq/slater commands, as they are essentially meant to be an update - the old commands are not able to handle systems with multiple cationic or anionic species as they give unphysical charge transfer (see the Zhou et al paper in my summary for context). With that in mind, would it not make sense to update the parent  qeq class as well? Otherwise wouldn't it be like creating another fix qeq/comb command, which is a standalone command from the other qeq styles? I think it would be preferable for this new style to fit in seamlessly with the others. That's why I included the definition of the additional parameters in the parent qeq style - it should not affect the existing qeq styles so long as the qfile has the proper format, e.g. none of the qeq styles use all of the parameters in the qfile. Let me know if this does not work though as I am not too familiar with how the overall LAMMPS code is structured but am always interested in learning.

Thanks!
Gabriel

src/QEQ/fix_qeq_ctip.cpp
```diff
@@ -0,0 +1,394 @@
...
+void FixQEqCTIP::compute_H()
+{
+  int inum, jnum, *ilist, *jlist, *numneigh, **firstneigh;
+  int i, j, ii, jj;
+  double **x;
+  double dx, dy, dz, r_sqr, r, reff;
+  double cutoffsq, cutoffcu, cutoff4, cdampcu, erfcc_cut, erfcd_cut, t_cut;
+  double erfcc, erfcd, t;
+  int elt1, elt2;
+  double shield[atom->ntypes][atom->ntypes], shieldcu[atom->ntypes][atom->ntypes], reffc[atom->ntypes][atom->ntypes], reffcsq[atom->ntypes][atom->ntypes], reffc4[atom->ntypes][atom->ntypes], reffc7[atom->ntypes][atom->ntypes];
```
rbberger: Variable length arrays (VLAs) are not valid C++ (it's a g++ extension). Please avoid.

src/QEQ/fix_qeq_ctip.cpp
```diff
@@ -0,0 +1,394 @@
...
+  double r = cutoff;
+  double rsq = r*r;
+  double r6 = rsq*rsq*rsq;
+  
+  double erfcd_cut = exp(-cdamp * cdamp * rsq);
+  double t_cut = 1.0 / (1.0 + EWALD_P * cdamp * r);
+  double erfcc_cut = (t_cut * (A1 + t_cut * (A2 + t_cut * (A3 + t_cut * (A4 + t_cut * A5)))) * erfcd_cut) / r;
+
+  int elt1;
+  double reff[atom->ntypes], reffsq[atom->ntypes], reff4[atom->ntypes], reff7[atom->ntypes], s2d_self[atom->ntypes];
```
rbberger: More VLAs.

akohlmey:
> With that in mind, would it not make sense to update the parent qeq class as well?

No. However flawed those styles may be from your perspective, they implement well documented algorithms and should keep doing that. You can just change them and produce unexpected results. Rather people would have to implement improved versions as derived classes.

src/QEQ/fix_qeq_ctip.cpp
akohlmey: @gplummer317 This *must* be corrected. Otherwise, we cannot merge your pull request, as it will no pass all tests.

For the pair style, you can just move the arrays as pointer variables into the header and allocate them with memory->create() and copy over the data in the setup_params() function (make certain to initialized the pointers to nullptr in the constructor and first destroy() and then create() them and also destroy() them in the destructor). This should give you a (small?) performance improvement, since you allocate and initialize them only once per run and not once for every call to compute().

For the fix a similar solution needs to be found. I've made some formal/cosmetic updates and merged with upstream.

akohlmey:
@gplummer317 Please let us know if you need assistance with the necessary changes. Please note that we so far have only done a partial review since the issues so far are fundamental and need to be resolved before a more detailed look is warranted. So there will be more detailed reviews. We have spent a lot of time improving the quality and reliability of LAMMPS we don't want to lose that.

gplummer317:
@akohlmey Thanks for your comments. Yes, I think it would be good to get some assistance with the changes you have recommended. I am not an experienced C++ developer, so most of my changes have been based on how I see things structured in the existing LAMMPS code. From my perspective, I don't really mind how these new features are implemented in LAMMPS, but rather I just want them to be accessible for the wider community to use. So, whatever you recommend for a seamless integration sounds fine to me, and hopefully I can learn some new things along the way.

akohlmey:
>  Yes, I think it would be good to get some assistance with the changes you have recommended.

Ok. I will work on this over the next few days and adjust the code the way I would prefer it.
I will ping you again, when it is ready for testing.

gplummer317:
> > Yes, I think it would be good to get some assistance with the changes you have recommended.
> 
> Ok. I will work on this over the next few days and adjust the code the way I would prefer it. I will ping you again, when it is ready for testing.

Thank you, Axel - I appreciate the assistance

src/QEQ/fix_qeq.cpp
akohlmey: Actually, it is better to leave them in the base class or else the entire file parser would have to be duplicated

src/QEQ/fix_qeq.cpp
akohlmey: fixed

src/QEQ/fix_qeq.cpp
akohlmey: this is another case of removing old changes by copying files instead of merging. fixed.

src/QEQ/fix_qeq.h
akohlmey: done. also for "alpha" which belongs to qeq/slater

src/QEQ/fix_qeq.h
akohlmey: exactly.

src/QEQ/fix_qeq_ctip.cpp
akohlmey: fixed

src/QEQ/fix_qeq_ctip.cpp
akohlmey: fixed

akohlmey:
@gplummer317 This should be close to full compliance with the requirements for adding to LAMMPS a summary of my changes yesterday and today:
- move pair style to EXTRA-PAIR package
- restore changes in upstream that were undone by copying over modified files instead of merging
- replacement of all variable length arrays. They are no longer allocated and initialized in every iteration but once
- various small changes to integrate the sources cleanly into the source tree
- various small changes to integrate the documentation cleanly into the manual
- whitespace fixes, clang-format
- file parsing in fix qeq works again for all other styles
- some updates to LAMMPS programming style, e.g. ewald constants imported from ewald_const.h
- added a force-style test to the unittest tree
- added metadata to the EAM file
- moved potential files to potentials folder (with update of README) and symlinks in the example folder
- recreated example logs for 1 process and 4 processes using the naming conventions from other examples

akohlmey:
All tests pass now.

gplummer317:
> All tests pass now.

@akohlmey Wow, that was very quick! I checked the log files that you generated, and everything seems to be in order with regards to how the potential and charge equilibration should be behaving. I don't have any further changes at this time. Thanks for your help!

megmcca (approved):
in file review, resolved one comment; successfully ran example with make and cmake builds; checked off most of PR checklist (license info does not appear to be needed).

edit to add: 'changes requested' appears to be outdated and now resolved

megmcca:
@stanmoore1 the changes you requested look to be resolved, could you take a look?

stanmoore1 (approved):
I approve

doc/src/pair_coul.rst
```diff
@@ -20,6 +20,7 @@
 .. index:: pair_style coul/msm
 .. index:: pair_style coul/msm/omp
 .. index:: pair_style coul/streitz
+.. index:: pair_style coul/ctip
```
sjplimp: this list should be in alphabetic order I believe?

doc/src/pair_coul.rst
```diff
@@ -64,6 +65,9 @@ Accelerator Variants: *coul/msm/omp*
 pair_style coul/streitz command
 ===============================
 
+pair_style coul/ctip command
+============================
```
sjplimp: ditto alphabetic here

doc/src/pair_coul.rst
```diff
@@ -139,6 +143,9 @@ Examples
    pair_style coul/streitz 12.0 wolf 0.30
    pair_coeff * * AlO.streitz Al O
 
+   pair_style coul/ctip 0.30 12.0
```
sjplimp: ditto here

doc/src/pair_coul.rst
sjplimp: just the section from debye to wolf
the cout/cut and tip4p variants are separate

doc/src/pair_coul.rst
```diff
@@ -281,6 +288,32 @@ Streitz-Mintmire parameterization for the material being modeled.
 
 ----------
 
+.. versionadded:: TBD
```
sjplimp: likewise these subsections should probably be in the same alphabetic order as the command listings at the top of the page - you can reorder the existing ones to match that and add tip

sjplimp: basically for a doc page this long and complex we want to make it as easy as possible for users to find the info they want

sjplimp (changes requested):
Do both the pair coul/ctip and fix qeq/ctip files check that the complememtary fix and pair styles are defined?  If I read the could/ctip doc page correctly, this is a requirement?

akohlmey:
@megmcca do you want to take care of the requested changes, or should I implement them?

megmcca:
@akohlmey i can take care of it this week

doc/src/pair_coul.rst
megmcca: from this, i take it we'll keep coul/cut and coul/tip4p ordered as-is, and move coul/ctip to start of the post-cut styles (instead of very strictly alphabetic)? i.e.:

```
pair_style coul/cut
pair_style coul/cut/gpu
...
pair_style coul/cut/global/omp
pair_style coul/ctip
pair_style coul/debye
...
```

src/QEQ/fix_qeq_ctip.cpp
```diff
@@ -0,0 +1,438 @@
...
+    }
+  }
+}
+
+/* ---------------------------------------------------------------------- */
+
+void FixQEqCTIP::extract_ctip()
+{
+  Pair *pair = force->pair_match("^coul/ctip",0);
+  if (pair == nullptr) error->all(FLERR,"No pair style coul/ctip for fix qeq/ctip");
```
megmcca, line 183: check for pair_style coul/ctip exists

doc/src/pair_coul.rst
sjplimp: yes, that would be fine

sjplimp (approved):

PR #4191, opened 2024-06-09
New pair style hybrid/molecular

akohlmey:
**Summary**

This pull request is an attempt to implement a pair style hybrid/molecular that would allow to define different pair styles  for intra- and inter-molecular interactions. The pair is derived from pair style hybrid/overlay but requires exactly two sub-styles, the first for intra-molecular interactions, the second for inter-molecular. Same as with hybrid, and hybrid/overlay, the same pair style may be listed twice and then differ only by their pair_coeff settings

**Related Issue(s)**

This was discussed during the LAMMPS Master Class Workshop 2024
Also on MatSci.org at https://matsci.org/t/using-different-force-field-for-particles-in-different-molecules-and-in-the-same-molecule/55404

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

This implementation piggybacks on the skip list neighbor list build, but uses an additional flag to indicate whether a regular, atom-type based skip list is desired or a mol ID based skip list for intra-/inter-molecular.

**Post Submission Checklist**

- [x] Need to test with OPENMP package
- [x] Need to test with GPU package
- [x] Need to test with KOKKOS package
- [x] Need to test with INTEL package


akohlmey:
@stanmoore1 this new pair style will currently be refused by KOKKOS since it is missing the corresponding KOKKOS pair style implementation. Also, the neighbor list build for skip list will need to be updated. So it will behave as expected.

I have not yet looked into how much effort it would be to add the KOKKOS changes. Perhaps we should briefly discuss on slack? I don't think it is very urgent since the use case is rather unusual.

jtclemm (approved):

PR #4193, opened 2024-06-10
Create new EXTRA-COMMAND package

akohlmey:
**Summary**

As discussed at the LAMMPS Master Class workshop the group2ndx and ndx2group commands should be moved to the EXTRA-COMMAND package. No other current commands make sense to be moved there.

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

There are some updates to error handling. Gromacs does not allow group names with space (LAMMPS apparently does even though I have not seen anybody using that). _ndx2group_ would already throw a cryptic error message on such files, this is now down in a more consistent way with a meaningful error message. _group2ndx_ will now skip over processing groups with whitespace in the name.

**Post Submission Checklist**


doc/src/Packages_details.rst
```diff
@@ -886,6 +887,22 @@ This package has :ref:`specific installation instructions <electrode>` on the
...
 ----------
 
+.. _PKG-EXTRA-COMMAND:
+
+EXTRA-COMMAND package
+---------------------
+
+**Contents:**
+
+Additional compute styles that are less commonly used.
```
giacomofiorin: Did you mean specifically "compute styles" instead of "commands"?

giacomofiorin (approved):
As discussed earlier, these two commands are much more broadly useful than just supporting Colvars.

Thank you for moving them!

doc/src/Packages_details.rst
akohlmey: No. This is a copy-n-paste error.

sjplimp:
@akohlmey @giacomofiorin I think the doc page for the 2 commands should be expanded in 3 ways:
a) Give some sample use cases for these commands: for Gromacs compatibility or data exchanges, for Colvars, and for other use cases you envision
b) Describe the format of the file - if you don't do this then it is only a Gromacs-related tool and someone has to be familiar with Gromacs to use the command
c) Describe how to output this file from Gromacs, or to use it with Gromacs if LAMMPS outputs an altered file

giacomofiorin:
> @akohlmey @giacomofiorin I think the doc page for the 2 commands should be expanded in 3 ways: a) Give some sample use cases for these commands: for Gromacs compatibility or data exchanges, for Colvars, and for other use cases you envision b) Describe the format of the file - if you don't do this then it is only a Gromacs-related tool and someone has to be familiar with Gromacs to use the command c) Describe how to output this file from Gromacs, or to use it with Gromacs if LAMMPS outputs an altered file

@sjplimp Good suggestions.

@akohlmey I don't mind taking a stab at this, since we have already content from the Colvars doc. Let me know what you think.

akohlmey:
> Let me know what you think.

@giacomofiorin please go ahead. I have a few ideas of my own, but I can always edit them in later.

akohlmey:
@giacomofiorin didn't have the patience to wait any longer and added the changes I have in mind. Please let me know if you want to make any changes/updates to this.

giacomofiorin:
> @giacomofiorin didn't have the patience to wait any longer 

:-D  Okay, I guess.

> and added the changes I have in mind. Please let me know if you want to make any changes/updates to this.

What you wrote looks more or less like part of what I wanted, too.

The main reason why I haven't pushed anything yet (other than trying to finish unrelated work first) is that I also wanted to confirm how these two LAMMPS commands handle spaces within a group's name.

Because the GROMACS `make_ndx` tool does not allow them, they are not checking for them, which can lead to some interesting errors: 
https://gromacs.bioexcel.eu/t/colvar-output-force-and-work/3191/10

giacomofiorin:
LGTM

akohlmey:
> The main reason why I haven't pushed anything yet (other than trying to finish unrelated work first) is that I also wanted to confirm how these two LAMMPS commands handle spaces within a group's name.

That is an interesting subject.
I have never seen group names with spaces, but apparently that is allowed in LAMMPS if you put the name in quotes (I just checked it).

The ndx2group command would throw an error if there was a section with two or more words, but a cryptic one, so I changed the code to have more meaningful error messages.

The group2ndx command would happily create an index file with an invalid header, so I changed the code to skip over groups with whitespace in their name. The alternative would be to replace with whitespace with a non-whitespace character, e.g. '_' an underscore. But I'll think about this in case someone is actually complaining about not being able to write out groups with spaces in them.

akohlmey:
@sjplimp does this address your concerns?

PR #4197, opened 2024-06-13
library: add comm->procgrid to extract_global

rbberger:
**Summary**

Adds access to procgrid via the library interface

**Related Issue(s)**

**Author(s)**

@rbberger 

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/python/lammps/core.py
+++ b/python/lammps/core.py
@@ -891,7 +891,7 @@ class lammps(object):
     # set length of vector for items that are not a scalar
     vec_dict = { 'boxlo':3, 'boxhi':3, 'sublo':3, 'subhi':3,
                  'sublo_lambda':3, 'subhi_lambda':3, 'periodicity':3,
-                 'special_lj':4, 'special_coul':4 }
+                 'special_lj':4, 'special_coul':4, 'proc_grid':3 }
     if name in vec_dict:
       veclen = vec_dict[name]
     elif name == 'respa_dt':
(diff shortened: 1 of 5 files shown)


sjplimp:
@rbberger @akohlmey This looks fine.  However, note that procgrid[3] is only defined in comm.h for a logical 3d grid of procs (layout UNIFORM or NONUNIFORM).  It is not defined for a TILED layout, e.g. when using RCB via balance or fix balance.  So I am not sure if the lib interface should error check for that or return something different, e.g. mysplit[3][2] ??

rbberger:
@sjplimp we have other parts of the library interface that are accessible but make no sense to access in certain cases. E.g. when we have triclinic boxes some values are not defined. There are some other comm quantities as well that might be worth exposing, but this was the minimal change I needed for a collaborator, so I started this PR.

sjplimp:
So in all those cases, why don't we either print a warning (or trigger an error), or return something like a null ptr so the caller can check it.  If we return procgrid when the values are unset, that seems like a bad idea.  Ditto for anything similar for orthogonal vs triclinic.

rbberger:
Added the necessary check for layout to the new `proc_grid` global. Otherwise, it'll return a `nullptr`.

akohlmey:
@rbberger is there a specific reason the keyword is "proc_grid" instead of the name of the class member "procgrid"?

rbberger:
Was following the naming scheme of other globals, that don't necessarily exactly match the underlying variables.

akohlmey:
> Was following the naming scheme of other globals, that don't necessarily exactly match the underlying variables.

Hmm... In most cases they *do* match and most of the time when there is an underscore it is added to represent a dereference, e.g. 'respa_levels' for `respa->nlevels'.

Not that I am opposing it, but I prefer to have these match the variable name as much as possible since that is easier to remember.

akohlmey:
FWIW, I've also added some more flags from the Comm class so that one can check without having to deal with None.

rbberger:
Fine by me either way.

akohlmey (approved):

PR #4198, opened 2024-06-14
Add functionality to change and manage atom types through LAMMPS Pyth…

superde1fin:
**Summary**

This pull request adds functionality to create, delete, switch between, and reset type sets in the LAMMPS Python module. A type set is a collection of types assigned to each atom in LAMMPS, originally stored in the `Atom::type` array. The new functionality allows for multiple versions of this array to be stored and switched between efficiently. This includes:
- Addition of type sets (versions of the `Atom::type` array)
- Deletion of type sets
- Switching between type sets
- Resetting type sets to their original values

Additionally, two new global properties are introduced: `ntype_sets` and `current_typeset`. This functionality is particularly useful for evaluating global properties of specific structures in LAMMPS without the need for molecular dynamics simulations, enabling memory and complexity-efficient probing of type modifications on particular structures.

**Related Issue(s)**

This enhancement does not address a specific open GitHub issue but introduces new capabilities based on user needs and research requirements.

**Author(s)**

The functionality has been developed by Vasilii Maksimov, a member of the Wilkinson Glass Group at Alfred University.

**Licensing**

**Backward Compatibility**

The changes introduced in this pull request are backward compatible with existing inputs. The new functionality is additive and does not alter existing behavior.

**Implementation Notes**

The new functionality has been implemented by extending the LAMMPS Python module to support operations on type sets. The correctness of the implementation has been verified through unit tests and integration tests with the existing LAMMPS framework. The addition of `ntype_sets` and `current_typeset` global properties allows users to manage and query the state of type sets efficiently.

**Post Submission Checklist**

- [ ] The feature or features in this pull request are complete.
- [ ] Licensing information is complete.
- [ ] Corresponding author information is complete.
- [ ] The source code follows the LAMMPS formatting guidelines.
- [ ] Suitable new documentation files and/or updates to the existing docs are included.
- [ ] The added/updated documentation is integrated and tested with the documentation build system.
- [ ] The feature has been verified to work with the conventional build system.
- [ ] The feature has been verified to work with the CMake-based build system.
- [ ] One or more example input decks are included.

For more details and usage examples, please refer to the included documentation and example input decks.
P.S. This is my first attempt at modifying LAMMPS source code, so any commentary, critique, and suggestions from developers are more than welcome.

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -314,7 +314,8 @@ set(STANDARD_PACKAGES
   UEF
   VORONOI
   VTK
-  YAFF)
+  YAFF
+  VERSIONS)
 
 set(SUFFIX_PACKAGES CORESHELL GPU KOKKOS OPT INTEL OPENMP)
 
@@ -683,6 +684,10 @@ endif()
 
 if(PKG_H5MD)
   include(Packages/H5MD)
+endif()
+
+if(PKG_VERSIONS)
+  target_compile_definitions(lammps PUBLIC -DPKG_VERSIONS)
 endif()
 
 ######################################################################
--- a/python/lammps/core.py
+++ b/python/lammps/core.py
@@ -333,6 +333,9 @@ class lammps(object):
     self.lib.lammps_fix_external_set_vector_length.argtypes = [c_void_p, c_char_p, c_int]
     self.lib.lammps_fix_external_set_vector.argtypes = [c_void_p, c_char_p, c_int, c_double]
 
+    self.lib.lammps_check_symbol.argtypes = [c_char_p]
+    self.lib.lammps_check_symbol.restype = c_int
+
     #Type sets
     self.lib.lammps_change_typeset.argtypes = [c_void_p, c_int]
     self.lib.lammps_change_typeset.restype = c_int
@@ -1744,8 +1747,27 @@ class lammps(object):
     with ExceptionCheck(self):
       return self.lib.lammps_create_atoms(self.lmp, n, id_lmp, type_lmp, x_lmp, v_lmp, img_lmp, se_lmp)
 
+    
+
   #----------------------------------------------------------------
   #The follosing section provides a bridge to lammps functionality governing type sets' behavior
+
+  def check_symbol(self, symname):
+    """
+    Check if a specific preprocessor symbol is defined.
+
+    This function is a wrapper around the :cpp:func:`lammps_check_symbol`
+    function of the C-library interface.
+
+    It returns `True` if the symbol is defined, otherwise it returns `False`.
+
+    :param symbol: a string representing the preprocessor symbol to check
+    :type symbol: str
+
+    :return: `True` if the symbol is defined; `False` otherwise
+    :rtype: bool
+    """
+    return bool(self.lib.lammps_check_symbol(symname.encode()))
 
   def add_typeset(self, type_list):
     """
@@ -2488,3 +2510,4 @@ class lammps(object):
     computeid = computeid.encode()
     idx = self.lib.lammps_find_compute_neighlist(self.lmp, computeid, reqid)
     return idx
+
--- a/python/lammps/core.py
+++ b/python/lammps/core.py
@@ -333,6 +333,14 @@ class lammps(object):
     self.lib.lammps_fix_external_set_vector_length.argtypes = [c_void_p, c_char_p, c_int]
     self.lib.lammps_fix_external_set_vector.argtypes = [c_void_p, c_char_p, c_int, c_double]
 
+    #Type sets
+    self.lib.lammps_change_typeset.argtypes = [c_void_p, c_int]
+    self.lib.lammps_change_typeset.restype = c_int
+    self.lib.lammps_reset_typesets.argtypes = [c_void_p]
+    self.lib.lammps_reset_typesets.restype = c_int
+    self.lib.lammps_delete_typeset.argtypes = [c_void_p, c_int]
+    self.lib.lammps_delete_typeset.restype = c_int
+
     # detect if Python is using a version of mpi4py that can pass communicators
     # only needed if LAMMPS has been compiled with MPI support.
     self.has_mpi4py = False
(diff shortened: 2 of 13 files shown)


akohlmey:
@superde1fin Thanks for your submission.
I have some important questions, however: Why is it necessary to make changes to the LAMMPS C++ source code? And why are those features only accessible through the library interface?

You have access to per-atom data from the python wrapper through either `extract_atom()` where you can make local copies and changes to the existing type arrays or through `gather_atoms()` and `scatter_atoms()` and then all the functionality can be handled from local code alone and no changes to LAMMPS itself (or the library interface and the python module) would be required.

Another way to realize this functionality would be to use fix store/state. You can just create one fix instance for each set of types and then access those either directly or through atom style variables.
In a similar fashion you can use custom integer per-atom properties through fix property/atom. Those may even be per-atom arrays. Again, you can access those directly or through atom style variables.

So is there any particular reason why those alternatives cannot be used or are there any significant differences that *require* using modifications to the LAMMPS source code.

Also, there is the question, is there any need for this kind of feature outside of your own research?

To provide some context, we've had extended discussions for the "typelabel" feature that maps numerical types to strings and back, and changes have been integrated rather slowly and even though this can potentially enable LAMMPS to behave more like MD codes like NAMD or Gromacs, uptake and demand for this feature has been very slow.

Since the functionality you describe seems even more unusual, we need to think about this, since including your changes also means that those will need to be tested and maintained by the LAMMPS developers.

Whenever there is a request to add feature with "invasive" changes to the core functionality of LAMMPS (i.e. not add some optional code in some packages) then we are ultra-critical and want to make certain that there is no viable alternative.

Thanks in advance for understanding.

superde1fin:
Dear Dr. Kohlmeyer,

I understand your concern about the necessity of LAMMPS source code
modification. The reason for such addition is the attempt to make this
process as memory and computation efficient as possible. Indeed the methods
you have described to handle the behavior I have just added work, but to my
understanding a little less efficiently then a built-in approach. I might
be completely incorrect though. Mostly, I have added this code for my own
purposes, but decided to submit it as a pull request just in case the
developers of LAMMPS find it useful. In a different case, it would be
perfectly reasonable to neglect this pull request. As to the applications,
without going into much detail of my personal research (which as you have
correctly mentioned is only relevant to me) I believe this functionality
can be useful for any kind of position based energy landscape mapping using
LAMMPS. Again, thank you for considering the feature. Obviously feel free
to do whatever the developers team deems necessary. I just thought I could
share this in case somebody needed it.

Best regards,
Vasilii Maksimov

akohlmey:
> The reason for such addition is the attempt to make this process as memory and computation efficient as possible. Indeed the methods you have described to handle the behavior I have just added work, but to my understanding a little less efficiently then a built-in approach.

There is a simple way to prove this: make some benchmarks and let us know. I suspect that you are having a case of "premature optimization". That change of types would have to be outside a run or minimization and for that reason we don't put much effort in to making the "setup()" or "init()" functions maximally efficient, since the bulk of the time is spend outside of those during typical LAMMPS calculations.

I am also somewhat concerned about the correctness of your approach. I do not see any code to communicate the data you allocate during computations or when rebuilding the neighbor list or for other cases where atoms are communicated between subdomains (even applies for serial runs with periodic boundaries). You will need pack/unpack the data for forward, exchange and border communication. There will have to be a forward communication if you change types without also forcing a neighbor list rebuild as the types are also stored with ghost atoms. Forthermore, LAMMPS will regularly do a spatial sort of the atoms in the local arrays for better cache efficiency. Your way of storing will then change the order and the whole operation would be bogus.

More worrying is your loop over `Atom::natoms` in the library function. This is flat out wrong and will crash with a segmentation fault for any parallel calculations, since the per subdomain stored data has `Atom::nlocal` plus `Atom::nghost` atoms and such arrays are usually dimensioned with `Atom::nmax`. `Atom::natoms` is only equal to `Atom::nlocal` for serial calculations.

superde1fin:
Thank you so much for your explanation and remarks. I will make sure to
check everything you mentioned and will run some efficiency tests. It is
amazing that developers of LAMMPS are so quick to respond with useful
suggestions. Now Im glad I made the pull request.

Thank you again,
Vasilii

akohlmey:
> as memory [...] efficient

This is a rather pointless exercise unless you are doing simulations with billions of atoms, and even then the bulk of the memory in classical MD codes like LAMMPS is used for the neighbor lists (`natoms * <numneigh>`). Saving memory on per-atom data only makes sense if you do not use neighbor lists.

akohlmey:
You may want to check out Steve's lecture from the recent "Master Class" workshop: https://www.youtube.com/watch?v=fhNhZ6ilTOU
It discusses the various communication patterns in LAMMPS even though the more advanced ones: ring, irregular, and rendezvous do not apply to your feature.

akohlmey:
@superde1fin I had some exchanges with multiple core LAMMPS developers and the consensus was that we don't see a value for LAMMPS in your pull request for multiple reasons
- it is incompatible with atoms moving between subdomains or changing order in the local arrays due to sorting (mind you this also casts significant doubts on the validity of your research results using this feature).
- there are multiple ways to implement the same functionality with existing features, most of which do not have the issues listed above
- there is no corresponding LAMMPS command (which would mandate an implementation in the LAMMPS source) with access to the added functionality
- it can be implemented as pure python code even using the gather/scatter functions (and using global atom indexing instead of local indexing which avoids all the problems about parallel runs using subdomains)
- nobody could come up with an application where this would be useful
- your concerns about performance are misplaced

Thus our recommendations are:
- close this pull request
- consider one of the alternatives to implement this feature with less programming effort and thus less risk of unexpected bugs due to lack of familiarity with the inner workings of LAMMPS

I am sorry for the bad news, yet I believe - however painful and frustrating this may be in the short run - it will help you in the long run and give you more confidence in your research results.

superde1fin:
> @superde1fin I had some exchanges with multiple core LAMMPS developers and the consensus was that we don't see a value for LAMMPS in your pull request for multiple reasons
> 
>     * it is incompatible with atoms moving between subdomains or changing order in the local arrays due to sorting (mind you this also casts significant doubts on the validity of your research results using this feature).
> 
>     * there are multiple ways to implement the same functionality with existing features, most of which do not have the issues listed above
> 
>     * there is no corresponding LAMMPS command (which would mandate an implementation in the LAMMPS source) with access to the added functionality
> 
>     * it can be implemented as pure python code even using the gather/scatter functions (and using global atom indexing instead of local indexing which avoids all the problems about parallel runs using subdomains)
> 
>     * nobody could come up with an application where this would be useful
> 
>     * your concerns about performance are misplaced
> 
> 
> Thus our recommendations are:
> 
>     * close this pull request
> 
>     * consider one of the alternatives to implement this feature with less programming effort and thus less risk of unexpected bugs due to lack of familiarity with the inner workings of LAMMPS
> 
> 
> I am sorry for the bad news, yet I believe - however painful and frustrating this may be in the short run - it will help you in the long run and give you more confidence in your research results.

Dear Dr. Kohlmeyer,

Thank you for your consideration and advice. As I have expressed before, I mainly submitted this pull request to receive the commentary of the developing team which you have provided fully.

Best,
Vasilii

PR #4203, opened 2024-06-22

akohlmey:
**Summary**

This pull request collect multiple small changes and bugfixes that do not warrant a pull request of their own.

**Related Issue(s)**

Addresses a segfault mentioned here: https://matsci.org/t/about-run-style-respa/56075

**Author(s)**

Axel Kohlmeyer (Temple U), akohlmey@gmail.com

**Licensing**

**Backward Compatibility**

When updating LAMMPS from a git repository, a "make clean-all" is required in the doc folder

**Implementation Notes**

- include and enable sphinx-design extension for creating side-by-side formatting (in HTML)
- avoid segfaults when setting the timestep *after* defining a shake/rattle or Nose-Hoover fix with run_style respa. This causes calling `Fix::reset_dt()` and access pointers into the Respa class *before* they have been initialized during `Fix::init()`. Thus now also `Fix::reset_dt()` will initialize those pointers. Most other fixes are not affected since they set the variables taken from `Respa` only during `Fix::initial_integrate_respa()` of `Fix::final_integrate_respa()` and thus the pointers are initialized
- adjust `test_fix_timestep` program to trigger `Fix::reset_dt()` so that segfaults like above can be triggered
- update `test_fix_timestep` to use modernized accessor for looking up a fix instance by its id
- update documentation and reference it for `utils::bounds_typelabel()`; remove redundant "error" argument
- use consistent method to silence warnings about unused parameters for generic KOKKOS pair styles
- programming style refactoring in ML-POD package to be closer to the LAMMPS style
- documentation updates and corrections for ML-POD package and bioFF howto
- spelling and documentation fixes

**Post Submission Checklist**


sjplimp (approved):

akohlmey:
@exapde please review the last 6 commits in this pull request

exapde (commented):
The whitespace removal is good.

akohlmey:
> The whitespace removal is good.

Please also note the following:
- are you using `git add -A`? If yes, you should stop doing this as you added temporary files from (probably failed) attempts to run `ctest`. I personally prefer using `git gui` and selecting files or even individual lines or hunks of changes for commits so that those are `logical commits` even if there are multiple different changes in the same file. At any rate, you need to carefully check which files are added before committing
- the LAMMPS programming style states that member variables should be initialized in the constructor. I moved the corresponding code for the structs in the fitpod command accordingly
- compile time constants should use "static constexpr" instead of a define so we have proper type checking and checking whether a constant is used or not by the compiler
- the LAMMPS distribution has a bundled .clang-format file in the src folder. Thus using clang-format on files brings them automatically close(r) to the LAMMPS programming style (some things it messes up since it is not flexible enough) but those sections can be wrapped with `// clang-format off` / `// clang-format on`

exapde (approved):
Thanks for reformating and styling the code.

exapde (approved):
Thanks for make the changes to improve the code.

exapde:
> Please also note the following:
> 
> * are you using `git add -A`? If yes, you should stop doing this as you added temporary files from (probably failed) attempts to run `ctest`. I personally prefer using `git gui` and selecting files or even individual lines or hunks of changes for commits so that those are `logical commits` even if there are multiple different changes in the same file. At any rate, you need to carefully check which files are added before committing
> * the LAMMPS programming style states that member variables should be initialized in the constructor. I moved the corresponding code for the structs in the fitpod command accordingly
> * compile time constants should use "static constexpr" instead of a define so we have proper type checking and checking whether a constant is used or not by the compiler
> * the LAMMPS distribution has a bundled .clang-format file in the src folder. Thus using clang-format on files brings them automatically close(r) to the LAMMPS programming style (some things it messes up since it is not flexible enough) but those sections can be wrapped with `// clang-format off` / `// clang-format on`

I review your changes. They look very good. I will test the code more carefully when you merge your branch to the develop branch.

Thanks very much for the programing tips. They are very helpful.

src/ML-POD/compute_podd_atom.cpp
```diff
@@ -58,6 +58,8 @@ ComputePODDAtom::ComputePODDAtom(LAMMPS *lmp, int narg, char **arg) :
   pod = nullptr;
   elements = nullptr;
 
+  if (((((MAXBIGINT*3.0)*atom->natoms)*podptr->nClusters)*podptr->Mdesc) > (MAXSMALLINT*1.0))
```
akohlmey, line 61: @exapde the statement below in line 63 was flagged by static code analysis as potentially overflowing a 32-bit integer. So I added a check, but please be aware that what you are doing here can potentially "strangle" machines, since your memory consumption will grow quadratically or possibly even worse with the number of atoms.

You want to store per-atom data (so that is going to consume `sizeof(double)*Atom::nlocal*size_peratom_cols`). Thanks to domain decomposition the amount of per-processor storage can be reduced by using more processors. However, you also multiply with `Atom::natoms` which is the global number of atoms and then have multiple datums per-atom, some of which look like they could also grow as the total number of atoms grows. This can very, very quickly lead to huge memory consumption for all but the smallest systems with a small number of atoms.

In general, we **strongly** discourage any arrays of the dimension `Atom::natoms` because its memory consumption for larger systems cannot be contained by using more processors. But in this case there is also the fact that this is already a per-local atom property with some additional prefactors.

exapde: @akohlmey Thanks a lot for doing this. 

The purpose of these compute classes is to train the POD potential on DFT datasets, which are typically a few hundred atoms at most. They are never used to run MD simulations. 

I use Atom::natoms to store per-atom data mainly because LAMMPS does not provide MPI assembly for the compute classes. I have to tally the per-atom descriptors into the global descriptors, similar to how forces are tallied. While I can do that with the pair-pod class, I do not know how to do the same with the compute pod classes.

akohlmey: > While I can do that with the pair-pod class, I do not know how to do the same with the compute pod classes.

This is something where you should ask the LAMMPS developers. Do you have access to the LAMMPS developer slack? That is the place to ask.

You can do a reverse communication with computes just as well as it is done (automatically) after the forces are computed. You just needed to define a pack_reverse_comm() and unpack_reverse_comm() function and flag how many items you need to collect from ghosts to the principal atoms.
There are over 15 such computes in the LAMMPS distribution currently.
You should also check out the recent talk by Steve on the "Master Class" workshop: https://www.youtube.com/@templelammps/streams, i.e.  https://www.youtube.com/live/fhNhZ6ilTOU?feature=shared

PR #4212, opened 2024-06-28

### LINKED ISSUE #4208: Documentation Error for the C-API
The C-API reference for `lammps_extract_variable` (https://docs.lammps.org/Library_objects.html#_CPPv423lammps_extract_variablePvPKcPKc)

Lists two strings with wich to acces the length of a vector:
- "GET_VECTOR_SIZE"
- "LMP_SIZE_VECTOR"

However it seems that only "LMP_SIZE_VECTOR" actually works.

Because you're not supposed to free the pointer you get when asking for a Vector Style Variable except for when you also pass "LMP_SIZE_VECTOR" for the group parameter to get he size of the vector, it's easy to accidentally free the wrong data which results in a crash.

### LINKED ISSUE #4210: [BUG] Segfault for neb command in verbose mode when cores per replica exceeds 2
**Summary**

When I execute the `neb` command with `verbosity verbose`, LAMMPS segfaults if I use more than 2 cores per replica via the `-partition` command line flag. 

I am using stable release 2Aug2023 - Update 3. I experience the issue on both the Windows Installer build and RHEL 7 Linux. 

**Expected Behavior**

Not segfaulting

**Actual Behavior**

The NEB computation immediately crashes and logs show only the zeroth timestep NEB data. The Valgrind stack trace points to line 621 of neb.cpp within `NEB::print_status()`, shown below:

```C++
if (print_mode == VERBOSE) {
  freplica = new double[nreplica];
  MPI_Allgather(&fnorm2, 1, MPI_DOUBLE, &freplica[0], 1, MPI_DOUBLE, roots);
  fmaxatomInRepl = new double[nreplica];
  MPI_Allgather(&fnorminf, 1, MPI_DOUBLE, &fmaxatomInRepl[0], 1, MPI_DOUBLE, roots);
}
```

I think the issue is that when there are greater than 2 processors per replica, there are more than `nreplica` processes with `color=1`, so the `freplica` and `fmaxatomInRepl` arrays are then of insufficient size. Surrounding the Allgather statements with `if (me == 0)` seems to solve the issue, but I'm not sure if that is the correct way to fix this.

**Steps to Reproduce**

Execute the `neb` command in `verbose` mode with more than 2 processors per replica (e.g., using `-partition 8x3`). Example script and  output is attached: [MWE.zip](https://github.com/user-attachments/files/16025710/MWE.zip)

Thank you!

akohlmey:
**Summary**
 
This pull request combines multiple small changes and bugfixes.

**Related Issue(s)**

Fixes #4208 
Fixes #4210 
Addresses issue from https://matsci.org/t/detect-end-of-run-due-to-an-elapsed-timeout/56226

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- fixes a logic error in NEB with verbose output (see #4210)
- add special function `is_timeout()` to variables, so that one can detect in an input script whether the timer is expired
- the Timer class timeout should be stored in a double and it is now possible to add fractions of a second (or minute, or hour) without truncation or setting an infinite timeout
- correct and improve documentation for `lammps_extract_variable()`. Simplify behavior by selecting between NULL and non-NULL for the group argument.
- add tests for `lammps_extract_variable()` and `lammps_extract_variable_datatype()`
- drop INTEL package from cross-compiled Windows package (no significant benefit over OPENMP for non-Intel compilers)
-  mention that adding a pair style may need updates to `Install.sh` file

**Post Submission Checklist**


sjplimp:
@akohlmey What does cross-compiled Windows package mean?  I assume it is still possible to use the INTEL package optimizations on a Windows machine?

akohlmey:
> @akohlmey What does cross-compiled Windows package mean? I assume it is still possible to use the INTEL package optimizations on a Windows machine?

Cross-compiled Windows package means LAMMPS compiled on Linux for Windows. E.g. what I provide here: https://packages.lammps.org/windows.html

The INTEL package optimizations are really only optimizations (compared to OPENMP) when you compile with a compatible vectorizing compiler. So far this applies to the Intel compiler and some variant of the AMD/HIP compiler.
So if somebody compiles LAMMPS on Windows with the native Intel compilers, then it may still be possible. However, I have not installed the Intel compilers on Windows for a while, since the last time I tried, they made a mess out of my Windows virtual machine and I had to re-install it.

[Update:]
When using the Intel compilers in Visual Studio, then the INTEL package should be automatically selected with a different mechanism. Also when compiling with other ways, the preset is just a preselection, so adding `-DPKG_INTEL` or using "make yes-intel" is always an option.

Initially I had been looking for an option to reduce the number of symbols after adding the PLUMED package to the Windows compilation, as that bumped the number of exported symbols over 64k, and the INTEL package came out as the obvious case after running some benchmarks. Since then I have reliazed, that I *must* compile PLUMED as a plugin (for licensing reasons same as ML-PACE) and thus this isn't strictly necessary anymore, but the point of not having any significant measurable benefit and the complexity remains and thus I kept dropping it from the preset.

sjplimp (approved):

PR #4213, opened 2024-07-01
Add granular surfaces built from triangles (3d) or line segments (2d)

sjplimp:
**Summary**

Add capability for granular surfaces to be used in conjunction with granular particle models.  Surface in this context mean a collection of triangles (3d) or line segments (2d).  This is to allow more geometrically complex objects and boundary conditions to be represented within granular models.

Two flavors of surfaces are planned.

(1) Global surfaces, where a copy of the collection of (large) triangles is stored by every processor.
(2) Local surfaces, where the collection of (small, roughly particle-sized) triangles is distributed across processors
     in the same way that particles are.

**Related Issue(s)**

N/A

**Author(s)**

Steve, Joel Clemmer (SNL), Dan Bolintineanu (SNL)

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

**Post Submission Checklist**


akohlmey:
FYI, translating the manual results in the following reported issues:
```
Fix style entry surface/global is missing or incomplete in Commands_fix.rst
Fix style entry surface/global is missing or incomplete in fix.rst
fix index entry surface/global is missing
Found 1 issue(s) with style index
Found 92 packages
Package GRANSURF missing in Packages_list.rst
Package GRANSURF missing in Packages_details.rst
Found 2 issue(s) with package lists
```

akohlmey:
All doc pages for new commands should have a `.. versionadded:: TBD` so that they'll list the version where this feature appeared. For new sections to existing commands, e.g. molecule, the tag should be added with the new paragraph describing the added feature.

akohlmey:
@sjplimp FYI, I updated the failing death test for the molecule command to check for the modified error message.

akohlmey:
I would expect that support for reading triangles from STL files (or converting them to a molecule file?) would be highly desirable by users of this feature. The file format is quite simple and may be text or binary. A simple reader is in tools/std_bin2text.cpp and also in create_atoms.cpp.

sjplimp:
@akohlmey - responses to your comments
yes, I haven't started working on doc pages yet
re: STL, I agree this is a good idea, I'll add it to the to-do list

akohlmey:
@sjplimp You should merge with upstream now so you have the enhanced tokenizer class for more specific pattern matching of molecule file keywords.
Because of these changes, there is now a merge conflict. I you would prefer to have me do the merge and resolve the conflict, just let me know.

sjplimp:
@akohlmey done - thanks for the updated Molecule class

doc/src/fix_modify.rst
```diff
@@ -167,6 +169,19 @@ will have no effect on the motion of the rigid bodies if they are
 specified in the input script after the fix rigid command.  LAMMPS
 will give a warning if that is the case.
 
+<<<<<<< HEAD
```
akohlmey: This looks like an incorrectly resolved merge conflict.

akohlmey:
@sjplimp This branch is breaking for example the test for atom style tri. When I compare the source to upstream I see comments about some hybrid style making settings, but what about the non-hybrid atom style? The existing default settings are removed, e.g. the default radius and that causes the test failures.

sjplimp:
@akohlmey We are trying to make atom_style tri now work both as a single style with point particles (old use case) as well as in hybrid mode (style = hybrid sphere tri) with finite-size spheres (new use case for granular).  Can you explain more about what is making your test fail (presumably the old case) ?  So that I can reproduce the issue.

akohlmey:
> Can you explain more about what is making your test fail (presumably the old case) ? So that I can reproduce the issue.

It is *our* test, not "mine".

The default(?) values for the atoms are no longer set. E.g. radius used to be 0.5 but now comes out as 0.0.

Check out for example: https://github.com/lammps/lammps/actions/runs/22791265403/job/66118133952?pr=4213#step:7:1311

According to the test log, the failing tests are in lines 1900 and 1901, 2022 and 2023, and 2026 and 2027 in the file `unittest/formats/test_atom_styles.cpp`.

jtclemm:
This is a test that'll need to be updated. Steve and I decided that it was meaningless/arbitrary to assign a particle radius of 0.5 to point particles when using atom style line/tri with the line/tri flag set to zero. This also produced unexpected conflicts when using hybrid atom styles. Extra logic was added to clarify how radii are calculated and ensure it works as intended.

I might have time to patch it this evening.

sjplimp:
@akohlmey I'm remembering more about how we got to this point.  The old usage of triangles (or lines) was only for use with point particles.  You could use atom_style tri by itself (no hybrid), and there is no specification of radii in the data file.  Instead the triflag of 0/1 meant point particle versus triangle.  The src/ASPHERE/pair_tri_lj.cpp code allows for point/point, tri/point, and tri/tri interactions and does not use the atom->radius vector.  

The new usage of granular triangles is only for use with spheroidal particles (can't use point particles).  To enable that you have to use atom_style hybrid tri sphere, so that the data file can now specify radii for non-tri particles.  The new pair surf/granular allows for sphere/sphere and tri/sphere interactions but not tri/tri.

So we are now explicitly setting radius = 0.0 for point particles and 1/2*diam for spheroids.  The logic in atom_style tri also has to reflect that for spherioids, there will be another atom style, which actually performs the setup for spheroids.  And it has
to work whether you use "sphere tri" of "tri sphere" in the atom_style hybrid command.

As @jtclemm said, that new logic is probably breaking some of the existing old tests for the old use case.

akohlmey:
@sjplimp @jtclemm There is now an option to draw smooth surfaces from triangle meshes, if you can devise suitable surface normals for the corners of the triangles (see PR #4925). I have already applied this to a few existing cases (e.g. ellipsoids, arrows, isosurface, and rounded polyhedron body particles). I am now wondering, if the changes in this pull request would benefit from that as well. Do you have any ideas on that?

doc/src/Howto_granular_surfaces.rst
```diff
@@ -0,0 +1,656 @@
...
+   For an object defined by two or more connected triangles/lines, it is
+   an error to assign a motion and not include all the connected
+   triangles/lines, since this would break the connections.  LAMMPS
+   checks this for *global* surfaces but only checks that the fix group
+   of any instances of :doc:`fix move <fix_move>` include all or none of
+   a set of connected *local* triangles/lines.
+
+----------
+
+Calculation of forces
```
sjplimp, line 356: The math symbols and eqs in this section did not show up well-formatted when I did a "make html" on my local repo.  Not sure why that is.  Also true in a few other places, but his section is math-rich.

doc/src/Howto_granular_surfaces.rst
```diff
@@ -0,0 +1,656 @@
...
+
+When designing geometries for granular surfaces, there are several
+things to keep in mind to avoid unintended force contributions and to
+improve efficiency.
+
+While convex corners are identified and used to censor forces from
+physically hidden sections of a wall, if a particle is not in contact
+with the entirety of a convex turn, then forces cannot be properly
+censored.  For example, consider a 2d system with a U shaped wall
+defined by 3 line segments (see figure).  If the width of the U is wider
```
sjplimp, line 625: Do not see this U-shaped figure ?

doc/src/Howto_granular_surfaces.rst
akohlmey: @sjplimp Please try running "make clean-all" in the doc folder, and try again. Watch for any warnings or errors during the process. If possible, also try make pdf to confirm the problem is with the MathJax setting and not LaTeX in general.

doc/src/Howto_granular_surfaces.rst
```diff
@@ -0,0 +1,656 @@
...
+(:math:`\delta_i = 0`), the limit in the convex scenario is only
+relevant for connections that are both flat and convex.  Secondly, if
+two flat lines are perfectly parallel, then :math:`\hat{n}_{f,i} =
+\hat{n}_{s,i} = \hat{n}_{s,j}` implying the concave/convex designation
+has no effect.  Lastly, if a point is shared by more than two lines,
+then LAMMPS finds which connecting line has a normal vector closest to
+:math:`\hat{n}_{s,i}` to determining whether its a concave or convex
+connection.
+
+To illustrate, these scenarios are visualized in the figure below.  In
```
sjplimp: I do not understand the 2 parts of this 4-circle figure with convex/concave line segments.  I assume the purpose of the figure is to show that the forces on the particle are "as expected" (in an intuitive sense) for several positions of the particle wrt the join-point of the 2 segments.  But I'm not getting that, either visually or from the text.  I'm also not clear on what the length of the arrows or their directions mean, since they always seem to be coming from a point on the surface ?

Also, I think the 2 gray arrows are confusing, they could be dropped.

I also don't understand why the paragraph seems to say there are 2 alternative ways to do it and one (or both) gives an unphysical wobble.  Why is it not saying "LAMMPS does it this way" and it is "correct" ?

doc/src/Howto_granular_surfaces.rst
```diff
@@ -0,0 +1,656 @@
...
+models with such particles.
+
+This page explains how you can also define granular surfaces which are
+a collection of triangles (3d systems) or line segments (2d systems),
+which act as boundaries interacting with the particles.  This is
+particularly useful for defining a complex wall or boundary geometry.
+As described below, particle/surface interactions can be specified
+with similar options as those for particle/particle interactions.
+
+----------
```
sjplimp, line 46: This is where I think it would be good to add a couple (or more) figures which illustrate this feature.  With pics of the kinds of cool granular surfs that can be defined.  To entice users to learn more about them.

They could be pulled from the examples.  Or if those are only toy problems, then from your archive of more interesting images with gran surfs.

I also think it would be good to list here the set of commands that are used for gran surfs, similar to what is at the top of the Howto granular doc page.

doc/src/pair_surf_granular.rst
```diff
@@ -0,0 +1,130 @@
...
+one of the two particles going to infinite radius and mass (flat
+surface).  Specifically, :math:`\delta` = overlap of particle with
+triangle/line, :math:`m_eff` = mass of particle, and the effective
+radius of contact `R_{eff}= R_i R_j/R_i + R_j` is set to the radius of
+the particle. See the :doc:`Howto granular surfaces
+<Howto_granular_surfaces>` doc page for information on how overlaps
+and normal vectors are calculated based on the geometry of the surface
+and when friction is transferred between lines/triangles.
+
+Pair coefficients are identical to those in the :doc:`granular
```
sjplimp: Like (almost?) all pair style doc pages, the syntax for the corresponding pair coeff command should be given here.  In particular it should explain that for the two type args, one needs to be for tri/line particles, and the other for spherical particle types -- is that true?

If not, it is somewhat confusing, and should also be explained, since that is different than how fix surf/global does it ?

The remaining args can simply be said to be the same and point to the pair granular doc page.

sjplimp:
@jtclemm - Read thru the doc pages.   Made a handful of small, cosmetic/typo edits.  Also added a few comments on bigger-issue questions.  I think we should add some nice pics of granular surf geometries and gransurf/particle simulations, so that users know what we are talking about.  This could be at the top of the Howto granular surfs doc page (see my comment).  Or it could be on the Pics/Movies web pages and we could point to them.  Or both places.

doc/src/Howto_granular_surfaces.rst
sjplimp: aah - that fixes the math formatting issue - thanks!

jtclemm:
> @jtclemm - Read thru the doc pages. Made a handful of small, cosmetic/typo edits. 

Thanks Steve, I'll work on fixing those issues. For images, how about one (or both) of these? I have no preference on the website vs. documentation Howto page.

If the Howto page, another option is to use dump image to generate the visuals. @akohlmey how well do you think the graphics package could handle these geometries?

image
image

akohlmey:
> . @akohlmey how well do you think the graphics package could handle these geometries?

Can you provide me with a couple of data files and input files? Just send them to my personal email, and I can see what I can do. If there is a way to compute a gradient map so there would not only be flat triangles but a smooth transition between them, we could potentially do better for the surfaces.

sjplimp:
@jtclemm - I like both these, they could be added to the Howto granular surf doc page.  The 2nd one is self-explanatory.
For the 1st one, I'm not clear what it is?  Maybe you need to zoom out further on the image?

jtclemm:
The 1st is a screw feeder, grains fall through the top pipe and the helix rotates to convey them. Could use this image instead? Or see if LAMMPS can directly render something comparable. 

screwfeeder2

akohlmey:
> Or see if LAMMPS can directly render something comparable.

Here is a quick attempt with LAMMPS:

image

[feeder.webm](https://github.com/user-attachments/assets/892a74d9-44d7-4e4d-9446-9c5d03be4b22)

doc/src/Howto_granular_surfaces.rst
jtclemm: > I do not understand the 2 parts of this 4-circle figure with convex/concave line segments. I assume the purpose of the figure is to show that the forces on the particle are "as expected"

Yes, that was the intention of the figure. The point they are originating from is arbitrary since it's just indicating the direction of the force. I could have them originate from the closest point of contact on a line, but then it'd just be a bunch of overlapping lines on the left (which maybe is fine). Can scrap it if it's not helpful.

For the alternate way, I can remove that. It was just explaining the bad consequence of picking what someone might naively think would be a better solution. But no point explaining what isn't done.

doc/src/Howto_granular_surfaces.rst
jtclemm: I suppose which fixes do we recommend people use with surfaces? For the local version there's fix move and set? Any other important ones?

akohlmey:
Here is a different viewing angle and particles colored by velocity:

image

doc/src/pair_surf_granular.rst
jtclemm: I added more context on how the two types of pair coefficients determine different contact models. But do we also want to repeat all possible options for contact models? So list normal (hooke, hertz, hertz/material dmt), tangential (linear_nohistory, linear, mindlin, mindlin/rescale, mindlin/force, mindline/rescale/force)... etc where each option then also has a different set of material parameters?

Dan actually has a rewrite of pair granular which is more organized with links to jump to each option. Maybe we can add links to those sections down the road? Otherwise, I could also copy + paste the pair granular documentation here? 

Regarding the two types being spheres/lines/tris, pair surface/granular has no restrictions on what atom types are which. That is handled by the neighbor list which is created by fix fsl. So any restrictions are enforced on that end.

jtclemm:
The animation is certainly nice, the rotating helix is something I couldn't figure out how to render externally. Can that format be used in a doc file?

akohlmey:
> Can that format be used in a doc file?

It won't work in a PDF and the file is too large (>8MB) to bundle it with the sources

akohlmey:
Here is an attempt for using LAMMPS' graphics facilities with the second demo system.
The added command lines for `dump image` are:
```
dump viz all image 1000 myimage-*.png diameter diameter tri index 1 1.0 size 800 600 zoom 3.1 &
   shiny 0.1 fsaa yes view 80 -20 box no 0.0 fix 1 const 2 0.5 center s 0.49883 0.55 0.495547 &
   ssao no 13251 0.3

dump_modify viz pad 9 backcolor black backcolor2 white lights 0.2 0.8 0.6 0.9 fcolor 1 silver &
   color map1 0.282 0.129 0.451 color map2 0.435 0.435 0.556 color map3 0.161 0.686 0.498 &
   color map4 0.741 0.875 0.149 amap min 20 cf 0.0 4 min map1 0.333 map2 0.667 map3 max map4
```
image

[itokawa.webm](https://github.com/user-attachments/assets/29b441a7-81d0-4ead-8091-a29350d7c8f3)

akohlmey:
The one thing I am realizing that we could add to dump image is an option to limit additional graphics elements with a region so you can "cut an STL object open". For the first example this can work already because you have typed tri particles.

jtclemm:
> The one thing I am realizing that we could add to dump image is an option to limit additional graphics elements with a region so you can "cut an STL object open". For the first example this can work already because you have typed tri particles.

That would be a nice little option. Particularly for containers. Does that exist for fix wall/region with geometric primitives?

And nice images/videos! I'm thinking there might be value using your images in the doc files since they are all-lammps. Thoughts @sjplimp? 

Edit: I don't think the asteroid makes for a good LAMMPS example since it's a bit expensive and complicated, but should I add the screw feeder with the dump image commands as an example?

sjplimp:
All the new images look nice, both from @jtclemm and @akohlmey   I like the screw feeder from Joel best, b/c it shows the triangulation of the feeder more clearly.  Maybe that is a setting for the in situ graphics.  For the asteroid, both Joel's and Axel's (the one with colored particles) are good.  I think the "cut STL open" as in Joel's pic can be done simply by using a geometric region to limit the dump image graphics?  Which may be what Joel has done with OVITO?  For the images added to the doc page, we should say what graphics tool produced them.

Adding the screw feeder as a LAMMPS example would be good, so that the Howto page pics could point to it.
It could include the dump image commands from Axel.

Finally, I don't think any of this needs to be perfect, we can just go with what you have.  It is more important to have the PR released and we can add better or newer images later.

sjplimp:
Re: cut the image - what I mean is, doesn't dump_modify region already work with dump image?  I thought it did at least for particles.

doc/src/Howto_granular_surfaces.rst
jtclemm: I forgot a % sign, should be there now.

jtclemm:
> Finally, I don't think any of this needs to be perfect, we can just go with what you have. It is more important to have the PR released and we can add better or newer images later.

Agreed. I'll opt for Axel's current images and include the example with the dump image command. The other Ovito images I generated for the granular paper so those will be used there. As you say, can always improve the dump commands to slice regions later.

I'll work on those updates. In the meantime I also replied to your comments above to fix some issues. 

Edit: added the new examples, please let me know what you think.

akohlmey:
> Re: cut the image - what I mean is, doesn't dump_modify region already work with dump image? I thought it did at least for particles.

for particles, you can already control what is output by defining the dump group.
but that would be different by graphics objects that are added through a fix or compute.
I have to review the code.

To answer the question about fix wall/gran/region. We don't have a fix visualization for that but you can visualize the region directly and region visualizations are a totally different beast. I have to re-read the code about how this would work.

Bottom line: the feeder example should be able to "slice" the surfaces with creative definition of the dump group, but the other example not.

I will investigate. It certainly is not something that belongs in this pull request but is a later add-on. Right now everything seems to work well enough for using dump image as it is. Another later option to explore would be to add an option to compute a gradient map as an additional option for smoother surfaces.

sjplimp:
@jtclemm new feeder example looks good including adding commented-out in situ graphics
@akohlmey per previous email, from my POV this is ready to merge when Joel approves the PR

jtclemm:
> @akohlmey per previous email, from my POV this is ready to merge when Joel approves the PR

I made a few more final clean ups and think the PR is ready to be merged (assuming I didn't break any tests). There's undoubtedly more work that could be done, but this PR is big enough so best to save that for later.

jtclemm (approved):

akohlmey:
> but this PR is big enough so best to save that for later.

There is one annoying detail. This adds the third STL file reader/parser and I would like to consolidate that and use the safer/cleaner text processing tools/classes we have available in LAMMPS.

akohlmey (approved):

jtclemm:
> There is one annoying detail. This adds the third STL file reader/parser and I would like to consolidate that and use the safer/cleaner text processing tools/classes we have available in LAMMPS.

Oof that is annoying, I didn't know that. Thanks for cleaning up! Nice to finally have this off my plate. 

For now...

PR #4215, opened 2024-07-01
Ensure atom map is reset by fix pour and fix deposit

sjplimp:
**Summary**

Fix pour and fix deposit both had a small bug with how they used the atom map.  The resetting of the atom map after insertion was conditional on "success" of the insertions.  It should instead be done regardless since it is cleared before insertion.  The success flag was checked for calling map_init(), and if it is not called the subsequent map_reset() could crash.  Fix pour had the additional issue that the "success" flag depends on the final insertion.  So numerous successful insertions could be treated as unsuccessful.

**Related Issue(s)**

N/A

**Author(s)**

Steve

**Licensing**

**Backward Compatibility**

N/A - previous code would either crash or throw valgrind errors.

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/doc/src/atom_modify.rst
+++ b/doc/src/atom_modify.rst
@@ -71,11 +71,11 @@ all atoms, e.g. in a data or restart file.
    atom IDs are required, due to how neighbor lists are built.
 
 The *map* keyword determines how atoms with specific IDs are found
-when required.  An example are the bond (angle, etc) methods which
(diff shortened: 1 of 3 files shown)


sjplimp:
@akohlmey Please check whether you think my logic on this change is correct.  I saw the error when using atom maps with the new features in PR #4213.

akohlmey (approved):

Issue #251, opened 2016-11-03
fix ave/atom etc should produce standard deviation


andeplane:
When using the ave/atom (and similar averaging commands) it makes sense to be able to produce the standard deviation for the quantity. 

For instance `stddev yes` will add N extra columns with the standard deviations for each N value.

What is your view on this?

rjones-sandia:
I think variances would be super useful given the stat mech context. I wrote a version of fix ave/time to use an online/real time estimate of the variance. I'd would like to add that to lammps proper

Reese Jones
rjones@sandia.gov<mailto:rjones@sandia.gov> 925 294-4744
"What's gone and what's past help should be past grief" - Paulina, The Winter's Tale.

On Nov 3, 2016, at 12:58 PM, Anders Hafreager <notifications@github.com<mailto:notifications@github.com>> wrote:


When using the ave/atom (and similar averaging commands) it makes sense to be able to produce the standard deviation for the quantity.

For instance stddev yes will add N extra columns with the standard deviations for each N value.

What is your view on this?

-

athomps:
I agree. Providing this as an option to fix ave/time would be very beneficial.  I never again want to figure out the correct form of this expression:

varx = (sumxsq - sumx*sumx/n)/n

Aidan



--
      Aidan P. Thompson
      01444 Multiscale Science
      Sandia National Laboratories
      PO Box 5800, MS 1322      Phone: 505-844-9702
      Albuquerque, NM 87185     Fax  : 505-845-7442
      E-mail:athomps@sandia.gov Cell : 505-218-1011

andeplane:
Ok, I can start thinking about implementing this. I remember a technique where you don't get troubles when adding small quantities to a large cumulative sum.

rjones-sandia:
the online/realtime algorithm is s bit more complex. give me a minute and I’ll try to dig up the code and a bit of documentation

Reese (Jones) 925 294-4744 rjones@sandia.gov<mailto:rjones@sandia.gov>






On Nov 3, 2016, at 1:19 PM, Anders Hafreager <notifications@github.com<mailto:notifications@github.com>> wrote:


Ok, I can start thinking about implementing this. I remember a technique where you don't get troubles when adding small quantities to a large cumulative sum.

rjones-sandia:
ok
  I think the algorithm embedded in the attached code (spawned from a now ancient version of fix_ave_time) is basically the one described in :

[...]

PR #4221, opened 2024-07-03
Adding variance computation option to averaging fixes

Bibobu:
**Summary**

Following a personal need for variance of averaged data and since issue #251 did not progress in recent years, this PR proposes an implementation of the Welford algorithm to compute variance of data in a single pass.

The numerical comparison was performed using the temperature of a melted LJ fluid. The values where recomputed using a python code which gave a relative difference of $10^-8$.

**Related Issue(s)**

Related to #251. This only adds the variance computation to ave/time, but the extension to other fixes is possible. It might be a bit tricky though. The algorithm is quite simple but requires intermediates vectors and arrays.

**Author(s)**

Germain Clavier, Unicaen germain.clavier@unicaen.fr

**Licensing**

**Backward Compatibility**

Should not affect backward compatibility.

**Implementation Notes**

The option can be turned on and only changes output files for now. The variances arrays cannot be accessed outside of the fix class.

**Post Submission Checklist**


akohlmey:
I think this feature should be enabled/disabled with a meaningful command line flag, "var" is rather nondescript and could be interpreted as related for variables and not variance
Then the computed data should also be made available though the compute_xxx() functions.

Is this compatible with the different averaging modes? Like running or window?

Any feature in LAMMPS must be documented!

Since we are currently in "stabilization mode" for the next stable release, this PR will not be considered for merging until after the stable release, so there would be ample time to port this to other averaging fixes. For as long as this is an optional feature, memory consumption is not a concern.

Bibobu:
> I think this feature should be enabled/disabled with a meaningful command line flag, "var" is rather nondescript and could be interpreted as related for variables and not variance Then the computed data should also be made available though the compute_xxx() functions.

Hi @akohlmey, I wasn't sure how to format the output so I preferred to leave it as is. A possibility would be to reformat the data the way I formatted the output when the option is turned on. That is, in vector mode, using an array twice the number of column with every variance column put right after the related averaged data. I think that would not require to much reformatting, though care must be taken with the arrays indices. In the case of the scalar mode, this might require switch from a vector of values to an array. I don't know how that could be consistently transferred to the other averaging fixes.

> Is this compatible with the different averaging modes? Like running or window?

As I understand the algorithm, the way the running and window averaging modes work is by accumulating and dividing the data. That is $E(X)=\frac{1}{N}\sum{}E_i(X)$ for $N$ epoch every Nfreq. The variance addition formula is $\sigma(X_1+X_2)=\sigma(X_1)+\sigma(X_2)+2Cov(X_1,X_2)$. So if the autocorrelation between the variables $E_i$ is low, the corresponding variance formula should reduce to $\sigma(X)=\frac{1}{N^2}\sum{}\sigma_i(X)=\frac{N\sigma(X)}{N^2}=\frac{\sigma(X)}{N}$. This is what I implemented using the norm value and the same list array as used in running and window mode. The hypothesis on the autocorrelation should be correct in the general case with sufficiently long epochs (just like in the block average hypothesis). I guess this can be checked by computing the autocorrelation of the averages in post processing.

> Any feature in LAMMPS must be documented!

Whoopsie. I forgot the doc file I modified in the last commit. Thanks for pointing it out.

> Since we are currently in "stabilization mode" for the next stable release, this PR will not be considered for merging until after the stable release, so there would be ample time to port this to other averaging fixes. For as long as this is an optional feature, memory consumption is not a concern.

Sure thing. This was a first step and the stars aligned for me to finish it. It is a first implementation in what I think is the easiest averaging fix. It should be rather straightforward with the propose formatting but the way of organizing the data might be a bit different. That I don't know.

Edit: I took the liberty to correct a typo and change the name of the PR to make it larger and more explicit. I will commit the mentioned doc file during the day and wait for feedback before working more on this.

Bibobu:
> > I think this feature should be enabled/disabled with a meaningful command line flag, "var" is rather nondescript and could be interpreted as related for variables and not variance Then the computed data should also be made available though the compute_xxx() functions.
> 
> Hi @akohlmey, I wasn't sure how to format the output so I preferred to leave it as is. A possibility would be to reformat the data the way I formatted the output when the option is turned on. That is, in vector mode, using an array twice the number of column with every variance column put right after the related averaged data. I think that would not require to much reformatting, though care must be taken with the arrays indices. In the case of the scalar mode, this might require switch from a vector of values to an array. I don't know how that could be consistently transferred to the other averaging fixes.
> 
> > Is this compatible with the different averaging modes? Like running or window?
> 
> As I understand the algorithm, the way the running and window averaging modes work is by accumulating and dividing the data. That is E(X)=1N∑Ei(X) for N epoch every Nfreq. The variance addition formula is σ(X1+X2)=σ(X1)+σ(X2)+2Cov(X1,X2). So if the autocorrelation between the variables Ei is low, the corresponding variance formula should reduce to σ(X)=1N2∑σi(X)=Nσ(X)N2=σ(X)N. This is what I implemented using the norm value and the same list array as used in running and window mode. The hypothesis on the autocorrelation should be correct in the general case with sufficiently long epochs (just like in the block average hypothesis). I guess this can be checked by computing the autocorrelation of the averages in post processing.
> 
> > Any feature in LAMMPS must be documented!
> 
> Whoopsie. I forgot the doc file I modified in the last commit. Thanks for pointing it out.
> 
> > Since we are currently in "stabilization mode" for the next stable release, this PR will not be considered for merging until after the stable release, so there would be ample time to port this to other averaging fixes. For as long as this is an optional feature, memory consumption is not a concern.
> 
> Sure thing. This was a first step and the stars aligned for me to finish it. It is a first implementation in what I think is the easiest averaging fix. It should be rather straightforward with the propose formatting but the way of organizing the data might be a bit different. That I don't know.
> 
> Edit: I took the liberty to correct a typo and change the name of the PR to make it larger and more explicit. I will commit the mentioned doc file during the day and wait for feedback before working more on this.

After reconsidering, I noticed that the average of standard deviation assumption did not hold for correlated values that someone would like the statistical values from, like temperature in NVT (duh...). It is easy to rewrite the computation so that the RUNNING style is performed correctly. However, the WINDOW style would require to store the values used for the computation and to recompute the variance from them every time after the max number of values is reached. This is not impossible as Axel suggested since this is an option, but this might stocking arrays which might become large in some cases like `fix ave/time 1 1000 1000 *some values* window 1000` which would be a $10^6$ values array per value. The computation in itself should not be that long compared to large MD runs. This should be benchmarked afterward.

Concerning the access to the data, an alternative idea I had would be to double the array sizes produced along their last dimension when activating the variance option, and store the values at the end. That means for $N$ values, the value stored at the $N+i$ location would be the variance of the $i_{th}$ value. Same would go in the vector style with a 3d array. This would allow to keep backward compatibility when not activating the option and allow to access the data from newer LAMMPS scripts. I convinced myself that the writing in output files should follow the same pattern for similar reasons.

Bibobu:
Sorry for the delay, I didn't have a lot of time to work on the correct implementation but I've been thinking about it and maybe putting it as an option is not really a proper "LAMMPS" way of doing.

@akohlmey, what is your opinion about putting this in another dedicated fix derived from the ave formalism that would compute the variance of the values with the same option and formalism? i.e. that would look like `fix 1 all variance/time 1 1 1 ...` to compute the variance on the data on its own? That might also make it easier to derived for all average styles.

I also have an implementation idea for the window style, but it could grow large with $N_{repeat}\times{}N_{freq}$ and would require redo the calculation every call once the window size is reached.

srtee:
Hi @Bibobu ,

When I started looking at this PR out of sheer curiosity I did think a `fix variance/time` derived fix would be more appropriate and straightforward. Similarly a `fix stddev/time` fix for that matter -- if we are creating a shortcut for

```
variable myVarsq equal v_myVar*v_myVar
fix a all ave/time ... v_myVar v_myVarsq
variable myVarvariance equal f_a[2]-f_a[1]*f_a[1]
```

(fair enough!) we might as well do one for the square root as well.

Note that the variance when combining _samples of the same variable_ does not involve any covariances or correlations. (You are thinking there of combining _two variables into a new variable_.)

So the procedure is straightforward:

* For `ave one`, use Welford's one-pass procedure per `Nfreq`;
* For `ave running`, after each one-pass calculation, use [Chan's parallel algorithm](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Parallel_algorithm) to update the running variance;
* For `ave window M`, after each one-pass calculation: update the running variance, store the one-pass values, reuse the existing `window` logic to check if there's an old value to remove, and if there is, remove it with a _negatively weighted_ parallel update.

As a last note -- there should be a `bessel yes/no` option for users to choose between applying the Bessel correction (divide by N-1) or not (divide by N).

Bibobu:
> Hi @Bibobu ,
> 
> When I started looking at this PR out of sheer curiosity I did think a `fix variance/time` derived fix would be more appropriate and straightforward. Similarly a `fix stddev/time` fix for that matter -- if we are creating a shortcut for
> 
> ```
> variable myVarsq equal v_myVar*v_myVar
> fix a all ave/time ... v_myVar v_myVarsq
> variable myVarvariance equal f_a[2]-f_a[1]*f_a[1]
> ```
> 
> (fair enough!) we might as well do one for the square root as well.

Thank you for your feedback @srtee. I have since a couple weeks a working version for the scalar computation, working for all styles of data collection. Unfortunately I've been busy with other things (the usual excuse) before implementing the vector one. I'll work on it again and finish it soon.

> Note that the variance when combining _samples of the same variable_ does not involve any covariances or correlations. (You are thinking there of combining _two variables into a new variable_.)

Edit: I realised I skipped to comment that indeed, I was obviously wrong on this and yes, in the case of the same variable, it made no sense. 😅

> So the procedure is straightforward:
> 
>     * For `ave one`, use Welford's one-pass procedure per `Nfreq`;
> 
>     * For `ave running`, after each one-pass calculation, use [Chan's parallel algorithm](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Parallel_algorithm) to update the running variance;
> 
>     * For `ave window M`, after each one-pass calculation: update the running variance, store the one-pass values, reuse the existing `window` logic to check if there's an old value to remove, and if there is, remove it with a _negatively weighted_ parallel update.

With enough tweaking, one can use the Welford's version with `one`, `running` and `window`, only this last case requiring to store the data and redo the computation when is has been done M times.  Yet it might be worth implementing it the way you're suggesting it for memory management..

> As a last note -- there should be a `bessel yes/no` option for users to choose between applying the Bessel correction (divide by N-1) or not (divide by N).

That's also a good idea IMHO.

Bibobu:
I think I managed to remove the changes to `fix_ave_time` in order to create a new `fix_var_time` (name can be changed if the ambiguity with "variance" is considered to important).

The new fix takes scalar values but allows for the computation using the same values as the one gathered by `fix_ave_time`. That is `fix var/time 1 10 1000` computes the variance using the same values as `fix_ave_time 1 10 1000`. It supports `one`, `running` and `window` modes.

The algorithm just needs to be ported to the vector mode.

akohlmey:
@Bibobu are you still planning to complete this? It has not seen any activity for two months.
How does this pull request relate to fix ave/momentum? Is its functionality still useful?

Ahead of the stable release we like to pare down the list of pending pull requests to those that are under active development and ideally keep the list of pending pull requests to one page, so that nothing gets overlooked.

Bibobu:
@akohlmey No. In the discussion related to ave/momentum, I sided with the argument that this version of the code was superfluous.

We can close the PR.

PR #4128, opened 2024-04-08
Draft: Use CMake CUDA/HIP language

rbberger:
**Summary**

This is WIP to work through what is needed to transform both GPU and the Kokkos packages to support CMake's CUDA/HIP language support and avoid the wrappers (`nvcc_wrapper`/`hipcc`). This uses [CMake features](https://cmake.org/cmake/help/latest/prop_tgt/CUDA_FATBIN_COMPILATION.html) that have only become available in CMake 3.27.

**Related Issue(s)**

**Author(s)**

@rbberger 

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

**Implementation Notes**

- ~also fixes a typo in the GPU package CUDA cubin header generation code~ (wasn't a bug, the generated name did have the extra `_`)

**Post Submission Checklist**

- [ ] The feature or features in this pull request is complete
- [ ] Licensing information is complete
- [ ] Corresponding author information is complete
- [ ] The source code follows the LAMMPS formatting guidelines
- [ ] Suitable new documentation files and/or updates to the existing docs are included
- [ ] The added/updated documentation is integrated and tested with the documentation build system
- [ ] The feature has been verified to work with the conventional build system
- [ ] The feature has been verified to work with the CMake based build system
- [ ] Suitable tests have been added to the unittest tree.
- [ ] A package specific README file has been included or updated
- [ ] One or more example input decks are included

**Further Information, Files, and Links**

PR #4233, opened 2024-07-14
Rewrite, modernize and complete cmake install

rbberger:
**Summary**

WIP: this is currently a proof-of-concept to improve the situation for plugin writers. I've explicitly not looked at platform specific issues like Windows builds, but maintained some of the helper functionality that was part of `LAMMPSCXXInterface.cmake`.

* update how LAMMPSConfig.cmake is generated; note there will be more changes necessary to make it complete, e.g. detecting more dependencies
* installs all headers necessary for plugins, including package headers like OPENMP
* make compilation definitions / compile options PUBLIC so they are inherited by plugins
* add `lammps_add_plugin` helper
* rewrote the `examples/plugins/CMakeLists.txt` to take advantage of the update CMake install

**Related Issue(s)**

**Author(s)**

@rbberger 

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**


rbberger:
@akohlmey this one might still take a bit, will need to recreate the windows environment somehow. in the meantime, you should probably cherry-pick 8119d8a7e7a9e69410c16e3f8099674f20857279

akohlmey:
> @akohlmey this one might still take a bit, will need to recreate the windows environment somehow. in the meantime, you should probably cherry-pick [8119d8a](https://github.com/lammps/lammps/commit/8119d8a7e7a9e69410c16e3f8099674f20857279)

Thanks, done.

akohlmey:
@rbberger can we (that is you and I) have a discussion within the next few weeks on how we tackle this and also PR #4128?

I suspect that raising the minimum CMake version to 3.20 is not going to be enough, but on the other hand for some newer CMake versions, the current scripting may not be compatible. So there would have to be some duplication based on CMake versions. But at the same time, we have the process of slowly removing functionality from the GNU make build and the time for a proper from-the-ground-up refactor of CMake would be once GNU make is history. This means, there is a lot of moving targets and we need some kind of strategy to navigate them.

rbberger:
Yes, let's figure out a time via Slack.

rbberger:
@akohlmey I don't have a windows system a the moment, could you check out why the path becomes `\\D:\\a\\lammps\\lammps\\build\\helloplugin.so` ?

akohlmey:
> @akohlmey I don't have a windows system a the moment, could you check out why the path becomes `\\D:\\a\\lammps\\lammps\\build\\helloplugin.so` ?

I will give it a look when I find some time between waiting for merges to pass testing. :wink:

akohlmey:
> @akohlmey I don't have a windows system a the moment, could you check out why the path becomes `\\D:\\a\\lammps\\lammps\\build\\helloplugin.so` ?

@rbberger I think I have figured it out. There is a problem with manipulating windows PATH variables, since they use a semi-colon as a separator (unlike unix with a colon). However, that conflicts with CMake's interpretation of a semi-colon as a separator of list items. We need to require for Windows CMake version 3.20 so we can use the cmake_path() command: https://cmake.org/cmake/help/v3.20/command/cmake_path.html

Will make some tests and update the PR branch accordingly, if they succeed.

akohlmey:
@rbberger I am stumped. It works nicely on my Windows virtual machine with Windows 11 and up-to-date MSVC, but fails on GitHub's Windows Server machine.

unittest/commands/CMakeLists.txt
```diff
@@ -2,16 +2,25 @@
 
 add_executable(test_simple_commands test_simple_commands.cpp)
 
-# tests for the plugin command require the PLUGIN package
-if(PKG_PLUGIN)
+# tests for the plugin command require the PLUGIN package and won't work on windows
+if((NOT (CMAKE_SYSTEM_NAME STREQUAL "Windows")) AND PKG_PLUGIN)
```
rbberger: @akohlmey why did you re-add this? It won't build the plugins in that case.

akohlmey: Not intentionally.

akohlmey: That would explain why it works on my virtual machine and not on GitHub: it uses the old plugins, but those are missing with the GitHub runner.

cmake/CMakeLists.txt
```diff
@@ -171,18 +178,20 @@ if(LAMMPS_CXX11)
 endif()
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF CACHE BOOL "Use compiler extensions")
+target_compile_features(lammps PUBLIC cxx_std_${CMAKE_CXX_STANDARD})
+set_target_properties(lammps PROPERTIES CXX_EXTENSIONS ${CMAKE_CXX_EXTENSIONS})
 # ugly hacks for MSVC which by default always reports an old C++ standard in the __cplusplus macro
 # and prints lots of pointless warnings about "unsafe" functions
 if(MSVC)
   if((CMAKE_CXX_COMPILER_ID STREQUAL "MSVC") OR (CMAKE_CXX_COMPILER_ID STREQUAL "Intel"))
-    add_compile_options(/Zc:__cplusplus)
```
akohlmey, line 178: @rbberger this is a problematic change. These MSVC custom settings are required for "everything" and they need to be applied consistently.

rbberger: Which targets did I miss?

akohlmey: I can't easily tell. But is exactly why the global option is convenient and easier. That way there is no question.

rbberger: I really want to get away from global manipulations. Convenience isn't as selling point, since you pay it in other ways later. But, you own CMake now. So ultimately it's your call. I'm telling you it's a bad idea and it's generally considered bad practice. I don't have time for the expensive back and forth, but I found it amusing to ask ChatGPT about the exact reasons...

akohlmey: I am aware of the bad practice (and some, but not all of the potential problems), but I am also concerned about keeping it easy to have people submit packages with CMake support, and not burden them with too much porting effort. At the moment, we have an advantage there, since I've seen people struggle providing GNU make support to their packages.

BTW: Another bad practice that we are currently employing is the extensive use of globbing. Including/excluding files based on package selection would be much more straightforward, if we didn't. My main concern is that if we build up all these extra requirements to contributors, however this would be following best practices, we will never be able to drop GNU make. ... and the latter currently is a higher priority for me at the moment.

I don't want to have a long-winded argument about principles at the moment either. My master plan is to keep changes small until we have eliminated GNU make support. Then we can do a deeper reorganization and at that point, employing more best practices should be on the agenda.

rbberger: So what are you proposing? Have the new target-based approach (which is useful for CMake clients) and keep the global options as catch all for the local build?

akohlmey: I have not looked at this in sufficient detail, so I cannot make a decision or proposal right now. I have to better understand which problem this PR is addressing first.

akohlmey:
We need to look into ways to have the LAMMPS python module installed as part of `make install`. I am thinking of just bypassing the complex processing with building a wheel that is done by `make install-python` and just copy the files at some suitable location under CMAKE_INSTALL_PREFIX and leave the rest (python dependencies, setting LD_LIBRARY_PATH or adding a file to /etc/ld.so.conf.d) to the user.

PR #4236, opened 2024-07-20
Replace (some) snprintf() in dump classes with fmt::sprintf()

akohlmey:
**Summary**

The code in the Dump*::convert_string() methods is quite complex and hard to read and writing to a buffer with changing offset is a bit risky. This collects the formatted line(s) into a std::string class using fmt::sprintf() (which add type safety thanks to throwing exceptions same as fmt::format()) and then copies that to the designated buffer and expanding the buffer as needed.

A secondary goal is to try silence new warnings from CodeQL: https://github.com/lammps/lammps/security

**Related Issue(s)**

This builds on PR #4232 by @rbberger 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues. Should be completely transparent. Unit tests on dumps pass.

**Implementation Notes**

fmt::sprintf() works a bit like fmt::format() but uses traditional printf() format strings.

**Post Submission Checklist**

- [ ] Make benchmake to compare the speed of "fmt::sprintf()" and "snprint()"


akohlmey:
@sjplimp I've added the line:
```
dump 1 all custom 1 big.dump id type element x y z xu yu zu fx fy fz vx fy fz
dump_modify 1 element X
```
to in.melt
and there is about 18% overhead for the output from the changes in this PR. Atom style dump is similar.

I expected less. Generally for its native processing fmt is much faster, so it is probably the printf() part that is creating overhead.

sjplimp:
Hmm - I don't really want any slowdown at all, since almost all simulations use these dump files.
Since the current code works fine, safety isn't a real concern.  I wouldn't want to change it unless you can make it faster and simpler/safer.

akohlmey:
> I wouldn't want to change it unless you can make it faster and simpler/safer.

At the moment, I suspect that requires a different approach that doesn't use printf() style format strings, but rather you would specify only the field width and the (floating point) precision and then use fmt::format(). We had discussed this before and were divided about whether this is an improvement or not.

 That would require a major rewrite and a different implementation. So I am closing this pull request for now.

PR #4239, opened 2024-07-22
Bug fix for determining when an occasional neighbor list needs rebuilding

### LINKED ISSUE #4231: [BUG] Wrong CNA value(and neighbor list) when LAMMPS is run in a loop using next
**Summary**

If LAMMPS is run in a loop by using ```next``` and ```jump``` and if there is no data writing activity either using the ```restart``` or ```dump``` command in between the loop then it produces the wrong Common Neighbor Analysis(CNA) values. Probably due to corrupted neighbor list.
More details is in this thread at [LAMMPS Discussion forum](https://matsci.org/t/wrong-cna-value-when-lammps-is-run-in-a-loop-using-next/56414/4).

LAMMPS version: LAMMPS (17 Apr 2024 - Development - patch_17Apr2024-219-g6d5aa1fbc6)
I was able to reproduce the issue in:
(1) Ubuntu(20.04.2 LTS) in WSL 
(2) Rocky Linux(9.2) in Stampede3 at Texas Advanced Computing Center(TACC).

**Expected Behavior**

```compute cna/atom``` should be able to detect fcc structure during the entire simulation in the attached LAMMPS input script.

**Actual Behavior**

The output of the CNA value changes abruptly when there is no writing activity in between the ```run``` within the ```loop```.  But if writing frequency is increased to write the data either using ```dump``` or ```restart``` during one loop then the problem goes away. So, just changing the dump frequency alters the simulation results. Log files also show the difference in the neighbor statistics.

**Steps to Reproduce**
Attached is a minimal LAMMPS input script to reproduce the problem. First, run the script as it is then change the ```dump``` frequency from 100 to 10 to see the difference.

#. The problem still appears whether using a single processor or multiple.
#. Using different potentials does not help.
#. Changing the boundary from “ppp” to “sss” to prevent atoms from hopping from one side to other does not help.
#. A dry run without any time integration does not show up the problem.

[script_LAMMPS_Basic.md](https://github.com/user-attachments/files/16200994/script_LAMMPS_Basic.md)

### PR DESCRIPTION
**Pull request by sjplimp**
**Summary**

Occasional neighbor lists (e.g. requested by a compute) were sometimes not re-built when they should be.  In particular if
the compute were triggered on the last step of one run (e.g. due to dump file output) and then perpetual lists were re-built on the same timestep at the beginning of the next run (during the setup).  A subsequent dump file output in the next run could use the outdated neighbor list.

The logic change to ensure this doesn't happen also removed the need for an optional "preflag" in the build_one() method.  This flag was used by several computes and fixes, but has been removed, b/c it should no longer be needed.

**Related Issue(s)**

Fixes #4231

**Author(s)**

Steve

**Licensing**

**Backward Compatibility**

Hopefully no change, but the logic for rebuilding occasional lists was changed.

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/src/neighbor.cpp
+++ b/src/neighbor.cpp
@@ -163,6 +163,9 @@ pairclass(nullptr), pairnames(nullptr), pairmasks(nullptr)
   npair_perpetual = 0;
   plist = nullptr;
 
+  npair_occasional = 0;
+  olist = nullptr;
+
   nrequest = maxrequest = 0;
   requests = nullptr;
   j_sorted = nullptr;
@@ -253,6 +256,7 @@ Neighbor::~Neighbor()
 
   delete[] slist;
   delete[] plist;
+  delete[] olist;
 
   for (int i = 0; i < nrequest; i++)
     if (requests[i]) delete requests[i];
@@ -1008,6 +1012,9 @@ int Neighbor::init_pair()
   for (i = 0; i < nrequest; i++) {
     requests[i]->index_pair = -1;
     flag = lists[i]->pair_method;
+
+    // 22 Jul 2024 NOTE, don't think flag = 0 occurs in current code
+
     if (flag == 0) {
       neigh_pair[i] = nullptr;
       continue;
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

**Comment by sjplimp** [comment]
@stanmoore1 Not sure if this CPU change in the logic used in the Neighbor::build() and build_one() methods, using the new olist vector of occasional neighbor lists, needs to be replicated for Kokkos.

File: src/neighbor.cpp [hunk 1]
```diff
@@ -1054,19 +1061,25 @@ int Neighbor::init_pair()
...
+  npair_perpetual = npair_occasional = nstencil_perpetual = 0;
   plist = new int[nlist];
+  olist = new int[nlist];
+  slist = new int[nstencil];
 
   for (i = 0; i < nlist; i++) {
     if (lists[i]->occasional == 0 && lists[i]->pair_method)
       plist[npair_perpetual++] = i;
-  }
+    if (lists[i]->occasional)
```
jtclemm: I know we determined this will never happen, should this still check `&& list[i]->pair_method` in case that ever becomes a possibility? Or just to mirror the above logic for perpetual lists?

sjplimp: yes, I forgot to add that - just pushed the change - good catch

jtclemm (approved):

PR #4145, opened 2024-04-22
Improve type label support

jrgissing:
**Summary**

add direct type label support for commands/fixes/computes that have 'type' arguments

- [x] create_atoms command
- [x] fix deposit
- [x] fix mol/swap
- [x] fix atom/swap
- [x] group command
- [ ] fix shake/rattle
  - [x] implemented but not committed. type labels could pretty easily interfere with 'a', 'b', etc. keywords
- [x] fix gcmc
- [x] fix widom
- [x] delete_bonds command
- [x] fix bond/create/break
- [x] fix adapt + adapt/fep
  - [x] add option to feed a type label into utils::bounds. returns nlo = nhi = label2type(type_label)
  - [x] I < J requirement relaxed for non-asterisk syntax
- [x] compute rdf
  - [x] could technically break there was a type label called 'cutoff'
- [ ] pair_style tracker
  - [x] not sure if can work, because commas used in lists
- [x] fix charge/regulation
- [ ] variable command (type==H)
  - [x] rely on label2type(kind,label)
- [x] Dump Output Support
  - [x] dump_modify 'types' for xyz
  - [x] 'typelabel' keyword for dump custom
- [ ] compute property/local - will require refactor of communication routines, suggest a separate PR
- [ ] others?

Existing/prior support includes:
- [x] data file/molecule templates
- [x] label2type function
- [x] set command
- [x] coeff commands
- [x] mass command

**Related Issue(s)**

none

**Author(s)**

Jake

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

yes

**Implementation Notes**

new string output (type labels) to dump files not added in a very extensible manner. can refactor if you think more string fields will be added to dump output

~~currently, 'dump_modify types numeric' (only relevant to dump xyz) does not revert a previous 'dump_modify types labels' command~~

**Post Submission Checklist**

- [x] The feature or features in this pull request is complete
- [ ] Licensing information is complete
- [ ] Corresponding author information is complete
- [ ] The source code follows the LAMMPS formatting guidelines
- [ ] Suitable new documentation files and/or updates to the existing docs are included
- [ ] The added/updated documentation is integrated and tested with the documentation build system
- [ ] The feature has been verified to work with the conventional build system
- [ ] The feature has been verified to work with the CMake based build system
[...]

PR #4240, opened 2024-07-24
Add some more typelabel support

akohlmey:
**Summary**

This pull request expands support for typelabels to a few more commonly used features in LAMMPS.

**Related Issue(s)**

 This expands on PR #4145 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

Type label support is added to the following features:
- neigh_modify exclude type
- fix shake/rattle

**Post Submission Checklist**

### DIFF
--- a/doc/src/fix_shake.rst
+++ b/doc/src/fix_shake.rst
@@ -27,9 +27,9 @@ Syntax
 
   .. parsed-literal::
 
-       *b* values = one or more bond types (may use typelabels)
-       *a* values = one or more angle types  (may use typelabels)
-       *t* values = one or more atom types (may use typelabels)
+       *b* values = one or more bond types (may use type labels)
+       *a* values = one or more angle types  (may use type labels)
+       *t* values = one or more atom types (may use type labels)
        *m* value = one or more mass values
 
 * zero or more keyword/value pairs may be appended
@@ -139,12 +139,15 @@ constrained if its type is in the list.
 
 .. versionchanged:: TBD
 
-The types may be given as typelabels for as long as none of the
-typelabels are called *b*, *a*, *t*, or *m*.  In those cases the type
-arguments would be ambiguous and thus support for typelabels will be
-disabled instead.
+The types may be given as type labels *only* if there is no atom, bond,
+or angle type label named *b*, *a*, *t*, or *m* defined in the
+simulation.  In that is the case, type labels cannot be used as
+constraint type index with these two fixes because the type labels would
+be incorrectly treated as a new type of constraint instead.  Thus,
+LAMMPS will print a warning and type label handling is disabled and
+numeric types must be used.
 
-For all constraints, a particular bond is only constrained if both
+For all constraints, a particular bond is only constrained if *both*
 atoms in the bond are in the group specified with the SHAKE fix.
 
 The degrees-of-freedom removed by SHAKE bonds and angles are accounted
--- a/doc/src/fix_shake.rst
+++ b/doc/src/fix_shake.rst
@@ -27,9 +27,9 @@ Syntax
 
   .. parsed-literal::
 
-       *b* values = one or more bond types
-       *a* values = one or more angle types
-       *t* values = one or more atom types
+       *b* values = one or more bond types (may use typelabels)
+       *a* values = one or more angle types  (may use typelabels)
+       *t* values = one or more atom types (may use typelabels)
        *m* value = one or more mass values
 
 * zero or more keyword/value pairs may be appended
@@ -136,6 +136,13 @@ constrained (within a fudge factor of MASSDELTA specified in
 ``src/RIGID/fix_shake.cpp``).  The *a* constraint lists angle types.  If
 both bonds in the angle are constrained then the angle will also be
 constrained if its type is in the list.
+
+.. versionchanged:: TBD
+
+The types may be given as typelabels for as long as none of the
+typelabels are called *b*, *a*, *t*, or *m*.  In those cases the type
+arguments would be ambiguous and thus support for typelabels will be
+disabled instead.
 
 For all constraints, a particular bond is only constrained if both
 atoms in the bond are in the group specified with the SHAKE fix.
--- a/doc/src/fix_shake.rst
+++ b/doc/src/fix_shake.rst
@@ -141,11 +141,11 @@ constrained if its type is in the list.
 
 The types may be given as type labels *only* if there is no atom, bond,
 or angle type label named *b*, *a*, *t*, or *m* defined in the
-simulation.  In that is the case, type labels cannot be used as
-constraint type index with these two fixes because the type labels would
-be incorrectly treated as a new type of constraint instead.  Thus,
-LAMMPS will print a warning and type label handling is disabled and
-numeric types must be used.
+simulation.  If that is the case, type labels cannot be used as
+constraint type index with these two fixes, because the type labels
+would be incorrectly treated as a new type of constraint instead.
+Thus, LAMMPS will print a warning and type label handling is disabled
+and numeric types must be used.
 
 For all constraints, a particular bond is only constrained if *both*
 atoms in the bond are in the group specified with the SHAKE fix.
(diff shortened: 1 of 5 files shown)


akohlmey:
@jrgissing I would like to make an effort to have support for typelabels added to all features that are likely used in Simon's tutorial. While there should be one example showing that numeric types and typelabels can be mixed, all other examples using typelabels (and we should have a mix of numeric types and typelabels since typelabels just add overhead in some cases) should be fully supported.
I just added support for two more features used in my current text example.

akohlmey:
@lammps/core do you have any LAMMPS commands or styles where you use numeric types (for atoms, bonds, angles, etc) that have not yet been converted to support typelabels. See PR #4145 and this one for commands that have been converted.

I think for fix shake this also resulted in a cleaner parsing of the command arguments...

Here is an example that makes used of typelables throughout (and thus requires the changes in this pull request):
[typelabel-example.tar.gz](https://github.com/user-attachments/files/16366647/typelabel-example.tar.gz)

src/RIGID/fix_shake.cpp
```diff
@@ -121,49 +121,40 @@ FixShake::FixShake(LAMMPS *lmp, int narg, char **arg) :
...
-    // break if keyword that is not b,a,t,m
-
-    } else if (isalpha(arg[next][0])) break;
-
-    // read numeric args of b,a,t,m
-
-    else if (mode == 'b') {
-      int i = utils::inumeric(FLERR,arg[next],false,lmp);
+    if ((strcmp(arg[next],"b") == 0) && (next+1 < narg)) {
+      int i = utils::expand_type_int(FLERR,arg[next+1],1,lmp);
```
jrgissing: I have been using enum values (e.g., Atom::ATOM) instead of numbers for the mode argument of expand_type_int

jrgissing:
I purposely did not add support for rattle/shake in the previous PR, because I am not sure how the treat the case of a type label being 'a', 'b', 't' or 'm'.

akohlmey:
> I purposely did not add support for rattle/shake in the previous PR, because I am not sure how the treat the case of a type label being 'a', 'b', 't' or 'm'.

That is why I refactored the parser. The old version would loop over individual "words" and then set "mode" to the character from the previous iteration. This is rather difficult to understand and error prone.

Instead, the parser is now processing the "mode" character **and** the following word in one go and then just increment the argument index twice. Thus the check that was previously triggered is skipped.

Update:
I changed my test example to use "a" for hydrogen and "b" for oxygen and there is no problem with the revised fix shake parser.

src/RIGID/fix_shake.cpp
akohlmey: Thanks for the reminder!
Laziness at work. Of course this should be using the enumerator. Will make the change ASAP.

jrgissing:
> > I purposely did not add support for rattle/shake in the previous PR, because I am not sure how the treat the case of a type label being 'a', 'b', 't' or 'm'.
> 
> That is why I refactored the parser. The old version would loop over individual "words" and then set "mode" to the character from the previous iteration. This is rather difficult to understand and error prone.
> 
> Instead, the parser is now processing the "mode" character **and** the following word in one go and then just increment the argument index twice. Thus the check that was previously triggered is skipped.
> 
> Update: I changed my test example to use "a" for hydrogen and "b" for oxygen and there is no problem with the revised fix shake parser.

I do not think this resolves all possible issues, because more than one (an unknown number of) types can be listed after each keyword

akohlmey:
> I do not think this resolves all possible issues, because more than one (an unknown number of) types can be listed after each keyword

Ok, so that is a problem. How about all typelabels would have to come with a "label:" or "l:" prefix, for example: b 1 2 l:HW-OW"?

jrgissing:
How about we require quotes around all types for a certain 'type kind' when using type labels? This would be in line with the LAMMPS parsing style that quotes can enclose a single argument

akohlmey:
> How about we require quotes around all types for a certain 'type kind' when using type labels? This would be in line with the LAMMPS parsing style that quotes can enclose a single argument

After some deliberating, I am moving to a different strategy. I'm adding a test whether there is a type label conflicting with the 'b', 'a', 't', or 'm' keyword and then disable the checking for a type label (with a suitable warning message). I think the chance of conflict is very low and in case there is a conflict, there is always the option to change the type labels until there is no more conflict. Not perfect, but simpler and more consistent for the common case of no conflicts.

jrgissing:
> > How about we require quotes around all types for a certain 'type kind' when using type labels? This would be in line with the LAMMPS parsing style that quotes can enclose a single argument
> 
> After some deliberating, I am moving to a different strategy. I'm adding a test whether there is a type label conflicting with the 'b', 'a', 't', or 'm' keyword and then disable the checking for a type label (with a suitable warning message). I think the chance of conflict is very low and in case there is a conflict, there is always the option to change the type labels until there is no more conflict. Not perfect, but simpler and more consistent for the common case of no conflicts.

OK that sounds good

src/neighbor.cpp
```diff
@@ -2705,8 +2705,8 @@ void Neighbor::modify_params(int narg, char **arg)
           memory->grow(ex1_type,maxex_type,"neigh:ex1_type");
           memory->grow(ex2_type,maxex_type,"neigh:ex2_type");
         }
-        ex1_type[nex_type] = utils::inumeric(FLERR,arg[iarg+2],false,lmp);
-        ex2_type[nex_type] = utils::inumeric(FLERR,arg[iarg+3],false,lmp);
+        ex1_type[nex_type] = utils::expand_type_int(FLERR, arg[iarg+2], Atom::ATOM, lmp);
+        ex2_type[nex_type] = utils::expand_type_int(FLERR, arg[iarg+3], Atom::ATOM, lmp);
```
jrgissing, line 2709: I would complain about not also supporting collection/type for neigh_modify, but that runs into the same parsing problem I had with pair_style tracker, where type labels are separated by commas

akohlmey: I think those options are quite uncommon and thus not a high priority

akohlmey:
> OK that sounds good

Just pushed the alternate implementation of type handling for fix shake/rattle

doc/src/fix_shake.rst
```diff
@@ -27,9 +27,9 @@ Syntax
 
   .. parsed-literal::
 
-       *b* values = one or more bond types
-       *a* values = one or more angle types
-       *t* values = one or more atom types
+       *b* values = one or more bond types (may use typelabels)
```
jrgissing: I think we have always used two words for 'type label'

doc/src/fix_shake.rst
```diff
@@ -137,6 +137,13 @@ constrained (within a fudge factor of MASSDELTA specified in
 both bonds in the angle are constrained then the angle will also be
 constrained if its type is in the list.
 
+.. versionchanged:: TBD
+
+The types may be given as typelabels for as long as none of the
```
jrgissing: rewording suggestion: "The types may be given as type labels only if there is no atom, bond or angle type label of *b*, *a*, *t*, or *m* defined in the simulation.  In this case, type label support is disabled because the type arguments could be ambiguous."

sjplimp: I suggest this altered 2nd sentence in Jake's suggestion, if it is correct:

If that is the case, type labels cannot be used as constraint values with these 2 fixes because those type labels would mistakenly be treated as a new category of constraint.  Use numeric types instead.

doc/src/fix_shake.rst
akohlmey: Ok. Made the change, also for fix shake docs.

doc/src/fix_shake.rst
akohlmey: Done

akohlmey:
I am marking this as ready for review now since there doesn't seem to be any other common use cases for (numeric) types where we could support type labels.

doc/src/fix_shake.rst
```diff
@@ -137,7 +137,17 @@ constrained (within a fudge factor of MASSDELTA specified in
 both bonds in the angle are constrained then the angle will also be
 constrained if its type is in the list.
 
-For all constraints, a particular bond is only constrained if both
+.. versionchanged:: TBD
+
+The types may be given as type labels *only* if there is no atom, bond,
+or angle type label named *b*, *a*, *t*, or *m* defined in the
+simulation.  In that is the case, type labels cannot be used as
```
jrgissing: In ...-> If...

akohlmey: Fixed. Thanks.

jrgissing (approved):

Issue #1377, opened 2019-03-24
[BUG] _Possible errors in unit conversion factors and/or units documentation_


tadmor:
**Summary**

As part of the extension of the OpenKIM testing framework to support LAMMPS simulator models, I have written a program to convert units between the LAMMPS unit systems.  To conform to LAMMPS, I have used physical constants from NIST (https://physics.nist.gov/cuu/Constants/Table/allascii.txt).  To test my code, I verified that I could reproduce the conversion factors given in update.cpp in the LAMMPS code.  I get agreement for almost all factors, but there are several discrepancies that I would like to bring to your attention.  These may be due to documentation errors on the LAMMPS units page site or bugs in update.cpp.

Here is a summary of the discrepancies I see:

1. Electron system (mvv2e factor). According to the LAMMPS documentation, the velocity units in electron are "Bohr/atomic time units [1.03275e-15 seconds]" (note this is not the standard atomic time units). See:

https://sourceforge.net/p/lammps/mailman/lammps-users/thread/BCA2BDB2-BA03-4280-896F-1E6120EF47B2@caltech.edu/

With this definition, I get a value of mvv2e = 1.0. However, the value in update.cpp is mvv2e = 1.06657236. This value is obtained if the velocity is in Bohr/femtosecond not Bohr/atu_electron. So it appears that either the documentation is wrong or the value in update.ccp is wrong.

2. Electron system (ftm2v factor). Same issue as in item 1. With the velocity in "Bohr/atomic time units 1.03275e-15 seconds]", I obtain ftm2v = 0.968288666437. Whereas the value in update.cpp is ftm2v = 0.937582899. As in 1, the value in update.cpp is obtained if the velocity is taken to be bohr/fs.

3. Electron system (vxmu2f). The units documentation does not provide dynamic viscosity units for the electron system. If I guess that the units are Pascal*second, then I get vxmu2f = 32.9114634037. The value in update.cpp is vxmu2f = 3.39893149e1. As in item 1, this discrepancy is corrected if the velocity is taken to be bohr/fs. (Note in addition to dynamic viscosity, the LAMMPS documentation page also does not provide units for the electron system for torque (I guessed hartree), and density (I guessed amu/bohr^3), and should be provided.)

[...]

PR #4241, opened 2024-07-25
Add new fix to compute force and torque due to electric potential

gsalkuin:
**Summary**

This PR adds a new fix that computes the force/torque on both charges and dipoles given an electric potential. Currently, the fix efield command does not compute the force on dipoles due to the gradient of the electric field. Instead of requiring the components of the electric field and its gradients, this fix only takes a Lepton expression of the form V(x,y,z).

**Related Issue(s)**

N/A

**Author(s)**

Gabriel Alkuino (Syracuse University) - gsalkuino@outlook.com

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

**Post Submission Checklist**


akohlmey:
@gsalkuin Thank you for your submission. Also, many thanks for making the effort to adapt to the LAMMPS programming style, providing (and integrating!!) documentation. You've made my job a lot easier. I've made a few more cosmetic changes on top of that and added a couple of tests for the automated testing. So, from the formal point of view this is ready to go it. I'll leave it to other LAMMPS developers that are more competent in this area of simulations to check the implementation itself.

gsalkuin:
Thank you for the positive feedback and for making the additional changes. This is my first PR and I'm still learning C++, but I hope to make more contributions in the future.

ndtrung81:
@gsalkuin thanks for the contribution! It's important to see that fix efield with a non-uniform e-field misses a force component. I have only a comment in the source code to make the for loop(s) over `nlocal` more efficient, and easier to port to accelerator styles.

gsalkuin:
@ndtrung81 Thanks for the feedback. You are right, I found that the loop is very inefficient when compared to fix efield for the uniform efield case. I think the main cause is lines 195-212 in `fix_epot_lepton.cpp`. Putting this outside the loop and adding a `has_ref` flag similar to `pair_lepton_coul.cpp` made it much faster when not all derivatives are nonzero.

akohlmey:
@gsalkuin it looks like we have a portability issue with Windows and Microsoft Visual C++
Do you have access to a Windows machine with Microsoft Visual Studio so that you can have a look for yourself?
FWIW, the "Community Edition" of Visual Studio is available cost free to individuals, thus it is straightforward to install the compiler (and IDE) and LAMMPS is well set up to be compiled straight from checking out the git repository, since git and CMake are supported directly.

Otherwise, I can take a look (I have a Windows Virtual machine that I can use for that purpose).

doc/src/fix_epot_lepton.rst
```diff
@@ -0,0 +1,140 @@
...
+Syntax
+""""""
+
+.. code-block:: LAMMPS
+
+   fix ID group-ID epot/lepton V ...
+
+* ID, group-ID are documented in :doc:`fix <fix>` command
+* style = *epot/lepton*
+* V = electric potential (energy/charge units)
```
srtee: Suggest electric field * distance units rather than energy/charge. That gets us back to volts in most unit systems other than `lj` (which is for advanced users anyway) and `electron` (where perhaps we could hack in an exception to use Hartrees per electron)

doc/src/fix_epot_lepton.rst
```diff
@@ -0,0 +1,140 @@
...
+.. note::
+
+   If you want the electric potential energy to be included in the
+   total potential energy of the system (the quantity being minimized),
+   you MUST enable the :doc:`fix_modify <fix_modify>` *energy* option for this fix.
+
+.. note::
+
+   The potential :math:`V` can only depend on the spatial coordinates :math:`x, y, z`.
+   It will not be updated with periodic boundary conditions and the :doc:`fix deform <fix_deform>` command.
```
srtee: Are unwrapped coordinates, box sizes, and triclinic angles available to Lepton? In which case users are free to hack together their own desired boundary-aware expressions.

EDIT: from the code, this fix works on _unwrapped_ coordinates. This should be clarified in documentation.

doc/src/fix_epot_lepton.rst
```diff
@@ -0,0 +1,140 @@
...
+
+Description
+"""""""""""
+
+.. versionadded:: TBD
+
+Add an electric potential :math:`V` that applies to a group of charged atoms a force
+:math:`\vec{F} = q \vec{E}` and to dipoles a force :math:`\vec{F} = (\vec{p} \cdot \nabla) \vec{E}`
+and torque :math:`\vec{T} = \vec{p} \times \vec{E}`, where :math:`\vec{E} = - \nabla V`.
+
```
srtee:
```suggestion

The fix also evaluates the additional electrostatic energy due to this potential when `fix/modify energy yes` is applied.
```

src/LEPTON/fix_epot_lepton.cpp
```diff
@@ -0,0 +1,322 @@
...
+  // check validity of lepton expression
+  // remove whitespace and quotes from expression string and then
+  // check if the expression can be parsed without error
+  expr = LeptonUtils::condense(arg[3]);
+  try {
+    auto parsed = Lepton::Parser::parse(LeptonUtils::substitute(expr, lmp));
+    auto phi = parsed.createCompiledExpression();
+  } catch (std::exception &e) {
+    error->all(FLERR, e.what());
+  }
```
srtee: Are there Lepton expressions which can be successfully parsed, but not differentiated (in `x`, `y`, or `z` for forces and 2D combos for torques)?

If yes, should we also try-catch differentiation errors here?

src/LEPTON/fix_epot_lepton.cpp
```diff
@@ -0,0 +1,322 @@
...
+  double **mu = nullptr, **t = nullptr;
+  int *mask = atom->mask;
+  imageint *image = atom->image;
+  int nlocal = atom->nlocal;
+
+  const double qe2f = force->qe2f;
+  const double qqr2e = force->qqr2e;
+
+  auto parsed = Lepton::Parser::parse(LeptonUtils::substitute(expr, lmp)).optimize();
+  Lepton::CompiledExpression phi, dphi_x, dphi_y, dphi_z, dphi_xx, dphi_xy, dphi_xz, dphi_yy, dphi_yz, dphi_zz;
```
srtee: Does `parsed` ever change? Can this be done during `setup` instead of during every step's `post_force`?

src/LEPTON/fix_epot_lepton.cpp
```diff
@@ -0,0 +1,322 @@
...
+    if (mask[i] & groupbit) {
+      if (region && !region->match(x[i][0], x[i][1], x[i][2])) continue;
+      domain->unmap(x[i], image[i], unwrap);
+      
+      // Substitute x, y, z if they exist           
+      for (size_t j = 0; j< cexprs.size(); j++) {
+        if (has_ref[j][0]) (*cexprs[j]).getVariableReference("x") = unwrap[0];
+        if (has_ref[j][1]) (*cexprs[j]).getVariableReference("y") = unwrap[1];
+        if (has_ref[j][2]) (*cexprs[j]).getVariableReference("z") = unwrap[2];
+      }
```
srtee: If I understand correctly this imposes forces calculated from _unwrapped_ coordinates, with lots of possible unintended consequences -- for example a quadratic field will impose a different force on charges depending on their image flag.

I would suggest enabling users to write their Lepton expressions with a wrapped `x` co-ordinate vs an unwrapped `ux` co-ordinate (= x + ix*lx -- which is itself problematic in NPT sims), or whatever equivalent is available in Lepton parsing.

It's messy either way -- then again, inhomogeneous electric fields in periodic simulations ARE inherently messy 😈

gsalkuin:
@akohlmey I think the issue is due to a missing `#include <array>`. I am able to compile it on my Linux and Windows machine now. Hopefully, it passes the checks.

src/LEPTON/fix_epot_lepton.cpp
```diff
@@ -0,0 +1,324 @@
...
+
+  const double qe2f = force->qe2f;
+  const double qqr2e = force->qqr2e;
+
+  auto parsed = Lepton::Parser::parse(LeptonUtils::substitute(expr, lmp)).optimize();
+  Lepton::CompiledExpression phi, dphi_x, dphi_y, dphi_z, dphi_xx, dphi_xy, dphi_xz, dphi_yy, dphi_yz, dphi_zz; 
+
+  dphi_x = parsed.differentiate("x").createCompiledExpression();
+  dphi_y = parsed.differentiate("y").createCompiledExpression();
+  dphi_z = parsed.differentiate("z").createCompiledExpression();
```
srtee: I'm genuinely not sure if maybe modern computers need less hand-holding -- but I feel like these CompiledExpressions should be defined in a header, allocated in an `allocate` function, and deallocated in the fix destructor, instead of being dynamically allocated and destroyed every step.

Also, since there are only ever 10 possible expressions (energy, 3 force, 6 torque), the code might actually be _more_ readable without the vectors and arrays.

src/LEPTON/fix_epot_lepton.cpp
```diff
@@ -0,0 +1,324 @@
...
+    q = atom->q;
+    phi = parsed.createCompiledExpression();
+    cexprs.push_back(&phi);
+  }
+  if (atom->mu_flag) {
+    mu = atom->mu; t = atom->torque;
+    dphi_xx = parsed.differentiate("x").differentiate("x").createCompiledExpression();
+    dphi_xy = parsed.differentiate("x").differentiate("y").createCompiledExpression();
+    dphi_xz = parsed.differentiate("x").differentiate("z").createCompiledExpression();
+    dphi_yy = parsed.differentiate("y").differentiate("y").createCompiledExpression();
```
srtee: Can the previous `parsed.differentiate("x")` (for example) be memoized for reuse? I'm not familiar with Lepton's use costs.

src/LEPTON/fix_epot_lepton.cpp
```diff
@@ -0,0 +1,324 @@
...
+  // check if reference to x, y, z exist
+  const std::array<std::string, 3> variableNames = {"x", "y", "z"}; 
+  std::vector<std::array<bool, 3>> has_ref;
+  for (auto &cexpr : cexprs) {
+    has_ref.push_back({true, true, true});
+    for (size_t i = 0; i < 3; i++) {
+      try { 
+        (*cexpr).getVariableReference(variableNames[i]);
+      }
+      catch (Lepton::Exception &) {
```
srtee: All of this try-catching feels expensive. If we can't move this out of post_force step, I wonder if we can just add `+0*(x+y+z)*(x+y+z)` to the original expression and call it a day. 😈

EDIT: do we also need to try-catch for the double derivatives if dipole torques are calculated?

src/LEPTON/fix_epot_lepton.cpp
```diff
@@ -0,0 +1,324 @@
...
+        t[i][2] += qqr2e * tz;
+        fsum[0] -= qqr2e * (mu[i][0] * ex + mu[i][1] * ey + mu[i][2] * ez);
+
+        fx = -qe2f *
+            (mu[i][0] * dphi_xx.evaluate() + mu[i][1] * dphi_xy.evaluate() +
+             mu[i][2] * dphi_xz.evaluate());
+        fy = -qe2f *
+            (mu[i][0] * dphi_xy.evaluate() + mu[i][1] * dphi_yy.evaluate() +
+             mu[i][2] * dphi_yz.evaluate());
+        fz = -qe2f *
```
srtee: You can save three `evaluate()`s by memoizing the 2D components? Not sure if Lepton does its own memoizing.

srtee (changes requested):

srtee:
Hi @gsalkuin , this is a great addition to the LAMMPS codebase -- even though I've marked my review as needing addressed, this is really well done (as well as including good documentation and tests!)

My main question is around the intended treatment of periodicity -- should the Lepton expression act on wrapped coordinates or unwrapped? There are problems either way -- acting on wrapped coordinates would impose artificial force-mismatches on molecules straddling periodic boundaries. But acting on unwrapped coordinates would result in a non-periodic simulation, where particles with similar box coordinates might experience completely different forces depending on their image flags. I am not familiar with periodic simulations involving inhomogeneous fields, so I am happy to defer to your expertise -- but periodicity should be more clearly documented.

My other main questions were around having lots of RAIIs being built and destroyed inside the `post_force` loop, and similar operations -- but if the performance hit is minimal, then you may consider my comments as premature optimisation and reply accordingly :)

Cheers,
Shern

src/LEPTON/fix_epot_lepton.cpp
```diff
@@ -0,0 +1,324 @@
...
+      // substitute x, y, z if they exist           
+      for (size_t j = 0; j< cexprs.size(); j++) {
+        if (has_ref[j][0]) (*cexprs[j]).getVariableReference("x") = unwrap[0];
+        if (has_ref[j][1]) (*cexprs[j]).getVariableReference("y") = unwrap[1];
+        if (has_ref[j][2]) (*cexprs[j]).getVariableReference("z") = unwrap[2];
+      }
+
+      // charges
+      // force = q E, potential energy = q phi
+      if (atom->q_flag) {
```
srtee: This only needs checking once per `post_force`, so moving it outside the `nlocal` loop could speed things up

src/LEPTON/fix_epot_lepton.cpp
```diff
@@ -0,0 +1,324 @@
...
+          v[3] = fx * unwrap[1];
+          v[4] = fx * unwrap[2];
+          v[5] = fy * unwrap[2];
+          v_tally(i, v);
+        }
+      }
+
+      // dipoles
+      // force = (mu dot D) E, torque = mu cross E, potential energy = - mu dot E
+      if (atom->mu_flag) {
```
srtee: As above -- maybe this can go outside `nlocal`

src/LEPTON/fix_epot_lepton.cpp
```diff
@@ -0,0 +1,324 @@
...
+             mu[i][2] * dphi_zz.evaluate());
+
+        f[i][0] += fx;
+        f[i][1] += fy;
+        f[i][2] += fz;
+
+        fsum[1] += fx;
+        fsum[2] += fy;
+        fsum[3] += fz;
+      }
```
srtee: This fix doesn't do `evtally` for dipoles? It's worth documenting in code comments if there's a reason

srtee (changes requested):

src/LEPTON/fix_epot_lepton.cpp
akohlmey: Yes, because the expression may reference LAMMPS variables, which are evaluated and substituted by `LeptonUtils::substitute()`. So the once per timestep processing is required.  The really expensive part of Lepton expressions is the evaluation (although it is faster than Python or LAMMPS variables).

gsalkuin:
Hi @srtee, 

Thanks for the detailed feedback and suggestions! I will try to answer what I can based on my limited expertise. I mostly use LAMMPS for mesh-free simulations in solid mechanics, so I rarely use PBC. The intended use case for this fix is to simulate electric/magnetic control, which usually has the field localized in a region. But that is an excellent point on the force becoming mismatched or unbounded in traditional MD use cases. I used unwrapped coordinates to avoid possible discontinuities in the potential, but perhaps this is not an issue for LEPTON, and wrapped coordinates will be better. I will defer this to your and other MD experts' judgment. I am also not that familiar with the many box-changing commands, but as long as the expression can be written as a function of "x", "y", "z" I think it should be pretty flexible.

As for the `post_force()` variable-reference checking, I think this is needed due to `LeptonUtils::substitute(expr, lmp)`. If a LAMMPS variable evaluates to zero at a particular time step, LEPTON may remove a reference during `optimize()` which will cause an error. This is indeed a very costly step. A quick fix I have is to define another function to check if there are no or only constant LAMMPS variables. The performance improvement is pretty significant, but it made the original code more bloated. Maybe there is a way this could be added in `lepton_utils.cpp` instead.

I will try to address your other concerns when time permits. Thanks!

src/LEPTON/fix_epot_lepton.cpp
akohlmey: Multiple evaluations of the same term should be avoided by caching the results.

akohlmey:
> so I rarely use PBC. The intended use case for this fix is to simulate electric/magnetic control, which usually has the field localized in a region. But that is an excellent point on the force becoming mismatched or unbounded in traditional MD use cases. I used unwrapped coordinates to avoid possible discontinuities in the potential, but perhaps this is not an issue for LEPTON, and wrapped coordinates will be better.

If you do not have PBC, wrapped and unwrapped coordinates are the same. Recent versions of LAMMPS will drop image flags != 0 for non-periodic dimensions when reading coordinates from data files.

The code should print a warning if there is a periodic dimension, and there should be an explanation in the documentation that using it for periodic boundaries is likely going to cause problems and unphysical behavior for atoms crossing the boundary.

src/LEPTON/fix_epot_lepton.cpp
gsalkuin: This is a good catch. I based my code on `fix_efield.cpp` where the `v_tally` is only done for charges under a constant E-field. Since the dipoles will only experience a force under an in-homogeneous field, I'm not sure if that should be included in the Virial.

akohlmey: Please note that LAMMPS has fix numdiff and fix numdiff/virial to validate forces and virials from potential energy through finite differences.

gsalkuin: Thanks, @akohlmey. I looked at the source code and it looks like they only compute the change in energy wrt atomic displacements. In this case, the force on the dipoles should be added to the virial. However, there would still be a missing rotational energy since the dipoles should have atom style sphere for this fix and `fix efield`.

gsalkuin:
A better approach I found is to use finite difference for the dipole force instead of having Lepton compile analytical 2nd derivatives. This way, Lepton only needs to try/catch $(V, E_x, E_y, E_z)$. I got ~3x speed increase on my single CPU tests, so I will probably go with this implementation.

gsalkuin:
Sorry for the late update; I was busy with travel.

Here are the key changes I made based on the excellent feedback I received:

1. The gradient force is now computed using the central difference method to avoid the expensive parsing of second derivatives. A new `step` keyword was added to control the step size.
2. Added a warning that unwrapped coordinates are used. Also made it clear in the documentation. (I did not add support for parsing wrapped coordinates, but I may consider this in the future.)
3. In the documentation, changed the electric potential units to `electric field * distance`. 
4. I also checked the unit conversions and decided to remove support for `electron`, `micro`, and `nano` units for now. For `micro` and `nano`, I believe the conversion factor `qe2f` is incorrect as pointed out in issue #1377. For `electron`, I think the conversion factor from `dipole * electric field` to `energy` units will be different from `qe2f`.
5. Moved `q_flag` and `mu_flag` checks outside the loop over `nlocal`. However, because $E$ is used by both charges and dipoles, I made separate loops for each of the 3 cases (charge only, dipole only, both) to minimize the number of LEPTON parses.
6. Added dipole force in virial to be consistent with other fixes.

Please let me know if you have any further feedback or questions. Thanks.

stanmoore1 (approved):
When the comments/feedback from other reviewers are addressed then I approve this PR. I don't have any additional comments or concerns.

sjplimp:
Thanks @gsalkuin for this addition.  I have 3 cosmetic suggestions.
(1) The first paragraph of the doc page description for both this command and fix efield should mention the other command as an alternative.  This is to guide users. E.g. fix efield could say you should use this other command if your efield has a gradient b/c that term is missing.  This command could say use fix efield if you have no gradient, or if the gradient effect is small and you want something faster.
(2) Adding an example input script or two in the examples dir, illustrating use of this command, would also be helpful to users.
(3) Naming of the command.  Since this command is converting a potential formula to an efield which is then applied to particles (same final step as fix efield), my preference would be to name this command fix efield/lepton.  So that when users scan the list of fixes, they appear together alphabetically.  If you don't like that, then fix epotential/lepton would be my 2nd choice, since epot is not as clear.

doc/src/Commands_fix.rst
```diff
@@ -69,6 +69,7 @@ OPT.
    * :doc:`eos/cv <fix_eos_cv>`
    * :doc:`eos/table <fix_eos_table>`
    * :doc:`eos/table/rx (k) <fix_eos_table_rx>`
+   * :doc:`epot/lepton <fix_epot_lepton>`
```
sjplimp: This is not in alphabetic order

gsalkuin:
Thanks, @stanmoore1 and @sjplimp!

@sjplimp, I'll work on making the changes you suggested. For (3), I didn't name it efield/lepton to avoid confusion since it doesn't follow the same syntax as efield and efield/tip4p. But if this isn't an issue, I can rename it to efield/lepton.

sjplimp:
@gsalkuin Having the same or similar syntax is more an issue when the new class is derived from a parent class.  In this case it is more a conceptual motivation.  Have all the fixes which apply an efield to charged particles, start with the same name.  Kind of like all the compute temp variants, many of which have different syntax, but all conceptually calculate a temperature.

Let's see if Stan or Axel have feedback on the (3) name change.

akohlmey:
> Let's see if Stan or Axel have feedback on the (3) name change.

I agree with calling it fix efield/lepton. It helps people look in the right places for what they need. All /lepton styles are by their very nature different from others, so there is not a problem with having a different syntax than fix efield.

sjplimp (approved):

akohlmey:
@gsalkuin I have submitted a pull request in your fork with some updates for this branch.

@srtee can you please confirm that all your concerns have been addressed.
Otherwise, I will merged this tomorrow or as soon as my changes are included.

srtee:
@akohlmey nothing major from me, just suggestions to documentation and code style (perhaps minor performance improvements). I hope to finish those within 24 hours from this comment but if they're not in before then that's on me, I'm happy to merge.

@gsalkuin very good job on moving the heavy Lepton machinery out of the inner loops!!

Notes to self:
- clarify role of itinerant polarisation in efield (affects energy, shouldn't affect force or else unphysical)
- make those "x, y, z" static constexpr and give better names?

srtee:
Hi @gsalkuin,

I've put my proposed changes in a PR to your branch, with more detailed comments there. In brief:

* I suggested some code changes that will hopefully make the code (slightly) more maintainable in future
* I moved those `if (atom->q_flag)` and `if (atom->mu_flag)` checks back into the `nlocal` loop -- there's a lot of code duplication when those are moved outside the loop. This was definitely a mistake _on my end_ suggesting moving those checks out, and your original design choice to have those checks inside the loop makes for much more readable code.
* I did suggest a different way of getting and storing the Lepton references -- I feel less confident about this one improving the code maintainability, compared to the other changes I suggested.

I did think about adding things to the documentation but -- it's on users to know the physics and realism of their simulations.

I also checked that the regression test you added does, in fact, fail when I randomly flip one of the force signs in the source code, good job!

To be clear -- if the changes I suggested make the code less sensible to you, rather than more, just let me know in this comment thread and I'm happy to approve the PR as-is. I think we should have this PR merged in pretty soon :)

gsalkuin:
Hi @srtee,

The changes you made are excellent—they make the code much more concise. I’ve merged your PR. Thanks!

srtee (approved):

PR #4243, opened 2024-07-26
Cleaning up and fixing bug in SPH package

### LINKED ISSUE #3831: [BUG] SPH/DPD packages incompatible with velocity remapping
**Summary**

In several mesoscale packages (SPH, DPD-MESO, DPD-SMOOTH, and MACHDYN), an extrapolated velocity (`avec->vest`) is used to calculate pair interactions or implement a modified integration scheme. However, many fixes in these packages neglect the fact that the regular velocity array `avec->v` may be augmented in `comm->forward_comm()` and `comm->borders()` calls if there is velocity remapping and an atom crosses a PBC (e.g. in Lees Edwards boundaries or nonzero `domain->h_rate` arrays). This leads to nonsensical dynamics at the border.

This is demonstrated using the SPH package in the attached input script. Under simple shear, particle dynamics go haywire along the simulation border in the gradient direction. Using a patch drafted in this commit, https://github.com/lammps/lammps/commit/6f66bf61c06c3847d4791849ca377217480c19a8, no such behavior is seen and a clean linear velocity gradient develops.

Thanks to Phil Taylor (SNL) for identifying the problem.

Development branch downloaded on 6/20/2023.

**Expected Behavior**

Using the above packages, velocity remapping should either a) produce correct dynamics at borders or b) trigger an error to prevent users from using a faulty combination of commands

**Actual Behavior**

Estimated/extrapolated velocities do not pick up the velocity remapping term leading to bogus forces at periodic borders.

**Steps to Reproduce**

In the attached input script, funky particle properties (e.g. rho) are seen at the interface after 5000 timesteps. For comparison, one can switch remap v to remap x (or used the linked patch) and the behavior disappears. 

[shear.lmp.txt](https://github.com/lammps/lammps/files/11806541/shear.lmp.txt)

jtclemm:
**Summary**

This PR fixes a bug in the SPH package where velocities are only partially remapped by a deformation fix. Essentially, the SPH integrator uses a velocity estimate which is not updated when an atom crosses a PBC. 

Error messages were also added to several other fixes in the DPD packages or MACHDYN package that similarly use velocity estimates. I suppose this could be replaced with a warning though since I didn't confirm 100% they were affected.

I also cleaned up various aspects of the SPH package and fixed a few minor bugs (wrong style names or invalid `pair->single()` methods).

**Related Issue(s)**

This closes [#3831](https://github.com/lammps/lammps/issues/3831).

**Author(s)**

Joel Clemmer (SNL)

**Licensing**

### DIFF
--- a/doc/src/fix_meso_move.rst
+++ b/doc/src/fix_meso_move.rst
@@ -247,6 +247,11 @@ defined by the :doc:`atom_style sph <atom_style>` command.
 
 All particles in the group must be mesoscopic SPH/SDPD particles.
 
+versionchanged:: TBD
+
+This fix is incompatible with deformation controls that remap velocity,
+for instance the *remap v* option of :doc:`fix deform <fix_deform>`.
+
 Related commands
 """"""""""""""""
 
--- a/doc/src/fix_mvv_dpd.rst
+++ b/doc/src/fix_mvv_dpd.rst
@@ -97,6 +97,11 @@ These fixes are part of the DPD-MESO package. They are only enabled if
 LAMMPS was built with that package. See the :doc:`Build package
 <Build_package>` page for more info.
 
+versionchanged:: TBD
+
+This fix is incompatible with deformation controls that remap velocity,
+for instance the *remap v* option of :doc:`fix deform <fix_deform>`.
+
 Related commands
 """"""""""""""""
 
--- a/doc/src/fix_rigid_meso.rst
+++ b/doc/src/fix_rigid_meso.rst
@@ -353,6 +353,11 @@ defined by the :doc:`atom_style sph <atom_style>` command.
 
 All particles in the group must be mesoscopic SPH/SDPD particles.
 
+versionchanged:: TBD
+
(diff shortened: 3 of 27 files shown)


akohlmey:
@jtclemm I am thinking that we should perhaps mention the now unsupported combination of settings triggering the error checks in the "Restrictions" section of the respective commands. Possibly prefixed with a ".. versionchanged:: TBD" so it will become visible when exactly the behavior was changed. What do you think?

akohlmey:
FWIW, I am also backporting the changes to the maintenance branch.

jtclemm:
> @jtclemm I am thinking that we should perhaps mention the now unsupported combination of settings triggering the error checks in the "Restrictions" section of the respective commands.

Yes, that is a good idea. I will add some notes to the doc files. 

Ideally I would have liked to patch all packages, but I don't want to make substantial changes to key fixes in packages I'm unfamiliar with. Also because this is likely an issue that very few (if any) users might run into...

sjplimp (approved):

PR #4242, opened 2024-07-25
Collected small fixes and updates for LAMMPS and LAMMPS-GUI

akohlmey:
**Summary**

This pull request combines multiple small bugfixes and updates that do not warrant a pull request of their own.

**Related Issue(s)**

Fixes the crash reported here https://github.com/lammpstutorials/lammpstutorials-article/pull/10#issuecomment-2249925159
Fixes #4246 
Includes #4247 
Closes #4249 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

No known issues. All tests pass.

**Implementation Notes**

The following individual changes are included:
- use "is" instead of "==" for comparison to None in Python
- fix crash in OpenMP version of ReaxFF that crashes in hbondchk due to not looping over ghost atoms
- correct error output for molecule files
- make native trajectory reader detect triclinic boxes reliably for various revisions of the trajectory file format
- add entry to open Simon Gravelle's LAMMPS tutorial from the About menu in LAMMPS-GUI
- add option to LAMMPS-GUI to export collected thermo data to YAML file form the chart viewer
- add option to LAMMPS-GUI to recenter snapshot image on COM of selected group
- fix error in pair style born/gauss docs
- fix out-of-bounds memory access in pair style dpd/coul/slater/gpu
- update force style tests for bonded interfactions to also test KOKKOS/OpenMP. Imported from PR #4190 
- update tests for ValueTokenizer
- update library interface tests enabling KOKKOS to handle the case of enabled GPU support (which *must* be used).
- update conversion of string to integer using stroll() (available since c++11) and strod() instead of atoi(), atol(), atoll(), and atof() with checks for partial conversion and overflow. See issue #4246 
- some changes in PyLammps to better handle running with MPI. Raise exception on eval() if not called from MPI rank 0. Some updates for the documentation
- includes KOKKOS related fixes from PR #4247 
- includes KOKKOS GFX1103 GPU support from PR #4248 

**Post Submission Checklist**

- [x] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [x] The source code follows the LAMMPS formatting guidelines
- [x] The feature has been verified to work with the conventional build system
- [x] The feature has been verified to work with the CMake based build system

PR #4250, opened 2024-07-31

akohlmey:
**Summary**

This pull request contains multiple small fixes and updates.

**Related Issue(s)**

Among other changes, this addresses some issues that were caused by PR #4242.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

This includes the following individual changes:
- completes the conversion from atoi() and atof() to std::stoi() and std::stod()
- lock sphinx version to < 7.5 since 7.4.x is working but new version 8.0 clashes with breathe.
- when evaluating variable expressions, round to zero on underflows to avoid creating strings with numbers that will be refused for being out-of-range.

**Post Submission Checklist**

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -887,7 +887,7 @@ endif()
 include(Testing)
 include(CodeCoverage)
 include(CodingStandard)
-find_package(ClangFormat 8.0)
+find_package(ClangFormat 11.0)
 
 if(ClangFormat_FOUND)
   add_custom_target(format-src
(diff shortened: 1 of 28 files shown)


sjplimp:
@akohlmey I can't easily see from the PR diff on variable.cpp.  Does the underflow check (reset to 0.0) need to be done for per-atom or vector variables as well, i.e. element by element ?

akohlmey:
> @akohlmey I can't easily see from the PR diff on variable.cpp. Does the underflow check (reset to 0.0) need to be done for per-atom or vector variables as well, i.e. element by element ?

No, this is for `${}` expansions and similar cases where LAMMPS will read back the string it just wrote. So internally we can have an underflow (in the double variable) since Linux doesn't enable floating point traps for them.

sjplimp (approved):

PR #4252, opened 2024-07-31
Fix reading empty type label string in restart

### LINKED ISSUE #4251: [BUG] Invalid label map when reading restart
**Summary**

When reading a restart file, type labels are not checked for empty strings before adding to the label map. This results in label maps marked as complete if there is only one type, and invalid output from write_data that cannot be read by read_data.

LAMMPS (27 Jun 2024) on Ubuntu

**Expected Behavior**

Label maps read from restart files should not include empty strings, and data files written from the result should not contain label maps with empty strings

**Actual Behavior**

When an empty label is read from a restart file, the empty string is used as a key in the type label map, where there was no such entry in the system that wrote the restart file. If there is only one type, the existence of the key in the type label map incorrectly marks it as a complete map and can be written to data files. This results in a type label entry of just "1 ", which is invalid and rejected when reading the data file.

**Steps to Reproduce**

Input file 1:
```
region mybox block 0 1 0 1 0 1
create_box 1 mybox

pair_style none
mass 1 1

labelmap clear  # Initialise labelmap as empty

write_data out.data nocoeff  # No type labels written
write_restart out.restart
```

Input file 2:
```
read_restart out.restart
write_data out_2.data nocoeff
```

out_2.data will contain an invalid atom type label entry "1 ".

k-harris27:
**Summary**

Bugfix for type labels in label_map.cpp. When reading restart files, type labels are now checked for empty strings before using as keys in the label map. This prevents invalid data file output seen in issue #4251.

**Related Issue(s)**

fixes #4251

**Author(s)**

Kieran Harris, University of Manchester

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

Checks for empty strings before assigning label map entries. Tested using the example input files in issue #4251 .

**Post Submission Checklist**

### DIFF
--- a/src/label_map.cpp
+++ b/src/label_map.cpp
@@ -369,35 +369,35 @@ void LabelMap::read_restart(FILE *fp)
   for (int i = 0; i < natomtypes; i++) {
     charlabel = read_string(fp);
     typelabel[i] = charlabel;
-    typelabel_map[charlabel] = i + 1;
+    if (strlen(charlabel) > 0) typelabel_map[charlabel] = i + 1;
     delete[] charlabel;
   }
 
   for (int i = 0; i < nbondtypes; i++) {
     charlabel = read_string(fp);
     btypelabel[i] = charlabel;
-    btypelabel_map[charlabel] = i + 1;
+    if (strlen(charlabel) > 0) btypelabel_map[charlabel] = i + 1;
     delete[] charlabel;
   }
 
   for (int i = 0; i < nangletypes; i++) {
     charlabel = read_string(fp);
     atypelabel[i] = charlabel;
-    atypelabel_map[charlabel] = i + 1;
+    if (strlen(charlabel) > 0) atypelabel_map[charlabel] = i + 1;
     delete[] charlabel;
   }
 
   for (int i = 0; i < ndihedraltypes; i++) {
     charlabel = read_string(fp);
     dtypelabel[i] = charlabel;
-    dtypelabel_map[charlabel] = i + 1;
+    if (strlen(charlabel) > 0) dtypelabel_map[charlabel] = i + 1;
     delete[] charlabel;
   }
 
   for (int i = 0; i < nimpropertypes; i++) {
     charlabel = read_string(fp);
     itypelabel[i] = charlabel;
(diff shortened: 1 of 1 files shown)


akohlmey:
@k-harris27 thanks for your well done bug report and pull request. This is certainly a required fix.

@jrgissing after making some experiments on this, I am thinking we should make some improvements to the labelmap command and related functions.
- for data files only complete maps are written, but restarts also store incomplete ones. This should be made consistent.
- we could also write incomplete maps, if we have a placeholder for empty assignment, e.g. NULL
- using NULL would also allow to selectively delete labelmap entries: `labelmap atom 2 NULL` would clear only the entry for atom type 2. Right now we only have `labelmap clear`
- there should be more tests for this. I already have a few ideas.
What do you think? I am in favor of also storing incomplete maps in data files.

sjplimp (approved):
Jake should have the final say on this PR

jrgissing:
@k-harris27 thanks for your report

@akohlmey I agree with your suggestions. I recall that we initially shied away from incomplete maps in data files because it was deemed as easy thing for the user to complete the map, while saving extra checks in various places on the LAMMPS side. But, I think many users will only need use of a partial map

akohlmey:
> @akohlmey I agree with your suggestions. I recall that we initially shied away from incomplete maps in data files because it was deemed as easy thing for the user to complete the map.

We have been very paranoid and thorough on the original implementation. So a few adjustments are not a big deal, if they are justified. But for that to work, we need to have cases where they matter.

I think the key idea here is to accept NULL as a special type label that means "deleted". This is in keeping with how LAMMPS uses NULL in other places. But I would not have come up with the idea without this issue and pull request.

I have some prototype code already in a branch at home, i don't expect that there will be too many changes. Important is to also expand the tests in the unittest branch. I will see, if I can complete the branch to a state ready to submit it as a draft pull request and then we can iterate it.

@k-harris27 You have perfect timing! Your issue report and pull request came just at the right time. We're about to make a stable release and @jrgissing and I are working on adjusting a LAMMPS tutorial to demonstrate type labels. It would have been very embarrassing to have this all published soon with such a bug in the code.

jrgissing (approved):

PR #4254, opened 2024-08-01
Fixes and enhancements for type label support

akohlmey:
**Summary**

This updates the labelmap, write_data, and read_data commands to make them behave more consistently, more flexible and lift some restrictions that were originally imposed. In particular:

- assigning a type label of NULL will remove the entry from its label map
- write_data no longer requires complete labelmaps to write them to data files. In particular
   - In the XXX Type Labels sections, it will write a label of NULL if no assignment exists
   - When using `write_data xxx.data types labels` only types with a type label assignment will be written as label; all others as numerical type.
- When reading data files, LAMMPS will accept partial labelmaps and not abort. Thus the behavior for data files is closer to how restart files work.
- The unit test tester has been significantly expanded to test the new or modified features thoroughly.

**Related Issue(s)**

This tries to implement what was discussed on PR 4252
https://github.com/lammps/lammps/pull/4252#issuecomment-2260628245

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

Data files created by write_data differ when there are incomplete label maps.

**Implementation Notes**

Some internal APIs where changed, e.g. to remove no longer used parameters. The majority of changes are in the Atom and LabelMap classes. We probably also need to add (more) tests to the type label related functions in the Variable class and potentially make updates.

**Post Submission Checklist**


sjplimp (approved):
Please don't merge until @jrgissing approves

jrgissing:
@akohlmey is this to be merged before stable release?

akohlmey:
> @akohlmey is this to be merged before stable release?

Yes. I am making an exception to the pre-stable rules for all type label and LAMMPS-GUI stuff that is required to complete our revising Simon's tutorials (and paper). Since the tutorial is explicitly referring to the stable version (for good reasons), we should get everything related to type labels into the best possible state and not postpone changes to be done right after the stable release. 

I have more stuff for LAMMPS-GUI coming and some requires (small) changes to the core code as well.

akohlmey:
@jrgissing do you have any suggestions for additional changes, or see any possible problems with the changes included here?

src/atom.cpp
```diff
@@ -1251,7 +1250,6 @@ void Atom::data_atoms(int n, char *buf, tagint id_offset, tagint mol_offset,
               error->one(FLERR, "Invalid atom type {} in {}: {}", itype, location,
                          utils::trim(buf));
             type[nlocal - 1] = itype;
-            if (labelflag) type[nlocal - 1] = ilabel[itype - 1];
```
jrgissing, line 1254: I am confused about many of the changes in this PR. Simply removing lines like these essentially reverts the type label functionality within a data file, and eliminates one of the main benefits of the framework, which is to automatically sync types between multiple input files when using type labels.

akohlmey: Please provide an example illustrating this.

jrgissing: Please see attached for the example I was thinking of.
[in.zip](https://github.com/user-attachments/files/16517160/in.zip)

akohlmey:
This will require a fresh start.

PR #4255, opened 2024-08-01

### LINKED ISSUE #4253: [Feature Request] mpi4py 4.0.0 was released but LAMMPS requires mpi4py >=2,<=3
**Summary**

LAMMPS internally restricts the mpi4py version to be 2 or 3: 
https://github.com/lammps/lammps/blob/develop/python/lammps/core.py#L372

As mpi4py version 4 was just released it would be great if LAMMPS could add support for mpi4py version 4.

**Detailed Description**

mpi4py version 4.0.0 release https://github.com/mpi4py/mpi4py/releases/tag/4.0.0 

akohlmey:
**Summary**

This pull request combines multiple small changes and fixes that don't warrant a pull request on their own.

**Related Issue(s)**

Implements and thus fixes #4253

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues

**Implementation Notes**

- make sure there the internal dump style is deleted when there is an error in dump_modify. This is to avoid getting re-use of dump-ID WRITE_DUMP  errors when calling write_dump after a previously failed call.
- give access to the git revision info through the library interface
- flag mpi4py version 4.x.x as compatible with the LAMMPS python module for passing MPI communicators
- avoid problems in LAMMPS-GUI when trying to use atom diameters of 0 for images
- open either the development, stable, or release version of the manual from LAMMPS-GUI depending on the git branch of the LAMMPS version linked to it
- change the default of "Reformat on Enter" from yes to no for LAMMPS GUI
- new optional auto-save feature in LAMMPS-GUI that automatically saves the editor before starting a run or exiting (off by default)
- new button in LAMMPS-GUI's snapshot image viewer that toggles between shiny and matted surfaces (default can be set in preferences)
- limit the maximum number of threads used by default in LAMMPS-GUI to 16 to avoid performance issues on machines with very large numbers of threads
- add support and scripts to LAMMPS-GUI for building and distributing it as a flatpak package.
- fix memory leaks, file handle leaks, and uninitialized data access flagged by running valgrind on unittests

**Post Submission Checklist**


src/write_dump.cpp
```diff
@@ -50,15 +50,21 @@ void WriteDump::command(int narg, char **arg)
...
 
+  std::string dump_id = "WRITE_DUMP";
   auto dumpargs = new char *[modindex + 2];
-  dumpargs[0] = (char *) "WRITE_DUMP";                      // dump id
+  dumpargs[0] = (char *) dump_id.c_str();                   // dump id
   dumpargs[1] = arg[0];                                     // group
   dumpargs[2] = arg[1];                                     // dump style
   dumpargs[3] = utils::strdup(std::to_string(dumpfreq));    // dump frequency
 
+  // delete existing dump with same ID to avoid re-use of dump-ID error
```
sjplimp: I don't understand this comment, or possibly the bullet point in the PR summary about this change.  What is a scenario when a dump with the same ID already exists?  Possibly just a library interface issue?  What happens if the user script previoulsy defined a dump with the same ID - we wouldn't want to delete it, correct?

doc/src/Build_development.rst
```diff
@@ -153,7 +153,12 @@ development headers to compile (if those are not found locally a recent
 version of that library will be downloaded and compiled along with
 LAMMPS and the test programs) and will download and compile a specific
 version of the `GoogleTest <https://github.com/google/googletest/>`_ C++
-test framework that is used to implement the tests.
+test framework that is used to implement the tests.  Those unit tests
+may be combined with memory access and leak checking with valgrind
+(see below for how to enabled it).  In that case, running so-called
```
jtclemm: (see below for how to *enable* it)

src/write_dump.cpp
akohlmey: The scenario can happen when you issue a write_dump command twice, but there is an error in the dump_modify part. You have to use the library interface to come across it (it happened to me when testing some feature in LAMMPS-GUI) so that exceptions are caught and thus the application doesn't crash instantly. The annoying part is that you cannot easily resolve this without explicitly using undump.

> What happens if the user script previoulsy defined a dump with the same ID - we wouldn't want to delete it, correct?

That is a tricky issue that applies to a lot of "internal" commands. Usually we derive the internal ID from the parent command/style. This doesn't exist in this special case.

The alternative would be to put a try/catch around the call to dump_modify and delete the dump if the dump_modify fails and then rethrow the error. That makes for rather messy output though.

akohlmey: Here is a small reproducer script:
```python
#!/usr/bin/env python3

from lammps import lammps

lmp = lammps()
lmp.file('in.melt')
try:
  lmp.command('write_dump all image image.jpg type type modify adiam 1 0')
except:
  print("write_dump should fail so this is OK")
try:
  lmp.command('write_dump all image image.jpg type type modify adiam 1 1')
except:
  print("write_dump should not fail so this is not OK")
```
Without the change you get:
```
Dangerous builds not checked
ERROR: Illegal dump_modify command (src/dump_image.cpp:1562)
Last command: write_dump all image image.jpg type type modify adiam 1 0
write_dump should fail so this is OK
ERROR: Reuse of dump ID: WRITE_DUMP (src/output.cpp:748)
Last command: write_dump all image image.jpg type type modify adiam 1 1
write_dump should not fail so this is not OK
Total wall time: 0:00:00
```
But you *should* get:
```
Dangerous builds not checked
ERROR: Illegal dump_modify command (src/dump_image.cpp:1562)
Last command: write_dump all image image.jpg type type modify adiam 1 0
write_dump should fail so this is OK
Total wall time: 0:00:00
```

I'll try again with the try/catch. I had tried this before, but I see now a cleaner alternative.

sjplimp:
@akohlmey Don't see our earlier discussion above, so copying info here:

I asked: What happens if the user script previoulsy defined a dump with the same ID - we wouldn't want to delete it, correct?

You replied: That is a tricky issue that applies to a lot of "internal" commands. Usually we derive the internal ID from the parent command/style. This doesn't exist in this special case.

In the other cases, an earlier input script command by the user, defined an ID which is the same as the ID another LAMMPS command wants to generate internally.  That is treated as an error and LAMMPS stops.  So the solution is for the user to redefine the ID they have control over.

In this case for this PR, the code behavior is to remove the original command (dump in this case) with the matching ID.  Which will happen silently.  I see why it is likely to be OK, b/c the original command was left by an error when accessing via the lib interface.  But what about the rare cases when the user specified the ID (e.g. input script read by lib interface command), and then we whack it.  It seems inconsistent.  Can we at least issue a warning, so the user knows that it happened ?

akohlmey:
> Can we at least issue a warning, so the user knows that it happened ?

There is no more need for that. I reverted to use a try/catch block around the dump_modify command. Previously, that would be a mess because deleted the dump and then issued a *new* error message, but I then realized that I could just rethrow the same exception that I caught and the mess was resolved.

So if the dump-ID already exists, the "dump" command inside the write_dump will error out and there is no problem.
The issue was with faulty "dump_modify" arguments that would trigger an error, but not delete the (temporary) dump so the write_dump command would fail the next time around.  With the current code this will not happen because the error is intercepted,  the dump then deleted before the error processing continue.

sjplimp:
@akohlmey I see - thanks for the explanation

sjplimp (approved):

PR #4262, opened 2024-08-05
Type label support for pair_style commands

jrgissing:
**Summary**

Add type label support for certain pair_style commands that have atom, bond, or angle type arguments

- [x] lj/cut/tip4p/long (KSPACE)
  - [x] corrected*
- [x] lj/long/tip4p/long (KSPACE)
  - [x] corrected*
- [x] tip4p/long (KSPACE)
  - [x] corrected*
- [x] tip4p/long/soft (FEP)
  - [x] corrected*
- [x] lj/cut/tip4p/long/soft (FEP)
  - [x] corrected*
- [x] tip4p_cut (MOLECULE)
  - [x] corrected*
- [x] pair_lj_cut_tip4p_cut (MOLECULE)
  - [x] corrected*
- [ ] eff - skip (essentially an element list, similar to other pair styles)
- [x] pair_coeff for pair_style atm
- [x] pair_coeff for hbond/dreiding/lj
- [x] pair_coeff for hbond/dreiding/morse
- [x] e3b
  - [x] corrected*
- [x] srp
  - [x] corrected*
- [x] pair_write
- [x] add checks for legal atom types
- [ ] Update TIP4P HOWTO with type labels?
- [ ] others? (above found by 'grep "=.\*atom type" pair_\*' in docs/src)

*refactored to not evaluate type labels in pair_style command immediately, but instead when coeff() is called

**Related Issue(s)**

none

~~OPT, GPU, OPENMP packages not checked~~ (should be fine, settings not read in, in these routines)

**Author(s)**

Jake

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**


jrgissing:
I am realizing that pair_style is often called before types or type labels are defined. In the case of tip4p styles, for example, there is no check for whether typeO is a valid atom type. Perhaps these type of checks can be added in a routine that is called later than settings(), such as ~~init_style()~~ coeff()

sjplimp (approved):

akohlmey:
@jrgissing I figured it out. It is much less severe than I suspected. When calling `Pair::coeff()` after a restart the type strings are empty and the corresponding variables for types already set. You can see the error when running:
```
TEST_ARGS=-v ctest -R tip4p -V
```
This will not only show the test results but also echo the commands and their output. 

Changes are in commit 5da737fd11de324e891099c3c3b551c8e22c32a7

jrgissing:
Thank you! I suspect I will need to port that fix over to some of the non-TIP4P styles, even if they are not triggering an unit test failure. I should be able to do that later tonight

PR #4263, opened 2024-08-06
Collected small fixes and updates for LAMMPS and LAMMPS-GUI

### LINKED ISSUE #4271: [BUG] Examples in package "RHEO" do not run
**Summary**

Trying to run any of the example scripts fails with "ERROR: Must have only one fix rheo defined (../pair_rheo.cpp:466)"

Running LAMMPS 27 Jun 2024 Develop branch on Ubuntu 22.04

**Actual Behavior**

Trying to run any of the provided example scripts results in the same error

akohlmey:
**Summary**

This contains mostly updates and improvements for LAMMPS-GUI and associated changes to LAMMPS.

**Related Issue(s)**

Implements the feature suggested here: https://matsci.org/t/lammps-gui-test-version-for-macos-windows-and-linux/49921/66

Fixes #4271 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

The following individual changes to LAMMPS-GUI are included:
- add "Inspect restart file" feature into LAMMPS-GUI that creates three windows after reading the restart into a separate LAMMPS instance with: 1) info command output, 2) the data file written from the restart, 3) a snapshot image (which can be rotated, zoomed and more)
- CMake scripting improvements to building flatpak bundles of LAMMPS-GUI
- Menu items and "wizard" dialogs that help setting up tutorials by downloading required files on demand
- Completer for "extra" arguments to read_data 

The following individual changes to LAMMPS are included:
- fix bug in pair style rheo
- add timestep info (update->dt and update->ntimestep) to "info system"
- implement and document a new "-restart2info" command line flag to inspect restart files
- add an undocumented "noinit" flag to write_dump (same as for write_data) to silence the warning about writing a dump without a system init. This is also applied when LAMMPS is run with -restart2dump.
- improve error handling when parsing the MEAM user parameter file: catch value tokenizer exceptions record line numbers and report filename and line number with the error
- force style unit test for 2NN MEAM potentials
- fix various minor issues detected by @ndtrung81 using his new regression testing script

**Post Submission Checklist**


akohlmey:
@sjplimp after we have the "restart file inspector" feature in LAMMPS-GUI, how about also adding a -restart2info flag to the console version of LAMMPS?

For example a restart written at the end of the peptide example would produce the output:
```

Printed on Mon Aug 05 23:08:28 2024

System information:
Units         = real
Atom style    = full
Atom map      = array
Molecule type = standard
Atoms     =         2004,  types =       14,  style = lj/charmm/coul/long
Bonds     =         1365,  types =       18,  style = harmonic
Angles    =          786,  types =       31,  style = charmm
Dihedrals =          207,  types =       21,  style = charmm
Impropers =           12,  types =        2,  style = harmonic
Special bond factors lj =    0        0        0       
Special bond factors coul =  0        0        0       
Kspace style = none

Dimensions = 3
Orthogonal box = 27.371366 x 27.371367 x 27.371367
Boundaries = p,p p,p p,p
xlo, xhi = 36.840194, 64.21156
ylo, yhi = 41.013691, 68.385058
zlo, zhi = 29.768095, 57.139462

Current timestep number = 300
Current timestep size = 2

Group information:
Group[ 0]:     all              (static)
Group[ 1]:     peptide          (static)

Compute information:
Compute[  0]:  thermo_temp,      style = temp,             group = all
Compute[  1]:  thermo_press,     style = pressure,         group = all
Compute[  2]:  thermo_pe,        style = pe,               group = all

Fix information:

```

sjplimp:
@akohlmey sure, that sounds useful

stanmoore1:
I like this too.

akohlmey:
> I like this too.

@sjplimp @stanmoore1 done.

doc/src/Howto_lammps_gui.rst
```diff
@@ -533,13 +535,32 @@ keyboard, the first of those entries is chosen.
...
+.. versionadded:: 1.6
+
+When LAMMPS-GUI is asked to "Inspect a Restart", it will read the
+restart file into a LAMMPS instance and then open three different
+windows.  The first window is a text viewer with the output of an
+:doc:`info command <info>` with system information stored in the
+restart.  The second window is text viewer containing a data file
+generated with a :doc:`write_data command <write_data>`.  The third
+window is a :ref:`Snapshot Image Viewer <snapshot_viewer>` containing a
+visualization of the system in the restart.
```
jtclemm, line 562: I'm not familiar with how the image viewer works, but I'm just curious, how demanding could this be if you unintentionally inspected a restart file from a very large system on, say, a laptop?

akohlmey: Good point. That could be a problem. I would expect less of a problem from the image, but from converting the restart to a data file and then show that. The entire file is stored in memory. 

It is easy to add a check for the file size and then define some threshold and have a dialog "You are trying to inspect a very large restart, do you really, *really*, **REALLY** want that?" or something like that.

I've never run a LAMMPS simulation big enough where that made a difference, but I've also always been working on machines with plenty of RAM.

Does anybody have an idea of a meaningful cutoff value? 1GB? 25% RAM? 5% RAM?

jtclemm: What if you opened the first window, let the user know how many atoms there were, then presented a button which could be clicked to open windows 2 & 3?

Although, it might get old if you have to constantly click an extra confirmation button for small systems...

akohlmey: By that point you would have already read the entire restart into memory. And the number of atoms is not that reliable a measure, since the storage required can change a lot depending on the atom style: "atomic" has far fewer fields that need to be allocated and filled than, say "full".

jtclemm: Ah, if you have to read the whole restart file to get natoms that defeats the purpose of my suggestion. Could have been useful to quickly access high level details of a large restart file even on an underpowered machine.

I'd vote for 1 GB as a fair warning - if it's too big for notepad maybe a data file shouldn't be opened in text.

akohlmey: I'll make some experiments with the rhodo benchmark. Have it write out a restart file and compare memory used by LAMMPS after reading it with a single processor, size of data file and then see what size of restart file will require a large chunk of RAM. I think 1GB on the restart file might be too large. I think 1GB RAM could be ok assuming that even underpowered devices have around 4GB RAM these days (hell, even smartphones have that).

akohlmey: As expected with atom style atomic the memory consumption is less. So for a restart file of about 250MByte containing 3M atoms, the memory consumption for reading the restart and viewing the data file is a bit under 1GByte RAM. For atom style full with the rhodo benchmark, I need about 1.2GB RAM for 1M atoms and the same size restart file. So the warning limit is now empirically determined to be at restart files with over 250MBytes.

akohlmey: done

akohlmey:
If anybody feels adventurous, I just put some pre-compiled binary packages of the latest version of this PR branch on https://download.lammps.org/testing/  (as version 1.6.7). This has the "restart file inspector" and non-functional prototypes for the "Start tutorial" Wizards. Feedback is highly welcome. Also there is the -restart2info command line flag.
For those using Linux, there is either a flatpak bundle or a .tar.gz. Instructions for flatpak are in the manual: https://docs.lammps.org/latest/Tools.html#pre-compiled-executables, for the .tar.gz just unpack, change to the LAMMPS_GUI directory and execute the commands from there. Windows and macOS work the usual way.

doc/src/read_restart.rst
```diff
@@ -47,6 +47,12 @@ Restart files are saved in binary format to enable exact restarts,
 meaning that the trajectories of a restarted run will precisely match
 those produced by the original run had it continued on.
 
+Some information about a restart file can be gathered directly from the
```
sjplimp, line 50: Don't see that this new command-line option is added to the Run_options.rst doc page ?

doc/src/Run_options.rst
```diff
@@ -503,6 +504,36 @@ e.g. the *nfile* and *fileper* keywords.  See the
 
 ----------
 
+.. _restart2info:
+
+**-restart2info restartfile keyword ...**
```
akohlmey, line 509: @sjplimp how about this?

doc/lammps.1
```diff
@@ -226,6 +226,20 @@ arguments of the "dump" command. See the
...
+Write infomation about the <restart file> previously written by
+.B LAMMPS
+to the screen and immediately exit.  Following <restart file>
+argument, additional keywords for the
+.B LAMMPS
+"info" command may be added to increase the amount of information
+written. By default "system" "group" "fix" "compute" are already
+set. See the
+.B LAMMPS
+manual for details on either of the two commands.
```
sjplimp: I don't understand what "either of the two commands" means ?

doc/lammps.1
```diff
@@ -226,6 +226,20 @@ arguments of the "dump" command. See the
...
 manual for details on either of the two commands.
 .TP
+\fB\-r2info <restart file> <keyword> ...\fR or
+\fB\-restart2info <restart file> <keyword> ...\fR
+Write infomation about the <restart file> previously written by
+.B LAMMPS
+to the screen and immediately exit.  Following <restart file>
+argument, additional keywords for the
+.B LAMMPS
+"info" command may be added to increase the amount of information
```
sjplimp, line 236: This should be a link to the info command?

doc/lammps.1
akohlmey: cut-n-paste error. Should be 'for details on the "info" command.'

doc/lammps.1
akohlmey: ??? You cannot link from an nroff file.

sjplimp: sorry, I thought I was reading this from the Run_options.rst file

doc/src/Run_options.rst
```diff
@@ -503,6 +504,36 @@ e.g. the *nfile* and *fileper* keywords.  See the
...
+
+The specified restartfile name may contain the wild-card character "\*".
+The restartfile name may also contain the wild-card character "%".  The
+meaning of these characters is explained on the :doc:`read_restart
+<read_restart>` documentation.  The use of "%" means that a parallel
+restart file can be read.  Note that a filename such as file.\* may need
+to be enclosed in quotes or the "\*" character prefixed with a backslash
+("\") to avoid shell expansion of the "\*" character.
+
+The :doc:`info command <info>` is already followed by the most useful
```
sjplimp: I think I know what you mean by this and the 2-line equivalent above, but it is somewhat confusing.  Maybe something like this:

Optional keywords following the restart filename can be specified to output additional details about the information contained in the restart file.   The list of allowed keywords are those which the info command uses.  Note that the info keywords group, compute, fix are automatically included.  Additional keywords like coeffs or comm can be specified.

akohlmey: rewritten

doc/lammps.1
```diff
@@ -226,6 +226,20 @@ arguments of the "dump" command. See the
 .B LAMMPS
 manual for details on either of the two commands.
 .TP
+\fB\-r2info <restart file> <keyword> ...\fR or
+\fB\-restart2info <restart file> <keyword> ...\fR
+Write infomation about the <restart file> previously written by
```
jtclemm: typo: information

akohlmey: @jtclemm I am amazed by your ability to spot these kinds of typos. I noticed I make them regularly and am debating whether it is my "fast reading of words by shape rather than by letter" that makes me miss them or that the now almost 30 year old [IBM Model-M keyboard](https://en.wikipedia.org/wiki/Model_M_keyboard) that I am typing this on is finally starting to wear out. But then again, the one in the office is five years older and I have been using it since I was a graduate student .

jtclemm (approved):

doc/lammps.1
jtclemm: Unfortunately any ability I may have to spot typos does not translate to not making typos myself ;)

doc/src/Run_options.rst
```diff
@@ -503,6 +504,37 @@ e.g. the *nfile* and *fileper* keywords.  See the
...
+
+The specified restartfile name may contain the wild-card character "\*".
+The restartfile name may also contain the wild-card character "%".  The
+meaning of these characters is explained on the :doc:`read_restart
+<read_restart>` documentation.  The use of "%" means that a parallel
+restart file can be read.  Note that a filename such as file.\* may need
+to be enclosed in quotes or the "\*" character prefixed with a backslash
+("\") to avoid shell expansion of the "\*" character.
+
+Optional keywords may follow the restartfile.  Those would have to be
```
sjplimp: These must be valid keywords ...

Also on the info doc page, the args are "computes" and "fixes" (plural) which is not what it says here or up above in the code-block example.

Also, the info doc page section on Syntax should clarify that all the keywords except "out" and "styles" do not take any values.  It was a bit confusing to me to read.  I think it should read like most (if not all) of the other doc pages, e.g. the "keywords" section of the velocity command:

args = one or more keywords may be appended
keyword = out, all, system, ...
.. parsed-literal ...
only the out and styles keywords require values, the others are just keywords
out value = screen, log, append filename ...
styles value = all, angle, ...

sjplimp: I just realized the info styles keyword can have multiple values, so that should be clarified in the Syntax section as well.  At least one of the Examples should illustrate that.

sjplimp: Also, this -restart2info section should clarify if everything after the initial "restartfile" arg can be legitimate syntax for the "info" command.  That is more than just a list of keywords, b/c the info command allows more than that ?  Does the code simply terminate that info syntax with a command-line arg with a leading dash?

akohlmey: > Does the code simply terminate that info syntax with a command-line arg with a leading dash?

The argument passing code was just copied from -restart2data and -restart2dump, so it works the same way. I don't see any remarks about leading dashes in their docs.

akohlmey: > everything after the initial "restartfile" arg can be legitimate syntax for the "info" command. That is more than just a list of keywords, b/c the info command allows more than that ?

Docs currently say: "Optional keywords may follow the restartfile argument.  Those would
have to be valid keywords for the :doc:`info command <info>`." that says, *any* valid info command argument can be used. This is the same as for the other -restart2xxx only than some keywords are already added to info. This is not a big deal since the same keyword may be used multiple times without bad effects and you can use "all" without causing an error.

If we need changes beyond that, we will need them for the other -restart2XXX commands as well.

akohlmey: @sjplimp Please note that the info command was originally conceived for interactive use in lammps-shell, so it does argument parsing in a way that is convenient for interactive use and therefore differs from other parts of LAMMPS:
- keywords may be abbreviated up to the characters that makes them unique (so fix and fixes is valid)
- unknown keywords or keywords with unknown arguments are ignored
- keywords that select different info sections are cumulative. You can turn sections on, but not off. "all" just turns all on and then keywords like "fixes" or "computes" are already active

I am surprised that this is causing such a big fuss now after it has been around for quite some time.

doc/src/Run_options.rst
```diff
@@ -503,6 +504,37 @@ e.g. the *nfile* and *fileper* keywords.  See the
...
+.. versionadded:: TBD
+
+Write out some info about the restart file and and immediately exit.
+This is the same operation as if the following 2-line input script were
+run:
+
+.. code-block:: LAMMPS
+
+   read_restart restartfile
+   info system group compute fix
```
sjplimp: computes fixes

doc/src/Run_options.rst
```diff
@@ -503,6 +504,37 @@ e.g. the *nfile* and *fileper* keywords.  See the
...
+The specified restartfile name may contain the wild-card character "\*".
+The restartfile name may also contain the wild-card character "%".  The
+meaning of these characters is explained on the :doc:`read_restart
+<read_restart>` documentation.  The use of "%" means that a parallel
+restart file can be read.  Note that a filename such as file.\* may need
+to be enclosed in quotes or the "\*" character prefixed with a backslash
+("\") to avoid shell expansion of the "\*" character.
+
+Optional keywords may follow the restartfile argument.  Those would
+have to be valid keywords for the :doc:`info command <info>`.  The most
```
sjplimp: These must be valid keywords ...

sjplimp (approved):

doc/src/Run_options.rst
sjplimp: For your first 2 bullets, the info doc page does not explain that (it should).
For the -restart2info doc page, you shouldn't use "fix compute" in one place and "fixes computes" in another.
The Syntax section of the info doc page should be formatted like every other LAMMPS command doc page - that is all I am saying.  You can make the changes in this PR or another PR or I can do it.

Issue #4274, opened 2024-08-13
LAMMPS PDF Documenation review


rbberger:
Current version: https://docs.lammps.org/Manual.pdf

Pages reviewed:
- [x] 123 - 399
- [x] 400 - 672
- [x] 673 - 948
- [x] 949 - 1220
- [x] 1221 - 1497
- [x] 1498 - 1770
- [ ] 1771 - 2042
- [ ] 2043 -  2314
- [x] 2315 - 2589
- [x] 2590 - 2838

sjplimp:
Here are my comments on 2590-2838

(a) 2678 - pair style tracker - do not see it in alphabetic list of styles, p 127
(b) 2837 - dump styles chapter is empty
(c) 2867 - Biblio chapter has a single entry
(d) 2869 and 2871 - Python and Fortran pages are near empty ?
(e) 2885, 2886 - index pages have 2-column overlaps

akohlmey:
> Here are my comments on 2590-2838
> 
> (a) 2678 - pair style tracker - do not see it in alphabetic list of styles, p 127
It is at the top where the other "non-standard" pair styles like hybrid, list, or kim are listed.

>  (b) 2837 - dump styles chapter is empty

Should be fixed with the latest changes in #4275 

> (c) 2867 - Biblio chapter has a single entry

That would be the *second* bibliograpy. This is generated when some file uses the LaTeX \cite like mechanism. Since we do not use it and use references and put the citations at the bottom of each .rst file, the (first) bibliography is compiled manually (it needs updating...)

> (d) 2869 and 2871 - Python and Fortran pages are near empty ?

There is only one Fortran module and two Python modules properly referenced in the documentation.
Changed the sphinx configuration to skip creating those.

> (e) 2885, 2886 - index pages have 2-column overlaps

This is a tricky one. Some indexed lines are very long. I've tried to shorten some by changing the sphinx configuration. For example for all C++ code the "LAMMPS_NS::" prefix is now dropped. That has the benefit that classes and functions are sorted into sections by their name and don't all show up under "L".
I've also included some LaTeX hack that will configure the index to be single column.

CI run just finished. The resulting PDF is here: https://ci.lammps.org/job/dev/job/pull_requests/job/build_docs/10336/artifact/lammps/doc/Manual.pdf

rbberger:
@akohlmey I've noticed that the CI generated PDF is a bit different to the one produced locally with newer Sphinx. Should we use a newer container just for the docs? Example:

[...]

PR #4275, opened 2024-08-13
Collected fixes and changes for LAMMPS and LAMMPS-GUI

akohlmey:
**Summary**

This collects small fixes, many from automated code analysis and compiler warnings.
Also, this contains the combined results of the documentation review #4274 

**Related Issue(s)**

Fixes #4274 
Closes #4281 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

No known issues.

**Implementation Notes**

The following individual changes  for LAMMPS are included:
- fix multi-thread bug when creating images with SSAO
- initialize class member pointers to null in constructor
- better error checking and error messages in `compute mliap`
- modernize checking for compatible fixes in pair styles lubricate, lubricate/poly, lubricateU, lubricateU/poly, brownian, brownian/poly
- modernize access to fixes in fix electrode/conp
- documentation review to complete issue #4274 
- update sphinx configuration settings to address issues mentioned in #4274
- changes from PR #4276 
- changes from PR #4277 
- changes from PR #4278 
- changes from PR #4279
- changes from PR #4283 
- changes from PR #4285 
- update Bibiliography.rst

The following individual changes for LAMMPS-GUI are included:
- skip adding data that deviates too much (5 sigma) from the previous 20 data points for cleaner graphs
- add option to add a smoothed plot to the chart using window averaging with a polynomial (Savitzy-Golay filter)
- update graph in Charts window only after a given minimum delay (default 500 milliseconds)

**Post Submission Checklist**

- [ ] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [x] The source code follows the LAMMPS formatting guidelines
- [ ] Suitable new documentation files and/or updates to the existing docs are included
- [ ] The added/updated documentation is integrated and tested with the documentation build system
- [x] The feature has been verified to work with the conventional build system
- [x] The feature has been verified to work with the CMake based build system
- [ ] Suitable tests have been added to the unittest tree.

PR #4277, opened 2024-08-14
Documentation updates/fixes

rbberger:
**Summary**

**Related Issue(s)**

#4274 

**Author(s)**
@rbberger 

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/doc/src/Commands_compute.rst
+++ b/doc/src/Commands_compute.rst
@@ -1,19 +1,21 @@
-.. table_from_list::
-   :columns: 3
+.. only:: html
 
-   * :doc:`General commands <Commands_all>`
-   * :doc:`Fix styles <Commands_fix>`
-   * :doc:`Compute styles <Commands_compute>`
-   * :doc:`Pair styles <Commands_pair>`
-   * :ref:`Bond styles <bond>`
-   * :ref:`Angle styles <angle>`
-   * :ref:`Dihedral styles <dihedral>`
-   * :ref:`Improper styles <improper>`
-   * :doc:`KSpace styles <Commands_kspace>`
-   * :doc:`Dump styles <Commands_dump>`
+   .. table_from_list::
+      :columns: 3
+   
+      * :doc:`General commands <Commands_all>`
+      * :doc:`Fix styles <Commands_fix>`
+      * :doc:`Compute styles <Commands_compute>`
+      * :doc:`Pair styles <Commands_pair>`
+      * :ref:`Bond styles <bond>`
+      * :ref:`Angle styles <angle>`
+      * :ref:`Dihedral styles <dihedral>`
+      * :ref:`Improper styles <improper>`
+      * :doc:`KSpace styles <Commands_kspace>`
+      * :doc:`Dump styles <Commands_dump>`
 
-Compute commands
-================
+Compute styles
+==============
 
 An alphabetic list of all LAMMPS :doc:`compute <compute>` commands.
 Some styles have accelerated versions.  This is indicated by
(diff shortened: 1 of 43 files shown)


doc/src/Commands_compute.rst
```diff
@@ -1,27 +1,13 @@
-.. table_from_list::
```
akohlmey, line 1: FYI these repeated lists were a special request by @sjplimp.
You could prefix all but the first with `.. only:: html` since this was for the HTML version only.

rbberger: I know, but they're beyond obsolete and duplicating the exact same navigation that is to the very left in the HTML version. Unless @sjplimp insists on keeping them, I would remove them.

sjplimp: I'm not sure what this is referring to, and if it is about whether or not it should be in the HTML version or PDF file or both ?

rbberger: lammps-docs-discussion

rbberger: In the HTML version it is redundant. In the PDF version it is more than 10 times the same table with no real purpose. So option A (as Axel suggests) is only show it in the HTML version, Option B (my preference) is get rid of the redundancy completely.

sjplimp:
I see what you mean now - thanks for the explanation.  For the PDF, you can remove all those small tables, as I don't think they are that useful in a PDF file.  For the HTML, I'd like to keep all of them.  The reason is I think the initial page (with General Commands) is likely the most-visited page in the LAMMPS manual (it's linked directly from the home page).  My reason for asking for the small table to be added to each sub-table page was to make it as simple-as-possible for users to find the command(s) they want.  When I go to that page I often flip back and forth between the different sub-tables (pair, fix, compute styles, etc) using the small table of links to look for 2 or 3 related commands. I think it's simpler to use as a compact table of all the command categories. I typically only use the sidebar to skip between top-level chapters..

rbberger:
@akohlmey finished my review of my part. feel free to discard 84c102fe369eadb7284f0b9bee27035bb4dc9db5, it's just a proposal and showcases how customizations are done.

akohlmey:
> @akohlmey finished my review of my part. feel free to discard [84c102f](https://github.com/lammps/lammps/commit/84c102fe369eadb7284f0b9bee27035bb4dc9db5), it's just a proposal and showcases how customizations are done.

@rbberger had to back it out because it made the PDF look strange:  https://ci.lammps.org/job/dev/job/pull_requests/job/build_docs/10376/artifact/lammps/doc/Manual.pdf
There must be some unwanted interaction between different LaTeX "hacks".

Perhaps you can try again after #4275 is merged and we have a new baseline?

rbberger:
> Perhaps you can try again after #4275 is merged and we have a new baseline?

@akohlmey looks like if we just avoid using `afterpage` and just do the regular commands right after the maketitle command it works as intended (the second page gets black too, which would be the backside of the first page).

akohlmey:
> > Perhaps you can try again after #4275 is merged and we have a new baseline?
> 
> @akohlmey looks like if we just avoid using `afterpage` and just do the regular commands right after the maketitle command it works as intended (the second page gets black too, which would be the backside of the first page).

Perhaps this could be something to look into at the off-site meeting in ABQ? I have a few more ideas how we might improve the design, e.g. by using some more of the admonitions for the "Restrictions" "See also" and "Defaults" sections at the end of most pages. Aidan wanted to do some collective hacking and when many of us are in the same room it will be easier to collect feedback on what is an improvement and what not.

So far, I am already extremely happy with what we got from the last week. Many little pieces but altogether quite an improvement. Just finished tweaking some more of the tables with explicit column widths. There is much, much less now in the PDF version that is hard to read.

sjplimp:
@rbberger I changed my mind on the multiple HTML mini-tables.  I now think it is fine to remove all of them except the top-level one (above the General Commands table).  Navigating between the tables via the sidebar list or the top-right Commands link works well.

PR #4299, opened 2024-08-28
Allow the definition of point in region plane using variables

evoyiatzis:
**Summary**

Allow the definition of point in region plane using variables 

**Related Issue(s)**

No related issue

**Author(s)**

E. Voyiatzis at NovaMechanics Ltd (email evoyiatzis@gmail.com)

**Licensing**

**Backward Compatibility**

**Implementation Notes**

The modifications in the code at region_plane.h and region_plane.cpp follow the approach in region_sphere.cpp and region_sphere.h A xstyle/ystyle/zstyle int variable stores if the point is defined as constant or via variable and the xstr/ystr/zstr stores the name of the variable if required. The init method calls variable_check to check if the definition and the numerical value of the point is sound. The deconstructor cleans the memory from the xstr/ystr/zstr variables. 

**Post Submission Checklist**

A set of test files is included in the attached zip file

[equilibrated_germanene_sheet_50.0.zip](https://github.com/user-attachments/files/16780858/equilibrated_germanene_sheet_50.0.zip)


doc/src/region.rst
```diff
@@ -34,10 +34,11 @@ Syntax
        *ellipsoid* args = x y z a b c
          x,y,z = center of ellipsoid (distance units)
          a,b,c = half the length of the principal axes of the ellipsoid (distance units)
-           x,y,z,a,b and c can be a variable (see below)
+         x,y,z,a,b and c can be a variable (see below)
        *plane* args = px py pz nx ny nz
          px,py,pz = point on the plane (distance units)
          nx,ny,nz = direction normal to plane (distance units)
+         px,py,pz can be a variable
```
akohlmey:
```suggestion
           px,py,pz can be a variable
```
The list is more readable when the "can be a variable" statements are indented.

doc/src/region.rst
```diff
@@ -203,12 +204,13 @@ and with radius as its radius.
 
 The *radius* value for styles *sphere* and *cylinder*, and the
 parameters a,b,c for style *ellipsoid*, can each be specified as an
-equal-style :doc:`variable <variable>`.  Likewise, for style *sphere*
```
akohlmey, line 206: For better readability for the documentation source, there should be two blanks after a full stop. We usually add them when revising documentation. There is a lot of legacy text without. Removing those extra blanks just creates extra work for the folks maintaining the manual.

doc/src/region.rst
evoyiatzis: @akohlmey  thanks for the explanation !

doc/src/region.rst
```diff
@@ -18,13 +18,13 @@ Syntax
        *delete* = no args
        *block* args = xlo xhi ylo yhi zlo zhi
          xlo,xhi,ylo,yhi,zlo,zhi = bounds of block in all dimensions (distance units)
-         xlo,xhi,ylo,yhi,zlo,zhi can be a variable
+           xlo,xhi,ylo,yhi,zlo,zhi can be a variable
```
akohlmey:
```suggestion
           xlo,xhi,ylo,yhi,zlo,zhi can be a variable (see below)
```

doc/src/region.rst
```diff
@@ -38,6 +38,7 @@ Syntax
        *plane* args = px py pz nx ny nz
          px,py,pz = point on the plane (distance units)
          nx,ny,nz = direction normal to plane (distance units)
+           px,py,pz can be a variable
```
akohlmey:
```suggestion
           px,py,pz can be a variable (see below)
```

jmgoff (approved):

Issue #4194, opened 2024-06-12
[Feature Request] Allow ReaxFF CPU version to work for >3 million atoms per MPI rank


hagertnl:
**Summary**

When running with > 3 million atoms per MPI rank in the LAMMPS ReaxFF benchmark, strange memory access faults occur where Kokkos reports that a memory allocation error occured while trying to allocate sizes on the order of 1e10 GB, which is a completely bogus number for a problem of this size.

Here is an example output from the Kokkos Serial backend with 7.25 million atoms:
```
Exception: Kokkos failed to allocate memory for label "qeq/kk:jlist".  Allocation using MemorySpace named "Host" failed with the following error:  Allocation of size 1.718e+10 G failed, likely due to insufficient memory.  (The allocation mechanism was standard malloc().)
```

We have identified at least 1 source of integer overflow in https://github.com/lammps/lammps/blob/7ef9a93a75e873484d9aad8119e91c461f130d1c/src/KOKKOS/fix_qeq_reaxff_kokkos.cpp#L307 . We suspect there are more beyond that, but we have not yet worked around this first source of overflow.

A feasible way to reproduce this is in a 1-rank CPU-only simulation using Kokkos Serial backend with > 3 million atoms on a node with at least 256 GB available memory. `run 1` should be sufficient to reproduce the memory fault, but we do `run 100`.

This is relevant for running the ReaxFF benchmark on high-memory (>64GB per GPU) GPU series. 64 GB GPUs max at around 2.1 million atoms.

**LAMMPS Version and Platform**

Latest `develop` branch with the Kokkos+Serial backend using PrgEnv-amd on Frontier.

**Expected Behavior**

We expect the simulation to consume around 160 GB of memory per rank, for a 7.25-million atom system.

**Actual Behavior**

Memory allocations of greater than 1e10 GB are made, causing a crash.

**Steps to Reproduce**

Replicate the LAMMPS-provided ReaxFF 58-atom benchmark on a single MPI rank by dimensions of 50x50x50 to form a 7.25mil-atom system, and run for 1 or more steps. The memory access fault should happen at the first step. Here is the stack trace from a run that has `-ftrapv` set in amdclang, which aborts when an integer overflow is detected:
```
(gdb) bt
[...]

PR #4301, opened 2024-08-29
Gracefully error out if integer overflow in ReaxFF or QEq

stanmoore1:
**Summary**

Gracefully error out if integer overflow in ReaxFF or QEq

**Related Issue(s)**

https://github.com/lammps/lammps/issues/4194

**Author(s)**

Stan Moore (SNL), in discussion with Steve Plimpton

**Licensing**

**Backward Compatibility**

Yes

### DIFF
--- a/src/QEQ/fix_qeq.cpp
+++ b/src/QEQ/fix_qeq.cpp
@@ -237,8 +237,9 @@ void FixQEq::reallocate_storage()
 
 void FixQEq::allocate_matrix()
 {
-  int i,ii,inum,m;
+  int i,ii,inum;
   int *ilist, *numneigh;
+  bigint m;
 
   int mincap;
   double safezone;
@@ -261,7 +262,10 @@ void FixQEq::allocate_matrix()
     i = ilist[ii];
     m += numneigh[i];
   }
-  m_cap = MAX((int)(m * safezone), mincap * MIN_NBRS);
+  bigint m_cap_big = MAX((int)(m * safezone), mincap * MIN_NBRS);
+  if (m_cap_big > MAXSMALLINT)
+    error->one(FLERR,"Number of neighbors too large in fix qeq/reaxff");
+  m_cap = m_cap_big;
 
   H.n = n_cap;
   H.m = m_cap;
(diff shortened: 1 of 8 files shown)


stanmoore1:
@akohlmey I'm done working on this, it should be ready for review

src/QEQ/fix_qeq.cpp
```diff
@@ -261,7 +262,10 @@ void FixQEq::allocate_matrix()
     i = ilist[ii];
     m += numneigh[i];
   }
-  m_cap = MAX((int)(m * safezone), mincap * MIN_NBRS);
+  bigint m_cap_big = MAX((int)(m * safezone), mincap * MIN_NBRS);
```
akohlmey:
```suggestion
  bigint m_cap_big = MAX((bigint)(m * safezone), (bigint)mincap * MIN_NBRS);
```
Shouldn't  there be casts to bigint here?

src/REAXFF/fix_qeq_reaxff.cpp
```diff
@@ -360,7 +361,10 @@ void FixQEqReaxFF::allocate_matrix()
     i = ilist[ii];
     m += numneigh[i];
   }
-  m_cap = MAX((int)(m * safezone), mincap * REAX_MIN_NBRS);
+  bigint m_cap_big = MAX((int)(m * safezone), mincap * REAX_MIN_NBRS);
```
akohlmey:
```suggestion
  bigint m_cap_big = MAX((bigint)(m * safezone), (bigint)mincap * REAX_MIN_NBRS);
```
Same as in fix_qeq.cpp

stanmoore1:
@akohlmey good catch, I will make the change

stanmoore1:
I don't think we need the second cast since it is a minimum size, e.g. the default mincap = 50, and the REAX_MIN_NBRS = 100 --> 5000 by default.

akohlmey:
> I don't think we need the second cast since it is a minimum size, e.g. the default mincap = 50, and the REAX_MIN_NBRS = 100 --> 5000 by default.

It doesn't hurt and compilers may throw warnings for comparing different size integers.

stanmoore1:
@akohlmey I tweaked it a bit, I think everything has to get promoted to `double` for the comparison anyway. If you want to change it that is fine with me

akohlmey (approved):

PR #4304, opened 2024-08-31
Add lammps_extract_atom_size() call to library interface and equivalent to python and fortran modules

### LINKED ISSUE #4302: [BUG] `Atom::extract_datatype` returns incorrect datatype for custom properties
For `i2_<name>` `Atom::extract_datatype` should return `LAMMPS_INT_2D`, however `LAMMPS_INT` gets returned instead.
The same happens for `d2_<name>`. Here, `LAMMPS_DOUBLE` gets returned instead of `LAMMPS_DOUBLE_2D`. I assume that this is not intentional?

This is the logic for `Atom::extract_datatype` for custom properties:

```c++
if (utils::strmatch(name,"^[id]2?_")) {
  int which = 0, array = 0;
  if (name[0] == 'd') which = 1;
  if (name[1] == '2') array = 1;
  
  int index,flag,cols;
  if (!array) index = find_custom(&name[2],flag,cols);
  else index = find_custom(&name[3],flag,cols);
  
  if (index < 0) return -1;
  if (which != flag) return -1;
  if ((!array && cols) || (array && !cols)) return -1;
  
  if (which == 0) return LAMMPS_INT;
  else return LAMMPS_DOUBLE;
}
```

I  think this should be

```c++
if (utils::strmatch(name,"^[id]2?_")) {
  int which = 0, array = 0;
  if (name[0] == 'd') which = 1;
  if (name[1] == '2') array = 1;
  
  int index,flag,cols;
  if (!array) index = find_custom(&name[2],flag,cols);
  else index = find_custom(&name[3],flag,cols);
  
  if (index < 0) return -1;
  if (which != flag) return -1;
  if ((!array && cols) || (array && !cols)) return -1;
  
  if (which == 0 && array == 0) return LAMMPS_INT;
  if (which == 0 && array == 1) return LAMMPS_INT_2D;
  if (which == 1 && array == 0) return LAMMPS_DOUBLE;
  else return LAMMPS_DOUBLE_2D;
}
```
instead.

### LINKED ISSUE #4303: [Feature Request] expose `Atom::find_custom` through the C-API
**Summary**

Currently, it's not possible to determine the number of per-atom entries of a custom property through the C-API.
Exposing `Atom::find_custom` would fix this.

**Detailed Description**

I'm currently in the process of writing/improving the [csmix-mit/LAMMPS.jl](https://github.com/cesmix-mit/LAMMPS.jl) julia wrapper library for lammps. Wrapping the `lammps_extract_atom` function for custom properties in a safe manner requires me to know how many columns said property has.

As an alternative, I could imagine exposing a `lammps_extract_atom_cols` method, which returns the number of columns for any per-atom property (i.e build-in and custom ones). Currently, I have to hardcode the sizes of the build-in properties, following the table listed [here](https://docs.lammps.org/Classes_atom.html#_CPPv4N9LAMMPS_NS4Atom7extractEPKc). Providing a general method to query the sizes of per atom data could be an elegant solution to both of these problems.

Unfortunately, I don't have much experience with C++/C development. I'm not sure wheter I would be able to implement this request myself.

akohlmey:
**Summary**

This adds a new function lammps_extract_atom_size() to the C library interface that reports the sizes of per-atom vectors and arrays. The implementation is also ported to Python, Fortran, Swig, and the library plugin loader.
This enables determining the number of columns for per-atom arrays which is otherwise not directly accessible.
While some of it (from atom styles) can be hard-coded, custom properties are completely dynamic.

**Related Issue(s)**

Fixes #4302.
Includes and thus closes #4309 

Implements and thus closes #4303. This implementation is more general than requested and follows similar functionality in lammps_extract_compute() and lammps_extract_fix().

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

In the process the extract_atom_iarray() and extract_atom_darray() from the NumPy wrapper were removed. They have been deprecated since 2020 and don't seem to be used anywhere in the LAMMPS distribution.

**Implementation Notes**

The actual functionality is implemented in the Atom class as "extract_size()" method. This follows the implementation of lammps_extract_datatype() which forwards to "Atom::extract_datatype()". Using the new function allows to correctly set the array dimensions for NumPy and Fortran arrays in the respective modules.

Additionally a few minor changes are included:
- change a zero size allocation with "new" in pair style quip to be nonzero to avoid some compilers choking on it
- add the "noreturn" attribute to the templated versions of Error::one() and Error::all() in order to reduce false positives from static code analysis
- step version number of LAMMPS-GUI
- fix logic bug in Atom::find_custom_ghost()
- make atom style oxdna compatible with comm_modify vel yes
- add flag to version.h for the development version
- add GitHub workflow for checking for variable length arrays using -Werror=vla with GCC
- add GitHub workflow for running the unit test suite with Linux built with -DLAMMPS_BIGBIG
- speed up GitHub workflow for unit testing with MSVC++ significantly through using ninja and ccache
- update docs for GitHub CLI with some useful examples.

**Post Submission Checklist**


rbberger (approved):
LGTM

Issue #4329, opened 2024-09-18
Deprecate PyLammps and show how to use lammps Package instead


rbberger:
About 8 years ago when the LAMMPS Python package was still very rudimentary and not well documented, I've created the `PyLammps` interface which wraps the `lammps`  package in a more Pythonic way. It was mostly designed for smaller examples and for interactive use in Jupyter Notebooks. The vision at that time was that it might be useful for tutorials.

Things have significantly improved, especially in 2020 where we did major refactors of the Library/Python interface and added more [extensive documentation](https://docs.lammps.org/Python_head.html). I believe it's time to retire PyLammps and direct people to using the `lammps` Package directly. We might be able to factor out some parts that people may find useful (like a `CmdWrapper`, IPython-plotting). But overall, the amount of people getting misled into using `PyLammps` instead of `lammps` for actual real work has been getting alarming. Specifically for running with `mpi4py` the interface becomes very fragile due to output parsing.

So the plan is two fold:
- deprecate the `PyLammps` interface, possibly extract useful features in their own helpers
- rewrite the [PyLammps Howto](https://docs.lammps.org/Howto_pylammps.html) in terms of the `lammps` interface using modern facilities.

rbberger:
@lammps/core FYI

sjplimp:
@rbberger sounds like a good idea to me

akohlmey:
Implemented in #4369 
Final removal scheduled after the next stable release in summer 2025 together with a whole bunch other legacy features

PR #4331, opened 2024-09-19

### LINKED ISSUE #4334: Failure to correctly parse an XYZ file
Summary
The new version of LAMMPS is unable to correctly parse the xyz files

LAMMPS Version
LAMMPS (29 Aug 2024 - Development - patch_29Aug2024-269-g805c2ae131

Platform
Platform Independent : tested on both MacOS Sonoma (Apple M3) and Rocky Linux 8.8 (Computing Cluster)

Description
I am trying to read a trajectory dumped into an xyz file and rerun the calculation to calculate the components of the forces acting on the atoms. When I tried to do the calculation using the version **LAMMPS (29 Aug 2024 - Development - patch_29Aug2024-269-g805c2ae131)**, it failed with the following error message:
```
Exception: stoi: no conversion
MPI_ABORT was invoked on rank 0 in communicator MPI_COMM_WORLD
  Proc: [[1845,0],0]
  Errorcode: 1

NOTE: invoking MPI_ABORT causes Open MPI to kill all MPI processes.
You may or may not see output from other processes, depending on
exactly when Open MPI kills them.
```

I was able to get the expected behavior by replacing the command ```mytype = std::stoi(line);``` at line 185 in src/reader_xyz.cpp to be ```mytype = atoi(line);```.

The expected behavior is a successful completion of the calculation with the following data
```
Per MPI rank memory allocation (min/avg/max) = 11.71 | 11.71 | 11.71 Mbytes
    PotEng    
-145.97803    
-148.7211     
-150.78172    
-152.24371    
-153.03339    
-153.22808    
-152.92975    
-152.19228    
-151.09856    
-149.9398     
Loop time of 0.000468583 on 1 procs for 10 steps with 3 atoms

Total wall time: 0:00:00
```

The xyz file which I used is copied below: 
```
3       
 i =        1, time =       85.000, E =       -76.4214103553
  O         0.0000000000         0.0000000000         0.2044070700
  H         0.6542503700         0.0000000000        -0.5095819300
  H        -0.6542503700         0.0000000000        -0.5095819300
3       
 i =        2, time =       90.000, E =       -76.4261078367
  O         0.0000000000         0.0000000000         0.1844585500
  H         0.6840662200         0.0000000000        -0.4996076700
  H        -0.6840662200         0.0000000000        -0.4996076700
3       
 i =        3, time =       95.000, E =       -76.4293408482
  O         0.0000000000         0.0000000000         0.1645120900
  H         0.7138756700         0.0000000000        -0.4896344400
  H        -0.7138756700         0.0000000000        -0.4896344400
3       
 i =        4, time =      100.000, E =       -76.4312805439
  O         0.0000000000         0.0000000000         0.1398309900
  H         0.7354607900         0.0000000000        -0.4772938900
  H        -0.7354607900         0.0000000000        -0.4772938900
3       
 i =        5, time =      105.000, E =       -76.4319496950
  O         0.0000000000         0.0000000000         0.1176405500
  H         0.7608740900         0.0000000000        -0.4661986700
  H        -0.7608740900         0.0000000000        -0.4661986700
3       
 i =        6, time =      110.000, E =       -76.4314366521
  O         0.0000000000         0.0000000000         0.0954539000
  H         0.7862800900         0.0000000000        -0.4551053400
  H        -0.7862800900         0.0000000000        -0.4551053400
3       
 i =        7, time =      115.000, E =       -76.4295434217
  O         0.0000000000         0.0000000000         0.0753198300
  H         0.8167986700         0.0000000000        -0.4450383100
  H        -0.8167986700         0.0000000000        -0.4450383100
3       
 i =        8, time =      120.000, E =       -76.4270152371
  O         0.0000000000         0.0000000000         0.0508873100
  H         0.8378091900         0.0000000000        -0.4328220500
  H        -0.8378091900         0.0000000000        -0.4328220500
3       
 i =        9, time =      125.000, E =       -76.4235815793
  O         0.0000000000         0.0000000000         0.0264605500
  H         0.8588119800         0.0000000000        -0.4206086700
  H        -0.8588119800         0.0000000000        -0.4206086700
3       
 i =       10, time =      130.000, E =       -76.4187707203
  O         0.0000000000         0.0000000000         0.0036618400
  H         0.8854050400         0.0000000000        -0.4092093100
  H        -0.8854050400         0.0000000000        -0.4092093100
```

### LINKED ISSUE #4340: Contradicting information in fix GCMC manual
**Summary**

It is unclear whether min/max keyword of fix gcmc applies as atom or molecules

**Details**

At the beginning of the fix gcmc manual it reads:
"
  max value = Maximum number of molecules allowed in the system
  min value = Minimum number of molecules allowed in the system
"
where later it is written:

"The max and min keywords allow for the restriction of the number of atoms in the simulation. "

So it is unclear whether the algorithm uses atom number or molecule number in min/max keyword of fix gcmc.

akohlmey:
**Summary**

This pull request combines multiple small fixes and changes.

**Related Issue(s)**

Fixes #4334 
Fixes the bug reported here: https://matsci.org/t/bug-in-fix-adapt/57890
Fixes #4340 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

The change to MS-MEAM changes the results and thus break backwards compatibility for MS-MEAM inputs. Unit tests and example log files have been updated or re-created accordingly.

**Implementation Notes**

- add new function `lammps_expand()` to the library interface to support #4329
- fix cut-n-paste bug for angle styles in fix adapt
- in xyz reader class, we now require a typelabel map to map string labels to atom types. There is no change for xyz files that have a numeric type (like those produced by LAMMPS by default).
- add a call to `apt-get update` before installing packages in workflows to avoid failures due to out-of-sync package indices
- add extract() method to bond style harmonic/shift/cut and angle style cosine/squared
- enable and apply clang-format and modernize fix plumed
- fix issue in lammps-open.py unit test with mpi4py version 4
- add a missing override qualifier
- improve ASPHERE/line examples and update logs
- update and correct description of running LAMMPS on Windows
- clarify fix gcmc documentation of min/max keywords
- bugfix for MS-MEAM from Mike Baskes (via Aidan)

**Post Submission Checklist**


doc/src/read_dump.rst
```diff
@@ -230,13 +230,18 @@ will then have a label corresponding to the fix-ID rather than "x" or
...
+(like when they are created with default setting by :doc:`dump style
+<dump>` *xyz*), that number will be used at atom type. The the atom
+label is a string, it is required that a map is created using the
+:doc:`labelmap command <labelmap>` that maps those labels to the
+(numeric) atom types that LAMMPS uses.  The xyz format dump file does
+not store atom IDs, so these are assigned consecutively to the atoms as
+they appear in the dump file, starting from 1.  Thus you should ensure
+that order of atoms is consistent from snapshot to snapshot in the XYZ
+dump file.  See the :doc:`dump_modify sort <dump_modify>` command if the
+XYZ dump file was written by LAMMPS.
```
jtclemm: I didn't fully understand what was meant by the new text, but if I understood correctly, how about:

If the label provided for atoms is an integer number 
(like when they are created with default setting by :doc:`dump style
<dump>` *xyz*), that number will be used as the atom type. If the atom
label is a string, it is required that a map is created using the
:doc:`labelmap command <labelmap>` that maps those string labels to the
(numeric) atom types that LAMMPS uses.  The xyz format dump file does
not store atom IDs, so these are assigned consecutively to the atoms as
they appear in the dump file, starting from 1.  Thus you should ensure
that the order of atoms is consistent from snapshot to snapshot in the XYZ
dump file.  See the :doc:`dump_modify sort <dump_modify>` command if the
XYZ dump file was written by LAMMPS.

(also fixed a few minor typos)

src/library.cpp
```diff
@@ -501,6 +501,51 @@ void lammps_error(void *handle, int error_type, const char *error_text)
...
 
+/* ---------------------------------------------------------------------- */
+
+/** expand a single LAMMPS input line from a string.
+ *
+\verbatim embed:rst
+
+This function tells LAMMPS to expand the string in *cmd* like it would process
+an input line fed to :cpp:func:`lammps_command` **without** executing it.
+The entire string is considered as command and need not have a
```
jtclemm: "as a command" ?

doc/src/read_dump.rst
sjplimp: A related Q - the PR header says that a type label map is now required for reading XYZ files.  But I think some XYZ files just use all numeric fields, including for the type.  I think older XYZ files might even skip the type (all atoms the same type).  So is what is implemented in the PR only requiring a type label map for XYZ files which use non-numeric strings in place of numeric types?  Like what Joel is saying above?

akohlmey: Previously, the label would be converted by atoi(), which does not flag an error with incorrect or incomplete conversions and would either return any leading numbers or 0. Now we are always returning a valid type. Either the number we have read (and checked to be valid) or the number from the labelmap.

I have reformulated the text a little bit for more clarity.

@sjplimp the description says that the typemap is required for "string labels" it said nothing about numeric labels. I've added a remark that those remain unchanged.

src/library.cpp
akohlmey: copy-and-modify oversight. Should be improved now.

hammondkd (approved):

doc/src/read_dump.rst
```diff
@@ -230,13 +231,27 @@ will then have a label corresponding to the fix-ID rather than "x" or
...
+fields are supported.  There are many variants of the XYZ file format.
+LAMMPS will read the number of atoms from the first line of each frame,
+ignore the second (title) line, and then read one line for each atom in the format:
+
+.. parsed-literal::
+
+   <label> <x coordinate>  <y coordinate> <z coordinate>
+
+
+If label is an integer number as label (like with XYZ files created by
```
jtclemm: I am admittedly still confused what "If label is an integer number as label" means. What does each of the two words "label" correspond to? The column label? A type label (from label map)? Or the keyword label? 

I'm pretty sure you mean a type label (and maybe the last two words "as label" are vestigial), but the term is very oversubscribed in this doc page so I think it has to be exceedingly clear.

Few other things that are tripping me up/typos:

1. Line 246, do you mean "will be used as the atom type"?
2. Line 249, is there an extra word or word missing in "one string label and that numeric type is stored"? I don't understand what this means.
3. Line 252, "ensure that *the* order of atoms"

doc/src/read_dump.rst
akohlmey: @sjplimp @jtclemm is this now sufficiently clear in the pull request? If not, please let me know where there are still problem. thanks.

doc/src/read_dump.rst
sjplimp: If the atom label is a numeric integer (like with XYZ ...)

doc/src/read_dump.rst
```diff
@@ -230,13 +231,27 @@ will then have a label corresponding to the fix-ID rather than "x" or
...
+created with default settings by :doc:`dump style <dump>` *xyz*), that
+number will be used at atom type.  If the atom label is a string, then
```
sjplimp: as the atom type.

doc/src/read_dump.rst
```diff
@@ -230,13 +231,27 @@ will then have a label corresponding to the fix-ID rather than "x" or
...
+it is required that a map is created using the :doc:`labelmap command
```
sjplimp: then a map must have been created using ...

doc/src/read_dump.rst
```diff
@@ -230,13 +231,27 @@ will then have a label corresponding to the fix-ID rather than "x" or
...
+<labelmap>`. This map needs to associate each (numeric) atom type with
+one string label and that numeric type is stored.  The xyz format dump
```
sjplimp: new paragraph starting: The xyz format...

akohlmey: all done.

jtclemm (approved):

sjplimp (approved):

doc/src/read_dump.rst
akohlmey: @jtclemm it seems you are looking at the version *before* commit https://github.com/lammps/lammps/pull/4331/commits/a31f083babfc98c90d7a55c68eef42f07576b33a

2. The labelmap must be a so-called "complete" map. That is there must be a (unique) string for each atom type. That is a requirement of the labelmap command. The map is used to determine the atom type (which is an integer). I have further rewritten this as:
```
This map needs to associate each (numeric) atom type with
a string label. The numeric atom type is stored internally.
```

3. fixed. that was an old issue and repeated a few lines below in the documentation of the *molfile* format.

jtclemm: Yes, that was a comment from last week.

PR #4339, opened 2024-09-25
Fixing bugs in GRANULAR and mistakes in RHEO documentation

### LINKED ISSUE #4216: [BUG] _fix nonaffine/displacement not work in rerun
**Summary**

I use rerun to calculate nonaffine displacement by using fix nonaffine/displacement command, but results give all zero value for this fix calculation when I dump it.

LAMMPS (17 Apr 2024) on Windows

**Expected Behavior**

This fix should give the non-zero value. Rather than using rerun, I re-perform the simulation, it works. 

**Actual Behavior**

But when I use rerun, I only get zero value, which is, of course, wrong result.

**Steps to Reproduce**

units 		real
dimension 	3
processors 	* * *
boundary 	p p p
atom_style 	charge

read_data	L300K.dat

include		pot.mod

timestep      1

#fix 1 all nve

fix 3 all nonaffine/displacement 100 d2min custom 3.5 fixed 0
dump            3 all custom 100 stressd.text id type  x y z f_3[1]

rerun  stressy.text first 5610000 every 100000 dump x y z

I also try to change the value after the keyword “nonaffine/displacement” to match the dump file, but none of them works.

### LINKED ISSUE #4337: [BUG]  GRANULAR package fix_pour:  wrong insertion volume calculation (particularly in in SI units)
**Summary**
The fix_pour does not insert the proper number of particles every the proper number of time steps if the region is a box of width along the y direction smaller than one.

This is due to the line   `if (dy < 1.0) dy = 1.0;` in this portion of code leads to a wrong volume calculation, especially in SI units where lengths are typically much smaller than one.

https://github.com/lammps/lammps/blob/b95f7c2ab2b9cba44c3f690b9bfb16b36cc11b98/src/GRANULAR/fix_pour.cpp#L226-L232

Develop branch, last commit, Ubuntu 20.04

**Expected Behavior**

It should compute the volume using the actual width of the box, even if smaller than one. 

**Steps to Reproduce**

```
atom_style      sphere
units           si

region          boxreg block 0 0.016 0 0.064 0 0.016
create_box      2 boxreg
change_box      all boundary p p f

region          insertbox block 0 0.016 0.0064 0.0576 0.0128 0.016 units box
neighbor        0.0008 bin
neigh_modify    exclude type 1 1  # do not compute fixed particles interaction forces

fix             bot_wall all wall/gran granular hooke 263000 0.3 damping coeff_restitution tangential mindlin 75142.8571428572 1 0.4 rolling none twisting none zplane 0 NULL

fix             1 all nve/sphere   # mandatory for granular style simulations
fix             grav all gravity 9.81 vector 0 0 -1 

comm_modify     vel yes   # mandatory for granular style simulations
thermo_style    custom step atoms ke
thermo_modify   lost warn
thermo          ${n_print}
thermo          500

timestep        1.31805410714104e-06
fix             ins all pour 25000 2 10 region insertbox diam range 0.00064 0.00096  dens 2650 

timer           loop
run             100000
```

### LINKED ISSUE #4338: [BUG]  GRANULAR package fix_pour: frequency of insertion not updated if timestep changed after defining fix_pour in input script
**Summary**

The frequency of insertion (`nfreq` in fix_pour.cpp) is not updated if timestep dt is changed _after_ defining the fix pour in the input script. This is unexpected, since all examples using the fix pour set the time step after defining the fix pour.

Develop branch, last commit, Ubuntun 20.04

**Expected Behavior**

From the examples, we expect that change of the time step, even after defining the fix pour, would be taken into account in the pouring of particles. Indeed, the number of timesteps between two insertion events is defined as the time needed for a particle to reach the bottom of the insertion zone divided by the time step dt.

Otherwise, we should modify the examples, and explicitly mention this restriction in the documentation.

**Actual Behavior**

Changing `timestep        0.001` in [this example](https://github.com/lammps/lammps/blob/develop/examples/granular/in.pour.flatwall)  to `timestep        0.0001` will not change the number of steps between insertion events, while it should multiply it by 10.

Moving the line `timestep        something` before:

https://github.com/lammps/lammps/blob/b95f7c2ab2b9cba44c3f690b9bfb16b36cc11b98/examples/granular/in.pour.flatwall#L49-L50

will update it properly.

jtclemm:
**Summary**

This PR includes 3 sets of changes:
1) Patches to fix pour to resolve the issues listed below
2) Clarifying error messages in fix pour
3) Minor updates to RHEO documentation
4) Adding a safeguard to fix gran/wall that blocks use of the *region* option (only applicable to fix gran/wall/region, its child class) which otherwise causes a segfault
5) Clarifying that fix nonaffine/displacement does not currently work with the rerun command

**Related Issue(s)**

Fixes #4337
Fixes #4338
Closes #4216

**Author(s)**

Joel Clemmer (SNL)

**Licensing**

**Backward Compatibility**

NA

**Implementation Notes**

A big chunk of the constructor in fix pour was moved to the `init()` method. This ensures calculations involving the timestep or total # of atoms are not performed until a run command is issued. This is necessary if a user defines the timestep or adds atoms after a fix pour command. Affected examples were updated.

**Post Submission Checklist**


src/GRANULAR/fix_pour.cpp
```diff
@@ -310,6 +216,108 @@ void FixPour::init()
   if (fixgrav->varflag != FixGravity::CONSTANT)
     error->all(FLERR, "Fix gravity for fix pour must be constant");
 
+  // Perform one time initialization operations
+  // outside of constructor in case fix definition precedes timestep defintion/atom creation
+
+  if (!initialize_flag) {
```
akohlmey, line 222: What if an input has multiple "run" statements and the timestep is changed in between?

jtclemm: I tested this and it looks like that scenario was already covered by this:
https://github.com/lammps/lammps/blob/3a53b01ed9dbd1af82e2b5af83a493b53c4760d7/src/GRANULAR/fix_pour.cpp#L1054-L1057

The only time this wasn't triggered seems to have been when the timestep was changed between the fix declaration and the first run command (so there was nothing to reset).

cgadal:
What do you think about adding to the fix pour documentation a quick sentence describing how the insertion time steps and number of particles inserted per insertion are calculated? 
For example: 

The interval between two insertion events is calculated as the time a particle needs to fall across the entire height of the insertion region, taking into account any region's motion. The number of particles inserted per insertion event is calculated as $\phi V_{r}/V_{p}$, where $\phi$ is the required volume fraction, $V_{r}$ the volume of the region and $V_{p}$ the volume of a particle.

jtclemm:
> What do you think about adding to the fix pour documentation a quick sentence describing how the insertion time steps and number of particles inserted per insertion are calculated? For example:
> 
> The interval between two insertion events is calculated as the time a particle needs to fall across the entire height of the insertion region, taking into account any region's motion. The number of particles inserted per insertion event is calculated as ϕ V r / V p , where ϕ is the required volume fraction, V r the volume of the region and V p the volume of a particle.

That might be useful, but it isn't strictly true since there are many conditions that can affect the actual insertion rate. Maybe something like this is a bit more accurate?

```
The exact details of the insertion procedure depend on many factors
(the range of particle diameters inserted, whether molecules are
inserted, potential failed particle insertions due to overlaps). However, 
in the simplest scenario where monodisperse atoms are always successfully 
inserted, the procedure works as follows. First, the number of 
timesteps between two insertion events is calculated as the time a
particle needs to fall across the entire height of the insertion 
region, taking into account a region's motion and gravity. Next, the 
target number of particles inserted per insertion event is calculated
as the product of the volume fraction and the volume of the insertion
region divided by the volume of a particle (or area in 2D). Insertion
events are then repeated until N particles have been inserted, where the
final insertion event only inserts the remainding number of particles 
needed to reach N total. At the beginning of the run, the insertion
interval and the target number of inserted particles per insertion event
are reported in the log/screen.
```

I can see the value of explaining this, but the level of detail and its conditions may be a bit much. 

@akohlmey Do you have any thoughts?

akohlmey:
> @akohlmey Do you have any thoughts?

Fix pour has been around a very long time and people have used it with great success. So I think for the most part the docs should be ok "as is" or else people would have commented on it before.

That said, I am not using it much, so perhaps @sjplimp has some idea.

At any rate, this kind of change can always be done later.

sjplimp:
I think the long paragraph by @jtclemm would be useful for some users, but is TMI for most users.

I wonder if this kind of info could be included in the doc page in a new colored section (similar to a NOTE) with a label like "Implementation Details".  So it would be easier for the majority of readers to skip.

jtclemm:
> I wonder if this kind of info could be included in the doc page in a new colored section (similar to a NOTE) with a label like "Implementation Details". So it would be easier for the majority of readers to skip.

@sjplimp Maybe like this? 

https://github.com/lammps/lammps/blob/4436197081e7812de45edeee657f72a4ea3a9a8e/doc/src/fix_pour.rst?plain=1#L158-L171 

I'm not familiar with an instance in the doc pages with an "implementation details" tag.

I also streamlined the text.

sjplimp:
yes, but not as a blue-boxed note, but as a green-boxed Implementation
Details, or the like
So Sphinx would have to support that

Steve

akohlmey:
@jtclemm you can use:
```rst
.. admonition:: Implementation Notes
   :class: Hint
```
instead of
```rst
.. note::
```

jtclemm:
> @jtclemm you can use:

That's good to know about, thanks!

src/GRANULAR/fix_wall_gran.cpp
```diff
@@ -165,6 +165,9 @@ FixWallGran::FixWallGran(LAMMPS *lmp, int narg, char **arg) :
     wallstyle = REGION;
     idregion = utils::strdup(arg[iarg+1]);
     iarg += 2;
+    // This option is only compatible with fix wall/gran/region
+    if (strcmp(style, "wall/gran/region") != 0)
```
akohlmey:
```suggestion
    if (!utils::strmatch(style, "^wall/gran/region"))
```
This would be future-proof in case the fix is ported on an accelerator package

akohlmey:
@jtclemm here is one observation about the RHEO package that showed up in static code analysis with coverity scan.
While it is not incorrect as such, there are many complaints that there is no test for a null pointer after the dynamic casts.

What the code analysis tool does not know, is that with the `modify->get_fix_by_style()` or `modify->get_compute_by_style()` you are already limiting what can be chosen.

In that context it is rather unfortunate that fix rheo is named without a suffix like /basic or /general or /nve or similar.
This makes looking up fixes by styles in a way that is compatible with accelerators rather difficult and adding checks for null pointers returned by a dynamic cast more important to avoid a segfault in case a check for a style returned the wrong fix instance.

Bottom line: there is nothing wrong, and using the dynamic casts is correct and better than anything else we did in the past. We just need to be aware of the potential ambiguities for the future and perhaps make some adjustments when some refactoring of the package would be planned.

jtclemm:
> @jtclemm here is one observation about the RHEO package that showed up in static code analysis with coverity scan. While it is not incorrect as such, there are many complaints that there is no test for a null pointer after the dynamic casts.
> 
> What the code analysis tool does not know, is that with the `modify->get_fix_by_style()` or `modify->get_compute_by_style()` you are already limiting what can be chosen.
> 
> In that context it is rather unfortunate that fix rheo is named without a suffix like /basic or /general or /nve or similar. This makes looking up fixes by styles in a way that is compatible with accelerators rather difficult and adding checks for null pointers returned by a dynamic cast more important to avoid a segfault in case a check for a style returned the wrong fix instance.
> 
> Bottom line: there is nothing wrong, and using the dynamic casts is correct and better than anything else we did in the past. We just need to be aware of the potential ambiguities for the future and perhaps make some adjustments when some refactoring of the package would be planned.

Adding null ptr checks would be easy and I'm happy to do it after all the dynamic casts, but I'm not quite sure what you mean with accelerator compatibility. Do you mean if one was to potentially look for "rheo/kokkos"? 

I suppose it may not be too late to also rename the fix something else...

akohlmey:
> but I'm not quite sure what you mean with accelerator compatibility. Do you mean if one was to potentially look for "rheo/kokkos"?

If styles are ported to accelerator packages, a suffix ("/gpu", "/kk", "/omp") gets appended to the style name. Thus we can match the unaccelerated and the accelerated version using `utils::strmatch("^name")`, since that will also match "name/gpu", "name/kk", or "name/omp". This works for fix rheo/oxidation, fix rheo/thermal and so on, but it fails for fix rheo, since then `utils::strmatch("^rheo")` will match *all* of those fixes, so you would have to use strcmp() with checking for each accelerator. If the fix was renamed, that ambiguity would go away. Before we implemented utils::strmatch, there were lots of cases where accelerated version were not detected or there were long lists of strcmp() calls.

akohlmey:
> Adding null ptr checks would be easy

As I tried to point out, that would at the moment only have the purpose of appeasing the static code analysis tool, since it doesn't know what we know of how you get the pointer that you are casting. There are plenty of places in LAMMPS where the dynamics casts are unchecked. It would only serve a real purpose in the code, if for one reason or another, there would be unexpected ambiguities looking up the fixes or computes.

stanmoore1:
I will say that it is good practice in many cases to check for `nullptr`, otherwise it can lead to bugs. Here was one case that gave unexpected behavior until we added the check: https://github.com/lammps/lammps/blob/develop/src/KOKKOS/pair_pace_kokkos.cpp#L257-L260

In other cases we may already check a flag inside the class before the cast, or somehow otherwise know that it is a valid cast, and then checking isn't necessary.

akohlmey:
@jtclemm How do you want to proceed with the PR?
It looks ready to me.
We won't get an approval from Steve until he returns from vacation. Do you still want somebody else (Dan?) have a look before merging? Perhaps sending a reminder may help.

We are reaching the point where we have enough changes for a feature release and for that I would like to merge as many PRs that are ready, especially since this is now at the beginning of the release cycle.

If you have some additional fixes or updates, you can always submit another PR or add them to my latest "collection" PR.

jtclemm:
> @jtclemm How do you want to proceed with the PR?
> It looks ready to me.

I agree. I think the only input needed from Steve was regarding that implementation note. We could either just temporarily remove it from this PR or leave it in and revise/remove it down the road if needed. I have no strong opinions.

I'll ping Dan again and maybe we can give him till tomorrow, but we already discussed offline the gist of changes to fix pour and he agreed.

akohlmey:
> I think the only input needed from Steve was regarding that implementation note.

The way I read his comments, the only missing piece was the color scheme and we sorted that out.

dsbolin (approved):

akohlmey (approved):

PR #4344, opened 2024-10-03
Prepare development branch for requiring C++17 and deprecating GNU make support for some packages

### LINKED ISSUE #3840: C++14/C++17 in LAMMPS
## Summary

It's been 5 years since issue #1659 (2019), where we kept track of the C++11 transition for LAMMPS, and 4 years (2020) since C++11 was allowed in LAMMPS core.

In a similar way, this issue now documents the current state of common compilers among popular Linux distributions.
By now, moving to C++14 is a trivial change, since most compilers at least default to C++14, if not already C++17.

The only exception is the default RHEL7 compiler, which barely supports C+11. I personally don't consider this a blocker, given that most HPC centers, even if they still use RHEL7-based systems, will provide newer GCC compilers.

As a side note, GCC versions prior to v9.x should not be used for C++17 due to various quirks and bugs.

Currently, C++17 is currently only required by Kokkos 4.

C++20 will be required in Kokkos 5.

## Proposal to assess impact
```c++
// for C++14
#ifndef LAMMPS_CXX14
#if __cplusplus <= 201402L
  #error LAMMPS is planning to transition to C++14. Do disable this error please use a C++14-compliant compiler or define LAMMPS_CXX14 in your makefile
#endif
#endif

// alternative for C++17
#ifndef LAMMPS_CXX17
#if __cplusplus <= 201703L
  #error LAMMPS is planning to transition to C++17. Do disable this error please use a C++17-compliant compiler or define LAMMPS_CXX17 in your makefile
#endif
#endif
```

## Target Platforms

| OS |  g++ Version | g++ default | clang++ Version | clang++ default |
| --- | ---- | --- | --- | ---|
| Ubuntu 18.04 LTS   | 7.4.0 | C++14    | 6.0.0 | C++14 |
| Ubuntu 20.04 LTS (EOS 04/2023)   |  9.4.0  | C++14  | 10.0.0 | C++14 |
| Ubuntu 22.04 LTS (EOS 04/2027)   |  11.3.0 | C++17  | 14.0.0 | C++14 |
| Ubuntu 24.04 LTS              |  13.2.0 | C++17  | 18.1.3 | C++17 |
| RHEL7 (EOL July 2024)    | 4.8.5 | C++98 (up to C++11 with `-std=c++11`) | 3.4.2 | C++98 (up to C++11 with `-std=c++11`) |
| RHEL 8    | 8.5.0 | C++14 | 15.0.7 | C++14 |
| RHEL 9    | 11.3.1 | C++17 | 15.0.7 | C++14 |
| Fedora 37 (EOL 12/15/2023)     | 12.2.0 | C++14    | 15.0.7 | C++14 |
| Fedora 38     | 13.0.1 | C++17    | 16.0.5 | C++17 |
| Fedora 40     | 14.2.1 | C++17    | 18.1.6 | C++17 |
| OpenSuse Leap 15 | 7.3.9 | C++14    | 15.0.7 | C++14 |
| OpenSuse Tumbleweed | 13.1.1 | C++17    | 16.0.6 | C++17 |

### Testing default C++ compiler standard
```
g++ -dM -E -x c++  /dev/null | grep -F __cplusplus
```

## References:
 - https://en.cppreference.com/w/cpp/compiler_support

### LINKED ISSUE #4418: Kokkos without Makefiles
This issue is mostly to track https://github.com/kokkos/kokkos/issues/7621

We will have to figure out what to do in LAMMPS e.g. if we drop Kokkos support in our Makefile system.

akohlmey:
**Summary**

This pull request adds warnings when compiling LAMMPS with a compiler that is not C++17 compatible.
Also it adds warnings when using the GNU make build procedure with certain packages.
KOKKOS was added to the list since the Kokkos library developers plan to drop support for makefiles.
And a warning is printed if the CMake version is not 3.20 or later.

**Related Issue(s)**

This was discussed and agreed upon at the recent LAMMPS Developer meeting in Albuquerque.
Closes #3840 
Closes #4418 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

The changes try to make the C++ warning only show up with compilers that do not support C++17 by default. Most current compilers do, only those on outdated systems (RHEL 7, or Ubuntu 18) will produce the warnings.

**Post Submission Checklist**


athomps (approved):
I approve

rbberger (approved):

ndtrung81 (approved):

cmake/CMakeLists.txt
```diff
@@ -3,6 +3,9 @@
 # CMake build system
 # This file is part of LAMMPS
 cmake_minimum_required(VERSION 3.16)
+if(CMAKE_VERSION VERSION_LESS 3.20)
+  message(WARNING "LAMMPS is planning require at least CMake version 3.20 by Summer 2025. Please upgrade!")
```
stanmoore1: LAMMPS is planning _to_ require at least

src/MAKE/Makefile.mpi
```diff
@@ -7,12 +7,12 @@ SHELL = /bin/sh
 # specify flags and libraries needed for your compiler
 
 CC =		mpicxx
-CCFLAGS =	-g -O3 -std=c++11
+CCFLAGS =	-g -O3 # -std=c++17
```
stanmoore1, line 10: Why is `-std=c++17` commented out, unlike the previous?

akohlmey: Because we do not yet *enforce* the C++17 requirement. As we determined from the survey of current Linux distributions and their default compilers, many (if not most) of them will already default to C++17 and they can compile LAMMPS without any problems. People with a non-C++17 compiler would get an error from the compiler if the flag was uncommented. But they would also get an error from the check in lmptype.h which has the instruction to add `-DLAMMPS_CXX11` to avoid this error.

Of course the comment will be removed *after* the next stable release when we actually will require C++17

stanmoore1: Got it, thanks for the explanation

stanmoore1 (approved):

junghans (approved):

PR #4346, opened 2024-10-03
features + bug fixes for granular simulations w/ Kokkos

### LINKED ISSUE #4219: [BUG] _Wrong forces input for bulk stress tensor calculation in pair granular/hooke/history/kk_
**Summary**

Miscalculated stress tensor with pair granular/hooke/history/kk under isotropic compression of an FCC granular sample (see Ref. 1 for more information about the input file and system details). 

version: 17 Apr 2024

**Expected Behavior**
The diagonal component of stress tensor must be almost equal, and the off-diagonal ones must be nearly zero.

**Actual Behavior**

Unphysical differences for the diagonal stress components and high values of the off-diagonal ones.

**Bug Identification**
By delving into the file pair_gran_hooke_history_kokkos.cpp, I identified that the problem is related to the force input in procedures *ev_tally_xyz_atom* and *ev_tally_xyz* in *PairGranHookeHistoryKokkos::operator()*. Both receive the accumulated forces *fx_i*, *fy_i*, and *fz_i* as input, following the code notation; therefore, there is an overaccumulation of stresses.

**Potential Solution**
I have addressed the issue using the pair forces *fx*, *fy*, and *fz*; the outcome results are the expected ones. This is because the stress for every pair force must be computed before the summation is correctly done. This bug does not affect the force calculations but the resulting stress and barostat simulations.

*replace current code lines in PairGranHookeHistoryKokkos::operator()*
'''
if (EVFLAG == 2)
      ev_tally_xyz_atom<NEIGHFLAG, NEWTON_PAIR>(ev, i, j, fx_i, fy_i, fz_i, delx, dely, delz);
if (EVFLAG == 1)
      ev_tally_xyz<NEWTON_PAIR>(ev, i, j, fx_i, fy_i, fz_i, delx, dely, delz);
'''
*by*
'''
if (EVFLAG == 2)
      ev_tally_xyz_atom<NEIGHFLAG, NEWTON_PAIR>(ev, i, j, fx, fy, fz, delx, dely, delz);
if (EVFLAG == 1)
      ev_tally_xyz<NEWTON_PAIR>(ev, i, j, fx, fy, fz, delx, dely, delz);

'''

**Steps to Reproduce**
Since I wanted to reproduce the results of the cited paper, I also modified the Hooke model to have the Hertz model instead, creating a new pair: pair granular/hertz/history/kk. That is done by multiplying tangential and normal forces by the polyhertz similar to pair_gran_hertz_history.cpp from the GRANULAR package. The .h and .cpp files of the hertz pair are also shared here. Perhaps this pair interaction or a version of it can be deployed in LAMMPS.

1. Compile lammps with kokkos and granular packages, including the ,h and .cpp files.
2. Run with the attached input file and the initial positions file.
'''
mpirun -np 1 lmp -k on g 1 -sf kk -pk kokkos newton off neigh half comm device gpu/aware on binsize 0.0105-in Isotropic_mu0_0.0.in
'''

[attached files](https://drive.google.com/file/d/1CNB_xPTOUBgqsybQ2OV1vE9MkApbjnEV/view?usp=sharing).

1 - Salomon, J., O’Sullivan, C., & Patino-Ramirez, F. (2024). On data benchmarking and verification of discrete granular simulations. Data in Brief , 53 , 110252.
Data: [Validation and Benchmark Dataset for Discrete Element Method Simulations](https://zenodo.org/records/10411062)

cjknight:
**Summary**

This is a collection of added features and bug fixes encountered when trying to run a Kokkos workload that included the gran/hooke/history pair_style, deform fix, and temp/deform compute.

* Added brownian/kk pair style
* gran/hooke/history
  * fixed call to virial function: e.g. fx_i was being used instead of fx
  * fixed issue with global virial not calculated correctly
  * might be simpler fix, but this is largely copy+paste from lj/cut/dipole/cut/kk (minus energy)
* npair_kokkos
  * fixed issue where particle pairs were missing 
* compute temp/deform
  * fixed host sync issue to correct output
* fix deform
  * fixed a performance regression where host sync only needs to occur in one scenario
  * I believe this is correct based on version before recent refactoring
  * It was a significant performance degradation in the particular workload I’m working with (Loop time of 45 vs. 9 seconds on single A100 GPU)
* gran/hooke/history comm memory leak
  * the prior size of comm array was being checked for growing the array or not
  * this value fluctuates over course of run leading to memory leak causing GPU runs to quickly crash
  * new variable introduced to track largest size of array; stable simulation for millions of steps (i.e. GPU memory usage plateaued)
* I’m withholding a kokkos pairstyle for the time being derived from lubricate
  * The modification to pair_lubricate.cpp/.h can be discarded for now if preferred

**Related Issue(s)**

Fixes #4219

**Author(s)**

Chris Knight (ANL)
Suman Chakraborty (Purdue)

**Licensing**

**Backward Compatibility**

No issues, but this will change results for the better.

**Implementation Notes**

Several test simulations were validated on Polaris@ALCF comparing energies, forces, and virials for tens of steps as well as comparison of CPU & GPU results from multi-million step simulations.

**Post Submission Checklist**


stanmoore1:
@cjknight Wow good work, thanks! We'll take a look.

akohlmey:
@cjknight check out issue #4219 this seems related.

cjknight:
Ooooh, yes, it is. That's the same fix here. 

I believe that one-liner is enough for correct per-atom virials, but the total pressure still didn't match cpu runs. Copying over and massaging the virial tally logic from other pair style gave me consistent virial-related results: cpu, kokkos/openmp, kokkos/cuda.

akohlmey:
@cjknight please go to the src folder, type `make fix-whitespace` and commit the changes. That should recover the programming style test.

stanmoore1:
@cjknight I made a few tweaks, can you retest? Especially `compute temp/deform/kk`, I don't think the sync to host is needed--if it is needed then I think there is a bug somewhere else. Otherwise this looks great.

stanmoore1:
Also need to update the docs for `brownian/kk` pair style

cjknight:
@stanmoore1  I believe the problem in this particular workload is that the output of `compute temp/deform/kk` is immediately consumed by another compute that runs on the host. Here's a sorta minimal reproducer for this one.

```
$ cat in.test
variable	rate equal 1
variable	fric equal 50000000*${rate}

units		lj
atom_style	sphere
dimension	3 
boundary        p p p
newton		off

variable        L equal 30
lattice         fcc  1.07
region          box  block 0 ${L} 0 ${L} 0 ${L}
create_box      1 box
create_atoms    1 box

pair_style      gran/hooke/history ${fric} ${fric} 0 0 1 0 
pair_coeff      * *

comm_modify	vel yes
change_box	all triclinic

fix		tintegrate all nve/sphere

fix		def all deform 1 xz erate ${rate} remap v units box

compute		deformT all temp/deform
compute		realT all temp/sphere bias deformT
thermo_modify	temp realT

thermo_style    custom step temp c_deformT c_realT ke pe etotal press
thermo		1

timestep	0.000001
run             1
```

cpu run
```
   Step          Temp        c_deformT       c_realT         KinEng         PotEng         TotEng         Press     
         0   0              122.98532      61.492373      0              0              0              131.59307    
         1   0              122.98532      61.492373      0              0              0              131.59307    
```

gpu run
```
   Step          Temp        c_deformT       c_realT         KinEng         PotEng         TotEng         Press     
         0   0              122.98532      61.492373      0              0              0              131.59307    
         1   0              122.98532      0              0              0              0              0            
```

I get correct thermo output adding back in a couple `atomKK->sycn(Host, V_MASK);` calls. I'd have to look again what setup is doing differently that step 0 output matches.

stanmoore1:
@cjknight OK thanks I'll take a look

cjknight:
Kokkos support now mentioned in the top-level pair commands and brownian pages. i'm sure i'm missing where else it needs to be called out...

akohlmey:
> i'm sure i'm missing where else it needs to be called out...

If you run "make html" in the doc folder, it should translate the manual and run a list of checks and mention missing items or collisions.
Any warnings produced by this are hints of missing changes or typos.

cjknight:
thanks @akohlmey !!

stanmoore1:
@cjknight can you also add the boiler plate note as in other styles (if it isn't in there already)? It goes near the bottom. Thanks

```
----------

.. include:: accel_styles.rst

----------
```

I see how to fix the sync issue, will require a small refactor...

stanmoore1:
I confirmed that the refactor fixed the issue in the reproducer above. I will run this through Kokkos regression testing, then it should be ready to merge from my POV.

cjknight:
Looks good to me with the "real" workload I've been playing with. 

It's a shame adding X_MASK to that sync increases runtime so much (10.2 vs. 9.2 seconds).

stanmoore1:
>It's a shame adding X_MASK to that sync increases runtime so much (10.2 vs. 9.2 seconds).

@cjknight do you mean this line? https://github.com/lammps/lammps/pull/4346/files#diff-5740126004ad4ff0b529785a58530fe1697bbe5a7c49e5af4c0608729d4f34e1R206

In the script above, it uses `compute temp/sphere` which isn't ported to Kokkos. That slowdown would likely go away and even get a speedup if someone ported `compute temp/sphere`.

cjknight:
Yeah, that's the line I was referring too.

I can look into porting temp/sphere, but not in this PR.

stanmoore1:
The issue is that `compute temp/sphere` moves all the data to host, then marks everything including "x" as modified because as a non-Kokkos style we don't know what was actually changed. Then the new "x" sync moves data even though it isn't actually needed. The easy solution would be to change the masks in  `compute temp/sphere` to only move the needed data. Better would be to port it to Kokkos, but that is a lot more work.

akohlmey:
@cjknight @stanmoore1 This is now merged with upstream, so the automated tests will pass. I also re-checked the docs and they are consistent (there is one warning, but that is anticipating a change from another PR).
I manually triggered the full non-KOKKOS regression tests. If those pass, this PR should be ready for merge.

stanmoore1:
@cjknight I think that slowdown should be gone with 2c73866.

jtclemm (approved):
I went through the changes and everything looks good to me, but I don't think my review is the important one here.

akohlmey:
> I went through the changes and everything looks good to me, but I don't think my review is the important one here.

Thanks anyway. When making changes to accelerator packages, there is always the lingering doubt that one is missing something that is required by the model and that the optimizations/corrections will work for the test case(s) at hand, but may fail for others in unexpected ways. So to have a quick "sanity check" is very helpful.

Specifically for neighbor lists and most types of granular simulations we are rather light on tests.

stanmoore1:
Regression test failure seems unrelated to this PR:
```
  File "/data/builder/workspace/dev/pull_requests/ubuntu/regression_tests/cmake_regression_mpi/pyenv/lib/python3.10/site-packages/nose/suite.py", line 106, in _set_tests
    if isinstance(tests, collections.Callable) and not is_suite:
AttributeError: module 'collections' has no attribute 'Callable'
Traceback (most recent call last):
```

stanmoore1:
Internal Kokkos regression tests (OpenMP and CUDA) passed with no new failures.

stanmoore1:
This PR is ready to merge from my POV.

PR #4349, opened 2024-10-06

akohlmey:
**Summary**

This pull request contains multiple small changes and fixes

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

The following individual changes are included.
- simplify the preset for the IntelLLVM compiler. Use local linalg since local BLAS/LAPACK may not be ABI compatible.
- use `-fp-model precise` with IntelLLVM compilers.
- use std::vector<>.data() consistently in the ELECTRODE package and fix sgcmc to access the underlying pointer
- Skip tests with KokkosOpenMP when GPU support is included (e.g. CUDA, HIP, or SYCL) since KOKKOS enforces use of GPUs in that case, so we cannot test with OpenMP only as desired.
- relax some force-style test epsilon values for compatibility with IntelLLVM compilers
- must use fabs() instead of abs() when operating on floating point numbers in fix qeq/ctip
- reformat files with clang-format
- fix bug when building Kokkos atom map on the host (i.e. `-pk kokkos atom/map no`) using a "hash" style map. The shallow copy of `Kokkos::UnorderedMap` is not working as expected, need to use an alias instead
- consolidate kspace communication enumerators in kspace.h header
- remove unused code in PPPM/MSM styles of the DIELECTRIC package
- allow Sphinx 8.x for building the documentation
- replace globbed toctrees for pair, bond, angle, dihedral, improper, computes, and fixes with python script generated files so that duplicate entries are avoided (and Sphinx won't complain about them or pick the wrong one as the canonical one)

**Post Submission Checklist**


robeme (approved):
As far as ELECTRODE is concerned, everything is fine. However, @ludwig-ahrens and I noticed that some efield unit tests did not complete. We traced this back to the missing prerequisite `atom dipole` in the yamls `fix-timestep-efield_dipole_const.yaml` and `fix-timestep-efield_dipole_variable.yaml`. Would it be a good idea to add this prerequisite in both yamls?

akohlmey:
> We traced this back to the missing prerequisite `atom dipole` in the yamls `fix-timestep-efield_dipole_const.yaml` and `fix-timestep-efield_dipole_variable.yaml`. Would it be a good idea to add this prerequisite in both yamls?

@robeme Thanks for the note. These tests were added only recently by @alphataubio.
Yes, the dependency needs to be added. This is done in commit e6c9353ccf028b9a9ad928bc8bc8709075af59b4

akohlmey:
@ndtrung81 please have a look at the changes I made to the dielectric package.

ndtrung81:
@akohlmey sorry for the bug in `msm_dielectric.cpp`. Thanks for fixing it! Also for cleaning up `pppm*/dielectric` and `pppm/gpu`.

akohlmey:
@lammps/core this is waiting now for almost a week for a review...

ndtrung81 (approved):

PR #4353, opened 2024-10-09
CMake: install shared libraries on Windows with correct target

martok:
**Summary**

This fixes a CMake install issue on Windows: DLLs built as SHARED must be installed using the RUNTIME target, unlike Linux and Mac where .so are LIBRARY. Installing the DLL in `CMAKE_INSTALL_BINDIR` also means it is installed next to the exe which depends on it.

See [Documentation](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Install.html#installing-targets) and the examples in [Exporting Targets](https://cmake.org/cmake/help/latest/guide/importing-exporting/index.html#exporting-targets).

**Author(s)**

Sebastian Hütter (OvGU)

**Licensing**

**Backward Compatibility**

No change.

**Implementation Notes**

In theory, one could switch out which Target to use depending on the platform, but including both does not do anything negative. Windows builds continue to include the static lib while for *nix nothing changes.

**Post Submission Checklist**


rbberger (approved):
LGTM

PR #4356, opened 2024-10-11
add extract() function to few angle and bond styles 

evoyiatzis:
**Summary**

Add or expand extract() method to bond class2, angle class2/p6, angle cosine/buck/6d, bond morse, bond nonlinear, angle cross, angle amoeba, angle mesocnt, angle gaussian and bond gaussian

**Related Issue(s)**

No issues

**Author(s)**

E. Voyiatzis at NovaMechanics Ltd - email [evoyiatzis@gmail.com](mailto:evoyiatzis@gmail.com)

**Licensing**

**Backward Compatibility**

It will not break backward compatibility

**Post Submission Checklist**


evoyiatzis:
Is there a yaml file for unit test of angle_style cosine/buck6d? I cannot find one in the unittest/force-styles/tests/ folder.

akohlmey:
> Is there a yaml file for unit test of angle_style cosine/buck6d? I cannot find one in the unittest/force-styles/tests/ folder.

Not all styles have tests. Some cannot be tested this way, some require some work, some are simply not yet done.
WYSIWYG!

sjplimp (approved):

evoyiatzis:
bond_style mesocnt inherits from bond harmonic so I guess it does not require an extract () method right?

sjplimp:
@akohlmey Why does bond harmonic derive from bond mesocnt - that seems illogical on the face of it.

evoyiatzis:
> @akohlmey Why does bond harmonic derive from bond mesocnt - that seems illogical on the face of it.

@akohlmey  @sjplimp 
My bad ! Indeed it is mesocnt that it is derived from harmonic

sjplimp:
No, I mistyped - I saw that bond mesocnt derives from bond harmonic. 
However:
(a) Normally this would be a bond harmonic/mesocnt command, not bond mesocnt, to indicate the derivation.
(b) The doc page is incomplete - it doesn't say how the chiral indices n,m are used to alter the harmonic formula.
      I assume they alter K, but it should be explained.
(c) I think a better way to implement this would to use it with bond hybrid:
     Use standard bond harmonic for bond types that are "custom"
     Use bond harmonic/mesocnt for bond types that are "C", and drop the "C or custom" param

Steve

akohlmey:
> @akohlmey Why does bond harmonic derive from bond mesocnt - that seems illogical on the face of it.

@sjplimp That is not what @evoyiatzis wrote. You switched it around. Apparently bond style mesocnt is a harmonic bond with a different way to enter its parameters.

@evoyiatzis You are correct. The mesocnt bond style uses all the functionality from bond style harmonic except for how to enter the parameters, so the extract function does not need to be re-implemented. It makes little sense in the context of such a model anyway. It would make more sense to export the unit conversion factor (eunit), but that is a local variable...

sjplimp:
@akohlmey I did switch it around by accident.  What I meant by illogical is that the name of the mesocnt style is not harmonic/mesocnt - see (a) in my 2nd comment.

akohlmey:
> @akohlmey I did switch it around by accident. What I meant by illogical is that the name of the mesocnt style is not harmonic/mesocnt - see (a) in my 2nd comment.

It is not going to be changed in this pull request. You need to contact the MESOCNT style author.
I personally, don't see much harm in this choice: it is in a package, it is only of use with other styles of that package, it has a unique behavior in the input, there is no value for any use beyond this package, so its inheritance is a minor concern.

PR #4357, opened 2024-10-13

akohlmey:
**Summary**

This pull request combines multiple small fixes and corrections.

**Related Issue(s)**

Addresses the issues raised in https://matsci.org/t/inconsistencies-in-the-documentation-of-compute-dihedral-local/58245

Added Howto section on OPLS-AA from https://matsci.org/t/issues-with-opls-force-field-for-ethylene-glycol-simulation/58296/11

Avoids incompatible fix atom/swap inputs as shown in: https://matsci.org/t/atom-swap-issue-with-hybrid-overlay/58345

Disables support for bonded interactions in parallel for stress/mop and stress/mop/profile see
https://matsci.org/t/missing-bond-contributions-from-compute-stress-mop/58455

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- update the documentation to correct examples for compute reduce using local data to include `inputs local`. This is requires since the refactor of compute reduce in LAMMPS version  21 Nov 2023
- let fix nve error out when used with additional argument in case we can safely detect that it was not instantiated from a derived class
- disallow using compute stress/mop and stress/mop/profile with bonded interactions in parallel
- update false positives list for spellchecking the manual
- silence compiler warnings
- add check to fix atom/swap to disallow swaps between different hybrid sub-styles
- remove unneeded check for `Pair::single()` availability from compute property/local
- Howto section on OPLS force fields
- remove dysfunctional filter for outliers from chartviewer class in LAMMPS-GUI
- reposition logwindow cursor before inserting/appending text in LAMMPS-GUI

**Post Submission Checklist**


doc/src/compute_reduce_chunk.rst
```diff
@@ -37,55 +37,57 @@ Description
...
+*min* or *max* options find the minimum or maximum value of the per-atom
+values for each chunk.
+
+Note that only atoms in the specified group contribute to the reduction
+operation.  If the *chunkID* compute returns a 0 for the chunk ID of an
+atom (i.e., the atom is not in a chunk defined by the :doc:`compute
+chunk/atom <compute_chunk_atom>` command), that atom will also not
+contribute to the reduction operation.  An input that is a compute or
+fix may define its own group which affects the quantities it returns.
+For example, a compute with return a zero value for atoms that are not
```
simongravelle:
```suggestion
For example, a compute will return a zero value for atoms that are not
```

doc/src/compute_reduce_chunk.rst
```diff
@@ -37,55 +37,57 @@ Description
...
 
 For each atom, this compute accesses its chunk ID from the specified
-*chunkID* compute.  The per-atom value from an input contributes
-to a per-chunk value corresponding the the chunk ID.
+*chunkID* compute.  The per-atom value from an input contributes to a
+per-chunk value corresponding the the chunk ID.
 
 The reduction operation is specified by the *mode* setting and is
 performed over all the per-atom values from the atoms in each chunk.
 The *sum* option adds the pre-atom values to a per-chunk total.  The
```
simongravelle:
```suggestion
The *sum* option adds the per-atom values to a per-chunk total.  The
```

doc/src/compute_reduce_chunk.rst
```diff
@@ -37,55 +37,57 @@ Description
...
+or atoms in a spatial bin.  See the :doc:`compute chunk/atom
+<compute_chunk_atom>` and :doc:`Howto chunk <Howto_chunk>` doc pages for
+details of how chunks can be defined and examples of how they can be
+used to measure properties of a system.
 
 For each atom, this compute accesses its chunk ID from the specified
-*chunkID* compute.  The per-atom value from an input contributes
-to a per-chunk value corresponding the the chunk ID.
+*chunkID* compute.  The per-atom value from an input contributes to a
+per-chunk value corresponding the the chunk ID.
```
simongravelle:
```suggestion
per-chunk value corresponding the chunk ID.
```

hothello (commented):
Hi Axel, thank you for the prompt commit. Could you please remove the following lines?
```
* :doc:`pair_style <pair_buck>` buck
* :doc:`pair_style <pair_buck>` buck/coul/cut
* :doc:`pair_style <pair_buck>` buck/coul/long
* :doc:`pair_style <pair_lj>` lj/cut
```
Sorry for the inconvenience.

akohlmey:
> Could you please remove the following lines?

done

Bibobu (commented):
I have a remark to the changes proposed by @hothello to the OPLS forcefield.

As far as I know the improper terms are not described by the harmonic improper_style but by cvff. This allows to reproduce the form using only the second term of the dihedral OPLS form with $d=-1$ and $n=2$. See for example [this paper](https://onlinelibrary.wiley.com/doi/10.1002/jcc.20962). I also think this is the one used by the LigParGen server that many people use.

doc/src/Howto_bioFF.rst
```diff
@@ -236,6 +236,32 @@ documentation for the formula it computes.
 
 * :doc:`special_bonds <special_bonds>` dreiding
 
+OPLS-AA
+-------
+
+OPLS-AA is a general force field for atomistic simulation of proteins in
+their native environment, developed at the `Jorgensen group
```
simongravelle:
```suggestion
their native environment, developed by the `Jorgensen group
```

doc/src/Howto_bioFF.rst
```diff
@@ -236,6 +236,32 @@ documentation for the formula it computes.
...
+<https://traken.chem.yale.edu/oplsaam.html>`_ at Purdue University and
+later at Yale University.
+
+This force field is based on atom types mapping to specific functional
```
simongravelle:
```suggestion
This force field is based on atom types mapped to specific functional
```

doc/src/Howto_bioFF.rst
```diff
@@ -236,6 +236,32 @@ documentation for the formula it computes.
...
+groups in organic and biological molecules. The atom type includes an
```
simongravelle:
```suggestion
groups in organic and biological molecules. Each atom type includes an
```

doc/src/Howto_chunk.rst
```diff
@@ -58,28 +58,30 @@ chunk ID for an individual atom can also be static (e.g. a molecule
...
-atom.  This means you can write your own compute or fix to output a
-per-atom quantity to use as chunk ID.  See the :doc:`Modify <Modify>`
-doc pages for info on how to do this.  You can also define a :doc:`per-atom variable <variable>` in the input script that uses a formula to
-generate a chunk ID for each atom.
+:doc:`computes <compute>`, :doc:`fixes <fix>`, and :doc:`variables
+<variable>` to be used to define chunk IDs for each atom.  This means
+you can write your own compute or fix to output a per-atom quantity to
+use as chunk ID.  See the :doc:`Modify <Modify>` doc pages for info on
+how to do this.  You can also define a :doc:`per-atom variable
+<variable>` in the input script that uses a formula to generate a chunk
```
simongravelle, line 66:
```suggestion
<variable>` in the input script that uses a formula that generates a chunk
```

hothello:
> I have a remark to the changes proposed by @hothello to the OPLS forcefield.
> 
> As far as I know the improper terms are not described by the harmonic improper_style but by cvff. This allows to reproduce the form using only the second term of the dihedral OPLS form with d = − 1 and n = 2 . See for example [this paper](https://onlinelibrary.wiley.com/doi/10.1002/jcc.20962). I also think this is the one used by the LigParGen server that many people use.

Hi Germain,

It's a fair point. The OPLS-AA library shipped with Moltemplate was converted from a [TINKER .prm file](https://dasher.wustl.edu/tinker/distribution/params/oplsaa.prm), where the improper angles are described with an Amber-style quadratic term -- see the [Amber manual](https://ambermd.org/doc12/Amber23.pdf) page 287.

In a [recent issue](https://github.com/jewettaij/moltemplate/issues/104) raised by @DrDomenicoMarson, a new OPLS-AA/M library has been created directly from the SI of a recent paper by Prof. Jorgensen, and indeed the improper style is `cvff`.

I have added a note to the list of potential functions.

doc/src/Howto_chunk.rst
akohlmey: I think the original formulation is better here.

akohlmey:
> > I have a remark to the changes proposed by @hothello to the OPLS forcefield.
> > As far as I know the improper terms are not described by the harmonic improper_style but by cvff. This allows to reproduce the form using only the second term of the dihedral OPLS form with d = − 1 and n = 2 . See for example [this paper](https://onlinelibrary.wiley.com/doi/10.1002/jcc.20962). I also think this is the one used by the LigParGen server that many people use.

Some OPLS inputs also use improper style fourier.

> I have added a note to the list of potential functions.

I have generalized and updated the section to be more generic about OPLS.  Using information from my personal experience and from Wikipedia.

sjplimp:
Two comments/Qs:

Is the added restriction on atom/swap b/c the code does not (or cannot easily) do that correctly?
I don't see a conceptual reason for why swapping between hybrid sub-styles should not be allowed.
In some cases two sub-styles are the same pair styles, just with different params.

I don't think of OPLS as a bio FF - possilby OPLS should be its own howto page, or the Howto_bioFF name of that section should be generalized?

akohlmey:
> Two comments/Qs:
> 
> Is the added restriction on atom/swap b/c the code does not (or cannot easily) do that correctly?

Yes. The problem is with the neighbor lists. For hybrid pair styles, you have skip-lists. If you swap atoms (or rather their types) outside the confines of the skip list, you are in trouble, because the parameters are not available. For example, if you have eam for atom types 1 and 2 and lj/cut of type 3 and you want to swap types 2 and 3, then suddenly you are trying to get the EAM parameters for atom type 3, but that had be set to NULL and thus you will get a segfault since the map file has a -1. So both atom types that you want to swap have to have the same combination of sub-styles.

> I don't see a conceptual reason for why swapping between hybrid sub-styles should not be allowed. In some cases two sub-styles are the same pair styles, just with different params.

> I don't think of OPLS as a bio FF - possilby OPLS should be its own howto page, or the Howto_bioFF name of that section should be generalized?

It is. The latest versions explicitly mention proteins and DNA. Its core is derived from Amber, but then its non-bonded parameters have not only optimized on biomolecular structures (like Amber, CHARMM, GROMOS etc.) but also on general solvents. For that reason OPLS is very often used for general mixtures or solvents because the others fail there and the others are more often used for biomolecules, because more people already do. So people that enter the simulation business will follow what others have done.

sjplimp:
@akohlmey Thanks for the clarifications.

evoyiatzis (approved):
@akohlmey thanks for disabling the parallel execution of the stress calculations

doc/src/Howto_bioFF.rst
```diff
@@ -236,6 +236,40 @@ documentation for the formula it computes.
...
+OPLS (Optimized Potentials for Liquid Simulations) is a general force
+field for atomistic simulation of organic molecules in solvent.  It was
+developed by the `Jorgensen group
+<https://traken.chem.yale.edu/oplsaam.html>`_ at Purdue University and
+later at Yale University.  Multiple versions of the OPLS parameters
+exist for united atom representations (OPLS-UA) and for all-atom
+representations (OPLS-AA).
+
+This force field is based on atom types mapped to specific functional
+groups in organic and biological molecules. Each atom includes an
```
jtclemm: a static

src/EXTRA-COMPUTE/compute_stress_mop.cpp
```diff
@@ -133,27 +133,28 @@ ComputeStressMop::ComputeStressMop(LAMMPS *lmp, int narg, char **arg) : Compute(
...
-  if (domain->triclinic != 0)
-    error->all(FLERR, "Compute stress/mop is incompatible with triclinic simulation box");
-
   // 2D and pressure calculation in the Z coordinate
   if (domain->dimension == 2 && dir == Z)
     error->all(FLERR, "Compute stress/mop is incompatible with Z in 2d system");
 
+  // orthogonal simulation box
+
+  if (domain->triclinic != 0)
```
jtclemm: This error check would be better moved into `init()`. Currently, I can issue a `change_box all triclinic` command after the compute is defined and bypass this check in examples/packages/mop/in.compute_stress_mop.

src/EXTRA-COMPUTE/compute_stress_mop_profile.cpp
```diff
@@ -134,14 +134,14 @@ ComputeStressMopProfile::ComputeStressMopProfile(LAMMPS *lmp, int narg, char **a
   // 3D only
 
   if (domain->dimension == 2 && dir == Z)
-    error->all(FLERR, "Compute stress/mop/profile incompatible with Z in 2d system");
+    error->all(FLERR, "Compute stress/mop/profile is incompatible with Z in 2d system");
 
   // orthogonal simulation box
 
   if (domain->triclinic != 0)
```
jtclemm: Ditto

src/compute_dihedral_local.cpp
```diff
@@ -75,47 +76,45 @@ ComputeDihedralLocal::ComputeDihedralLocal(LAMMPS *lmp, int narg, char **arg) :
...
   if (nvar) {
-    if (!setflag)
-      error->all(FLERR,"Compute dihedral/local variable requires a set variable");
+    if (!setflag) error->all(FLERR, "Compute dihedral/local variable requires a set variable");
     for (int i = 0; i < nvar; i++) {
       vvar[i] = input->variable->find(vstr[i]);
-      if (vvar[i] < 0)
-        error->all(FLERR,
-                   "Variable name for copute dihedral/local does not exist");
+      if (vvar[i] < 0) error->all(FLERR, "Variable name for copute dihedral/local does not exist");
```
jtclemm: copute -> compute

jtclemm (approved):

PR #1760, opened 2019-11-05
added internal fix dummy command to enable more control of fix ordering

sjplimp:
**Summary**

Add a fix dummy command which other styles can use to create a placeholder fix (early in the input script) which is replaced later with the desired fix.  This enables the ordering of fixes to correspond to the listing of commands in the input script.  This addresses a bug reported where granular history was not preserved correctly with fix deform or fix balance when they migrated atoms w/out FixNeighHistory having been invoked soon enough to assign history to atoms.

**Related Issues**

N/A

**Author(s)**

Steve

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

Should not, but regression testing should be done.

**Implementation Notes**

_Provide any relevant details about how the changes are implemented, how correctness was verified, how other features - if any - in LAMMPS are affected_

**Post Submission Checklist**

_Please check the fields below as they are completed **after** the pull request has been submitted. Delete lines that don't apply_

- [ ] The feature or features in this pull request is complete
- [ ] Licensing information is complete
- [ ] Corresponding author information is complete
- [ ] The source code follows the LAMMPS formatting guidelines
- [ ] Suitable new documentation files and/or updates to the existing docs are included
- [ ] The added/updated documentation is integrated and tested with the documentation build system
- [ ] The feature has been verified to work with the conventional build system
- [ ] The feature has been verified to work with the CMake based build system
- [ ] A package specific README file has been included or updated
- [ ] One or more example input decks are included

**Further Information, Files, and Links**

_Put any additional information here, attach relevant text or image files, and URLs to external sites (e.g. DOIs or webpages)_

PR #4359, opened 2024-10-21
Adding new force options to BPM package

jtclemm:
**Summary**

This PR includes a new optional multibody force contribution to the BPM/spring bond style and an anharmonic term to pair BPM/spring. The multibody term is designed to allow users to simulate non-trivial Poisson's ratios. The anharmonic force can help minimize contact overlaps.

**Related Issue(s)**

NA

**Author(s)**

Joel Clemmer (SNL)

**Licensing**

**Backward Compatibility**

NA

**Implementation Notes**

NA

**Post Submission Checklist**

NA


akohlmey:
@jtclemm FYI, I've added some changes:
- delete the internally created fix in destructor (if present) to avoid leakage
- implement the write_data() method for the bond style (it seemed trivial enough)
- minor programming style changes
- update to upstream

Please check that I didn't break anything.

akohlmey:
@jtclemm I noticed some memory corruption and an unexpected error message from the automatic "quick" regression test. The issue was with in.bpm.impact.spring. This should be resolved with the changes in commit 53b1ba61317e6ad552b2e9b1acdcbfedfd92f639

Update: some more small tweaks and were added for more consistent behavior.

sjplimp (approved):

jtclemm:
@akohlmey Thanks! Let me take a look at the changes. 

One initial comment. I don't think reading/writing bond coefficients from a datafile is possible with the bond style since it initializes a fix in the constructor. I'll check to confirm.

However, maybe the pair style still can still write its coefficients. I'll look into that.

akohlmey:
> One initial comment. I don't think reading/writing bond coefficients from a datafile is possible with the bond style since it initializes a fix in the constructor. I'll check to confirm.

That is irrelevant. The data file section is the Bond Coeffs section which will be fed line-by-line to the bond_coeff command. I don't see why this should not work. You just need to account for the different number of arguments depending on the bond style settings.

jtclemm:
> That is irrelevant. The data file section is the Bond Coeffs section which will be fed line-by-line to the bond_coeff command. I don't see why this should not work. You just need to account for the different number of arguments depending on the bond style settings.

I remember this was an issue I had when I first wrote the bond style. The problem is that fixes can't be defined before read data. I.e. if the bond_style is defined first:
    "ERROR: Fix command before simulation box is defined"
and if read data is called first:
    "ERROR: Must define bond_style before Bond Coeffs"
Not clear to me how this could be resolved.

> I noticed some memory corruption and an unexpected error message from the automatic "quick" regression test. The issue was with in.bpm.impact.spring. This should be resolved with the changes in commit

Hmm moving the definition of `id_fix_bond_history` to the parent class will prevent one from defining multiple bpm bond styles, as with bond hybrid. Is there a particular problem with having the string defined in the child classes?

akohlmey:
> I remember this was an issue I had when I first wrote the bond style. The problem is that fixes can't be defined before read data. I.e. if the bond_style is defined first: "ERROR: Fix command before simulation box is defined" and if read data is called first: "ERROR: Must define bond_style before Bond Coeffs" Not clear to me how this could be resolved.

Isn't that what the fix DUMMY instances are for?
Can you provide me with some example inputs to test this issue with?
Also for the other issue mentioned below?
I had no problem with the examples in examples in examples/bpm/impact
 
> > I noticed some memory corruption and an unexpected error message from the automatic "quick" regression test. The issue was with in.bpm.impact.spring. This should be resolved with the changes in commit
> 
> Hmm moving the definition of `id_fix_bond_history` to the parent class will prevent one from defining multiple bpm bond styles, as with bond hybrid. Is there a particular problem with having the string defined in the child classes?

There was a conflict there before. If you want unique names for hybrid, we need to append "instance_me" to the fix id (in fact for all of them), then it doesn't matter where the name is defined. The old flow of control was a bit rather confusing.
I can revise my changes this way. Any particular reason these using instance_total instead of instance_me?

akohlmey:
The following change should work around the fix defined before box issue:
```
 diff --git a/src/modify.cpp b/src/modify.cpp
  index 1a9c056027..a7d0000bc0 100644
  --- a/src/modify.cpp
  +++ b/src/modify.cpp
  @@ -827,7 +827,7 @@ Fix *Modify::add_fix(int narg, char **arg, int trysuffix)
     const char *exceptions[] =
       {"GPU", "OMP", "INTEL", "property/atom", "cmap", "cmap3", "rx",
        "deprecated", "STORE/KIM", "amoeba/pitorsion", "amoeba/bitorsion",
  -     nullptr};
  +     "DUMMY", nullptr};
     // clang-format on
   
     if (domain->box_exist == 0) {
```

jtclemm:
Here are the scripts I used to test data files, running 1 then 2
[in.part1.txt](https://github.com/user-attachments/files/17496732/in.part1.txt)
[in.part2.txt](https://github.com/user-attachments/files/17496731/in.part2.txt)

I also didn't view writing bond coeff to a data file as important since bond history is only transferred across restart files - like neighbor history for granular pair styles. I think the doc files may mention that bond styles write no data to data files... I'll check this too.

jtclemm:
> Isn't that what the fix DUMMY instances are for?

The DUMMY instances were a hack to try and shoe horn history fixes (both neighbor and bond) in before any other fixes that might trigger irregular comm calls (like fix deform). Otherwise if an irregular comm call was made, it wasn't guaranteed that fix neigh/history would first call it's `pre_exchange()` method and pack up the most recent history data. 

Introduced here: https://github.com/lammps/lammps/pull/1760

jtclemm:
> I had no problem with the examples in examples in examples/bpm/impact

The rheo/oxidation example uses a hybrid bond style. Although it currently works because the child class still overrides 
Edit:, adding link: https://github.com/lammps/lammps/blob/e214f60d4fd9725f5c6e4365590fa734261f84f9/src/RHEO/bond_rheo_shell.cpp#L55

> There was a conflict there before. If you want unique names for hybrid, we need to append "instance_me" to the fix id (in fact for all of them), then it doesn't matter where the name is defined. The old flow of control was a bit rather confusing. I can revise my changes this way. Any particular reason these using instance_total instead of instance_me?

I remember there was a significant discussion over using `instance_me` in the bond styles and the ultimate decision was to avoid it. It looks like the discussion starts somewhere around here: https://github.com/lammps/lammps/pull/2906#issuecomment-1045241387, but I haven't read through it yet.


> The following change should work around the fix defined before box issue:

That does partially work, but bond bpm defines the BOND_HISTORY fix in settings so that still triggers an error. It defines this in settings because for multiple instances of history to work, they need to know about one another to handle bond breaking. So the order is:
settings -> define history
init_style -> look up other histories
~~Maybe this can be adjusted by moving something to `init()`. But I'll have to first familiarize myself with how `init()` is different than `init_styles`.~~ Nevermind, I now remember that `init()` calls `init_style()`. Like with dummy, a lot of these choices boil down to how to set the order of operations to ensure things can't break.

akohlmey:
@jtclemm This is a real "mind twister(tm)" problem. I will have to sit back and think about it and map out the different scenarios. Unfortunately, I am currently distracted by urgent Temple HPC work and thus it may take a couple of days before I can make a suggestion. It may be simpler to roll back the write_data changes and forget about it.
But I still have a nagging feeling that this will just stop us from identifying a proper solution. Restarting under these circumstances is quite challenging, I had to deal with a similarly complex situation some time ago with fix alchemy. It took me a long time to make a choice that work well enough.

jtclemm:
> Unfortunately, I am currently distracted by urgent Temple HPC work and thus it may take a couple of days before I can make a suggestion. It may be simpler to roll back the write_data changes and forget about it.
> But I still have a nagging feeling that this will just stop us from identifying a proper solution.

I'm in no particular rush to get this PR merged, so I'm content either rolling back changes or waiting. If we do wait, here's a reminder of the two pending issues:
1) How/whether to support bpm bond styles writing coeff to data files - resolving this means moving fix declarations later in the bond setup (like BOND_HISTORY) or adding exceptions to modify.cpp (like DUMMY)
2) How to allow multiple bpm bond styles to create different fix IDs (instance_me vs. instance_total vs. manually defining in child classes).

stanmoore1 (approved):

akohlmey:
@jtclemm I hope that everything is back in working order and the issues with passing undefined fix id strings around resolved. Data file writing is removed as it is too complex with the chain for fixes creating fixes (since bond.cpp enables it by default is must be disabled with writedata = 0, similar to single_enable for pair styles). Otherwise and empty section is written that gives trouble when reading.

I had to make some small tweak for restarting since Bond::settings() is not run when using reading from a restart, so that the bond history fix is created. Without I would get segfaults or not all restart info would be reloaded.

jtclemm:
Thanks Axel, I just double checked that data file still operate as they used to (nothing is written, nothing is expected). I think this PR is good to go from my perspective (keeping in mind that it might be good to revisit how bond/fix IDs and coeff are written/read from restart/data files in the future).

akohlmey:
> keeping in mind that it might be good to revisit how bond/fix IDs and coeff are written/read from restart/data files in the future

I don't think that both will be possible. There are too many diverging constraints and requirements.

PR #4369, opened 2024-11-04
Python interface updates

rbberger:
**Summary**

- deprecates the PyLammps interface
- adds the `cmd` property to `lammps.lammps`, which maintains the wrapper from PyLammps for writing Python function calls that get converted into `lammps.command` calls.
- automatically wraps Python functions and lambdas in a way that make them callable from fix python/invoke, passing them a `lammps.lammps` instance instead of a pointer.
- adds new `lammps.ipython` package to replace some of the Jupyter notebook functionality,  including adding a magic to suppress/capture LAMMPS output.

**Related Issue(s)**

#4329

**Author(s)**

@rbberger 

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**


doc/src/Howto_pylammps.rst
```diff
@@ -1,564 +1,6 @@
...
-Feedback and Contributing
--------------------------
-
-If you find this Python interface useful, please feel free to provide feedback
-and ideas on how to improve it to Richard Berger (richard.berger@outlook.com). We also
-want to encourage people to write tutorial style IPython notebooks showcasing LAMMPS usage
-and maybe their latest research results.
+The PyLammps interface is deprecated and will be removed in a future release of
+LAMMPS. As such, the PyLammps version of this tutorial has been removed and is
+replaced by the :doc:`Python_head`.
```
megmcca, line 6: we could add the version of the last LAMMPS release that supports PyLammps here as well (presumably the last stable release).

megmcca (approved):
this looks great! consolidating how users interact with LAMMPS through Python will make things much easier to learn.  nice and clear updates to documentation and iPython too.

~~my only optional suggestion is to add a line in the `Howto_pylammps.rst` that informs readers about the last stable LAMMPS version that supports it, for example in case a user needs to test old code before porting it.~~ <-- pre-coffee brain! as @akohlmey mentioned we can mark the final version in the docs of the summer 2025 "important legacy" release.

doc/src/Howto_pylammps.rst
akohlmey: @megmcca please note the comment `will be removed in a future release of LAMMPS`, so it is still functional. In fact, all the unit tests pass and none has been removed, yet. I expect we will pull the plug *after* the *next* stable release where a bunch of other disruptive changes will take place (see #4344). The Summer 2025 stable release will become another "important legacy" release https://www.lammps.org/download.html

akohlmey:
> pre-coffee brain!

@megmcca no big deal. I appreciate your commitment to LAMMPS to even *do* a review on a Sunday morning. :smile:

akohlmey:
@rbberger FYI, I just tested the howto and added some more links and explanations, added some separators for better readability, and removed references to PyLammps functions or properties not supported in the lammps python module.

I think the "eval()" functionality may be something worth adding to the C library interface. This way should be issued on *all* MPI ranks of the LAMMPS instance, same as when the input file is processed and immediate variables are expanded. This will not result in a hang if variable functions like `count()` are used that require collective MPI calls.

akohlmey:
> I think the "eval()" functionality may be something worth adding to the C library interface.

Check out commit dce1231052f7329a4fa211d51d5ca655266dbd9c

akohlmey (approved):

Issue #4360, opened 2024-10-22
[Feature Request] Add library API to schedule compute time steps


rbberger:
**Summary**

Add library API for calling `addstep` on computes.

**Detailed Description**

Some computes like `stress/atom` require the virial to be computed in the current time step.

When accessing such a compute via the library interface (`extract_compute`) at an arbitrary time step (nothing that matches up with either thermo or dump), there is currently no way to add such a step to the compute (via `addstep`).

Without the added step, the integration loop will not schedule the virial computation. Since we would like to run our Python analysis independent of both thermo and dump time steps, we would need library API methods to accomplish this.

@sjplimp what do you think?

**Further Information, Files, and Links**

sjplimp:
I think it would be good to enable this via the lib interface.
Can you explain your use case in more detail (are you not doing integration steps at all)?
  E.g.  I think this would work fine if you were doing "run 0" repeatedly.
Also, what library calls are you proposing?

Steve

rbberger:
@sjplimp this is a large simulation with a online local Python analysis happening at some interval that is independent of thermo and dumps disabled or very infrequent due to the large data volume.

sjplimp:
I think what I don't understand is this:
The Python lib interface code wants to compute the stress/atom for a snapshot of atoms on a particular timestep.
(a) Where is it getting this snapshot from?
To do this it has to invoke the pair style on the snapshot (and build a neighbor list, etc).
(b) Assuming it isn't doing normal step-to-step time integration, why doesn't it do this via a "run 0"
     command, which would build the neigh list, invoke the pair style, and set the addstep flags to ensure
     the pair style also tallied virial info, so that compute stress/atom could be invoked after the "run 0"

Also I don't understand what you mean by an "online local Python analysis" - possibly you mean it is running
at the same time as the large simulation is running?  If that is the case, my (a) question is relevant.
And am I correct that the Python code still needs to invoke the pair style on the snapshot?

rbberger:
The Python code runs as part of the regular integration loop at fixed intervals, at end_of_step (fix python/invoke). That's the "online" part. Pair style computation has happened and the neighbor list is accessible. What is missing is the necessary addstep for compute stress/atom so that the virial tally gets scheduled.

So the library API I'm looking for is being able to call addstep for a specific compute. Each Python invocation would then add the next scheduled step. fix invoke/python can't do this since it doesn't know which computes are needed.

[...]

PR #4380, opened 2024-11-12
Python and library API changes for computes

rbberger:
**Summary**

- ensures that fix python/invoke registers computes' next timestep
- add library API for clearing and adding new time steps
- update interfaces for fortran, swig, and plugins accordingly
- add missing documentation

**Related Issue(s)**

#4360 

**Author(s)**

@rbberger 

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**


akohlmey (changes requested):
@rbberger I tried to resolve the merge conflict and looked into propagating the library interface extension to all the other related interfaces (Fortran, plugin, SWIG), but ran into some issues due to the use of the "bigint" versus "int" type.

Even though it is counter-intuitive, it would be easier to make this portable by making the nextstep argument of the compute_addstep variants a (void?) pointer. The correct size could easily be recovered *inside* the function from the size defines and we can extract the size of an bigint though the interface for the calling code and thus call the function with the correct argument (or just exit with a runtime error, if there is a mismatch). That would also allow to get rid of the ifdef in the function prototype in library.h and simplify extracting the documentation.

Also, for a person that occasionally programs LAMMPS itself, the names of the library functions are a bit unexpected. I would have expected those to be called `lammps_clearstep_compute()` and `lammps_addstep_compute()` and `lammps_addstep_compute_all()` to align better with the corresponding methods in the Modify class.

rbberger:
@akohlmey did the changes you mentioned. don't have anything against them.

akohlmey:
> @akohlmey did the changes you mentioned. don't have anything against them.

@rbberger Thanks a lot. I appreciate it. I will look into it when I have a bit more time. Currently wrestling with GPFS troubles when updating from CentOS 7 to MultiLinux (aka Liberty Linux) 7.

akohlmey (approved):

akohlmey:
@rbberger have 90% of the missing changes done. Am lacking a good idea for how to test python and fortran. Do you have something, so you can check that I didn't mess up something that you need?

If you want, we can merge "as is". I will need to do a general "sync" of the various types of interfaces build on top of the LAMMPS C-style library interface at some time in the future.

akohlmey:
The two missing Fortran interface functions are added.

src/PYTHON/fix_python_invoke.cpp
```diff
@@ -91,6 +94,8 @@ int FixPythonInvoke::setmask()
 
 void FixPythonInvoke::end_of_step()
 {
+  modify->clearstep_compute();
```
sjplimp, line 114: Can you add a comment line or two to explain why you are wrapping the callback with clearstep/addstep?
I.e. in case the Python method accesses compute info that depends on pair energy/virial data being calculated on specific timesteps.   Ditto at the end of the constructor.

sjplimp: And ditto in post_force().

sjplimp:
@rbberger I believe there is a subtle issue with the use of addstep_compute_all() in fix python/invoke being only in the constructor.

If the input script logic is this:

fix ... python/invoke 100 post_force ,,,     # called every 100 steps
minimize ...   # runs for 125 steps
run 10000  # starts at step 125

then a compute called back to by python/invoke on step 200 will not have its eflag/vflag timestep trigger set.
b/c FixPyInvoke::post_force() wlll not have been called on step 100.

The solution is to do like fix ave/time does and add an init() function to FixPyInvoke which makes a call to addstep_compute_all(nextvalid()).  This will be called on step 125 and set the trigger for step 200.
For the added init() code, you should also add a comment like the one in FixAveTime::init() to clarify why this is being done.

On a related note, the fix python/invoke doc page does not have this section: 
Restart, fix_modify, output, run start/stop, minimize info
which all fixes should typically have.

In this case, its content should be the exact same text as the fix nve doc page.
Which would explain (relevant to this issue) that fix python/invoke is not invoked during energy minimization.

Steve

akohlmey:
@rbberger please assign the pull request to me, when Steve's concerns are addressed and I will merge it.

src/PYTHON/fix_python_invoke.cpp
rbberger: Added comments.

rbberger:
@sjplimp made the suggested changes.

sjplimp (approved):
thanks

PR #4383, opened 2024-11-14

### LINKED ISSUE #4384: [BUG] "fix adapt/fep" does not adapt the pair coefficients derived from mixing rules
**Summary**

"fix adapt/fep" does not adapt pair coefficients derived from mixing rules

LAMMPS (2 Aug 2023 - Update 3)
CentOS Stream release 9
AMD EPYC 9124 16-Core Processor

**Expected Behavior**

If some pair coefficients are derived from mixing rules ( but not explicitly specified by pair_coeff), they could be adapted by fix adapt/fep.

**Actual Behavior**

If some pair coefficients are derived from mixing rules ( but not explicitly specified by pair_coeff), they could NOT be adapted by fix adapt/fep. This behavior is not written in the [manual ](https://docs.lammps.org/fix_adapt_fep.html )

**Steps to Reproduce**

Here is some simplified scripts that could possibly reflect the problem I met in my simulation:

pair_style hybrid/overlay lj/cut/soft 1 0.5 12.0
pair_coeff 1 1 lj/cut/soft 0.886 2.933 1
pair_coeff 2 2 lj/cut/soft 0.886 2.933 1
variable timestep equal ramp(0,100)
fix ADAPT all adapt/fep 1  pair lj/cut/soft lambda 1 2 v_timestep

In this way, the energy between the particles of type1 and type2 will not be adapted during the simulation. (This energy is non-zero)

If you add "pair_coeff 1 2 lj/cut/soft 0.886 2.933 1" before the "fix adapt/fep", the energy between type 1 and type 2 will then be adapted during the simulation.

The [manual ](https://docs.lammps.org/fix_adapt_fep.html ) does say that pair_coeff need to be used before fix adapt/fep. But I think it just implies that the coefficients derived from mixing rules should also be adapted. Or the manual should be clearer to specify that coefficients derived from mixing rules should be explicitly specified using pair_coeff before "fix adapt/fep".

akohlmey:
**Summary**

**Related Issue(s)**

Addresses the invalid angle issue mentioned here: https://matsci.org/t/stillinger-weber-error/58812/5
Implements the suggestion from here: https://matsci.org/t/maintaining-positional-uncertainty-for-changing-temperatures/58898/2
Closes #4384 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- Handle the issue of invalid forces for 180 degree angles in pair style sw/table/angles due to rounding and division by zero
- Updates the documentation for fix adapt and fix adapt/fep to require explicit pair_coeff commands for all parameters that are supposed to be changed. Pair coefficients derived from mixing are not suitable.
- Avoid access to uninitialized per-atom data with fix reaxff/species.
- Trigger citation reminder for label maps only when they are used.
- Avoid dereferencing reaxff object when pair style reaxff is not used
- Accept fix qtpie/reaxff also for pair style reaxff/omp
- Add force-style unittest for fix qtpie/reaxff with pair style reaxff and reaxff/omp.
- Use utils::open_potential() to read qtpie/reaxff gaussian exponent file.
- Add support for specifying the fix spring/self spring constant as equal or atom style variable.

**Post Submission Checklist**


akohlmey:
@jrgissing this PR contains the changes that I suggested to @jtclemm to make the labelmap code promote the labelmap paper less aggressively, i.e. only when the maps are actually used. If you would like this to be done differently, you should be able to pull and modify the PR branch yourself (best via the gh command).

src/command.h
```diff
@@ -20,7 +20,7 @@ namespace LAMMPS_NS {
 
 class Command : protected Pointers {
  public:
-  Command(class LAMMPS *lmp) : Pointers(lmp) {};
+  Command(class LAMMPS *lmp) : Pointers(lmp), copymode(0) {};
```
akohlmey, line 23: @stanmoore1 we both overlooked this one. Fortunately, Coverity Scan reported it after its regular Thursday night check on the develop branch...

jrgissing:
> @jrgissing this PR contains the changes that I suggested to @jtclemm to make the labelmap code promote the labelmap paper less aggressively, i.e. only when the maps are actually used. If you would like this to be done differently, you should be able to pull and modify the PR branch yourself (best via the gh command).

Yes, I took a look at this and it seems to be working as it should, now. thanks for fixing

akohlmey:
@stanmoore1 @alphataubio FYI, I just added support for specifying the fix spring/self spring constant as a variable. This is missing KOKKOS support, though. Can you look into implementing it?

There are unit tests (using just a constant value, so the results are the same as for the constant value). But to run it with KOKKOS+OpenMP you have to remove the kokkos_omp entry in skip_tests.

alphataubio:
> specifying the fix spring/self spring constant as a variable

anything that's using variable.cpp is not really worth running in KOKKOS because it has to go back  and forth to the host to compute the variable, especially if it's at every timestep.

unless someone implements variable_kokkos.cpp to make variable stuff KOKKOS aware, and sorry that will not be me anytime soon

stanmoore1:
Yes the insides of variable class will likely never be ported to KOKKOS. For SPARTA I put sync/modify inside the variable class, but for now I think we just wrap the call with sync all, modify all which is a trivial change.

akohlmey:
For the time being, I added a check so that the KOKKOS version will error out cleanly for non-constant K. Better than some random crash.

akohlmey:
> anything that's using variable.cpp is not really worth running in KOKKOS because it has to go back and forth to the host to compute the variable, especially if it's at every timestep.

That very much depends on the variable. Simple equal style variables that just have some simple evaluation using only internal info and no references to fixes, computes or atoms would really only require to transfer the new K value. For spring/self a common case would be to couple the value of K to ramp that changes with the timestep.

stanmoore1:
>Simple equal style variables that just have some simple evaluation using only internal info and no references to fixes, computes or atoms would really only require to transfer the new K value.

The problem is we don't know if the equal style variable is computed by summing over positions or forces or some other atom array, so everything has to be transferred to the CPU just in case.

akohlmey:
> Yes the insides of variable class will likely never be ported to KOKKOS. For SPARTA I put sync/modify inside the variable class, but for now I think we just wrap the call with sync all, modify all which is a trivial change.

The "kstyle" class member indicates whether it is CONSTANT, EQUAL, or ATOM style. You can just call the CPU version for the latter two cases. For the current version you would have to wrap the fix command into "suffix off / suffix on" for variable K so that it the CPU version with suffixes.

stanmoore1:
@akohlmey yes true. That would be really easy too.

akohlmey:
> The problem is we don't know if the equal style variable is computed by summing over positions or forces or some other atom array, so everything has to be transferred to the CPU just in case.

True, and I don't even want to think about what it would take to flag whether a variable evaluation is "trivial" or "complex".

src/REAXFF/fix_qtpie_reaxff.cpp
```diff
@@ -199,7 +199,9 @@ void FixQtpieReaxFF::pertype_parameters(char *arg)
   if (comm->me == 0) {
     gauss_exp[0] = 0.0;
     try {
-      TextFileReader reader(gauss_file,"qtpie/reaxff gaussian exponents");
+      FILE *fp = utils::open_potential(gauss_file, lmp, nullptr);
+      if (!fp) throw TokenizerException("Fix qtpie/reaxff: coul not open gauss file", gauss_file);
```
jtclemm: Typo: "coul"

src/fix_spring_self.cpp
```diff
@@ -33,8 +35,7 @@ using namespace FixConst;
 /* ---------------------------------------------------------------------- */
 
 FixSpringSelf::FixSpringSelf(LAMMPS *lmp, int narg, char **arg) :
-  Fix(lmp, narg, arg),
-  xoriginal(nullptr)
+  Fix(lmp, narg, arg), xoriginal(nullptr), kstr(nullptr), kval(nullptr)
```
jtclemm, line 38: Should `kval` be in the destructor?

src/REAXFF/fix_qtpie_reaxff.cpp
akohlmey: fixed

src/fix_spring_self.cpp
akohlmey: yes, as should be kstr.

jtclemm (approved):

PR #4390, opened 2024-11-21
Add support to region plane for having the normal vector read from variables

evoyiatzis:
**Summary**

Add support to region plane for having the normal vector read from variables

**Related Issue(s)**

No related issue

**Author(s)**

E. Voyiatzis at NovaMechanics Ltd (email [evoyiatzis@gmail.com](mailto:evoyiatzis@gmail.com))

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/src/region_plane.cpp
+++ b/src/region_plane.cpp
@@ -22,13 +22,14 @@
 
 using namespace LAMMPS_NS;
 
+enum { CONSTANT, VARIABLE };
+
 /* ---------------------------------------------------------------------- */
 
-RegPlane::RegPlane(LAMMPS *lmp, int narg, char **arg) :
-    Region(lmp, narg, arg), xstr(nullptr), ystr(nullptr), zstr(nullptr)
+RegPlane::RegPlane(LAMMPS *lmp, int narg, char **arg) : Region(lmp, narg, arg),
+    xstr(nullptr), ystr(nullptr), zstr(nullptr), nxstr(nullptr),
+    nystr(nullptr), nzstr(nullptr)
 {
-  xvar = yvar = zvar = 0.0;
-
   options(narg - 8, &arg[8]);
 
   if (utils::strmatch(arg[2], "^v_")) {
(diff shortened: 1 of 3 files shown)


src/region_plane.cpp
```diff
@@ -61,15 +65,41 @@ RegPlane::RegPlane(LAMMPS *lmp, int narg, char **arg) :
...
   if (varshape) {
     variable_check();
     RegPlane::shape_update();
   }
 
-  normal[0] = xscale * utils::numeric(FLERR, arg[5], false, lmp);
-  normal[1] = yscale * utils::numeric(FLERR, arg[6], false, lmp);
-  normal[2] = zscale * utils::numeric(FLERR, arg[7], false, lmp);
-
   // enforce unit normal
```
sjplimp, line 103: I think this piece of code will throw an error if all 3 normal args are variables?
B/c length of current normal = zero?

src/region_plane.cpp
```diff
@@ -174,6 +207,22 @@ void RegPlane::shape_update()
...
+
+  if (nxstyle == VARIABLE) normal[0] = xscale * input->variable->compute_equal(nxvar);
+
+  if (nystyle == VARIABLE) normal[1] = yscale * input->variable->compute_equal(nyvar);
+
+  if (nzstyle == VARIABLE) normal[2] = zscale * input->variable->compute_equal(nzvar);
+
+  // enforce unit normal
+
+  double rsq = normal[0] * normal[0] + normal[1] * normal[1] + normal[2] * normal[2];
```
sjplimp, line 221: The common use case will be that none of the normal components are variables.  So it would be better if the conversion to a unit vector were done only once in that case (as in the prior code in the constructor).  Only do it here if one or more components are variables.

evoyiatzis: I included a check and if all components are constants then the function will return and avoid renormalization

src/region_plane.cpp
```diff
@@ -174,6 +207,24 @@ void RegPlane::shape_update()
...
+  if (nystyle == VARIABLE) normal[1] = yscale * input->variable->compute_equal(nyvar);
+
+  if (nzstyle == VARIABLE) normal[2] = zscale * input->variable->compute_equal(nzvar);
+
+  if ((nxstyle == CONSTANT) && (nystyle == CONSTANT) && (nzstyle == CONSTANT)) return;
+
+  // enforce unit normal
+
+  double rsq = normal[0] * normal[0] + normal[1] * normal[1] + normal[2] * normal[2];
+  if (rsq == 0.0)
```
sjplimp, line 222: I think there is a logic problem with this normalization of the norm vector.  If some components are CONSTANT and some are VARIABLE, then the CONSTANT components are rescaled and will thus not remain constant at each timestep.

I think you either need to have a 2nd copy of the norm vector (initial and normalized) so that the initial copy can persist.  Or you need to force the user to make all components constant or all components variable.  Not sure what your intent is.

akohlmey: > Or you need to force the user to make all components constant or all components variable

Since this is just *one vector*, this requirement seems very reasonable to me.

evoyiatzis:
@sjplimp I will contact you via email to get your thoughts on the logic and the user cases most likely to be encountered 

In my case, two components are defined by variables and the third one is constant but zero.  My initial thoughts were limited to such a setting.

PR #3465, opened 2022-09-29
Packed Clebsch-Gordon look-up tables for SNAP/Kokkos

weinbe2:
**Summary**

This PR provides marginal performance improvements to kernels relevant for Clebsch-Gordon contractions (`ComputeZi`, `ComputeYi`, `ComputeYiFromZlist`) by refactoring the index look-up table `idxz` into the minimum number of 128-bit loads (three, three, and one, respectively). It also embeds the values stored in `idx_cgblock`, avoiding the extra load that had a serial dependency on loading `j`, `j1`, and `j2` from `idxz`. More details are below.

Performance optimizations range from 2-5% for “linear” snap, as a function of (reasonable) problem sizes and “2J” value, more details below. Very tiny atom counts (<= 512) and quadratic SNAP as a whole sees less of a speed-up.

**Related Issue(s)**

Iterative improvement on #2599 , #2905 , other Kokkos/SNAP PRs

**Author(s)**

Evan Weinberg (NVIDIA). eweinberg [at] nvidia.com preferred, evansweinberg [at] gmail.com for "long-lived".

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

No issues.

**Implementation Notes**

In the current `develop` branch (as of https://github.com/lammps/lammps/commit/afc35aa7b0db1a72b8a4bd0ba9d88f356f7aa490 ), the look-up table for Clebsch-Gordon contractions `idxz` stores its values a discrete 32-bit integers. There is also a separate data structure, `idx_cgblock`, that stores offsets into the Clebsch-Gordon coefficient data structure `cglist`. In total, this corresponds to 4, 10, and 11 `int32` loads for `ComputeYiFromZlist`, `ComputeZi`, and `ComputeYi`, respectively. In addition, there is a serial dependency between loading j/j1/j2 and loading an offset from `idx_cgblock`.

This PR refactors this to pack these values into three `int4` data structures. In this case, there are now 1, 3, and 3 128-bit transactions for `ComputeYiFromZlist`, `ComputeZi`, and `ComputeYi`, respectively. While this does lead to “wasted” bits on face value, it reduces the number of memory transactions and improves cache accesses, leading to what should be a net improvement. Further, since it packs the value from `idx_cgblock` offset, it removes a serial dependency.

[...]

PR #2599, opened 2021-02-11
Kokkos SNAP cleanup/optimizations --- templating, Kokkos::LaunchBounds, tuning

weinbe2:
**Summary**

Note: this PR is still WIP because I'm doing performance checks across a wider variety of SNAP sample inputs, as opposed to the spot tests I've done thus far (2J8, 2J14 Tungsten benchmarks). The core of the work is complete. I don't expect any surprises. 

This PR is a "miscellaneous" collection of updates to SNAP potential within the Kokkos package when using the Cuda backend (though it should generalize well to any GPU). It does not include any algorithmic overhauls. The PR features multiple discrete and largely orthogonal commits related to code readability and cleanliness, empirical performance stability across different CUDA toolkit versions, and some small (~1-2%) performance boosts. More details on each commit are given under Implementation Notes. Readability, cleanliness, and overall design choices are of course all up to discussion; I'm not married to any one convention.

This PR has no effect to the CPU Kokkos backend (it may actually give a ~1% speed-up, for much the same reason other SNAP PRs have caused slowdowns---no good reason). This has no effect on the CPU implementation in the SNAP package, which has historically been faster than the Kokkos/CPU backend.

Overall performance notes below, with some per-PR speedups given under Implementation Details. For robustness, I note that `master` is specifically the `patch_10Feb2021` tag at time of filing.

**Related Issue(s)**

Iterative improvement on #2549, other Kokkos/SNAP PRs.

**Author(s)**

Evan Weinberg (NVIDIA). eweinberg [at] nvidia.com preferred, evansweinberg [at] gmail.com for "long-lived".

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

No issues.

**Implementation Notes**

* https://github.com/lammps/lammps/commit/d23755085467590d5382322139f6cd933b64c302 +https://github.com/lammps/lammps/commit/334c0d9969a1c164b39654258cbd083a85c5ca46: Replace the `dir` runtime variable in the `SNAKokkos` class with a template parameter to the `*ComputeFusedDeidrj` tag. Latter commit fixes a host CPU compile-time error uncovered when compiling with just the `Serial` Kokkos backend. Generally iso-performance.
[...]

PR #4391, opened 2024-11-22
SNAP/Kokkos CPU+GPU code unification and other optimizations

weinbe2:
**Summary**

This PR unifies the CPU and GPU codepaths as much as possible in the Kokkos implementation of the SNAP force evaluation; of the core kernels, only ComputeUi and ComputeDuidrj/Deidrj still have separate codepaths between the CPU and the GPU. As a byproduct of this work and other included optimizations performance increases by ~10% on NVIDIA H100 80GB.

As a byproduct of this work, the PR also improves performance of the CPU backend, both serial and OpenMP, but in general it is still less performant than the “default” implementation in the ML-SNAP package and unsurprisingly less performant than the AVX-512 implementation in the INTEL package.

As part of the implementation, this PR generally improves compile time by avoiding redundant compilation of CPU and GPU kernels. This is with the new `if constexpr (host_flag)` machinery.

There are now separate codepaths for chemsnap and non-chemsnap simulations, greatly reducing integer arithmetic for some kernels in the non-chemsnap case. Additional parallelism was exposed in ComputeBeta which is beneficial at low atom counts.
An additional launch bound argument was used in ComputeYi to further stabilize performance across CUDA versions. It also improved performance on H100 by reducing register requirements, boosting occupancy.

This PR was only tested on NVIDIA GPUs but the optimization efforts were done in good faith towards other architectures; I believe that the overall reduction in integer arithmetic with the non-chemsnap codepaths plus reducing the additional offset calculation by removing the AoSoA data layouts should be beneficial across the board. Launch bounds in `pair_snap_kokkos.h` for other architectures may need to be re-tuned; without access to other architectures I could not take this on myself.

Performance improvements are marginal (but non-zero) for chemsnap; the benefits for linear snap and quadratic snap are much more notable.

**Related Issue(s)**

Iterative improvement on #2599 , #2905 , #3465 , other Kokkos/SNAP PRs...

**Author(s)**

**Licensing**

**Backward Compatibility**

No issues.

**Implementation Notes**

The core requirement for unifying the CPU and GPU codepaths was unifying the CPU and GPU data structures. This was enabled by removing the AoSoA data layouts but preserving the tiled data access patterns. Extra padding is added to the views to preserve cache alignment; the memory overheads of the extra padding just make the size of the more-flattened Views equal to their original AoSoA form.

A special “dispatcher” utility function was added to simplify launching the GPU versions of kernels as Rank-3 MDRangePolicy calls and launch the CPU versions of kernels as simple RangePolicy calls (with optional 2-d MDRangePolicy implementations available at compile time). This is the `snap_get_policy` dispatcher in `pair_snap_kokkos.h`.

The separate codepaths for chemsnap and non-chemsnap were implemented via separate kernels for each case. In the non-chemsnap codepath, the “for” loops over element ids are eliminated and “zero” is baked in for the element number offset, greatly reducing the number of integer instructions.

The second launch bound argument, “min blocks”, was set to a value of “4” for the ComputeYi and Zi kernels for the CUDA backend and left to a default of “0” on other backends, but this is easy to modify as it is implemented as a new `static constexpr int` variable.

Correctness was verified by comparing the results for various flavors of SNAP between the current `develop` branch and this PR branch.

Performance was checked on an NVIDIA H100 80GB HBM3 using CUDA 12.6.3. “m-at/s” refers to million atom timesteps per second. Performance table for a few flavors of SNAP (linear, chemSNAP, carbon qSNAP):

| benchmark | atoms | develop | PR | speed-up |
|---|---|---|---|---|
| ECP 2J8 | 2000 atoms | 2.101 m-at/s | 2.262 m-at/s | 1.08x |
| ECP 2J14 | 2000 atoms | 0.306 m-at/s | 0.351 m-at/s | 1.15x |
| ECP 2J8 | 512,000 atoms | 3.715 m-at/s | 4.152 m-at/s | 1.12x |
| pSNAP | 512,000 atoms | 2.464 m-at/s | 2.585 m-at/s | 1.05x |
| GB qSNAP | 373,248 atoms | 2.651 m-at/s | 3.100 m-at/s | 1.17x |

**Post Submission Checklist**

None.


stanmoore1:
@weinbe2 sorry for the delay, this all looks great, nice work! My test showed that Kokkos SNAP on CPU is as fast or even a little faster than vanilla SNAP. Currently this code path is blocked because it falls back to the vanilla CPU code. I think it now makes sense to allow Kokkos Serial and OpenMP, i.e. remove checks like this: https://github.com/lammps/lammps/pull/4391/files#diff-36b9dfd9a71b342a96de48e3a23c4f966ec640df5ebcbe105f4fc5a1aeaa4d97R85-R92. Do you agree?

weinbe2:
> @weinbe2 sorry for the delay, this all looks great, nice work! My test showed that Kokkos SNAP on CPU is as fast or even a little faster than vanilla SNAP. Currently this code path is blocked because it falls back to the vanilla CPU code. I think it now makes sense to allow Kokkos Serial and OpenMP, i.e. remove checks like this: https://github.com/lammps/lammps/pull/4391/files#diff-36b9dfd9a71b342a96de48e3a23c4f966ec640df5ebcbe105f4fc5a1aeaa4d97R85-R92. Do you agree?

I support allowing the Kokkos Serial and OpenMP backend now (especially because you can still opt-out), especially if you're seeing perf parity/a perf boost from using it. I still saw it being a smidge slower in my (very) limited testing, but our mixed experiences suggest we're at that "break-even point" where one difference or another can push the needle on one CPU vs another and we may as well not torture ourselves about it.

Do you mind popping the change in, since it looks like you have the power to make the change? I'm not sure if there's some extra documentation you'd like to add along with it as well.

stanmoore1:
@weinbe2 thanks, yes I will make the change. I don't think any documentation is needed but I will double check.

stanmoore1 (approved):
I approve, nice work @weinbe2

PR #4412, opened 2024-12-12
Revert "Disallow newton on with full neigh list"

jhelie:
This reverts commit 27954609b8f1c2f3333064f61dd27ad971ea6132.

**Summary**

We use Kokkos for simulating many-body interactions using an in-house ML FF. We need the ability to use `newton on` as well as `neigh full` together.

I couldn't find an explanation of why this was disallowed in [the PR that recently banned this possibility](https://github.com/lammps/lammps/pull/4346). What was the rationale? 

cc @stanmoore1 @nw13slx 

**Related Issue(s)**

**Author(s)**

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/src/KOKKOS/kokkos.cpp
+++ b/src/KOKKOS/kokkos.cpp
@@ -633,22 +633,13 @@ void KokkosLMP::accelerator(int narg, char **arg)
   // set neighbor binsize, same as neigh_modify command
 
   force->newton = force->newton_pair = force->newton_bond = newtonflag;
-  newton_check();
+
+  if (neigh_thread && newtonflag)
+    error->all(FLERR,"Must use KOKKOS package option 'newton off' with 'neigh/thread on'");
 
   neighbor->binsize_user = binsize;
   if (binsize <= 0.0) neighbor->binsizeflag = 0;
   else neighbor->binsizeflag = 1;
-}
-
-/* ---------------------------------------------------------------------- */
-
-void KokkosLMP::newton_check()
-{
-  if (neighflag == FULL && force->newton)
-    error->all(FLERR,"Must use 'newton off' with KOKKOS package option 'neigh full'");
-
-  if (neigh_thread && force->newton)
-    error->all(FLERR,"Must use 'newton off' with KOKKOS package option 'neigh/thread on'");
 }
 
 /* ----------------------------------------------------------------------
--- a/src/KOKKOS/kokkos.h
+++ b/src/KOKKOS/kokkos.h
@@ -64,7 +64,6 @@ class KokkosLMP : protected Pointers {
   static void initialize(const Kokkos::InitializationSettings&, Error *);
   static void finalize();
   void accelerator(int, char **);
-  void newton_check();
   bigint neigh_count(int);
 
   template<class DeviceType>
--- a/src/accelerator_kokkos.h
+++ b/src/accelerator_kokkos.h
@@ -59,7 +59,6 @@ class KokkosLMP {
   void accelerator(int, char **) {}
   int neigh_list_kokkos(int) { return 0; }
   int neigh_count(int) { return 0; }
-  void newton_check() {};
 };
 
 class AtomKokkos : public Atom {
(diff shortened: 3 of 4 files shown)


stanmoore1:
@jhelie you can still use `newton on` with `neigh full` together in the current code, no need to revert commit https://github.com/lammps/lammps/commit/27954609b8f1c2f3333064f61dd27ad971ea6132. For example see https://github.com/lammps/lammps/blob/develop/src/KOKKOS/pair_snap_kokkos_impl.h#L100-L110.

Unfortunately the nomenclature here is a little confusing but I will try to explain. This option was originally written for simple pairwise and some manybody potentials. In that case, a "FULL" neighborlist means work duplication and no thread atomics, while "HALF" means no work duplication and thread atomics. For "FULL" and a pairwise potential, you do twice the work as a "HALF" neighbor list and avoid the write conflict in the force array. This is also an option with the manybody EAM potential and at one point "FULL" without atomics was even supported by Tersoff and Stillinger-Weber, but the "FULL" option required even more than 2x work duplication (again avoiding thread atomics) and was slower so the "FULL" option was removed.

Unfortunately SNAP and other ML potentials break this mold since they require a full list AND thread atomics, so in the current nomenclature they are classified as a "HALF" list to the user, even if they use a full list in the code. I know this doesn't make a lot of sense so if anyone has suggestions on a better naming convention I'm receptive to that.

The problem is that using a "full" neighbor list plus `newton on` for a simple pairwise potential leads to unnecessary overhead. `Newton on` spawns a "reverse" communication, but the forces have already been computed for the ghost atoms with a full list, so it is pure overhead. So that is why the option was banned. It doesn't affect ML potentials that use a full list with atomics.

So bottom line is this PR is not necessary, you can still do what you need with your ML potential, even though the naming is funky. See SNAP and PACE as examples.

stanmoore1:
The problem is basically that we template the functors on `FULL` and `HALF`, and with `FULL` it turns off the atomics which for SNAP and other ML potentials is not what we want. So we have to use `HALF` instead for the templates, even though it is actually using a full list under the hood. I know this is confusing, maybe the templates should use `ATOMIC` and `NONATOMIC` instead of `HALF` and `FULL` but it made sense before all the ML potentials were added!

stanmoore1:
@jhelie I'm going to close this PR since I'm confident the current code can be made to work with your ML potential. If you have more questions, please feel free to respond here, on MatSci, or LAMMPS developer slack channel. Thanks.

akohlmey:
> We need the ability to use `newton on` as well as `neigh full` together.

You need to explain what you are using this for. Apart from the special *additional* implications that @stanmoore1 mentioned, the choices of "newton on" and a full neighbor list contradict each other. A full neighbor list is by its very definition implying "newton off" since *all* pairs are listed twice (local or including pairs with ghosts). You can still skip half the interactions manually inside your pair style, either for all pairs or only pairs that straddle subdomain boundaries. This is done for example for the two-body part of pair style sw on the CPU.

stanmoore1:
@akohlmey right, this is why to the user the option of `newton on` and `neigh full` is now disabled and will error out. However in the code a developer can still use a full neigh list plus `newton on` as is done for SW 3-body, SNAP, PACE, FLARE, etc. etc.

jhelie:
Thanks both for the very quick reply and apologies for not following-up earlier. Understood, thanks for the pointers. Out of curiosity I do wonder what is the point of the code below in in the [SNAP example linked](https://github.com/lammps/lammps/blob/develop/src/KOKKOS/pair_snap_kokkos_impl.h#L108-L109) ? AIUI the neighbour list is forcefully set to `FULL`, so what is the point of this check - is a `HALF` list computed and used for something else? 

```cpp
neighflag = lmp->kokkos->neighflag;
if (neighflag == FULL)
  error->all(FLERR,"Must use half neighbor list style with pair snap/kk");
```

stanmoore1:
SNAP requires thread atomics, which is accomplished by using a `HALF` list in the KOKKOS package settings. As was mentioned previously, a better keyword could be `ATOMIC` instead of `HALF` but it is just nomenclature.

PR #4419, opened 2024-12-19
add area and volume preserving barostats

srtee:
**Summary**

Allows barostats to preserve volume or area through deviatoric barostatting (setting suitable hydrostatic stress trace to zero).

**Author(s)**

Shern Tee (Griffith University), s.tee@griffith.edu.au / shernren@gmail.com

**Licensing**

**Backward Compatibility**

Only new input options -- should not modify existing behaviour.

**Implementation Notes**

Added `vol_preserve_flag` and `area_preserve_flag` to `FixNH` (!!), and offsets `omega_dot` vectors as needed during `nh_omega_dot()` so that the necessary sums are zero (thus preserving volume or area as needed).

Technically either option removes one barostat DOF. Anybody who cares enough will be able to modify DOF as needed (I'm not sure even existing `iso` / `aniso` / `tri` options track their DOFs to similar accuracy ...).

**Post Submission Checklist**

Follows from MatSci discussion: https://matsci.org/t/molecular-dynamics-simulation-in-the-n-v-a-0-t-ensemble/59675/4


doc/src/fix_nh.rst
```diff
@@ -398,6 +402,32 @@ Dullweber-Leimkuhler-McLachlan integration scheme
 giving better energy conservation and allows slightly longer timesteps
 at only a small additional computational cost.
 
+.. versionadded:: latest
```
akohlmey, line 405:
```suggestion
.. versionadded:: TBD
```

src/fix_nh.cpp
```diff
@@ -467,6 +476,36 @@ FixNH::FixNH(LAMMPS *lmp, int narg, char **arg) :
...
+  if (area_preserve_flag) {
+    if (p_flag[0] == 0 || p_flag[1] == 0)
+      error->all(FLERR,"Fix {} must barostat both x and y dimensions to preserve xy area");
+    if (pcouple == XYZ || pcouple == XY)
+      error->all(FLERR,"Fix {} cannot couple x and y dimensions while preserving xy area", style);
+    if (pcouple == XZ || pcouple == YZ)
+      error->all(FLERR,"Fix {} cannot couple z dimension while preserving xy area", style);
+    if ((p_start[0] + p_start[1] != 0.0 ||
+         p_stop[0] + p_stop[1] != 0.0) && comm->me == 0)
+      error->warning(FLERR,"Fix {} has x and y pressures that do not sum to zero but is preserving xy area", style);
```
akohlmey, line 503:
```suggestion
      if (comm->me == 0)
        error->warning(FLERR,"Fix {} has x and y pressures that do not sum to zero but is preserving xy area", style);
```

akohlmey (changes requested):
These changes need to be confirmed with all classes that derive from FixNH or classes that derive from those derived classes.
```
$ grep -l public.FixNH.\{ *.h */*.h
fix_nh_sphere.h
fix_nph.h
fix_npt.h
fix_nvt.h
fix_nvt_sllod.h
ASPHERE/fix_nh_asphere.h
BODY/fix_nh_body.h
EFF/fix_nh_eff.h
GPU/fix_nh_gpu.h
INTEL/fix_nh_intel.h
KOKKOS/fix_nh_kokkos.h
OPENMP/fix_nh_omp.h
SHOCK/fix_nphug.h
UEF/fix_nh_uef.h
```
and it is going to be best to add corresponding tests to `unittest/force-styles/tests`

jtclemm:
Not sure how important uniform notation is, but the functionality of _preserve xyz_ would be enabled in fix deform using _box volume_. But the _box_ argument is also designed to provide additional functionality, so not exactly one-to-one.

athomps:
I have not looked at this in detail.  The functionality is useful and the core code looks good. My only immediate concern is the restriction: `all pressures sum to zero to preserve xyz volume`.  By fixing the volume, the resultant average pressure from the simulation will also be fixed to some value, but that value will not be zero. Maybe a better restriction would be `all target axial stresses must be equal and all target shear stresses must be zero` i.e. zero deviatoric stress.

athomps:
I agree with @jtclemm that this is similar to the `volume` keyword in `fix deform`. It also has something in common with the existing `couple` keyword in `FixNH`.  Combining these two ideas, I think it would not be hard to convert this to a simple and general optional keyword `volume` or `anticouple`, similar to `couple`:

`volume value = none or xyz or xy or yz or xz`

In the code, this would have the effect of offsetting `omega_dot` for the anti-coupled dimensions so that they sum to zero. At each timestep, the corresponding scale factors applied to the anti-coupled dimensions will multiply to 1, so area or volume is preserved.  I also see no need to check the values of the target stress values. Users should be free to (ab)use this as they like.  The only checks that are strictly needed are the same as the dozen or so checks for the `couple` keyword.

akohlmey:
@srtee This has not seen any activity for two months. Are you still planning to work on it?
Ahead of the stable release we are trying to pare down the pull request list to those under active development to ideally keep the list of pending pull requests down to just one page (everything on the second page and beyond gets rarely looked at).

One way to move forward could be to close this pull request and either resubmit it later when you have time, or submit a feature request issue that then points to the closed pull request as a starting point for somebody to take over and flag it as "volunteer needed".

srtee:
I'm not planning to work actively on this. Let's just close the PR for now -- if someone else has a feature request mentioning this, we can reopen it then.

akohlmey:
This was converted into issue #4623

PR #4422, opened 2024-12-19

akohlmey:
**Summary**

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- make unit tests for format style variables independent of rounding issues (triggered by most recent MS Visual C++ on Windows 11 24H2)
- work around bug in FindVTK.cmake for some versions of VTK that does not enable C before checking for C components)
- add check to `init()` functions of various fix styles to make certain an internal pressure or temperature compute is still valid and not manually replaced via unfix/fix which would bypass the corresponding check in fix_modify.

**Post Submission Checklist**

### DIFF
--- a/src/fix_press_berendsen.cpp
+++ b/src/fix_press_berendsen.cpp
@@ -266,19 +266,21 @@ void FixPressBerendsen::init()
   // set temperature and pressure ptrs
 
   temperature = modify->get_compute_by_id(id_temp);
-  if (!temperature)
-    error->all(FLERR, "Temperature compute ID {} for fix press/berendsen does not exist", id_temp);
-
-  if (temperature->tempbias)
-    which = BIAS;
-  else
-    which = NOBIAS;
+  if (!temperature) {
+    error->all(FLERR, "Temperature compute ID {} for fix {} does not exist", id_temp, style);
+    if (temperature->tempflag == 0)
+      error->all(FLERR, "Compute ID {} for fix {} does not compute a temperature", id_temp, style);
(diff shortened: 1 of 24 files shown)


src/fix_press_berendsen.cpp
```diff
@@ -266,17 +266,22 @@ void FixPressBerendsen::init()
...
-  if (!temperature)
-    error->all(FLERR, "Temperature compute ID {} for fix press/berendsen does not exist", id_temp);
-
-  if (temperature->tempbias)
-    which = BIAS;
-  else
-    which = NOBIAS;
+  if (!temperature) {
+    error->all(FLERR, "Temperature compute ID {} for fix {} does not exist", id_temp, style);
+    if (temperature->tempflag == 0)
```
jtclemm, line 271: Does fix press/berendsen even use the temperature? I see compute calls, but they look unnecessary.

Or is its output somehow needed for compute pressure?

sjplimp: not sure what you mean - press/berend creates a pressure compute and passes it the temp compute it also created - then in end_of_step() it invokes both and uses the result in couple()

sjplimp (approved):

src/fix_press_berendsen.cpp
jtclemm: Ah I missed that the id was passed to the pressure compute.

jtclemm (approved):

PR #4424, opened 2024-12-19
Kinetic Monte Carlo fix

jtavenner-dev:
**Summary**

Integrated MD-KMC functionality to the development branch. MD-KMC implementation as described in "Molecular Dynamics based kinetic Monte Carlo simulation for accelerated diffusion"

**Related Issue(s)**

Current changes force the entire MC package to depend on the Voronoi package, as used in the current MD-KMC implementation. This is likely undesirable but the authors are unsure of the desired implementation approach.

**Author(s)**

Jacob Tavenner - NASA Ames Research Center
tavennerjacob@gmail.com

**Licensing**

**Backward Compatibility**

No issues related to backward compatibility are known, and the code has been tested under the current development branch.

**Implementation Notes**

The fix uses the Voronoi LAMMPS package to calculate a Voronoi region around target atoms. The correctness of the algorithm was calculated based on matching expected diffusion profiles as demonstrated in the manuscript.

**Post Submission Checklist**

https://doi.org/10.1016/j.commatsci.2022.111929


akohlmey:
> Current changes force the entire MC package to depend on the Voronoi package, as used in the current MD-KMC implementation. This is likely undesirable but the authors are unsure of the desired implementation approach.

This is not a problem. There are methods to skip compiling/installing the fix if the voronoi package is not installed/enabled. I can take a look and make the necessary changes.

akohlmey:
@jtavenner-dev the changes for a conditional build are now in place. It was pretty straightforward to add, since already fix sgcmc had a similar dependency (on the MANYBODY package).

I've also removed commented out debug code, made some formatting and LAMMPS programming convention related changes, fixed a few spelling, formatting and linking issues in the manpage and replaced calls to pow(x,2.0) - which are very slow - with calls to MathSpecial::square() - which are much faster.

jtavenner-dev:
The changes look good, thanks for reviewing and removing the debug and updating the programming convention, etc.

The solution for the conditional build is ideal, I am glad to see that as an option from the MANYBODY example and now replicated here.

akohlmey:
@sjplimp @athomps Can you please have a look at this or recommend somebody else to have a look?

sjplimp:
@akohlmey I will take a look

sjplimp:
Hi Jacob @jtavenner-dev  - thanks for submitting this new KMC fix - it looks
interesting.  I have a couple general comments about things I don't
understand.  Possibly looking at your recent paper would help me, but
I don't have access to it.  If you can email it to me, I will take a
look.  But I also think users familiar with KMC may have the same Qs,
so clarifying them in the new doc page would be good.  I also have
some more minor comments about things I did not understand from the
doc page, which should be clarified.  I have not looked at the code
itself yet, except see (15) below.

General Qs:

(1) Not clear why you call this KMC behavior.  I think of KMC as
defining a bunch of relative probabilities for competing "events",
then choosing one event based on those probs, then updating the
probs, and repeating.  Maybe you are doing something like that, but it
is not clear from the doc page.  So I think the doc page needs to
explain more precisely what the algorithm is and how it works in parallel
and why it is KMC-(like).

Specifically:
(a) is a swap an event ?
(b) what does it mean to "attempt" an event, 
    - in KMC I think there is always an event selected and performed, no attempts (unlike Metropolis MC)
(c) what is the set of events which KMC is choosing from ?
    (c1) is it each pair of atoms which are Voronoi neighbors? (subject to the keywords)
    - Note: I think the definition/meaning of a Voronoi neighbor and why it makes
      the algorithm KMC-like should be explained on this doc page
    (c2) or do you pick an atom randomly and the set of events to choose from 
         is only swaps with that atom's neighbors, which is not as KMC-like
(d) for (c1) or (c2) do you compute all the probabilities for each potential event
    in the set, then use the KMC rule to pick one event with the correct probability ?
    - how do you do that efficiently if (c1) is your algorithm ?
    - do you redo the whole procedure after each performed event or re-use
      previously computed probabilities ?
    - after you perform an event, do you update the Voronoi neighbor lists
      to reflect that some atoms now have different neighbors ?

(2) Similar Qs as (1) but for how the algorithm works in parallel

(a) Is this algorithm performed by each proc simultaneously
  on only its subset of atoms and their neighbors ?
  Or is it done across the entire system to select a single event (true KMC) ?
  In parallel, can an atom swap with an atom owned by another proc ?
(b) What makes this fix and its algorithm inefficient (if anything does) ?
  E.g. too many atoms/proc - do you expect it to scale linearly (in a performance sense) as N/P (same as MD itself)
(c) Should a user expect to get the same statistical answer no matter
  how many procs the simulation is run on ?
  might not be the case if atoms owned by different procs cannot swap

More minor comments - almost all about the doc page:

(3) For "types" keyword, you allow 2 or more args and type labels.  How
  do you determine the end of the args?  What if a type label is "ke" ?
  The more typical syntax in LAMMPS for something like this is:
  types Nt 1 4 5 foo 7 8
  where Nt = # of args to follow = 6 in this case
(4) The "diff" keyword takes an atom type but not a type label ?
(5) What should the user choose for R ?
    In the prob formula it uses r_0. If R = r_0, I suggest using
    R0 instead or R in the input syntax and in the formula.
    If not, you need to explain how r_0 is calculated.
(6) syntax of "rates" keyword - I suggest
    rates values = V1 V2 ... Vntypes
      V1,V2,... = numeric weighting for each of Ntypes
    Note: can these be floating point weights?
(7) Once you have explained the algorithm at the start of the doc page,
  then have a section (line separated) which explains the keywords
  and what they do in the nomenclature of the algorithm
  e.g. diff = only one atom type is allowed for the I atom in each
    event pair
  e.g. rates = alter the probabilities of the prob formula for IJ,
    is it a pre-factor ?
    of possibly for (c2) it alters how you randomly pick an atom I ?
(8) Are there some of the keywords which are not allowed with other keywords ?
(9) For "region" keyword, do both atom I and J need to be in the region, or just I ?
(10) For "voro" keyword, explain/warn that if the groups do not match,
  then LAMMPS will not complain, but it means some atoms will
  have no Voronoi neighbors and thus no possible swaps.
(11) Symmetry of swaps - is an IJ and JI swap equally probable?
  Seems like they may not be if region or other keywords are specified ?
  If not, this might break detailed balance for MC ?
(12) Since the "voro" keyword is required, I suggest making it a required arg w/out the "voro"
  e.g. N X seed T R voro-ID keyword values ...
(13) Semigrand is listed as a keyword but not explained.
(14) Default for "diff" keyword is listed as no, which doesn't make sense

One comment on the code:

(15) You are using RanPark for both "equal" and "unequal" random
number generators (RNGs).  Unequal meaning different on different
procs.  But you are using the same seed on all procs for the
unequal. That could definintely lead to unwanted correlations between
events on different procs (if that's how you are using it).  I suggest
you look at how fix langevin defines an unequal RNG using RanMars.
That is the standard LAMMPS way to do it.  It does make restarting
much harder, b/c each proc has its own RNG and state.  But we
generally don't worry about that for any use of unequal RNGs.
I.e. fix langevin does not attempt to restart exactly.  Conceptually it should be
fine for a restarted run versus a non-restarted run to be
statistically similar - both for Langevin and definitely for anything
with MC in its name :) 
So I would advise using RanMars and not worrying about exact restarts.

regnar-exa:
Hi Steve @sjplimp hopefully I can answer a few of your questions here. I've emailed you copies (to sjplimp@gmail.com) of both the paper outlining the method implemented here (ComMatSci 2023), and one where we used it for calculating precipitation behavior (JCP 2024). I think it's good to clarify the questions related to the doc page as well.

Responses:

(1) Calling this KMC is mostly from internal usage and development when outlining the original goal of the fix to be able to simulate the diffusion of a species through a solid by uncoupling its movement from the MD timescale while keeping the MD physics between moves. The selection of an event probability is done as in Metropolis MC so that we can capture the proper energetic behavior as given through MS calculations. This is the reason to name the fix neighbor/swap, as the fix is designed to enable kMC-like behavior, though the underlying algorithm is closer to a Metropolis MC. I will adjust the reference to kMC in the doc accordingly.

Specifics:
(a) A swap is selected from the Voronoi neighbors of an atom with a probability-weighted based on the distance of the neighbor atom to the central atom (i.e. if all neighbor atoms are valid types, their distance-based probabilities given by p_{ij} are summed and then one is randomly selected based on those probabilities and the RNG).
(b) Once a neighbor has been selected to be swapped, the swap is evaluated using the Metropolis criterion to determine if it is kept (as in the atom/swap fix)
(c) Reference to kMC may be confusing. In the doc page, it is only referenced once. The behavior is closer to (c2), where a random (valid) atom is selected from the entire simulation, then the local probability to determine which neighbor to swap to is evaluated. Can expand doc on what a Voronoi neighbor is with reference to the Voronoi tesselation
(d) Only the local probabilities for a single atom are computed to determine which swap will be performed. Swap acceptance is also subject to the metropolis selection.
- the procedure is re-done each time since the distance probability calculation is minimal compared to the energy evaluation
- the atom types are swapped to avoid neighbor list re-calculation within the MC steps

(2)

(a) It is performed once across the entire system to simulate a single event. Earlier versions enabled parallel events, but this also leads to errors at the processor bounds where the neighbors are on different procs. Since energy evaluation is also required for the Metropolis evaluation all procs are called at that point.
(b) The inefficiency is from energy evaluations of the MC done across all procs rather than just for atoms within the interatomic potential cutoff, i.e. comparing total system energy is unnecessary since for large systems only a small number of atoms will have a change in energy due to the swap. This is an inefficiency shared with all the current MC implementations to my knowledge.
(c) Yes, the same statistical answer should be given regardless of the number of procs since atoms can swap across proc boundaries. This has been tested.

(3) Currently it only accepts types if they are numeric. i.e. the stopping criterion is non-numeric character. It seems this behavior is consistent with other fixes but it can be adjusted if desired.
(4) Currently only uses numeric atom types as the 'type' keyword.
(5) Change the label to R0 in the doc, it is the same. Will update the doc with reasoning, since the probability is weighted behavior is not significantly sensitive to the exact value of R0.
(6) Have changed keywording as described. Yes, floating point is accepted if there is a particular method of notation for the doc?
(7) Will update the doc with a keyword section. Is there an example of an existing LAMMPS fix with such a section to follow the same structure?
(8) Only types and diff are exclusionary since they are two different methods of specifying the desired atoms to swap. Is there an accepted way to note this in the doc? Likewise, one of these keywords is required.
(9) Both atoms need to be in the region to swap. If an atom J is not in the region it will not be computed for the local selection probability.
(10) Adding this to the doc. I am not aware of a way to check this in code to throw a warning?
(11) Yes, the swaps are not always equally likely since only one atom type is selected as type I, and either swapped with all other atom types (diff keyword) or selected atom types (type keyword). It does not appear to break the MC balance since if both I and J types are able to be selected the probabilities of the swap happening are equal. Need to make this clearer in the docs.
(12) Will update to require voro.
(13) Doc error, not a valid keyword
(14) Removed for clarity.

(15) Random unequal is unnecessary since only one event occurs at a time, will remove (currently unused holdover).

jtavenner-dev:
Hi Steve @sjplimp hopefully I can answer a few of your questions here. I've emailed you copies (to sjplimp@gmail.com) of both the paper outlining the method implemented here (ComMatSci 2023), and one where we used it for calculating precipitation behavior (JCP 2024). I think it's good to clarify the questions related to the doc page as well.

Responses:

(1) Calling this KMC is mostly from internal usage and development when outlining the original goal of the fix to be able to simulate the diffusion of a species through a solid by uncoupling its movement from the MD timescale while keeping the MD physics between moves. The selection of an event probability is done as in Metropolis MC so that we can capture the proper energetic behavior as given through MS calculations. This is the reason to name the fix neighbor/swap, as the fix is designed to enable kMC-like behavior, though the underlying algorithm is closer to a Metropolis MC. I will adjust the reference to kMC in the doc accordingly.

Specifics:
(a) A swap is selected from the Voronoi neighbors of an atom with a probability-weighted based on the distance of the neighbor atom to the central atom (i.e. if all neighbor atoms are valid types, their distance-based probabilities given by p_{ij} are summed and then one is randomly selected based on those probabilities and the RNG).
(b) Once a neighbor has been selected to be swapped, the swap is evaluated using the Metropolis criterion to determine if it is kept (as in the atom/swap fix)
(c) Reference to kMC may be confusing. On the doc page, it is only referenced once. The behavior is closer to (c2), where a random (valid) atom is selected from the entire simulation, then the local probability to determine which neighbor to swap to is evaluated. Can expand doc on what a Voronoi neighbor is with reference to the Voronoi tesselation
(d) Only the local probabilities for a single atom are computed to determine which swap will be performed. Swap acceptance is also subject to the metropolis selection.
- the procedure is re-done each time since the distance probability calculation is minimal compared to the energy evaluation
- the atom types are swapped to avoid neighbor list re-calculation within the MC steps

(2)

(a) It is performed once across the entire system to simulate a single event. Earlier versions enabled parallel events, but this also leads to errors at the processor bounds where the neighbors are on different procs. Since energy evaluation is also required for the Metropolis evaluation all procs are called at that point.
(b) The inefficiency is from energy evaluations of the MC done across all procs rather than just for atoms within the interatomic potential cutoff, i.e. comparing total system energy is unnecessary since for large systems only a small number of atoms will have a change in energy due to the swap. This is an inefficiency shared with all the current MC implementations to my knowledge.
(c) Yes, the same statistical answer should be given regardless of the number of procs since atoms can swap across proc boundaries. This has been tested.

(3) Currently it only accepts types if they are numeric. i.e. the stopping criterion is non-numeric character. It seems this behavior is consistent with other fixes but it can be adjusted if desired.
(4) Currently only uses numeric atom types as the 'type' keyword.
(5) Change the label to R0 in the doc, it is the same. Will update the doc with reasoning, since the probability is weighted behavior is not significantly sensitive to the exact value of R0.
(6) Have changed keywording as described. Yes, floating point is accepted if there is a particular method of notation for the doc?
(7) Will update the doc with a keyword section. Is there an example of an existing LAMMPS fix with such a section to follow the same structure?
(8) Only types and diff are exclusionary since they are two different methods of specifying the desired atoms to swap. Is there an acceptable way to note this in the doc? Likewise, one of these keywords is required.
(9) Both atoms need to be in the region to swap. If an atom J is not in the region it will not be computed for the local selection probability.
(10) Adding this to the doc. I am not aware of a way to check this in code to throw a warning?
(11) Yes, the swaps are not always equally likely since only one atom type is selected as type I, and either swapped with all other atom types (diff keyword) or selected atom types (type keyword). It does not appear to break the MC balance since if both I and J types can be selected the probabilities of the swap happening are equal. Need to make this clearer in the docs.
(12) Will update to require voro.
(13) Doc error, not a valid keyword
(14) Removed for clarity.

(15) Random unequal is unnecessary since only one event occurs at a time, will remove (currently unused holdover).

sjplimp:
@jtavenner-dev Thanks for the detailed response - all sounds good.
 I will read the revised doc page when you have it.

I suggest an organization similar to this:

explain what kinds of models the command is good for, how different than other related MC
  options in LAMMPS, e.g. fix atom swap or others ? 
describe the algorithm in some detail, so the user knows what the new fix is doing
  (a) an atom I is randomly selected
  (b) one of its Voronoi neighbors J is selected via your (1a) above
  (c) the before/after energy of the IJ swap is calculated, mention it is a global calc (cost similar to timestep)
  (d) the Metropolis criterion is applied (show eq) to accept/reject the swap
  (e) repeat X times
  and how all of this works in parallel (how do 2 atoms on different procs get swapped)
highlight any performance issues - for example with the energy calculation or parallel communication
then a section of the doc page on all the optional keywords, describe how each of them affects a,b,c,d
restrictions:
  does this work with any pair style, or with KSpace ?
  a Voronoi neigh can be further away than the pair style or comm cutoff - is that allowed ?

sjplimp:
Also, I did not receive an email with a PDF of your paper ...

jtavenner-dev:
Sounds good, will update doc with a more detailed description of the algorithm and relation to other LAMMPS features. Is sjplimp@gmail.com the correct address?

sjplimp:
yes, that's my email

jtavenner-dev:
I have sent the papers again, hopefully you have recieved it this time?  If not it looks like I may be able to upload them here.

sjplimp:
still no paper email, I have no idea why

jtavenner-dev:
Odd. Hopefully this linking works.
[2023_MDkMC_Algorithm_Tavenner.pdf](https://github.com/user-attachments/files/18605422/2023_MDkMC_Algorithm_Tavenner.pdf)
[2024_Ni3AlPrecip_Tavenner.pdf](https://github.com/user-attachments/files/18605423/2024_Ni3AlPrecip_Tavenner.pdf)

sjplimp:
that worked for the 2 papers - thanks

akohlmey:
@sjplimp @jtavenner-dev can you please post an update on the state of this pull request?

jtavenner-dev:
Thanks for the bump. I have now added the updated doc page for review.

jtavenner-dev:
All changes are in from my end. Please let me know if further adjustments are needed for approval.

sjplimp:
@jtavenner-dev - thanks for updates.  I am reading the new doc page now and it addresses my suggestions.  I am making a few edits for clarity, and will email a new version to you, hopefully later today.  Too many changes to try and do it via GHub comments.

athomps (approved):
I approve

jtavenner-dev:
I have updated for all edits of the doc now.

akohlmey:
@sjplimp This is waiting on your final review now.

sjplimp:
@jtavenner-dev thanks - I will read the new doc page today

doc/src/fix_neighbor_swap.rst
```diff
@@ -0,0 +1,259 @@
...
+   fix ID group-ID neighbor/swap N X seed T R0 voro keyword values ...
+
+* ID, group-ID are documented in :doc:`fix <fix>` command
+* neighbor/swap = style name of this fix command
+* N = invoke this fix every N steps
+* X = number of swaps to attempt every N steps
+* seed = random # seed (positive integer)
+* T = scaling temperature of the MC swaps (temperature units)
+* R0 = scaling swap probability of the MC swaps (distance units)
+* voro = valid voronoi compute id (compute voronoi/atom)
```
sjplimp: I think this line vanishes, b/c it is now a keyword?

doc/src/fix_neighbor_swap.rst
```diff
@@ -0,0 +1,259 @@
...
+* two or more keyword/value pairs may be appended to args
+* keywords *types* and *diff* are mutually exclusive, but one must be specified
```
sjplimp, line 22: * keyword *voro* is required
* keywords types and diff are mutually ...

doc/src/fix_neighbor_swap.rst
```diff
@@ -0,0 +1,259 @@
...
+* keyword = *types* or *diff* or *ke* or *region* or *rates*
```
sjplimp, line 23: *voro* or *types* or *diff* or *ke* ...

doc/src/fix_neighbor_swap.rst
```diff
@@ -0,0 +1,259 @@
...
+atoms *I* of the first atom type will be selected.  Only atoms *J* of the
+remaining atom types will be considered as potential swap partners.
+
+The keyword *diff* take a single atom type as its value.  Only atoms
+*I* of the that atom type will be selected.  Atoms *J* of all
+remaining atom types will be considered as potential swap partners.
+This includes the atom type specified with the *diff* keyword to
+account for self-diffusive hops between two atoms of the same type.
+
+The keyword *voro* is required.  Its *vID* value is the compute-ID of
```
sjplimp: Put the description on the *voro* keyword before the types/diff keywords, following the order it is listed at the top of the doc page

Also vID --> voro-ID

doc/src/fix_neighbor_swap.rst
```diff
@@ -0,0 +1,259 @@
...
+be selected than atoms of type 2.  If the *rates* keyword is not used,
+
+
+Restart, fix_modify, output, run start/stop, minimize info
+""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
+
+This fix writes the state of the fix to :doc:`binary restart files
+<restart>`.  This includes information about the random number generator
+seed, the next timestep for MC exchanges, the number of exchange
+attempts and successes, etc.  See the :doc:`read_restart <read_restart>`
```
sjplimp: Remove "etc", since it seems like you've listed everything ?

doc/src/fix_neighbor_swap.rst
```diff
@@ -0,0 +1,259 @@
...
+The keyword *rates* can modify the swap rate based on the type of atom
+*J*.  Ntype values must be specified, where Ntype = the number of atom
+types in the system.  Each value is used to scale the probability
+weighting given by the equation above.  In the third example command
+above, a simulation has 3 atoms types.  Atom *I*s of type 1 are
+eligible for swapping.  Swaps may occur with atom *J*s of all 3 types.
+Assuming all *J* atoms are equidistant from an atom *I*, *J* atoms of
+type 1 will be 3x more likely to be selected as a swap partner than
+atoms of type 2.  And *J* atoms of type 3 will be 6.5x more likely to
+be selected than atoms of type 2.  If the *rates* keyword is not used,
```
sjplimp, line 196: Incomplete final sentence - I think you mean something like ... not used, the probability for swapping with *J* atoms of all types is the same.

sjplimp (changes requested):
Looks good - just a few minor comments on the new doc page

doc/src/fix_neighbor_swap.rst
```diff
@@ -0,0 +1,259 @@
...
+
+fix neighbor/swap command
+=========================
+
+Syntax
+""""""
+
+.. code-block:: LAMMPS
+
+   fix ID group-ID neighbor/swap N X seed T R0 voro keyword values ...
```
sjplimp: word "voro" also vanishes here, b/c now a keyword ?

doc/src/fix_neighbor_swap.rst
```diff
@@ -0,0 +1,259 @@
...
+* seed = random # seed (positive integer)
+* T = scaling temperature of the MC swaps (temperature units)
+* R0 = scaling swap probability of the MC swaps (distance units)
+* voro = valid voronoi compute id (compute voronoi/atom)
+* two or more keyword/value pairs may be appended to args
+* keywords *types* and *diff* are mutually exclusive, but one must be specified
+* keyword = *types* or *diff* or *ke* or *region* or *rates*
+
+  .. parsed-literal::
+
```
sjplimp, line 26: *voro* value = voro-ID = ID of compute voronoi/atom command

sjplimp (changes requested):
added a couple more edits

doc/src/fix_neighbor_swap.rst
jtavenner-dev: It was a keyword initially, but current version has it as mandatory since otherwise is a required keyword. The compute ID input is in the class initialization not the options now. (lines 90-98 of fix_neighbor_swap.cpp)

doc/src/fix_neighbor_swap.rst
jtavenner-dev: See above comment, is not keyword since required.

doc/src/fix_neighbor_swap.rst
jtavenner-dev: Same as other voro edits.

doc/src/fix_neighbor_swap.rst
jtavenner-dev: Same as other voro edits.

doc/src/fix_neighbor_swap.rst
jtavenner-dev: Same as other voro edits.

doc/src/fix_neighbor_swap.rst
jtavenner-dev: Same as other voro edits.

doc/src/fix_neighbor_swap.rst
sjplimp: I don't care which way you do it, but I think as it stands that the description of the syntax and examples are inconsistent.

(a) The 3 examples show "voro voroN" as a keyword/arg pair, and the command at the top of the page shows "R0 voro keyword", so inconsistent.

(b) If voro is not a keyword, what are the 2 required keywords?  One is types or diff.  Is there another?  I'm thinking it used to be voro, so now only 1 required keyword.

(c) To avoid confusion, please do not use "voro" in the syntax description.  Use voro-ID, and state that it is the ID of a compute voronoi/atom command.

doc/src/fix_neighbor_swap.rst
jtavenner-dev: Finished sentence, thanks for catching.

doc/src/fix_neighbor_swap.rst
jtavenner-dev: Yes, updated sentence appropriately.

doc/src/fix_neighbor_swap.rst
jtavenner-dev: Thanks, I had not caught those inconsistencies in the current version. Have updated to be consistent in the latest push.

doc/src/fix_neighbor_swap.rst
```diff
@@ -0,0 +1,264 @@
...
+fix neighbor/swap command
+=========================
+
+Syntax
+""""""
+
+.. code-block:: LAMMPS
+
+   fix ID group-ID neighbor/swap N X seed T R0 voro keyword values ...
```
sjplimp: voro --> voro-ID

doc/src/fix_neighbor_swap.rst
```diff
@@ -0,0 +1,264 @@
...
+
+* ID, group-ID are documented in :doc:`fix <fix>` command
+* neighbor/swap = style name of this fix command
+* N = invoke this fix every N steps
+* X = number of swaps to attempt every N steps
+* seed = random # seed (positive integer)
+* T = scaling temperature of the MC swaps (temperature units)
+* R0 = scaling swap probability of the MC swaps (distance units)
+* voro-ID = valid voronoi compute id (compute voronoi/atom)
+* two or more keyword/value pairs may be appended to args
```
sjplimp: sorry - is it "two required" or just "one" now ?

sjplimp:
@jtavenner-dev I think we have converged - just address my final comment and Q and I will approve this for merging.
Thanks for developing this model and contributing it to LAMMPS, and for patience with my many (!) requests.

doc/src/fix_neighbor_swap.rst
jtavenner-dev: Yes, just one I had been searching for where the two required was noted. Have changed.

doc/src/fix_neighbor_swap.rst
jtavenner-dev: Updated.

jtavenner-dev:
Thanks for all the reviews and for making sure it makes sense to others reading it from the LAMMPS docs.

sjplimp (approved):

sjplimp:
@akohlmey I believe this one is ready to merge ...

PR #4427, opened 2024-12-27

akohlmey:
**Summary**

This pull request combines multiple small sets of changes and fixes.

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- continued checking that internal computes for temperature, pressure, or potential energy are not externally replaced by incompatible computes. Bugfixes for a couple of previously added/modified checks.
- workaround to silence bogus GCC warnings about overflowing buffer size in memcpy()
- LAMMPS-GUI updates for corresponding LAMMPS tutorial paper submission
- Highlight warnings and errors in LAMMPS-GUI output window
- use "command-line" consistently and drop "line" where not applicable
- Update, correct, and partially reformat "Speed GPU" and "Speed INTEL" pages

**Post Submission Checklist**


src/MACHDYN/atom_vec_smd.cpp
```diff
@@ -201,7 +201,7 @@ void AtomVecSMD::write_data_restricted_to_general()
 
   int nlocal = atom->nlocal;
   memory->create(x0_hold,nlocal,3,"atomvec:x0_hold");
-  if (nlocal) memcpy(&x0_hold[0][0],&x0[0][0],3*nlocal*sizeof(double));
+  if (nlocal) memcpy(&x0_hold[0][0],&x0[0][0],(3*nlocal*sizeof(double))&memcpymask);
```
akohlmey: @sjplimp @stanmoore1 this set of changes (inspired by a similar change in libfmt to work around the same kind of compiler issue) should silence the bogus compiler warnings. Please let me know if I missed a spot.

src/atom_vec.cpp
```diff
@@ -2330,19 +2330,19 @@ void AtomVec::write_data_restore_restricted()
   // no other write_data Velocities fields are Nx3 double arrays
 
   if (v_hold) {
-    memcpy(&v[0][0],&v_hold[0][0],3*nlocal*sizeof(double));
+    memcpy(&v[0][0],&v_hold[0][0],(3*nlocal*sizeof(double))&MEMCPYMASK);
```
sjplimp: just a Q about this - say nlocal is only a bit smaller than MAXINT (obviously unlikely) - does 3*nlocal
overflow before the multiply by sizeof(double) promotes it to a 64-bit result ?

sjplimp (approved):

src/atom_vec.cpp
akohlmey: > does 3*nlocal overflow before the multiply by sizeof(double) ?

Looks like it. Check out: https://godbolt.org/z/WPzzrThPz

sjplimp: So you're saying the MEMCPYMASK will prevent all the (bogus) compiler warnings?
But we should also (not in this PR) make sure we are not potentially causing these kinds of overflows?
Either by ordering the multiplies small to large, or by assigning the arg to a bigint (with appropriate casting)
and using the bigint as the arg?
We do the latter in other places in the code already, but undoubtably not everywhere.

akohlmey: > So you're saying the MEMCPYMASK will prevent all the (bogus) compiler warnings?

It does for me.  ...and it is independent of the overflow issue. It just reduces the possible size from 64-bits to 63-bits which GCC for unknown reasons has decided is the allowed maximum.

> make sure we are not potentially causing these kinds of overflows?

I am genuinely surprised it hasn't shown with any of the static code analysis tools we use (Coverity Scan, CodeQL, or compiler warnings with `-Wall -Wextra`). There were numerous overflow warnings that I have fixed over the years. Most often inconsequential ones, like in the `memory_usage()` functions. There are over 200 files where there is now a cast to double for the first number:
```
lammps/src$ grep -l bytes.*+=.*double *.cpp */*.cpp | wc -l
203
```

> Either by ordering the multiplies small to large, or by assigning the arg to a bigint (with appropriate casting)
and using the bigint as the arg?

Without parenthesis, expressions are evaluated left to right, so the simplest is to cast the first number or variable to the final data type. 

> We do the latter in other places in the code already, but undoubtably not everywhere.

There is too much code to check, so it needs some tool to automate this. Something like clang-tidy or a compiler warning should do. But then again, there are use cases, where the wrap around, specifically for integers is desired, so the check has to be more narrow, e.g. when passing something to a function that accepts a size_t argument or similar.

doc/src/Build_extras.rst
```diff
@@ -421,9 +421,9 @@ minutes to hours) to build.  Of course you only need to do that once.)
       cmake build system.  The ``lib/kim/Install.py`` script supports a
       ``CMAKE`` environment variable if the cmake executable is named other
       than ``cmake`` on your system.  Additional environment variables may be
-      provided on the command line for use by cmake.  For example, to use the
+      provided on the command-line for use by cmake.  For example, to use the
```
jtclemm: I thought the hyphen was only if it was an adjective. Could be wrong though.

akohlmey: I have not found an authoritative statement on this. It would not be the first inconsistency in english grammar or spelling :wink:. Feel free to make corrections as you see fit.

For several cases like this, I have reformulated the text to not use 'command-line' or 'command line'. I'll do the same for this.

PR #4431, opened 2025-01-08
Interactive Molecular Dynamics [IMD] Version 3 implementation

ljwoods2:
**Summary**

The [IMDv3 protocol](https://imdclient.readthedocs.io/en/latest/protocol_v3.html) allows analyzing trajectory data as it is generated without writing to disk by sending the data through a socket to an IMD client. Version 3 builds on the already implemented version 2 in 3 ways: 
- New packet types useful for analysis (64-bit time and integration step, velocity and force data, and box dimensions)
- Fixing the packet order within each simulation integration step (not required in v2 which was only used for visualization, but necessary for analysis to properly associate i.e. coordinates with a particular time)
- Sending the packet types the client can expect from the simulation engine in each frame at the beginning of the session (for parsing/error handling purposes).

This PR updates the `fix_imd` with the new version of the protocol without breaking existing LAMMPS scripts that use version 2 by adding an optional 'version' keyword that allows users to opt-in.

For broader context, streaming trajectories directly into MDAnalysis is a [major milestone for version 3.0 of MDA](https://www.mdanalysis.org/2023/10/25/towards_3.0/) since it will allow monitoring a simulation with python code as well as high-throughput, femtosecond-scale analysis that would be expensive to save to disk. Our team is concurrently developing NAMD and GROMACS implementations as well for greater interoperability.

The python MDAnalysis client for IMDv3 is available in the [imdclient repo](https://github.com/Becksteinlab/imdclient) and is already fully compatible with these changes.

**Related Issue(s)**

**Author(s)**

Author of these changes:

- Lawson Woods, School of Molecular Sciences at Arizona State University, lawsonw84@gmail.com

Other members of the development team for the MDAnalysis IMD client + protocol version 3 who can maintain these changes:

- Oliver Beckstein, Center for Biological Physics, Arizona State University, obeckste@asu.edu
- Amruthesh Thirumalaiswamy, School of Molecular Sciences, Arizona State University, athiru12@asu.edu
- Matthias Heyde, School of Molecular Sciences, Arizona State University, mheyden1@asu.edu
- Heekun Cho, School of Molecular Sciences, Arizona State University, hcho96@asu.edu

**Licensing**

**Backward Compatibility**

No changes will break backward compatibility- IMDv2 will be used by default unless the `version 3` keyword is provided to the fix.

**Implementation Notes**

Few changes are made to v2 code- only some memory deallocations, which were updated for efficiency and consistency with how memory is handled in the v3 part of the code, and removal of some unused packet parsing code for packets which are never received by the simulation engine (IMD_ENERGIES, IMD_FCOORDS, IMD_IOERROR, which are only ever *sent* by the engine) or are parsed only when a handshake occurs (IMD_GO) rather than generally during a simulation. These small changes make the implementation more consistent with the protocol and with the original GROMACS and NAMD implementations. Overall, no changes to the actual functionality of the original code were made.

The v3 code is split into separate methods to allow minimal changes to the original class. These methods (`handle_client_input_v3` and `handle_output_v3`) implement the new packets. 

One notable change is that v3 packets are sent to the client using `end_of_step` rather than `post_force` (as in v2) to avoid sending half-step velocities. This was done for consistency with the GROMACS and NAMD implementations.

We build these changes into a [container](https://github.com/Becksteinlab/streaming-md-docker) and run [regression tests](https://github.com/Becksteinlab/imdclient/blob/97cc13a891b1197e84fd78021679881657c482f9/imdclient/tests/test_lammps.py) on it against the MDAnalysis client for IMDv3, ensuring data that comes through the socket via IMD exactly aligns with disk trajectory output and that our new packets work as they are intended. 

No other features in LAMMPs should be impacted.

**Post Submission Checklist**

Example `.in` file that uses the fix changes: https://github.com/MDAnalysis/imd-workshop-2024/blob/main/streaming-101/LAMMPS/MD/lammps_v3.in

Example python script for parsing the simulation run with the above script (requires conda or pip installing `imdclient`): https://github.com/MDAnalysis/imd-workshop-2024/blob/main/streaming-101/LAMMPS/MDAnalysis/demo.ipynb


akohlmey:
@ljwoods2 thanks for your submission. I have made some small updates and cleanups to it. Unfortunately, you are using an "institutional" GitHub account, which means, I cannot push them directly into the pull request branch. Thus I have manufactured a pull request over here: https://github.com/Becksteinlab/lammps/pull/3
If you merge it, it will update the pull request branch and with those changes, the branch should be ready for merging after a positive review from a second LAMMPS core developer.

sjplimp:
Thanks for submitting this.  Are there updates/extensions or new input scripts which should be added to examples/PACKAGES/imd to illustrate how to use this new IMD version?  Or are the changes primarily internal and not at the user level?

akohlmey:
> One notable change is that v3 packets are sent to the client using `end_of_step` rather than `post_force` (as in v2) to avoid sending half-step velocities. This was done for consistency with the GROMACS and NAMD implementations.

@ljwoods2 This means, that for v3 communication you cannot correctly have forces added from the IMD client to the simulation via  fix imd. This is how using a haptic device with VMD with IMDv2 works:

[Haptic device] <=> [VRPN] <=> [VMD] <=> [MD engine/fix imd]

That is VMD receives the forces from the haptic device through the VRPN server and sends any force feedback information to the device via VRPN as well. This information is then scaled, if requested, and sent to the MD engine. Some more details on this should be in http://dx.doi.org/10.1007/978-3-642-17274-8_38

Any forces added at end_of_step are missing during `Fix::final_integrate()` for the second velocity half-update. They will be used during `Fix::initial_integrate()`, but after that forces are cleared.

For the immediate pull request, this details and the consequences should be documented in the fix imd documentation, specifically under "Restrictions". Perhaps, the IMDv3 protocol should not allow sending forces from the IMD client to the MD engine. So there could be a (single shot?) warning via `utils::logmesg()` , if that happens.

Longer term, you have to make a choice if you want to support this feature (and thus eventually also update VMD to IMDv3) or not. One way to support this could be to make a copy of the last set of received forces and re-add them in `Fix::post_force()`. Another way could be to switch back to communicating in ``Fix::post_force()`` but not send the current velocities, but make a copy during `Fix::initial_integrate()` and require that fix imd is coming before any time integration fixes (i.e. before they can do a half-update of the velocities).

ljwoods2:
@akohlmey I don't think I fully understand the problem- can you help me see what I'm missing?

The intention was that incoming IMDv3 packets from the client are parsed during `post_force`, including the `MDCOMM` force packet, which should then be applied immediately (inside of `handle_client_input_v3`). 

Outbound communication from LAMMPS to the client then occurs only in `end_of_step` (via `handle_output_v3`).

No forces can be added to the simulation during `end_of_step`- there can be forces collected and sent to the client for analysis/monitoring (`IMD_FORCES`), but not `IMD_MDCOMM` input forces parsed from the client.

Is the intended behavior I described incorrect, or would the intended behavior work, but I implemented it wrong?

ljwoods2:
@sjplimp I added an example along with MDAnalysis usage in 15999f6518249d6b3104c502b1928dda0fca476f

akohlmey:
> I don't think I fully understand the problem- can you help me see what I'm missing?

@ljwoods2 I have to apologize. I have been lazy and based my statements entirely on your statement in the description in the pull request description. I will have to carefully study the source code.

akohlmey:
> Is the intended behavior I described incorrect, or would the intended behavior work, but I implemented it wrong?

I carefully reviewed the source code (and made some more cosmetic changes) and can confirm that your changes are correct and my statements incorrect. There is nothing that needs to be changed for supporting interaction with sending forces to LAMMPS. Again, I have to apologize for being too lazy to check the details right away and thus avoid the confusion.

sjplimp (approved):

ljwoods2:
@akohlmey Thank you for taking the time to get this into shape and making quick work of it, it's greatly appreciated. We're super excited for people to be able to use this.

orbeckst:
Thank you to everyone!

src/MISC/fix_imd.cpp
```diff
@@ -1128,23 +1367,442 @@ void FixIMD::post_force(int /*vflag*/)
...
+          double yz = domain->yz;
+      for (int i = 0; i < nlocal; ++i) {
+        if (mask[i] & groupbit) {
+          int ix = (image[i] & IMGMASK) - IMGMAX;
+          int iy = (image[i] >> IMGBITS & IMGMASK) - IMGMAX;
+          int iz = (image[i] >> IMG2BITS) - IMGMAX;
+
+          if (domain->triclinic) {
+            buf[idx].tag = tag[i];
+            buf[idx].x = x[i][0]; + ix * xprd + iy * xy + iz * xz;
```
akohlmey, line 1676: > buf[idx].x = x[i][0]; + ix * xprd + iy * xy + iz * xz;

@ljwoods2 is the first semicolon here placed correctly??

This looks like you want to unwrap coordinates, but this code does not do it.
It will be the same as
`buf[idx].x = x[i][0];`

I am currently reviewing compiler warnings with warning level cranked to the maximum and came across warnings for this one and the following lines.

ljwoods2: @akohlmey You're absolutely right, this is a mistake. This is meant to be an unchanged copy of the same logic from `handle_step_v2` (lines 1343-1353):

```
if (domain->triclinic) {
            buf[nme].tag = tag[i];
            buf[nme].x   = x[i][0] + ix * xprd + iy * xy + iz * xz;
            buf[nme].y   = x[i][1] + iy * yprd + iz * yz;
            buf[nme].z   = x[i][2] + iz * zprd;
          } else {
            buf[nme].tag = tag[i];
            buf[nme].x   = x[i][0] + ix * xprd;
            buf[nme].y   = x[i][1] + iy * yprd;
            buf[nme].z   = x[i][2] + iz * zprd;
          }
```

Didn't catch this since our CI runs scripts with unwrap set to off, thank you for catching this

ljwoods2: quick PR to fix: #4448

akohlmey: > quick PR to fix: #4448

Thanks for confirming. FWIW, the change was already included in PR #4445

PR #4432, opened 2025-01-09
[Community project] Add short explanations and URL to them for common errors and warnings

akohlmey:
**Summary**

This pull request adds small paragraphs to errors that are frequently asked about and that do not have a simple single cause and thus need some background information to understand and correct them.

In addition, we should make a concerted effort to use the recently expanded Error class APIs: https://docs.lammps.org/latest/Developer_notes.html
https://docs.lammps.org/latest/run.html

This particularly applies to styles that can have a variety of arguments and references to other styles like pair style hybrid, compute reduce, fix ave/*, fix property/atom, compute property atom, dump style custom/cfg/xyz.
Errors that are flagged during a run (e.g. init(), setup(), compute(), post_force(), etc.) should be changed to
`error->all(FLERR, Error::NOLASTLINE, ...)` or `error->one(FLERR, Error::NOLASTLINE, ...)`.

The initial pull request implements one new utils::errorurl() case and a few calls to Error::all() or Error::one() to demonstrate the kind of changes to code and documentation required.

**Related Issue(s)**

https://matsci.org/t/what-are-the-most-confusing-error-messages-or-warnings-in-lammps/60131

**Author(s)**

The LAMMPS core developers

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

This makes use of the `utils::errorurl()` function which takes a number as an argument and outputs a string with a URL linking directly to the corresponding section in the (online) manual. Suggestions for suitable error messages can be easily found from looking at the most viewed topics in the LAMMPS forum. Most are about errors https://matsci.org/c/lammps/40?ascending=false&order=views

Here are some errors that were culled from the URL above.:
- [x] 1. Unknown identifier in data file (@akohlmey)
- [x] 2. Incorrect format in ... section of data file (@akohlmey)
- [x] 3. Illegal variable command: expected X arguments but found Y (@akohlmey), partially obsoleted by commit 4f6c3d12f7c59cfde2db6eed6eca498ee5629fcd
- [x] 4. Out of range atoms - cannot compute ... (@akohlmey)
- [ ] 5. Bond atoms missing
- [ ] 6. Non-numeric atom coords - simulation unstable
- [ ] 7. Non-numeric pressure - simulation unstable
- [ ] 8. Lost atoms
- [x] 9. Too many neighbor bins (@jtclemm)
- [ ] 10. Unrecognized pair style ... is part of ... package which is not enabled in this LAMMPS binary
- [ ] 11. All pair coeffs are not set
- [ ] 12. Segmentation fault - Address not mapped
- [ ] 13. Substitution for illegal variable
- [ ] 14. Bond atom missing in image check
- [x] 15. Cannot use neighbor bins - box size \<\< cutoff (@jtclemm)
- [ ] 16. Did not assign all atoms correctly
- [x] 17. Domain too large for neighbor bins (@jtclemm)
- [ ] 18. `hbondcheck` or `bondcheck` failed
- [ ] 19. Numeric index ... is out of bounds
- [ ] 20. Variable formula ... is accessed out-of-range
- [ ] 21. Incorrect args for pair coefficients (@megmcca)
- [ ] 22. X was not tallied on needed timestep
- [ ] 23. Molecule auto special bond generation overflow
- [ ] 24. Molecule topology/atom exceeds system topology/atom

This may also be extended to warnings:
- [ ] 25. Inconsistent image flags
- [ ] 26. No fixes with time integration, atoms won't move
- [ ] 27. System is not charge neutral, net charge = ...

Some files that would benefit from using the "pointer" to indicate which command line argument failed
- [ ] pair style hybrid and variants
- [x] pair style meam (@akohlmey)
- [ ] compute reduce and variants
- [ ] fix ave/* variants (some are already adapted in PR #4440)
- [ ] compute property/atom
- [ ] dump styles custom, cfg, xyz.

**Post Submission Checklist**


akohlmey:
@megmcca FYI, I've started adding some troubleshooting advice.
As @sjplimp mentioned in the call, we need to see where more detailed explanations are available: then we can either reduce the text in Error_details.rst to the essential bit and refer to the detailed docs or move content the other way.
Detailed explanations in the doc pages are often hard to find and not everybody reads all the way to the bottom.

So some redundancy may be acceptable, but then this also increases the maintenance effort to keep these in sync.

Also, you are now "owner" of this pull request and can re-assign it to me, once it is (sufficiently) complete.

sjplimp:
I am fine with cherry-picking the handful of useful error/warning messages from Sections 11.5 and 11.6 of the manual and putting them on a new manual page where all messages can be linked to by output from the running code.  Then just removing 11.5 and 11.6 from the manual.  Perhaps the current section 11.2 is a start in that direction?

akohlmey:
@megmcca if you need help to get started with this, please ping me on the developer slack.

akohlmey:
@megmcca FYI, I have submitted the current (incomplete) state of this PR branch as PR #4453 so we can have included bugfixes in the upcoming release, but also to possibly get some feedback on the improvements from users.

PR #4435, opened 2025-01-11
Hybrid scaled peratom

athomps:
**Summary**

Extends hybrid/scaled to handle atom-style variables as scale factors

**Related Issue(s)**

None

**Author(s)**

Aidan Thompson

**Licensing**

**Backward Compatibility**

None

**Implementation Notes**

This implementation only treats force contributions from a sub-style with atom-style scale factor, while potential energy and per-atom contributions are zero for the sub-style.  A fully consistent treatment of atom-weighted energy and force contributions would require modifying the code for the force calculation in each sub-style i.e. not general (see https://arxiv.org/abs/2411.03002)

**Post Submission Checklist**


athomps:
I was not able to reproduce the segfault on `MolPairStyle:hybrid-scaled` on my local machine.

akohlmey:
> I was not able to reproduce the segfault on `MolPairStyle:hybrid-scaled` on my local machine.

I was, but also you should be able to get useful information from valgrind using `ctest -R hybrid-scaled -T memcheck -V`

The problem is that "atomvar" was not saved and restored in the restart functions. That led to atomvar being a NULL pointer after a restart leading to segfaults when trying to access it. This is fixed in commit 494c2fc033035eb41e58d21585fe9b9853d6a5ac

FWIW: this also seems to be missing the corresponding implementation in `single()`. I only see changes to `compute()`.

src/pair_hybrid_scaled.cpp
```diff
@@ -32,9 +32,14 @@ using namespace LAMMPS_NS;
...
 PairHybridScaled::PairHybridScaled(LAMMPS *lmp) :
-    PairHybrid(lmp), fsum(nullptr), tsum(nullptr), scaleval(nullptr), scaleidx(nullptr)
+    PairHybrid(lmp), fsum(nullptr), tsum(nullptr), scaleval(nullptr), scaleidx(nullptr),
+    atomvar(nullptr), atomscale(nullptr)
 {
   nmaxfsum = -1;
+
+  // set comm size needed by this Pair (if atomscaleflag)
+
+  comm_forward = 1;
```
stanmoore1, line 42: Shouldn't `comm_forward = 1` depend on `atomscaleflag`? Or it doesn't matter?

src/pair_hybrid_scaled.cpp
```diff
@@ -66,18 +73,35 @@ void PairHybridScaled::compute(int eflag, int vflag)
...
+  int atomscaleflag = 0;
   if (nvars > 0) {
     auto vals = new double[nvars];
+    auto vars = new int[nvars];
     for (int k = 0; k < nvars; ++k) {
       int m = input->variable->find(scalevars[k].c_str());
       if (m < 0)
         error->all(FLERR, "Variable '{}' not found when updating scale factors", scalevars[k]);
-      vals[k] = input->variable->compute_equal(m);
+
```
stanmoore1, line 84: Could pull this out of compute so it isn't repeated every timestep, but maybe it doesn't matter.

stanmoore1 (approved):
I approve. I had a few minor questions/comments.

akohlmey:
@athomps if you want me to merge this, you have to unassign yourself and assign me. But first please address @stanmoore1's questions.

src/pair_hybrid_scaled.cpp
athomps: Yes, it could be set to zero if not `atomscaleflag`, but we don't set `atomscaleflag` until inside `compute()`. If I understand correctly, the size of `comm->maxforward` will usually be greater than 1, so there is no extra memory allocated.

src/pair_hybrid_scaled.cpp
athomps: This is necessary if the value of the variable changes between calls to `compute()`. I did not add this, it was there already.

athomps:
I addressed @stanmoore1 questions, it is ready to go.

stanmoore1 (approved):

PR #4438, opened 2025-01-15

### LINKED ISSUE #4436: Deprecate/Remove KNL support
Time to remove some outdated targets.

- [ ] Deprecate/Remove code
- [ ] Remove documentation
- [ ] Deprecate examples/benchmarks on website

References:
- https://www.phoronix.com/news/GCC-15-Drops-Intel-Xeon-Phi
- https://www.phoronix.com/news/LLVM-19-Drops-Xeon-Phi
- https://www.nersc.gov/news-publications/nersc-news/nersc-center-news/2023/nersc-and-the-hpc-community-bid-farewell-to-cori-supercomputer/

### LINKED ISSUE #4437: [BUG]  Cannot always obtain the correct neighbor list when `pair` contains different cutoff for different type.
**Summary**

`fix` and `compute` cannot always obtain the correct neighbor list when `pair` contains different cutoff for different type.

For example, type 1 has cutoff 3.6, and type 2 has cutoff 6.0. When a cutoff of 6.0 is required, the actual cutoff obtained for type 1 is still 3.6 (even if set through `comm_modify cutoff 10.0`), and only when a higher cutoff is required, such as 8.0, the correct neighbor list can be obtained.

This issue can lead to incorrect results during `compute rdf`.

LAMMPS version: stable_29Aug2024_update1
Platform: Windows 10

**Steps to Reproduce**

Here is the in file that generates incorrect result when calculating rdf due to this issue:

```text
units           metal
boundary        p p p
timestep        0.002

lattice         fcc 3.61
region          box block 0 5 0 5 0 5
create_box      2 box
create_atoms    1 box
mass            1 63.55
mass            2 63.55

velocity        all create 6000 4587281

comm_modify     cutoff 10.0

pair_style      lj/cut  3.6
pair_coeff      1 * 0.167 2.314  3.6
pair_coeff      2 * 0.167 2.314  6.0

fix             1 all nvt temp 6000 6000 0.2
run             10000

compute         rdf1 all rdf 1000 1 1 cutoff 6.0
compute         rdf2 all rdf 1000 1 1 cutoff 8.0
fix             2 all ave/time 1 10000 10000 c_rdf1[*] file rdf1.dat mode vector
fix             3 all ave/time 1 10000 10000 c_rdf2[*] file rdf2.dat mode vector
run             10000
```

rdf result:

Image

akohlmey:
**Summary**

This pull request combines multiple small changes and bugfixes.

**Related Issue(s)**

Closes #4436 
Fixes #4437

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- fix off-by-one bug when parsing the "inputs" keyword in compute reduce
- add deprecation message for Intel Xeon Phi and KNL accelerator support
- fix bug in morphing neighbor list builds: when a request with a cutoff  is processed, we need to also check the smallest possible pair-wise cutoff.
- address minor memory management issues flagged by static code analysis
- enhance the `Error::all()` and `Error::one()` methods for argument parsing, so it can in addition print the line of input how it was parsed and indicate which argument was causing the problem, but for other errors, even suppress repeating the last line of input
- some example uses for the enhanced error message APIs
- added sections to the documentation for both, users and developers. This explains the possible outputs and their meaning for the former and which error functions to call and how for developers.

**Post Submission Checklist**


jtclemm (approved):

athomps (approved):
The carat symbol pointing to bad argument is a very nice feature. One thing that might help new developers to quickly understand the new usage would be to replace the integer constants with more descriptive variable names e.g. in `fix_ave_time.cpp` replace `6` with `ifirst`.  It is true that `6` is a pretty good name all by itself, but something like `ifirst` would be even more informative.

akohlmey:
> integer constants with more descriptive variable names e.g. in `fix_ave_time.cpp` replace `6` with `ifirst`. It is true that `6` is a pretty good name all by itself, but something like `ifirst` would be even more informative.

`ifirst` would be confusing here since it is not the index of to the first argument, but the first which does not have a fixed position (index 0 - 5). Something like `ioffset` would be more appropriate. I have made that change and converted all errors to use the new style in that fix.

PR #4440, opened 2025-01-22

akohlmey:
**Summary**

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

This pull request removes backward compatibility with `compute_modify extra and `compute_modify dynamic`. Now the more accurate commands `compute_modify extra/dof and `compute_modify dynamic/dof` are required. This was present since 2017, so plenty of time has passed since.

**Implementation Notes**

- avoid multiple initialization and potential double free in dump image
- avoid integer overflow when computing size of memory allocation in fix imd
- add support for `modify_param()` functions in computes, similar to fixes. Compute `temp/drude` uses it.
- add check for missing drude atoms to compute temp/drude and stop with an error instead of a segmentation fault
- add the same check also to fix drude, so that inconsistent drude particle bonds can be identified
- remove a line of dead code from LAMMPS python wrapper
- download plumed-2.9.3 by default
- remove unnecessary CPU <--> GPU data movement when computing pressure for `fix npt/kk` which led to a huge slowdown on GPUs. This perf regression came from https://github.com/lammps/lammps/pull/4346.

**Post Submission Checklist**


src/compute_pressure.cpp
```diff
@@ -132,6 +133,11 @@ ComputePressure::ComputePressure(LAMMPS *lmp, int narg, char **arg) :
   vector = new double[size_vector];
   nvirial = 0;
   vptr = nullptr;
+
+  // KOKKOS
```
sjplimp, line 137: Is this what affected the Kokkos overhead slowdown?

sjplimp (approved):

src/compute_pressure.cpp
stanmoore1: Yes, `compute pressure` doesn't use any atom arrays, and needed to be marked as such, or else a ton of data movement between GPU and CPU occurred due to this change: [https://github.com/lammps/lammps/commit/8827dec5a91472eab3063fdb3108ab03db0dd327#diff-3b9190ecdb0f077e35caa96545a[…]8d6d765f335e39e30494911b8d6fR204](https://github.com/lammps/lammps/commit/8827dec5a91472eab3063fdb3108ab03db0dd327#diff-3b9190ecdb0f077e35caa96545a2db01ff558d6d765f335e39e30494911b8d6fR204)

PR #4441, opened 2025-01-23
Changes to LAMMPS for FitSNAP-ReaxFF

alphataubio:
**Summary**

- [x] ~~Howto ReaxFF [preview](https://alphataubio.com/fasthtml/Howto_reaxff.html)~~
- [x] ~~91 reaxff force fields added in `potentials` with doi links in howto to related journal articles~~
- [x] ~~`tools/reaxff/reaxff-format-ff.py` to properly reformat reaxff force fields copy/pasted from journal articles~~
- [ ] ~~`bond_scan` command~~
- [x] ~~`pair_coeff_reaxff()` python interface to set reaxff force field with a string instead of reading from a file~~
          failed test no *fmemopen* on windows 😡 , refactored FitSNAP-ReaxFF code to avoid this, no longer needed
- [ ] ~~`set_reaxff_parameters()` python interface to modify specific reaxff force field parameters~~
- [x] `fix acks2/kk` bugfix
- [ ] ~~`fix reaxff/species` output option from PR#4320~~
- [ ] ~~`reaxff-chenoweth2008a.ff` example figure 15 with fix reaxff/species output option~~
- [ ] ~~`reaxff-trnka2018.ff` NEB example figure S6~~
- [ ] ~~`reaxff-komissarov2021.ff` bond scan example figure 4~~

**Related Issue(s)**

- FitSNAP PR https://github.com/FitSNAP/FitSNAP/pull/266

- https://matsci.org/t/malloc-tiny-free-list-remove-ptr-internal-invariant-broken-next-ptr-of-prev/60351

**Author(s)**

@alphataubio 

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**


akohlmey:
@alphataubio FYI, there is no chance that all those ReaxFF potentials are going to be included into the LAMMPS distribution. It is a long-standing policy to only include a few potential files for use in the examples and to have an example for the syntax. Which also renders your ReaxFF howto effectively superfluous (I don't see any significant new information that is not already provided in the reaxff pair style documentation, only the massive list of potentials and references).

I suggest you convert that into a separate package, combined with the processing tool, and then we can put a link to that package into the documentation and on the LAMMPS homepage. Or you can contribute the potentials to the NIST interatomic potentials website: https://www.ctcms.nist.gov/potentials/

src/library.cpp
```diff
@@ -2146,6 +2150,245 @@ int lammps_map_atom(void *handle, const void *id)
...
+\endverbatim
+ *
+ * \param  handle  pointer to a previously created LAMMPS instance
+ * \param  id      void pointer to the atom ID (of data type tagint, i.e. 32-bit or 64-bit integer)
+ * \return         local atom index or -1 if the atom is not found or no map exists
+ * */
+
+
+
+void lammps_set_reaxff_atm_parameter(void *handle, int type, int parameter_index, double value) {
```
akohlmey, line 2168: @alphataubio you're going to hate me for this, too, but this is too specific an interface for too narrow a scope.
So in the interest of simpler maintenance and a more general scope, this needs to be done differently. We have a rather generic mechanism to get access to internal data from within LAMMPS and those are the "extract()" functions. And as it turns out, the library interface *already has*, e.g., `lammps_extract_pair()` and `lammps_extract_pair_dimension()`. So no addition needed.

To make it work for your case, you just need to expand the extract() function in the PairReaxff class to return the different arrays that you need. Since you get the true pointers, you can modify their values and no additional function calls are needed. However, you have to be *very* careful since there are some constraints about this internal data: it has to be symmetric, there are mixing rules applied, some values converted, and some values are derived from other values. It is very easy to shoot yourself in the foot.
See the file `reaxff_ffield.cpp`.

akohlmey (changes requested):
Since `Howto_reaxff.rst` is now gone, all the other modifications to the manual build can be removed as well.

The added features in the library interface can be based on existing functions, provided pair_reaxff.cpp is updated as needed and then most of the functionality from those library functions can be realized outside of LAMMPS.

The changes to fix acks2/reaxff seem unrelated and you have to discuss them with @stanmoore1.

akohlmey:
@alphataubio and while my reputation is already ruined, here is another comment that you may find irritating. Your commit messages of the kind "Update filename.cpp" are annoyingly useless for both reviewing pull requests and debugging and code maintenance. I can easily see from looking at the commit or pull requests, which files are modified, so the update statement is not adding any value. What is missing, however, is information about *what* in the file was changed and particularly *why* it was changed. This is useful information when debugging and trying to track down where something went wrong. The more explanatory information the commit message contains, the easier it is to find a proverbial smoking gun and also to understand what problem you are trying to solve, so that one can immediately determine if there are alternatives.

evoyiatzis:
@alphataubio regarding the collection of reaxff files, if you agree i can submit some of them to openkim knowledge base (https://openkim.org/) alongside a short description so that people can use them automatically.

src/library.cpp
alphataubio: > We have a rather generic mechanism

not generic enough

> And as it turns out, the library interface already has, e.g., lammps_extract_pair() and lammps_extract_pair_dimension(). So no addition needed.

extract_pair() only works with up to 2d array of doubles. i'd need 4d arrays of reax structs going through the cpython interface to do it the way you're asking. **AND** that still wouldn't work for 0-X-Y-0 torsion parameters. **PLUS** i'd have to rewrite all the reaxff_ffield.cpp code in python from scratch "to be symmetric, there are mixing rules applied, some values converted, and some values are derived from other values". Apologies that's not a can of 🪱 🪱 🪱 i'm willing to open when i'm already done writing it in cpp and it works *fast* and *tested*.

> It is very easy to shoot yourself in the foot.

i prefer not to eat the 🪣 of 🐛 🐞 🪳 you're proposing

> See the file reaxff_ffield.cpp.

i'm intimately familiar with reaxff_ffield.cpp having been through it at least a hundred times over the past month. i even rewrote it polymorphically in earlier versions of fitsnap-reaxff to pass a string from python instead of reading from a file. github compile test failed because windows doesn't have fmemopen(). i refactored fitsnap-reaxff instead to workaround this limitation of windows.

i'm closing this PR also 🪂 . i'll just put my version of library.cpp and core.py in the fitsnap repo and write instructions in the fitsnap-reaxff documentation to copy them into lammps/src to build a specialized version of lmp that works with fitsnap-reaxff.

please put the acks2/kk bugfix commit into your next small changes PR after @stanmoore1 approves it. it's only 3 lines. thank you.

alphataubio:
> while my reputation is already ruined

when you're in a hole stop digging

alphataubio:
> @alphataubio regarding the collection of reaxff files, if you agree i can submit some of them to openkim knowledge base (https://openkim.org/) alongside a short description so that people can use them automatically.

thanks @evoyiatzis for your help. im not done properly reformatting all the reaxff potentials copy/pasted from journal articles. I didn't copy the files available in commercial SCM software to avoid copyright issues. I only included potentials openly available in supplementary materials. My original intent was to have all of them bundled in lammps with doi links in documentation but openkim is already integrated with lammps AFAIK so that's a nice workaround you propose. 👏 👏 👏 

ill tag you from the fitsnap repo when it's done. in the meantime, you can start writing your summaries by looking at abstracts of journal articles in the table of fitsnap-reaxff doc (https://alphataubio.com/fitsnap/reaxff.html) with doi links in bibliography.

PR #4345, opened 2024-10-03
Update embedded copy of libfmt to version 11.0.2

akohlmey:
**Summary**

This updates the embedded version of libfmt

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer and the libfmt developers

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

N/A

**Implementation Notes**

This carries forward all LAMMPS customizations including the consteval skipping for C++20 and beyond (see PR #4343).

**Post Submission Checklist**

- [x] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [x] The source code follows the LAMMPS formatting guidelines
- [x] The feature has been verified to work with the conventional build system
- [x] The feature has been verified to work with the CMake based build system

PR #4442, opened 2025-01-23
Freeze fmt library at version 10.2.1 and add changes that prepare LAMMPS for transition to std::format

akohlmey:
**Summary**

This reverts the recent updates to fmtlib-11 (PR #4426, PR #4345)  back to version 10.2.1. Recent versions of fmtlib have caused increasing problems with compilers without providing features that we use. For version 10.2.1 we have implemented workarounds that are rather small changes. For version 11 these changes are more drastic.

Ultimately, we want to eventually drop bundling fmtlib and instead switch to using std::format as is included in C++20 and later. We can stick with version 10.2.1 until that time and won't need to update it again.

Also, there are some more changes included that make the code comply with the C++20 requirement (for both, fmt::format and std::format) that the format string is a constexpr. 

Finally, there is a new function utils::print() replacing fmt::print() that lifts some restrictions versus fmt::print() for compiling LAMMPS with C++20 and later. It will be required when switching to std::format since this does not include a function equivalent to fmt::print(). To help with that a new python script has been added to the style checker tools that can detect fmt::print() and - if enabled - also replace it with utils::print().

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/a

**Implementation Notes**

N/A

**Post Submission Checklist**


stanmoore1 (approved):
I approve

sjplimp (approved):

PR #4443, opened 2025-01-25
Reaxff/species issues

### LINKED ISSUE #4373: [BUG] Incorrect molecule IDs from fix reaxff/species
**Summary**

The molecule IDs returned by `fix reaxff/species` are sometimes incorrect.

LAMMPS version: LAMMPS (29 Aug 2024 - Update 1)
Platform: Ubuntu 20.04.2 LTS (fossa-ditto X84)
This error also occurs on Red Hat Enterprise Linux 8.5 (Ootpa).

**Steps to Reproduce**

I have attached a minimal working example (`in.test`) with the data file and the reaxff force field file needed to reproduce this issue (see [test.zip](https://github.com/user-attachments/files/17680105/test.zip)). It takes 14s to run with `mpirun -np 16 ~/lammps-29Aug2024/build/lmp -in in.test -log log.run` on my system. 

The data file contains 30 hydrocarbon molecules, which each contain 38 atoms, and 555 oxygen molecules. We can see from `dump.run` that the molecule IDs (the fourth column) are from 1 to 585 for timestep 230, with all molecules being assigned the correct ID. For instance, the first 38 atoms are assigned mol ID 1 and the last two atoms (an oxygen molecule) are assigned mol ID 585.

In contrast, there are errors in the molecule IDs for timestep 1150. For instance, the dump file contains
``` 
atom ID, element, q, mol ID, x, y, z
1997 O -0.00372084 459 -7.62194 -9.64018 13.2621
1998 O -0.0155002 2217 -7.11975 -9.47061 14.3768
...
2249 O -0.0068809 585 0.0830279 -3.45521 9.10559
2250 O -0.0172618 459 -0.990076 -3.51245 9.65433
```
Evidently from the positions, the oxygen with atom ID 1997 is not bonded with the oxygen with atom ID 2250. It is also unclear why the oxygen with ID 1998 is assigned a molecule ID of 2217 when there are only 585 molecules (as also confirmed by the species file output by `fix reaxff/species`). I'd have expected the oxygens with atom IDs 1997 and 1998 to have been assigned mol ID 459, as in the other timesteps in the dump file. Similarly, I'd have expected the oxygens with atom IDs 2249 and 2250 to have been assigned mol ID 585.

I should note that this error does not occur with this example when running with only one core.

I'd be happy to provide more examples if needed. I have a lot of dump files in which the molecule IDs are randomly wrong, i.e., at some time steps they appear to be correct and in others they appear randomly messed up.

### LINKED ISSUE #4398: [BUG] Incorrect centre of mass with fix reaxff/species position option
**Summary**

Following from #4373, there are a couple more inconsistencies with the `fix reaxff/species` command. In the documentation for fix reaxff/species, it states 

> The optional keyword position writes center-of-mass positions of each identified molecules to file filepos every posfreq timesteps. 

In addition, the species position file contains `CoM_x CoM_y CoM_z`. However, the positions given for each molecule in the file are actually average positions.

LAMMPS version: LAMMPS (29 Aug 2024 - Update 1)
Platform: Ubuntu 20.04.2 LTS (fossa-ditto X84)
This error also occurs on Red Hat Enterprise Linux 8.5 (Ootpa).

**Steps to Reproduce**

This can easily be reproduced by running the simple example of one water molecule in [reproduce.zip](https://github.com/user-attachments/files/17968783/reproduce.zip). The dump file at time step 1 is 
```
ITEM: TIMESTEP
1
ITEM: NUMBER OF ATOMS
3
ITEM: BOX BOUNDS pp pp pp
-5.0000000000000000e+01 5.0000000000000000e+01
-5.0000000000000000e+01 5.0000000000000000e+01
-5.0000000000000000e+01 5.0000000000000000e+01
ITEM: ATOMS id type xs ys zs
3 2 0.491929 0.5 0.505744
1 2 0.508071 0.5 0.505744
2 1 0.5 0.5 0.500004
```
and the species position file is
```
Timestep 1 NMole 1  NSpec 1  xlo -50.000000  xhi 50.000000  ylo -50.000000  yhi 50.000000  zlo -50.000000  zhi 50.000000
ID	Atom_Count	Type	Ave_q		CoM_x		CoM_y		CoM_z
1	3	H2O	0.00000000 	0.50000000 	0.50000000 	0.50383077
```
It is evident that (2 * 0.505744 + 0.500004) / 3 = 0.5038307, which is not the centre of mass z position. The documentation also states

> From the third line downward, each molecule writes a line of output containing the following information: molecule ID, number of atoms in this molecule, chemical formula, **total charge**, and center-of-mass xyz positions of this molecule.

However, this file gives average charge, rather than total charge.

This issue could easily be fixed by changing the documentation and updating `fix_reaxff_species.cpp` to write `av_x ax_y ax_z` instead of `CoM_x CoM_y CoM_z`. That being said, I think the centre of mass is probably more widely reported/used than the average position of the atoms in a molecule. In addition, I believe total charge is more relevant than average charge since the charges of each atom in a molecule can be largely different such that the average charge is not particularly useful, especially not without a measure of the spread of the charges. Therefore, I think what the documentation claims is probably the desired behaviour, but not the implemented behaviour.

jrgissing:
**Summary**

Fixes several bugs identified by @navlalli 

Fix incorrect treatment of 'molecule IDs' output by reaxff/species

Also, address 'center of mass' vs 'geometric center' and partial charge summation issues

**Related Issue(s)**

Fixes #4373
Fixes #4398

**Author(s)**

JG

**Licensing**

**Backward Compatibility**

code produces different values for Com and Tot_q, which previously had column title Ave_q in the reaxff/species position file, to be consistent with documentation

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/doc/src/fix_reaxff_species.rst
+++ b/doc/src/fix_reaxff_species.rst
@@ -200,8 +200,8 @@ The 2 values in the global vector are as follows:
 The per-atom vector stores the molecule ID for each atom as identified
 by the fix.  If an atom is not in a molecule, its ID will be 0.
 For atoms in the same molecule, the molecule ID for all of them
-will be the same and will be equal to the smallest atom ID of
-any atom in the molecule.
+will be the same, and molecule IDs will range from 1 to the number
+of molecules.
 
 No parameter of this fix can be used with the *start/stop* keywords of
 the :doc:`run <run>` command.
--- a/src/REAXFF/fix_reaxff_species.cpp
+++ b/src/REAXFF/fix_reaxff_species.cpp
@@ -947,6 +947,8 @@ void FixReaxFFSpecies::DeleteSpecies(int Nmole, int Nspec)
   int ndel, ndelone, count, count_tmp;
   int *Nameall;
   int *mask = atom->mask;
+  double *mass = atom->mass;
+  double *rmass = atom->rmass;
   double localmass, totalmass;
   std::string species_str;
 
@@ -999,7 +1001,8 @@ void FixReaxFFSpecies::DeleteSpecies(int Nmole, int Nspec)
         Name[itype]++;
         count++;
         marklist[nmarklist++] = i;
-        localmass += atom->mass[atom->type[i]];
+        if (rmass) localmass += rmass[i];
+        else localmass += atom->mass[atom->type[i]];
       }
     }
 
--- a/src/REAXFF/fix_reaxff_species.cpp
+++ b/src/REAXFF/fix_reaxff_species.cpp
@@ -836,7 +836,8 @@ void FixReaxFFSpecies::WritePos(int Nmole, int Nspec)
   int count, count_tmp, m, n, k;
   int *Nameall;
   int *mask = atom->mask;
-  double totq, totq_tmp, avx[3], avx_tmp, box[3], halfbox[3];
+  double *rmass = atom->rmass;
+  double totq, totq_tmp, com[3], com_tmp, thism, totm, box[3], halfbox[3];
   double **spec_atom = f_SPECBOND->array_atom;
 
   if (multipos) OpenPos();
@@ -864,7 +865,8 @@ void FixReaxFFSpecies::WritePos(int Nmole, int Nspec)
 
     count = 0;
     totq = 0.0;
-    for (n = 0; n < 3; n++) avx[n] = 0.0;
+    totm = 0.0;
+    for (n = 0; n < 3; n++) com[n] = 0.0;
     for (n = 0; n < nutypes; n++) Name[n] = 0;
 
     for (i = 0; i < nlocal; i++) {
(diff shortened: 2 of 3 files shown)


src/REAXFF/fix_reaxff_species.cpp
```diff
@@ -387,11 +381,25 @@ void FixReaxFFSpecies::init()
...
     f_SPECBOND = dynamic_cast<FixAveAtom *>(modify->add_fix(fixcmd));
+
+    // create a fix to point to fix_property_atom for storing clusterID
+    fixcmd = fmt::format("clusterID_{} all property/atom d_clusterID ghost yes", id);
+    f_clusterID = dynamic_cast<FixPropertyAtom *>(modify->add_fix(fixcmd));
+
+    // per-atom property for clusterID
+    int flag,cols;
+    int index1 = atom->find_custom("clusterID",flag,cols);
+    clusterID = atom->dvector[index1];
```
jtclemm, line 389: If this array grows, can its pointer change such that it's better to save the value of `index1` and redefine `clusterID` as needed?

jrgissing: I agree that would probably be safer

doc/src/fix_reaxff_species.rst
```diff
@@ -200,8 +200,8 @@ The 2 values in the global vector are as follows:
 The per-atom vector stores the molecule ID for each atom as identified
 by the fix.  If an atom is not in a molecule, its ID will be 0.
```
navlalli, line 201: Currently, this fix assigns a non-zero molecule ID to atoms that are not bonded to any other atoms. I don't think this is a problem provided

> If an atom is not in a molecule, its ID will be 0.

is updated.

akohlmey: @navlalli I disagree with your assessment. It has been a long-standing convention in LAMMPS to assign a molecule ID 0 to atoms that are not bonded to anything. So it would be worth the extra effort to ensure that what fix reaxff/species determines as molecule ID would be consistent with that.

navlalli: @akohlmey I assumed the current behaviour is the intended behaviour since it allows you to distinguish between single-atom species using molecule ID alone. This is useful, for example, when using the `position` keyword with `fix reaxff/species`, which creates a file that is sorted according to molecule ID at each time step.

sjplimp (approved):

sjplimp:
Nice work @jrgissing

PR #4447, opened 2025-01-29
Updates to recover failed post-merge tests

akohlmey:
**Summary**

Multiple issues that caused post-merge tests to fail after recent merges are addressed.

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer (Temple U)

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- Accumulated spelling and formatting issues in the manual were resolved that failed the documentation build
- Run and regression tests on the CPU failed due to dropping backward compatibility for `compute_modify extra` and `compute_modify dynamic`. This only affected the example inputs in the lammps-testing repository (see commit fd2ce671ad00d81cbc420c4949c2fef55d0e2b31). The example inputs in the LAMMPS distribution are already up-to-date.
- Correct dependencies for fix efield/lepton tests so that tests are not run if LEPTON package is missing. This caused failure of the GPU unit tests, since they don't include the LEPTON package.
- Failing clang compilation was fixed by including the libstdc++-12-dev package in Ubuntu 22.04LTS_gpu container in lammps-testing repo. This prohibited the update to Ubuntu 22.04LTS which will be needed after the next stable release
- Catch error exceptions now outside init_lammps() function and shut down output capture as needed. This avoids segfaults with "cannot have multiple output captures. Error is show with -V and without having to use TEST_ARGS=-v
- Set lmp pointer to NULL in cleanup_lammps() to force a crash on use after delete errors or avoid double deletes.
- Fix bug when setting up test environment for TestMliapPyUnified which caused a segfault. This still fails to fully reproduce lj/cut results.

**Post Submission Checklist**

### DIFF
--- a/doc/src/Build_basics.rst
+++ b/doc/src/Build_basics.rst
@@ -203,7 +203,7 @@ LAMMPS.
       check if the detected or selected compiler is compatible with the
       C++ support requirements of LAMMPS and stop with an error, if this
       is not the case.  A C++11 compatible compiler is currently
-      required, but a transition to require C++17 is in progess and
+      required, but a transition to require C++17 is in progress and
       planned to be completed in Summer 2025. Currently, setting
       ``-DLAMMPS_CXX11=yes`` is required when configuring with CMake while
       using a C++11 compatible compiler that does not support C++17,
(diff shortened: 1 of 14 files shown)


akohlmey:
@stanmoore1 the unittest refactor included here should do away with those cryptic unit test segfaults on output captures and instead print a meaningful error message, e.g. when the LAMMPS initialization failed due to a missing or incorrect prerequisite line in the yaml file for a test or when the syntax for a command has changed but a test not updated accordingly.

sjplimp (approved):

PR #4451, opened 2025-01-30
Step version strings for next feature release

akohlmey:
The next feature release is planned for 4 February 2025


akohlmey:
@stanmoore1 I have added some instructions for building a couple of the pre-compiled binary packages. If you have some time, it would be nice if you could try them out. You need apptainer/singularity installed locally (I have it installed as RPM package from Fedora) for building the static executable with a special musl-libc cross-compiler gcc (included in the container). For building the flatpak bundle, you need the flatpak and flatpak-builder RPM.
For testing you can just untar the static archive and try to use the executable (it should work on *any* x86_64 Linux box, no matter how old even though you are building on a very recent Linux distribution).

For the GUI package, you can install it on user level with `flatpak install --user <flatpak-bundle-name>`
See: https://docs.lammps.org/latest/Howto_lammps_gui.html#linux-on-x86-64

Let me know if you run into any problems.

sjplimp (approved):

.github/release_steps.md
```diff
@@ -1,42 +1,54 @@
 # LAMMPS Release Steps
 
-The following notes chronicle the current steps for preparing and publishing LAMMPS releases.  For
-definitions of LAMMPS versions and releases mean, please refer to [the corresponding section in the
-LAMMPS manual](https://docs.lammps.org/Manual_version.html).
+The following notes chronicle the current steps for preparing and
+publishing LAMMPS releases.  For definitions of LAMMPS versions and
+releases mean, please refer to [the corresponding section in the LAMMPS
```
jtclemm: Spotted another minor error, I think this was either supposed to be "For definitions of LAMMPS versions and *what* releases mean" or preferably "For definitions of LAMMPS versions and releases."

jtclemm (approved):

PR #4454, opened 2025-01-31
Add support to region plane for having the normal vector read from variables

evoyiatzis:
Summary

Add support to region plane for having the normal vector read from variables
It draws from the comments and review in #4390

Related Issue(s)

No related issue

Author(s)

E. Voyiatzis at NovaMechanics Ltd (email [evoyiatzis@gmail.com](mailto:evoyiatzis@gmail.com))

Licensing

Backward Compatibility

**Implementation Notes**

**Post Submission Checklist**


sjplimp:
@evoyiatzis Thanks for submitting - I will take a look.  I assume this also needs a doc page update?

evoyiatzis:
@sjplimp yes you are right. I updated the doc page accordingly. thanks for the time to review.

doc/src/region.rst
```diff
@@ -210,9 +210,11 @@ equal-style :doc:`variable <variable>`.  Likewise, for style *sphere*
 and *ellipsoid* the x-, y-, and z- coordinates of the center of the
 sphere/ellipsoid can be specified as an equal-style variable.  And for
 style *cylinder* the two center positions c1 and c2 for the location of
-the cylinder axes can be specified as a equal-style variable.  For style
-*cone* all properties can be defined via equal-style variables.  For
-style *plane* the point can be defined via equal-style variables.
+the cylinder axes can be specified as a equal-style variable.  For styles
+*cone* and *plane* all properties can be defined via equal-style variables.
+For style *plane*, there is the restriction that all three components
```
sjplimp: This statement is not as clear as the error message in the region_plane.cpp code.
So make it say either all constants or all defined by e-style variables.

doc/src/region.rst
```diff
@@ -210,9 +210,11 @@ equal-style :doc:`variable <variable>`.  Likewise, for style *sphere*
...
+of the direction normal to the plane should be either constant or
+defined via equal-style variables.
 
 If the value is a variable, it should be specified as v_name, where
```
sjplimp, line 220: Somewhere on the region doc page I think you should add a Note (which will appear in a blue box, see the HTML version of the region doc page).  Which cautions users whenever they use a variable to define a region property, to ensure that the variable does not cause any part of the region boundary to move too far within a single timestep. "Too far" would be more than a small fraction of a particle "size".  Otherwise bad dynamics can result.  The Note could also say this is true of the move/rotate options.  It could also highlight region styles for which this is easy to mess up.  E.g. region plane normals (this PR), since a small change in normal could shift the region surface far away from the region point by a large displacement.

sjplimp:
@evoyiatzis I will approve this PR after you amend the doc page in 2 places -- see my comments - thanks

doc/src/region.rst
evoyiatzis: @sjplimp what is a particle "size"? The simulations are usually carried out for point particles (such as atoms) which have no radius - of course granular materials are an obvious exception.

sjplimp: I was using "size" generically to mean the approx distance of closest approach for 2 particles, but it would be good to clarify that.  So sigma for LJ particles, or diameter for granular particles, etc.  Small fraction of that distance should ensure there is no large jump in the wall force in a single timestep,

evoyiatzis:
@sjplimp I have amended the doc page according to your comments. Is the manual clear enough now?

doc/src/region.rst
```diff
@@ -225,6 +226,21 @@ keywords for the simulation box parameters and timestep and elapsed
 time.  Thus it is easy to specify a time-dependent radius or have
 a time dependent position of the sphere or cylinder region.
 
+.. note::
+
+   Whenever a region property, such as a coordinate or an upper/lower
+   bound, is defined via an equal-style variable, it must be ensured
+   that the variable does not cause any of the region boundaries to move
+   too far within a single timestep. Otherwise, bad dynamics will occur.
+   By too far it is meant a small fraction of the approximate distance of
```
sjplimp: "Too far" means a small ...

doc/src/region.rst
```diff
@@ -225,6 +226,21 @@ keywords for the simulation box parameters and timestep and elapsed
 time.  Thus it is easy to specify a time-dependent radius or have
 a time dependent position of the sphere or cylinder region.
 
+.. note::
+
+   Whenever a region property, such as a coordinate or an upper/lower
+   bound, is defined via an equal-style variable, it must be ensured
```
sjplimp: ... via an equal-style variable, the variable should not cause any of the region boundaries ...

sjplimp (commented):
thanks for the note - just suggested a couple re-wordings
I'll go ahead and approve.

sjplimp (approved):

PR #4456, opened 2025-01-31
[Community project] Add short explanations and URL to them for common errors and warnings

akohlmey:
**Summary**

This pull request adds small paragraphs to errors that are frequently asked about and that do not have a simple single cause and thus need some background information to understand and correct them.

In addition, we should make a concerted effort to use the recently expanded Error class APIs: https://docs.lammps.org/latest/Developer_notes.html
https://docs.lammps.org/latest/run.html

This particularly applies to styles that can have a variety of arguments and references to other styles like pair style hybrid, compute reduce, fix ave/*, fix property/atom, compute property atom, dump style custom/cfg/xyz.
Errors that are flagged during a run (e.g. init(), setup(), compute(), post_force(), etc.) should be changed to
`error->all(FLERR, Error::NOLASTLINE, ...)` or `error->one(FLERR, Error::NOLASTLINE, ...)`.

The initial pull request implements one new utils::errorurl() case and a few calls to Error::all() or Error::one() to demonstrate the kind of changes to code and documentation required.

**Related Issue(s)**

https://matsci.org/t/what-are-the-most-confusing-error-messages-or-warnings-in-lammps/60131

**Author(s)**

The LAMMPS core developers

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

This makes use of the `utils::errorurl()` function which takes a number as an argument and outputs a string with a URL linking directly to the corresponding section in the (online) manual. Suggestions for suitable error messages can be easily found from looking at the most viewed topics in the LAMMPS forum. Most are about errors https://matsci.org/c/lammps/40?ascending=false&order=views

**Important Note:** Please use the number of the entry in the list below as the utils::errorurl() number, so there is no conflict from changes by different people.

Here are some errors that were culled from the URL above.:
- [x] 1. Unknown identifier in data file (@akohlmey)
- [x] 2. Incorrect format in ... section of data file (@akohlmey)
- [x] 3. Illegal variable command: expected X arguments but found Y (@akohlmey), partially obsoleted by commit 4f6c3d12f7c59cfde2db6eed6eca498ee5629fcd
- [x] 4. Out of range atoms - cannot compute ... (@akohlmey)
- [x] 5. Bond atoms missing (@jtclemm)
- [x] 6. Non-numeric atom coords - simulation unstable (@megmcca)
- [x] 7. Non-numeric pressure - simulation unstable (@megmcca)
- [x] 8. Lost atoms (@megmcca)
- [x] 9. Too many neighbor bins (@jtclemm)
- [x] 10. Unrecognized pair style ... is part of ... package which is not enabled in this LAMMPS binary (@megmcca)
- [ ] 11. All pair coeffs are not set (@megmcca)
- [x] Segmentation fault - Address not mapped (@akohlmey), handled by the OS. added a generic paragraph
- [x] 12. fmt::format_error (@akohlmey)
- [x] 13. Substitution for illegal variable (@megmcca)
- [x] 14. Bond atom missing in image check (@akohlmey)
- [x] 15. Cannot use neighbor bins - box size \<\< cutoff (@jtclemm)
- [x] 16. Did not assign all atoms correctly (@akohlmey)
- [x] 17. Domain too large for neighbor bins (@jtclemm)
- [x] 18. `hbondcheck` or `bondcheck` failed (@akohlmey)
- [x] 19. Numeric index ... is out of bounds (@akohlmey)
- [x] 20. Variable formula ... is accessed out-of-range (@akohlmey)
- [x] 21. Incorrect args for pair coefficients (@megmcca)
- [x] 22. X was not tallied on needed timestep (@athomps)
- [x] 23. Molecule auto special bond generation overflow (@akohlmey)
- [x] 24. Molecule topology type exceeds system topology type (@akohlmey)
- [x] 25. Molecule topology/atom exceeds system topology/atom (@akohlmey)
- [x] 30. Variable evaluation before simulation box is defined (@akohlmey)
- [x] 31. Invalid thermo keyword 'X' in variable formula (@akohlmey)

This may also be extended to warnings:
- [x] 26. Molecule attributes do not match system attributes (@akohlmey)
- [x] 27. Inconsistent image flags (@akohlmey)
- [x] 28. No fixes with time integration, atoms won't move (@megmcca)
- [x] 29. System is not charge neutral, net charge = ... (@jmgoff )

Some files that would benefit from using the "pointer" to indicate which command line argument failed
- [ ] pair style hybrid and variants (@megmcca)
- [x] pair style meam (@akohlmey)
- [x] fix ave/* variants (@akohlmey)  (some were adapted in PR #4440, some in PR #4453, and the rest in PR #4459)

**Post Submission Checklist**


megmcca:
@akohlmey the initial explanation i've written for error 21 "Incorrect args for pair coefficients" could apply equally well to bond, angle, dihedral, and improper coefficient error messages as well. 
if there's no other suggestion on how to handle those, i'll add those in next

akohlmey:
@megmcca I've resolved the merge conflict with upstream and added some words about segmentation faults to the "Troubleshooting" section. There is a problem with printing an utils::errorurl() line from a segfault, since this is triggered from the outside. That said, we could could install a signal handler and make that throw an exception with a suitable text.

I vaguely recall a discussion some time ago with @stanmoore1 about doing something like this. My recollection is that at the time it didn't seem to be that good an idea, but since now all errors throw a C++ exception, this would be much easier to handle (like Error::one()) and there could be other uses for signal handlers like forcing to write a restart at the next step, or expire a timer like it is done with fix halt, or forcing some output about the progress of the simulation, or catching floating-point exceptions, or ???. The main challenge is still to make it work in parallel, i.e. there must be a systematic way to setup all non-terminal signal handlers, so that the information is made consistent across all MPI ranks.

akohlmey:
@athomps FYI, you used the wrong index (21) for your addition. Also, it was missing the added calls to utils::errorurl() in the code. The latter could be easily added with the creative use of the "sed" command, so I did that right away.

sjplimp (approved):

megmcca:
i bet i've accidently introduced a typo into the detailed errors page. i'll resolve it today.

akohlmey:
@megmcca Please have a final look at everything. This should now be very close to being ready for merge.

akohlmey:
@megmcca as a reminder of the "github development protocol". You need to assign the PR to me if you want me to merge it.

jtclemm:
@megmcca @akohlmey I just proofread errors_details.rst to fix typos and try to improve clarity.

In particular, I had trouble understanding some phrases in hint 12 and errors 1, 12, 13, and 20 so I rewrote them. It'd be great if you could look at those edits to make sure I didn't change the intended meaning.

Regarding error 20, I also doubted the claim about indexing that "any value smaller than 1 will trigger [an] error". Is this supposed to be true? Last week I absentmindedly tried to access the last element of a fix as `f_myfix[-1]` and I'm pretty sure LAMMPS just outputted zeros.

jtclemm (approved):

akohlmey:
> Last week I absentmindedly tried to access the last element of a fix as `f_myfix[-1]` and I'm pretty sure LAMMPS just outputted zeros.

That would be a bug. We should track it down and address it. Can you create a reproducer input for this?

jtclemm:
> > Last week I absentmindedly tried to access the last element of a fix as `f_myfix[-1]` and I'm pretty sure LAMMPS just outputted zeros.
> 
> That would be a bug. We should track it down and address it. Can you create a reproducer input for this?

Sure, it was when printing a compute/fix value using `thermo_style`, e.g.:

```
# 3d Lennard-Jones melt

units           lj
atom_style      atomic

lattice         fcc 0.8442
create_box      1 box
create_atoms    1 box
mass            1 1.0

pair_style      lj/cut 2.5

compute         1 all msd
fix             1 all nve
thermo_style    custom step c_1[-1]

run             250
```

akohlmey:
> Sure, it was when printing a compute/fix value using `thermo_style`, e.g.:

Thanks. Yes, the logic for the checks in Thermo is a bit strange for historic reasons (this is a part of LAMMPS that has been around a long time and thus a bit convoluted). Fortunately, we can replace those checks with cleaner checks thanks to using the ArgInfo class for parsing compute/fix/etc references.

With fixes it it may return dummy values, but for computes, you can easily trigger segfaults.

akohlmey:
@jtclemm I've tried to update the logic and checking in thermo data access to computes, fixes and variables. It should now catch more cases. For fixes and variables with variable length vectors or arrays, I have left the behavior of returning 0.0 for (upper) out-of-range access in place. I've also tried to add "pointers for custom thermo_style commands.

jtclemm:
> @jtclemm I've tried to update the logic and checking in thermo data access to computes, fixes and variables. It should now catch more cases. For fixes and variables with variable length vectors or arrays, I have left the behavior of returning 0.0 for (upper) out-of-range access in place. I've also tried to add "pointers for custom thermo_style commands.

Cool, thanks! It didn't take me very long to spot my mistake and correct it, but an error message will definitely save me time in the future when I almost certainly make the same mistake again. I'm too used to Python.

akohlmey:
@megmcca I am assuming control over the pull request now and will merge it and then create a new branch and PR for any remaining "spring cleaning" work related to error messages. For example we can prune the existing lists with comments on error messages and warnings. In order to have some feedback about problems we may have introduced, we need to get this to the public. Sorry for bending the rules a little bit and thanks for all the effort you have invested in this and generally for taking initiative.

PR #4459, opened 2025-02-01

### LINKED ISSUE #4452: [BUG] _Issues from defining `__INTEL_COMPILER` as `__INTEL_LLVM_COMPILER`_
**Summary**

`__INTEL_COMPILER` defined as `__INTEL_LLVM_COMPILER` causes problems for use with oneDPL, and may cause other problems for those using features unique to `icc`, but missing from `ic[p]x`.

**Details**
oneDPL recently had an issue created: https://github.com/uxlfoundation/oneDPL/issues/2026

Attempting to compile LAMMPS for Intel GPU through Kokkos faced compile time errors.

I found within LAMMPS this line of code:
https://github.com/lammps/lammps/blob/48893236ecc13b95b5c2bdcce99fa635067a2a97/src/INTEL/intel_preprocess.h#L23

This is not recommended and causes problems for oneDPL (and likely other codes which have optimizations enabled based on compiler version).

`__INTEL_LLVM_COMPILER` is a predefined macro for the intel llvm compiler  `ic[p]x` as described [here](https://www.intel.com/content/www/us/en/developer/articles/technical/use-predefined-macros-for-specific-code-for-intel-dpcpp-compiler-intel-cpp-compiler-intel-cpp-compiler-classic.html).
`__INTEL_COMPILER` is a predefined macro for intel C++ compiler `icc` as described [here](https://www.intel.com/content/www/us/en/docs/cpp-compiler/developer-guide-reference/2021-10/use-predefined-macros-to-specify-intel-compilers.html).

These are two different compilers with different versioning spaces and different feature sets.  There exist some features within `icc` which do not exist within `ic[p]x`.

One example of such a feature that is in `icc` but not `ic[p]x` is `#pragma omp simd early_exit` as described [here](https://www.intel.com/content/www/us/en/docs/cpp-compiler/developer-guide-reference/2021-10/omp-simd-early-exit.html).  There are others as well like `#pragma omp ordered simd monotonic`.  

oneDPL uses these predefined macros to detect specific versions of specific compilers and enable certain codepaths for optimization.  Modifying this macro forces oneDPL to erroneously detect the `icc` compiler and enable a codepath which is not supported from `ic[p]x`.

Please see the oneDPL issue linked above for more details on the specific problem which arose.

Its unclear the purpose of this definition in your codebase, but I would recommend removing it and making a more specific fix without modifying these specific macros.

### LINKED ISSUE #4457: [BUG] Error in creating atoms using a mol file with many mol ids
**Summary**

When creating atoms using a single molecule file that contains many mol ids, one can easily trigger a seg fault. Create atoms will assign molecule IDs to all of the atoms created, but the loop is faulty and it continues to assign gradually increasing mol IDs to nonexistant atoms - eventually seg faulting if it exceeds the buffer on peratom arrays. Loop here:
 https://github.com/lammps/lammps/blob/ea1607f1d8a70941cc675f18c4255c25c95c459f/src/create_atoms.cpp#L580C1-L583C41

I'll poke around a bit more, but I'm unfamiliar with the molecule code so I'm not sure I can identify the appropriate solution.

Current development branch

**Expected Behavior**

The loop in create atoms should abort once all mol IDs are assigned and not seg fault.

**Actual Behavior**

A seg fault.

**Steps to Reproduce**

The attached molecule file is the smallest example I can get to crash on my system. 

[mol.txt](https://github.com/user-attachments/files/18624128/mol.txt)
[input.txt](https://github.com/user-attachments/files/18624129/input.txt)

akohlmey:
**Summary**

This pull request combines multiple small fixes and changes

**Related Issue(s)**

Fixes #4457 
Fixes #4452

**Author(s)**

Axel Kohlmeyer, Temple U
Joel Clemmer, SNL

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- bugfix for creating molecules with multiple molecule IDs from a single molecule file to avoid segfaults and memory corruption
- plug some memory and file descriptor leaks detected by running ctest with -T memcheck
- do not define __INTEL_COMPILER to _INTEL_LLVM_COMPILER. This breaks any code that contains features that will only work for the Intel compiler classic. Instead we check for either.
- Renumbering detailed error descriptions and utils::errorurl() calls to match numbers in list of PR #4456 
- Add some missing calls to utils::errorurl() in packages
- Simplify appending string return by utils::errorurl()
- Add some more errorurl() cases related to the molecule command
- Modernize the error message handling in the molecule command and specifically for pointing to molecule file where an error happened during reading of the file
- Remove `#include <cstdbool>` which is deprecated in C++17 and removed in C++20 and not needed anyway
- Add yaml files to test hbond/dreiding potentials
- remove some dead code in recently added files of the GRANULAR package
- update reference log files for several ReaxFF example inputs
- remove multi/old from comm_modify and neighbor command documentation. Add deprecation warning when used.

**Post Submission Checklist**


jtclemm:
I agree it has been long enough and we can start to move towards deprecating *multi/old*. However, could we add a note on the deprecation warnings to please contact the developers if there's a reason to keep it. I don't think we expect it to ever be faster than *multi* and I can't think of any systems which would support *multi/old* but not *multi*, but it's possible.

doc/src/minimize.rst
```diff
@@ -89,8 +89,8 @@ be more robust than previous line searches we have tried.  The
 backtracking method is described in Nocedal and Wright's Numerical
 Optimization (Procedure 3.1 on p 41).
 
-The :doc:`minimization styles <min_style>` *quickmin*, *fire* and
-*fire/old* perform damped dynamics using an Euler integration step.  Thus
+The :doc:`minimization styles <min_style>` *quickmin*, and *fire*
```
jtclemm: An extra comma snuck in between *quickmin* and *fire*.

akohlmey: No accident. This is called an "Oxford comma" (or serial comma): https://en.wikipedia.org/wiki/Serial_comma

jtclemm: I am a big proponent of oxford commas, but I was referring to the fact that there are now only 2 styles listed.

stanmoore1: Right, oxford commas are only used for a list of 3 or more.

akohlmey: Ooops. Didn't notice that. Thanks for noticing.

jtclemm (approved):
Thanks for the extra note on *multi/old*, hopefully no one contacts about it!

sjplimp (approved):

PR #4466, opened 2025-02-06

akohlmey:
**Summary**

This pull request combined multiple small changes and fixes

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- sync C++ standard selection and settings for plugin builds with main CMakeLists.txt file
- Replace non-standard variable length arrays in ML-RANN package with std::vector<double> and std::vector<std::vector<double>>
- Update information about steps for making a LAMMPS release
- Update main README file
- Update information and CMake script for building the LAMMPS manual
- Make SafeFilePtr class generally available in new safe_pointers.h header and apply to a few more use cases
- Fix rare bug with KOKKOS package on GPUs when `exchange comm` was on host, and `fix balance` was used. With `pair pace/kk`, it manifest as an abort with this message `"Encountered very small distance. Stopping."`

**Post Submission Checklist**


evoyiatzis:
In fix_adapt.cpp line 431, I think the error message should be "for bond hybrid" instead of "for pair hybrid". Similarly, in line 465 the error message should be "for angle hybrid"

akohlmey:
> In fix_adapt.cpp line 431, I think the error message should be "for bond hybrid" instead of "for pair hybrid". Similarly, in line 465 the error message should be "for angle hybrid"

@evoyiatzis Thanks for the note. Should be corrected in commit 31697bac8d41566005609c4ba6cb84c625e11010

src/angle_write.cpp
```diff
@@ -79,13 +80,13 @@ void AngleWrite::command(int narg, char **arg)
...
+  SafeFilePtr fp;
   std::string coeffs_file = table_file + ".tmp.coeffs";
   if (comm->me == 0) {
 
-    fp = fopen(coeffs_file.c_str(), "w");
-    force->angle->write_data(fp);
-    fclose(fp);
+    FILE *coeffs = fopen(coeffs_file.c_str(), "w");
+    force->angle->write_data(coeffs);
+    fclose(coeffs);
```
rbberger:
```suggestion
    {
      SafeFilePtr coeffs = fopen(coeffs_file.c_str(), "w");
      force->angle->write_data(coeffs);
    }
```

src/safe_pointers.h
```diff
@@ -0,0 +1,57 @@
...
+  SafeFilePtr &operator=(const SafeFilePtr &) = delete;
+
+  ~SafeFilePtr()
+  {
+    if (fp) fclose(fp);
+  }
+
+  SafeFilePtr &operator=(FILE *_fp)
+  {
+    fp = _fp;
```
rbberger, line 50: Creates a dangling handle if `fp` if it already set. I got a suggestion for all of this, but more tomorrow.

src/safe_pointers.h
```diff
@@ -0,0 +1,58 @@
...
+  {
+    if (fp && (fp != _fp)) fclose(fp);
+    fp = _fp;
+    return *this;
+  }
+  operator FILE *() const { return fp; }
+
+ private:
+  FILE *fp;
+};
```
rbberger: Here is a replacement that just uses `std::unique_ptr` (needs `#include <memory>` on top). See https://godbolt.org/z/enK4jYY57 for validation code.
```suggestion
class SafeFilePtr {
 public:
  SafeFilePtr(FILE * f) : fp(f, std::fclose) {}
  SafeFilePtr() : SafeFilePtr(nullptr) {}
  SafeFilePtr(SafeFilePtr &) = delete;
  SafeFilePtr& operator=(SafeFilePtr&) = delete;
  
  operator FILE*() const {
    return fp.get();
  }

 private:
  std::unique_ptr<FILE, int (*)(FILE*)> fp;
};
```

src/safe_pointers.h
rbberger: See below.

src/safe_pointers.h
akohlmey: Thanks for the suggestion. It is quite sophisticated. In this case, however, I prefer the more explicit version for better readability and accessibility for less experienced C++ programmers.
I am open to revisit this for other use cases of the same programming pattern. After all, I've named the new header file "safe_pointers.h" because I expect that other, similar will be added over time.

rbberger: :roll_eyes:

athomps (approved):
I approve.

athomps (approved):
I approve.

Issue #4350, opened 2024-10-06
[Feature Request] Support dihedral and improper changes in fix adapt


evoyiatzis:
**Summary**

It would be useful to extent fix adapt in a way that on-the-fly modifications / changes in the parameters of dihedral/improper potentials are supported.

**Detailed Description**

Similar to attributes bond and angle, two new attributes dihedral and improper would be added so that the user can modify / change the parameters of dihedral and improper potentials. The code in fix adapt should be expanded in similar way that was modified to cater for the angle attribute (PR Fix adapt angle #3229). The dihedral and improper classes should get an extract() method, something like 

virtual void *extract(const char *, int &) { return nullptr; }

and each dihedral / improper potential would override  accordingly.

I can add the extract method for several dihedral/improper potentials and the corresponding unit tests in the next weeks if this addition would be useful even without the modifications of fix adapt.

PR #4467, opened 2025-02-06
extend fix adapt to treat improper angles

evoyiatzis:
**Summary**

Add support to fix adapt for on-the-fly adjustment of improper style parameters. Similar to bond and angle style params. Support for 9 improper styles was added: amoeba, class2, cossq, cvff, distance, distharm, fourier, harmonic, inversion/harmonic, ring, umbrella, sqdistharm.

**Related Issue(s)**

Addresses partially feature request #4350

**Author(s)**

E. Voyiatzis at NovaMechanics Ltd (email evoyiatzis@gmail.com)

**Licensing**

**Backward Compatibility**

**Implementation Notes**

The implementation is heavily based on PR Fix adapt angle #3229 

Currently, all improper styles have fixed number of parameters. Thus, the reinitflag is always equal to 1 (it would be 0 if the number of parameters would vary) and the reinit method is never called. The implementation is useful only if in the future there is an addition of an improper style with variable number of parameters. 

**Post Submission Checklist**


akohlmey:
@evoyiatzis two comments/questions
- are you *trying* to kill the machines of the LAMMPS testing cluster at Temple? There is no need to git push after *every* commit. Each push will trigger a new full test run with over 50 individual test jobs, plus over 10 runs on GitHub runners. Please note that this is **very** old hardware that we have recycled after it was used in an HPC cluster for more than eight years and has been now running as a Jenkins cluster about just as long a time. It will be easier and faster to test compilation and selected unit tests locally, anyway, and then push to GitHub only every once in a while. Also, by keeping those machines busy, you create a backlog and prevent other tests from running for a long time. Of course, if you send us some significant $$s to replace those machines with new, more capable hardware, we would not have that issue.
- before making those many rather invasive changes to many force styles, it would be better to contact the LAMMPS developers and see if they are comfortable with them. Otherwise, you may be wasting a lot of time (plus the effort to revert everything).

evoyiatzis:
@akohlmey 

- apologies for abusing the system and causing issues to other users
- no invasive change was actually required. Renaming variables in several force styles was a wrong choice to satisfy a kind of personal syntactic sugar. Wasting my time and effort in this case should teach me to do better in the future although I am moving out of science / coding very soon.

sjplimp (approved):

PR #4469, opened 2025-02-10
WIP: Revise and Modernise MEAM

martok:
**Summary**

When I did the original Fortran->C++ translation, my C++ skill was... not great, so we mostly ended up with a C-ish literal translation. In the years since I've learned a bit (at least what Mozilla thinks is good C++), so my plan is to rework everything to be more maintainable and extensible.

This will take a while (next stable release, probably), but I'm creating the PR in advance to get early visibility - consider it as an RFC for now. Frequent rebase intended.

Planned (so far):

- [ ] Logical modules with enough compiler visibility for sensible inlining
- [ ] Reduce manual unrolls
- [ ] De-spaghettify reference lattice handling
- [ ] Properly handle reference lattices that are not A-B symmetric
- [ ] Detect and warn for not-actually-implemented features instead of randomly crashing (or worse, not crashing)
- [ ] Provide OpenMP parallelization
- [ ] Check and update literature references (eqn numbers etc.)

Some of this I've already done in prototypes over the years, the final version should properly integrate.

**Author(s)**

Sebastian Hütter, OvGU

**Licensing**

**Backward Compatibility**

There will be (hopefully) no functional changes *in correct usage*. Some results might change due to numerics, but this is expected to be minimal. I am willing to take minor performance regressions if it leads to better code, but at the same time compilers got smarter, so we might end up unchanged. Undefined / buggy behavior might change, ideally towards failing properly.

Apologies in advance to @stanmoore1 for the KOKKOS port. In an ideal world, things get easier since I need to shuffle some things for OMP anyway...

**Implementation Notes**

... will update as things progress ...

**Post Submission Checklist**


martok:
Hm, I can't reproduce the failure in test 296 - AtomicPairStyle:meam_ms locally. Could be a compiler issue? Mine:
```
OS: Linux "Ubuntu 22.04.5 LTS" 5.15.0-134-generic x86_64
Compiler: GNU C++ 11.4.0 with OpenMP 4.5
```

Other than that we're getting somewhere, only a few places with explicit lattice handling left which I intend to replace with parameterized routines. So far no performance regressions - it actually got faster by about 5%. Compilers today are apparently better than manual optimisation 22 years ago.

sjplimp:
One suggestion for a PR like this - (your call, you can ignore if you don't think it is useful):
(1) keep the old MEAM files for now
      e.g. rename the old files to be pair_style meam/old and the new ones to be pair_style meam
      this will makes it easy to build one exe with both versions, to facilitate testing and perf comparisons
(2) you could even release the PR that way, with both versions, to get feedback from users on bugs, changed results or perf changes
      then after ~6 months, use a new PR to whack the old version

martok:
> One suggestion for a PR like this - (your call, you can ignore if you don't think it is useful): (1) keep the old MEAM files for now e.g. rename the old files to be pair_style meam/old and the new ones to be pair_style meam this will makes it easy to build one exe with both versions, to facilitate testing and perf comparisons

Oh, right, like we did with meam/c during the fortran transition. That is a very good hint, I will do that. This should also let me get this work done without breaking KOKKOS at all.

From a git history perspective, is it better to have the "old" version be the copy or the new one?

sjplimp:
>> From a git history perspective, is it better to have the "old" version be the copy or the new one?
@akohlmey can answer that Q better than I

I think for the initial public release (with both versions) it would be better to have the new one be "meam", so
that it is used by default.  Otherwise users will not think to look for a new version.

stanmoore1:
@martok any update on this PR?

martok:
> @martok any update on this PR?

Not as much as I'd like since the last push (I have a more recent rebase and some local changes pending but didn't want to cause too much noise here). The architectural changes all work, look sane (IMO) and in many places actually improve performance, so that's nice. However I ran into a logical issue in the implementation of AxBy (x!=y) reference structures that still needs some drawing and 3d modelling and counting of coordination shells to sort out. In principle not more than maybe two weeks of actual work, but so far I've gotten in around a day a month, summer has been way more stressful than I had expected.

What I can't really give is a proper time table for the remaining work...

akohlmey:
@martok This pull request has been sitting idle for a couple of months now and there hasn't been any feedback or collaboration beyond the initial exchanges. Thus, I suggest you close it for the time being and resubmit when it is close(r) to completion or when you need specific feedback. This way it doesn't clog up our list of pending pull requests without being in need of reviews or feedback. Thanks.

martok:
@akohlmey Yeah, you're probably right. We spoke about it elsewhere, surprise teaching obligations eat a lot of time.

I'll close it for now and keep working on it on my fork. Is there an easy way to run the tests that are run as merge checks individually? Some manual trigger or something? I don't use Github actions all that much.

akohlmey:
> Is there an easy way to run the tests that are run as merge checks individually?

In `.github/workflows/` you can change in the YML files the string `lammps/lammps' to 'martok/lammps' and 'develop' to 'meam-cpp-refactor' and the checks on GitHub should be run whenever you push. I suggest you edit only the following files:

`check-cpp23.yml`, `check-vla.yml`, `compile-msvc.yml`, `style-check.yml`, `unittest-arm64.yml`, `unittest-linux.yml`, and `unittest-macos.yml`.

Of course, these changes will need to be reverted when you resubmit the branch.

PR #4473, opened 2025-02-14

akohlmey:
**Summary**

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

Some additional arguments for styles that were previously ignored will now trigger an error.

**Implementation Notes**

- new compute vacf/chunk command modeled after compute msd/chunk
- improved and more modern and careful error checking
- make SafeFilePtr class implementation compatible with C++11
- make certain that Contacs::varflag is always initialized
- correct cutneighmin in the Neighbor class in case there are neighbor list requests with a shorter cutoff (e.g. from hybrid pair styles). This recovers using the "trim" neighbor list builder for these cases.

**Post Submission Checklist**


src/BOCS/compute_pressure_bocs.cpp
```diff
@@ -139,7 +139,8 @@ void ComputePressureBocs::init()
   if (keflag) {
     temperature = modify->get_compute_by_id(id_temp);
     if (!temperature)
-      error->all(FLERR,"Could not find compute pressure/bocs temperature compute {}", id_temp);
+      error->all(FLERR, Error::NOLASTLINE,
```
jtclemm, line 142: I hadn't noticed the addition of `Error::NOLASTLINE` before, does its inclusion do anything different than the regular `error->all` calls w/o the argument?

jtclemm (approved):

src/BOCS/compute_pressure_bocs.cpp
akohlmey: The purpose is to suppress a "Last command: run ###" message that is appended by default (for backward compatibility and corresponds to `Error::NOPOINTER`, i.e print  the last command, but do not point to any keyword in it) and can be confusing for all errors happening in the "middle" or a run, e.g. during setup() or init().
I've tried to explain the current status here:
https://docs.lammps.org/Developer_notes.html#errors-warnings-and-informational-messages

jtclemm: Ah, that makes sense. I had picked up on the new use of arg indices before but not that.

PR #4480, opened 2025-02-24
Fix bad value for factor_sqrt with dpd/tstat pair_style

jdevemy:
**Summary**

The temperature diverges when using the `dpd/tstat` pair style because the `factor_sqrt` variable has not the right value.

**Related Issue(s)**

**Author(s)**

Julien Devémy
Alain Dequidt

**Licensing**

**Backward Compatibility**

**Implementation Notes**

The fix consist in recomputing the sqrt in the neighbors loop, so it's not optimal but it fixes the problem.

The issue may also be present in other `dpd` pair style, like `dpd` or `dpd/gpu`...

**Post Submission Checklist**

Here attached a data and input file to test.

[dpd_temp_bug.tar.gz](https://github.com/user-attachments/files/18938359/dpd_temp_bug.tar.gz)

Without the fix (blue), the temperature diverges, but with the fix (orange), it converges to the target (100K):

image

### DIFF
--- a/src/DPD-BASIC/pair_dpd_tstat.cpp
+++ b/src/DPD-BASIC/pair_dpd_tstat.cpp
@@ -91,7 +91,7 @@ void PairDPDTstat::compute(int eflag, int vflag)
     for (jj = 0; jj < jnum; jj++) {
       j = jlist[jj];
       factor_dpd = special_lj[sbmask(j)];
-      factor_sqrt = special_sqrt[sbmask(j)];
+      factor_sqrt = sqrt(factor_dpd);
       j &= NEIGHMASK;
 
       delx = xtmp - x[j][0];


akohlmey:
@jdevemy Thanks for your submission. This issue happens because in the current code the values of the special_sqrt[] array are pre-computed during Pair::init_syle(). However, in your specific input, you have different special_bonds values for different pair styles. Those get overridden in PairHybrid::compute(). Thus, had you used in your input:
```
special_bonds lj/coul 0.9999999999999999 1. 1.
pair_modify pair lj/cut special lj 0. 1. 1.
```
the issue would not have shown.

The better solution is re-initialize the special_sqrt[] array inside the Pair::compute() function of each pair style as it is done in commit d79be293f53ed4ea6bc09b6b01659d8d4c1d2789

PR #4481, opened 2025-02-26

### LINKED ISSUE #4487: [BUG] No neighbors when full neighborlist requested by a fix is shared by a pair_style
**Summary**

This bug occurs when all the following are true:

- A LAMMPS fix requests a full neighborlist with a modified cutoff, e.g using: `neighbor->add_request(this, NeighConst::REQ_FULL)->set_cutoff(rcutoff);`
- One or more pair_styles are defined which request half neighborlists with cutoff radii smaller than the cutoff used in the fix neighborlist.
- No pair_styles are defined which request a full neighborlist, such that the half neighborlist pair_styles get their neighborlists by pruning the full neighborlist requested by the fix, i.e:
```
  (1) pair lj/cut, perpetual, half/full trim from (2)
      attributes: half, newton on
      pair build: halffull/trim/newton
      stencil: none
      bin: none
  (2) fix debug, perpetual
      attributes: full, newton on, cut 8
      pair build: full/bin/atomonly
      stencil: full/bin/3d
      bin: standard
```

Under these conditions, i atoms in the half neighborlists accessed by the pair_styles have no j atoms, that is num_neigh(i) = 0 for all i, and therefore all computed forces and energies are 0. 

This bug was first discovered on LAMMPS release version 4 Feb 2025 and has been reproduced on the current development branch - LAMMPS (4 Feb 2025 - Development - patch_4Feb2025-76-gf0bc9ddcc9).

**Expected Behavior**

When using a pair_style that uses a half-full neighborlist constructed by pruning a full neighborlist requested by a fix, the forces and energies should be the same regardless of the cutoff specified when the fix requests the neighborlist.

**Actual Behavior**

Currently, if a half neighborlist used in a pair_style is pruned from a full neighborlist requested by a fix (where the fix neighborlist cutoff is greater than the pair_style neighborlist cutoff), then all forces and energies computed are exactly 0. This is because the i atoms in the neighbor_list accessed by the pair_style erroneously have no neighbors. 

This bug does not occur if the fix neighborlist cutoff is less than the pair_style neighborlist cutoff.

**Steps to Reproduce**

Simple example code to reproduce this bug is provided here:

[debug_files.zip](https://github.com/user-attachments/files/19073114/debug_files.zip)

*Steps*:
1. Build LAMMPS with the plugin package enabled
2. Build the `fix debug` plugin provided 
```
mkdir build
cd build
cmake .. -D LAMMPS_SOURCE_DIR=/path/to/lammps/src \
cmake --build . -j 1
```
3. Add the correct path to the plugin `.so` file at the top of `debug_input_file.in`.
4. Change the cutoff radius passed to `fix debug` in the input file. The bug will occur if it is greater than 5.0 (the cutoff specified in the defined lj/cut pair_style).
5. Run `lmp -in debug_input_file.in`. If the cutoff passed to `fix debug` was less than 5.0, the following thermo output should be seen:
```
Setting up pre force
Per MPI rank memory allocation (min/avg/max) = 8.774 | 8.774 | 8.774 Mbytes
   Step         PotEng         KinEng    
         0  -1294.9007      0            
Loop time of 1.328e-06 on 1 procs for 0 steps with 4000 atoms
```
and there should be non-zero forces present in the dump file `debug.lammpstrj`. 

If the cutoff passed to `fix debug` was greater than 5.0, the following thermo output should be seen:
```
Setting up pre force
Per MPI rank memory allocation (min/avg/max) = 7.248 | 7.248 | 7.248 Mbytes
   Step         PotEng         KinEng    
         0   0              0            
Loop time of 1.275e-06 on 1 procs for 0 steps with 4000 atoms
```
and all forces in the dump file `debug.lammpstrj` should be 0. 

This bug was found as it currently makes it impossible to use the [ML-MIX LAMMPS plugin](https://github.com/kermodegroup/ML-MIX) to force-mix two pair_styles that use half neighborlists. This is because the region building fix used in this package requests full neighborlists.

### LINKED ISSUE #4489: [BUG] Segfault with fix pimd/langevin
**Summary**

fix pimd/langevin segfaults when running with multiple processors and method `pimd`.

`develop`

**Expected Behavior**

Error message if multiproc isn't supported or implementation that allows this.

**Actual Behavior**

Segfault, due to accessing `bufsortedall` with `nullptr` in `FixPIMDLangevin::spring_force()`

**Steps to Reproduce**

```
# input script
...
fix  foo all pimd/langevin method pimd ...
```

```
srun -n 112 lmp -partition 4x28 -in in.script
```

### LINKED ISSUE #4491: [BUG] _Wrong coordination number (CN) with compute rdf
**Summary**

The coordination number calculation with `compute rdf` is wrong.
``

LAMMPS (29 Aug 2024)

Rocky Linux release 9.2 (Blue Onyx)

System: Lenovo NeXtScale nx360
CPU: Intel E5-2683v4 2.1 GHz

**Expected Behavior**

Correct CN calculation. For this system, the CN should be 4 beyond 3.2 Å.

**Actual Behavior**

Wrong CN calculation. The CN is 1 beyond 3.2 Å, and the maximum CN (at the cutoff of 12.96 Å) value is 1.

**Steps to Reproduce**

Remove ".txt" from file names.

Run the input file test_rdf.lmp with the data file test_rdf.data and the parameter file 512_acetamide_128_LiTFSI.settings. RDF data is output to test.rdf.

Input files:
[512_acetamide_128_LiTFSI.settings.txt](https://github.com/user-attachments/files/19111191/512_acetamide_128_LiTFSI.settings.txt)
[test_rdf.data.txt](https://github.com/user-attachments/files/19111189/test_rdf.data.txt)
[test_rdf.lmp.txt](https://github.com/user-attachments/files/19111190/test_rdf.lmp.txt)

RDF output:
[test.rdf.txt](https://github.com/user-attachments/files/19111215/test.rdf.txt)

### LINKED ISSUE #4499: [BUG] Restart file is invalid
**Summary**

We found what we believe to be a LAMMPS bug: a restart file being written by LAMMPS, and that cannot be read back by the same version of LAMMPS, on the same computer.

A full reproducer for the bug can be found there: https://filesender.renater.fr/?s=download&token=b8a4f79b-b276-4868-8f57-6dcce6d47b5d

We have reproduced this bug on:
- LAMMPS 4 Feb 2025 static Linux binaries for x86_64
- Another Linux x86_64 install of the 29 Aug 2024 version

**Expected Behavior**

The restart files written by the following command in `mof.input`:

```
restart                 100 restart.*
```

should later be read fine, in `mof_restart.input`, by this command:

```
read_restart            restart.100
```

**Actual Behavior**

The output from running `mof_restart.input` is the following error:

```
LAMMPS (4 Feb 2025)
OMP_NUM_THREADS environment is not set. Defaulting to 1 thread. (src/comm.cpp:99)
Reading restart file ...
  restart file = 4 Feb 2025, LAMMPS = 4 Feb 2025
  restoring atom style full from restart
ERROR: Invalid flag in header section of restart file (src/read_restart.cpp:855)
Last input line: read_restart restart.100
```

**Steps to Reproduce**

- `lmp -in ./mof.input`
- `lmp -in ./mof_restart.input`

The input files (and data file) are available here: [restart_bug.tar.gz](https://github.com/user-attachments/files/19249160/restart_bug.tar.gz)

akohlmey:
**Summary**

**Related Issue(s)**

Fixes #4487
Fixes #4491 
Fixes #4489 
Fixes #4499 
Fix halt universe flag was inspired by this discussion https://matsci.org/t/quitting-all-partitions/61686/
SPIN package fixes and example update inspired by https://matsci.org/t/spin-example-input-script-crashing-at-time-step-0/61762

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- fix issue when trimming neighbor lists where the trimmed list request does not have a custom cutoff
- fix normalization issue with compute rdf when atom types were swapped on input
- fix bug when restarting from a general triclinic geometry
- do not allow using fix pimd/langevin in pimd mode with multiple processors per bead
- add test to ReaxFF parameter file parser to error out with a meaningful error when the "lgvdw" flag is used with an incompatible force field file
- add "universe" option to fix halt that will synchronize a stop across multiple partitions
- add note about limited dynamic group support for fix reaxff/species doc page
- set stores_ids flag to 1 for several fixes that store atom ID related or dependent data
- make certain nstyle member variable is initialized before use in RegionPlane class
- bugfix from @ndtrung81 to avoid problems when using read_dump in cases where not all MPI ranks receive atoms
- fix some minor memory leaks in the SPIN package
- correctly tally "Sync" time when enabled and thus don't double count it in "Other"
- enable recently release sphinx 8.2.x
- document (some) internal styles
- correct and clarify the documentation for bond styles harmonic/shift and harmonic/shift/cut
- documentation fixes and cleanups
- modernize some error messages
- update SPIN/read_restart examples and logfiles

**Post Submission Checklist**


akohlmey:
@stanmoore1 @jtclemm Please check out the changes to neighbor list trimming based on the bug report in #4487. In the test case the custom cutoff is not set for the neighbor list that is trimmed and then it is trimmed to cutoff 0.0.

I could not come up with an idea for how to do the same for "size" style neighbor lists, so I effectively disabled trimming when there is no custom cutoff.

stanmoore1:
Neighlist changes likely need to be ported to KOKKOS

akohlmey:
> Neighlist changes likely need to be ported to KOKKOS

Yes, they do. I had a look, but it is beyond my level of understanding of KOKKOS to add access to data like the pairwise cutoff or atom type, so your assistance is needed here.

jtclemm:
> @stanmoore1 @jtclemm Please check out the changes to neighbor list trimming based on the bug report in #4487. In the test case the custom cutoff is not set for the neighbor list that is trimmed and then it is trimmed to cutoff 0.0.

Those changes look correct to me. I can't see any issues that could emerge.

> I could not come up with an idea for how to do the same for "size" style neighbor lists, so I effectively disabled trimming when there is no custom cutoff.

Can the same strategy, defining `cut_trim` as `cutneighsq[itype][jtype]` if no custom cutoff is defined, not work?

akohlmey:
> Can the same strategy, defining `cut_trim` as `cutneighsq[itype][jtype]` if no custom cutoff is defined, not work?

On second thought, I don't think it is necessary. In the "size" style neighborlist build, the cutoff to compare to is computed from `radius[i] + radius[j] + skin`, so there should be nothing to trim unless a shorter(!) custom cutoff is set. It would make no sense to look for a per-type cutoff.

sjplimp (approved):

jtclemm (approved):

PR #4485, opened 2025-03-01
Address several issues with pair style dispersion/d3

### LINKED ISSUE #4484: ERROR:Pair_style dispersion/d3 needs 4 arguments (…/pair_dispersion_d3.cpp:133)
**Summary**

When I tried to use pair_style dispersion/d3 to combine my machine learning potential for molecular dynamics simulations, I wrote the following line for the pair style:
pair_style hybrid/overlay nep nep nep nep nep dispersion/d3 zero pbe 12.0 6.0
, where “nep” is an external package for a machine learning potential that has been compiled into LAMMPS via ([GitHub - brucefan1983/NEP_CPU: CPU version of NEP](https://github.com/brucefan1983/NEP_CPU)) . However, I encountered an error during runtime:
ERROR: Pair_style dispersion/d3 needs 4 arguments (…/pair_dispersion_d3.cpp:133) .
The error indicates that the style requires 4 arguments, but I did provide four parameters (dispersion/d3zero pbe 12.0 6.0dispersion/d3) after . I would like to request assistance in resolving this issue.

[akohlmey](https://matsci.org/u/akohlmey) think This is due to an unfortunate choice of the authors of the dispersion/d3 pair style.When using any hybrid pair styles, the arguments are “partitioned” into entries for the individual pair styles by scanning the command for names of valid pair styles. As it happens “zero” is a vaild pair style name and thus pair style hybrid will assign no arguments to the dispersion/d3 sub-style and assume that there also is a zero sub-style with three arguments.

lammps-4Feb2025

**Expected Behavior**

I'd like to solve the problem of zero recognising errors when using pair_style hybrid/overlay

akohlmey:
**Summary**

This pull request fixes several issues with pair style dispersion/d3

**Related Issue(s)**

Fixes #4484 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

Yes.

**Implementation Notes**

The keyword "zero" cannot be used with hybrid pair styles, because there is a pair style "zero" and the list of arguments are split between sub-styles based on known pair styles. Thus the "zero" keyword is interpreted as a new sub-style.
The keyword is renamed in this pull request to "original" ("zero" remains recognized for backward compatibility, but is undocumented).

Furthermore, the detection of elements is made more thorough and errors are thrown, if needed.

Also, the pair style must require the use of "metal" units and that check is added.

**Post Submission Checklist**

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -72,6 +72,8 @@ src/EXTRA-COMMAND/ndx_group.*       @akohlmey
 src/EXTRA-COMPUTE/compute_stress_mop*.* @RomainVermorel
 src/EXTRA-COMPUTE/compute_born_matrix.* @Bibobu @athomps
 src/EXTRA-FIX/fix_deform_pressure.* @jtclemm
+src/EXTRA-PAIR/pair_dispersion_d3.* @soniasolomoni @arthurfl
+src/EXTRA-PAIR/d3_parameters.h      @soniasolomoni @arthurfl
 src/MISC/*_tracker.*                @jtclemm
 src/MC/fix_gcmc.*                   @athomps
 src/MC/fix_sgcmc.*                  @athomps
--- a/doc/src/pair_dispersion_d3.rst
+++ b/doc/src/pair_dispersion_d3.rst
@@ -63,7 +63,8 @@ factor, and :math:`f_n^{damp}` are damping functions.
 
 Available damping functions are the original "zero-damping" (*original*)
 :ref:`(Grimme1) <Grimme1>`, Becke-Johnson damping (*bj*) :ref:`(Grimme2)
-<Grimme2>`, and their revised forms (*bjm*) :ref:`(Sherrill) <Sherrill>`.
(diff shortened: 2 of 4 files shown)


akohlmey:
@soniasalomoni @arthurfl Please have a look at the changes proposed in this pull request to address issue #4484 and some more problems noticed in the process.

Please also note that the "zerom" damping function is not documented and thus some information about it needs to be added.

arthurfl:
Thanks for the feedback and the quick fix, these changes look good to me. Concerning the "zerom" damping function (that also concerns the "bjm" variant), it is already referred to in the docs with the citation from Sherrill et al. These are simply new sets of parameters. Should we add more detail?

akohlmey:
> These are simply new sets of parameters. Should we add more detail?

My mistake, I misread the statement before the last quote and didn't notice it was plural when I was trying to map the keywords to the references. I've updated this accordingly and we should be fine now.

sjplimp (approved):

PR #4502, opened 2025-03-16

### LINKED ISSUE #4503: [BUG] utils::print() fails silently without ERROR or WARNING
**Summary**

while writing code that uses utils::print(), a problem with the fmt is silently ignored without an ERROR or WARNING. utils::fmtargs_print() has a catch that _" ; // do nothing "_.

https://github.com/lammps/lammps/blob/2240430d5eae6a330dfff9a7cf1cefbfd381b832/src/utils.cpp#L273-L280

```
mitch@macmini build-charmm % lmp_psf -h
zsh: command not found: lmp_psf
mitch@macmini build-charmm % ./lmp_psf -h
Large-scale Atomic/Molecular Massively Parallel Simulator - 4 Feb 2025 - Development
Git info (alphataubio-psf / patch_4Feb2025-566-ge30221d6e4-modified)
OS: Darwin 23.4.0 arm64
Compiler: Clang C++ Apple LLVM 15.0.0 (clang-1500.3.9.4) with OpenMP not enabled
C++ standard: C++17
Embedded fmt library version: 10.2.0
MPI v3.1: Open MPI v5.0.6, package: Open MPI brew@Sonoma-arm64.local Distribution, ident: 5.0.6, repo rev: v5.0.6, Nov 15, 2024

Installed packages:
CHARMM COLVARS EXTRA-COMMAND EXTRA-DUMP KSPACE MOLECULE RIGID 
```

**Expected Behavior**

should give an ERROR or at least a WARNING

**Steps to Reproduce**

```
    utils::logmesg(lmp, "*** i {}\n", i);
    utils::print(fp, "{} {}\n", i)
  }
```

**Suggested Fix**

https://github.com/lammps/lammps/pull/4245/commits/f3c53867283805423e885249450c0ba7fe645369

### LINKED ISSUE #4504: Python lib install unable to circumvent "externally-managed-environment" error
**Summary**

Unable to use "make install-python", or rather the "install.py" script to install lammps into an "externally-managed-environment", even only on the user level, because there is not way to pass the "--break-system-packages" argument to "pip" through it.

lammps-2025.2.4 ; Ubuntu 24.04.2 LTS ; hardware irrelevant

**Details**

Newer versions of pip don't let a user install libs into an "externally-managed-environment", which is basically the system python, neither for the whole system or just the user. Instead they print this message:
```
error: externally-managed-environment
× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.12/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
```
So currently one can only install the LAMMPS python library in a virtual environment, because there is no way to pass the "--break-system-packages" parameter to the "install.py" script.

I circumvented the problem by adding that parameter myself in lines 148 and 157 of the "install.py" script, where "pip" is called, but there should be some nicer way to get that done by an average user. Like passing it as a parameter to the script. And "make install_python" should be able to get around that problem also. Maybe some message when it initially fails on how to circumvent it if one wants to (like pip itself does).

P.S. I might be wrong, but line 153 in "install.py" seems like it might be a bug... Because Python's "String.find" returns "-1" when the argument is not found in the string, and for the purposes of the "if" statement -1 equals TRUE, and maybe that wasn't the intent here?

### LINKED ISSUE #4505: [BUG] lammps can't find shared libraries without LD_LIBRARY_PATH
**Summary**

LAMMPS (built locally) doesn't "see" the "liblammps.so" shared library in the "/usr/local/lib" folder

LAMMPS (4 Feb 2025 - Development - patch_4May2022-12521-gde246dd05a-modified) 
on Ubuntu 24.02 LTS

**Expected Behavior**

It's expected for programs to "see" shared libraries in the "/usr/local/lib" folder, or any folders in the conf files in "/etc/ld.so.conf.d" (libc.conf specifically contains /usr/local/lib)

**Actual Behavior**

One has to put /usr/local/lib explicitly in the LD_LIBRARY_PATH env var and export it for LAMMPS to work...

**Steps to Reproduce**

```
export LD_LIBRARY_PATH=""
lammps
```
gives
```
lammps: error while loading shared libraries: liblammps_mine.so: cannot open shared object file: No such file or directory
```
and
```
export LD_LIBRARY_PATH=/usr/local/lib
lammps
```
gives
```
LAMMPS (4 Feb 2025 - Development - patch_4May2022-12521-gde246dd05a-modified)

```

akohlmey:
**Summary**

**Related Issue(s)**

fixes #4503 
fixes #4504 
closes #4505 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- address several minor issues that were flagged by static code analysis
- revert macro for angstrom character back to working version
- update WHAM source download to fetch version 2.1.0 which includes our customizations
- do not silence format errors in utils::print() (will have to flag implicitly non-throwing functions with noexcept(false) instead)
- add note that when using CMake and "make install" to a system folder, you also need to run `ldconfig`.
- document `-D DOWNLOAD_POTENTIALS=off` for installing LAMMPS in offline mode or to avoid big downloads
- document how to handle "externally managed environment" errors when installing the LAMMPS Python module
- add always visible "Go to top" anchor at bottom of the navigation sidebar that should jump to the top (this is using plain html and sometimes requires clicking twice, it might be better implemented in JavaScript)

**Post Submission Checklist**


doc/utils/sphinx-config/_themes/lammps_theme/layout.html
```diff
@@ -172,6 +173,11 @@
         </div>
         {%- endblock %}
       </div>
+      <div style="padding: 4px;">
```
rbberger: This will get hidden by the fixed Versions bar.

rbberger: You can do this and get rid of the image.
```html
<div style="position: fixed; bottom: 0px; width: 300px; height: 64px; padding: 4px; background-color: #272525; text-align: center">
<a href="#sidebar-top-target" class="fa fa-caret-up" style="color: white"> Go to top</a>
</div>
```
Note, this still breaks rendering on mobile devices. For that we would have to put this into proper CSS rules that are conditioned on the device size.

akohlmey: Ok. I have replaced it. Any idea why I need to click twice (with either version) on this to have the desired effect?

jtclemm (approved):

PR #4507, opened 2025-03-19
Add fix qeq/rel/reaxff

### LINKED ISSUE #4458: [Feature Request] QEq with alternate efield implementation
**Summary**

Using `fix efield` with `fix qeq/reaxff` does not allow for periodic boundaries in the direction of an external electric field and can also lead to unphysical long-range charge transfer. Using electric potential differences to incorporate an electric field, as in `fix qtpie/reaxff`, could lead to more physical behaviour with `fix qeq/reaxff` and an electric field and also allow for periodic boundaries in the direction of an external electric field.

**Detailed Description**

This could be easily implemented by inheriting from `fix qtpie/reaxff`. 

@stanmoore1 is this something that you think it would be worth including in the REAXFF package as a new fix? It should be said that this method of incorporating an electric field is based on physical intuition, rather than having a theoretical foundation, but it does lead to more physical behaviour than the current method of incorporating an electric field with `fix qeq/reaxff` and it does also permit periodic boundaries in all directions.

I'd be happy to implement this fix myself if you think it is worth including. I have a couple of questions:
* Should the fix be restricted so that it requires `fix efield` to be in use, since the behaviour is the same as `fix qeq/reaxff` in the absence of an electric field and `fix qeq/reaxff` will be faster?
* What would be an appropriate name for this fix?

navlalli:
**Summary**

`fix qeq/rel/reaxff` accounts for an external electric field using electric potential differences, as in `fix qtpie/reaxff`. The electric field implementation with `fix qeq/rel/reaxff` is advantageous over that of `fix qeq/reaxff` in that it permits an external electric field in the direction of periodic boundaries and in that it doesn't worsen long-range charge transfer.

**Related Issue(s)**

Closes #4458 

**Author(s)**

Navraj S Lalli (Imperial College London)

**Licensing**

**Backward Compatibility**

**Implementation Notes**

This fix is implemented by adding the files fix_qeq_rel_reaxff.cpp and fix_qeq_rel_reaxff.h to the REAXFF package.

**Post Submission Checklist**


stanmoore1 (approved):
This looks good to me. @navlalli thank you for writing the code cleanly.

akohlmey (changes requested):
@navlalli I think we need a better name for this fix. Typically, when we have variants of an existing fix we add `/<something>` or in this case rather have `qeq/<something>/reaxff. Adapting the current name would lead to `qeq/r/reaxff` where it is not clear what the '/r/' stands for. Some brainstorming with @stanmoore1 led to: qeq/potential/reaxff, qeq/epot/reaxff, qeq/potdiff/reaxff, qeq/dedr/reaxff
My preference would be either qeq/epot/reaxff or qeq/potdiff/reaxff.
@sjplimp any opinion or alternate suggestion

navlalli:
The `r` stands for relative because this fix uses electric potential differences and therefore only depends on the relative position of atoms, i.e., the position of atoms relative to nearby atoms (the effective electronegativity of each atom is independent of the reference point chosen as the position of 0 electric potential). This is in contrast with `fix qeq/reaxff` and the use of `fix efield` where the effective electronegativity depends on the absolute position of each atom, i.e., the position from the origin or whatever point is defined as the position of 0 electric potential. 

How about `fix qeq/rel/reaxff` or `fix qeq/erel/reaxff`?

If it helps, I'm planning on publishing some work with this method so could link that to the docs to make it clear where the name comes from (once it's out).

src/REAXFF/fix_qeqr_reaxff.h
```diff
@@ -0,0 +1,38 @@
...
+#else
+
+#ifndef LMP_FIX_QEQR_REAXFF_H
+#define LMP_FIX_QEQR_REAXFF_H
+
+#include "fix_qtpie_reaxff.h"
+
+namespace LAMMPS_NS {
+
+class FixQEqrReaxFF : public FixQtpieReaxFF {
```
akohlmey: @navlalli the documentation says that fix qeqr/reaxff would be exactly like fix qeq/reaxff when used without an electric field. How does this add up when the fix is derived from fix qtpie/reaxff instead?

doc/src/fix_qeq_rel_reaxff.rst
```diff
@@ -0,0 +1,195 @@
...
+
+where *itype* is the atom type from 1 to Ntypes. Note that eta is
+defined here as twice the eta value in the ReaxFF file.
+
+The overlap integrals :math:`S_{ij}` are computed by using normalized 1s
+Gaussian type orbitals. The Gaussian orbital exponents, :math:`\alpha`,
+that are needed to compute the overlap integrals are taken from the file
+given by *gfile*.  This file must contain one line for each atom type
+and provide the Gaussian orbital exponent for each atom type in units of
+inverse square Bohr radius.  Each line should be formatted as follows:
```
srtee, line 102: Strongly suggest using native distance units (1/Angstrom squared). Users should not have to juggle multiple different (and easily confused) distance unit systems in one input script -- that's the developers' job. Provide the conversion factor 1 inverse square Bohr = 3.57107 inverse square angstrom in the documentation if needed.

src/REAXFF/fix_qeq_rel_reaxff.cpp
```diff
@@ -0,0 +1,112 @@
...
+          if (efield->varflag != FixEfield::ATOM) {
+            phib = -factor * (x[j][0] * efield->ex + x[j][1] * efield->ey + x[j][2] * efield->ez);
+          } else {    // atom-style potential from FixEfield
+            phib = efield->efield[j][3];
+          }
+          sum_n += (chia + scale * (phia - phib)) * overlap;
+          sum_d += overlap;
+        }
+      }
+      chi_eff[i] = sum_n / sum_d;
```
srtee:
```suggestion
          sum_n += scale * phib * overlap;
          sum_d += overlap;
        }
      }
      chi_eff[i] = chia + scale * phia - sum_n / sum_d;
```

I assume this is the same -- it saves a few FLOPs and is more readable (especially in lining up the source code with the docs description)

src/REAXFF/fix_qeq_rel_reaxff.cpp
```diff
@@ -0,0 +1,112 @@
...
+        if (dist < dist_cutoff) {
+          expb = gauss_exp[type[j]];
+
+          // overlap integral of two normalised 1s Gaussian type orbitals
+          p = expa + expb;
+          m = expa * expb / p;
+          overlap = pow((4.0 * m / p), 0.75) * exp(-m * dist * dist);
+
+          if (efield->varflag != FixEfield::ATOM) {
+            phib = -factor * (x[j][0] * efield->ex + x[j][1] * efield->ey + x[j][2] * efield->ez);
```
srtee, line 96: I am assuming that `j` is from a neighlist and therefore `x[j]` will be correctly unwrapped?

src/REAXFF/fix_qeq_rel_reaxff.cpp
```diff
@@ -0,0 +1,112 @@
...
+        phia = -factor * (x[i][0] * efield->ex + x[i][1] * efield->ey + x[i][2] * efield->ez);
+      } else {    // atom-style potential from FixEfield
+        phia = efield->efield[i][3];
+      }
+
+      sum_n = 0.0;
+      sum_d = 0.0;
+
+      for (j = 0; j < nt; j++) {
+        dist = distance(x[i], x[j]) * ANGSTROM_TO_BOHRRADIUS;    // in atomic units
```
srtee: - as above -- juggling distance units introduces more fragility into the code
- if the square distance is just `dx*dx + dy*dy + dz*dz` then it shouldn't be shunted into a different function which will make the code harder to maintain
- furthermore, if the square distance is being used throughout and not the distance itself, then using square distance consistently saves quite a few square roots (cutoff checking can be done using rsq < cutsq as is done elsewhere throughout LAMMPS)

src/REAXFF/fix_qeq_rel_reaxff.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+        if (dist < dist_cutoff) {
+          expb = gauss_exp[type[j]];
+
+          // overlap integral of two normalised 1s Gaussian type orbitals
+          p = expa + expb;
+          m = expa * expb / p;
+          overlap = pow((4.0 * m / p), 0.75) * exp(-m * dist * dist);
```
srtee:
```suggestion
          // double root2, root4;
          root2 = 2.0*sqrt(m/p);
          root4 = sqrt(root2);
          overlap = root2 * root4 * exp(-m * dist * dist);
```

More importantly, it looks like `overlap` is some distance-independent prefactor times a Gaussian in distance squared -- the prefactor should absolutely be precomputed and stored in some `prefactor[itype][jtype]` table.

doc/src/fix_qeq_rel_reaxff.rst
```diff
@@ -0,0 +1,195 @@
+.. index:: fix qaq/rel/reaxff
```
srtee:
```suggestion
.. index:: fix qeq/rel/reaxff
```

doc/src/fix_qeq_rel_reaxff.rst
```diff
@@ -0,0 +1,195 @@
...
+efield <fix_efield>` since :doc:`fix qeq/reaxff <fix_qeq_reaxff>` should
+be used without an external electric field.  Charges are computed with
+the QEqR method by minimizing the electrostatic energy of the system in
+the same way as the QEq method but where the absolute electronegativity,
+:math:`\chi_i`, of each atom in the QEq method is replaced with an
+effective electronegativity given by
+
+.. math::
+   \chi_{\mathrm{r}i} = \chi_i + \frac{\sum_{j=1}^{N} \beta(\phi_i - \phi_j) S_{ij}}
+                                      {\sum_{m=1}^{N}S_{im}},
```
srtee, line 55:
```suggestion
   \chi_{\mathrm{r}i} = \chi_i + \frac{\sum_{j=1}^{N} \beta(\phi_i - \phi_j) S_{ij}}
                                      {\sum_{j=1}^{N}S_{ij}},
```

I have the same question now about the original QTPIE method and documentation: what is the difference between the index `j` in the numerator and index `m` in the denominator? Are they the same, referring to all nearby overlapping atoms (in which case they should both run over the same index), or are `j` and `m` two different sets of atoms (which doesn't seem to be reflected in the code)?

doc/src/fix_qeq_rel_reaxff.rst
```diff
@@ -0,0 +1,195 @@
...
+
+.. parsed-literal::
+
+   itype chi eta gamma
+
+where *itype* is the atom type from 1 to Ntypes. Note that eta is
+defined here as twice the eta value in the ReaxFF file.
+
+The overlap integrals :math:`S_{ij}` are computed by using normalized 1s
+Gaussian type orbitals. The Gaussian orbital exponents, :math:`\alpha`,
```
srtee, line 98: Best to state the exact function used -- in particular whether the charge density follows exp(- alpha r^2) or exp(-0.5 alpha r^2), both of which see common (and incompatible) usage.

srtee (commented):
In addition to the code comments, I am not sure what the physical motivation is for this method compared to QTPIE.

Consider a charge distribution being perturbed by an electronegativity difference of 0.1 eV between two points -- the goal of QTPIE is to damp the perturbation when those two points are far separated, and QTPIE makes no distinction between whether that electronegativity difference comes from chi (between atomic nuclei+core electrons) or phi (from an external field).

Is it superior to have a new method that does? Why should charge transfer be range-independent under electronegativity difference, but range-dependent under an external field?

src/REAXFF/fix_qeqr_reaxff.h
navlalli: https://github.com/lammps/lammps/blob/fc9d7bc181b45c88fd87bfd4fd320cf474c6d912/src/REAXFF/fix_qeq_rel_reaxff.cpp#L110

This line sets the effective electronegativity of each atom to the absolute electronegativity of each atom (when no efield is applied), which aligns with `fix qeq/reaxff`.

doc/src/fix_qeq_rel_reaxff.rst
navlalli: Each sum is independent, so two different indexes are used. I think it is clearer to use two separate indexes rather than `j` in both. But you are correct that they could be interchanged

navlalli:
> In addition to the code comments, I am not sure what the physical motivation is for this method compared to QTPIE.
> 
> Consider a charge distribution being perturbed by an electronegativity difference of 0.1 eV between two points -- the goal of QTPIE is to damp the perturbation when those two points are far separated, and QTPIE makes no distinction between whether that electronegativity difference comes from chi (between atomic nuclei+core electrons) or phi (from an external field).
> 
> Is it superior to have a new method that does? Why should charge transfer be range-independent under electronegativity difference, but range-dependent under an external field?

This fix is merely an alternative method to implementing an external electric field with `fix qeq` and leads to more physical charges than using `fix qeq` and `fix efield` for non-perfectly-conducting systems.

doc/src/fix_qeq_rel_reaxff.rst
navlalli: I'll link a paper in a few weeks, rather than making the documentation longer.

doc/src/fix_qeq_rel_reaxff.rst
navlalli: This is a good point and next time I will do it like you said. Given that this is already used in `fix qtpie/reaxff`, I think it would probably be less harmful to leave it as is. That being said, I modified the code so that I am not changing between Angstrom and Bohrradius throughout.

src/REAXFF/fix_qeq_rel_reaxff.cpp
navlalli: Thanks for these points. I've updated the code to implement these changes

src/REAXFF/fix_qeq_rel_reaxff.cpp
navlalli: Good point, I've updated the code to precompute it.

src/REAXFF/fix_qeq_rel_reaxff.cpp
navlalli: Correct, I've included this change :)

navlalli:
Thank you for the reviews @stanmoore1 and @srtee ! And for updating the name @akohlmey ! Please let me know if there is anything else.

stanmoore1:
@srtee 
>I am not sure what the physical motivation is for this method compared to QTPIE

As I understand, QTPIE requires more parameters than QEq, which may not be available with a given forcefield. This fix allows QEq to be used with fields in directions with periodic BCs, which isn't supported by the original `fix qeq/reaxff`

sjplimp:
@navlalli amd @akohlmey - I agree with Axel on the "qeqr" name not being sufficiently descriptive.  Of the other options you proposed - I like fix qeq/erel/reaxff.  Then be sure to explain in the doc page (if not already done) what the name means.  I.e. your
comment above: 

The erel stands for (what?) because this fix uses electric potential differences and therefore only depends on the relative position of atoms, i.e., the position of atoms relative to nearby atoms (the effective electronegativity of each atom is independent of the reference point chosen as the position of 0 electric potential).

akohlmey (approved):

PR #4511, opened 2025-03-21
Allow extended XYZ dump

### LINKED ISSUE #4471: [Feature Request] Dumping xyz files following extended XYZ convention
**Summary**

Dumping xyz files is pretty basic and does not allow for the extended convention. LAMMPS could include such feature.

**Detailed Description**

The description of the extended xyz format can be found [here](https://web.archive.org/web/20190811094343/https://libatoms.github.io/QUIP/io.html#extendedxyz).

Currently, ``dump simulation all xyz 200 simulation.xyz`` produces:
produces something like:
```
2000
 Atoms. Timestep: 0
1 0 0 0
1 1.4325 1.4325 1.4325
...
```
when it could produce something like:
```
2000
Lattice="27.2175 0.0 0.0 0.0 27.2175 0.0 0.0 0.0 27.2175" Properties=type:I:1:pos:R:3 Timestep=0
1 0 0 0
1 1.4325 1.4325 1.4325
...
```
In addition, it would be nice to add more data, like ``dump simulation all xyz 200 simulation.xyz id type x y z c_atom_pe c_atom_ke c_atom_st[1] c_atom_st[2] c_atom_st[3]``:
```
2000
Lattice="27.2175 0.0 0.0 0.0 27.2175 0.0 0.0 0.0 27.2175" Properties=type:I:1:pos:R:3:c_atom_pe:R:1:c_atom_ke:R:1:c_atom_st:R:3 Timestep=0
1 0 0 0 <more data columns here>
1 1.4325 1.4325 1.4325 <...>
...
```

This would help reading xyz files, specially for code developed by users. This format is used in [CascadesDB](https://cascadesdb.iaea.org/) and it would be nice to have a proper implementation of this format, since many simulations there were done with LAMMPS.

I would like to participate in the implementation of this feature (Python user).

fxcoudert:
**Summary**

Allow extended XYZ (`extxyz`) as a variant of the `dump xyz` style.

**Related Issue(s)**

Fixes https://github.com/lammps/lammps/issues/4471

**Author(s)**

François-Xavier Coudert, CNRS, fxcoudert@gmail.com

**Licensing**

**Backward Compatibility**

The pull request does not break backward compatibility.

**Implementation Notes**

This is a first version of the pull request, I would really welcome comment.

**Post Submission Checklist**

### DIFF
--- a/src/dump_xyz.cpp
+++ b/src/dump_xyz.cpp
@@ -15,6 +15,7 @@
 #include "dump_xyz.h"
 
 #include "atom.h"
+#include "domain.h"
 #include "error.h"
 #include "label_map.h"
 #include "memory.h"
@@ -32,10 +33,20 @@ static constexpr int DELTA = 1048576;
 DumpXYZ::DumpXYZ(LAMMPS *lmp, int narg, char **arg) : Dump(lmp, narg, arg),
   typenames(nullptr)
 {
-  if (narg != 5) error->all(FLERR,"Illegal dump xyz command");
+  if (narg < 5) error->all(FLERR,"Illegal dump xyz command");
   if (binary || multiproc) error->all(FLERR,"Invalid dump xyz filename");
 
-  size_one = 5;
+  extxyz_flag = 0;
+
+  int iarg = 5;
+  while (iarg < narg) {
+    if (strcmp(arg[iarg],"extxyz") == 0) {
+      extxyz_flag = 1;
+      iarg++;
+    } else {
+      error->all(FLERR,"Invalid attribute {} in dump xyz command",arg[iarg]);
+    }
+  }
 
   buffer_allow = 1;
   buffer_flag = 1;
@@ -44,7 +55,13 @@ DumpXYZ::DumpXYZ(LAMMPS *lmp, int narg, char **arg) : Dump(lmp, narg, arg),
 
   delete[] format_default;
 
-  format_default = utils::strdup("%s %g %g %g");
+  if (extxyz_flag) {
+    size_one = 11;
+    format_default = utils::strdup("%s %g %g %g %g %g %g %g %g %g");
+  } else {
+    size_one = 5;
+    format_default = utils::strdup("%s %g %g %g");
+  }
 
   ntypes = atom->ntypes;
   typenames = nullptr;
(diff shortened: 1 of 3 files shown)


akohlmey:
> This is a first version of the pull request, I would really welcome comment.

Thanks for you contribution.

Rather than adding a flag to dump style xyz, I would prefer you create a new DumpExtXYZ class and place it in the EXTRA-DUMP package folder and then just override or extend functions from the base class as needed. That would be cleaner than having all those if statements everywhere.

src/dump_xyz.cpp
```diff
@@ -159,8 +176,22 @@ void DumpXYZ::write_header(bigint n)
...
     if (!fp) error->one(FLERR, "Must not use 'run pre no' after creating a new dump");
 
-    auto header = fmt::format("{}\n Atoms. Timestep: {}", n, update->ntimestep);
-    if (time_flag) header += fmt::format(" Time: {:.6f}", compute_time());
+    std::string header;
+
+    if (extxyz_flag) {
+      header = fmt::format("{}\nTimestep={}", n, update->ntimestep);
+      if (time_flag) header += fmt::format(" Time={:.6f}", compute_time());
+      header += fmt::format(" pbc=\"{} {} {}\"", domain->xperiodic ? "T" : "F",
```
akohlmey, line 184: Shouldn't there be more fields here for extxyz, e.g. potential energy etc.?
You'd have to create the necessary compute instances internally similar to how e.g. the Thermo class does it.

akohlmey:
@fxcoudert please let me know, if you need help with the changes I proposed. It should be straightforward for me to update your pull request along the lines of my suggestions and then leave it to you to complete the supported properties stored in the comment line.

akohlmey:
Another point is that it could be helpful to add some test to `unittest/formats` that checks the output created by the new dump style for validity. If there is a (simple) validator in the libAtoms package, we could also try to import that (although that will likely be in Fortran, which makes things a bit complicated after LAMMPS has become Fortran-free (except for the Fortran module and external Fortran code in the QUIP package)

akohlmey:
@fxcoudert can you please update us on your intentions and timeline for this PR?

fxcoudert:
> Rather than adding a flag to dump style xyz, I would prefer you create a new DumpExtXYZ class and place it in the EXTRA-DUMP package folder and then just override or extend functions from the base class as needed. That would be cleaner than having all those if statements everywhere.

Just to check that I get it right: the idea would be that the `DumpExtXYZ` class inherits from `DumpXYZ`? I think that works, expect that I will have to modify a bit the argument checking of `DumpXYZ` to be more flexible.

I'll work on it in the next few days, when I get time.

I would welcome someone having a look at the handling of the various cell settings and output of the unit cell lattice. Have I covered all the possible situations?

akohlmey:
> I think that works, expect that I will have to modify a bit the argument checking of DumpXYZ to be more flexible.

Dump style XYZ doesn't take any special arrangements, so that should not be a problem. Any tweaks (e.g. which fields and data to include can be set via dump_modify and that should work nicely with polymorphism.

I would suggest to require dump_modify elements unless there are type label maps that you can use. Also which per-atom or global data to include can be adjusted with dump_modify.

fxcoudert:
> Dump style XYZ doesn't take any special arrangements, so that should not be a problem. Any tweaks (e.g. which fields and data to include can be set via dump_modify and that should work nicely with polymorphism.

That means I have to first understand how dump_modify works 😄

akohlmey:
As I mentioned, I am willing to create a basic template that you would then test, correct, expand and document.

fxcoudert:
> As I mentioned, I am willing to create a basic template that you would then test, correct, expand and document.

For sure, if you have time, I would welcome that! I have never coded in LAMMPS before, so I'm having to learn (some of) its internals.

akohlmey:
> That means I have to first understand how dump_modify works 😄

By overriding the modify_params() method.

akohlmey:
@fxcoudert I cannot push into your branch directly since your GitHub account looks like an "institutional" that usually disable this at the global level by default. I have thus submitted a pull request in your fork that shall update your PR branch.

fxcoudert:
That's weird, cause I have this enabled:

Capture d’écran 2025-03-26 à 20 59 00

In any case, thanks! I'll have a look and update the PR in the next couple of days.

akohlmey:
> That's weird, cause I have this enabled:

Yes, but if that setting is overridden at the institutional level, it becomes meaningless. I've seen this a lot over the years with accounts that were sponsored by an institution in contrast to "free" personal accounts.

akohlmey:
Here are some ideas for enhancements:
- dump_modify force yes/no enables/disables printing forces (default no)
- dump_modify vel yes/no enables/disables printing velocities (default yes)
- dump_modify pe compute-ID add potential energy to the comment line (may use compute thermo_pe which always exists)
- dump_modify temp compute-ID add temperature to the comment line (may use compute thermo_temp which always exists)
- dump_modify press compute-ID add pressure  like above

fxcoudert:
Closing in favour of https://github.com/lammps/lammps/pull/4519

PR #4513, opened 2025-03-22
Add a Howto showing how to convert a bulk system to a slab

akohlmey:
**Summary**

Converting a pre-equilibrated bulk system to a slab geometry requires some trickery when trying to do this with LAMMPS itself. This adds a Howto document discussing the problems and suggesting an approach to address them using the Rhodo benchmark input as example.

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

N/A

**Post Submission Checklist**


doc/src/Howto_bulk2slab.rst
```diff
@@ -0,0 +1,138 @@
...
+since the image flags indicate how many times the box length in x-, y-,
+or z-direction needs to be added or subtracted to get the "unwrapped"
+coordinates.  By changing the box dimension this distance is changed and
+thus those atoms move unphysically relative to their neighbors with zero
+image flags.  Setting image flags forcibly to zero creates problems because
+that could break apart molecules by have one atom of a bond on the top
+of the system and the other at the bottom.
+
+.. _bulk2slab:
+.. figure:: JPG/rhodo-both.jpg
```
ndtrung81, line 22: The snapshots looks more like a lipid bilayer than a Rhodopsin protein in water to me.

doc/src/Howto_bulk2slab.rst
```diff
@@ -0,0 +1,138 @@
...
+   change_box     all z final $(zlo-2.0*v_delta) $(zhi+2.0*v_delta) &
+                      boundary p p f
+   read_dump      rhodo-unwrap.lammpstrj 0 x y z box no replace yes
+   kspace_modify  slab 3.0
+
+Specifically, the :doc:`variable delta <variable>` (set to 10.0)
+represents a distance that determines the amount of vacuum added: we add
+twice its value in each direction to the z-dimension; thus in total
+:math:`40 \AA` get added.  The :doc:`reset_atoms image all
+<reset_atoms>` command shall reset any image flags to become either 0 or
```
ndtrung81, line 67: Comment: The command `reset_atoms image all` is very important for the trickery here. To create the 2 interfaces, I think the value of `delta` would need to be sufficiently large compared to the molecule dimensions (e.g. end-to-end distance, or 2 times the radius of gyration). Of course, this is system specific: short chains like in the example, or really long chains wrapping around the original box multiple times.

ndtrung81 (approved):
This is a nice Howto page.

doc/src/Howto_bulk2slab.rst
akohlmey: Yes, Rhodopsin is a transmembrane protein. I just took the data file from the bench folder.

doc/src/Howto_bulk2slab.rst
akohlmey: I guess I need to add a disclaimer, that this will not work for all systems, especially polymers where chains could be connected across periodic boundaries. Those cannot be converted to slabs. There may be other pathological cases.

jrgissing (approved):

jrgissing:
You might also add the term 'free surface' somewhere for search keyword purposes

akohlmey:
> You might also add the term 'free surface' somewhere for search keyword purposes

Good point. Thanks. Will do.

doc/src/Howto_bulk2slab.rst
akohlmey: done

PR #4514, opened 2025-03-23
Improve handling argument errors with library interface functions

### LINKED ISSUE #4118: [Feature Request] Make lammps_extract_fix consistent with lammps_extract_compute for global arrays
**Summary**

While  `lammps_extract_compute` returns a pointer to a scalar, vector or array, consistent with the `LMP_TYPE_*` argument given to it, `lammps_extract_fix` always returns a pointer to a single value, which is inconsistent with the referenced data structure and withnthe behavior of  `lammps_extract_compute`.

I propose making the behavior consistent by `lammps_extract_compute` returning a pointer to a [copy of a] data structure described by the Type constant.

**Detailed Description**

The C API `lammps_extract_compute` function, when called with style constant `LMP_STYLE_GLOBAL` returns a pointer to a scalar, a vector or an array, depending on the Type constant.

On the other hand,  `lammps_extract_fix`, when called with the same constants, returns a pointer to a single value, regardless whether it was called with `LMP_TYPE_SCALAR`, `LMP_TYPE_VECTOR` or `LMP_TYPE_ARRAY`. This is done by copying the value and returning a pointer to it.

Even though it is explictly warned in the documentation for the C interface and @akohlmey added that info to the Python API documentation, the behavior is inconsistent with `lammps_extract_compute` and with the underlying referenced data structure.

Since  `lammps_extract_fix` already allocates memory and returns the pointer to it, it may be possible to allocate memory for the whole fix data structure (vector, array) and return a pointer to it.

This issue was discussed in the [forums](https://matsci.org/t/lmp-numpy-extract-fix-returns-a-sugle-float-when-used-with-lmp-style-global-and-lmp-type-array/54242) and this feature request suggested there.

Thank you!

### LINKED ISSUE #4417: [Feature Request] Library Interface: Error when using functions like extract_compute with invalid name
**Summary**

`lmp.numpy.extract_compute("bad_name", ...)` will fail with a `nullptr` access error. It would be more useful if an exception would be thrown that tells us e.g. that "bad_name" doesn't exist.

**Detailed Description**

While the `extract_compute` has error handling, there is no error thrown in this case. For bad names it just silently returns a `nullptr`, making it indistinguishable from other `nullptr` cases.

akohlmey:
**Summary**

This pull request implements several features that avoid crashes of the LAMMPS C-library interface functions and improve the behavior when invalid or incorrect arguments are used.

**Related Issue(s)**

Fixes #4417 
Fixes #4118 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No incompatible changes to the library interface are intended.

**Implementation Notes**

The following features are implemented:
- add function to error class that allows to suppress printing error messages immediately, make available in library interface
- add checks to lammps_extract_compute(), lammps_extract_fix() that throw an exception when a non-existing id is used or the unsupported data type and style are accessed

TODO list:
- [x] add check for valid handle to library functions starting at line 3254 in library.cpp
- [x] consolidate calls to Error class functions in library.cpp: consistent format, use Error::NOLASTLINE
- [x] add support for lammps_set_show_error() to library plugin interface
- [x] add support for lammps_set_show_error() to Fortran module
- [x] add support for lammps_set_show_error() to SWIG interface

**Post Submission Checklist**

- [ ] Suitable tests have been added to the unittest tree


akohlmey:
@rbberger This tries to implement your feature request from last December. I have been testing it with the following simple test program (using the melt input with the run command replaced by `run 0 post no`):
```c++
#include <iostream>
#include "library.h"

int main(int, char**)
{
        char errmesg[2048];
        void *handle = lammps_open_no_mpi(0, nullptr, nullptr);
        lammps_file(handle, "in.melt");
        double natoms = lammps_get_natoms(handle);
        double *dvalue = (double *) lammps_extract_compute(handle, "thermo_pe", LMP_STYLE_GLOBAL, LMP_TYPE_SCALAR);
        if (lammps_has_error(handle)) {
                lammps_get_last_error_message(handle, errmesg, 2048);
                std::cerr << errmesg;
        } else {
                std::cout << "value = " << *dvalue / natoms << "\n";
        }
        dvalue = (double *) lammps_extract_compute(handle, "thermo_xxx", LMP_STYLE_GLOBAL, LMP_TYPE_SCALAR);
        if (lammps_has_error(handle)) {
                lammps_get_last_error_message(handle, errmesg, 2048);
                std::cerr << errmesg;
        } else {
                std::cout << "value = " << *dvalue / natoms << "\n";
        }
        dvalue = (double *) lammps_extract_compute(handle, "thermo_pe", LMP_STYLE_ATOM, LMP_TYPE_VECTOR);
        if (lammps_has_error(handle)) {
                lammps_get_last_error_message(handle, errmesg, 2048);
                std::cerr << errmesg;
        } else {
                std::cout << "value = " << *dvalue / natoms << "\n";
        }
        lammps_close(handle);
        return 0;
}
```

This produces the output:
```
[...]
Per MPI rank memory allocation (min/avg/max) = 4.187 | 4.187 | 4.187 Mbytes
   Step          Temp          E_pair         E_mol          TotEng         Press     
         0   3             -6.7733681      0             -2.2744931     -3.7033504    
Loop time of 7.81e-07 on 1 procs for 0 steps with 4000 atoms

value = -6.77337
ERROR in lammps_extract_compute(): Compute thermo_xxx does not exist
ERROR in lammps_extract_compute(): Compute thermo_pe does not compute per-atom vector or array
Total wall time: 0:00:00
```
I've opted to output the name of the library function instead of source file and line. This seemed more helpful, specifically if this is run from the python module.

I'm planning on further checks, e.g. if "handle" is NULL, or other dereferences of potential NULL pointers. I am also thinking about adding a library function that can turn off printing of the error message in the Error class *before* the exception is thrown and thus avoid double printing of errors when the error is checked and re-thrown in the python module.

src/library.cpp
```diff
@@ -79,6 +79,38 @@ static void ptr_argument_warning()
...
+{
+  try {
+    lammps_throw_error(fname, fmt::vformat(format, args));
+  } catch (fmt::format_error &e) {
+    lammps_throw_error(fname, e.what());
+  }
+  exit(1); // to trick "smart" compilers into believing this does not return
+}
+
+template <typename... ARgs>
```
rbberger: ARgs is weird. Just call it Args.

akohlmey: turns out this code is not needed at all.

akohlmey:
Here is a Python demo:
```python
from lammps import lammps, LMP_STYLE_ATOM, LMP_STYLE_GLOBAL, LMP_TYPE_SCALAR, LMP_TYPE_VECTOR

lmp = lammps()
lmp.set_show_error(0)
lmp.file('in.melt')
natoms = lmp.get_natoms();
dvalue = -100.0
try:
  dvalue = lmp.extract_compute("thermo_pe", LMP_STYLE_GLOBAL, LMP_TYPE_SCALAR)
  print("value = ", dvalue/natoms)
except Exception as fail:
  print(fail)
try:
  dvalue = lmp.extract_compute("thermo_xxx", LMP_STYLE_GLOBAL, LMP_TYPE_SCALAR)
  print("value = ", dvalue/natoms)
except Exception as fail:
  print(fail)
try:
  dvalue = lmp.extract_compute("thermo_pe", LMP_STYLE_ATOM, LMP_TYPE_VECTOR)
  print("value = ", dvalue)
except Exception as fail:
  print(fail)

lmp.close();
```
and the output:
```
[...]
Per MPI rank memory allocation (min/avg/max) = 4.187 | 4.187 | 4.187 Mbytes
   Step          Temp          E_pair         E_mol          TotEng         Press     
         0   3             -6.7733681      0             -2.2744931     -3.7033504    
Loop time of 7.02e-07 on 1 procs for 0 steps with 4000 atoms

value =  -6.773368053259247
ERROR: lammps_extract_compute(): Compute thermo_xxx does not exist (src/library.cpp:2461)
ERROR: lammps_extract_compute(): Compute thermo_pe does not compute per-atom vector or array (src/library.cpp:2501)
Total wall time: 0:00:00
```

rbberger (approved):

PR #4516, opened 2025-03-24
wrong qi*qj in the cuda code

Eddy-Barraud:
Hello,

1)
I did quite a big mistake in the CUDA code of the pair_style dpd/coul/slater/long
For some reason I put cutsq[mtype].z (which is the squared Slater cutoff) in place of the particle j charge: extra[j].x

Sorry for that.

2) 
I also modified the example in "examples/PACKAGES/dpd-basic/dpd_coul_slater_long". The read_data function is not reading the Slater boolean, even if present in the file, resulting in no direct contribution <=> E_coul = 0. Therefore, I added pair_coef redefinition. Now you can see the big difference with and without GPU 😅 
Outputs:
[gpu_cpu.txt](https://github.com/user-attachments/files/19432585/gpu_cpu.txt)

**Post Submission Checklist**


akohlmey:
> The read_data function is not reading the Slater boolean, even if present in the file, resulting in no direct contribution

It *is* reading the data correctly since it is the same function that also reads the pair_coeff command. The problem is/was that the *write_data* functionality had a logic bug and would write out the reverse logic and thus the data file, which was apparently created by write_data, had the reversed boolean data. Correcting the data file has the same effect as adding the pair_coeff statements.

I've also recreated the log files with the changed source and backported the changes to the maintenance branch for the third update to the 29Aug2024 stable release.

ndtrung81 (approved):
@Eddy-Barraud thanks for the bug fix. Sorry I missed this bug when reviewing the original code, `prefactor` is already qj, i.e. `extra[j].x` from line 482. Adding the comments is helpful here.

PR #4517, opened 2025-03-25
An Updated Formulation of the Langevin GJF Method

talinke:
**Summary**

We provide an updated formulation of the GJF method in the langevin framework. The fundamental functionality of the fix langevin is maintained, including the results obtained from the gjf option. In the new implementation, we have isolated noise and friction into two independent steps using the splitting formulation of the GJF method. This rewrite of the same algorithm allows a better integration with the NVE/fix langevin structure in LAMMPS. As a result, the new formulation solves several artifacts of the current implementation, including
-	Restart glitch in restart using either vfull or vhalf
-	Unnatural placing in input script of fix langevin gjf that differed from fix langevin
-	Now allows for user to compare implementation directly to equations in publication

In addition, it improves memory performance, declutters the code by separating position and velocity treatment from force update and shortens code by 100+ lines.

**Related Issues**

N/A

**Authors**

Tim Linke (UC Davis, LLNL)
Niels Gronbech-Jensen (UC Davis)

**Licensing**

**Backward Compatibility**

Outside of the gjf option, all changes to the fix langevin are fully backward compatible and reproduce the same exact results. The gjf on-site and half-step options are statistically consistent with old versions. The gjf option is no longer compatible with the scale, tally or zero keyword due to the separation of the gjf option and the force update.

**Implementation Notes**

This fix is implemented by editing the files fix_langevin.cpp and fix_langevin.h. The documentation fix_langevin.rst is updated and now also clarifies correct attribution to the fix_langevin algorithm. The example/gjf files are updated, and new log files are provided. Correctness was verified via a direct comparison with the current implementation.

**Post Submission Checklist**

The new implementation closely follows equation 24 in Gronbech-Jensen, J. Stat. Phys. 191, 137 (2024). Per conversation with @athomps, a future PR can contribute a standalone GJ/GJF fix to include compatibility with additional options such as tally, zero and scale.


akohlmey:
@talinke The support for GJF in fix langevin is such a messy bit of code. There is very little code sharing and lots of hacks and workarounds. I am very worried to pile more changes on top of it.
We have several other Langevin implementations that use separate fixes, in part also including a suitable matching time integrations.

What do you think about unbundling GJF from fix langevin and making it its own fix, possibly with inclusion of time integration? Check out `src/EXTRA-FIX/fix_gle.cpp/h` and `src/EXTRA-fFIX/fix_gld.cpp/h` for example.

@athomps @stanmoore1 any thoughts?

talinke:
We agree that the support for GJF in fix langevin was a messy bit of code. With this PR, we remove the hacks and workarounds and fix important issues exposed to the user in the process. Per conversation with @athomps, we propose to maintain a basic GJF as an on/off switch in the fix langevin and add a new fix in EXTRA-FIX for more advanced usage of GJ methods. To do so, this PR is important to correct the standing issues you mentioned with the current implementation.

akohlmey:
> Per conversation with @athomps, we propose to maintain a basic GJF as an on/off switch in the fix langevin and add a new fix in EXTRA-FIX for more advanced usage of GJ methods.

Sorry, but that is half-measures and as such - in my opinion - worse than a clean split. Just leave support for the keyword(s), so that people that want to use fix langevin with GJF can get an meaningful message about switching
to a different fix. Ideally, you want to keep the (initial) command line for the new fix the same.

Please note that fix langevin has been ported to KOKKOS some time ago (and, if I understood correctly, the GJF parts have caused significant headaches) and thus any changes you make / keep in the regular fix langevin need to be ported to the KOKKOS version, too. @athomps may not be aware of this, but I hope @stanmoore1 can chime in and explain the problems with KOKKOS.

src/fix_langevin.cpp
```diff
@@ -378,61 +335,145 @@ void FixLangevin::setup(int vflag)
...
       for (int i = 0; i < nlocal; i++)
         if (mask[i] & groupbit) {
-          dtfm = force->ftm2v * 0.5 * dt / rmass[i];
-          v[i][0] += dtfm * f[i][0];
-          v[i][1] += dtfm * f[i][1];
-          v[i][2] += dtfm * f[i][2];
-          lv[i][0] = v[i][0];
-          lv[i][1] = v[i][1];
-          lv[i][2] = v[i][2];
+          dtfm = dtf / rmass[i];
```
akohlmey, line 384: when making these kind of changes, the fix will actually do time integration (although only in addition to changes made by fix nve. This should be flagged with setting `time_integrate = 1;` in the constructor (provided this code path is used). This also would likely require checking and confirming the order of fixes. You cannot undo what fix nve has done, if it has not yet done it because of being (re-) defined later. This change confirms my belief that this would be better kept as a separate fix with full time integration included (and thus no longer requiring fix nve).

src/fix_langevin.cpp
```diff
@@ -1140,9 +1030,6 @@ void FixLangevin::copy_arrays(int i, int j, int /*delflag*/)
 int FixLangevin::pack_exchange(int i, double *buf)
```
akohlmey, line 1030: If you are changing the amount of data in pack/unpack routines, you have to update the corresponding buffer size information in the constructor.

akohlmey (changes requested):
After looking at (some of) the changes I more firmly than before believe that GJF functionality should be removed from fix langevin and made a separate fix in the EXTRA-FIX package.

Please also keep in mind that any changes to fix_langevin.cpp/h need to be ported to KOKKOS/fix_langevin_kokkos.cpp/h.  Also, all compilation and unit tests have to pass.

athomps:
@akohlmey raises valid points, particularly about KOKKOS support, and also the mixing of time integration between `fix nve` and `fix langevin` commands.  I think @talinke should focus on creating a stand-alone clean and comprehensive implementation of Gronbech-Jensen integrators for Langevin dynamics.  once that is done, we can remove/deprecate the `gjf` option from `fix langevin`

src/fix_langevin.cpp
talinke: Great point! We will be sure to include this in the new PR.

talinke:
Thank you for your comments! We will refocus our efforts on a standalone fix for advanced usage of the gjf methods.

src/fix_langevin.cpp
talinke: Great catch, thank you.

PR #4519, opened 2025-03-27
Extended XYZ dump

### LINKED ISSUE #4471: [Feature Request] Dumping xyz files following extended XYZ convention

fxcoudert:
**Summary**

Provide an extended XYZ (extxyz) dump style.

**Related Issue(s)**

Fixes https://github.com/lammps/lammps/issues/4471

**Author(s)**

François-Xavier Coudert, CNRS, [fxcoudert@gmail.com](mailto:fxcoudert@gmail.com)
Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

The pull request does not break backward compatibility.

**Post Submission Checklist**

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -71,6 +71,7 @@ src/EXTRA-COMMAND/group_ndx.*       @akohlmey
 src/EXTRA-COMMAND/ndx_group.*       @akohlmey
 src/EXTRA-COMPUTE/compute_stress_mop*.* @RomainVermorel
 src/EXTRA-COMPUTE/compute_born_matrix.* @Bibobu @athomps
+src/EXTRA-DUMP/dump_extxyz          @fxcoudert
 src/EXTRA-FIX/fix_deform_pressure.* @jtclemm
 src/EXTRA-PAIR/pair_dispersion_d3.* @soniasolomoni @arthurfl
 src/EXTRA-PAIR/d3_parameters.h      @soniasolomoni @arthurfl
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -71,7 +71,7 @@ src/EXTRA-COMMAND/group_ndx.*       @akohlmey
 src/EXTRA-COMMAND/ndx_group.*       @akohlmey
 src/EXTRA-COMPUTE/compute_stress_mop*.* @RomainVermorel
 src/EXTRA-COMPUTE/compute_born_matrix.* @Bibobu @athomps
-src/EXTRA-DUMP/dump_extxyz          @fxcoudert
+src/EXTRA-DUMP/dump_extxyz.*          @fxcoudert
 src/EXTRA-FIX/fix_deform_pressure.* @jtclemm
 src/EXTRA-PAIR/pair_dispersion_d3.* @soniasolomoni @arthurfl
 src/EXTRA-PAIR/d3_parameters.h      @soniasolomoni @arthurfl
--- a/src/EXTRA-DUMP/dump_extxyz.cpp
+++ b/src/EXTRA-DUMP/dump_extxyz.cpp
@@ -1,0 +1,159 @@
+/* ----------------------------------------------------------------------
+   LAMMPS - Large-scale Atomic/Molecular Massively Parallel Simulator
+   https://www.lammps.org/, Sandia National Laboratories
+   LAMMPS development team: developers@lammps.org
+
+   Copyright (2003) Sandia Corporation.  Under the terms of Contract
+   DE-AC04-94AL85000 with Sandia Corporation, the U.S. Government retains
+   certain rights in this software.  This software is distributed under
+   the GNU General Public License.
+
+   See the README file in the top-level LAMMPS directory.
+------------------------------------------------------------------------- */
+
+#include "dump_extxyz.h"
+
+#include "atom.h"
+#include "domain.h"
+#include "error.h"
+#include "label_map.h"
+#include "memory.h"
+#include "update.h"
+
+#include <cstring>
+
+using namespace LAMMPS_NS;
+
+static constexpr int ONELINE = 128;
+static constexpr int DELTA = 1048576;
+
+/* ---------------------------------------------------------------------- */
+
+DumpExtXYZ::DumpExtXYZ(LAMMPS *lmp, int narg, char **arg) : DumpXYZ(lmp, narg, arg)
+{
+
+  size_one = 11;
+  delete[] format_default;
+  format_default = utils::strdup("%s %g %g %g %g %g %g %g %g %g");
+
+  // use type labels by default if present
+  if (atom->labelmapflag) {
+    typenames = new char *[ntypes + 1];
+    for (int itype = 1; itype <= ntypes; itype++) {
+      typenames[itype] = utils::strdup(atom->lmap->typelabel[itype - 1]);
+    }
+  }
+}
+
+/* ---------------------------------------------------------------------- */
+
+void DumpExtXYZ::init_style()
+{
+  if (!typenames)
+    error->all(FLERR, Error::NOLASTLINE,
+               "Must use either typelables or dump_modify element with dump style extxyz");
+
+  DumpXYZ::init_style();
+}
+
+/* ---------------------------------------------------------------------- */
+
+int DumpExtXYZ::modify_param(int narg, char **arg)
+{
+  int rv = DumpXYZ::modify_param(narg, arg);
+  if (rv > 0) return rv;
+  return 0;
+}
+
+/* ---------------------------------------------------------------------- */
+
+void DumpExtXYZ::write_header(bigint n)
+{
+  if (me == 0) {
+    if (!fp) error->one(FLERR, Error::NOLASTLINE, "Must not use 'run pre no' after creating a new dump");
+
+    std::string header;
+
+    header = fmt::format("{}\nTimestep={}", n, update->ntimestep);
+    if (time_flag) header += fmt::format(" Time={:.6f}", compute_time());
+    header += fmt::format(" pbc=\"{} {} {}\"", domain->xperiodic ? "T" : "F",
+                          domain->yperiodic ? "T" : "F", domain->zperiodic ? "T" : "F");
+    header +=
+        fmt::format(" Lattice=\"{:g} {:g} {:g} {:g} {:g} {:g} {:g} {:g} {:g}\"", domain->xprd, 0.,
+                    0., domain->xy, domain->yprd, 0., domain->xz, domain->yz, domain->zprd);
+    header += " Properties=species:S:1:pos:R:3:vel:R:3:forces:R:3";
+    utils::print(fp, header + "\n");
+  }
+}
+
+/* ---------------------------------------------------------------------- */
+
+void DumpExtXYZ::pack(tagint *ids)
+{
+  int m, n;
+
+  tagint *tag = atom->tag;
+  int *type = atom->type;
+  int *mask = atom->mask;
+  double **x = atom->x;
+  double **v = atom->v;
+  double **f = atom->f;
+  int nlocal = atom->nlocal;
+
+  m = n = 0;
+  for (int i = 0; i < nlocal; i++)
+    if (mask[i] & groupbit) {
+      buf[m++] = tag[i];
+      buf[m++] = type[i];
+      buf[m++] = x[i][0];
+      buf[m++] = x[i][1];
+      buf[m++] = x[i][2];
+      buf[m++] = v[i][0];
+      buf[m++] = v[i][1];
+      buf[m++] = v[i][2];
+      buf[m++] = f[i][0];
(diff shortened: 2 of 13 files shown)


fxcoudert:
Work in progress, will continue improving it over the next few days.

- Now with more flexibility, allowing to dump (or not) per-atom properties: forces, velocities, mass.

fxcoudert:
There is one quirk in the implementation of the printing of per-atomic properties. Since their number can be changed, the way things currently work with a single `printf` statement and format leads to the following situation:
- we need to pass a fixed number of arguments
- so when the format string is short, not all arguments are used
- I think it's still safe, as long as the buffer is big enough: we're dereferencing the next values without using them, and they're still inside the buffer (so no out-of-bounds access).

To be safe, it'd be better to add some padding to the buffer. I'm not sure where (I don't really understand the logic for the existing version).

----

Other possibility: get rid of the format string, and output each element independently. This might have poorer performance.

akohlmey:
> way things currently work with a single `printf` statement and format leads to the following situation

When adding or removing output, you can change the default format string. Any custom format must be reset. You can also change the "size_one" value for the pack/unpack buffer accordingly and the pack routine can skip those not needed.

I can provide an example patch of how this could be done, if needed.

akohlmey:
@fxcoudert please avoid force pushes to branches once they have been made public like in a pull request. They make my life as a maintainer miserable, since they create diverging git histories and any branches that I keep with possible updates will become invalid or I have to deal with messy merges of overlapping changes. This means that once the PR is submitted you should not do any kind of rebasing or appending to commits after having pushed a change.
This makes for somewhat messy git histories, but helps enormously to avoid errors from merging diverging branches and also aids in tracking down changes.

fxcoudert:
> When adding or removing output, you can change the default format string. Any custom format must be reset. You can also change the "size_one" value for the pack/unpack buffer accordingly and the pack routine can skip those not needed.

I have done so: changing the format string, changing the packing, and size_one. It works fine.

The question are the `snprintf` and `fprintf` calls in `convert_string` and `write_lines`. In a call to `printf`, the number of arguments passed has to be constant, known at compile-time. So when I do this:

```
        snprintf(&sbuf[offset], maxsbuf - offset, format, typenames[static_cast<int>(mybuf[m + 1])],
                 mybuf[m + 2], mybuf[m + 3], mybuf[m + 4], mybuf[m + 5], mybuf[m + 6], mybuf[m + 7],
                 mybuf[m + 8], mybuf[m + 9], mybuf[m + 10], mybuf[m + 11]);
```

The format has between 4 and 11 arguments, and I need to pass it 11 (the longest possible case). It will only print the right number. But are there any circumstances where that reference to `mybuf[m + 11]` while `size_one == 5` leads to out-of-bounds access? Dereferencing outside the array bounds is invalid behaviour, even if you don't use the value after that.

Or asked in another way: how is `buf` sized? How were the values of `ONELINE` and `DELTA` chosen?

akohlmey:
Please add the following change:
```
git diff .github/CODEOWNERS
  diff --git a/.github/CODEOWNERS b/.github/CODEOWNERS
  index 5016d9a644..d286a6fa15 100644
  --- a/.github/CODEOWNERS
  +++ b/.github/CODEOWNERS
  @@ -71,6 +71,7 @@ src/EXTRA-COMMAND/group_ndx.*       @akohlmey
   src/EXTRA-COMMAND/ndx_group.*       @akohlmey
   src/EXTRA-COMPUTE/compute_stress_mop*.* @RomainVermorel
   src/EXTRA-COMPUTE/compute_born_matrix.* @Bibobu @athomps
  +src/EXTRA-DUMP/dump_extxyz          @fxcoudert
   src/EXTRA-FIX/fix_deform_pressure.* @jtclemm
   src/EXTRA-PAIR/pair_dispersion_d3.* @soniasolomoni @arthurfl
   src/EXTRA-PAIR/d3_parameters.h      @soniasolomoni @arthurfl
```
This way you get automatically added as a reviewer if there changes to your contributed code.

akohlmey:
> The format has between 4 and 11 arguments, and I need to pass it 11 (the longest possible case). It will only print the right number. But are there any circumstances where that reference to `mybuf[m + 11]` while `size_one == 5` leads to out-of-bounds access?

Just have multiple snprintf() calls depending on the value of size_one. With velocities and forces being options, there need to be three versions for `size_one==5`, `size_one==8`, and `size_one==11`. If size_one has a different value, just error out.

> Or asked in another way: how is buf sized?

The size of buf is determined using the value of `size_one`. LAMMPS tries to collate communications that happen at the same time into a single buffer and then do the size left/right, top/down, back/front exchanges. The pack/unpack functions (or equivalent) are called before/after the communication.

> How were the values of ONELINE and DELTA chosen?

They are defined at the top of the dump_xyz.cpp file.

.github/CODEOWNERS
```diff
@@ -71,6 +71,7 @@ src/EXTRA-COMMAND/group_ndx.*       @akohlmey
 src/EXTRA-COMMAND/ndx_group.*       @akohlmey
 src/EXTRA-COMPUTE/compute_stress_mop*.* @RomainVermorel
 src/EXTRA-COMPUTE/compute_born_matrix.* @Bibobu @athomps
+src/EXTRA-DUMP/dump_extxyz          @fxcoudert
```
akohlmey:
```suggestion
src/EXTRA-DUMP/dump_extxyz.*          @fxcoudert
```

src/EXTRA-DUMP/dump_extxyz.cpp
```diff
@@ -0,0 +1,259 @@
...
+{
+  int m, n;
+
+  tagint *tag = atom->tag;
+  int *type = atom->type;
+  int *mask = atom->mask;
+  double **x = atom->x;
+  double **v = atom->v;
+  double **f = atom->f;
+  double *mass = atom->mass;
```
akohlmey, line 180:
```suggestion
  double *mass = atom->mass;
  double *rmass = atom->rmass;
```
required for the change below.

src/EXTRA-DUMP/dump_extxyz.cpp
```diff
@@ -0,0 +1,259 @@
...
+        buf[m++] = v[i][1];
+        buf[m++] = v[i][2];
+      }
+      if (with_forces) {
+        buf[m++] = f[i][0];
+        buf[m++] = f[i][1];
+        buf[m++] = f[i][2];
+      }
+      if (with_mass) {
+        buf[m++] = mass[type[i]];
```
akohlmey:
```suggestion
        if (rmass) {
          buf[m++] = rmass[i];
        } else {
          buf[m++] = mass[type[i]];
        }
```
This is for compatibility with atom styles that have per-atom masses.

fxcoudert:
@akohlmey sorry I need a bit more hand-holding. I thought I had understood how to dump potential energy, temperature and pressure. And it works, only as long as the `thermo` frequency is the same as the `dump` frequency.

Otherwise, I get this (the link on the error actually does not work):

```
ERROR: Energy was not tallied on needed timestep
For more information see https://docs.lammps.org/err0022 (src/compute_pe.cpp:88)
```

I thought I could handle it very nicely with `invoked_flag`, only dump when it is available. But somehow it's not working as I expect. Guidance would be welcome.

_After that, I'll add tests, clean up the doc, and I think I'll consider it done for a first implementation. We can refine later if needed._

akohlmey:
> Otherwise, I get this (the link on the error actually does not work):

The link does not (yet) work because the underlying feature is very new. It will work after the next feature release planned for next week.  At the moment, you would have to look at https://docs.lammps.org/latest/Errors_details.html#err0022

This error happens, because you need to add this line in the constructor:
```
  clearstep = 1;
```
This indicates that the dump may invoke computes and tells the Modify class to collect data for computes that are only collected on-demand (pe, stress) during the force evaluation. While testing this, I came across a bug in the code triggering this collection.  This works now for fixed intervals. Somehow it is still broken for `dump_modify <id> every/time` and I have not yet figured out why and whether it affects other dump styles as well.

> I could handle it very nicely with invoked_flag

The value of the invoked_flag is the last timestep when it was invoked.

> Guidance would be welcome

The idea of using modify->get_compute_by_id() is sound, but it has one small flaw: those computes may be replaced with thermo_modify. There are two ways out of this: have dump style extxyz create and query its own computes or have direct access to the computes used by the Thermo class. I have opted for the second way, since we would also need to add support for dump_modify to replace those computes. I think it is quite desirable to have the same data as thermo output.

There is a third way to access thermo data. The last values of thermo output is cached and can be retrieved. This way, you don't need to worry about the computes being usable, but then the data may be outdated and you are dependent on the column header names (which can be changed).

Neither solution is ideal.

akohlmey:
> Somehow it is still broken for `dump_modify <id> every/time` and I have not yet figured out why and whether it affects other dump styles as well.

This issue should be fixed now as well. Looks like not a lot of people have used this feature in combination with dumping a per-atom compute to a dump file.

This makes two bugs found and fixed that must have been around for at least a couple of years with a single new feature.

sjplimp (changes requested):
This all looks fine - thanks for contributing.  Except a more LAMMPS-like name for
the command and source file would be:
dump xyz/extended
dump _xyz_extended.cpp/h`
Not clear if there is a reason not to do that?

akohlmey:
> Not clear if there is a reason not to do that?

@sjplimp 
There is a *very* good reason to *keep* "extxyz": it is the name used by the document describing the format.
The names "extxyz" or "exyz" are quite well established to identify files with this format. The name "xyz/extended" is rather clunky in comparison and will not be as easily identified as what it is representing than "extxyz".

fxcoudert:
Several "extended xyz" formats have been proposed over the years. The most used nowadays, and best formalized, is "ExtXYZ". So I think it is a clearer name, with less ambiguity.

sjplimp:
I see your points.  However, my concern is for how LAMMPS users find the new Extended XYZ dump option, or learn about it if they are not familiar with it.  LAMMPS doc pages have many alphabetized lists of commands.  If all the xyz-related dump styles are together in the lists, then they are easy to find and compare.  The bigger LAMMPS gets, the harder it is to find stuff. 

An example are EAM or Tersoff related pair styles like Finnis-Sinclair EAM.  We call it eam/fs, not fseam.  However we call Modified EAM by meam, not eam/mod.  This is b/c it is actually a completely different set of formulas than EAM and is known by practitioners as MEAM.  There is also a so-called modified Tersoff potential, which we call tersoff/mod, not modtersoff or mtersoff.  B/c it is a small tweak to standard Tersoff.

In my view Extended XYZ is one of a family of XYZ formats, so a name like xyz/extended or xyx/ext would be better for the LAMMPS community.  Likewise if support for other enhanced XYZ formats were added later, the LAMMPS names would be xyx/something.

But it's your call - I'm fine if you both decide to keep this as dump extxyz.  Over the years, we have not been totally consistent ourselves with what I've described.

akohlmey:
> If all the xyz-related dump styles are together in the lists, then they are easy to find and compare. The bigger LAMMPS gets, the harder it is to find stuff.

I don't think this is much of a problem in this specific case. The page with all dump style variants is rather small and since extxyz is an established keyword for this specific format, it is harder to find it, if you call it xyz/extended. On the dump.html page, it is hard to find *any* dump style that does not have its own page.

> But it's your call - I'm fine if you both decide to keep this as dump extxyz.

Then please approve so that your block to merge from requesting changes is lifted.

sjplimp (approved):

Issue #4526, opened 2025-04-02
[BUG] ML-IAP broken with Kokkos when using Python


rbberger:
**Summary**

The ML-IAP package, when being build with KOKKOS and Python support fails when trying to run the `examples/mliap/*.pytorch*` examples.

One reason is due to PyTorch changes, which is now more restrictive about the pickle format. One workaround is:

```
export TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD=1
```

Otherwise, the pickled files can not be loaded. Either document this or change something in the save / load logic to avoid this issue.

The second, more severe reason is this line:

https://github.com/lammps/lammps/blob/8defe0e7981d2fef26de959c9d876d350dba77d9/src/KOKKOS/mliap_model_python_kokkos.cpp#L81

It is not valid (was it ever?), since apparently the Python parts don't  use `coeffelem` at all. And since `read_coeff` is overwritten, it remains `nullptr`. Trying to create a `k_coeffelem` from it will always fail and cause a segfault.

**LAMMPS Version and Platform**

`develop`

**Expected Behavior**

Example runs.

**Actual Behavior**

PyTorch error during `load`. With workaround, segfault.

**Steps to Reproduce**

```bash
# example that triggers this:
cd examples/mliap
../../build/lmp -in in.mliap.pytorch.Ta06A -k on g 1 -sf kk -pk kokkos newton on neigh half
```

**Further Information, Files, and Links**

bathmatt:
So, the LAMMPS_NS::MLIAPModelPythonKokkos<Kokkos::Cuda>::read_coeffs reads the coeffs but doesn't copy them into model->coeffelem, or even allocate that array.  We can check on the call on line 81 of 
```
 if (coefffilename && model->coeffelem) MLIAPModelKokkos<DeviceType>::set_k_coeffelem(); 
```
to avoid this, or we can reach into the python and try to fill that data locally, I'm not sure the right thing to do.

bathmatt:
That change provides this 
```
Per MPI rank memory allocation (min/avg/max) = 19.37 | 19.37 | 19.37 Mbytes
   Step          Temp          E_pair        c_energy        TotEng         Press         v_press    
         0   300           -11.85157      -11.85157      -11.813095      2717.1661     -2717.1661    
        10   296.01467     -11.851059     -11.851059     -11.813095      2697.4796     -2697.4796    
        20   284.53666     -11.849587     -11.849587     -11.813095      2289.1527     -2289.1527    
[...]

PR #4532, opened 2025-04-04

akohlmey:
**Summary**

This pull request combines multiple small changes and bug fixes that don't warrant a pull request of their own.

**Related Issue(s)**

Might affect issue #4526 
Addresses bug report from https://matsci.org/t/segementation-fault-of-any-fix-using-atomstyle-template/62905

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- add feature to LAMMPS-GUI to visualize "molecules" from the molecule command
- fix bug in neighbor list code for atom style template when using templates without bonds and thus without special neighbors
- remove unused cutghost member from pair styles mliap and mliap/kk and disable support for the corresponding flag with unified potentials (that setting was leaking memory and producing incorrect forces/energies, too).
- plug several memory leaks when using pair style mliap/kk
- plug memory leak when using SO3 descriptor with pair style mliap/kk
- plug memory leaks when using pair styles pace/kk and pace/extrapolation/kk
- avoid uninitialized data access in pair style coul/long/kk
- plug memory leaks when calling ev_init() directly or indirectly from KOKKOS styles
- add post_constructor() to fix acks2/reaxff to reorder operations for consistency and to avoid memory leak
- add explanation for "boost neigh_modify one" errors to Error_details.html page and modify the corresponding error message to include the URL
- fix bug in OpenMP pragmas in pppm/dielectric and pppm/disp/dielectric
- fix memory leaks in timestep test binary
- fix out-of-bounds memory access in testing the fortran interface to fix external
- add some more valgrind suppressions

**Post Submission Checklist**


akohlmey:
@stanmoore1 so far KOKKOS package related changes were only tested with Kokkos/OpenMP through the testers in the unittest tree.

akohlmey:
@lubbersnick FYI, this pull request removes the option to request neighbors of ghost atoms for mliap unified. This is because, 1) I could not find any code in LAMMPS that uses this feature, 2) it was leaking memory when using it with KOKKOS, and 3) it was producing inconsistent results in the `test_mliappy_unified.cpp` test (for quite some time now, it did work at some time, or else I would not have added the test).

If this setting is needed we can restore this, but then we need two things to happen:
- we need somebody to address the issues with respect to leaking memory in Kokkos/OpenMP and the inconsistent results.
- we need to have some (more) example/code included in LAMMPS that allows us to test and verify this feature. we cannot maintain code where we cannot easily validate its correctness.

lubbersnick:
Neighbors of ghost atoms are required in any MLIAPUnified model which requires more than two message-passing hops.  (Although the recent Kokkos-only changes can get around this at the expense of using far more communication). This is a generic property of NN MLIAPs; this capability is not an edge case for MLIAPUnified. There are a lot of other potentials in LAMMPS which require `ghostneigh=1`. We followed the recipes--I believe based on Tersoff/GPU.  Is there any indication of why that might be wrong?

We have an example which shows where the feature is needed here:

https://github.com/lanl/hippynn/blob/development/examples/lammps/in.mliap.unified.hippynn.Al.ghostneigh#L27

I can encapsulate an example such as this into LAMMPS. I did discuss this briefly with @athomps recently. I think the biggest obstacle is probably that adding pytorch into automated build systems seems to add a lot of overhead to making the environment. I don't know if you have any advice on that possibility. It might also be possible to implement a toy example in numpy - more similar to how the test you mentioned is structured.

akohlmey:
> Is there any indication of why that might be wrong?

I have been tracking down memory issues with valgrind and they went away when removing the cutghostneigh array. So it has not been implemented correctly.

It is very difficult to maintain code that isn't been used by anything in the distribution. and having to install pytorch just for testing is a very big ask. Pytorch is a bit of a monster. I think the recent breakage was exposed by changes where the existing code was very obviously (to me and Aidan) incorrect yet giving the expected results, but the refactored/corrected code broke the test. Because of the complications from building the cython based code and testing without installing pytorch, there is no automated testing and thus over time there is breakage to be expected.

It may be simpler for me to re-introduce the feature in a consistent fashion rather then restoring the inconsistent and rather messy code that was there before. My problem is that my experience in KOKKOS programming is rather limited, but after having plugged some issues over the last week, I may be in a better situation of being aware of the pitfalls with allocations in the base class and the derived classes.

Would there be a way that *you* can set up some automated testing for your code that is tracking the LAMMPS develop branch so we get feedback on when things go wrong faster.

I'll provide an update after experimenting for a bit to restore the flag.

lubbersnick:
Yes, I also do have Kokkos experience either, and was myself only involved with the non-Kokkos version. If the non-kokkos version works, I think I need to pass the buck to @bathmatt .  (Also in that case, perhaps we can simply disable the feature on kokkos side only?). Please let me know if it is broken in both kokkos, and non-kokkos, or if only one of those.

Regardless we will look into automated testing of our LAMMPS interface. This is also not trivial for us, so I don't have an exact timeline. At the same time, I am in complete agreement that having this code regularly tested would be highly beneficial.

akohlmey:
@lubbersnick I have restored the option to request neighbor lists with neighbors of ghosts, but I have not restored the other bits and pieces that in the descriptor classes that were inconsistent. With this my test executable now is reporting correct results (as it should) for both neighbor list settings. This is testing the plain non-KOKKOS pair style only. So this is good news. I will now try to expand the test executable to also test with Kokkos/OpenMP to see how this affects the results.

If you are curious, you could try to check out and compile LAMMPS from this specific branch and check if it works for your "real" applications not just the dummy LJ test that I am using.

akohlmey:
@lubbersnick after some more tests and updates here is the "best and final" version  that I can do.
I've found and fixed a small memory leak in the .pyx files related to storing element names. Also the handling of element names in the C++ code is now consistent. The plain and Kokkos/OpenMP version of the test give results consistent to the LJ reference, the version with the extended neighbor list not. But the calculations finish and do not crash. Also valgrind is only complaining about MPI and Python internal stuff as far as I can tell, and they are too many for creating valgrind suppressions.

lubbersnick:
Thanks, we will test with this version, although the time until that is complete is not yet known, so do not let us be blocking on your PR.

akohlmey:
@ndtrung81 Somehow I managed to break the pppm/dielectric unit test, but I have no clue how I did this.

akohlmey:
> @ndtrung81 Somehow I managed to break the pppm/dielectric unit test, but I have no clue how I did this.

@ndtrung81 found and fixed the bug. See commit fc3d295560590e3843de6ccdbd9a1f9cb20fb62b

ndtrung81:
@akohlmey thanks for the quick fix!

akohlmey:
> @akohlmey thanks for the quick fix!

I got lucky.

doc/src/Errors_details.rst
```diff
@@ -982,3 +982,52 @@ order of preference there are:
...
+   there is not enough space left and then allocates a new page.
+
+#. There are fewer calls to the memory allocator functions (which can be
+   time consuming for long-running jobs and fragmented memory space) and
+   the resulting neighbor lists are close to each other physically which
+   improves cache efficiency.
+
+This is controlled by the two parameters "one" and "page", respectively,
+that can be set via the :doc:`neigh_modify command <neigh_modify>`. The
+parameter "one" is the estimate for the number of entries in a single
```
jtclemm: This sounds a bit like the number of pairs in the entire neighbor list (extensive). Doesn't *one* specify the number of neighbors per atom (intensive)?

src/OPENMP/npair_bin_omp.cpp
```diff
@@ -207,9 +207,9 @@ void NPairBinOmp<HALF, NEWTON, TRI, SIZE, ATOMONLY>::build(NeighList *list)
               if (molecular != Atom::ATOMIC) {
                 if (!moltemplate)
                   which = find_special(special[i], nspecial[i], tag[j]);
-                else if (imol >= 0)
-                  which = find_special(onemols[imol]->special[iatom], onemols[imol]  ->nspecial[iatom],
-                                       tag[j] - tagprev);
+                else if ((imol >= 0) && onemols[imol]->special)
+                  which = find_special(onemols[imol]->special[iatom],
+                                       onemols[imol]  ->nspecial[iatom], tag[j] - tagprev);
```
jtclemm: is the space in `onemols[imol]  ->nspecial[iatom]` intentional?

doc/src/create_atoms.rst
```diff
@@ -395,6 +396,12 @@ correct number of particles are inserted, in a perfectly random
 fashion.  Which lattice sites are selected will change with the number
 of processors used.
 
+.. versionadded:: TBD
+
+The *group* keyword adds the newly created atoms to the named
+:doc:`group <group>`.  If the group does not yet exist it will be
+created.  All created atoms are always added to the group "all".
```
jtclemm: Maybe worth specifying that only the last group keyword will be used? I.e. one cannot list two groups and expect atoms to be added to both.

doc/src/Errors_details.rst
akohlmey: Rephrased.

src/OPENMP/npair_bin_omp.cpp
akohlmey: No.

doc/src/create_atoms.rst
akohlmey: Good point.

jtclemm (approved):

src/OPENMP/npair_bin_omp.cpp
jtclemm: Oh I see, that was an existing space that was overlooked. It is mind numbing updating all the npair classes...

akohlmey: That is why I often use macros in Emacs for that. :robot:

doc/src/Errors_details.rst
```diff
@@ -982,3 +982,57 @@ order of preference there are:
...
+   larger than individual lists, LAMMPS just "fills up" the page until
+   there is not enough space left and then allocates a new page.
+
+#. There are fewer calls to the memory allocator functions (which can be
+   time consuming for long-running jobs and fragmented memory space) and
+   the resulting neighbor lists are close to each other physically which
+   improves cache efficiency.
+
+This is controlled by the two parameters "one" and "page", respectively,
+that can be set via the :doc:`neigh_modify command <neigh_modify>`. The
```
sjplimp, line 1009: Why does this sentence say estimate?  The neigh_modify doc page says one is the max # of neighbors one atom can have.   If an atom has more than "one" neighbors, does the current code throw an error?  I think it used to.

akohlmey: Because the (embedded doxygen) documentation for MyPage said so: 
```
 * \param  user_maxchunk   Expected maximum number of items for one chunk
```
But looking at the `vget()` and `vgot()` methods this is a hard requirement. So I will have to update both, the explanation and the embedded documentation for MyPage.

akohlmey: @sjplimp done

PR #4536, opened 2025-04-06
Remove CMAKE_TUNE_FLAGS

### LINKED ISSUE #4525: [BUG] Setting CMake global compile flags can cause issues
**Summary**

`CMAKE_TUNE_DEFAULT` adds compiler flags globally. This can affect builds of utility libraries such as libyaml and google test. E.g. if NVCC specific flags are added, gcc and gfortran compilers in those utilities will error out.

They become dependent on the cmake execution order, which is why it may only show up after certain CMake runs.

Further motivation for #4233 and a target based approach. CMake generator expressions may need to guard these kind of flags.

It's also bad practice to prefix any variable with `CMAKE_`. LAMMPS should use its own `LAMMPS_` prefix for such variables.

`develop`

**Expected Behavior**

No errors.

**Actual Behavior**

Errors due to unknown flags `-Xcudafe --diag_suppress=unrecognized_pragma,--diag_suppress=128` for both gcc and gfortran.

**Steps to Reproduce**

akohlmey:
**Summary**

This pull request removes the CMAKE_TUNE_FLAGS setting from LAMMPS' CMake configuration.
The setting was created for convenience, but has caused as much (or more?) problems than it has solved.
There are also several CMake best practices that its use violates.

**Related Issue(s)**

Fixes #4525 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

No. You have to set `CMAKE_<language>_FLAGS` or `CMAKE_<language>_FLAGS_<build type>` instead.

**Implementation Notes**

This implements similar functionality by appending the settings to the `CMAKE_<language>_FLAGS_<build type>` variables. Since `CMAKE_<language>_FLAGS` overrides the build type specific settings, it is not appended there unless the variable is set (and the build type specific flags correspondingly ignore).

**Post Submission Checklist**


akohlmey:
@stanmoore1 @rbberger @giacomofiorin This will require extensive testing for the various use cases where `CMAKE_TUNE_FLAGS` was previously applied. I'm submitting this in its unfinished and untested stage to get feedback on the overall concept.

cmake/CMakeLists.txt
```diff
@@ -102,49 +102,49 @@ include(CheckIncludeFileCXX)
...
+  set(CMAKE_CXX_FLAGS_${BTYPE} "${CMAKE_CXX_FLAGS_${BTYPE}} -Minform=severe")
 endif()
 
 # this hack is required to compile fmt lib with CrayClang version 15.0.2
 # CrayClang is only directly recognized by version 3.28 and later
 if(CMAKE_VERSION VERSION_LESS 3.28)
   get_filename_component(_exe "${CMAKE_CXX_COMPILER}" NAME)
   if((CMAKE_CXX_COMPILER_ID STREQUAL "Clang") AND (_exe STREQUAL "crayCC"))
-    set(CMAKE_TUNE_DEFAULT "-DFMT_STATIC_THOUSANDS_SEPARATOR")
+    set(CMAKE_CXX_FLAGS_${BTYPE} "${CMAKE_CXX_FLAGS_${BTYPE}} -DFMT_STATIC_THOUSANDS_SEPARATOR")
```
rbberger: This is FMT specific and should be a `target_compile_definition` on the library target.

cmake/CMakeLists.txt
```diff
@@ -102,49 +102,49 @@ include(CheckIncludeFileCXX)
 if(CMAKE_CXX_COMPILER_ID STREQUAL "Intel")
   if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
     if(CMAKE_CXX_COMPILER_ID STREQUAL "Intel")
-      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /Qrestrict")
+      set(CMAKE_CXX_FLAGS_${BTYPE} "${CMAKE_CXX_FLAGS_${BTYPE}} /Qrestrict")
```
rbberger: What is the rationale to set the Config flags instead of just `CMAKE_CXX_FLAGS`? Both `CMAKE_CXX_FLAGS` and `CMAKE_CXX_FLAGS_<Config>` are passed, in that order, to the compiler.

doc/src/Build_settings.rst
```diff
@@ -659,40 +658,3 @@ code has to be set up to *catch* exceptions thrown from within LAMMPS.
...
-done in the ``main()`` function, this applies only to the standalone
-executable, not the library.
-
-.. tabs::
-
-   .. tab:: CMake build
-
-      .. code-block:: bash
-
-         -D CMAKE_TUNE_FLAGS=-DLAMMPS_TRAP_FPE
```
rbberger, line 688: This should be it's own CMake option and enable the `LAMMPS_TRAP_FPE` define if set.

cmake/CMakeLists.txt
akohlmey: done

doc/src/Build_settings.rst
akohlmey: I am removing this setting. It doesn't work as expected most of the time and I don't know of anyone outside of me that has ever used it.

giacomofiorin (approved):

giacomofiorin:
> @stanmoore1 @rbberger @giacomofiorin This will require extensive testing for the various use cases where `CMAKE_TUNE_FLAGS` was previously applied. I'm submitting this in its unfinished and untested stage to get feedback on the overall concept.

Hi @akohlmey! As far as I am concerned this looks great to me. I have never used `CMAKE_TUNE_FLAGS`. Flags shared by Colvars and LAMMPS are communicated using `target_compile_definitions()` or `target_compile_options()`. If there is something else I can look into please let me know?

akohlmey:
> If there is something else I can look into please let me know?

No, I am mostly waiting on @stanmoore1 and @rbberger to let me know if this PR is not creating unexpected problems, specifically with KOKKOS and external ML packages.

PR #4542, opened 2025-04-10
Integrate header-only JSON library

akohlmey:
**Summary**

This is integrates the header-only JSON library we discussed at the last LAMMPS developer call. 

**Author(s)**

Axel Kohlmeyer, Temple U.

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

Same as with the {fmt} library, the name of the namespace and the include guards are changed so there is no conflict with any application also using the same JSON library.

**Post Submission Checklist**


akohlmey:
@jrgissing it might be helpful if you or your student can provide additional features to this pull request using the JSON class, so we get a better impression of what the final code would look like.

We can probably play around with this branch for a bit and then at one of the next meetings discuss it and either keep it or start over. Most likely the latter.

Please note that nothing here is set in stone. I've prepared this pull request because I find it much easier to discuss a real application with real source code compared to hypothetical implementations.

jrgissing:
thanks Axel! I have not looked closely yet, but in this thermo output demo, is the entire json file also being stored in memory?

jrgissing:
after reading your description a little closer, a couple thoughts:

although nlohmann/json is probably indispensable for reading/parsing and possibly writing short but complicated files, it may be a hindrance for writing simple, but long and repetitive jsons. For example, this is probably not scalable to dump files. However, a 'streaming' version of json may be possible by deleting/overwriting the final (or last couple) '}' in the json file. 

Also, can we use 'run_1', 'run_2' as keys for multiple runs?

akohlmey:
> but in this thermo output demo, is the entire json file also being stored in memory?

Yes. This is a shortcoming of going through the json class for outputting the file.
Thus this is not a very good use case, but I have deliberately used it to demonstrate this.
I don't think that this is a good solution.

Output of a file/data bypassing the class library is what is done for thermo_style yaml, where the YAML data is used instead of the (default) "line" format or the sometimes used "multi" format.

> However, a 'streaming' version of json may be possible by deleting/overwriting the final (or last couple) '}' in the json file.

That is a very bad idea as it has a very high risk of creating corrupted output files.

Yes, also dump files is a problem, if you want a single file. But here it would be simpler to use the json class since you could do the same as for dump style cfg and require that the output has a "*" and then write just one file per timestep with the timestep included into each dump file. Though, unlike with the native dump, you will need to collect the entire data on a single processor and thus you have a similar scaling problem as with the thermo  output option.

> Also, can we use 'run_1', 'run_2' as keys for multiple runs?

It doesn't make sense with the current implementation unless you do a similar thing than what a hypothetical dump style json would do (as suggested above) and use a different name for the output (and require an output with a "*"). This - same as dump style cfg - has the potential to clutter your output folder with producing a large number of files. Otherwise, you would just grow the memory consumption and - if you write out at intermediate steps - the performance would get increasingly worse as the amount of data grows since you have to overwrite the entire file every time.

Bottom line, using a json class is not well suited for continuous long "streaming" kind of outputs. It reminds me a bit of the limitations of XML. If you try to create standard conforming JSON output, it will be challenging. I think the worst issue is that the file will only be properly closed and in a correct format if the simulation finishes cleanly. YAML is a bit more forgiving in that respect (at the expense of worse performance in parsing), for example you can section its output with "---" and "..." for which I have not found an equivalent in JSON. This will also be a problem with output for fix ave/* and so on.

I don't have a good suggestion for how to move forward on this. The implementation of the class is nice and easy to use (once you figured out the basics), but it works best for small files like molecule files or potential files and forces inefficient use for anything big or continuous (especially for cases where you would run with many parallel processes). It is a bit easier to work around these limitations with the YAML format, but if you would want to use a YAML class for convenience, the same issues would pop up. And having to deal with both, YAML and JSON at the same time doesn't sound like a good solution either.

jrgissing:
> > However, a 'streaming' version of json may be possible by deleting/overwriting the final (or last couple) '}' in the json file.
> 
> That is a very bad idea as it has a very high risk of creating corrupted output files.

How so? To be clear, I mean that we *always* write a valid json, and then add data after deleting the trailing curly bracket. Seems like it would solve both the memory and the streaming or 'unexpected of run' issues.

akohlmey:
> How so? To be clear, I mean that we _always_ write a valid json, and then add data after deleting the trailing curly bracket. Seems like it would solve both the memory and the streaming or 'unexpected of run' issues.

Not all networked/parallel filesystems are as reliable as a local Linux file system. You can easily have corrupted data if - by accident - multiple processes try to append to the same file. I think some of the fix ave/* styles allow rewriting pieces of the output, but that is optional and incurs the same risks. Just look for the calls to `platform::fseek()` and `platform::ftell()` which try to provide consistent behavior with a subset of features across all platforms.

Already with Windows native text output, you have a problem since you have an inconsistent behavior with respect to newlines and carriage return characters between reading and writing. So seeking backward is completely unreliable.
After 30+ years of running on clusters, I'd rather create 1000s of files and concatenate them later than trying to append, let alone partially rewrite a file. Already the though makes the hairs on my back stand up (and there are far too many of those).

akohlmey:
Just to be clear, I don't mind if we drop the JSON thermo output and just have this pull request only integrate the JSON library and then have a second (or third) pull request with actual applications.

If desired, I could add instead a dump style json that would be similar to dump style cfg, but in JSON format as discussed earlier.

akohlmey:
One more thought, since we want to make JSON the standard non-native format for any kind of structured data, there should be some (mini) tutorial on how to use it (also pointing to example applications, that is why I added the thermo stuff) and - more importantly - documenting the conventions for how to structure the data trees in the files.
A starting point for that would be what is already established in either YAML or NetCDF or HDF5 (or rather H5MD).

jrgissing:
I agree that it would be useful to have one PR that just introduces the library, so that subsequent PRs can utilize it

akohlmey:
@jrgissing how do you want to proceed with this?
- merge "as is"?
- drop the thermo code and then merge?
- convert the thermo code to output JSON without using the class (same as YAML)?
- add the hypothetical dump style json as application example
Please let me know.

jrgissing:
I vote 'drop the thermo code and then merge', and hash out other details on another PR

akohlmey:
> I vote 'drop the thermo code and then merge', and hash out other details on another PR

Ok. I took out the relevant commits with a rebase and forced-push.

rbberger:
I'm very much against polluting src with more non-LAMMPS code. What happened to keeping things like this in lib?

jrgissing:
> I'm very much against polluting src with more non-LAMMPS code. What happened to keeping things like this in lib?

the json library seems very different than anything in lib and more similar to the fmt library, in that calls to json are likely to be embedded in the core code (see Axel's recently reverted thermo output in this PR, or #4547)

rbberger:
Honestly, the fmt library should be in lib too. But this was a one-off that is supposed to go away soon. Merging this would "normalize" further inclusions like this...

akohlmey:
> What happened to keeping things like this in lib?

LAMMPS doesn't follow the usual conventions. "lib" is meant for libraries in support of specific *packages*; "kokkos" for KOKKOS, "gpu" for GPU, "poems" for POEMS, "colvars" for COLVARS and so on. It has become a bit more convoluted with "linalg" and "lepton" which are used by multiple packages.

The main problem with "lib" is in the GNU make based build system. 1) It requires a "make lib-xxx" to compile its contents and/or set up the symlinks and 2) it requires a "yes-XXXX" to include the settings from the lib folder in the main makefile.

It **would** work well and I would agree with moving it there, if we put all support for JSON format input or output into a specific JSON package. However, my understanding from the last two developer meetings is that JSON support is supposed to be "everywhere". I even see the "molecule/json" command in EXTRA-COMMANDS from PR #4547 as a temporary solution and would expect a tighter integration at a later point, e.g. the current molecule command would become molecule/lammps and a new wrapper would added that detects the format of the molecule file and then launches either molecule/lammps or molecule/json or ...?

rbberger:
If `lib` has a legacy problem (that hopefully will go away soon too), maybe the answer is to have something like a `third_party`, `extern`, or `utils` folder. Either in the root folder, or I'd even be ok with `src/extern`. My main concern is making an already convoluted `src` folder worse by not grouping these into their own space.

akohlmey:
@rbberger here is a proposal that tries to address your concern but also something that was still worrying me.
The folder with the class has been moved to a "third_party" top level folder.

In "src" I've added a little wrapper "json.h" which - unlike "json.hpp" conforms to the header file name convention of LAMMPS and makes the json class directly an alias in the LAMMPS_NS namespace, so one does not have to remember the awkward "nlohmann_lmp::json" namespace.

I've chosen "third_party" because that could be used for a different approach to the "cannot automatically download" third-party sources on restricted machines. We could offer *two* versions of the LAMMPS sources in the future, the current variant with manual included, and a "complete" package with all third-party sources directly included in the tarball as well.

jrgissing (approved):

PR #4543, opened 2025-04-10
adaptive-precision interatomic potentials (APIP)

d-immel:
**Summary**

The APIP package provides adaptive-precision interatomic potentials (APIP) according to Immel, Drautz and Sutmann, J. Chem. Phys. 162, 114119 (2025).

**Related Issue(s)**

None.

**Author(s)**

David Immel¹³, Ralf Drautz², Godehard Sutmann¹²

1: Jülich Supercomputing Centre (JSC), Institute for Advanced Simulation (IAS), Forschungszentrum Jülich, Jülich, Germany
2: Interdisciplinary Centre for Advanced Materials Simulations (ICAMS), Ruhr Universität Bochum, Bochum, Germany
3: d.immel@fz-juelich.de

**Licensing**

**Backward Compatibility**

This changes do not break backward compatibility.

**Implementation Notes**

Energies and forces calculated with the adaptive-precision versions of ACE and EAM were validated against the normal EAM and ACE implementations of the ML-PACE and the MANYBODY package as shown in examples/PACKAGES/apip/in.validate .

An adaptive-precision potential is conservative when all switching parameters are constant over time; the energy is conserved in this case as demonstrated by examples/PACKAGES/apip/in.vacancy.const.lambda .

The energy is not conserved when the switching parameter is updated by fix lambda due to the neglection of the gradient of the switching parameter; however, this is compensated by fix lambda_thermostat as described in our paper and shown in examples/PACKAGES/apip/in.vacancy.dynamic.lambda .

**Post Submission Checklist**

Immel, Drautz and Sutmann, J. Chem. Phys. 162, 114119 (2025).
https://doi.org/10.1063/5.0245877


d-immel:
@junghans This is the pull request you wanted to be informed about.

junghans:
The CMake part looks good to me, given there isn't much.

src/atom_vec_apip.cpp
akohlmey, line 1: Couldn't this file be put in the APIP package folder? Same for the header?

src/atom.h
akohlmey, line 1: @sjplimp what do you think of the per-atom property names?
All those lambdas make me a bit dizzy and confused. Lambda can mean so many different things.
Perhaps we should start a new tradition and put some package related string in there: `apip_lam`, `apip_lam_inp`, `apip_e_smpl`, `apip_e_cmplx', etc.

src/atom_vec_apip.cpp
d-immel: Yes. I never thought about that. src/APIP is the better dir.

src/atom.h
d-immel: I don't care about the internal names; I'm happy to follow your suggestions. Should this also affect the names visible to the user, e.g. in dump_custom.cpp?

akohlmey: @d-immel Let's not rush into action before we have a second opinion. Generally, we prefer to have consistent names internally and externally. It is less confusing.  Like all things that are present in LAMMPS for a long time, we only discovered over time how important consistency and recognizability are. Some ambiguities have been resolved over time, but others remain in place for support with legacy applications and use cases.

ndtrung81:
@d-immel I have a couple of suggestions for improving consistency with other packages:
1. To signal that the fixes are provided by the APIP package, I would rename `FixLambda` to `FixLambdaAPIP`, `FixLambdaThermostat` to `FixLambdaThermostatAPIP` and `FixApipAtomWeight` to `FixAtomWeightAPIP`
2. `FixStyle(lambda_thermostat,FixLambdaThermostatAPIP)` -> `FixStyle(lambda/thermostat,FixLambdaThermostatAPIP)`
3. The pair style classe names:  `PairEAMapip` -> `PairEAMAPIP` and `PairEAMFSapip` -> `PairEAMFSAPIP` and similar ones for the APIP suffix (and `/apip` in the style names).

d-immel:
@ndtrung81 

>     1. To signal that the fixes are provided by the APIP package, I would rename `FixLambda` to `FixLambdaAPIP`, `FixLambdaThermostat` to `FixLambdaThermostatAPIP` and `FixApipAtomWeight` to `FixAtomWeightAPIP`
> 
>     2. `FixStyle(lambda_thermostat,FixLambdaThermostatAPIP)` -> `FixStyle(lambda/thermostat,FixLambdaThermostatAPIP)`

According to my understanding, the internal class names and the user-visible command names should correspond to each other apart from the leading Fix/Pair/Compute and upper/lower case. Your suggestions affect only the class names and would introduce inconsistencies in the naming scheme. Is this intentional? I can of course change names, but I would rather do it once than twice.

>     3. The pair style classe names:  `PairEAMapip` -> `PairEAMAPIP` and `PairEAMFSapip` -> `PairEAMFSAPIP` and similar ones for the APIP suffix (and `/apip` in the style names).

Good point. Capital letters are used for the most abbreviations in class names. I'll change this as soon as we agree on the other aspekt.

sjplimp:
Hi @d-immel - thanks for submitting your work to LAMMPS as a new APIP package.
This seems like a powerful capability.

Here are my thoughts on naming conventions.  Once these are addressed, I will read the new
doc pages and provide any additional feedback.  It looks like this package is well-designed to
be separate from main LAMMPS, so I don't anticipate any big issues with merging it.

(a) internal variable names

I agree with Axel that everything being added to atom.cpp/h should be prefixed with "apip", e.g. 

double *apip_lambda;
double *apip_e_simple;
int apip_lamdba_flag;

As Axel noted, the developers should probably start doing this for some other existing packages as well.

Note this does not restrict how you use these variables in files within the APIP package.  E.g. you can do this in one of your own files if you like:

double *lambda = atom->apip_lambda;

Exception: you need to use the "apip" names in atom_vec_apip.cpp/h.

(b) naming of input script commands and assocated class files

I agree with Trung that all commands in the APIP package should  have "apip" (non-capitalized) in their input script form,
which will affect the doc files as well:

fix lambda/apip
pair_style eam/fs/apip

And likewise in the file name, and in the file class.  It would be fine to sometimes not capitalize APIP in the class name to avoid
something like EAMFSAPIP, if you prefer.  Or to use an underscore: EAMFS_APIP.

(c) access to per-atom APIP quantities from input script

dump custom - don't add anything there
fix store/state - don't add anything there
variable - don't add anything there
compute property/atom - this is the place to add them with "apip" names - dump custom, fix store/state, variable can access them via compute prop/atom
set - OK to add lambda here, but make it apip_lambda
read_dump - same as set, use apip_lamba

Not clear why you are adding options to write many APIP variables to a dump file (now via compute property/atom), but only to read lambda from a dump file?  What will be the typical use-case for a user?  Maybe lambda is needed to restart a simulation from a dump file, but the other APIP per-atom quantites are derived anew and thus don't need to be read-in ?

d-immel:
Hi @sjplimp , thanks for your thoughts and the outlook on how to proceed.

I'll follow your suggestions regarding (a) internal variable names and (b) naming of input script commands and assocated class files.

(c) access to per-atom APIP quantities from input script

Using set to initialise lambda and reading lambda from a dump file to restart is sufficient for simulations with constant switching parameters lambda. Thus, lambda as exception that remains in set and read_dump is fine for me. This is exactly the intended and typical use-case for users. I was not aware of the compute property/atom command when I started working on this. The many APIP variables in dump are legacy issues. I'll clean them up and maybe add some in property/atom.
For dynamic switching parameters, one needs to use binary restart files anyway.

d-immel:
@sjplimp @ndtrung81 As discussed, I've changed
1. internal variable names
2. naming of input script commands and assocated class files
3. access to per-atom APIP quantities from input script.

Furthermore, I resolved the merge conflict arising from the recently updated set command.

Do you have any additional feedback?

ndtrung81:
@d-immel thanks for the updates which look good to me. I also tested the 4 example input scripts, all run successfully. The reference log files need to be updated with the changes in the input scripts.

jmgoff (approved):
APIP package and relevant changes are well-documented.

The package itself, the new fixes, computes, and pair_styles look good and are consistent with other similar fixes, computes, etc.

Examples offer a good number of use cases and have been tested successfully.

The names of the new per-atom properties are a little lengthy in comparison to other per-atom properties. However I think this is appropriate, at least for now, because the names are appropriately descriptive.

d-immel:
@ndtrung81 The reference log files are updated.
@jmgoff Thank you for the review.

src/atom.cpp
```diff
@@ -3304,6 +3333,17 @@ int Atom::extract_datatype(const char *name)
   if (strcmp(name,"curvature") == 0) return LAMMPS_DOUBLE;
   if (strcmp(name,"q_unscaled") == 0) return LAMMPS_DOUBLE;
 
+  // PACE package
```
sjplimp: this should be APIP package ?

doc/src/Howto_apip.rst
```diff
@@ -0,0 +1,200 @@
+Adaptive-precision interatomic potentials (APIP)
+================================================
+
+The :ref:`APIP package <PKG-APIP>` allows to use adaptive-precision potentials
```
sjplimp: ... PKG-APIP enables use of adaptive-precision potentials as described in (citation)...

doc/src/Howto_apip.rst
```diff
@@ -0,0 +1,200 @@
...
+according to :ref:`(Immel) <Immel2025_1>`.
+The potential energy :math:`E_i` of an atom :math:`i` described by an
```
sjplimp, line 30: I suggest adding a sentence or two here which explains/motivates up front what an APIP is (in words) before
showing the energy formula.  Something like:

An APIP model is one which combines a traditional empirical potential (fast, less accurate) with
a machine learned potential (slower, more accurate) using a lambda parameter to weight
the contribution of each component to an atom's energy and force.  As explained below, lambda
values can be defined for each atom and adjusted dynamically during a simulation.

Also both here in the Howto and in the Packages doc page I suggest adding a sentence up front to say that in this context, precision = accuracy of the potential.  Because "precision" has an alternate meaning in MD, especially for GPUs. E.g. 32-bit versus 64-bit precision.

doc/src/Howto_apip.rst
```diff
@@ -0,0 +1,200 @@
...
+
+   How to select the values of the parameters of an adaptive-precision
+   interatomic potential is discussed in detail in :ref:`(Immel) <Immel2025_1>`.
+
+
+.. tabs::
+
+   .. tab:: dynamic switching parameter
+
+      The affected parts of a LAMMPS script can look as follows:
```
sjplimp, line 88: can remove this line, since the code is in the Examples section

sjplimp: Lines like these would appear in the input script:

doc/src/Howto_apip.rst
```diff
@@ -0,0 +1,200 @@
...
+
+      An adaptive-precision EAM-ACE potential, for which the switching parameter
+      :math:`\lambda` is calculated from the CSP, is defined via
+      :doc:`pair_style hybrid/overlay <pair_hybrid>`.
+      The fixes ensure that the switching parameter is calculated, the energy conserved,
+      the weight for the load balancing calculated and the load-balancing itself is done.
+
+   .. tab:: constant switching parameter
+
+      The affected parts of a LAMMPS script can look as follows:
```
sjplimp: Lines like these would appear in the input script:

sjplimp (changes requested):
All looks good - thanks for the refactoring.  Just a couple minor doc page suggestions, then I will approve the PR.

d-immel:
@sjplimp Thank you for the review. I've updated the doc pages as suggested.

sjplimp (approved):

PR #4554, opened 2025-04-17
Add pair style lj/pirani

mateors6:
**Summary**

This pull adds a new pair style: an Improved Lennard-Jones (ILJ) potential, by Pirani et al. Phys. Chem. Chem. Phys., 2008,10, 5489-5503

**Related Issue(s)**

**Author(s)**

Mateo Rodríguez (mateorsuarez@gmail.com) ( Molecular Interactions Group INTERMOL IFF-CSIC, http://intermol.iff.csic.es)

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**


doc/src/pair_lj_improved_cut.rst
```diff
@@ -0,0 +1,153 @@
+.. index:: pair_style lj/improved/cut
+
+pair_style lj/improved/cut command
+=============================
```
akohlmey:
```suggestion
==================================
```
Underlines must match the length of the text or sphinx is unhappy.

akohlmey:
@mateors6 I've added some simple optimizations (you want to check with some existing input how much of a difference that makes), fixed a few issues with the doc file, and added a test input.
That test produces invalid forces for multi-cutoff r-RESPA, so I have (temporarily) disabled it.  Please test carefully and either correct the issue or remove the code. Single-cutoff r-RESPA passes.

mateors6:
Dear @akohlmey,

Thank you. With your optimizations, now it runs around 1% faster than before (with a little run, I'll see how it works with a longer one). 
I have updated the .cpp file, now both single-cutoff and multi-cutoff r-RESPA should be working. I'm going to check if I have updated the file correctly. 

Thanks for your help.

akohlmey:
> I have updated the .cpp file, now both single-cutoff and multi-cutoff r-RESPA should be working

Forces seem OK, but energy not.

mateors6:
> > I have updated the .cpp file, now both single-cutoff and multi-cutoff r-RESPA should be working
> 
> Forces seem OK, but energy not.

Yes, I am checking it. I will update once it's solved.

akohlmey:
@mateors6 here is another simple optimization (reduce pointer chasing when updating forces).

mateors6:
> @mateors6 here is another simple optimization (reduce pointer chasing when updating forces).

That's interesting! I didn't know it was better that way. 

If I have uploaded the .cpp file correctly, now the energy calculation with r-RESPA should be OK.

akohlmey:
> That's interesting! I didn't know it was better that way.

There are 3 simple changes that can be applied to most pair styles that can improve performance.
- cache outer loop forces in local variables (which I just applied)
- change from array of arrays to array of structs for positions and forces
- move if statements (on whether to compute and tally energy and stress) out of the inner loops by templating the compute function
You can see these in many pair styles of the OPT and OPENMP packages.

mateors6:
> > That's interesting! I didn't know it was better that way.
> 
> There are 3 simple changes that can be applied to most pair styles that can improve performance.
> 
> * cache outer loop forces in local variables (which I just applied)
> * change from array of arrays to array of structs for positions and forces
> * move if statements (on whether to compute and tally energy and stress) out of the inner loops by templating the compute function
>   You can see these in many pair styles of the OPT and OPENMP packages.

That's great, thank you. I'll take a look at it.

megmcca (approved):

sjplimp (changes requested):
I would like to look this over next week to see if I have comments

sjplimp:
Here is my first (perhaps only) comment - I don't think the name lj/improved is sufficiently descriptive.
"Improved" can mean a number of things in the context of an empirical potential.
If you look at all the other pair lj variants in LAMMPS, they all have more descriptive names than "improved".
And their authors probably all think their potential is "improved" in some respect for some use cases vs pure LJ.
So what would be a more descriptive name?  Possibly lj/pirani.  You also don't need to have "cut" in the name
unless you are planning a lj/something/long variant in the future.  Did they give this potential a name
in the original paper - it they simply called it "improved LJ", I'm still thinking "no" for LAMMPS.

mateors6:
> Here is my first (perhaps only) comment - I don't think the name lj/improved is sufficiently descriptive.
> "Improved" can mean a number of things in the context of an empirical potential.
> If you look at all the other pair lj variants in LAMMPS, they all have more descriptive names than "improved".
> And their authors probably all think their potential is "improved" in some respect for some use cases vs pure LJ.
> So what would be a more descriptive name?  Possibly lj/pirani.  You also don't need to have "cut" in the name
> unless you are planning a lj/something/long variant in the future.  Did they give this potential a name
> in the original paper - it they simply called it "improved LJ", I'm still thinking "no" for LAMMPS.

It has been named 'Improved Lennard-Jones' since it has more parameters than the LJ and the LJ can be recovered from this one. So, in the end, it's just a more complex approach. Also, it "eliminates most of the inadequacies at short- and long-range of the LJ model.", as stated in the original paper, where it is named 'ILJ'. 
Probably lj/pirani is a suitable name. We are not planning on adding new variants.
I'll ask my supervisors to see what they think about this topic.
Thanks for your comment.

PR #4559, opened 2025-04-20
Added atomic energy option for energy change calculation in fix sgcmc

athomps:
**Summary**

This pull request adds an atomic energy option for energy change calculation in fix sgcmc.  

**Related Issue(s)**

This provides a tradeoff between the general but very inefficient global energy change calculation and the efficient but potential-specific local energy change calculation.  Currently, all energy models in LAMMPS except the EAM pair style must use the global calculation.  The new method has been implemented for EAM, it gives the same answer as the existing local EAM method and is somewhat slower, but still much faster than the global method. Extending it to other energy 
models requires implementing `Pair::compute_atomic_energy()`, but no further modification to fix sgcmc code is required.

**Author(s)**

Aidan Thompson

**Licensing**

**Backward Compatibility**

No issues

**Implementation Notes**

Correctness was verified using examples/sgcmc/in.sgcmc.eam.  Identical results were obtained with and without `atomic/energy yes`, running on 1 or 8 MPI ranks.  Speed difference range from 2x-20x.  On 1 MPI rank, atomic energy method was 100x faster than generic method.  This speed-up increases as more MPI ranks are added (generic method is strictly serial).

**Post Submission Checklist**


tools/coding_standard/README
```diff
@@ -6,3 +6,8 @@ whitespace.py    detects TAB characters and trailing whitespace
 homepage.py      detects outdated LAMMPS homepage URLs (pointing to sandia.gov instead of lammps.org)
 errordocs.py     detects deprecated error docs in header files
 versiontags.py   detects .. versionadded:: and .. versionchanged:: with pending version date
+
```
akohlmey, line 9: Even simpler is going to the "src" folder and typing "make check", or one of the possible "make check-XXX" or "make fix-XXX" commands. E.g. "make fix-whitespace" to remove known whitespace issues.
Please see: https://docs.lammps.org/Build_development.html#coding-style-utilities and https://docs.lammps.org/Tools.html#coding-standard

src/pair.h
```diff
@@ -156,6 +158,7 @@ class Pair : protected Pointers {
   virtual void compute_inner() {}
   virtual void compute_middle() {}
   virtual void compute_outer(int, int) {}
+  virtual double compute_atomic_energy(int, NeighList *) { return 0.0; }
```
akohlmey, line 161: This function and the associated flag should be documented in https://docs.lammps.org/Modify_pair.html

src/MC/fix_sgcmc.cpp
```diff
@@ -989,6 +1023,70 @@ void FixSemiGrandCanonicalMC::flipAtomGeneric(int flipAtom, int oldSpecies, int
...
+  for(int jj = 0; jj < jnum; jj++) {
+    int j = jlist[jj];
+    Enew += force->pair->compute_atomic_energy(j, neighborList);
+  }
+
+  atom->type[flipAtom] = oldSpecies;
+
+  deltaE = Enew - Eold;
+
+  //  printf("In computeEnergyChangeEatom() deltaE = %g\n", deltaE);
```
akohlmey: Please remove debugging comments.

src/MC/fix_sgcmc.cpp
```diff
@@ -175,15 +177,15 @@ FixSemiGrandCanonicalMC::FixSemiGrandCanonicalMC(LAMMPS *_lmp, int narg, char **
         targetConcentration[i] = utils::numeric(FLERR, arg[iarg], false, lmp);
         targetConcentration[1] -= targetConcentration[i];
       }
-      for (int i = 1; i <= atom->ntypes; i++, iarg++) {
```
akohlmey, line 178: Is this a bugfix that would have to be backported to the stable version?

akohlmey (changes requested):
Please see my comments to individual changes.

src/MC/fix_sgcmc.cpp
athomps: Yes, I noticed that any keywords added after `variance ...` were being ignored.  This change fixed it.

src/MC/fix_sgcmc.cpp
athomps: I removed debugging comments from 3 other files in MANYBODY and MC too, added in a separate commit.

megmcca (approved):

athomps:
@akohlmey Let me know if there are still issues that I have not addressed.

akohlmey (approved):

akohlmey:
> @akohlmey Let me know if there are still issues that I have not addressed.

Nope. I was just waiting for an approval.

PR #4567, opened 2025-04-23
New fix set command

sjplimp:
**Summary**

Implement a new fix set command which allows set command operations to be invoked every N steps during a simulation.
This is only useful for atom properties which are set via an atom-style variable, which presumably changes the values during a run.

IMPORTANT note for developers: The fix set command itself is simple, but it relies on a refactored version of the set command to allow it be invoked either as a standalone command or via fix set.  The refactoring was straightforward, but tedious.  So possibly bugs were introduced in the set command.  It would be useful to run a full regression test on this PR before merging.

**Related Issue(s)**

N/A

**Author(s)**

Steve

**Licensing**

**Backward Compatibility**

Possible that bugs were introduced in set command - see Summary above

**Implementation Notes**

**Post Submission Checklist**


sjplimp:
@akohlmey @athomps - Aidan is going to do some testing with this draft PR on one of his models where it will be useful.  This PR still needs a doc page for the new fix set command

akohlmey:
> So possibly bugs were introduced in the set command. It would be useful to run a full regression test on this PR before merging.

@sjplimp I have "repaired" what I could and the PR now passes all integration, regression, and unit tests.

athomps:
@akohlmey Let me know if there are still issues that I have not addressed.

sjplimp:
@akohlmey I added the fix set doc page, and would like @athomps to review this - then it should be ready for you to look over and merge

jtclemm:
This looks quite useful but I had one immediate question. How does fix set handle modifying a property that is not typically forwarded to ghost atoms? E.g. charge or finite size radius (without using the dynamic setting)? Does calling fix set trigger a neighbor list build/exchange whenever it runs? 

Reminder from conversation: this could happen with regular *set* command if someone uses *run pre no*, should update note on *run* doc page to warn against this as well.

sjplimp:
@jtclemm Added a reneighbor option, also added info on run doc page.
@akohlmey Once @jtclemm or @athomps approves this PR, it should be ready for you to check/approve.

src/fix_set.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+FixSet::FixSet(LAMMPS *lmp, int narg, char **arg) : Fix(lmp, narg, arg)
+{
+  if (narg < 8) error->all(FLERR, 1, "Illegal fix set command: need at least eight arguments");
+
+  nevery = utils::inumeric(FLERR, arg[3], false, lmp);
+  if (nevery <= 0) error->all(FLERR, "Fix {} Nevery must be > 0", style);
+
+  reneighbor = utils::inumeric(FLERR, arg[4], false, lmp);
+  if (reneighbor = 0 || reneighbor > 1)
```
jtclemm: Accidental assignment to zero.

Should it also be `reneighbor != 1`? Not sure why it's `>1`.

src/set.cpp
```diff
@@ -42,739 +42,418 @@
...
+      process_shape(iarg,narg,arg,action);
+      invoke_choice[naction++] = &Set::invoke_shape;
+    } else if (strcmp(arg[iarg],"smd/contact/radius") == 0) {
+      action->keyword = SMD_CONTACT_RADIUS;
+      process_smd_contact_radius(iarg,narg,arg,action);
+      invoke_choice[naction++] = &Set::invoke_smd_contact_radius;
+    } else if (strcmp(arg[iarg],"smd/mass/density") == 0) {
+      action->keyword = SMD_MASS_DENSITY;
+      process_smd_mass_density(iarg,narg,arg,action);
+      invoke_choice[naction++] = &Set::invoke_density;
```
jtclemm: `invoke_smd_mass_density`?

src/fix_set.cpp
sjplimp: good catch - I meant it to be error if < 0 or > 1

src/set.cpp
sjplimp: yes, correct

sjplimp:
@jtclemm ready for re-review

jtclemm (approved):

akohlmey:
@athomps this is now waiting on you.

athomps (approved):
I approve. In my tests, the fix set command works as advertised.

PR #4574, opened 2025-04-28
New Fix for GJF Langevin Dynamics

talinke:
**Summary**

We provide a new standalone fix for a comprehensive formulation of the GJ methods. This fix provides a significant improvement to the current implementation found within the fix langevin, keyword gjf. In the new implementation, we have isolated noise and friction into two independent steps using the splitting formulation of the GJF method. This rewrite of the same algorithm allows a better integration with the time integration structure in LAMMPS, and it allows for exact restart capabilities. As a result, the new formulation solves several artifacts of the fix langevin keyword gjf implementation, including
-	Unnatural placing in input script of fix langevin gjf
-	Unnatural damping factor restriction
-	Removed need for NVE integration
-	Now allows for user to compare implementation directly to equations in publication

In addition, and most importantly, the new fix now contains several GJ methods, accessible through the keyword method. The user may also choose between using the half-step or on-site velocity through the keyword vel.

**Related Issue(s)**

PR #4517

**Author(s)**

Tim Linke (UC Davis)
Niels Grønbech-Jensen (UC Davis)

**Licensing**

**Backward Compatibility**

Prior usage of the GJF method was provided using the fix langevin with keyword gjf. This new fix paves the way for more sophisticated usage of any of the GJ methods (including GJF), and is the first of two steps in separating the GJ methods from the fix langevin framework. Consequently, fix langevin keywords such as tally, scale, zero or angmom are no longer supported in this new fix. Outside this new fix, all changes are fully backward compatible and reproduce statistically consistent results.

**Implementation Notes**

This fix is implemented by creating the files fix_langevin_gjf.cpp and fix_langevin_gjf.h. The documentation fix_langevin_gjf.rst is created, and the new fix is listed where relevant. The example/gjf files are updated to showcase the usage of the new keywords, and a comparison between different thermostats is provided in a new README. New log files are provided. The implementation was tested by comparing to the existing fix langevin implementation using various setups to ensure full functionality. No other features in LAMMPS are affected.

**Post Submission Checklist**

The new implementation closely follows equation 24 in Gronbech-Jensen, J. Stat. Phys. 191, 137 (2024). Per conversation with Aidan Thompson and Axel Kohlmeyer, a future PR will depreciate the keyword gjf in the fix langevin framework.


akohlmey:
@athomps @sjplimp What do you think about the choice of the fix name? To be consistent with other similar fixes like fix ffl, gle, or gld it may better to use the name fix gjf. Those all include time integration (like this one), whereas fix langevin/gjf may imply it is more similar to fix langevin and thus does not include time integration.

sjplimp:
I agree with @akohlmey that fix langevin/glf is not an ideal name, for the reasons he stated.  Also calling it langevin/glf implies that it is derived from fix langevin, which I don't think is correct (either code-wise or conceptually).  I also think the names gjf, ffl, gle, gld all suffer from being too short and non-obvious to users.  But I don't have a better suggestion than gj or gjf, unless the authors have some more descriptive standalone name for the integration algorithm than Niels name.  It would also pay to think ahead to future enhancements, as hinted at in the summary.  If Langevin effects and keywords will be added, would that be two fixes, fix gjf and fix gjf/langevin ?  Or you mention this fix as a gateway to other related algs - will those be added later as new fix gjf/suffix commands?  Or will all of those variants be implemented in this single new fix, with optional keywords?

Also, is the eventual goal to remove the current GJ options for fix langevin which I think you are saying will become inferior implementations?  If so, that is fine by me, but @athomps may want to weigh in.

sjplimp:
I just saw the "deprecate" comment at the end of the summary - all good.

talinke:
Thank you @akohlmey and @sjplimp for the comments. I agree that the new implementation is a significant departure from the old fix langevin algorithm. However, given that 'gjf' has been a longstanding option of the fix, we propose that the name langevin/gjf will make it clear to users that the base functionality they seek is the same as the old implementation.

Regarding future work: this fix now enables many more enhancements to be added later within the new fix, like new velocity definitions. Per conversation with @akohlmey, it is difficult to foresee the functionalities users may be interested in, so we added typical thermostat functions such as a temperature bias. The new formulation allows for future GJF enhancements to be integrated as keywords.

akohlmey:
> However, given that 'gjf' has been a longstanding option of the fix, we propose that the name langevin/gjf will make it clear to users that the base functionality they seek is the same as the old implementation.

I don't think that calling this fix langevin/gjf will achieve that for existing users and for the reason @sjplimp mentioned it would be misleading for new users.
An much better approach to educate existing users would IMO be to not remove the gjf keyword from fix langevin (only the code once would be removed after the new fix is fully tested) and then the error message should inform users that this functionality is now available in fix gjf. Same in the documentation and we also have this page: https://docs.lammps.org/Commands_removed.html

talinke:
That is a great idea - thanks for the additional insights. I've updated the name according to your and @sjplimp's suggestion.

akohlmey:
@talinke thanks for making the changes. It is much appreciated.
I've added the info about the pending changes in fix langevin to the manual and also added a deprecation warning to fix langevin itself. This looks good from the technical point of view.

@athomps can you please do the final review on the science for this contribution?
I propose we leave fix langevin unchanged except for the added warning and then remove
the GJF code from fix langevin after the stable release this summer.

talinke:
@akohlmey thank you for adding the depreciation warning. For completeness, I've also updated a citation mismatch in the standard fix langevin: the implemented algorithm follows the BBK method rather than Schneider & Stoll (Niels had noticed this).

megmcca (approved):

PR #4581, opened 2025-05-01
Enable equal- and atom-style variables to invoke Python functions more directly with args

sjplimp:
**Summary**

Add support for the formulas in equal- and atom-style variables to include a Python function wrapper and pass the Python function arguments from the formula.  The wrapper syntax includes the name of the python-style variable which links the variable to a Python-coded function, using the existing python command capabilities.  The wrapper makes it each to call the Python function directly with LAMMPS-based arguments.  And to invoke the function once-per-atom in the case of an atom-style variable, passing it per-atom values as arguments.

The syntax of the python command adds an `"input" option to use iv_name in addition to v_name.  IV stands for internal var.  This means the Python function will directly access internal vars for numeric values which are set by the Python function wrapper in the Variable class.  Likewise, an option was added to the Python function invocation to request it return a numeric value directly to the Variable class (w/out conversion to a string and back to a numeric value), when invoked via an equal-style or atom-style variable.

Also made the creation of internal-style variables, by various commands which require them, automatic, so that lines like this:
variable xx internal 0.0
are not needed in the input script.  The one exception is the fix controller command, where other (potentially arbitrary) commands need to know the name of the internal-style variable before fix controller creates it.

The performance for atom-style variables is better than expected.  Less than a second of CPU time for 100Ks of simple function invocations.

Added an examples/python/in.python.wrap script which illustrates the use of a Python function wrapper in both an equal-style and atom-style variable.

**Related Issue(s)**

N/A

**Author(s)**

Steve

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**


sjplimp:
@akohlmey The spell checker flagged the variable.rst table with an error - I can't figure out the issue.  Also not clear if you want me to add a couple of the new pseudo words for spell check skip, or you do that.

akohlmey:
> @akohlmey The spell checker flagged the variable.rst table with an error - I can't figure out the issue.

There is a "Malformed table" error, which is independent of the spell checker.
That usually means, that the ASCII-art table lines do not match up.

>  Also not clear if you want me to add a couple of the new pseudo words for spell check skip, or you do that.

I *have* to do it, if you don't do it. Of course, it is much less of a hassle if you don't
leave all the things you break for me to clean up. ;-)

But then again, if I have the branch already checked out and and see what the issues are, it is
easier to just fix it than to explain how to fix it. So I'm cleaning it up (again).

doc/src/variable.rst
```diff
@@ -1161,6 +1180,79 @@ variable name.
...
+floating point argument, and returns its truncated integer value.  In
+this case, the Python code for truncate() is included in the *python*
+command; it could also be contained in a file.  See the :doc:`python
+<python>` command doc page for details.
+
+The *variable pyarg1* command defines an internal-style variable.  It
+MUST have the name pyarg1.  If the Python function has *N* arguments,
+*N* internal-style variables MUST be defined with names *pyarg1*,
+*pyarg2*, ... *pyargN*.  Note that multiple Python function wrappers
+can use the same internal-style variables.
```
rbberger: This is weird and unintuitive. We should not have to expose a implementation detail like this.

sjplimp: What do you suggest instead? The code logic is that when the variable formula which includes a Python wrapper function is invoked, the formula evaluates the sub-formulas for each of the N args to the Py function. It needs to put the N results somewhere before triggering invocation of the Py function.  Internal variables are an obvious choice since the Py function evaluation already allows for that and the args are always numeric. How about the variable class queries the Python class where the function is registered to get the names of the internal vars used in the "python" command for that function?  Then they could be any internal var names the user wished.

rbberger: Can you register internal variables on the fly when needed?

sjplimp: Not sure if by "register" you mean create or set the value.  If the python command specifies the name(s) of them,
then the variable formula eval can set the values (many times), so long as it can query the names from the Python class.

This is similar to what the create_atoms command with its var/set option does.  See its doc page in the section where the
sinusoidal image of atoms is - it shows the input script commands that do that.

If what you are wanting is to create internal variables behind the scenes based on names specified in the input script, I think that could work -- I think I like that idea.

But things like the following would need to be disallowed:  create_atoms var uses a variable which uses a python wrapper function and they both use the same internal variable.  Then I think the values could step on each other.
So if doing auto-creation, it would sometimes be OK if the variable already existed, other times not.  I'd have to think about that some more.  Probably it would always be OK if the auto-created internal variables did not already exist.

What I would not want to do is create/destroy internal vars each time the formula for the Python function wrapper was triggered.

You can see all the current uses of internal vars in LAMMPS if you do this: grep internal-style doc/src/*rst

rbberger: What I mean is creation behind the scenes during first use with a automatic internal name.

sjplimp: I'm assuming by "automatic" you mean LAMMPS chooses a name for the internal-style variable,
which is hidden from the user.

This could work for the use case of this PR, b/c the python command could create one or more of them,
and the variable command doesn't need to know their names to use a Python function wrapper. It could query the Python class for the index of the internal-style var for each function arg.  I.e. the formula with the Py func wrapper does not specify the names of the internal-style vars.

But that's not the way other uses of internal-style vars operate.
The create_atoms var/set use case requires the name of the internal-style var be known by the input script, b/c
the user has to specify an equal-style var formula which accesses one or more internal-style var names.
Ditto for fix deposit, compute bond/local, compute angle/local, etc

Also ditto for fix controller, and there it is some other "unknown" LAMMPS command which uses the internal-style var.
See input scripts in examples/controller as examples.

Also ditto for the lib interface lammps_set_internal_varlable, which requires the external caller to know the name of the var.

So I would say we could get rid of the need for the input script to have lines like:
variable xx internal 0.0
by having (in the case of this PR), the python command create N internal-style vars with names the user specifies in the python command.
This would also work for all other use cases listed above, except fix controller (see the example scripts).
And the lib interface, b/c it is supposed to allow an external code to be the "driver" of the internal var, similar to fix controller.

sjplimp:
@rbberger Here is what I changed in the PythonImpl class to enable what I suggested above -- see what you think ?

(a) Added iv_name as an "inputs" option to trigger create_entry() to create the internal variable if needed.  It also trigger invoke_function() to use pre-computed values from internal variables as arguments.
(b) Added an extra arg to invoke_function() to allow it to return a numeric value directly, rather than as a string.  When called from Variable methods for equal- and atom-style formula evaluations, this saves converting and reconverting the numeric value.
When called by Variable::retrieve() nothing is changed, even if return is numeric.
(c) Added a sister method wrapper_match(), similar to old variable_match() (now called function_match()), to check consistency of a variable Python wrapper function with the python command's definition.  I added a NOTE to both suggesting we print the error messages there, rather than duplicate the messages 3x in variable.cpp
(d) Similar to (a), I changed various other commands which use internal vars to auto-create them if they do not exist, so the user of  commands like:
variable xx implicit 0.0
is now no longer needed. (One exception, see summary at top).

There is a also a NOTE in invoke_function() about how to handle use of the "invoke" keyword in the python command, wrt a return value, which I think may be confusing to users, since I believe the return value is inaccessible.

sjplimp:
@rbberger I have updated the doc pages for this new version.  Also fixed a bug I found in testing.
From my perspective this is now ready-to-go - pending your feedback or edits.  Also after addressing the NOTE
comments to you in the code and python.rst doc page.

I also tested 2 funky use cases with recursion - they seemed to work fine:
a) eval of a python-style var which used an arg which was another python-style var
b) variable formula with nested Python function wrappers like this: variable xnew atom py_foo(py_foo2(x))
And the Variable class error checks if you accidentally trigger an infinite recursion.

ndtrung81 (approved):

akohlmey:
@rbberger please assign this pull request to me when you have no more requests for changes or questions and I can have a final look at technical details and formal issues.

src/PYTHON/python_impl.cpp
```diff
@@ -416,10 +440,22 @@ int PythonImpl::find(const char *name)
...
+   ensure the Python function produces an output
+   ensure the Python function outputs to the matching python-style variable
+   ensure a string is returned only if retrieve() is the caller
+--------------------------------------------------------------------- */
+
+int PythonImpl::function_match(const char *name, const char *varname, int numeric)
 {
+  // NOTE to Richard - any reason not to put error messages here instead of in Variable class ?
+  //   error messages appear 3x in Variable
+
```
rbberger: @sjplimp ~The main reason is to not have PythonImpl be a Pointers subclass. We can avoid that by passing `error` or raising an exception instead.~ Nevermind, that's not even true. So no reason.

akohlmey: moved error handling to PythonImpl class and out of Variable class.

akohlmey:
@sjplimp I resolved the remaining NOTEs.

PR #4582, opened 2025-05-01
Fix torch.load in MLIAP for PyTorch >= 2.6 

fglines-nv:
**Summary**

PyTorch recently made a change in 2.6 to by default only load weights from pickles files and no objects for security reasons. (See release notes: https://pytorch.org/blog/pytorch2-6/)

Since the `torch.load` call in the MLIAP pairstyle always loads a `MLIAPUnified` object and not just weights, we need to pass the kwarg `weights_only=false` to `torch.load`. 

**Related Issue(s)**

**Author(s)**

Forrest Glines [fglines@nvidia.com](mailto:fglines@nvidia.com)
**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/src/KOKKOS/mliap_unified_couple_kokkos.pyx
+++ b/src/KOKKOS/mliap_unified_couple_kokkos.pyx
@@ -498,7 +498,7 @@ cdef public object mliap_unified_connect_kokkos(char *fname, MLIAPDummyModel * m
         unified = LOADED_MODEL
     elif str_fname.endswith(".pt") or str_fname.endswith('.pth'):
         import torch
-        unified = torch.load(str_fname)
+        unified = torch.load(str_fname,weights_only=False)
     else:
         with open(str_fname, 'rb') as pfile:
             unified = pickle.load(pfile)
--- a/src/ML-IAP/mliap_unified_couple.pyx
+++ b/src/ML-IAP/mliap_unified_couple.pyx
@@ -363,7 +363,7 @@ cdef public object mliap_unified_connect(char *fname, MLIAPDummyModel * model,
         unified = LOADED_MODEL
     elif str_fname.endswith(".pt") or str_fname.endswith('.pth'):
         import torch
-        unified = torch.load(str_fname)
+        unified = torch.load(str_fname,weights_only=False)
     else:
         with open(str_fname, 'rb') as pfile:
             unified = pickle.load(pfile)


fglines-nv:
For the sake of SEO, the error without this PR when using PyTorch >=2.6 looks like this:
```
Traceback (most recent call last):
  File "mliap_unified_couple_kokkos.pyx", line 501, in mliap_unified_couple_kokkos.mliap_unified_connect_kokkos
  File "/usr/local/lib/python3.12/dist-packages/torch/serialization.py", line 1455, in load
    raise pickle.UnpicklingError(_get_wo_message(str(e))) from None
_pickle.UnpicklingError: Weights only load failed. This file can still be loaded, to do so you have two options, do those steps only if you trust the source of the checkpoint.
        (1) Re-running `torch.load` with `weights_only` set to `False` will likely succeed, but it can result in arbitrary code execution. Do it only if you got the file from a trusted source.
        (2) Alternatively, to load with `weights_only=True` please check the recommended steps in the following error message.
        WeightsUnpickler error: Unsupported global: GLOBAL hippynn.interfaces.lammps_interface.mliap_interface.MLIAPInterface was not an allowed global by default. Please use `torch.serialization.add_safe_globals([MLIAPInterface])` or the `torch.serialization.safe_globals([MLIAPInterface])` context manager to allowlist this global if you trust this class/function.

Check the documentation of torch.load to learn more about types accepted by default with weights_only https://pytorch.org/docs/stable/generated/torch.load.html.
ERROR: Running mliappy unified module failure. (src/KOKKOS/mliap_unified_kokkos.cpp:248)
```

akohlmey:
@fglines-nv can you please state whether this change is backward compatible to older versions of torch? And if yes, how far.

fglines-nv:
> @fglines-nv can you please state whether this change is backward compatible to older versions of torch? And if yes, how far.

I had to look up when `weights_only` was added as a parameter. PyTorch 1.10 (from October 2021), so we'd be backwards compatible back to that version.

rbberger:
Just for completeness. This and a workaround via an environment variable was also mentioned in #4526. But I agree that changing this directly in code is preferable.

stanmoore1 (approved):
I approve.

stanmoore1:
Seems like this PR should close issue https://github.com/lammps/lammps/issues/4526

PR #4584, opened 2025-05-02
Update fix qtpie/reaxff and fix qeq/rel/reaxff docs

navlalli:
**Summary**

Add citation to recently published work that details the implementation of `fix qtpie/reaxff` and `fix qeq/rel/reaxff`.

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

PR #4587, opened 2025-05-05

### LINKED ISSUE #4526: [BUG] ML-IAP broken with Kokkos when using Python

### LINKED ISSUE #4583: [BUG] ComputeRHEOInterface::store_forces() cannot compute the non-pressure forces correctly in some cases.
**Summary**

Non-pressure forces are calculated in `ComputeRHEOInterface::store_forces()` for density reconstruction. If I understand it correctly, `store_forces()` uses `fix setforce` to identify the solid particles, because we often set the forces of stationary solid particles to zero by `fix 1 wall setforce 0 0 0`. 

Image

This is OK in most cases with only one `fix setforce`, but the non-pressure forces of fluid will be incorrect in cases of two or more `fix setforce`. For example, the walls on the left and right are of different types:

```
create_atoms       1 region fluid
create_atoms       2 region walls_left
create_atoms       3 region walls_right

group              fluid type 1
group              rig_left type 2
group              rig_right type 3

fix                5_left rig_left setforce 0.0 0.0 0.0
fix                5_right rig_right setforce 0.0 0.0 0.0
```

Since there are two `fix setforce`, `fixlist.size()` will be two, and Line 356 will be executed **twice** for a fluid particle. For this fluid particle with atomID=i, before `store_forces()` is called, `fp_store[i]` is the pressure force (with a force unit) stored in Line 354-364 of `PairRHEO::compute(int, int)`.
When it is executed for the first time (`fix_5_left`), `fp_store[i][a] = (f[i][a] - fp_store[i][a]) * minv;` will assign the non-pressure force to `fp_store[i]` with an acceleration unit (force/mass). However, when it is executed for the second time (`fix_5_right`), `fp_store[i]` will become unphysical because `f[i]` is a force, but `fp_store[i]` is now an acceleration (force/mass), so they cannot be subtracted!

A remedy is `fix setforce` only once by the union operation:

```
group    rig_all union rig_left rig_right
fix      5 rig_all setforce 0.0 0.0 0.0
```

But I suggest that the source code be modified for general cases. If `modify->get_fix_by_style("setforce")` is really to check the status, I wonder why not use more intuitive `(status[j] & PHASECHECK)`. What if a moving boundary is used rather than a stationary boundary so that there is no `fix setforce`?

Latest version: 2Apr2025

### LINKED ISSUE #4586: [Feature Request] Expose bond, angle, dihedral and improper neighbor lists through the C-API
**Summary**

It already is possible to write custom pair styles using the C-API through fix external. However, defining for example custom Bond styles would require access to the bond neigbor list, which currently isn't possible.

**Detailed Description**

I am aware that `lammps_gather_bonds` etc. exists. However using this isn't feasable when running lammps on a cluster with a high number of MPI ranks. From my understanding of the code, this would simply require exposing the `num_bond`, `bond_type`, `bond_atom`, etc. pointers from the `Atom` class through the C-API.

### LINKED ISSUE #4589: [BUG] A bug of FixRHEOPressure::pre_force(int)
**Summary**

Image
`FixRHEOPressure::calc_pressure` has two arguments, one of which is `int i`. It can be easily deduced from `int type = atom->type[i];` that `i` is a local ID of an atom. However, in `FixRHEOPressure::pre_force(int)`, the type-ID instead of atom-ID is passed to `FixRHEOPressure::calc_pressure`:

Image

That is to say, `type[i]` in `FixRHEOPressure::pre_force(int)` should be changed into `i`.

2Apr2025

akohlmey:
**Summary**

**Related Issue(s)**

Implements and thus closes #4586 
Closes #4526
Fixes #4589
Fixes #4583

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- provide access to bondlist, anglelist, dihedrallist, and improperlist created by the Neighbor class through the library interface
- provide access to bond topology related per-atom data through the library interface
- avoid nvcc warnings for all Clang/LLVM derived compilers not only Clang
- small updates to LAMMPS-GUI docs
- include PR #4582
- include PR #4584 
- small optmization of GRANULAR package by replacing pow() with faster alternatives
- fixes a typo in recently added changes to GRANULAR package
- fixes two issues in RHEO package from @jtclemm 

**Post Submission Checklist**


jtclemm:
Hey @akohlmey, would it be easy to incorporate these two commits: https://github.com/lammps/lammps/commit/7c30111fa06e88bad0381c748d169ce4f97c497c and https://github.com/lammps/lammps/commit/dc07a1471ec938aaa62f452942fd46c3adf76a3b ? They are just quick fixes for little bugs in the RHEO package, closing issues #4589 and #4583.

Otherwise, I can just create a separate small PR,.

akohlmey:
> would it be easy to incorporate these two commits: [7c30111](https://github.com/lammps/lammps/commit/7c30111fa06e88bad0381c748d169ce4f97c497c) and [dc07a14](https://github.com/lammps/lammps/commit/dc07a1471ec938aaa62f452942fd46c3adf76a3b) ?

Yes very easy.

src/library.cpp
```diff
@@ -1255,6 +1255,7 @@ be called without a valid LAMMPS object handle (it is ignored).
 * :ref:`Image masks <extract_image_masks>`
 * :ref:`System status <extract_system_status>`
 * :ref:`System sizes <extract_system_sizes>`
+* :ref:`Neigbor list settings <extract_neighbor_settings>`
 * :ref:`Atom style flags <extract_atom_flags>`
```
jtclemm, line 1259: typo, neigbor

src/library.cpp
```diff
@@ -1395,6 +1406,31 @@ internally by the :doc:`Fortran interface <Fortran>` and are not likely to be us
    * - nbodies
      - number of atoms that have body data (see :doc:`the Body particle HowTo <Howto_body>`)
 
+.. _extract_neighbor_settings:
+
+**Neigbor list settings**
```
jtclemm: ditto

jtclemm (approved):

PR #4590, opened 2025-05-09
Make fix SRD compatible with PBCs & fix deform remap v, and add profile-unbiased thermostat

### LINKED ISSUE #4602: [BUG] Velocity profiles with fix srd and fix deform shear flow
**Summary**

When combining `fix srd tstat yes` and `fix deform`, the velocity profile has a clear kink near the periodic boundary in gradient direction. Without thermostat, the fix fails to maintain shear flow.

LAMMPS (7 Feb 2024), Debian 12

**Expected Behavior**

We expect a linear velocity profile with and without thermostat.

**Actual Behavior**

`fix deform remap v` should lead to linear velocity profiles, also combined with `fix srd`. (1) However, with `tstat yes`, there are clear kinks, deviations from a line, in the velocity profile near the periodic boundary in gradient direction. Below, I provide a simple input deck to reproduce this issue. (2) Also, when turning the `fix srd` thermostat off, the shear profile is not maintained. When we equilibrate with `tstat yes` and then turn to `tstat no`, the shear profile quickly vanishes starting from the periodic boundary, leading to a constant velocity throughout the box. This is in contrast to the documentation which states "if no thermostatting is performed (the default), then it may take a long time for the SRD fluid to come to equilibrium with a velocity profile that matches the simulation box deformation".

**Steps to Reproduce**

See the minimal input deck: shear a simple SRD fluid with `fix deform` and measure the velocity profile with and without thermostat.

A possible explanation: SRD propagates particles ballistically in the streaming step. In regular intervals, it sorts them into the compartments of a grid and performs effective collisions in each cell, coupling the momenta. The grid is randomly shifted every time a collision step is executed. This leads to bins going through the periodic boundaries. If I understand correctly, communication takes care of that. However, I suspect that velocities are not adapted when a bin goes through the periodic boundary in gradient direction.

I tried to solve this in PR #4590 . Given my limited understanding of LAMMPS' communication, I am unsure whether my approach was valid. In any case, these bugs are in the current version of LAMMPS.

The kinks in the velocity profile:

Vanishing profile when turning off the thermostat:

Input deck and Python script plotting the velocity profile:
[in.txt](https://github.com/user-attachments/files/20459896/in.txt)
[profile.py.txt](https://github.com/user-attachments/files/20459953/profile.py.txt)

taenzel:
**Summary**

This PR addresses issues of `fix srd` related to the fact that SRD collision cells / bins ignore velocity jumps at the PBCs from `fix deform`. It also introduces a profile-unbiased thermostat for SRD simulations under shear with `fix deform`, controlled via a new keyword `unbiased`.

**Related Issue(s)**

Fixes #4602

**Author(s)**

Victor Taenzel, University of Freiburg

**Licensing**

**Backward Compatibility**

The new profile-unbiased thermostat is disabled by default, preserving existing behavior and compatibility with current input scripts.

**Implementation Notes**

1. Periodic Boundaries in SRD Collision Cells
  In a box with $N$ collision cells, LAMMPS sets up $N+1$ bins. This happens in the context of the grid shifts of the SRD method. Communication then takes care of linking the first and last bins. However, under shear flow, the velocity of particles in the fastest and slowest bin do not match.
    - This suppresses a linear shear profile with `tstat no`. This is due to the following: first, it couples the fastest and the slowest particles without incorporating the velocity shift. Second, in gradient direction, momentum is currently only transported through the Lees-Edwards periodic boundary during the streaming step, but not during the typically frequent collision steps. These then only transport momentum "within" the box, leading to an averaging effect of the velocity, as there is no source for momentum, which would otherwise be the Lees-Edwards boundaries. Even when we set up a velocity profile with `velocity ramp`, this effect quickly leads to a constant velocity throughout the box. Thus, the current promise of the documentation that "for streaming simulations, if no thermostatting is performed (the default), then it may take a long time for the SRD fluid to come to equilibrium with a velocity profile that matches the simulation box deformation" does not hold true, as it will never happen. 
    - The current profile-biased thermostat leads to distinct kinks of the velocity profile near the periodic boundaries (see the picture below). 

To fix these issues, the velocities are temporarily shifted, which is reversed after the rotation operation and thermostat of the collision step. This only affects the "fastest" bin, as bin shifts are unidirectional. Tests confirmed energy conservation and a velocity profile without strong kinks under shear.

2. Profile-Unbiased Thermostat (PUT):
The correct treatment of periodic boundaries allowed the implementation of a profile-unbiased thermostat.
    - With `unbiased=no` (default): $v_\mathrm{new} = v_\mathrm{stream} + \zeta (v-v_\mathrm{com})$
    - With `unbiased=yes` (new): $v_\mathrm{new} = v_\mathrm{com} + \zeta (v-v_\mathrm{com})$

  The current thermostat, accessible with `unbiased=no`, risks suppressing interesting flow patterns around suspended particles, and is generally not used by the SRD/MPCD community. The implementation was verified to maintain linear velocity profiles and create them from quiescent initial conditions. In the latter case, it generates a velocity profile from $-v_\mathrm{max}/2$ to  $v_\mathrm{max}/2$ instead of from $0$ to $v_\mathrm{max}$. If this is problematic was also asked [here](https://matsci.org/t/velocity-profile-under-deformation/63177).

Please note that `rescale yes` will not conserve momentum and energy, which is problematic when using `unbiased yes`.

**Post Submission Checklist**

Image 1: The strong kink in the velocity profile with default profile-biased thermostat, is solved by following Lees-Edwards boundaries during cell assignment.

Image 2: Velocity profile under shear flow, starting from a quiescent state. Averages are conducted over 500 and 5000 collision steps.

Example input file for pure SRD fluid under shear with PBT and PUT.
[in.srd.txt](https://github.com/user-attachments/files/20119998/in.srd.txt)


jtclemm:
I'm not a heavy user of the package, but from my familiarity with it, these changes all look correct to me. 

Only one comment, I didn't find *put* to be a very intuitive argument name. It seems like a pretty niche acronym. What about *unbiased* or *profile/unbiased* or something more explicit?

taenzel:
Thank you for the feedback! A more explicit name would make a lot of sense.

In the meantime, I have learned that my code changes need to be adapted for multiprocessing. If everything goes well, I will update the code asap.

EDIT: I updated the description of the PR.

sjplimp:
@taenzel Thanks for working on these issues with fix srd in conjunction with fix deform.
Are you saying your most recent changes fix the issues in parallel as well, not just in serial?
E.g. the 2 plots you included above would be reproduced when running on an arbitrary # of procs?

taenzel:
Yes @sjplimp, the provided plots can be reproduced in a parallelized run. I have mostly tried 8 processes. The pure SRD suspensions seem to work fine now.

However, I have now tested embedding spherical particles in the SRD suspension. Apparently, there is another bug in the code, unrelated to my changes, that prevents me from properly testing my modifications. When switching to the new unbiased thermostat, a jump in the center of mass velocity occurs. The same happens in the unmodified, standard LAMMPS when turning off the thermostat. I reported this in issue #4611 and do not know how to solve it.

sjplimp:
Thanks @taenzel for addressing the Qs of myself and @jtclemm.  I will approve this PR for merging.

sjplimp (approved):

PR #4594, opened 2025-05-13
Collected small changes

### LINKED ISSUE #4592: [Feature Request] expose `eflag` and `vflag` to C-API for fix external
**Summary**

Currently its not possible to know wheter the virial or energy contributions of a fix external are required for the current timestep, as `eflag` and `vflag` are not exposed through the C-API. Exposing these flags would make it possible to only calculate the energies and virials in the callback in fix external for the timesteps where they are actually needed.

**Detailed Description**

One possible way to do this would be to make `vflag` and `eflag` available through `lammps_extract_setting` or `lammps_extract_global`, if these flags are indeed a global value (which I'm not sure is the case).

Another approach would be to extend the api with functions that may look like this:
```C
double *lammps_fix_external_get_virial_global(void *handle, const char *id, int *vflag_global)
double **lammps_fix_external_get_virial_atom(void *handle, const char *id, int *vflag_atom)
double *lammps_fix_external_get_energy_global(void *handle, const char *id, int *eflag_global)
double *lammps_fix_external_get_energy_atom(void *handle, const char *id, int *eflag_atom)
```
These would then return pointers to `user_virial`, `vatom`, `user_energy`, and `eatom`, respectively, and also set the corresponding flag. This would also eliminate a few unecessary copy operations.

akohlmey:
**Summary**

This pull request combines multiple small changes and fixes that don't justify a pull request on their own.

**Related Issue(s)**

Closes #4592
Closes #4595 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

The sign of *dx*, *dy*, and *dz* in `compute bond/local` and `compute pair/local` is no longer determined by their atom IDs but by their order in the bond-list or neighbor-list.
This is for consistency with *fx*, *fy*, *fz*. The respective atom IDs can by obtained
with `compute property/local`.

**Implementation Notes**

- make `eflag_global`, `eflag_atom`, `vflag_global`, and `vflag_atom` of the `Update` class accessible through the C-library interface
- enable the PotentialFileReader class to adjust the line buffer size of the internal TextFileReader class instance
- documentation updates
- add section about OPC water model (used by the latest Amber force field revision) to TIP4P howto
- add generic section about conventional force field to BioFF Howto section
- update outdated fix imd examples with GPU acceleration (after being unchanged for 15 years)
- delete a couple assignments to self in KOKKOS
- avoid type clash with constant string argument for some MPI implementations in LATBOLTZ package
- correct check for 32-bit integer overflow in `compute podd/atom`
- includes PR #4595
- remove directionCorrection from `compute bond/local` and `compute pair/local` according to discussion below

**Post Submission Checklist**


akohlmey:
@sjplimp do you think that the patch suggested in https://matsci.org/t/outputs-from-pair-local-command/63507/2 to get a force sign consistent with the distances makes sense to include?

sjplimp:
@akohlmey I see the inconsistency mentioned on MatSci.  But I think it would be better to change dx,dy,dz to be wrt atom I (i,e. distance of atom J from atom I), just like fx,fy,fz are the force on atom I due to atom J.  I don't recall what the motivation was to have dx,dy,dz be the distance from the higher ID to the lower ID.  I.e. remove the "directionCorrection" logic from the current version.  The user can also get the IDs of the I and J atoms in the dump file via compute property/local

akohlmey:
> @akohlmey I see the inconsistency mentioned on MatSci. But I think it would be better to change dx,dy,dz to be wrt atom I (i,e. distance of atom J from atom I), just like fx,fy,fz are the force on atom I due to atom J. I don't recall what the motivation was to have dx,dy,dz be the distance from the higher ID to the lower ID. I.e. remove the "directionCorrection" logic from the current version. The user can also get the IDs of the I and J atoms in the dump file via compute property/local

@sjplimp It looks like you approved it here: https://github.com/lammps/lammps/pull/3032

I guess we need to ask @GenieTim for clarification.

sjplimp:
A better way to say it (e.g. in the doc page) would be that (dx,dy,dz) is the vector displacement = Xi - Xj
force is the magnitude of the force with a sign = positive for repulsion, negative for attraction
(fx,fy,fz) is force*(dx,dy,dz) or effectively the force of atom J on atom I
compute pair/local is coded the same way as all the pair styles (e.g. pair lj/cut), if the directionCorrection is removed

sjplimp:
Looked at the older PR - @GenieTim added (dx,dy,dz) to be like compute bond/local.  But bonds are ID of I < ID of J by construction, not like pair interactions, which come from neigh lists.  So I think making dx,dy,dz and fx,fy,fz be consistent with pair styles is the way to do it.  Also compute pair/local can possibly have the IDs of I,J be the same. which directionCorrection cannot account for consistently.  E.g. if you wanted to compute a virial from the sum of compute pair/local outputs.

sjplimp:
PR #3032 also added dx,dy,dz to compute bond/local and the doc page sentences about lower ID and higher ID to both
pair/local and bond/local. I think that doc page info as well as directionCorrection should be removed from both commands.
Dx,dy,dz and fx,fy,fz will be consistent in both local commands if they are just done the way pair and bond styles do the calculations.  I think the logic to skip duplicate pairs or bonds in both local styles when newton is off, is what confused the issue.

akohlmey:
@sjplimp I've tried to implement your suggestions. Please have another look and suggest corrections where needed.

sjplimp (changes requested):
Code looks perfect - just one comment on the 2 doc pages for pair/local and bond/local

doc/src/compute_bond_local.rst
```diff
@@ -64,22 +64,32 @@ All these properties are computed for the pair of atoms in a bond,
...
+The value *force* is the magnitude of the force acting between the pair
+of atoms in the bond, which is positive for a repulsive force and
+negative for an attractive force.
 
-The values *fx*, *fy*, and *fz* are the xyz components of
-*force* between the pair of atoms in the bond. For bond styles that apply
-non-central forces, such as :doc:`bond_style bpm/rotational
-<bond_bpm_rotational>`, these values only include the :math:`(x,y,z)`
-components of the normal force component.
+The values *fx*, *fy*, and *fz* are the :math:`(x,y,z)` components of
```
sjplimp, line 87: Here is what I suggest, and similar for pair local doc page:

The values *fx*, *fy*, and *fz* are the :math:`(x,y,z)` components of
the force on the first atom I in the bond due to the second atom J.  Mathematically, this is obtained by multiplying the components dx, dy, dz of the distance vector described above by *force*.  For bond styles that apply non-central forces ...

sjplimp, line 87: Here and in the pair local doc, I would say it this way:
The values *fx*, *fy*, and *fz* are the :math:`(x,y,z)` components of the force on the first atom I in the bond due to the second atom J in the bond.  Mathematically, they are obtained by multiplying the dx, dy, dz components of the distance vector described above by *force*.  For bond styles that apply non-central ...

sjplimp: sorry, it seemed to require my comment twice ...

akohlmey: > they are obtained by multiplying the dx, dy, dz components of the distance vector described above by _force_.

That would not be correct. The value of force is the internally computed force multiplied by *r*.
One could say, that it is a vector obtained from multiplying with a unit vector created from dx, dy, dz.

sjplimp: yes - good catch - how about "obtained by converting the (dx,dy,dz) distance vector (above) to a unit vector and multiplying by the *force* quantity (above)"

sjplimp (approved):
approved after making the one change you caught

doc/src/compute_bond_local.rst
akohlmey: I had already made a change that says the same.

PR #4597, opened 2025-05-15
Small CMake  updates

akohlmey:
**Summary**

This adds some small updates to CMake scripting for improved consistency.

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

This updates the project() command to include some more (optional) information and always enable C as programming language.

~~Also we are checking for and using only the C-language interface to MPI, so we should use the imported target `MPI::MPI_C` and not `MPI::MPI_CXX`. This avoids a bogus dependency on libmpicxx and avoid compilation failures on MPI libraries that have dropped the deprecated C++ MPI class.~~

For recent GCC or Clang compilers the flag `--fuse-ld` needs to be augmented with `--ld-path`. We detect if the latter is supported and add it, if yes.

For cross-compiling to Windows we remove support for the obsolete MPICH-1.4.1 binaries and now support 64-bit Windows and MS-MPI only.

Update the version string to to be `<year>.<month>.<day>.<tweak>` with `<tweak>` being 0 for regular releases, 1-n for update releases, and 99 for development or maintenance.

**Post Submission Checklist**


cmake/CMakeLists.txt
```diff
@@ -394,7 +400,7 @@ if(PKG_ADIOS)
 endif()
 
 if(NOT CMAKE_CROSSCOMPILING)
-  find_package(MPI QUIET)
+  find_package(MPI QUIET COMPONENTS C)
```
rbberger: This is wrong.

cmake/Modules/LAMMPSInterfacePlugin.cmake
```diff
@@ -163,12 +164,12 @@ if(BUILD_MPI)
 
       ExternalProject_get_property(mpi4win_build SOURCE_DIR)
       file(MAKE_DIRECTORY "${SOURCE_DIR}/include")
-      add_library(MPI::MPI_CXX UNKNOWN IMPORTED)
```
rbberger, line 166: We're a C++ code. The correct way to use the **C bindings** of MPI with C++ is to use the MPI_CXX target. I've spent a significant amount of time correcting this in other places. Don't go backwards.

rbberger (changes requested):

cmake/Modules/LAMMPSInterfacePlugin.cmake
rbberger: If there is still any relevant MPI-2 implementation out there (I'm guessing Windows), use `MPI_CXX_SKIP_MPICXX` prior to `find_package(MPI)` to set the right defines.

rbberger: FYI, the MPI-2 C++ interface is what you get with

```cmake
find_package(MPI COMPONENTS MPICXX)
target_link_libraries(... MPI::MPI_MPICXX)
```
unlike the regular C bindings via
```cmake
find_package(MPI COMPONENTS CXX)
target_link_libraries(... MPI::MPI_CXX)
```

cmake/CMakeLists.txt
```diff
@@ -711,6 +717,7 @@ foreach(PKG_LIB POEMS ATC AWPMD H5MD)
     file(GLOB_RECURSE ${PKG_LIB}_SOURCES CONFIGURE_DEPENDS
       ${LAMMPS_LIB_SOURCE_DIR}/${PKG_LIB}/[^.]*.c ${LAMMPS_LIB_SOURCE_DIR}/${PKG_LIB}/[^.]*.cpp)
     add_library(${PKG_LIB} STATIC ${${PKG_LIB}_SOURCES})
+    target_compile_definitions(${PKG_LIB} PUBLIC -DMPICH_SKIP_MPICXX -DOMPI_SKIP_MPICXX=1)
```
rbberger: You avoid this by using `MPI_CXX_SKIP_MPICXX` from the MPI package.

cmake/Modules/LAMMPSInterfacePlugin.cmake
akohlmey: Using `MPI_CXX` incorrectly links with libmpicxx.so. We do *not* use (or want to use) any MPI C++ interface. Neither the deprecated MPI class version not the MPI:: namespace. We only use the MPI C interface thus we should not even link with a C++ MPI library.
Whether our code is written in C++ is irrelevant in that respect. We only use the C interface.

cmake/CMakeLists.txt
akohlmey: I *want* the C interface to MPI. How can this be wrong?

cmake/Modules/LAMMPSInterfacePlugin.cmake
rbberger: I understand what you think you want. But this is not it. You're fixing the wrong thing.

rbberger: > Using MPI_CXX incorrectly links with libmpicxx.so.

This is not true in general.

cmake/CMakeLists.txt
rbberger: Because you want the C interface when compiling with a C++ compiler.

akohlmey:
@rbberger I have rolled back the changes to use the MPI_C target instead of MPI_CXX. We are too close to the stable release to have a controversial pull request get merged.

Despite using the corresponding CMake variable, the C++ MPI library is needlessly linked and thus it is also not possible without an unwarranted effort to debug whether the "skip MPICXX" defines are actually applied correctly.

sjplimp (approved):

cmake/CMakeLists.txt
```diff
@@ -196,6 +203,10 @@ if((CMAKE_SYSTEM_NAME STREQUAL "Windows") AND BUILD_SHARED_LIBS)
   set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
 endif()
 
+# do not include the (obsolete) MPI C++ bindings which makes for leaner object files
+# and avoids namespace conflicts. Put this early to increase its visbility.
+set(MPI_CXX_SKIP_MPICXX TRUE)
```
rbberger: Change to CACHE variable and FORCE it.

rbberger (approved):

PR #4599, opened 2025-05-21

### LINKED ISSUE #4607: Update fmt version to 11.1.0+ to support Clang++ 20
**Summary**

PR #4442 has introduced a breaking change for anyone wanting to compile using Clang++ 20.

LAMMPS version - stable_29Aug2024_update2
Clang++ version - 20.1.2 (0ubuntu1)

**Details**

PR #4426 and PR #4440 updated fmt version to 11.1.0, which was then later reverted to 10.2.1 by PR #4442.
This has caused an incompatibility with Clang++ 20 due to an issue (fmtlib/fmt#4177) which as been resolved (fmtlib/fmt#4187) as of fmt version 11.1.0.

I understand #4442 wishes to drop bundling fmt altogether and use std::format instead, which would be a great solution. However, the resolve to stick with 10.2.1 until then will remove the ability to compile with Clang++ 20.

My current work around, using official releases, is to compile using patch_19Nov2024 which supposedly bundles fmt 11.0.2. I don't know why, but Clang++ 20 compiles with no issue using this patch.

akohlmey:
**Summary**

This pull request combines multiple small fixes and corrections.

**Related Issue(s)**

Fixes #4607 
Fixes issue reported here: https://matsci.org/t/substantially-different-number-of-neighbours-when-simulating-eam-on-gpu/60211/

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- let `utils::expand_args()` stop with a descriptive error message, when a vector-style variable reference is provided with a wildcard without the required upper boundary
- refactor `fix neighbor/swap` for improved errors and modernized C++ code. Also remove dead code and redundant implementation
- add more in-depth discussion of which factors affect LAMMPS' serial and parallel performance
- place a prominent warning on the hybrid pair style documentation for when to *not* use it; based on recent recurring posts on MatSci.org
- other small documentation updates
- remove unused local variable in `Input::substitute()`
- avoid problems compiling the bundled fmtlib with modern compilers and C++20 or later enabled
- expand C++23 compatibility workflow to check clang in addition to gcc
- add bugfix for eam/gpu with OpenCL from @ndtrung81 
- improve error messages in `read_data` command.

**Post Submission Checklist**


akohlmey (commented):
@jtavenner-dev I had to do some significant refactoring of your contributed code to modernize some parts and have it better aligned with current programming conventions. Also, a couple minor bugs were fixed and I added support for type labels.

Unfortunately, I could not find an example input in the `examples` tree, so this is currently untested. Can you have a closer look and - ideally - provide some small, simple examples so that I can check as well.

jtavenner-dev:
Thanks for the code modernization updates and clearer error handling. I have fixed a minor bug unrelated to your changes regarding reading the options. I've also tested some short example scripts.

However, I'm not sure how to add to this pull request. I've updated the changes to a branch in sync with this pull, but can't seem to push any changes. Not sure if this is just permissions-related or if I'm missing something else.

akohlmey:
> However, I'm not sure how to add to this pull request. I've updated the changes to a branch in sync with this pull, but can't seem to push any changes. Not sure if this is just permissions-related or if I'm missing something else.

You cannot push because you have no write permission for the LAMMPS repository.
Just push it to your fork and let me know the name of the branch and then I can pull it myself.
or create a diff and attach it here.

jtavenner-dev:
Sounds good. I was trying to push to one of your forked lammps branches, so permissions make sense. I've synced the changes here and my updates to my lammps fork on the 'develop' branch.

akohlmey:
> Sounds good. I was trying to push to one of your forked lammps branches, so permissions make sense. I've synced the changes here and my updates to my lammps fork on the 'develop' branch.

@jtavenner-dev The changes are now integrated. I also moved and renamed the example folder, trimmed the inputs and make them use a potential that we already have included.

jtavenner-dev:
> @jtavenner-dev The changes are now integrated. I also moved and renamed the example folder, trimmed the inputs and make them use a potential that we already have included.

Looks good to me, thanks.

doc/src/pair_hybrid.rst
```diff
@@ -100,6 +100,52 @@ first is assigned to intra-molecular interactions (i.e. both atoms
...
+.. admonition:: When **NOT** to use a hybrid pair style
+   :class: warning
+
+   Using pair style *hybrid* can be very tempting to use if you need a
+   **many-body potential** supporting a mix of elements for which you
+   cannot find a potential file that covers *all* of them.  Regardless
+   of how this is set up, there will be *errors*.  The major use case
+   where the error is *small*, is when the many-body sub-styles are used
+   on different objects (for example a slab and a liquid, a metal and a
+   nano-machining work piece).  In that case the *mixed* terms **must**
```
sjplimp: "must" implies that LAMMPS will throw an error.  I don't think that is the case ?
I think you mean "should" (to minimize errors).  LAMMPS will allow you to do otherwise
but that way leads to darkness ...

sjplimp:
@akohlmey I think it would be good for @athomps to read the new cautionary text in pair_hybrid.rst.  To see if it all makes sense to him and he agrees with it.

doc/src/pair_hybrid.rst
akohlmey: Yup. It is a common mistake to make for somebody who is a native German speaker.

There are exceptions to this discussed further down in the documentation, which require editing of the potential file. So with "should" that is covered. I don't want to mention the exception here, since that applies only to experts.  This warning and its intensity is aimed at the many non-experts that are using hybrid styles incorrectly since they don't understand force fields well enough and the more complex the description becomes the smaller the intended effect.

akohlmey:
> @akohlmey I think it would be good for @athomps to read the new cautionary text in pair_hybrid.rst. To see if it all makes sense to him and he agrees with it.

So add him to the list of reviewers.

akohlmey:
@jtavenner-dev I've made some additional changes based on compiler warnings in commit 3dadcab0476230e37e99636199ed8822a6f5b953.

This removes dead code, unused arguments, unused variables etc. Please check.

There is one remaining issue:
In the `pick_i_swap_atom()` function the second `MPI_Allreduce()` call is bogus since it uses `i = -1`. Please advise on how to correct this.

ndtrung81 (approved):

sjplimp (approved):
But still please wait on Aidan for pair hybrid doc changes

akohlmey:
@jtavenner-dev please contact me ASAP!

We have to discuss your contributed fix. I've run into more issues when trying to test with valgrind's memory checker and have to conclude that there is some non-obvious fundamental problem that needs to be resolved before we can properly release it. We usually don't perform this kind of check automatically, so it went under the radar. But with the experience of the recent issues, I went ahead and confirmed the problem. I've guessed some additional changes that would avoid the out-of-bonds memory accesses, but those lead to other problems crashing the force/energy computation. Addressing this is beyond my level of understanding of the method in your code and the implementation of it. In my experience, the best way to resolve this would be that you revert to the version before pull request and either re-apply apply the changes in pieces (some are obviously correct and straightforward) or implement alternate fixes.

There are two possible ways to proceed right now since the next feature release is imminent and I need to merge this pull request very soon:
- I remove the fix entirely (after creating a branch with the current state so you can compare things)
- I leave everything "as is", but place an error message that prevents people from using the feature unless the detected issues are resolved.

jtavenner-dev:
> @jtavenner-dev please contact me ASAP!

> @jtavenner-dev I've made some additional changes based on compiler warnings in commit [3dadcab](https://github.com/lammps/lammps/commit/3dadcab0476230e37e99636199ed8822a6f5b953).
> 
> This removes dead code, unused arguments, unused variables etc. Please check.
> 
> There is one remaining issue: In the `pick_i_swap_atom()` function the second `MPI_Allreduce()` call is bogus since it uses `i = -1`. Please advise on how to correct this.

@akohlmey I've been unable to sync to the specific previous commit [3dadcab] of the PR, but have synced one of my branches to check the current end of this PR branch. The current changes make the command hang without error or the simulation progressing. I believe this is due to the removal of the MPI_ALLreduce, but have been unable to verify with difficulty getting my local branch to sync to a previous commit of this PR.

The issue with removing the second MPI command is that not all processors then reach a communication state, and therefore, the process is hung. It is possible that an alternative MPI is needed, in the initial development I was unable to get things working properly with a MPI send/recieve commands but that should what needs to happen there. I.e, the first MPI command should be sending the local data to sync with the others at the second.

This is true for both pick_i_swap_atom and pick_j_swap_neighbor.

akohlmey:
> I believe this is due to the removal of the MPI_ALLreduce, but have been unable to verify with difficulty getting my local branch to sync to a previous commit of this PR.

But that call is using i=-1 and that is plain wrong and is flagged by valgrind as out-of-range memory access and would eventually cause segmentation faults or corrupted data.

This is why I am suggesting, you go back to the state of the code in develop and create a new feature branch and have another feature branch with all the changes here. Then you can compare the two branches side by side and first transfer *only* your and my changes that objectively correct. Then you can review my additional changes and decide on a case-by-case basis whether they are correct or some other change is needed. E.g. what about all those `if (xxx){ double r = yyy;}` that have no effect and I thus removed. Perhaps the solution is to not declare the variable which shadows a variable outside that block.
All the changes I made are either because the code has no effect or because parameters are not used or because they have invalid memory access. But I don't know the logic of the science and the algorithm, so it is up to you. I *know* that some of my changes are not correct. Otherwise, I would not have pinged you. But I *also know* that the current code in develop is not correct either, even after fixing the indexing bug.

athomps (approved):
I approve. The pair hybrid admonition is entirely correct and appropriate.

Issue #4598, opened 2025-05-16
[Feature Request] Make use of atom-ID, molecule-ID, group-ID and others consistent


akohlmey:
**Summary**

The documentation has a mix of conventions for IDs (ex. atom): atom ID, atom-ID, atomID
For consistency we should stick with one version. The dominant one is with a blank, but that can cause confusion when used as a column header (should there be one or two columns), so the version with a dash should be better.

**Detailed Description**

Having a consistent nomenclature avoids confusion. Also for computes and fixes and alike we have (as .rst source):
```
.. code-block:: LAMMPS

   fix ID group-ID nve

* ID, group-ID are documented in :doc:`fix <fix>` command
```
This should become:
```
.. code-block:: LAMMPS

   fix fix-ID group-ID nve

* fix-ID, group-ID are documented in :doc:`fix <fix>` command
```
and thus reduce ambiguity (and thus make it easier for AI tools to make sense of the manual 😉  ...and humans, too 😁 ).

akohlmey:
This is planned as part of the efforts to review the documentation before the stable release.

akohlmey:
@athomps If there is time, we may discuss this on the next developer call.

jrgissing:
Note that 'atom1', 'atom2', etc. are also often used in the docs to refer to atom-IDs. I don't think it is a problem, just mentioning because we are using that convention within 'format' strings for the molecule JSON.

akohlmey:
> Note that 'atom1', 'atom2', etc. are also often used in the docs to refer to atom-IDs. I don't think it is a problem, just mentioning because we are using that convention within 'format' strings for the molecule JSON.

@jrgissing @salamaaya I am just coming across this while refactoring the Molecule class and folding in contents from PR #4547. There is need for further discussion and creating a consistent format and way to process the JSON files. I will thus soon post a pull request with my current status and some questions/suggestions.

PR #4600, opened 2025-05-21
Refactor Molecule class to add support for JSON format molecule files

akohlmey:
**Summary**

In the pull request the Molecule class is updated to transparently parse either native molecule files as well as JSON format files. This requires some refactoring to reduce redundancy and accommodate the differences in how both types of files need to be parsed and processed.

**Related Issue(s)**

#4598 
Includes all relevant changes from PR #4547 thus closes #4547

**Author(s)**

Axel Kohlmeyer, (akohlmey@gmail.com), Temple U
Aya Salama (asalama0204@gmail.com), Stevens Institute of Technology
Jacob Gissinger (jgissing@stevens.edu), Stevens Institute of Technology

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

The general flow of control for JSON files is different from native files.
The native files are read twice line-by-line on MPI rank 0 and the lines are broadcast to and then processed on all processors. JSON files are parsed into a `json` class on MPI rank 0 in
one go. The JSON parsing is tried first and if it throws an exception the file is assumed to be in native format and the parsing continues accordingly. For JSON files the data in the `json` class instance is then read, processed, and the result broadcast.

**Post Submission Checklist**

https://json-schema.org/


akohlmey:
@salamaaya @jrgissing here is my current status of the refactor.
For the JSON format, I am trying to keep more-or-less the same flow of control as for the native format. But while I am advancing, I see some opportunities to do things a bit different because the JSON format has more structure.

Some thoughts (in random order), where I hope to get your feedback:
- for simplicity and consistency, we should keep all keywords lowercase
- as noted in issue #4598 there is a mix of conventions in LAMMPS. The issue is mostly about the documentation, so it doesn't concern the file conventions much. so it would be quite conceivable to use only "id" where it is obvious from the context which id is referred to, e.g. in the "atoms", "bonds" etc. sections. Also the use of "atom1", "atom2" feels natural compared to something "atom-ID-1". An exception would be the "molecules" section, where we have "atomid" and "molid".
- I see the keyword "schema" pointing to the documentation. How about we create a real JSCON-scheme, so that files can be validated? See https://json-schema.org/ and e.g. https://github.com/python-jsonschema/jsonschema
- I am thinking that unlike for the native format, sections do not have to have exactly the number of entries for all items. In the native format this is so you can skip over it and know when the next section starts. Skipped items would retain their default value, e.g. molecule-ID 0 or charge 0.0. Since the "coords" section defines how many atoms are in the file, there cannot be uninitialized positions. In sections like "types" or "masses" leaving the default would result in an error during validation. Not sure if it is a good idea to add a keyword that would provide a default value.
- A similar question concerns the "format" entry. Should we require an exact match or only error out when a required entry is missing and ignore others. A related question is whether the order of columns can be arbitrary. I think being relaxed here could help reusing the same file, but it is a bit extra programming effort

jrgissing:
I agree that all keys should be lowercase. However, the 'format' should focus on human readability, and for some reason I find 'atomid' much less readable than 'atomID' or 'atom-ID'. I do not think that we should throw an error on format strings, that is just meant for humans to easily see what is going on.

By official schema, does that mean a new extension for each LAMMPS JSON we come up? If so, does that mean an additional third-party library would have to be imported to verify these JSON schemas?

Default values is always a little dangerous, I am usually a fan of making users type out their own 'default' values.

One other issue we are running into, is that sometimes we want to read a JSON object into the molecule class instead of a file. This comes up when a molecule JSON object is embedded in another JSON. Currently @salamaaya is working around this by writing out a new file in JSON format (containing just the molecule JSON object), reading that, and then deleting the file.

akohlmey:
> find 'atomid' much less readable than 'atomID' or 'atom-ID'.

Let's compromise on 'atom-id' then.

> By official schema, does that mean a new extension for each LAMMPS JSON we come up? If so, does that mean an additional third-party library would have to be imported to verify these JSON schemas?

The schema documents the exact conventions for how to structure the JSON file. A validator can be used to verify that not only the file has the correct syntax but also follows the correct conventions for which field to put where. A validator would would be an external program that uses the schema file to validate that a JSON file is conforming to the requirements.

> I do not think that we should throw an error on format strings, that is just meant for humans to easily see what is going on.

Sorry, but that statement makes little sense. To the parsing the format list should define the order of the columns in the data section. So it is much more than just something to make the file easier to read. Anything else could be quite confusing. Specifically when manipulating a dataset with a program, it can result in a reordering.

> One other issue we are running into, is that sometimes we want to read a JSON object into the molecule class instead of a file. This comes up when a molecule JSON object is embedded in another JSON

This may be possible to implement with a custom constructor. But that is something to look into after the core of the functionality is sorted out. In my experience, things will get more convoluted and ugly to just figure out a good way to finish the implementation before one can look at making it nice(r) and clean(er).

akohlmey:
@jrgissing Here is a minimal example for JSON schema validation (after spending about 1 hour reading and learning some basics on https://json-schema.org):
The file [twomols.json](https://github.com/user-attachments/files/20380078/twomols.json) is an example molecule file and [molecule-schema.json](https://github.com/user-attachments/files/20380080/molecule-schema.json) is a corresponding schema file describing how the data is organized.

I install a validator written in python with
```
pip install --user check-jsonschema
```

and then run the validation:
```
check-jsonscheme --schemafile molecule-schema.json twomols.json
```
This should be OK. This currently does not support typelabels, so if I modify the molecule file accordingly, it will fail the validation with:
```
$ check-jsonschema --schemafile molecule-schema.json twomols.json
Schema validation errors were encountered.
  twomols.json::$.types.data[0][1]: 'A' is not of type 'number'
  twomols.json::$.types.data[1][1]: 'B' is not of type 'number'
```
This would be useful for any software or script that generates a JSON format file for LAMMPS.

jrgissing:
I see, that is a nice idea. 

Btw, it looks like type labels can be supported with: "type": ["integer", "string"] in schema file.

It seems like this would be useful during a LAMMPS run, but I am not sure how something like this could be integrated.

akohlmey:
> Btw, it looks like type labels can be supported with: "type": ["integer", "string"] in schema file.

As I wrote, I literally spent only an hour reading about the schema format.
 
> It seems like this would be useful during a LAMMPS run, but I am not sure how something like this could be integrated.

You are missing the point entirely. The purpose of the schema is
- document the format
- allow to validate that a file is in the correct format and uses the conventions correctly. This latter point would apply to people writing tools that *generate* molecule files or when people manually write / edit them. The validation is a simple way to check if it will be parsed at all. There will still be additional checks inside of LAMMPS, since the data also has to make sense on a physics level, not just formally. Thus there is no value in integrating this validation into LAMMPS itself.

akohlmey:
@jrgissing @salamaaya here is another update and further exploration of how we can add JSON format and now with the additional option to process JSON objects.

This required some changes.
- I've created a minimal constructor that just sets up an empty object
- The processing of the command line arguments for the "molecule" command is now a `command()` method
- The distribution of the JSON class data is now done by serializing it to a binary vector, broadcasting it and deserializing it. This requires more memory, but then there is only one communication. Molecule files are supposed to be "small", so I think we can do this. It simplifies the code quite a bit
- Now there is a `from_json()` method that takes a JSON object as argument. This is used by the `command()` function, but can also be used elsewhere. I added an overload to `Atom::add_molecule()` for example. This step makes pimplification extremely inconvenient, so I am including "json.h" into the header now, which violates a LAMMPS design goal and is very invasive since atom.h is included in many places. This is not yet permanent. I want to look into slimming down the JSON class at a later point anyway. Perhaps we need a wrapper.
- I've added a "units" global property to the JSON schema and parser
- Also, I am starting with rather strict validation and checking of syntax

I am not happy with how everything currently is done. But we can look into improving things later. First we need all required functionality working. There is a lot of work still to be done.

jrgissing:
@akohlmey this feature is coming together nicely. If @salamaaya removes changes to core files, can we merge #4547 into this PR, so we can continue to update our example files?

akohlmey:
> If @salamaaya removes changes to core files, can we merge #4547 into this PR, so we can continue to update our example files?

@jrgissing I prefer doing this in steps myself and manually. It has been a rather iterative process so far to try and keep everybody (Steve, you, me) happy with the results. Eventually, everything from #4547 that is relevant should show up here. My time to work on this, however, is currently a bit limited due to some unexpected tasks I need to do for Temple HPC (which is where most of my salary comes from). This is a required feature for the next feature release so it is near the top of my TODO list.

jrgissing:
ok. Note that this is a case where I would suggest 'merging' rather than 'closing' the original PR. We will continue to update the bond/react example there.

akohlmey:
@jrgissing @salamaaya after a long wait we have some more progress in the branch again. There is still too much redundant code for my taste, but with the current strategy it should be possible to complete this up to the point of the bond topology. For "special", "shake", and "body", i want to use nested objects (e.g. `moldata["shake"]["flags"]["data"]` instead of `moldata["shake_flags"]["data"]`) as that makes more sense and comes at no extra cost (unlike for the traditional molecule file format). But that may be after the feature release.

akohlmey:
@jrgissing @salamaaya I've moved the example files over from PR #4547, updated and ran them to produce log files.
The schema file is also uploaded to download.lammps.org.
The special, shake and body sections still need to be implemented but we can skip them for now. Major missing piece is the updated documentation.

Let me know, if there still is something missing from #4547 that needs to be moved.

doc/src/molecule.rst
```diff
@@ -669,6 +680,189 @@ the file format.
...
+     - double
+     - no
+     - overrides the auto-computed total mass for the template
+   * - inertia
+     - list with 6 doubles
+     - no
+     - overrides the auto-computed moments of inertia
+   * - coords
+     - a data block
+     - yes
```
jrgissing: note that the Coords section is not required in a native molecule file. for example, coords are not used for most 'fix bond/react' features, and I have seen quite a few examples (from other groups) that do not bother to write out coords

akohlmey: @jrgissing Thanks for the note. I was not aware of this. We use the size of the "coords" entry to infer the number of atoms. Should I change this to the "types" section?

src/molecule.cpp
```diff
@@ -138,43 +152,1120 @@ Molecule::Molecule(LAMMPS *lmp, int narg, char **arg, int &index) :
...
+      error->all(FLERR, Error::NOLASTLINE, "JSON molecule data with unsupported revision {}", rev);
+  } else {
+    error->all(FLERR, Error::NOLASTLINE, "JSON data does not contain required 'revision' field");
+  }
+
+  // optional fields
+
+  if (moldata.contains("units") && (comm->me == 0)) {
+    if (std::string(moldata["units"]) != update->unit_style)
+      error->warning(FLERR, "Inconsistent units in JSON molecule data: current = {}, JSON = {}",
```
jrgissing: if we are including this key (great idea, btw!), should this just be a warning? seems like an error to me

akohlmey: I was mirroring the behavior from the "native" file, where there can be a "units = xxx" in the title line.

The only problem with making this an error would be that I think "real" and "metal" are equivalent for molecule templates (using Angstroms, electron charges, and AMU), so there should be a special case check for those and make them not an error.

doc/src/molecule.rst
jrgissing: yes, that should work (no other option, i guess)

jrgissing (approved):
barring my couple (mostly minor) comments above, this looks good to me. excited to extend and test more rigorously soon

akohlmey:
> barring my couple (mostly minor) comments above, this looks good to me. excited to extend and test more rigorously soon

Updates are under way.

Issue #1624, opened 2019-08-06
[Feature Request] KOKKOS fix rigid


ag949:
**Summary**
 
Porting fix rigid to KOKKOS

**Detailed Description**

In the past I've benefited significantly from using LAMMPS accelerated with KOKKOS. I am trying to run large scale simulations in GPUs with rigid bodies. Is porting fix rigid planned for the future?

jaj52:
This will be super useful!

Pakketeretet2:
I would like to take a look because I am interested in it myself but I would have to bother @stanmoore1 a lot with Kokkos-related questions, since porting a complicated fix like rigid seems to be a lot harder than porting a pair style. 
Are you interested in fix rigid or fix rigid/small?

ag949:
This sounds great, thank you. As you say, it doesn't look trivial.  We would be interested in fix rigid.

stanmoore1:
I'm happy to answer questions about Kokkos.

Pakketeretet2:
OK, I will take a look at it. Unfortunately, I am more interested in rigid/small but I'll see what I can do for either. I think rigid/small would be more suitable for GPUs and the like because it seems more similar to what Hoomd-Blue does.
@stanmoore1, should a simple port be reasonably performant or are there special things I should worry about when porting fixes like this? Or should I just port and get it working first, and then worry about performance?

stanmoore1:
@Pakketeretet2 yes a simple port is fine, i.e. just thread over loops of atoms. After we have a port then we'll worry about performance, I can do some tuning. I used c++11 lambdas extensively in #1650 and it made porting easier since I didn't have to mess with functors or tags.

Pakketeretet2:
OK, since fix propel/self is pretty much done from a programming point of view I have started with this now.

stanmoore1:
Great, feel free to ping me either here on Github or via email if you have questions.

Pakketeretet2:
Will do. For a second I couldn't find the lambdas anywhere but I found them. They look nicer than C++11 lambdas so they were hard to spot. :)

Pakketeretet2:
The latest changes with minimize seem to cause linking problems with clang on my laptop, but it might be something dumb on my part. I'll look into it more tomorrow.

akohlmey:
> The latest changes with minimize seem to cause linking problems with clang on my laptop, but it might be something dumb on my part. I'll look into it more tomorrow.

You are building LAMMPS with cmake, right?
[...]

PR #4609, opened 2025-05-29
fix rigid/small/kk

anjohan:
**Summary**

This pull request provides a Kokkos accelerated version of a subset of the functionality in `fix rigid/small`.

**Related Issue(s)**

#1624 This PR resolves that issue for a subset of use cases.

**Author(s)**

Anders Johansson, Sandia National Laboratories
Stan Moore, Sandia National Laboratories

**Licensing**

Claude Code was heavily used to fix merge conflicts, add missing features, and review and debug the code.

**Backward Compatibility**

No issues expected

**Implementation Notes**

The implementation is largely a translation of the non-Kokkos version.

There are two algorithmic differences:
- A body sendlist. As part of `pre_neighbor`, I construct a list of the rigid bodies that will need to be sent during forward/reverse communications on the subsequent timesteps. This avoids having to loop over all the atoms to check if they own rigid bodies on subsequent timesteps. In the example image below, packing forward communication on the left processor now only requires looping over the two rigid bodies that need to be communicated to the right processor (the dashed arrows).
pack_forward
- Packing of "holes" that have been left behind when bodies are sent away during exchange, as shown in the image below. Doing this in parallel requires an initial parallel scan for cumulatively counting the number of rigid bodies that need to be moved and their counts (6=0 and 8=1) below, followed by another parallel scan to cumulatively count the number of open spots (0=0 and 4=1 below), which are filled with the corresponding rigid bodies (6 and 8) in the `final` statement.
packing_holes

Beyond the new source files, the main modification is the new reverse communication for fixes on device, as well as domain and math functions.

Necessary TODOs:
- [ ] Test triclinic boxes to check new mathextra/domain functions
- [ ] Test a system where not all atoms belong to rigid bodies
- [ ] Add error/warning for untested functionality (langevin, restart files, set molecule)
- [ ] Add acceptably sized test. The scripts in examples/rigid/ are quite forgiving due to the limited number of interactions. I made my own test system of 1000 8-atom rigid cubes that helped me discover many more bugs. After running 10000 steps of this system, the final temperature of the OpenMP+MPI and CUDA+MPI runs both agree with the non-Kokkos code to 11 significant digits. The size of this data file is 649 kb, which is probably a bit too big to include - I will make a smaller one. What is an acceptable size? (@akohlmey )

Possible TODOs:
- [ ] NVE, NVT and NPT support
- [ ] Optimize langevin to not use host code
- [ ] Reimplement setup w/rendezvouz in Kokkos
- [ ] Optimize exchange to only send necessary data (similar to fix shake)

Other issues:
- [ ] Due to the lack of DualViews and related bookkeeping that mirror the host pointers, fix rigid/small/kk currently requires device sorting and communication. When running Kokkos with Serial or OpenMP, this requires the `sort device comm device` command line flags to override the LAMMPS default. Furthermore, this disallows the simultaneous use of any other fix that does not also support device sorting and communication.
- [ ] The use of the non-Kokkos setup routines requires host communication during the setup phase, while device communication is required during a run. I made a minor modification to CommKokkos to change the check for device communication from once to every timestep to allow these flags to be dynamic. This should probably be changed to "check during setup, and once during the run" to avoid the minor overhead.

**Post Submission Checklist**


akohlmey:
@anjohan this branch has been idle for quite some time now. Can you please provide some update on its progress? Please also note that it is in need of an update with upstream and resolution of the resulting merge conflicts.

akohlmey:
@anjohan In light of Aidan's advice to be more "rigid" in handling idle pull requests, I am closing this for now. Please resubmit when the feature is close to completion.

stanmoore1:
Update from Claude:

## Full status update — `fix rigid/small` Kokkos port (PR #4609 work)

This summarizes all the work done on top of the original PR (13 commits, `084135f`…`6453055`). The original PR was ~39k commits behind `develop` and used device-only `Kokkos::View`s that manually shadowed the base-class host pointers; the bulk of the work was merging it up, converting to tied `DualView`s, completing the NH/langevin styles, and getting MPI migration correct.

### 1. Merge up to current `develop` (`084135f`)
Merged `origin/develop` into the branch. `develop` had independently grown the infrastructure the PR had hacked in (per-call host/device comm dispatch in `CommKokkos`, `reverse_comm_device` in `fix.h`, per-fix legacy sort fallback in `AtomKokkos`), so the conflicts resolved by taking develop's side and dropping the PR's `comm_kokkos`/`atom_kokkos`/`fix.h` workarounds.

### 2. MathExtra (`8acae2d`)
Added the rigid-body helpers develop's `math_extra_kokkos.h` lacked (`angmom_to_omega`, `q_to_exyz`, etc.), inline, matching develop's style. Did **not** add the PR's separate `math_extra_kokkos.cpp`.

### 3. Core DualView conversion (`a5b0690`) — the main rework
Converted the 8 owned device-only Views to `DualView`s tied to the base host pointers via `memoryKK->grow_kokkos`. This is what lets the unmodified host setup/restart routines stay on the host while the device kernels see the same memory, and removes the PR's requirement for device-only comm/sort. Body setup (`create_bodies`/rendezvous) is intentionally kept on the host (setup-only, STL maps + MPI rendezvous callback, no per-timestep cost); the `Body` array is bridged by `copy_body_host/device`.

### 4. Tests (`17d2920`)
Added committed `data.rigid.cubes` (+ generator `gen_cubes.py`), the `in.rigid.cubes` driver, and triclinic + partial-rigid variants, plus reference logs.

### 5. Nose–Hoover integrators (`1e258f2`)
Added `rigid/{nve,nvt,npt,nph}/small/kk` (+ `/host`). NH chain math runs on the host (all scalar global state, per the design decision), with per-body kernels on the device.

### 6. Device langevin (`994bed2`)
Moved the langevin thermostat onto the device with a Kokkos RNG pool.

### 7–10. Robustness, docs, and the MPI migration fixes

- `f2a086f`, `932651c`: got atom **exchange** correct under MPI. First via the proven host exchange path (flushing the tied DualViews host↔device around migration), then fixing the host/device path *detection* so the flush is skipped only when the device exchange is genuinely active (`exchange_comm_device && !commKK->exchange_comm_legacy`).
- `6453055`: got the **device** exchange path itself working under MPI (see detailed comment above) — root cause was `sort_legacy` permanently falling back to host sort because `sort_device` was set too late, letting a host sort corrupt the device exchange's body bookkeeping. Fixed by setting `sort_device=1` in the constructor, handling the pre-setup `sort_kokkos` call, and requiring exchange+sort to run on the same side.
- `aea0bfc`: guard against a missing body owner in `reset_atom2body` (avoids an OOB `d_bodyown(-1)` read).
- `f749819`, `8038bac`, `1e53768`: document the host-side design decisions, the new accelerator variants in `fix_rigid.rst`, and the experimental status of langevin/`infile`.

### Verification status

**Bit-identical to non-Kokkos**, line-for-line through full runs:
- **Host path** (default `-k on -sf kk`) and **device path** (`-pk kokkos comm device sort device`), at **np 1/2/4**, for: `cubes`, `tri` (triclinic), `partial` (partial-rigid), and `nve`/`nvt`/`npt`/`nph`.
- **Kokkos/OpenMP, 4 threads vs non-Kokkos CPU** (single rank): all 7 deterministic styles identical to full printed precision.

**Validated statistically only (by design):**
- `langevin` — the device Kokkos `Random_XorShift64` pool does not reproduce the host `RanMars` stream bit-for-bit, so trajectories differ while fluctuating around the target temperature. Flagged experimental in the docs.

### Known limitations / not done
- Langevin under MPI is experimental (warned in code + docs).
- Reading body properties from a file (`infile`) and molecule insertion (`set_molecule`, deposit/pour) are experimental.
- `gravity` keyword not supported on device.
- Mixed comm/sort modes (device exchange + host sort, or vice versa) are rejected with a clear error rather than supported.
- Exchange selective-packing optimization (shake-style variable stride) was scoped but intentionally not implemented — the current fixed-stride exchange is correct and verified; this is a bandwidth optimization for a future commit.
- No GPU/CUDA build was available in this environment, so device-path verification used the Kokkos OpenMP/Serial backends (where `LMPDeviceType == LMPHostType`); the logic is backend-independent but a CUDA+MPI run is still worth doing before merge.

akohlmey:
@stanmoore1 I have some bugfixes for KOKKOS/GPU from merging this PR into a branch with GPU enabled unittesting and other pending pull requests included. How do you want to handle this? Do you want me to cherry pick these (rather mechanical and local) fixes here or just give you the hashes/commits?

akohlmey:
> Here are the comments from Claude Code (Opus 4.8) on the unit test failures

---

Since the PR notes it hasn't been exercised on GPUs, here are results from a KOKKOS/HIP build (gfx1103, host=OpenMP): the five `fix rigid/*/small/kk` force-style tests (`rigid_small`, `rigid_nve/nvt/npt/nph_small`) currently fail. All are GPU-only (they pass / are skipped on a host-only KOKKOS build, since host space == device space). Two separate problems, the first of which I have fix candidates for:

### 1. Setup-time DualView "concurrent modification" aborts (fix candidates available)

On a GPU build the styles aborted in `VerletKokkos::setup()` with Kokkos `DualView` concurrent host/device modification errors, from two independent causes:

- **`setup_device_push()` on `bodyown` et al.** — the setup-time device exchange/sort (enabled by `exchange_comm_device`/`sort_device` in the ctor) marks the body DualViews device-modified, then `setup_device_push()`'s `modify_host()` aborts. Since those device copies are pre-body placeholders at that point and the host setup routines have just written the authoritative arrays, a `clear_sync_state()` before the `modify_host()` block resolves it.
- **`setup()` / NH `setup()` on `atom:x`** — both mark `modified(Host, datamask_modify)` but, unlike the port's own `setup_pre_neighbor()`, never reconcile to the device before `ModifyKokkos::setup()` marks `modified(Device, …)`. Adding `sync(execution_space, datamask_read)` before `setup_device_push()` (matching `setup_pre_neighbor`) fixes it.

I have these as two small commits (one for `fix_rigid_small_kokkos.cpp`, one for `fix_rigid_nh_small_kokkos.cpp`) that you're welcome to cherry-pick if useful: https://github.com/lammps/lammps/commit/bd307c9479e50dff8515cfff278bcae9ea1f99a4 and https://github.com/lammps/lammps/commit/24119e27b344b3b625f0081a1aefe3f49108792f

### 2. `dof()` routes a `DOF` reverse comm to the device packer (needs a design decision)

With the aborts fixed, all five tests then hit:

```
ERROR: attempting invalid reverse comm on device (src/KOKKOS/fix_rigid_small_kokkos.cpp:1730)
```

Root cause: the thermo `temp` column calls `compute temp -> fix->dof()` every output step during the run. `FixRigidSmallKokkos::dof()` delegates to `FixRigidSmall::dof()`, which does `commflag = DOF; comm->reverse_comm(this, 3)`. Because `reverse_comm_device = 1` is set for the run, `CommKokkos::reverse_comm(fix)` routes this to `pack_reverse_comm_kokkos`, which only accepts `commflag == FORCE_TORQUE` and errors.

There's a secondary issue in the same path: the `if (!setupflag)` guard in `dof()` skips the host flush of the body arrays during the run, so even with the comm routed correctly the DOF count would be read from stale host data.

Two ways to resolve it, depending on your preference:

- **Keep `dof()` on the host path (simplest).** In `FixRigidSmallKokkos::dof()`, flush the arrays `FixRigidSmall::dof()` reads (`bodyown`/`bodytag`/`atom2body`, and `atom->mask` for the group) to the host *unconditionally* (drop the `!setupflag` guard), and temporarily clear `reverse_comm_device` around the `FixRigidSmall::dof()` call so `CommKokkos` uses the host `CommBrick::reverse_comm` + the base host packer that handles `commflag == DOF`. `dof()` is an infrequent diagnostic query, so the host round-trip is cheap.
- **Teach the device packer about `DOF`.** Extend `pack_/unpack_reverse_comm_kokkos` to handle `commflag == DOF` (3 per-atom values) so it stays on-device. More code, and needs the DOF inputs resident on the device.

I'd lean toward the first. Happy to put up a PR against this branch with whichever you prefer.

<sub>(Analysis and the candidate patches above were produced with AI assistance / Claude, reviewed by me.)</sub>

stanmoore1:
I think this is working on the CPU with Kokkos Serial/OpenMP backends. It needs more GPU testing though.

akohlmey:
> I think this is working on the CPU with Kokkos Serial/OpenMP backends. It needs more GPU testing though.

Yup. Running this on my AMD APU now.

akohlmey:
@stanmoore1 here are my findings (and those of Claude). Initially, there were still segfaults, but at a different location than previously, but then also test failures. This now passes the built-in tests for all variants including KOKKOS/OpenMP and KOKKOS/HIP without changes to the YAML files.

---

Follow-up to my earlier note on the GPU runs: with the setup-time DualView abort and the `dof()` reverse-comm routing resolved, I ran the five `fix rigid/*/small/kk` force-style tests on a KOKKOS/HIP build (gfx1103, host=OpenMP, `KOKKOS_PREC=double`) and chased the remaining failures all the way down. There were **four distinct GPU-only bugs** — each invisible on a host-only KOKKOS build, where host space == device space hides every stale-sync, and each masked behind the previous crash. All four are fixed; **all 133 `FixTimestep` tests now pass** and a CPU vs `-sf kk` rigid melt (`examples/rigid/in.rigid.small.kk`) is **bit-identical through 500 steps**. Every fix is contained to `src/KOKKOS/fix_rigid_small_kokkos.{cpp,h}` — no base-class or core edits.

The common thread: a base-class (KOKKOS-unaware) host write to a tied DualView leaves it marked **device-modified** from the constructor's setup-time exchange, so a later `sync<Device>()`/sort silently operates on stale device data. The cure each time is `clear_sync_state()` + `modify_host()` right after the host write.

### The four bugs

1. **Setup segfault in `dof()`.** `ModifyKokkos::setup()` runs `compute[]->setup()` (compute temp → `dof()`) *before* the fix `setup()` loop. `dof()`'s `sync_host()` pulled the stale device-modified placeholder over the valid host `atom2body` that `setup_pre_neighbor()` had just written, so `dof()` indexed out of bounds. Fix: reconcile the body DualViews to host-authoritative at the end of `setup_pre_neighbor()`.

2. **"Invalid reverse comm on device" on the 2nd run.** With `reinitflag` on (default), every run re-runs the host body build (`setup_bodies_static/dynamic`), whose non-`FORCE_TORQUE` forward/reverse comms (`FULL_BODY`/`INITIAL`/`XCM_MASS`/`ITENSOR`/`VCM_ANGMOM`) and `pre_neighbor`/`reset_atom2body`/`image_shift` calls were routed to the device packers because run 1's `setup_device_push()` had turned on device comm and `setupflag`. Fix: a `setup_host_rebuild` guard (the three overrides previously keyed only on `!setupflag`) plus clearing the device comm flags around the base call — restoring run-1 conditions.

3. **Wrong rigid-body setup velocities (the subtle one).** A single GPU run started with the wrong temperature/KE even though PE (positions) matched the CPU exactly — `set_v()` was leaving raw, unprojected velocities. The setup-time atom sort fires from inside `setup_bodies_static()` (still `setupflag==0`, after `create_bodies()`), and `sort_kokkos()` sorted the **stale device copy** of the body arrays (the `sync<Device>()` was a no-op against the placeholder device-modified flag), permuting them out of alignment with the atoms the sort reordered. `reset_atom2body()` then found `bodyown[map(bodytag[i])] < 0` for the displaced atoms and left `atom2body = -1`; `set_v()` skips those. Fix: in the `setupflag==0` branch of `sort_kokkos()`, make the device match the authoritative host *before* the sort and sync the sorted result back to the host *after*. (`atom_modify sort 0 0.0` was the bisection knob that pinned this to the sort.)

4. **Restart abort: "exchange and sorting on the same side".** The with-rmass restart variant restores per-atom rmass via `fix property/atom` (`property/atom/kk`), which supports device *sort* but not device *exchange*. It forces `CommKokkos` onto a host exchange (sets `kokkos->exchange_comm_legacy`) while the sort stays on the device, and the fix's hard same-side `error->all` fired. `fix property/atom` + `fix rigid/small` is a normal combination that works on the CPU, so erroring was wrong. Fix: instead of erroring, tie `sort_device` to the exchange side — `sort_device = exchange_comm_device && !kokkos->exchange_comm_legacy` — so a forced host exchange pulls the sort onto the host too (set in `setup_pre_neighbor()` before the setup-time sort, repeated in `setup_device_push()`). No-op in the normal all-device case.

### Incorporating

Four cherry-pickable commits, one per bug, in this order (each builds; the suite goes green only once all four are in):

```
e1c91afc8f  KOKKOS: fix setup-time segfault in fix rigid/small/kk dof()
8abb19d2e9  KOKKOS: fix invalid device reverse comm on 2nd run of fix rigid/small/kk
fd03e74b08  KOKKOS: fix wrong rigid-body setup velocities from device sort in fix rigid/small/kk
e0d5b9a269  KOKKOS: let fix rigid/small/kk follow a forced host exchange onto the host sort
```

They are a clean contiguous range on `akohlmey:preview` touching only `fix_rigid_small_kokkos.{cpp,h}`, so `git cherry-pick e1c91afc8f^..e0d5b9a269` applies them as a unit, or take them individually. Happy to open a PR against this branch with whichever subset you prefer, or to fold any of them differently.

stanmoore1:
All known issues addressed. This is ready for final testing (CPU and GPU).

akohlmey:
> All know issues addressed. This is ready for final testing (CPU and GPU).

Unittests for Kokkos/OpenMP pass.

akohlmey:
@stanmoore1 I am getting the following with Kokkos/HIP when running the unit tests for all fix rigid/small tests
```
[ RUN      ] FixTimestep.kokkos_omp
/home/akohlmey/compile/lammps-claude/unittest/force-styles/test_fix_timestep.cpp:1131: Skipped
Cannot test KOKKOS/OpenMP with GPU support enabled

[  SKIPPED ] FixTimestep.kokkos_omp (0 ms)
[ RUN      ] FixTimestep.kokkos_serial
/home/akohlmey/compile/lammps-claude/unittest/force-styles/test_fix_timestep.cpp:1154: Skipped
Cannot test KOKKOS/Serial with threading support enabled

[  SKIPPED ] FixTimestep.kokkos_serial (0 ms)
[ RUN      ] FixTimestep.kokkos_gpu
Kokkos::DualView::modify_host ERROR: Concurrent modification of host and device views in DualView "fix:vatom"

Backtrace:
```

stanmoore1:
Thanks I'll take a look

stanmoore1:
@akohlmey now all tests are passing on H100 as far as I can tell. Can you check HIP?

akohlmey:
> @akohlmey now all tests are passing on H100 as far as I can tell. Can you check HIP?

Done. Looking good. I will have to do a systematic check next week, if there are any KOKKOS or other package related bugfixes in the preview branch that have not made it back to develop. I don't expect any, but better safe than sorry. :safety_pin: 

```
$ ctest -R small
Test project /home/akohlmey/compile/lammps-claude/build-hip
    Start 676: FixTimestep:rigid_nph_small
1/6 Test #676: FixTimestep:rigid_nph_small ......   Passed    1.93 sec
    Start 678: FixTimestep:rigid_npt_small
2/6 Test #678: FixTimestep:rigid_npt_small ......   Passed    1.36 sec
    Start 682: FixTimestep:rigid_nve_small
3/6 Test #682: FixTimestep:rigid_nve_small ......   Passed    1.35 sec
    Start 684: FixTimestep:rigid_nvt_small
4/6 Test #684: FixTimestep:rigid_nvt_small ......   Passed    1.35 sec
    Start 686: FixTimestep:rigid_small
5/6 Test #686: FixTimestep:rigid_small ..........   Passed    1.35 sec
    Start 687: FixTimestep:rigid_small_dipole
6/6 Test #687: FixTimestep:rigid_small_dipole ...   Passed    1.33 sec

100% tests passed, 0 tests failed out of 6

Label Time Summary:
no_respa    =   8.67 sec*proc (6 tests)

Total Test time (real) =   8.70 sec
```

stanmoore1:
Everything is working on H100 GPU. This is ready to merge from my POV.

akohlmey (approved):

PR #4614, opened 2025-06-05

akohlmey:
**Summary**

This pull request combines multiple small changes and fixes that do not warrant a pull request of their own.

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- accept denormal floating-point numbers when converting with `ValueTokenizer::next_double()` or `utils::numeric()`
- add overloads to `utils::print()` so it can be used without providing a FILE pointer which will use `stdout`.
- bugfix from @stanmoore1 for KOKKOS memory leak fixes from PR #4532. Per-atom data must be initialized to zero.
- update list of known LAMMPS commands for LAMMPS-GUI

**Post Submission Checklist**


src/tokenizer.cpp
```diff
@@ -365,6 +367,11 @@ double ValueTokenizer::next_double()
     return val;
     // rethrow exceptions from std::stod()
   } catch (std::out_of_range const &) {
+    // could be a denormal number. try again with std::strtod().
+    char *end;
+    auto val = std::strtod(current.c_str(), &end);
+    // return value of denormal
+    if ((val != 0.0) && (val > -HUGE_VAL) && (val < HUGE_VAL)) return val;
```
jtclemm: I'm not sure I understand this, could a value of zero ever enter this catch statement?

akohlmey: I copied this from a stackoverflow response to separate denormal numbers from NaN and Inf and I think you are correct: the first comparison can be dropped in this case due to the context.

jtclemm (approved):

sjplimp (approved):

PR #4615, opened 2025-06-06
Roll back neighbor swap changes and flag items that need fixing

akohlmey:
**Summary**

This rolls back all non-trivial changes in the `fix neighbor/swap` and inserts "TODO" comments explaining why the following line of code can (and should?) be improved.

**Related Issue(s)**

Discussion in PR #4599 

**Author(s)**

Jacob Tavenner, NASA Ames Research Center, tavennerjacob@gmail.com
Axel Kohlmeyer, Temple U, akohlmey@gmail.com

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

When running the bundled example, valgrind's memory checker flags an illegal memory access:
```
==2465299== Invalid read of size 2
==2465299==    at 0x484CE40: memmove (vg_replace_strmem.c:1415)
==2465299==    by 0xE2A43FC: UnknownInlinedFun (string_fortified.h:29)
==2465299==    by 0xE2A43FC: typerep_do_unpack.constprop.0 (typerep_yaksa_pack.c:318)
==2465299==    by 0xE2CD9AD: UnknownInlinedFun (typerep_yaksa_pack.c:425)
==2465299==    by 0xE2CD9AD: do_localcopy.lto_priv.0 (utils.c:74)
==2465299==    by 0xE2CDD9C: MPIR_Localcopy (utils.c:324)
==2465299==    by 0xE247923: MPIDI_POSIX_mpi_release_gather_gather.lto_priv.0 (release_gather.h:302)
==2465299==    by 0xE247C87: MPIDI_POSIX_mpi_allreduce_release_gather (posix_coll_release_gather.h:325)
==2465299==    by 0xE24855C: UnknownInlinedFun (posix_coll.h:292)
==2465299==    by 0xE24855C: UnknownInlinedFun (shm_coll.h:47)
==2465299==    by 0xE24855C: MPIDI_Allreduce_intra_composition_gamma.lto_priv.0 (ch4_coll_impl.h:595)
==2465299==    by 0xE261938: UnknownInlinedFun (ch4_coll.h:362)
==2465299==    by 0xE261938: UnknownInlinedFun (ch4_coll.h:470)
==2465299==    by 0xE261938: MPIR_Allreduce (mpir_coll.c:4849)
==2465299==    by 0xE095221: UnknownInlinedFun (c_binding.c:3955)
==2465299==    by 0xE095221: PMPI_Allreduce (c_binding.c:4008)
==2465299==    by 0x71CF5B5: LAMMPS_NS::FixNeighborSwap::pick_i_swap_atom() (fix_neighbor_swap.cpp:561)
==2465299==    by 0x71D0442: LAMMPS_NS::FixNeighborSwap::attempt_swap() (fix_neighbor_swap.cpp:426)
==2465299==    by 0x71D088A: pre_exchange (fix_neighbor_swap.cpp:401)
==2465299==    by 0x71D088A: LAMMPS_NS::FixNeighborSwap::pre_exchange() (fix_neighbor_swap.cpp:376)
==2465299==  Address 0xab7c6fbc is 4 bytes before a block of size 65,536 alloc'd
==2465299==    at 0x4846607: realloc (vg_replace_malloc.c:1801)
==2465299==    by 0x6901769: LAMMPS_NS::Memory::srealloc(void*, long, char const*) (memory.cpp:99)
==2465299==    by 0x65AA4CE: grow<int> (memory.h:79)
==2465299==    by 0x65AA4CE: LAMMPS_NS::AtomVec::grow(int) (atom_vec.cpp:214)
==2465299==    by 0x6A5E3CD: LAMMPS_NS::Replicate::command(int, char**) (replicate.cpp:313)
==2465299==    by 0x6875047: LAMMPS_NS::Input::execute_command() (input.cpp:867)
==2465299==    by 0x6875DBE: LAMMPS_NS::Input::file() (input.cpp:313)
==2465299==    by 0x20296E: main (main.cpp:65)
```

**Post Submission Checklist**


akohlmey:
@jtavenner-dev Here is a branch with the rollback of my (too aggressive) changes and adding "TODO" comments instead. I've also grouped each logical set of changes into individual commits so you can see which corrections and improvements are already included. Please let me know what to do with each "TODO" item.

jtavenner-dev:
@akohlmey Thanks for starting this branch! I've started looking to compare the previous changes, this helps to focus on where specific modifications need to be made. I'll work on addressing the changes and memory access issues and ping for an update here once I've resolved that.

akohlmey:
@jtavenner-dev don't hesitate to ask, if you have any questions about the TODO items or if you need assistance.

jtavenner-dev:
@akohlmey Thanks for outlining the TODO, has been helpful for working through the memory leak problems. I have pushed an updated version with my edits to my lammps fork, with branch also named 'roll-back-neighbor-swap'. I don't believe I have permission to push here since it's a pull request from your branch.

There is at least one remaining problem with the MPI_Allreduce command in the pick_i_swap_atom. I believe I need to separate the send and receive buffer commands, since one is acting as a send and one as a receive. Still working to figure out what the proper commands for that should be.

akohlmey:
> I believe I need to separate the send and receive buffer commands, since one is acting as a send and one as a receive. Still working to figure out what the proper commands for that should be.

If you just want to communicate id_center, MPI_Bcast() would do the trick.

akohlmey:
@jtavenner-dev I pulled your branch, removed the "fixed" comments, merged with the latest develop from upstream and posted it here again.

akohlmey:
@jtavenner-dev is there anything I can help you with so that we get this pull request over the finishing line?

jtavenner-dev:
@akohlmey can you check the memory leak behavior with the changes updated to my branch? I believe I have resolved the access to the invalid address in the original MPI_Allreduce, but I am having trouble verifying the memory behavor with valgrind, as it seems to still record some memory loss in some blocks. I am not very well versed in reading the valgrind output regarding memory performance. Not sure if this is still related to any issues in the new fix.

Thanks for all the help getting this sorted.

akohlmey:
@jtavenner-dev This is really curious. I think I am running into a so-called Heisenbug here.
When I run the updated code, I get segmentation faults and valgrind is complaining about wild memory access. Upon inspection of the resulting core dump with a debugger I see that "i" is a large number, much larger than the number of atoms. Thus accessing atom->type[i] crashes.

*But* if I add a print statement, or something else that accesses the "i" variable, there is no more problem. So I am adding a couple of sanity checks to get a clean error exit of LAMMPS if one of the indices is off. This way we have a specific error message instead of a nondescript core dump.

Other than that, I don't see any memory access issues with the two checks in place.
So this seems to be ready now.

Using valgrind can be tricky these days, since many compiler runtimes have false positives.
That is why we provide additional "suppressions" in the tools/valgrind folder. I copy them to $HOME/.local/share/valgrind/ and then have a .valgrindrc file with:
```
--leak-check=full
--show-leak-kinds=all
--track-origins=yes
--suppressions=/home/akohlmey/.local/share/valgrind/FlexiBLAS.supp
--suppressions=/home/akohlmey/.local/share/valgrind/GTest.supp
--suppressions=/home/akohlmey/.local/share/valgrind/Kokkos.supp
--suppressions=/home/akohlmey/.local/share/valgrind/LLVM.supp
--suppressions=/home/akohlmey/.local/share/valgrind/MPICH.supp
--suppressions=/home/akohlmey/.local/share/valgrind/OpenMP.supp
--suppressions=/home/akohlmey/.local/share/valgrind/OpenMPI.supp
--suppressions=/home/akohlmey/.local/share/valgrind/Python3.supp
--suppressions=/home/akohlmey/.local/share/valgrind/libc.supp
--suppressions=/home/akohlmey/.local/share/valgrind/readline.supp
```
That keeps the bogus messages at a minimum. Although, the suppression files need regular updating for new compilers and new false positives. You still always have to check whether LAMMPS allocated the memory that is apparently leaked, or some runtime library (for OpenMP, OpenCL, LLVM, etc.)

src/MC/fix_neighbor_swap.cpp
```diff
@@ -0,0 +1,949 @@
...
+      if (qmax != qmin)
+        error->all(FLERR, Error::NOLASTLINE, "All atoms of a swapped type must have same charge.");
+    }
+  }
+
+  memory->create(sqrt_mass_ratio, atom->ntypes + 1, atom->ntypes + 1,
+                 "neighbor/swap:sqrt_mass_ratio");
+  for (int itype = 1; itype <= atom->ntypes; itype++)
+    for (int jtype = 1; jtype <= atom->ntypes; jtype++)
+      sqrt_mass_ratio[itype][jtype] = sqrt(atom->mass[itype] / atom->mass[jtype]);
```
jtclemm: There should be an error if atoms have an `rmass` instead of a type mass.

Edit: or a warning like in fix atom/swap.

jtavenner-dev:
@akohlmey Thanks for the review and overview of the valgrind memory. Glad things seem clean now.

I have updated a few items, it is possible the odd behavior with the "i" variable was since it was being re-typed inside of the root statement.

Have also added error handling for when the typed atom masses are requested and per-atom masses exist.

akohlmey:
> I have updated a few items, it is possible the odd behavior with the "i" variable was since it was being re-typed inside of the root statement.

I don't understand what you mean by that. Can you quote an example?

akohlmey:
> Have also added error handling for when the typed atom masses are requested and per-atom masses exist.

Actually, I am currently adapting your code for checking that the charges are consistent with types for this, so that per-atom masses can be used with your fix.

jtavenner-dev:
"i" was being assigned as "int i = . . . " both at the beginning of the pick_i_swap_atom() routine and inside of the if statement, rather than the one inside the if statement being "i = . . . " since the typing was already set.

The handling of per-atom masses will require communicating the atom masses once atoms are selected. My change merely made it so the ke normalization would be incompatible, but the fix would otherwise work and not require sending the mass info. Fine to modify if you can readily add the necessary communication.

akohlmey:
> "i" was being assigned as "int i = . . . " both at the beginning of the pick_i_swap_atom() routine and inside of the if statement, rather than the one inside the if statement being "i = . . . " since the typing was already set.
> 
> The handling of per-atom masses will require communicating the atom masses once atoms are selected. My change merely made it so the ke normalization would be incompatible, but the fix would otherwise work and not require sending the mass info. Fine to modify if you can readily add the necessary communication.

The way I read the code, in the case of per-atom masses, I can use the new mtype[] array just like atom->mass and in the same was as is qtype[] for charges. I just need to follow the steps made for charges.

jtavenner-dev:
> The way I read the code, in the case of per-atom masses, I can use the new mtype[] array just like atom->mass and in the same was as is qtype[] for charges. I just need to follow the steps made for charges.

That will work assuming all atoms of a type still have the same mass despite being set to a per-atom mass type. The qtype handling here likewise requires all atoms of the same type to have the same charge, which is checked in the qtype initialization. Since the qtypes are then referenced according to the atom types during swapping.

src/MC/fix_neighbor_swap.cpp
akohlmey: @jtclemm good catch! thanks. I just added code that makes this fix handle per-atom-masses, too.

jtclemm:
@akohlmey does it make sense to port the changes to rmass to fix atom/swap? It should have the same restriction but currently issues a warning.

akohlmey:
> @akohlmey does it make sense to port the changes to rmass to fix atom/swap? It should have the same restriction but currently issues a warning.

I can have a look. Depends on how this is handled. In general, we want to be compatible with per-atom masses. I've been using it for simulations on isotope effects, where you have the same atom type (and thus potential) but different masses. In principle, one just has to follow what is done for charges.

akohlmey:
> @akohlmey does it make sense to port the changes to rmass to fix atom/swap? It should have the same restriction but currently issues a warning.

@jtclemm done in PR #4653

akohlmey:
@jtavenner-dev this should handle per-atom masses now, too.

jtavenner-dev:
> @jtavenner-dev this should handle per-atom masses now, too.

Great, thanks for adding that handling for the per-atom mass cases. I wasn't familiar with the mtype as I haven't used the per-atom mass settings, it makes sense that it functions mostly the same as qtype.

jtclemm (approved):

PR #4618, opened 2025-06-09
Feat: add new keyword `gauss_fflag` for fix electrode/conp

ChiahsinChu:
**Summary**

Currently, the Gaussian correction in the real space (by `gausscorr` method in `fix_electrode_conp.*`) is always appiled in fix electrode/conp. While it is reasonable for common calculation, I would like to make the `fflag` for `gausscorr` as a user-defined variable, which is always set as `true` now. It will be useful if the users are interested in charge distribution on-the-fly but want to use other force fields for short range interactions (e.g., machine learning potentials).  

**Related Issue(s)**

**Author(s)**

Jia-Xin Zhu (Xiamen University), jiaxinzhu@stu.xmu.edu.cn/xmuchiahsin@gmail.com

**Licensing**

**Backward Compatibility**

No.

**Implementation Notes**

I set a new user-defined keyword `gauss_fflag` to specifiy the `fflag` in `gausscorr `.
Aiming for calculation WITH Gaussian correction as before, no changes are required for the input files. 

**Post Submission Checklist**

### DIFF
--- a/src/ELECTRODE/fix_electrode_conp.cpp
+++ b/src/ELECTRODE/fix_electrode_conp.cpp
@@ -99,6 +99,7 @@ FixElectrodeConp::FixElectrodeConp(LAMMPS *lmp, int narg, char **arg) :
   write_inv = write_mat = write_vec = read_inv = read_mat = false;
   symm = false;
   ffield = false;
+  gauss_fflag = true;
   thermo_time = 0.;
 
   top_group = 0;
@@ -238,6 +239,8 @@ FixElectrodeConp::FixElectrodeConp(LAMMPS *lmp, int narg, char **arg) :
       symm = utils::logical(FLERR, arg[++iarg], false, lmp);
     } else if ((strcmp(arg[iarg], "ffield") == 0)) {
       ffield = utils::logical(FLERR, arg[++iarg], false, lmp);
+    } else if ((strcmp(arg[iarg], "gauss_fflag") == 0)) {
+      gauss_fflag = utils::logical(FLERR, arg[++iarg], false, lmp);
     } else {
       error->all(FLERR, "Unknown keyword {} for fix {} command", arg[iarg], style);
     }
--- a/src/ELECTRODE/fix_electrode_conp.cpp
+++ b/src/ELECTRODE/fix_electrode_conp.cpp
@@ -652,7 +652,7 @@ void FixElectrodeConp::setup_pre_reverse(int eflag, int vflag)
                    style);
   // correct forces for initial timestep
   ev_init(eflag, vflag);
-  gausscorr(eflag, vflag, true);
+  gausscorr(eflag, vflag, gauss_fflag);
   self_energy(eflag);
 }
 
@@ -783,7 +783,7 @@ void FixElectrodeConp::pre_force(int)
 void FixElectrodeConp::pre_reverse(int eflag, int vflag)
 {
   ev_init(eflag, vflag);
-  gausscorr(eflag, vflag, true);
+  gausscorr(eflag, vflag, gauss_fflag);
   self_energy(eflag);
 }
 
--- a/src/ELECTRODE/fix_electrode_conp.h
+++ b/src/ELECTRODE/fix_electrode_conp.h
@@ -98,6 +98,7 @@ class FixElectrodeConp : public Fix {
   double thermo_temp, thermo_time;                       // used by electrode/thermo only
   int thermo_init;                                       // initializer for rng in electrode/thermo
   bool ffield;                                           // possibly tweak electrode/conq's version
+  bool gauss_fflag;                                      // flag for modifying forces in gausscorr
   std::string fixname;    // used by electrode/ffield to set up internal efield
   bool intelflag;
   inline virtual void intel_pack_buffers() {}


akohlmey:
> It will be useful if the users are interested in charge distribution on-the-fly but want to use other force fields for short range interactions (e.g., machine learning potentials).

Sorry, but I have trouble following that logic. In order to correctly apply a KSpace solver, you need to apply the corresponding damping function on the real space coulomb interactions? How do you propose to do that with a machine learning potential?

The only way that I can think of would require to project point charges on the training set, subtract out the *full* coulomb interaction based on those point charges from the force training, but also train the potential to not only only predict forces but also partial charges. Those would then be used with a regular hybrid/overlay pair style using coul/long. That approach would not require your changes.

ChiahsinChu:
> How do you propose to do that with a machine learning potential?

I want to use the charge distributions generated via polarizable electrode (i.e., fix electrode/conp & kspace) as the input for my machine learning potentials.

> Those would then be used with a regular hybrid/overlay pair style using coul/long. That approach would not require your changes.

The point is that I would not include the real space part, neither in coul/long nor in gausscorr of fix conp. For the coul/long I can use hybrid/overlay and set its factor as zero. But for the gausscorr there is no way to remove its contributions on forces, as far as I understood.

srtee:
Hi @akohlmey and @ChiahsinChu ,

Let me give this a closer look by the end of the week but some quick clarifications:

@akohlmey if it helps, the "KSpace Gaussian" is different from the "ELECTRODE Gaussian". The latter is a much more short-ranged charge structure, which primarily gives the dynamic electrode charges a finite self-energy to allow for charge equilibration. The "ELECTRODE Gaussian" is separate from the KSpace Gaussian and since it is usually much shorter-ranged it does not generally perturb any long-ranged electrostatics.

@ChiahsinChu the ELECTRODE team has been refactoring the code base to shift this Gaussian term from `fix electrode/*` out to the pair style, for broader reasons. While it was acceptable at the start for the original `fix conp` to "hard-code" the Gaussian modifications as a `post_force` routine, as the usage broadens and different varieties of constant potential models emerge, it is impractical to maintain more and more different short-range electrostatic modifications under `fix electrode/*`. In principle such modifications should be coded as separate pair styles, not in the fix; in practice there is a small performance penalty and lots of possible bugs and undefined behaviours in delegating a "pair responsibility" to the fix like that.

The main work in doing this is: if short-range modification is the responsibility of the pair style then so is the reporting of the electrostatic potential, since keeping energy, force, and potential calculations consistent within each pair style is the most logical way to do things. So we are at work developing the necessary code. Let me have a chat with the rest of the team and see whether you could participate in code development and testing together with us to align our coding work.

ChiahsinChu:
> the ELECTRODE team has been refactoring the code base to shift this Gaussian term from `fix electrode/*` out to the pair style, for broader reasons.

Sounds good. Then I would expect there would be more flexible to defined short-range part in the polarizable electrode method. 

> Let me have a chat with the rest of the team and see whether you could participate in code development and testing together with us to align our coding work.

I'd like to!

ludwig-ahrens:
@ChiahsinChu You can access the developer version of the refactored code [here](https://github.com/robeme/lammps/blob/electrode/). There is an option to use a pair style with Gaussian corrections instead of adding the terms in the fix, as explained by Shern. Using the new `pair` keyword, no eta value will be set for the fix and all corrections for the forces and the conp algorithm will be added by the pair style. [Here](https://github.com/robeme/lammps/blob/electrode/examples/PACKAGES/electrode/graph-il/in.pair) is an example. The last parameter for `pair_coeff` sets the reciprocal width eta, with NULL for point charges. Using point charges for electrode atoms is not recommended but possible. For details, see our recent publication on charge equilibration [Savvidi et al.](https://doi.org/10.1063/5.0255823).

If I understand correctly, you want to turn off Coulomb forces in LAMMPS and calculate all forces with your machine learning potential. This should work. However, if you want to keep long-range interactions as it is done in the 4th generation NNPs by Behler and coworkers, things are more complicated. In particular there is a correction to the forces, dU/dq dq/dr, which is not implemented in the electrode package (cf. [Ko et al.](https://doi.org/10.1038/s41467-020-20427-2)).

ChiahsinChu:
> You can access the developer version of the refactored code 

Thanks for sharing. It is really useful! 

> If I understand correctly, you want to turn off Coulomb forces in LAMMPS and calculate all forces with your machine learning potential. 

Yes. I think I can use `hybrid/overlay` for coul/long/guass (with factor of 0.0) and mlp pair (with factor of 1.0).

> However, if you want to keep long-range interactions as it is done in the 4th generation NNPs by Behler and coworkers, things are more complicated. In particular there is a correction to the forces, dU/dq dq/dr, which is not implemented in the electrode package (cf. [Ko et al.](https://doi.org/10.1038/s41467-020-20427-2)).

I agree it is not so easy to implement 4th-HDNNP. Currently we just plan to use the polarizable electrode as a resonable boundary for electrolyte (see [here](https://arxiv.org/abs/2407.17740)), which is an approximation but work quite well (at least for planar electrodes).

akohlmey:
If this kind of flag is still needed, it should be added to the development version of the ELECTRODE package and then submitted to LAMMPS with the next update. Thus this pull request is not needed and therefore I will close it now.

PR #4630, opened 2025-06-17

### LINKED ISSUE #4632: [BUG] pair.cpp Out-of-bounds array access in cvatom allocation/initialization
In the Pair::ev_setup function, the cvatom array is allocated with a second dimension size of 9. However, a subsequent loop attempts to access cvatom[i][9]:

cvatom[i][9] = 0.0;

akohlmey:
**Summary**

This pull request combines multiple small changes that do not warrant a pull request on their own.

**Related Issue(s)**

Fixes #4632 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- fix bug in Fortran module where reallocation of global fix callback list did not happen as expected (reported by @rbberger)
- avoid copies of strings when setting error messages in the library interface
- error out in compute msd if the number of atoms in the compute group changes by other means than a dynamic group
- fix out-of-bound access in Pair class when initializing cvatom reported in Issue #4632. Also fix the same issue `angle.cpp`, `dihedral.cpp`, and `improper.cpp`
- replace rounding with `static_cast<int>(val + 0.5)` with C++11's `std::lround()`
- remove workaround of running sphinx-build twice for linking Fortran functions in manual since it has no effect anymore. Will need to solve the problem in the package itself.
- silence compiler warnings
- run force-style unit tests with Kokkos/Serial, if it is available and Kokkos/OpenMP is not
- correct reading shake and special sections of a molecule file in native format that were ignoring the atom-id and thus contradicting the documentation.
- remove some remnants of Python 2 support from LAMMPS python module
- update LAMMPS python module to follow suggestions form pylint (or add comments to ignore them)
- small update to the LAMMPS python module wrapper for NumPy to replace discouraged code with recommended version
- parsing of JSON is only attempted for molecule files with the ".json" extension
- add support for shake, special, and body sections to JSON format molecule data.
- stricter checking of header items in native molecule file format
- correct creating of body particles from molecule files with `create_atoms`, `fix deposit`, and `fix pour`
- body particles require the Masses section to provide the total mass of the body

**Post Submission Checklist**

### DIFF
--- a/python/lammps/data.py
+++ b/python/lammps/data.py
@@ -91,3 +91,8 @@ class NeighList:
         return numneigh, neighbors
 
     return -1, None
+
+# Local Variables:
+# fill-column: 100
+# python-indent-offset: 2
+# End:
--- a/python/lammps/data.py
+++ b/python/lammps/data.py
@@ -13,7 +13,7 @@
 
 """
 Data structures for LAMMPS Python module
-Written by Richard Berger <richard.berger@temple.edu>
+Written by Richard Berger <richard.berger@outlook.com>
 """
 
 class NeighList:
--- a/python/lammps/data.py
+++ b/python/lammps/data.py
@@ -11,82 +11,83 @@
 #   See the README file in the top-level LAMMPS directory.
 # -------------------------------------------------------------------------
 
-################################################################################
-# LAMMPS data structures
-# Written by Richard Berger <richard.berger@temple.edu>
-################################################################################
+"""
+Data structures for LAMMPS Python module
+Written by Richard Berger <richard.berger@temple.edu>
+"""
 
-class NeighList(object):
-    """This is a wrapper class that exposes the contents of a neighbor list.
+class NeighList:
+  """This is a wrapper class that exposes the contents of a neighbor list.
 
-    It can be used like a regular Python list. Each element is a tuple of:
+  It can be used like a regular Python list. Each element is a tuple of:
 
-    * the atom local index
-    * its number of neighbors
-    * and a pointer to an c_int array containing local atom indices of its
-      neighbors
+  * the atom local index
+  * its number of neighbors
+  * and a pointer to an c_int array containing local atom indices of its
+    neighbors
 
-    Internally it uses the lower-level LAMMPS C-library interface.
+  Internally it uses the lower-level LAMMPS C-library interface.
 
-    :param lmp: reference to instance of :py:class:`lammps`
-    :type  lmp: lammps
-    :param idx: neighbor list index
-    :type  idx: int
+  :param lmp: reference to instance of :py:class:`lammps`
+  :type  lmp: lammps
+  :param idx: neighbor list index
+  :type  idx: int
+  """
+  def __init__(self, lmp, idx):
+    self.lmp = lmp
+    self.idx = idx
+
+  def __str__(self):
+    # pylint: disable=C0209
+    return "Neighbor List ({} atoms)".format(self.size)
+
+  def __repr__(self):
+    return self.__str__()
+
+  @property
+  def size(self):
     """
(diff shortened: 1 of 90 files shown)


navlalli:
@akohlmey Could you please rename qeqr/reax to qeq/rel/reax?

https://github.com/lammps/lammps/blob/53fec5563cf00bcdda454ccd22c460a39b368228/src/REAXFF/fix_qtpie_reaxff.cpp#L496

akohlmey:
> @akohlmey Could you please rename qeqr/reax to qeq/rel/reax?
> 
> https://github.com/lammps/lammps/blob/53fec5563cf00bcdda454ccd22c460a39b368228/src/REAXFF/fix_qtpie_reaxff.cpp#L496

done.

akohlmey:
@sjplimp @jrgissing After testing and debugging molecule files in native and JSON format for an extended period, specifically on features that were missing pieces and thus were causing unexpected errors, I have to give up my plan that molecule files can have arbitrary names and we detect the format automatically during parsing. Since the JSON format is very strict, it happens too easily with manual editing of the file to introduce typos and then the error messages can be very confusing.

The way out is to do what @sjplimp has been asking for from the beginning: let the filename extension decide on the *intended* format. Thus I want to implement that the "native" format *requires* an extension ".mol" and the JSON format files the extension ".json". Any other extension would trigger an "unknown molecule file extension" error. That would keep the door open for any future addition of yet another file format in a clean way.

Any objections or different suggestion? Please let me know.

jrgissing:
I would suggest detecting .json and assuming all other files are native, for backward compatibility

jrgissing:
Here is routine that could be added to platform.cpp:

bool platform::has_extension(const std::string &file, const std::string &extension)
{
  std::size_t dot = file.find_last_of('.');
  if (dot != std::string::npos) {
    const std::string ext = file.substr(dot + 1);
    if (ext == extension) return true;
  }
  return false;
}

akohlmey:
> Here is routine that could be added to platform.cpp:

`if (utils::strmatch(filename, "\\.json$"))` works just as well. :wink:

python/lammps/data.py
```diff
@@ -11,82 +11,83 @@
 #   See the README file in the top-level LAMMPS directory.
 # -------------------------------------------------------------------------
 
-################################################################################
-# LAMMPS data structures
-# Written by Richard Berger <richard.berger@temple.edu>
-################################################################################
+"""
+Data structures for LAMMPS Python module
+Written by Richard Berger <richard.berger@temple.edu>
```
rbberger: If you're touching these lines, might as well update this to `richard.berger@outlook.com`

akohlmey: done

python/lammps/numpy_wrapper.py
```diff
@@ -464,51 +456,48 @@ def darray(self, raw_ptr, nelem, dim=1):
       if dim > 1:
         a.shape = (nelem, dim)
       else:
-        a.shape = (nelem)
+        a.shape = nelem
```
rbberger: Not sure about that one. It might have to be `a.shape = (nelem,)`

akohlmey: I've looked at the documentation and it says that using `a.shape` as a setter is discouraged and one should use `np.reshape(a,...)` and use `a.shape` only as a getter. For reshape a single integer is permitted, for shape there is no information. I've tested on an ubuntu 18.04LTS container and it accepts the single integer, same as on my Fedora 42 machine.

In conclusion, the current change is acceptable but to future-proof it, I should update this to use `np.reshape(a, ...) `

Bibobu (approved):
I approve this PR.

PR #4636, opened 2025-06-26
Add support for inter-/intra-molecular bonding to fix bond/create

wapisani:
**Summary**

I added support for inter- and intra-molecular bonding to fix bond/create by copying the relevant code from fix bond/react and modifying it appropriately.

**Related Issue(s)**

N/A

**Author(s)**

William A. Pisani, PhD
US Army ERDC Environmental Laboratory

**Licensing**

**Backward Compatibility**

Backward compatibility should not be affected since it is an optional keyword to fix bond/create. 

**Implementation Notes**

The relevant fix bond/react code was copied over to fix_bond_create.cpp and fix_bond_create.h. The copied code was then modified to match fix bond/create's error messages.

To check that the molecule keyword was working as expected, I ran three different simulations (one with molecule inter, one with molecule intra, and one without the molecule keyword) with fix bond/create and visualized the data files in OVITO Pro with the cluster analysis modifier. The data file read into the simulation was comprised of 400 molecules evenly split into two types of monomers. These monomers were composed of 17 and 19 atoms each, respectively. The simulations were run for enough time that bond formation occurred in sufficient amounts.

For the simulation with molecule inter, correctness was verified if there were no clusters with 17 or 19 atoms that had bonded to itself. For the simulation with molecule intra, correctness was verified if there _were_ clusters of 17 or 19 atoms that had bonded to itself. For the simulation without the molecule keyword, correctness was verified if the simulation proceeded as usual with the results of molecule intra and inter, but no error messages.

**Post Submission Checklist**

Three example input decks and a densified monomer test system as a gzipped LAMMPS data file are attached as a zip file. The fix bond/create commands with a 0 in the name (bond0a, bond0b, etc.) specify bonding between the two kinds of monomers while the commands with a 1 and a 2 in the name specify bonding between the same kind of monomers.
[bond_create_test_scripts_data_file.zip](https://github.com/user-attachments/files/20933963/bond_create_test_scripts_data_file.zip)

### DIFF
--- a/doc/src/fix_bond_create.rst
+++ b/doc/src/fix_bond_create.rst
@@ -21,7 +21,7 @@ Syntax
 * Rmin = two atoms separated by less than Rmin can bond (distance units)
 * bondtype = type of created bonds (integer or type label)
 * zero or more keyword/value pairs may be appended to args
-* keyword = *iparam* or *jparam* or *prob* or *atype* or *dtype* or *itype* or *aconstrain*
+* keyword = *iparam* or *jparam* or *prob* or *atype* or *dtype* or *itype* or *aconstrain* or *molecule*
 
   .. parsed-literal::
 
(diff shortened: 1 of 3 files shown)


megmcca (approved):
looks good, preserves default behavior and adds detailed error descriptions.

Issue #4625, opened 2025-06-13
TODO list for JSON support in LAMMPS


akohlmey:
## Summary

This issue is for collecting ideas and plans for additions and modifications to supporting JSON format files with LAMMPS.

### molecule command:
- [x] add support for "shake" section (@akohlmey)
- [x] add support for "special" section (@akohlmey)
- [x] add support for "body" section (@akohlmey)
- [x] add API to C-library interface to inject molecule templates from string (including Python support) (@akohlmey)
- [ ] move content from `Molecule::from_json()` that is processed twice into a separate function so we can do something similar to `Molecule::read(0); Molecule::allocate(); Molecule::read(1);` (@akohlmey)
- [x] add tests to the `unittest` tree for JSON format (@akohlmey)
- [ ] check if code sharing between native and JSON reader can be increased

### Further ideas:

- Implement a `group2molecule` command to create a molecule template from some subsystem?
- Add a "molecule write" keyword which allows to write out molecule files and thus would also allow to convert between native and JSON format files?

Where else should we add JSON format support? Are there any low-hanging fruit?

akohlmey:
@jrgissing @salamaaya with the latest changes in PR #4630 JSON support in the molecule command should be feature complete (compared to the native format) and (at least lightly) tested.
In the process I found and fixed a bunch of issues with the native format reader (which also needed to be backported to the stable version). The result is - unfortunately - extremely ugly and repetitive, so I am thinking about ways to improve it.

This should serve as a foundation for any new features you would need, but don't expect the implementation to remain this same. Thus, I propose that if you need a new feature, you describe it here and let me handle the implementation (it will be easier for me to navigate my own mess than for anybody else, and I don't want people to get used to bad examples and possibly follow them).

jrgissing:
more useful than group2molecule will be chunk or 'per-atom vector' to molecule. see #4640 for an example of using per-atom vector to write deleted molecules from 'fix reaxff/species' to JSON

akohlmey:
> more useful than group2molecule will be chunk or 'per-atom vector' to molecule. see [#4640](https://github.com/lammps/lammps/pull/4640) for an example of using per-atom vector to write deleted molecules from 'fix reaxff/species' to JSON

I think we are not communicating properly here. Your pull request and my suggestion address very different needs.

[...]

PR #4640, opened 2025-06-28
JSON output for fix reaxff/species delete keyword

jrgissing:
**Summary**

New ‘fix reaxff/species' feature that produces JSON output for ReaxFF that includes atom IDs, types and coords of deleted molecules. ‘JSON mode’ is triggered by detecting the extension of the name of delete species output file (e.g., ‘deleted_molecule.json’).

Thanks to Jacob Breese (OSU) for testing

**Related Issue(s)**

template for general JSON output from run command, so some discussion necessary for format. see attached for example file of deleted water molecules

might be wise to add to the general LAMMPS output howto about JSON per-timestep output, or once it is added for a few commands

related to #4625

**Author(s)**

Jake Gissinger (AFRL, Stevens)

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

[species-del.json](https://github.com/user-attachments/files/23152402/species-del.json)

[example_deleting_water.zip](https://github.com/user-attachments/files/20958879/example_deleting_water.zip)


src/output.cpp
```diff
@@ -628,6 +630,146 @@ void Output::write_restart(bigint ntimestep)
...
+  int tab = 4;
+  indent.resize(json_level*tab, ' ');
+  int json_init = 0;
+
+  // let's first condense all ivec values
+  std::unordered_set<int> unique_ivec(ivec, ivec + (size_t) atom->nlocal);
+  unique_ivec.erase(0);
+  std::vector<int> mivec(unique_ivec.begin(), unique_ivec.end());
+
+  int approxsize = atom->natoms / comm->nprocs;
```
akohlmey: Why not use `atom->nmax`?

src/label_map.h
```diff
@@ -43,10 +43,11 @@ class LabelMap : protected Pointers {
   void read_restart(FILE *fp);
   void write_restart(FILE *);
 
- protected:
```
akohlmey, line 47: Rather than making data members public (as was done in the past), we should follow modern C++ guidelines for new code and add accessor functions.

src/output.cpp
```diff
@@ -628,6 +630,146 @@ void Output::write_restart(bigint ntimestep)
...
+
+    const int nfields = 3;
+    int blocklengths[nfields] = {1, 1, 3};
+    MPI_Aint offsets[nfields];
+    offsets[0] = offsetof(Particle, tag);
+    offsets[1] = offsetof(Particle, type);
+    offsets[2] = offsetof(Particle, x);
+    MPI_Datatype types[nfields] = {MPI_INT, MPI_INT, MPI_DOUBLE};
+
+    MPI_Type_create_struct(nfields, blocklengths, offsets, types, &ParticleStructType);
```
akohlmey: This needs to be implemented in the STUBS library for serial compilation.

jrgissing:
I propose to change my current suggested "deleted_molecules" key to simply "molecules", perhaps adding a description subkey to identify them as deleted. The idea is for it to be the same all fixes that will output molecules, to make parsing more universal. in particular, we would like to visualize this output, and visualization programs cannot be expected to parse different keys for each fix

jrgissing:
note: I changed the "deleted_molecules" key to simply "molecules". the "subformat" key identifies the molecules as deleted.

doc/src/fix_reaxff_species.rst
```diff
@@ -140,28 +140,88 @@ be either a list of specific chemical formulae or a range of molecular
...
+deleted molecules in the style of the :doc:`JSON molecule file <molecule>`,
+where more discussion of JSON schema can be found. Here is an example of a
+JSON output file from a simulation during which one water molecule was
+deleted on the first timestep:
+
+.. code-block:: json
+
+   {
+       "application": "LAMMPS",
+       "format": "output",
```
akohlmey: @jrgissing to make this more consistent with the rest of LAMMPS and also be more future-proof, I suggest we call this format "dump", and the "subformat" (or perhaps "style"?) should be called something like "by-molecule" or "by-atom" to indicate how the data is grouped. The "fix reaxff/species: delete keyword" string would be more fitting for a "title". Then I would rather have "timesteps" instead of "run_output", so that "run_output": [ {"timestep": 1, "molecules" [...]},...] could become "timesteps": [{10, [...]}, 20, [...]},...].

The "by-atom" format would then be more like the conventional dump styles, only that rather than having one big blob per timestep, we would have sections like "type", "coord", "charge", "molecule", "force", "vel", "mass", ... 
There is already a by column version with dump style yaml.

akohlmey: The by-atom format could become something of an alternative to a data file, if topology data and type-label mapping would be included.

jrgissing: I agree with all these suggestions. For molecule vs atom, what about:
"grouping": "molecule"

akohlmey: hmmm... in the "by-atom" format, the "grouping" of information would technically speaking be by property.  So in that sense a keyword like "order" or "ordering" would be better. But to be perfectly frank, I am not 100% comfortable with either of the choices that you or I have mentioned (including "sub-format").

Now that I look at this for a second time, I am beginning to like something like "style": "atom" and "style": "molecule" more. *Definitely* without the "by-"; that was a bad idea.  It is cleaner to just have a single word for a keyword. I suggest you pick (for now) what *you* like best for as long as it is understood that it is not yet the final word and I can try, if I can come up with something I like 100% (and can convince you of the same).

Another idea (primarily for a future dump style json) would be to have a flag requesting to output only fields that are *different* from the previous step. That could be a big space saver for simulations with partially frozen content or when outputting topology data (like changing bonds/angles/dihedrals) for some atoms. That kind of format would need another keyword to indicate whether unchanged data is skipped or not. 

We would need to write down the final format(s) in a `tools/json/dump-schema.json` file. That can be done in multiple iterations if needed. we should just converge for the next feature release (planned for mid-to-late December if development remains as slow as the last weeks and I am spending as much time on writing papers as I did since the last release).  When I was adding JSON support to the molecule command, I was writing both at the same time and used the JSON schema to make sure my C++ code was doing what was intended and used the JSON schema as a form of documentation of the format specifics before writing the actual docs.

jrgissing: what about:
  "format": {
    "type": "dump",
    "output": "molecules"
  }

akohlmey: No, we should make the global keywords have the same layout for the different formats.
So it would be better to have:
```json
"format": "dump",
"output": "molecules",
```
and:
```json
"format": "dump",
"output": "atoms",
```

or use "style" instead of "output"... ;-)

jrgissing: OK, made those updates. Note that I would vote for leaving the 'timestep' key inside the 'timesteps' object for human readability and searchability. I should be able to get to the schema this weekend.

jrgissing:
@akohlmey ok, this is ready for another look

doc/src/fix_reaxff_species.rst
```diff
@@ -140,28 +140,90 @@ be either a list of specific chemical formulae or a range of molecular
...
+deletes molecules that are identified as "CO" and "CO2" in the species
+output file.  The *masslimit* keyword permits deletion of molecules with
+molecular weights between *massmin* and *massmax*.  The *specieslist* and
+*masslimit* keywords cannot both be used in the same *reaxff/species* fix.
+The *delete_rate_limit* keyword can enforce an upper limit on the overall
+rate of molecule deletion. The number of deletion occurrences is limited to
+Nlimit within an interval of Nsteps timesteps. Nlimit can be specified with
+an equal-style :doc:`variable <variable>`. When using the
+*delete_rate_limit* keyword, no deletions are permitted to occur within the
+first Nsteps timesteps of the first run (after reading a either a data or
```
jtclemm: Extra word? "after reading *a* either a data or"

doc/src/fix_reaxff_species.rst
```diff
@@ -140,28 +140,90 @@ be either a list of specific chemical formulae or a range of molecular
...
+restart file).
+
+The *delete* keyword can output information about the deleted molecules in
+either legacy format or JSON format. The latter is activated when the
+*filedel* argument has a '.json' extension.  The legacy format lists how
+many of each species is deleted, while the JSON format provides the atom ID,
```
jtclemm: Are deleted, I think

jtclemm (approved):
I'm not the best reviewer since I'm not familiar with JSON nor am I super familiar with the utilities in LAMMPS for file writing, but everything else looked good to me. Just spotted a few minor proofreading edits.

tools/json/dump-molecules-schema.json
```diff
@@ -0,0 +1,121 @@
...
+            "const": "LAMMPS"
+        },
+        "units": {
+            "enum": ["lj", "real", "metal", "si", "cgs", "electron", "micro", "nano" ]
+        },
+        "format": {
+            "type": "string",
+            "const": "dump"
+        },
+        "style": {
```
akohlmey, line 20: Thanks for providing the JSON schema file. I think we should change the "style" entry to
```JSON
"style": {
    "enum": ["molecules"]
},
```
That would make it easier to extend this for other JSON dump styles and we would only need one schema file.

tools/json/dump-molecules-schema.json
```diff
@@ -0,0 +1,121 @@
...
+        },
+        "title": {"type": "string"},
+        "schema": {"type": "string"},
+        "timesteps": {
+            "type": "array",
+            "items": {
+                "type": "object",
+                "required": ["timestep", "molecules"],
+                "properties": {
+                    "timestep": {
```
akohlmey, line 36: I understand your desire for readability, but I have concerns about how we can efficiently access the parsed data. If the timestep data is just a list of integer + object, we can easily (and quickly) access a timestep by its value and extract the list of all included timesteps. By making it a "timestep": value / "molecules": list, you would have to search through *all* steps to find a specific one, if I am not mistaken. We could require starting at step 0 and equal distancing, but that would be a limitation, that was not so long ago removed from (most of) the regular dump styles.

tools/json/dump-molecules-schema.json
```diff
@@ -0,0 +1,121 @@
...
+                      "type": "integer",
+                      "minimum": 0
+                    },
+                    "molecules": {
```
akohlmey, line 40: It may not be needed and have much meaning for ReaxFF output, but should the molecules have molecule IDs (for example when outputting a trajectory by molecule from a future dump style json)?

tools/json/dump-molecules-schema.json
```diff
@@ -0,0 +1,121 @@
...
+                                                    {"type": "number"},
+                                                    {"type": "number"}
+                                                ],
+                                                "items": false
+                                            }
+                                        }
+                                    }
+
+                                },
+                                "bonds": {
```
akohlmey, line 91: This probably should also include angles, dihedrals, and impropers.
And how about other properties supported by molecule files?
Or Should we wait until we have a proper use case and then declare a revision 2 of the format?

tools/json/dump-molecules-schema.json
jrgissing: Are you suspecting that additional keys within 'style' will be necessary? Do you have any examples in mind? I am already suspecting that certain fix dumps will need to include fix-specific metadata on a per-molecule basis.

tools/json/dump-molecules-schema.json
jrgissing: ok

tools/json/dump-molecules-schema.json
jrgissing: This may be useful for future fix-bond dumps, but ReaxFF uses atom_style charge, so I don't you can assign molecule IDs in the first place, nor hope to keep consistent values during reactions. Seems like a v2 addition

tools/json/dump-molecules-schema.json
jrgissing: I think angles, dihedrals, impropers could be a possible v2 thing. I added bonds because fixed-bond applications would certainly need to include those.

jrgissing: or, perhaps a better option is to simply use a $ref to our LAMMPS molecule JSON schema, to clearly support the full (and potentially changing) molecule spec. We should just make sure that per-molecule metadata could be associated with each molecule.

tools/json/dump-molecules-schema.json
akohlmey: I would prefer to have one single schema describing a general JSON dump format. As mentioned in an earlier discussion, I can imagine that we also implement a dump style json and that could have the data organized similar to dump style custom or dump style yaml in columns and listed by atom ID (aka "style": "atom") or it could be ordered by property (-> "style": "property"), i.e. each section is organized more like a molecule JSON file or a data file with "coord", "type", "molecule", "charge", "velocity", "force", etc. sections and each containing data for all atoms.

tools/json/dump-molecules-schema.json
akohlmey: fix reaxff/species computes the clusterID array which could be used in place of a molecule ID. Thuse will change, when there are reactions, of course, but without they are consistent, since the order of the clusterIDs is derived from the smallest atom-ID in the cluster. We have at times recommended people to overwrite the molecule IDs with that information.

akohlmey: At any rate, the molecule ID would be an optional field and thus can be omitted, if so desired or for situations where the information is not available.

tools/json/dump-molecules-schema.json
akohlmey: Interesting idea. In that case there could be a Molecule class instance that is cleared and filled with available data for each molecule, then you can call `Molecule::to_json()` and with the thus created json class instance output the resulting serialized data structure using `json::dump(2)` and only create the JSON output manually for the outer loops. That would automatically apply the molecule JSON schema, but would have the downside of lots of redundant output.

jrgissing: I was only referring to modifying the schema. So far, I like the idea of having full control of the output. Either way, I propose to modify the 'molecules' object to put 'LAMMPS molecule objects' inside a data key:
``` 
           "molecules": [
                {
                    "data": {

                        "types": {
                            ...
                        },
                        "coords": {
                            ...
                        }
                    }
                },
```

JacksonBurns:
Hi @jrgissing this is a follow up to your talk at AIChE 2025, where you requested audience input about LAMMPS/JSON integration.

I'm far from an expert user, but I think it might be useful to offer some compatibility with rdkit's Molecule JSON format, [the docs for which are here](https://www.rdkit.org/docs/RDKit_Book.html#rdkitjson-format). Not sure if your current spec is compatible already, or if some converter would be needed, but just putting it out there. Great talk!

jrgissing:
> I'm far from an expert user, but I think it might be useful to offer some compatibility with rdkit's Molecule JSON format, [the docs for which are here](https://www.rdkit.org/docs/RDKit_Book.html#rdkitjson-format). Not sure if your current spec is compatible already, or if some converter would be needed, but just putting it out there. Great talk!

Thanks for the suggestion! Since it is not directly related to this PR, it would help to convert this suggestion into its own Issue

PR #4646, opened 2025-07-03
Reduce Impact of LAMMPS-GUI ChartViewer Performance Regression and Apply clang-tidy Fixes

akohlmey:
**Summary**

This pull request fixes a performance regression in LAMMPS-GUI caused by using smoothing in the Chart Viewer window for trajectories that produce a lot of thermo data (e.g. by running for a long time).

Also a lot of minor programming style improvements and fixes suggested by clang-tidy were applied.

**Related Issue(s)**

Multiple LAMMPS-GUI users on MatSci.org have reported slowdown with long runs.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

To reduce the overhead for live updates of the plots of thermo data, the default setting was changed to only display the "raw" plot by default.

The issue was primarily caused by using `QList<>::append()`. Instead we are now using a copy constructor to make a copy of the raw data and update only the y-component with the filtered data. In the chart the list of points is integrated with `replace()` instead of `clear()` and `append()`. 

Also, as warning note was added to the LAMMPS-GUI howto discussing the negative performance impact of having too frequent and too much thermo data.

**Post Submission Checklist**

**LAMMPS-GUI Binaries**

Pre-compiled binary packages with LAMMPS-GUI created from this branch have been uploaded to https://download.lammps.org/testing/

**Example Input Deck to Reproduce the Slowdown**
```
units           lj
atom_style      atomic

lattice         fcc 0.8442
region          box block 0 2 0 2 0 2
create_box      1 box
create_atoms    1 box
mass            1 1.0

velocity        all create 3.0 2131513 loop geom

pair_style      lj/cut 2.5
pair_coeff      1 1 1.0 1.0
neighbor        0.3 bin
neigh_modify    every 20 delay 0 check no one 100000 page 10000000

# change thermo to 100 or 10000 to see performance impact of frequent thermo output
# and compare the time for "Output" and "Other" for the three run segments.
# also compare to running LAMMPS on the command line.
thermo 1000
fix             1 all nve

run 1000000
run 1000000
run 1000000

```
For old executables, the performance with the default setting of plotting both, raw and smoot, should be quite bad.


doc/src/Howto_lammps_gui.rst
```diff
@@ -357,17 +357,31 @@ property can be shown at a time.  The plots are updated regularly with
...
+means that the smoothing window includes 10 points each to the left and
+the right of the current data point for a total of 21 points and a
+fourth order polynomial is fitted to the data in the window.
+
+.. admonition:: Slowdown of Simulations from Charts Window
+   :class: warning
+
+   Using frequent thermo output or having smoothing enabled during long
+   simulations creating many data points for the chart window can result
+   in a significant slowdown of that simulation as it progresses. This
```
simongravelle: Alternative for first sentence: Using frequent thermo output or having smoothing enabled during long simulations, which create many data points for the chart window, can result in a significant slowdown of the simulation as it progresses.

doc/src/Howto_lammps_gui.rst
```diff
@@ -357,17 +357,31 @@ property can be shown at a time.  The plots are updated regularly with
...
+   can be confirmed when an increasing percentage of the total run time
```
simongravelle:
```suggestion
   slowdown can be confirmed when an increasing percentage of the total run time
```

doc/src/Howto_lammps_gui.rst
```diff
@@ -357,17 +357,31 @@ property can be shown at a time.  The plots are updated regularly with
...
+   is spent in the "Output" or "Other" sections of the "MPI task timing
+   breakdown".  It is thus recommended to use a large enough value as
```
simongravelle:
```suggestion
   breakdown".  It is thus recommended to use a large enough value as an
```

doc/src/Howto_lammps_gui.rst
```diff
@@ -357,17 +357,31 @@ property can be shown at a time.  The plots are updated regularly with
...
+   argument for the :doc:`thermo command <thermo>` and to select
+   plotting only the "Raw" data in the *Charts Window* during such
+   simulations. It is always possible to switch between the different
+   display styles for charts during the simulation or after the
```
simongravelle: either during the simulation or after it has finished.

tools/lammps-gui/lammpsgui.cpp
```diff
@@ -2045,29 +2045,29 @@ void LammpsGui::setup_tutorial(int tutno, const QString &dir, bool purgedir, boo
...
+            case 4:
+                weburl = weburl.arg(tutno).arg("nanosheard-electrolyte");
+                break;
+            case 5:
+                weburl = weburl.arg(tutno).arg("reactive-silicon-dioxide");
+                break;
+            case 6:
+                weburl = weburl.arg(tutno).arg("water-adsorption-in-silica");
+                break;
+            case 7:
```
simongravelle, line 2066: seems like the tutorial 8 is missing from here

simongravelle (commented):
Hi @akohlmey, I only went through the changes in the documentation, but did not yet test the new version of the GUI. I will try to test it tonight.

simongravelle:
OK I tested the linux  (non flatpack) version of the GUI from there https://download.lammps.org/testing/ with a few of the tutorials, and it seems to work perfectly.

simongravelle (approved):

tools/lammps-gui/lammpsgui.cpp
akohlmey: Oops. Good catch. Perhaps you were still working on updating the website when I wrote this.

doc/src/Howto_lammps_gui.rst
akohlmey: Thanks, I went for something slightly different, emphasizing the thermo interval and enabling smoothing are two different factors.

PR #4652, opened 2025-07-09
New FENIX package

Matthew-Whitlock:
**Summary**

Adds online process recovery and memory-based checkpoint/restart support using the [Fenix library](https://www.github.com/sandialabs/fenix.git)

**Related Issue(s)**

**Author(s)**

Matthew Whitlock, Sandia National Labs
Corresponding email: matthew@mwhitlo.dev

**Licensing**

**Backward Compatibility**

Maintains backward compatibility

**Implementation Notes**

Relies on the changes in the feature/cpp branch of Fenix

To catch the exceptions thrown by Fenix upon detecting a failure, the fenix command "steals" the input processing by calling `input->file()` instead of directly returning. This works out fine, but may be fragile to changes to the `input->file()` function.

Since LAMMPS checkpointing uses direct file writes, memory-based checkpointing is implemented in a less-than-ideal manner. We create a tempfile (hopefully memory-backed by the OS), write the checkpoint to that file, copy the file contents to a memory buffer, then store to Fenix (which again moves the data). In an ideal world, LAMMPS would update to using a serializer-based approach for checkpointing. Something like Boost's serialize library or [DARMA's Magistrate](https://github.com/DARMA-tasking/magistrate). Direct file writing could be maintained as an option for backwards compatibility but be considered deprecated. This would be beneficial to the Fenix package, open flexibility for other resilience systems, and probably be more efficient for the current checkpointing system.

Many of the classes which write checkpoint data directly include checks to only write their global data on rank 0. To localize recovery, we want the global checkpoint data on all ranks. We achieve this by temporarily setting `comm->me` to 0. Ideally, classes would have two checkpoint functions, one to be called for serializing global data and one for serializing this rank's local data. The relevant function would only be called on the rank that should write the data, removing the checks from inside the functions and allowing the checkpointing system to decide that. Having a separate function for local data would simplify classes that currently manually merge local data onto rank 0, and would allow different checkpointing systems to break up local data in different formats.

Most recovery in LAMMPS is handled by passing data buffers to the various virtual read_restart functions, but the various force fields instead take a file pointer directly. They also manually broadcast each individual value read from the file on rank zero to all other ranks. We work around this by using the fmemopen function to create a memory-backed file to be read. Ideally, these classes would also recover by taking as input their (pre-broadcast) stored data in a buffer. This would allow us to avoid the POSIX-only fmemopen function and skip the communication during recovery.

**Post Submission Checklist**

- [ ] Suitable tests have been added to the unittest tree. (none)

Integrating a full serialization library would be the best choice for performance/reliability/flexibility, but may be more intrusive than desired. It would be simpler to just change the datatype passed to the read/write functions and then overload the fwrite family of functions for that datatype instead of FILE*. For a basic demo of that, see: https://github.com/Matthew-Whitlock/lammps/tree/serialization. Of course, none of this is a hard requirement for this FENIX package PR.


akohlmey:
@Matthew-Whitlock we should schedule a meeting with (at least) you, @sjplimp and me during or in the day before/after the upcoming LAMMPS user meeting in August. I have not looked to closely, but just your description scares me... a lot. Steve and I will want to better understand what you want to achieve and what the benefit to the users is going to be. You are touching a feature that is very deeply woven into the fabric of LAMMPS, so we are going to be extremely hesitant to make any changes. And if we would want to do then, we would be looking for something clean and generic that can also be used for other purposes and file formats.
But I am putting the carriage before the horse. First we need to understand the motivation and the benefit. For sure none of us will have a look at the details until after the workshop (and the upcoming LAMMPS stable release).

Matthew-Whitlock:
> @Matthew-Whitlock we should schedule a meeting with (at least) you, @sjplimp and me during or in the day before/after the upcoming LAMMPS user meeting in August. I have not looked to closely, but just your description scares me... a lot. Steve and I will want to better understand what you want to achieve and what the benefit to the users is going to be. You are touching a feature that is very deeply woven into the fabric of LAMMPS, so we are going to be extremely hesitant to make any changes. And if we would want to do then, we would be looking for something clean and generic that can also be used for other purposes and file formats. But I am putting the carriage before the horse. First we need to understand the motivation and the benefit. For sure none of us will have a look at the details until after the workshop (and the upcoming LAMMPS stable release).

I'm happy to have that meeting, it would be great to discuss this topic in depth. That said, I may have accidentally made this PR seem scarier than it is. I wrote a lot about changes to LAMMPS that would be nice to have, but none of them are necessary for this package -- the discussion there was more to explain the reasoning for why this package may be incompatible to use alongside some other packages/fixes/etc.

This code is functional as-is, and the vast majority of the work is contained within the optional FENIX package. Our changes to existing LAMMPS files are limited to:
  * A new utility function, which can be made internal to the package but is generic enough that it felt worthwhile to offer it there
  * Adding our checkpoint class as a friend to a few LAMMPS classes
  * ~8 changed lines of code otherwise
    * Of these, you'll likely only be unhappy about saving a copy of the argv pointer in the lammps class. Part of my motivation for opening this PR now is wanting to chat about a better way to handle the problem that this is trying to solve
  * No changes to LAMMPS' behavior at all unless the commands/fixes within are used.

But there's no rush to have this discussion, meeting in August works just as well and I'll chat with my group about gathering a few of us to make the pitch.

Matthew-Whitlock:
I've updated to fix a couple of bugs and rebased on top of the latest changes, this should be good for review again. I marked unit tests as completed, as it doesn't look like optional packages have unit tests. Is the conventional build system still something to check against?

akohlmey:
> as it doesn't look like optional packages have unit tests

Oh yes, but those are testing features/commands/styles not whole packages. Most test programs use available introspection to decide whether to run a specific test or not depending on available pre-requisites. You can see the coverage of the internal testing (does not include the regression tests the check the example inputs) here: https://download.lammps.org/coverage/

> Is the conventional build system still something to check against?

Not anymore since we branched off the 22 July 2025 stable release.
Legacy build support is now optional. You just need a dummy `Install.sh` file like for example `src/RHEO/Install.sh`

Matthew-Whitlock:
Speaking to the benefit of this package, it offers significantly faster checkpoint/restart as it currently stands. Memory-based checkpoints out-perform and out-scale the current file-based checkpoints by a large margin. Compared to using a luster filesystem and 8 nodes/file (found to be optimal for my tests), these memory checkpoints offer over 6x checkpoint speedups at even small scales (100 nodes, 100 ranks per node, 0.5Gb checkpoint data per node).

As work for a separate PR, my team has experience in localizing fault recovery using Fenix. Doing so will greatly improve scalability on less reliable clusters, but it requires first supporting online recovery as done with this package.

Matthew-Whitlock:
> Most test programs use available introspection to decide whether to run a specific test or not depending on available pre-requisites.

Gotcha, I'll dig into that a bit more to understand how to add unit tests. I should be able to set up a test of basic checkpoint serializing/deserializing without requiring Fenix as a prerequisite. I assume you don't want full tests against Fenix?

akohlmey:
@Matthew-Whitlock I am sorry that it takes so long to get this processed. I don't see much interest across my colleagues and most feel probably not experienced enough in technical HPC details, so they may hold back and let others figure it out. That usually means either Steve or Stan or me. On my side, there are currently too many open items that require my attention, and since few people complain about inefficient restarting (many LAMMPS applications do not use restarts at all and often run on desktops or laptops, often with Windows, there *are* a subset - often high-profile users, too - that do "run big" with LAMMPS but those are mostly in national labs or running on similarly well maintained facilities with good support and competent staff). Things for me went off-the-rails when I could not come to ABQ for the LAMMPS workshop this summer.

I am currently planning for putting all of GitHub processing on the back burner until next year. Just to catch up on other tasks.

akohlmey:
> I'll dig into that a bit more to understand how to add unit tests. I should be able to set up a test of basic checkpoint serializing/deserializing without requiring Fenix as a prerequisite.

For the most part we need to test restarting in combination with pair/bond/angle/dihedral/improper/fix styles and so on. This is done with the test programs in `unittest/force-styles`. I thing that would be overkill to change for your package.

What is more appropriate are the tests in `unittest/formats/`, especially `test_file_operations.cpp`.

You probably just want to set up a minimally representative system, I suggest one extra minimal with just pair style lj/cut or zero and then one with a molecular system and some fixes.
There is more information here: https://docs.lammps.org/Build_development.html#testing
and here: https://docs.lammps.org/Developer_unittest.html

Matthew-Whitlock:
@akohlmey I'm happy to wait for the new year, no worries! I just caught that you're working at Temple. I recently moved to Philadelphia, so meeting in person is an option if you still want to schedule a meeting when you get back to working on the GitHub side of things.

> I suggest one extra minimal with just pair style lj/cut or zero and then one with a molecular system and some fixes

I've added a unit test based on the write_restart test, but I'm not very familiar with the various options and fixes for LAMMPS. Are there any specific fixes that it would be best to test against?

akohlmey:
@Matthew-Whitlock This has been neglected for some time. While we haven't discussed it in depth, this pull request was mentioned in passing at multiple recent meetings of the LAMMPS core developers and the general sentiment was that people are content with the current functionality and want to wait until there are regularly reported issues with it before looking into making changes. This likely also extends to the changes in PR #4899. The main issue here is that the LAMMPS core team is lacking manpower to see such fundamental changes through. We have been experimenting with delegating refactoring tasks to AI coding agents, but those are not yet reliable enough for a code base as quirky as the core LAMMPS code (a lot of the fundamental designs go back over 20 years when LAMMPS was still written in Fortran).

It is always sad to turn down contributions considering the effort that has gone into it, but at the moment we have to face the realities of what we can manage and support over time. We have seen recently with some MPI-IO based optimized I/O styles that these contributions have a tendency to slowly and silently break in subtle ways when core functionality is extended and those optimizations are not maintained. We had to ultimately remove the code to protect LAMMPS users from relying on what had become unreliable code.

For the time being, I will close this pull request, but leave #4899 open and wait until we have had a more detailed discussion among the LAMMPS developers. My hope that it will gain traction is limited, though.

Matthew-Whitlock:
@akohlmey That's understandable. In fact, my initial motivation for #4899 came after noticing that an added restart field broke these changes. Part of the goal there was to allow packages like this to defer to the default read/write classes for deciding what belongs to a checkpoint and how to parse it - only overloading how/where to write and any targeted behavior changes.

If that one does manage to go farther (and I'm happy to update with any changes to make that more likely), then maybe something like this can be revisited in the future.

PR #4673, opened 2025-08-04
MLIAP Unified Generalized Interface

codeformore:
**Summary**

This implementation introduces the ability to extract named per-atom "extra properties" from machine-learned interatomic potentials (MLIAPs) implementing the unified interface, such as uncertainty metrics from ensemble models. It includes both infrastructure support in the MLIAP data handling and a new compute command for use in LAMMPS input scripts.

Key features include:

- New compute style: compute extraProperty/atom
- Kokkos support: Added as compute extraProperty/atom/kk
- Supports hybrid pair styles via the hybridIndex keyword on extraProperty/atom
- Automatic property retrieval through MLIAP infrastructure

**Author(s)**

Caden Fischer (Los Alamos National Laboratory) - [caden.fischer@gmail.com](mailto:caden.fischer@gmail.com)
Deepsana Shahi (Los Alamos National Laboratory) - [deepsanashahi@gmail.com](mailto:deepsanashahi@gmail.com)

**Licensing**

**Implementation Notes**

Key Components Implemented

- **`ExtraProperties` Class**
  - Added a new `ExtraProperties` class instantiated within `MLIAPData`.
  - Internally manages a `std::unordered_map<std::string, double**>`, where:
    - `std::string` is the name of the property (e.g., `"energy_std"`)
    - `double**` points to the associated per-atom data (dimensioned by number of atoms × property dimension)
  - The class is responsible for lifetime management of the memory associated with these properties.

- **`compute extraProperty/atom` Command**
  - A new `compute` style (`extraProperty/atom` and `extraProperty/atom/kk`) registers a request for a specific property name and its dimensionality.
  - The compute interacts with the `ExtraProperties` instance in `MLIAPData` to access the requested per-atom vectors.

- **Integration with MLIAP Unified Interface**
  - Introduced a new virtual method `compute_extra_property()` in the `MLIAPUnifiedInterface`.
  - LAMMPS uses this to request extra property data by name during force evaluation.
  - Implementers of this interface populate the data using `update_extra_property()`, ensuring the requested property is made available to LAMMPS via `ExtraProperties`.

Testing Strategy

- **Test Case**: Used ensemble models trained with `lanl/HIPPYNN` to compute and expose standard deviations for properties such as potential energy.
- **Input Decks**: Custom input scripts provided for testing this feature. These include:
  - Requests for `energy_std` [in.mliap.unified.hippynn.ensemble.Al.txt](https://github.com/user-attachments/files/21582028/in.mliap.unified.hippynn.ensemble.Al.txt)
  - Kokkos variant of `energy_std` [in.mliap.unified.hippynn.ensemble.kk.Al.txt](https://github.com/user-attachments/files/21582026/in.mliap.unified.hippynn.ensemble.kk.Al.txt)
  - Use of hybrid pair styles [in.mliap.unified.hippynn.ensemble.kk.Al.txt](https://github.com/user-attachments/files/21582026/in.mliap.unified.hippynn.ensemble.kk.Al.txt)
- **Test Matrix**:
  - **Backends**: Tested on both **non-Kokkos** and **Kokkos** LAMMPS builds
  - **Parallel Configurations**:
    - MPI ranks: 1, 2, 4, 8
    - Nodes: 1 or 2
    - Kokkos: 1 GPU per MPI task in Kokkos runs

All runs completed successfully and produced the expected per-atom extra property output.

**Post Submission Checklist**

### DIFF
--- a/src/ML-IAP/compute_extra_property_atom.cpp
+++ b/src/ML-IAP/compute_extra_property_atom.cpp
@@ -24,10 +24,9 @@ using namespace LAMMPS_NS;
 
 
 ComputeExtraPropertyAtom::ComputeExtraPropertyAtom(LAMMPS *lmp, int narg, char **arg) :
-    Compute(lmp, narg, arg), extra_property_data(nullptr), data(nullptr), descriptor(nullptr)
+    Compute(lmp, narg, arg), data(nullptr)
 {
   peratom_flag = 1;
-  nmax = 0;
   //Check for args here
   int curArg = 3;
   int nameFlag = 0, dimFlag = 0;
@@ -59,11 +58,8 @@ ComputeExtraPropertyAtom::ComputeExtraPropertyAtom(LAMMPS *lmp, int narg, char *
 
 ComputeExtraPropertyAtom::~ComputeExtraPropertyAtom()
 {
-  memory->destroy(extra_property_data);
-
   //data is managed by PairMLIAP so we should only set it to nullptr
   data = nullptr;
-  descriptor = nullptr;
 }
 
 /* ---------------------------------------------------------------------- */
@@ -81,37 +77,14 @@ void ComputeExtraPropertyAtom::init()
 
   //Get the pointers
   data = castedPair->data;
-  descriptor = castedPair->descriptor;
 
   //Register the extra_property compute with the data class
-  extra_property_index = data->register_extra_property(extra_property_name, size_peratom_cols);
+  data->register_extra_property(extra_property_name, size_peratom_cols);
 }
 
 /* ---------------------------------------------------------------------- */
 
 void ComputeExtraPropertyAtom::compute_peratom()
 {
-
-  //Resize or create eatom_stdev array if needed
-  if (extra_property_data == nullptr || atom->nmax > nmax) {
-    nmax = atom->nmax;
-    memory->create(extra_property_data, nmax, size_peratom_cols, "compute_extra_property_atom:extra_property_data");
-    array_atom = extra_property_data;
-  }
-
-  //Copy the values in data to eatom_stdev
-  for (int i = 0; i < atom->nlocal; i++) {
-    for (int j = 0; j < size_peratom_cols; j++) {
-      extra_property_data[i][j] = data->extra_properties[extra_property_index][i][j];
-    }
-  }
-}
-
-/* ---------------------------------------------------------------------- */
-
-double ComputeExtraPropertyAtom::memory_usage()
-{
-  double bytes = nmax * size_peratom_cols * sizeof(double); //extra_property_data
-
-  return bytes;
+  array_atom = data->extra_properties.get_pointer(extra_property_name);
 }
--- a/src/ML-IAP/compute_extra_property_atom.cpp
+++ b/src/ML-IAP/compute_extra_property_atom.cpp
@@ -19,6 +19,7 @@
 #include "memory.h"
 #include "error.h"
 #include "pair_mliap.h"
+#include "pair_hybrid.h"
 #include <cstring>
 using namespace LAMMPS_NS;
 
@@ -27,6 +28,7 @@ ComputeExtraPropertyAtom::ComputeExtraPropertyAtom(LAMMPS *lmp, int narg, char *
     Compute(lmp, narg, arg), data(nullptr)
 {
   peratom_flag = 1;
+  hybridIndex = -1;
   //Check for args here
   int curArg = 3;
   int nameFlag = 0, dimFlag = 0;
@@ -37,11 +39,17 @@ ComputeExtraPropertyAtom::ComputeExtraPropertyAtom(LAMMPS *lmp, int narg, char *
       extra_property_name = arg[curArg+1];
       nameFlag = 1; 
     }
-    if(strcmp(arg[curArg], "dim") == 0) { //Parse dim
+    else if(strcmp(arg[curArg], "dim") == 0) { //Parse dim
       size_peratom_cols = atoi(arg[curArg+1]);
       dimFlag = 1;
       if (size_peratom_cols <= 0) {
-        error->all(FLERR, "Compute extra_property/atom requires dim > 0.");
+        error->all(FLERR, "Compute extraProperty/atom requires dim > 0.");
+      }
+    }
+    else if(strcmp(arg[curArg], "hybridIndex") == 0) {
+      hybridIndex = atoi(arg[curArg+1]);
+      if (hybridIndex <= 0) {
+        error->all(FLERR, "Compute extraProperty/atom requires hybridIndex > 0.");
       }
     }
     curArg += 2; //Must put inside ifs if variable length args
@@ -50,7 +58,7 @@ ComputeExtraPropertyAtom::ComputeExtraPropertyAtom(LAMMPS *lmp, int narg, char *
   //If not all essential properties were set. Error
   if (nameFlag == 0 || dimFlag == 0)
   {
-    error->all(FLERR, "Compute extra_property/atom missing args (Expecting name and dim).");
+    error->all(FLERR, "Compute extraProperty/atom missing args (Expecting name and dim).");
   }
 }
 
@@ -68,12 +76,25 @@ void ComputeExtraPropertyAtom::init()
 {
   // Check if a pair style has been defined
   if (force->pair == nullptr)
-    error->all(FLERR,"Compute extra_property/atom requires a pair style be defined");
+    error->all(FLERR,"Compute extraProperty/atom requires a pair style be defined");
 
   // Check if it is safe downcast to PairMLIAP pair style
-  PairMLIAP* castedPair = dynamic_cast<PairMLIAP *>(force->pair);
-  if (castedPair == nullptr)
-    error->all(FLERR, "Compute extra_property/atom requires pair mliap to be active");  
+  
+  PairMLIAP* castedPair;
+  if (hybridIndex == -1) {
+    castedPair = dynamic_cast<PairMLIAP *>(force->pair);
+    if (castedPair == nullptr)
+      error->all(FLERR, "Compute extraProperty/atom requires pair mliap to be active");
+  } else {
+    PairHybrid* castedPairHybrid = dynamic_cast<PairHybrid *>(force->pair);
+    if (castedPairHybrid == nullptr)
+      error->all(FLERR, "Compute extraPorperty/atom with hybridIndex requires pair hybrid to be active");
+    if (hybridIndex > castedPairHybrid->nstyles)
+      error->all(FLERR, "Compute extraProperty/atom requires hybridIndex <= number of styles in pair hybird");
+    castedPair = dynamic_cast<PairMLIAP *>(castedPairHybrid->styles[hybridIndex-1]);
+    if (castedPair == nullptr)
+      error->all(FLERR, "Compute extraProperty/atom requires pair style at hybridIndex to be a pair mliap");
+  }
 
   //Get the pointers
   data = castedPair->data;
--- a/src/ML-IAP/compute_extra_property_atom.cpp
+++ b/src/ML-IAP/compute_extra_property_atom.cpp
@@ -1,0 +1,117 @@
+// clang-format off
+/* ----------------------------------------------------------------------
+   LAMMPS - Large-scale Atomic/Molecular Massively Parallel Simulator
+   https://www.lammps.org/, Sandia National Laboratories
+   LAMMPS development team: developers@lammps.org
+
+   Copyright (2003) Sandia Corporation.  Under the terms of Contract
+   DE-AC04-94AL85000 with Sandia Corporation, the U.S. Government retains
+   certain rights in this software.  This software is distributed under
+   the GNU General Public License.
+
+   See the README file in the top-level LAMMPS directory.
+------------------------------------------------------------------------- */
+
+#include "compute_extra_property_atom.h"
+#include "compute.h"
+#include "atom.h"
+#include "force.h"
+#include "memory.h"
+#include "error.h"
+#include "pair_mliap.h"
+#include <cstring>
+using namespace LAMMPS_NS;
+
+
+ComputeExtraPropertyAtom::ComputeExtraPropertyAtom(LAMMPS *lmp, int narg, char **arg) :
+    Compute(lmp, narg, arg), extra_property_data(nullptr), data(nullptr), descriptor(nullptr)
+{
+  peratom_flag = 1;
+  nmax = 0;
+  //Check for args here
+  int curArg = 3;
+  int nameFlag = 0, dimFlag = 0;
+  while (curArg < narg - 1) //All args come in pairs of two argName argValue
+  //If this changes you need to do bounds checking in each argName
+  {
+    if(strcmp(arg[curArg], "name") == 0) { //Parse name
+      extra_property_name = arg[curArg+1];
+      nameFlag = 1; 
+    }
+    if(strcmp(arg[curArg], "dim") == 0) { //Parse dim
+      size_peratom_cols = atoi(arg[curArg+1]);
+      dimFlag = 1;
+      if (size_peratom_cols <= 0) {
+        error->all(FLERR, "Compute extra_property/atom requires dim > 0.");
+      }
+    }
+    curArg += 2; //Must put inside ifs if variable length args
+  }
+
+  //If not all essential properties were set. Error
+  if (nameFlag == 0 || dimFlag == 0)
+  {
+    error->all(FLERR, "Compute extra_property/atom missing args (Expecting name and dim).");
+  }
+}
+
+/* ---------------------------------------------------------------------- */
+
+ComputeExtraPropertyAtom::~ComputeExtraPropertyAtom()
+{
+  memory->destroy(extra_property_data);
+
+  //data is managed by PairMLIAP so we should only set it to nullptr
+  data = nullptr;
+  descriptor = nullptr;
+}
+
+/* ---------------------------------------------------------------------- */
+
+void ComputeExtraPropertyAtom::init()
+{
+  // Check if a pair style has been defined
+  if (force->pair == nullptr)
+    error->all(FLERR,"Compute extra_property/atom requires a pair style be defined");
+
+  // Check if it is safe downcast to PairMLIAP pair style
+  PairMLIAP* castedPair = dynamic_cast<PairMLIAP *>(force->pair);
+  if (castedPair == nullptr)
+    error->all(FLERR, "Compute extra_property/atom requires pair mliap to be active");  
+
+  //Get the pointers
+  data = castedPair->data;
+  descriptor = castedPair->descriptor;
+
+  //Register the extra_property compute with the data class
+  extra_property_index = data->register_extra_property(extra_property_name, size_peratom_cols);
+}
+
+/* ---------------------------------------------------------------------- */
+
+void ComputeExtraPropertyAtom::compute_peratom()
+{
+
+  //Resize or create eatom_stdev array if needed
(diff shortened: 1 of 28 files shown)


src/pair_hybrid.h
```diff
@@ -22,6 +22,7 @@ PairStyle(hybrid/omp,PairHybrid);
 #define LMP_PAIR_HYBRID_H
 
 #include "pair.h"
+#include "compute_extra_property_atom.h"
```
akohlmey: This line must be removed

src/ML-IAP/pair_mliap.h
```diff
@@ -21,11 +21,13 @@ PairStyle(mliap,PairMLIAP);
 #define LMP_PAIR_MLIAP_H
 
 #include "pair.h"
+#include "compute_extra_property_atom.h"
```
akohlmey: This line must be removed

src/ML-IAP/pair_mliap.cpp
```diff
@@ -112,6 +112,10 @@ void PairMLIAP::compute(int eflag, int vflag)
   // calculate stress
 
   if (vflag_fdotr) virial_fdotr_compute();
+
+  // compute extra_properties
+
+  descriptor->compute_extra_properties(data);
```
akohlmey, line 119: This will fail without including the corresponding header, which *must* be included in the .cpp, but not the .h file. Please see my other comments.

src/ML-IAP/compute_extra_property_atom.h
```diff
@@ -0,0 +1,46 @@
...
+   DE-AC04-94AL85000 with Sandia Corporation, the U.S. Government retains
+   certain rights in this software.  This software is distributed under
+   the GNU General Public License.
+
+   See the README file in the top-level LAMMPS directory.
+------------------------------------------------------------------------- */
+
+#ifdef COMPUTE_CLASS
+// clang-format off
+ComputeStyle(extraProperty/atom,ComputeExtraPropertyAtom);
```
akohlmey: The command style name and the corresponding class, file, docfile names must be changed. In LAMMPS all keywords and style names are lower case. Since this is a ML-IAP specific command, it should contain "mliap" in the style, so something like "mliap/property/atom" could fit. "extraProperty" is too generic and not following the established conventions.

src/ML-IAP/compute_extra_property_atom.h
```diff
@@ -0,0 +1,46 @@
...
+
+class ComputeExtraPropertyAtom : public Compute {
+ public:
+  ComputeExtraPropertyAtom(class LAMMPS *, int, char **);
+  ~ComputeExtraPropertyAtom() override;
+  void init() override;
+  void compute_peratom() override;
+
+ private:
+  std::string extra_property_name;
```
akohlmey: Also here the "extra" should replaced. Since this is private to the compute, it could just be called "property_name".

src/ML-IAP/mliap_data.cpp
```diff
@@ -26,10 +26,51 @@
...
 
+ExtraProperties::ExtraProperties(class LAMMPS * lmp) : nproperties(0), nlistatoms(0), Pointers(lmp) {}
+
+ExtraProperties::~ExtraProperties() {
+  for (auto& pair : data) {
+    memory->destroy(pair.second);
+  }
+}
+
+int ExtraProperties::get_dim(std::string name) {
```
akohlmey, line 37: When passing strings, please use `const std::string &` wherever possible to avoid needless copies. This should be done throughout the entire pull request.

src/ML-IAP/compute_extra_property_atom.cpp
```diff
@@ -0,0 +1,111 @@
...
+  // Check if it is safe downcast to PairMLIAP pair style
+  
+  PairMLIAP* castedPair;
+  if (hybridIndex == -1) {
+    castedPair = dynamic_cast<PairMLIAP *>(force->pair);
+    if (castedPair == nullptr)
+      error->all(FLERR, "Compute extraProperty/atom requires pair mliap to be active");
+  } else {
+    PairHybrid* castedPairHybrid = dynamic_cast<PairHybrid *>(force->pair);
+    if (castedPairHybrid == nullptr)
```
akohlmey: This is an unusual way to identify a hybrid sub-style. It would be better, if you followed other examples in the LAMMPS code, so that we have a consistent interface.

src/ML-IAP/compute_extra_property_atom.cpp
```diff
@@ -0,0 +1,111 @@
...
+      nameFlag = 1; 
+    }
+    else if(strcmp(arg[curArg], "dim") == 0) { //Parse dim
+      size_peratom_cols = atoi(arg[curArg+1]);
+      dimFlag = 1;
+      if (size_peratom_cols <= 0) {
+        error->all(FLERR, "Compute extraProperty/atom requires dim > 0.");
+      }
+    }
+    else if(strcmp(arg[curArg], "hybridIndex") == 0) {
```
akohlmey: The "camel case" keyword capitalization conflicts with LAMMPS' conventions. Please also see my other comment about identifying and selecting hybrid sub-styles. Usually this is done by style name with an optional number following in case the same style name is used multiple times.

akohlmey (changes requested):
@codeformore Thanks for your contribution. Before even looking at any specific details of the implementation, I have some rather formal and fundamental suggestions for changes concerning the naming conventions of styles and keywords, and a few other LAMMPS programming style issues (especially for header file "hygiene" which is extremely important for a project as large as LAMMPS).

There will likely be suggestions from me on more details later on and also from other LAMMPS developers.  However, I thought it is best to do this in stages and not overwhelm you with a huge pile of comments.

akohlmey:
> Caden Fischer - [caden.fischer@gmail.com](mailto:caden.fischer@gmail.com) Deepsana Shahi

Please add your affiliation.

codeformore:
Tagging Ben Nebgen (Los Alamos National Laboratory) @bnebgen-LANL and Nick Lubbers (Los Alamos National Laboratory) @lubbersnick who worked with us on this feature.

src/pair_hybrid.h
codeformore: Hi @akohlmey,

Thanks for the feedback. Could you clarify why this `#include` needs to be removed? I currently have it in place to allow `ComputeExtraPropertyAtom` to be declared as a `friend` of `PairHybrid`, enabling access to its `styles` member.

That said, I can rework this to avoid the `friend` declaration and still maintain support for hybrid pair styles by adding a public member function to `PairHybrid`, such as:

```cpp
Pair* get_pair(char* keyword, int index = 1);
```
- ``keyword`` - The pair style name to get
- ``index`` - If only one pair style of type ``keyword`` in the styles, this is disregarded. If multiple, then the pair style at the specified index is selected. 

Would that approach be preferable, or is there another recommendation you'd suggest?

Thanks for your time!

akohlmey: > Thanks for the feedback. Could you clarify why this `#include` needs to be removed?

This is because we want to have as few include statements in the headers as possible to avoid conflicts and needless dependencies. LAMMPS is particularly vulnerable because of how styles are mapped to factory functions in the Force class (and equivalent). There *all* style headers need to be included and we have had multiple hard to track down issues because of that in the past. Thus either a simple forward declaration or a PIMPL (pointer to implementation) pattern needs to be used. There are more details about this in the various developer notes of the Programmer's Guide section of the LAMMPS manual.
We regularly use the IWYU tool from https://include-what-you-use.org/ to "sanitize" the list of include statements and it has made compiling LAMMPS faster and more reliable. That website has additional explanations why this is a good idea.

In this case there is no need to define the full class since you are not accessing any members, but only want to change access to class internals. The `friend` statement by itself is sufficient as it serves as a forward declaration as well. Just look at all the *other* friend statements in the same header.

> support for hybrid pair styles by adding a public member function to PairHybrid, such as:

Please see my other comment. We already *have* a function to look up pair style instances including hybrid sub-styles in the Force class with `Force::pair_match()`. See the `compute pair` documentation and implementation for how to use it and how to have an interface that is consistent throughout LAMMPS. With this function there is likely no need to even have the friend statement. We certainly want to avoid duplicated functionality and different interfaces for similar purposes. The `compute pair` command has similarities to your pull request, only that it operates on global data.

akohlmey: One more thing: please note that when looking up hybrid sub-styles, one has to be very careful with suffixes and exact matches. For technical reasons, the sub-style names are stored without suffix, if I remember correctly. If there is a a non-hybrid pair style, the suffix should be included but it is best if the compute command line allows to use the non-suffix name (i.e. matches with a "^mystyle/cut" regexp).

akohlmey:
@codeformore here is one more thing that occurred to me: your description of the `ExtraProperties` class has a lot of similarities with what `fix property/atom` does. You could delegate the entire data management to that fix (just create a suitable fix instance internally) since it supports per-atom vectors and arrays with integer and double data. As a bonus, you have built-in support for accessing the defined custom properties. Of course, you will still have to fill it with the desired data, but it could lower the amount of maintenance and debugging work to stick with some existing and well tested functionality.

codeformore:
Rebased onto develop branch.

src/pair_hybrid.h
codeformore: Resolved with https://github.com/lammps/lammps/pull/4673/commits/1736fc1631d005a10f0598e9149d0091e98f424b

src/ML-IAP/pair_mliap.h
codeformore: Resolved with https://github.com/lammps/lammps/pull/4673/commits/1736fc1631d005a10f0598e9149d0091e98f424b

src/ML-IAP/pair_mliap.cpp
codeformore: Resolved with https://github.com/lammps/lammps/pull/4673/commits/d53b29322d48001b3d3260c07d7fe6b9e87c9961

src/ML-IAP/compute_extra_property_atom.h
codeformore: Resolved with https://github.com/lammps/lammps/pull/4673/commits/4317158e7243d64f410cfe3afe4a6832671e16fa

src/ML-IAP/compute_extra_property_atom.h
codeformore: Resolved with https://github.com/lammps/lammps/pull/4673/commits/4317158e7243d64f410cfe3afe4a6832671e16fa

src/ML-IAP/mliap_data.cpp
codeformore: Resolved with https://github.com/lammps/lammps/pull/4673/commits/a34bed8391419bd27290427a0f26a95b6c7850f2

src/ML-IAP/compute_extra_property_atom.cpp
codeformore: Resolved with https://github.com/lammps/lammps/pull/4673/commits/4317158e7243d64f410cfe3afe4a6832671e16fa.

I now use the ``force->pair_match`` to get the pair style. Thanks for pointing me to that function!

src/ML-IAP/compute_extra_property_atom.cpp
codeformore: Resolved with https://github.com/lammps/lammps/pull/4673/commits/4317158e7243d64f410cfe3afe4a6832671e16fa

codeformore:
> @codeformore Thanks for your contribution. Before even looking at any specific details of the implementation, I have some rather formal and fundamental suggestions for changes concerning the naming conventions of styles and keywords, and a few other LAMMPS programming style issues (especially for header file "hygiene" which is extremely important for a project as large as LAMMPS).
> 
> There will likely be suggestions from me on more details later on and also from other LAMMPS developers. However, I thought it is best to do this in stages and not overwhelm you with a huge pile of comments.

I've pushed updates that address the feedback provided in the previous review. Please let me know if there's anything I missed. Thanks again for your time and thoughtful feedback.

bnebgen-LANL:
> @codeformore here is one more thing that occurred to me: your description of the `ExtraProperties` class has a lot of similarities with what `fix property/atom` does. You could delegate the entire data management to that fix (just create a suitable fix instance internally) since it supports per-atom vectors and arrays with integer and double data. As a bonus, you have built-in support for accessing the defined custom properties. Of course, you will still have to fill it with the desired data, but it could lower the amount of maintenance and debugging work to stick with some existing and well tested functionality.

Axel, thanks for the idea. You are probably correct in that doing it via a fix would possibly make things easier. We'd like to hold off on making this shift until another time since we've got this feature working, if that is OK. We aren't seeing any performance degradation doing it as a compute.

akohlmey:
> Axel, thanks for the idea. You are probably correct in that doing it via a fix would possibly make things easier. We'd like to hold off on making this shift until another time since we've got this feature working

@bnebgen-LANL Sorry, but you are missing my point. The issue is not compute versus fix. A compute is the appropriate class to provide the information you want to make available. My point is that your compute is *duplicating* functionality that already exists, i.e. mapping per-atom arrays to keywords. So, to reduce code duplication, you would just create a fix property/atom instance *internally* from within the compute and can then store data in it and refer to it.

I personally don't worry so much about performance, but I do worry about the effort required to *maintain* LAMMPS. The more duplicated functionality there is, the more work it is to keep things working and up-to-date. What is worse is that the interfaces are usually not exactly the same, which results in somewhat inconsistent behavior and that is something we want to reduce, if not avoid, for the sake of the users.

src/ML-IAP/compute_mliap_property_atom.cpp
```diff
@@ -0,0 +1,106 @@
...
+    castedPair = dynamic_cast<PairMLIAP *>(force->pair_match("mliap", 0, hybrid_index));
+    if (castedPair == nullptr)
+      error->all(FLERR, "Compute mliap/property/atom requires a mliap pair style at hybrid_index to be a pair mliap");
+  }
+
+  //Get the pointers
+  data = castedPair->data;
+
+  //Register the extra_property compute with the data class
+  data->register_extra_property(property_name, size_peratom_cols);
```
rbberger, line 98: As @akohlmey mentioned this can likely be replaced by something that internally creates a `fix property/atom`. See https://github.com/lammps/lammps/blob/develop/src/RHEO/compute_rheo_interface.cpp#L66 for an example on how to create a `d2_some_name X` array with `X` columns and retrieve its index via `atom->find_custom`.

bnebgen-LANL: Thanks Richard. Sorry to be a pain, but is there an example in kokkos? We would want to see the correct way to do this on both sides.

rbberger: @stanmoore1 It seems to me that `fix property/atom/kk` is missing full support for `IARRAY` and `DARRAY` and isn't creating any dual views for those. So I would expect the following in `atom_kokkos.h`:
```c++
DAT::tdual_int_3d k_iarray;
DAT::tdual_float_3d k_darray;
````
Also
```c++
DAT::tdual_int_2d k_ivector;
```
seems to be missing? Same story?

@akohlmey ~think we are also missing something equivalent to `find_custom` in `atom_kokkos.h` that gives us the kokkos views so we can utilize device data.~ Nevermind, it's the same index as on the host and the `k_*` members are public, so all you need is to cast `atom` to `AtomKokkos*` and access the public field with the same index.

rbberger: @bnebgen-LANL I'm not seeing a straight forward example that you can just copy. It looks more like the Kokkos code could be extended to a more general solution and you would be the first client of it.

codeformore: Thanks for the detailed suggestions and discussion.
 
I looked into replacing the `ExtraProperties` infrastructure with a fix-based approach using `fix property/atom` as suggested. While that pattern works well on the CPU implementation, it appears that the Kokkos version (`fix property/atom/kk`) currently lacks full support for `DARRAY` and `IARRAY`, including the necessary `tdual_*` views (as mentioned above). Without that, I don't see a straightforward way to access custom per-atom double arrays on device (please tell me if my assumption is incorrect), which we rely on for the modifications in the MLIAP unified interface.
 
Given this, I don't believe the fix-based pattern is currently viable on the Kokkos side without additional infrastructure. I'm happy to adapt to a unified `fix property/atom` interface in the future if support is added—but for now, the `ExtraProperties` approach provides the functionality we need for both non-Kokkos and Kokkos runs and keeps the CPU and Kokkos implementations similar.
 
Please let me know if you'd recommend a different approach given these constraints. I'm open to alternatives if there's a better-supported path. Thanks again for the feedback.

stanmoore1: It should be fairly trivial to add the missing dual views in `fix property/atom/kk`. We already have something similar in SPARTA.

bnebgen-LANL:
I'm happy to go with whatever the lammps maintainers want. However, Caden is only at LANL until Friday. Is there a way we could move forward with the merge and if/when Stan or somebody else implements the appropriate dual views, I can revisit this to implement it through the fix methodology?

akohlmey:
> Is there a way we could move forward with the merge

@bnebgen-LANL What is the rush? We generally prefer to iterate pull requests until everybody is satisfied. At the moment, the PR doesn't even pass the automated testing. So far, we only looked at some formal details. There hasn't even been a proper review, manual tests, checks of the documentation and so on. There also are some sweeping changes coming (KOKKOS with mixed and single precision, build system updates and more. Plus we have the LAMMPS workshop next week.
So there are likely weeks until this is going to be ready, even without the elimination of duplicated functionality.

akohlmey:
@bnebgen-LANL this has been idle for three months. Any updates?

If there are no plans to bring this pull request into a shape that it has merge conflicts resolved and passes the automated tests (or corrects false positives in the tests), then I suggest we close this for now and you resubmit when it is closer to being complete and reviewable.

bnebgen-LANL:
Sorry. The person working on this was out sick for a while. We should make more progress soon.

athomps:
Hi @bnebgen-LANL, sorry for not looking at this sooner.  Based on this discussion, in order for this to be accepted in to LAMMPS, I see two alternatives:

1. Leverage existing `fix properties/atom` rather than creating new `compute extraProperty/atom` command.
2. Leverage existing `fix pair` command.  This was created to extract per-atom uncertainty metric from `pair_style pace/extrapolation`.  It would be relatively easy to add an interface to extract pointer to uncertainties from `MLIAPData::extra_properties`, see `PairPACEExtrapolation::extract_peratom()`

athomps (changes requested):
Given the difficulties identified with the new `compute extraProperty/atom` command, I think it best to withdraw this PR until a clear path forward has been identified. [I have changed my mind on this, see below]

athomps (changes requested):
The best path forward for this is to leverage the existing fix pair command, which was created by @sjplimp for exactly this kind of use case. It so far just has one use instance:  to extract per-atom uncertainty metric from `pair_style pace/extrapolation`. It would be relatively easy to add a function `PairMLIAP::extract_peratom()` to extract pointers to uncertainties and other per-atom quantities from `MLIAPData::extra_properties`, see `PairPACEExtrapolation::extract_peratom()`. The atom uncertainties currently accessed from the compute class `ComputeMLIAPPropertyAtom` can just as well or better be accessed from the `FixPair` class. Since this can be done fairly quickly, I recommend keeping this PR open.

athomps:
@bnebgen-LANL this has been idle for nearly three months. Any updates?

bnebgen-LANL:
We'll have our next attempt ready on 2/23,  or you can close now.

akohlmey:
@bnebgen-LANL As suggested on Feb 10th we're closing this for now. Please resubmit when you have updates worth looking at.

PR #4663, opened 2025-07-22
Implement post stable release changes for minimum versions, build and package support

akohlmey:
**Summary**

After the 22 July 2025 release LAMMPS will require CMake 3.20, C++17.
The ATC, AWPMD, and POEMS packages are removed and others will be available only when compiling with CMake.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

No.

**Implementation Notes**

The following individual changes are included
- [x] raise minimum CMake version to 3.20
- [x] require and enable C++17
- [x] update documentation for new version requirements
- [x] remove ATC package and ATC related documentation and examples
- [x] remove AWPMD package and AWPMD related documentation and examples
- [x] remove POEMS package and POEMS related documentation and examples
- [x] update Purge.list with removed files
- [x] document removed packages and point to 22 July 2025
- [x] remove APIP package from legacy build
- [x] remove COLVARS package from legacy build
- [x] remove COMPRESS package from legacy build
- [x] remove ELECTRODE package from legacy build
- [x] remove GPU package from legacy build
- [x] remove H5MD package from legacy build
- [x] remove KIM package from legacy build
- [ ] remove KOKKOS package from legacy build
- [x] remove LEPTON package from legacy build
- [x] remove MACHDYN package from legacy build
- [x] remove MDI package from legacy build
- [x] remove MOLFILE package from legacy build
- [x] remove ML-HDNNP package from legacy build
- [x] remove ML-PACE package from legacy build
- [x] remove ML-POD package from legacy build
- [x] remove ML-QUIP package from legacy build
- [x] remove NETCDF package from legacy build
- [x] remove PLUGIN package from legacy build
- [x] remove PLUMED package from legacy build
- [x] remove PYTHON package from legacy build
- [x] remove QMMM package from legacy build
- [x] remove RHEO package from legacy build
- [x] remove SCAFACOS package from legacy build
- [x] remove VORONOI package from legacy build
- [x] remove VTK package from legacy build
- [x] document CMake vs. legacy build changes
- [x] remove legacy build for liblinalg.a

**Post Submission Checklist**

- [x] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [x] The source code follows the LAMMPS formatting guidelines
[...]

PR #4682, opened 2025-08-08
Remove neighbor style and comm mode multi/old

akohlmey:
**Summary**

The original implementation of neighbor style and comm mode multi has been phased out for quite some time. Most recently it was undocumented and only available for backward compatibility as multi/old. This pull request finally removes the actual code.

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

N/A

**Post Submission Checklist**


jtclemm:
I'm guessing no one ever replied to that extra message saying *multi/old* would be deprecated?

doc/src/Commands_removed.rst
```diff
@@ -12,6 +12,15 @@ stop LAMMPS and print a suitable error message in most cases, when a
 style/command is used that has been removed or will replace the command
 with the direct alternative (if available) and print a warning.
 
+Neighbor style and comm mode multi/old
+--------------------------------------
+
+.. deprecated:: TBD
+
+The orignal implementation of neighbor style multi and comm mode multi,
```
jtclemm: typo, *orignal*. ~The comm mode multi also didn't used to exist with *multi/old*.~ Or maybe I misremembered.

akohlmey:
> I'm guessing no one ever replied to that extra message saying _multi/old_ would be deprecated?

Exactly! Nada, zip, zilch.

jtclemm:
> > I'm guessing no one ever replied to that extra message saying _multi/old_ would be deprecated?
> 
> Exactly! Nada, zip, zilch.

In my opinion, I think it's safe to remove then. If there's no rush, I'd just like to wait until at least Tuesday though so I can check with some coworkers on Monday.

akohlmey:
> I'd just like to wait until at least Tuesday though so I can check with some coworkers on Monday.

@jtclemm FYI, a simple way to prevent a pull request from merging is to request changes.
See for example PR #4663 where Stan put in a request for changes, so he has time to update his KOKKOS test environment.

jtclemm (changes requested):
I got some feedback from various users about this deprecation before and during the workshop. Most were fine with deprecating multi/old, but there were two potential concerns brought up.

First was a concern that at intermediate size ratios (cut_large / cut_small ~ 1.5), that multi/old may outperform multi and bin. I tested this using the LJ melt example and extending cutoffs, but found the best performance was effectively always bin or multi (within fluctuations). If multi/old is ever optimal, it's only over a very narrow domain with very minimal savings. Data below.
 
Interestingly, in nearly monodisperse systems multi consistently built the neighbor list faster than multi/old (seems counterintuitive to me) but took longer to calculate forces (must be the ordering of atoms in the neighbor list vs. in memory?), but it was only ever a few percent either way. 

Second was a report of possible errors in multi. These are not certain and, if so, were seen quite early after multi was introduced. I suspect they may have been initial bugs that have already been fixed. But they are going to take a month to try and reproduce them to confirm. 

So revisit in September?


image

jtclemm:
> I got some feedback from various users about this deprecation before and during the workshop. Most were fine with deprecating multi/old, but there were two potential concerns brought up.

Thanks for waiting, no remaining concerns emerged so I think it's safe to go ahead with the deprecation. 

Could we add a message to comm.cpp and neighbor.cpp to acknowledge 'multi/old' was a previously recognized command? As in #4683, just something like
```
  } else if (strcmp(arg[1],"multi/old") == 0) {
     error->all(FLERR, "The multi/old option has been deprecated. Please see the documentation for more information");
  }
```

jtclemm (approved):
Can't figure out how to remove the old change request. Hoping this overrides.

akohlmey:
> Can't figure out how to remove the old change request.

Approving is the way to go.

akohlmey:
> Could we add a message to comm.cpp and neighbor.cpp to acknowledge 'multi/old' was a previously recognized command?

Done. See commit 56cabdce1520a353b302bfe58f6d207dba263ad7

PR #4683, opened 2025-08-08
Remove GJF implementation from fix langevin

akohlmey:
**Summary**

The GJF implementation has been moved to fix gjf (and improved) and the version in fix langevin is obsolete. This pull request removes it.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

N/A

**Post Submission Checklist**


src/fix_langevin.cpp
```diff
@@ -110,20 +105,6 @@ FixLangevin::FixLangevin(LAMMPS *lmp, int narg, char **arg) :
       else
         ascale = utils::numeric(FLERR, arg[iarg + 1], false, lmp);
       iarg += 2;
-    } else if (strcmp(arg[iarg], "gjf") == 0) {
```
jtclemm, line 113: How about preserving this keyword (at least temporarily) and trigger a distinct error stating that gjf is deprecated and moved into fix gjf?

jtclemm: One additional step to transition to deprecation. Could do the same for multi/old too.

akohlmey: I think this is overkill in both cases.

The GJF developers have stated that the existing implementation has issues and they originally submitted updates to fix langevin that were converted into fix gjf.
They *want* users to use the new fix and do *not* want to support the old, buggy version.

If somebody wants to absolutely go against that recommendation, it is always possible to revert to an older version of LAMMPS, like the current stable version (which will print just a warning).

Same for multi/old which has been undocumented for a long time.
I understand the need for some transition, but there is also the problem of being needlessly limited by having to carry around tons of legacy code. People that have not changed since, are not likely to change until we remove the option.

jtclemm: We might be talking past one another here. I'm simply suggesting keeping a note:

```
 } else if (strcmp(arg[iarg], "gjf") == 0) {
   error->all(FLERR, "The GJF option was deprecated. Please see the documentation for more information.");
 } else ...
```

so the code echos the documentation's deprecation message instead of a generic, "Invalid option gjf".

akohlmey: @jtclemm you are correct. I misunderstood. Please see the revised argument parsing (with generally updated error messages).

akohlmey:
@stanmoore1 I need a little help here to merge the changes to fix langevin/kk in this pull request with your single/mixed precision work.

jtclemm (approved):
The non-kokkos changes all look good to me. I assume Stan will handle whatever's left on the Kokkos side.

akohlmey:
> @stanmoore1 I need a little help here to merge the changes to fix langevin/kk in this pull request with your single/mixed precision work.

Nevermind. I think I've been able to sort this out.

PR #4686, opened 2025-08-09
Add feature to dump image to draw dynamic bonds 

akohlmey:
**Summary**

Add a feature to dump image that would draw bonds based on a cutoff to be used for simulations where bonds are implicit, for example with AIREBO, ReaxFF, Tersoff, etc.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

Bonds are derived by walking the neighbor list and finding pairs of atoms that are closer than the provided cutoff. This could be made more sophisticated by e.g. using a variable cutoff based on covalent radii or atom diameters.

The new feature is also accessible in the LAMMPS-GUI image viewer.

Plus bugfixes for LAMMPS-GUI for bugs found during the LAMMPS workshop tutorial

**Post Submission Checklist**

Below are two example images from the "breakable" input of the LAMMPS tutorial No.2 which uses the airebo pair style and atom style atomic.
```
dump viz all image 500 myimage.*.ppm type type size 1000 400 zoom 8.0 shiny 0.3 &
   box no 0.01 view 0 90 shiny 0.1 fsaa yes autobond 1.75 0.5 adiam 0.6 ssao yes 53262 0.6
dump_modify viz pad 5 backcolor white acolor 1 gray
```

myimage1
myimage2

Packages with pre-compiled binaries are currently available at https://download.lammps.org/testing/


akohlmey:
@simongravelle Had I known that it would take me only a few hours to implement this, I would have done it months ago and we would have it included in the stable release and the tutorial paper. This is far better than the "hack" in the tutorial. :frowning_face:

doc/src/Tools.rst
```diff
@@ -600,21 +600,23 @@ acceleration is available and enabled by default.
...
-the `Qt GUI framework <https://www.qt.io/product/framework>`_.
-Currently, Qt version 5.12 or later is required; Qt 5.15LTS is
-recommended; support for Qt version 6.x is available.  Building LAMMPS
-with CMake is required.
+.. versionchanged:: TBD
+
+LAMMPS-GUI version 1.7 and later is programmed in C++ based on the C++17
+standard and using the `Qt GUI framework
+<https://www.qt.io/product/framework>`_.  Currently, Qt version 5.15LTS
+or later is required commended; support for Qt version 6.x is available.
```
jtclemm: required commended -> required/recommended?

src/dump_image.cpp
```diff
@@ -1264,6 +1296,126 @@ void DumpImage::create_image()
...
   }
 
+  // render dynamic bonds for my atoms
+  // both atoms in pair must be selected and closer than bond cutoff for bond to be rendered
+  // if newton_bond is off, only render bond once
+  // render bond in 2 pieces if crosses periodic boundary
+
+  if (bondflag == AUTO) {
+    // grab some suitable neighbor list, if available
+    auto *list = neighbor->get_best_pair_list();
```
jtclemm, line 1306: Why not use `force->pair->list`? It seems most physically relevant.

akohlmey: That fails for hybrid pair styles. Originally, I tried creating my own (occasional) neighbor list request, but couldn't get it to work. Then I use force->pair->list, but ended up with the current solution as the most complete way.

jtclemm: If it is used with a hybrid pair style with two substyles with different cutoffs, is it then random whether or not it grabs the nlist that's trimmed vs. the nlist with a longer cutoff? Could a trimmed list could have a cutsq_custom less than bondcut^2?

(acknowledging that I have not really looked at how pair hybrid nlists work)

akohlmey: Hybrid neighbor lists are usually skip lists and those would be skipped (!) by the `get_best_pair_list()` function. but there are complete neighbor list request(s) from which the skip lists are created.

The (non-bonded) cutoffs for pair styles with implicit bonds are much longer than what would be needed for meaningful bonds. For performance reasons, many of those styles do a "trimmed" list internally, e.g. AIREBO for looping over triples and quadruples of atoms that are just a bond-length apart.

In a perfect world, I would submit a neighbor list request with a suitable custom cutoff, like it is done with `compute rdf`, for example. I tried a lot of things, but ultimately did not succeed. If there is more time, I would need to discuss with @sjplimp what can be done. For now the "solution" is adequate. After all, this is not for computing forces but only for visualization. So if it not perfect, it won't change the science.

jtclemm: That's fair, I wanted to make sure I understood. 

I didn't follow the changes to fix shake, but other than that, the PR looks good to me.

akohlmey: > I didn't follow the changes to fix shake

It looks more complex than it is. Previously, I was re-using lists that were also used during MD. But there were always cases, where the system would "explode" unexpectedly and the restraints would be all over the place. I found that the lists would become eventually inconsistent and point to the wrong atoms. Now, the code is collecting where it needs to apply bond restraints directly. It turned out to be rather low overhead, since I don't have to solve any constraint equations, just add forces from a harmonic spring.

akohlmey: Actually, those changes were not supposed to be in the PR. They are in #4684. I needed some changes from that branch, so I pulled the fix shake changes by accident as well and later I was too lazy to remove them. ;-)

jtclemm: Got it, I'll assume those changes will be reviewed more carefully in that PR.

jtclemm (approved):

PR #4692, opened 2025-08-14
fix pimd/langevin: small improvement for lj unit style

Yi-FanLi:
**Summary**

This PR improves the usage of running PIMD in the lj unit style. The previous implementation requres the user to specify the Planck constant and the mvv2e value in the unit of the LJ parameters. Instead, the new implementation simply requires the user to provide the unit style of the LJ parameters.

**Related Issue(s)**

**Author(s)**

Yifan Li
Department of Chemistry, Princeton University
yifanl0716@gmail.com

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**


src/REPLICA/fix_pimd_langevin.h
```diff
@@ -194,6 +196,30 @@ class FixPIMDLangevin : public Fix {
   int size_restart_global();
   int pack_restart_data(double *list);
   void restart(char *buf) override;
+
+   // Planck's constant for each LAMMPS unit style
```
akohlmey, line 200: These two maps should not be required to be in the header. Instead they could be moved to the .cpp file into an anonymous namespace.

akohlmey:
@Yi-FanLi Sorry for the long delay before looking at this. Work has been piling up and my priorities have changed a bit, too.

I am confused by the notion of having "unit styles" for reduced units. In reduced units everything is dimensionless, so where do those other units come from and what is the purpose?
In all but the most simple and fundamental cases, I fail to see the purpose of using reduced units in a way that those are linked to direct material properties. Why not just use a regular unit style and be done with it?

akohlmey:
@Yi-FanLi please provide some feedback and address the suggested bugfixes and improvements. Otherwise, I will close this pull request due to inactivity.

Yi-FanLi:
> @Yi-FanLi Sorry for the long delay before looking at this. Work has been piling up and my priorities have changed a bit, too.
> 
> I am confused by the notion of having "unit styles" for reduced units. In reduced units everything is dimensionless, so where do those other units come from and what is the purpose? In all but the most simple and fundamental cases, I fail to see the purpose of using reduced units in a way that those are linked to direct material properties. Why not just use a regular unit style and be done with it?

Hi @akohlmey, thanks for raising this question. I drafted a response to your comments last Thursday, but had not sent it because I was fully occupied since Friday. 

This is indeed a very good question. This question is related to another question I have been thinking about for a long time: should we support the `lj` unit style in `fix pimd/langevin`?

I did think a bit and my answer is "yes", since nothing prohibits me to do so. There are cases that need to run pimd with the reduced units, for example, in [Phase behavior of the quantum Lennard-Jones solid](https://pubs.aip.org/aip/jcp/article/153/7/074502/1064765/Phase-behavior-of-the-quantum-Lennard-Jones-solid). 

Then the question becomes: how should we determine the $\hbar$ constant used in `omega_np`?

(To be continued)

Yi-FanLi:
(Following my last comment)

When using reduced units we need to convert the physical Planck's constant, $h=6.62606896\times 10^{-34}\ \mathrm{J\cdot s}$,  into the reduced unit. Each system has their own reduced unit. For example, for Neon, $m=20.1797$ Dalton, $\varepsilon=3.0747\times 10^{-3}$ eV and $\sigma=2.7616$ Å (according to [Georgescu et al, JCP 2013](https://pubs.aip.org/aip/jcp/article/138/13/134502/72225/Mapping-the-phase-diagram-for-neon-to-a-quantum)). This set of units translates into $h=0.600\ \sigma\sqrt{m\varepsilon}$ in the reduced units. This implies that we should use `hplanck = 0.600` instead of `hplanck = 1.0` when calculating `omega_np`.

Yi-FanLi:
Physically, this factor `0.600` originates from the de Broglie wavelength of the atom, $\Lambda=h/\sqrt{m\varepsilon}$. [J. de Boer](https://www.sciencedirect.com/science/article/pii/S0079641708601006) defined a "quantumness" parameter, $\Lambda^{\ast}=\Lambda/\sigma$, which is the ratio between the de Broglie wavelength and $\sigma$. This de Boer parameter $\Lambda^{\ast}$ is nothing else but the Planck's constant in the reduced unit. When we are using the unit style `lj` in PIMD, we should assign the value of $\Lambda^{\ast}$ to `hplanck`.

Yi-FanLi:
Now the question becomes: how do we calculate the de Boer quantumness paramter? The equation is $\Lambda^{\ast}=h/\sigma\sqrt{m\varepsilon}$. To implement this, we have two options:

- Option 1: ask the user for the values of $m$, $\varepsilon$ and $\sigma$ and their units in the input file, and then calcualate $\Lambda^{\ast}$ internally. This is what I currently have in my branch, but may seem confusing. I will add more clarification in the document if we keep doing it in this way.

- Option 2: ask the user to provide the value of $\Lambda^{\ast}$ directly. Since there is the keyword `sp` which scales `hplanck`, we should can keep the code unchanged and just ask the user to set `sp` to the value of $\Lambda^{\ast}$. However, we should clarify this in the documentation.

@akohlmey Which option do you prefer? If you prefer option 1, I will improve the current code. If you prefer option 2, I will close this PR and open a new PR to update the documentation and example.

akohlmey:
> @akohlmey Which option do you prefer? If you prefer option 1, I will improve the current code. If you prefer option 2, I will close this PR and open a new PR to update the documentation and example.

Any solution where the settings are dependent on the material you are simulating, are not truly reduced units, but something else. If the generality is lost, then there is no more reason to use reduced units. Some people justify this because they are misunderstanding floating point math, but to me that is no justification.

If there is no (simple) way to have true reduced units, my vote is for not supporting them at all.

Yi-FanLi:
Hi @akohlmey, I totally agree that the generality of the reduced units should not be lost. In this sense it seems to me that option 2 is a better choise. We give the users the freedom to set the "quantumness parameter" just as they set the temperature or volume of the system. Then the users determine how to correspond the simulation results from the reduced units to physical units. This is already supported by the current code, but I need to make some clarifications in the document. Do you agree with me?

Yi-FanLi:
As discussed above, I am now supporting the `lj` unit style in the way of option 2 as done in #4756, which will close this PR.

PR #4695, opened 2025-08-14
New fix for Reactive Monte Carlo

vishnu2709:
**Summary**

This new fix performs Reactive Monte Carlo Molecular Dynamics (RMCMD) simulations for the simultaneous treatment of the doping reaction and morphology evolution in doped organic semiconductors. More details are provided in the documentation file, and also in our manuscript which uses this fix ([Raghuraman V, Verma A, Jackson N. All-Atom Reactive Monte Carlo Molecular Dynamics for Molecular Doping in Organic Semiconductors. ChemRxiv. 2025; doi:10.26434/chemrxiv-2025-713hw-v2](https://chemrxiv.org/engage/chemrxiv/article-details/688244a6728bf9025e09ab03)), currently submitted to peer-review.
Part of MC package.

**Related Issue(s)**

This is a new feature and does not address an existing issue.

**Author(s)**

Vishnu Raghuraman, University of Illinois at Urbana-Champaign
Institutional email: vishnura@illinois.edu
Long-lived email: vishnu.raman@gmail.com

**Licensing**

**Backward Compatibility**

No backward compatibility will be affected. 

**Implementation Notes**

The source code changes are entirely confined to the new fix files (fix_rmc_partial.h and fix_rmc_partial.cpp). No other changes to any core lammps classes have been made. It is added to the MC package. The implementation has been validated in our manuscript (referenced above), which also contains more details about the theory behind the method. The compilation was tested on our group cluster (dual socket Intel Xeon Gold CPU with 48 cores) with Intel oneAPI 2021 compilers, and on the Purdue Anvil system with GCC 11.2 compiler. 

**Post Submission Checklist**

The data and results of our above-mentioned manuscript is made available at Box
https://uofi.box.com/s/1sf6jvf3lo7ctgr4yaia3s55zpubn9e6

This contains input files, data files and output structures which may be helpful for people wanting to use the fix.


vishnu2709:
Removed a bunch of commented out debug code.

src/MC/fix_rmc_partial.cpp
```diff
@@ -0,0 +1,1446 @@
...
+      if (i < dopant_size)
+      {
+         dopant_mol_c[i] = new double [6];
+      }
+    }
+
+
+  if (comm->me == 0)
+  {   
+      fmt::print(screen, "{}\n", "###############################################");
```
akohlmey: This should use `utils::logmesg()`, and there is no need for a format `"{}\n", "####"` here, you can use just `"####\n"`.

src/MC/fix_rmc_partial.cpp
```diff
@@ -0,0 +1,1446 @@
...
+  type_dist = std::uniform_int_distribution<> (0,num_charge_states-1);
+  acc_dist = std::uniform_real_distribution<> (0,1);
+
+  // Separation between charge states
+  deltaQ = 1.0/(num_charge_states-1.0);
+  
+  // List of charges
+  charges = new double [num_charge_states];
+  charges[0] = 0;
+  for (int i=0;i<num_charge_states-1;i++)
```
akohlmey: Please adopt the LAMMPS programming style, where the opening brace is on the same line as the "for" or "if" or "else". A simple way to get close to it, is to apply clang-format.

src/MC/fix_rmc_partial.cpp
```diff
@@ -0,0 +1,1446 @@
...
+      total_charged_dopants = total_charged_dopants+num_dopant_charge[i];
+      total_charged_semiconductors = total_charged_semiconductors+num_semiconductor_charge[i];
+   }
+   double total_doping_efficiency = (double)total_charged_dopants/(double)n_dopant;
+
+   // Write out COM for all molecules, for RDF/morphology or other CG uses
+
+   if (comm->me == 0)
+   {
+      FILE *fcom = fopen("com.xyz", "w");
```
akohlmey: Having hard-coded file names is a very bad idea.

src/MC/fix_rmc_partial.cpp
```diff
@@ -0,0 +1,1446 @@
...
+      fmt::print(screen,"{} {}\n", "Acceptance rate: ",acceptance_rate);
+      fmt::print(screen, "{}\n", "###############################################");
+   }
+}
+
+
+FixRMCPartial::~FixRMCPartial()
+{
+   post_mortem();
+   delete[] dde;
```
akohlmey: There seem to be many more "new" operators than there are "delete" here. Have you checked your code against memory leaks?

src/MC/fix_rmc_partial.cpp
```diff
@@ -0,0 +1,1446 @@
...
+  
+  double total_energy = c_pe->compute_scalar();
+  update->eflag_global = update->ntimestep;
+
+  return total_energy;
+}
+
+int FixRMCPartial::determine_dopant_or_semiconductor(int mol_id)
+{
+   int indicator[comm->nprocs];
```
akohlmey: The arrays `indicator` and `global_indicator` are what is called "variable length arrays". This is a feature available in Fortran and (some versions of) C, but **NOT** in C++. Many C++ compilers accept those because of compatibility with C, yet this is essentially invalid C++. You can replace those with dynamically allocated storage using new/delete or `std::vector<int>`.

src/MC/fix_rmc_partial.cpp
```diff
@@ -0,0 +1,1446 @@
...
+  return global_mol_id;
+}
+
+void FixRMCPartial::calculateMoleculeCOM(double* com, struct Mol* molecule)
+{
+   double *local_com;
+   double unwrap[3];
+   double global_mass=0;
+   double local_mass=0;
+   local_com = new double [3];
```
akohlmey: Why do a dynamic allocation of a small fixed size? Where is the delete[]?

src/MC/fix_rmc_partial.cpp
```diff
@@ -0,0 +1,1446 @@
...
+   }
+   MPI_Allreduce(local_com, com, 3, MPI_DOUBLE, MPI_SUM, world);
+   MPI_Allreduce(&local_mass, &global_mass, 1, MPI_DOUBLE, MPI_SUM, world);
+   for (int j=0;j<3;j++)
+   {
+      com[j] = com[j] * (1.0/global_mass);
+   }
+   // map it back into the box
+   domain->remap(com);
+   //fmt::print(screen, "{} {} {}\n", com[0], com[1], com[2]);
```
akohlmey: please remove all debug comments

src/MC/fix_rmc_partial.cpp
```diff
@@ -0,0 +1,1446 @@
...
+     {
+        molecule_charge_states[i] = 0.0;
+        molecule_type[i] = determine_dopant_or_semiconductor(i+1);
+     }
+     num_dopant_charge[0] = n_dopant;
+     num_semiconductor_charge[0] = n_semiconductor;
+  }
+  else
+  {
+     TextFileReader molecule_type_handle("molecule_type.dat", "MType");
```
akohlmey: More hard-coded file names.

akohlmey (changes requested):
Thanks for your submission. Unfortunately, it already fails some elementary tests and has some significant programming style issues.
We would like all LAMMPS source code to have a mostly uniform programming style.

There are also some issues that we consider bad practice, like using variable-length arrays and hard-coded file names. `fmt::print()` should not be used at all (we want to eventually switch from `fmt::format()` to `std::format()` and `fmt::print()` is fmtlib extension. We have thus added `utils::print()`, however, in your case you should use `utils::logmesg()` so that data is not only printed to the screen and also that it correctly handles when either logfile or screen output has been turned off (and thus either `screen` or `logfile` are NULL).

There also should be a minimal and short example input to demonstrate the usage of the fix (not a proper application) which could also be used for testing (e.g. for memory leaks).

There will be more thorough reviews in the future, but for now, please work on the basic cleanup.

vishnu2709:
Thank you for your feedback. I will fix all the issues you have raised. I appreciate your patience.

src/MC/fix_rmc_partial.cpp
akohlmey: Here is an idea: since LAMMPS has now built-in support for reading (and writing) of JSON format files (see, e.g. the molecule command), you could combine the multiple input/output files into a single, hierarchical file (or most of them and just leave those that make sense to be in a different format). Each current, separate file would then become a section in the JSON format file. JSON files are also easy to be parsed in Python scripts and then the data is available as a tree of directories.

vishnu2709: that's a great idea, it would be super convenient to have a single file with all the RMC inputs instead of the multitude of files I have now. I will implement that.

vishnu2709:
@akohlmey thanks for your feedback and sorry for the delay. I have updated my fix according to the recommendations provided, which are mainly
1. Removed VLAs
2. Removed debug code.
3. Changed the if, else,for { to match stylistic guidelines
4. Replaced all fmt::print with utils::logmesg
5. Ensuring that all new operations have a delete, avoiding memory leak issues.
6. Removed all instances of hardcoded file names
7. Reworked the input such that all the required parameters are read from a JSON file, the name of which is provided when the fix is called (e.g. fix 1 all rmc/partial rmc_input.json)
8. Changed the documentation to reflect the above changes
9. Added example input/data/JSON file in examples/mc that run quickly.

Please let me know if anything else is needed. I'm grateful for your time and input.

src/MC/fix_rmc_partial.cpp
akohlmey: The use of `utils::logmesg()` was not as intended in many places and inefficient in others. I updated this myself.

src/MC/fix_rmc_partial.cpp
akohlmey: I applied clang-format myself.

src/MC/fix_rmc_partial.cpp
akohlmey: Execellent!

src/MC/fix_rmc_partial.cpp
akohlmey: I have plugged all memory leaks detected by valgrind.

src/MC/fix_rmc_partial.cpp
```diff
@@ -0,0 +1,1381 @@
...
+#include "kspace.h"
+#include "modify.h"
+#include "neighbor.h"
+#include "pair.h"
+#include "text_file_reader.h"
+#include "update.h"
+
+using namespace LAMMPS_NS;
+using namespace FixConst;
+
```
akohlmey, line 41: You may want to add here a "citeme" reminder to cite the most relevant publication describing the implementation of this fix.

src/MC/fix_rmc_partial.cpp
```diff
@@ -0,0 +1,1381 @@
...
+    angle_types = new int[num_charge_states];
+    if (rmcdata["angle_types"].size() != num_charge_states) {
+      error->all(FLERR, 3, "Number of angle types doesn't match the number of charge states\n");
+    }
+    for (int i = 0; i < num_charge_states; i++) angle_types[i] = rmcdata["angle_types"][i];
+    angle_list = new int *[num_angles];
+    for (int i = 0; i < num_angles; i++) {
+      angle_list[i] = new int[4];
+      for (int j = 0; j < 3; j++) angle_list[i][j] = rmcdata["angle_list"][i][j];
+      angle_list[i][3] = determine_molecule(angle_list[i][0]);
```
akohlmey, line 329: The original code had an `angle_list[i][4]` here, which is clearly an out-of-bounds access. This may have negatively impacted simulations with that code as that information would not have been stored correctly and thus some careful checking is strongly recommended.

doc/src/fix_rmc_partial.rst
```diff
@@ -0,0 +1,285 @@
...
+nvt or npt.
+
+.. image:: JPG/rmcmd-workflow.png
+   :align: center
+   :scale: 10%
+
+The figure above shows a schematic of the RMCMD cycle, using NVT MD as
+the fix for morphology.
+
+A step-by-step guide on running RMCMD will be made available shortly at
```
akohlmey, line 53: Referring to a website is not our preferred option, since web links tend to rot and go away eventually. Also, any referral to something that *may* happen in the future is not helping users.

I thus strongly suggest to include the step-by-step guide as a "Howto" document in the "Tutorials Howto" section of the LAMMPS manual. At least a rudimentary version for this pull request that may be improved/expanded later.

akohlmey (changes requested):
After looking over some of the details of your fix, it seems to me that it provide some rather specialized functionality for some specific applications and thus would be better be moved out of the MC package and added to the EXTRA-FIX package.

Correspondingly the examples should have their own folder under PACKAGES.

I also noted, that one cannot replace kspace style ewald with pppm without causing crashes due to floating point overflows and related issues. This needs to be documented as a restriction or changes implemented that make PPPM a viable option. It would be best if the code checks for it, too.

The example could be made even shorter. It also needs the logfiles for 1 and 4 MPI processes like with (most of) the other examples.

src/MC/fix_rmc_partial.cpp
```diff
@@ -0,0 +1,1381 @@
...
+  double total_charged_dopants = 0;
+  double total_charged_semiconductors = 0;
+  for (int i = 0; i < num_charge_states; i++) {
+    doping_efficiency[i] = (double) num_dopant_charge[i] / (double) n_dopant;
+    total_charged_dopants = total_charged_dopants + num_dopant_charge[i];
+    total_charged_semiconductors = total_charged_semiconductors + num_semiconductor_charge[i];
+  }
+  double total_doping_efficiency = (double) total_charged_dopants / (double) n_dopant;
+
+  // Write out COM for all molecules, for RDF/morphology or other CG uses
```
akohlmey, line 1278: Why write out a file with coordinates in a custom format?
Why not use the LAMMPS native format?
Or at the extxyz format?
Both of which have options to store the box dimensions?

sjplimp (changes requested):
I would like to review this PR when it is closer to merging

akohlmey:
@vishnu2709 due to inactivity (nearly two months) and lack of feedback, I will close this pull request now. Please resubmit when you have addressed the concerns and fixed the issues.

PR #4696, opened 2025-08-14
Add support to render regions to dump image

akohlmey:
**Summary**

This pull request implements something similar to the region2vmd command into dump image.
This feature is also exposed to the LAMMPS-GUI image viewer for convenient testing/debugging of regions.

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

Since the LAMMPS internal render facility is lacking some of the primitives of VMD, rendering is somewhat limited. Each object can be either rendered as a filled object, a wireframe, or a point cloud. For filled objects, the open faces information is respected.  Only primitive regions can be rendered as "filled" or "frame"; point clouds are supported for all types of regions. Supported styles for "filled" or "frame":
- [x] block
- [x] cone
- [x] cylinder
- [x] ellipsoid
- [x] prism
- [x] sphere

image 750

**Post Submission Checklist**


src/dump_image.cpp
```diff
@@ -39,15 +39,268 @@
...
+  // define triangle mesh
+  std::vector<triangle> trilist1;
+  trilist1.emplace_back(triangle{{pos1, pos4, pos5}});
+  trilist1.emplace_back(triangle{{pos5, pos4, pos2}});
+  trilist1.emplace_back(triangle{{pos2, pos4, pos6}});
+  trilist1.emplace_back(triangle{{pos6, pos4, pos1}});
+  trilist1.emplace_back(triangle{{pos5, pos3, pos1}});
+  trilist1.emplace_back(triangle{{pos2, pos3, pos5}});
+  trilist1.emplace_back(triangle{{pos6, pos3, pos2}});
+  trilist1.emplace_back(triangle{{pos1, pos3, pos6}});
```
rbberger: You can simplify this by either just doing:
```c++
trilist1.push_back({pos1, pos4, pos5});
```
or even better:
```c++
std::vector<triangle> trilist1 = {
    {pos1, pos4, pos5},
    {pos5, pos4, pos2},
    {pos2, pos4, pos6},
    {pos6, pos4, pos1},
    {pos5, pos3, pos1},
    {pos2, pos3, pos5},
    {pos6, pos3, pos2},
    {pos1, pos3, pos6}
};
```

src/dump_image.h
```diff
@@ -101,6 +101,8 @@ class DumpImage : public DumpCustom {
 
   class Image *image;    // class that renders each image
 
+  std::vector<class RegionInfo *> regions;
```
rbberger: Why not just make this `std::vector<class RegionInfo> regions;`?

src/dump_image.cpp
```diff
@@ -39,15 +39,268 @@
...
+namespace {
+
+using LAMMPS_NS::MathConst::MY_2PI;
+constexpr int RESOLUTION = 20;
+constexpr double RADINC = MY_2PI / RESOLUTION;
+
+using vec3 = std::array<double, 3>;
+using triangle = std::array<vec3, 3>;
+
+vec3 vecadd(const vec3 &a, const vec3 &b)
```
rbberger, line 72: This should probably all go into something like `geom.h`/`geom.cpp`, since it might be useful in multiple contexts.

akohlmey: We already have "math_extra.{h,cpp}". This is a bit of a testbed to see, if the functions there could be replaced. It just requires too much work to do all that right now.

src/dump_image.cpp
akohlmey: done

src/dump_image.h
akohlmey: I wanted to avoid putting the class definition into a style header and violate our own rules about keeping the namespace clean. However, I realized that I could put the class definition *inside* the DumpImage class and then the style header rules don't apply.

This also provides an opportunity to put re-usable functionality into a function and simplify the code a little bit.

src/dump_image.cpp
```diff
@@ -36,17 +36,253 @@
...
+        tri[i][2] = tri[i][2] * scale + center[2];
+      }
+      img->draw_cylinder(tri[0].data(), tri[1].data(), color, diameter, 3);
+      img->draw_cylinder(tri[0].data(), tri[2].data(), color, diameter, 3);
+      img->draw_cylinder(tri[1].data(), tri[2].data(), color, diameter, 3);
+    }
+  }
+
+  if (level >= 2) {
+    trilist = std::move(refine_triangle_list(trilist));
```
rbberger: std::move not necessary, it's done automatically.

src/dump_image.cpp
```diff
@@ -36,17 +36,253 @@
...
+{
+  std::vector<triangle> outlist;
+  for (const auto &tri : inlist) {
+    vec3 posa = vecnorm(vecadd(tri[0], tri[2]));
+    vec3 posb = vecnorm(vecadd(tri[0], tri[1]));
+    vec3 posc = vecnorm(vecadd(tri[1], tri[2]));
+    outlist.emplace_back(triangle{{tri[0], posb, posa}});
+    outlist.emplace_back(triangle{{posb, tri[1], posc}});
+    outlist.emplace_back(triangle{{posa, posb, posc}});
+    outlist.emplace_back(triangle{{posa, posc, tri[2]}});
```
rbberger: `push_back({a, b, c})`

src/dump_image.cpp
```diff
@@ -36,17 +36,253 @@
...
+        img->draw_cylinder(tri[0].data(), tri[2].data(), color, diameter, 3);
+        img->draw_cylinder(tri[1].data(), tri[2].data(), color, diameter, 3);
+      }
+    }
+  }
+
+  if (level == 4) {
+    trilist = std::move(refine_triangle_list(trilist));
+
+    for (auto tri : trilist) {
```
rbberger: `auto&`

src/dump_image.cpp
```diff
@@ -36,17 +36,253 @@
...
+        img->draw_cylinder(tri[1].data(), tri[2].data(), color, diameter, 3);
+      }
+    }
+  }
+
+  if (level >= 3) {
+    trilist = std::move(refine_triangle_list(trilist));
+
+    if (level == 3) {
+      for (auto tri : trilist) {
```
rbberger: `auto&`

src/dump_image.cpp
```diff
@@ -36,17 +36,253 @@
...
+        for (int i = 0; i < 3; ++i) {
+          auto scale = radscale(radius, tri[i]);
```
rbberger:
```
auto & t = tri[i];
t[0] = ...
```

src/dump_image.cpp
akohlmey: Thanks for the suggestions and thorough review. I've went one step further and wrapped the redundant code to resize and move to a function.

rbberger (approved):

PR #4698, opened 2025-08-16
fix pimd/langevin: support the Nose-Hoover chain thermostat

Lily200202:
**Summary**

This PR extends the thermostat options in `fix pimd/langevin` by adding Nose-Hoover chain (NHC) thermostat support alongside the existing PILE_L thermostat.

**Related Issue(s)**

**Author(s)**

Li Fu
fuli200202@163.com

**Licensing**

**Backward Compatibility**

No backward compatibility will be affected. Existing inputs using PILE_L thermostat will continue to work without modification.

**Implementation Notes**

The source code changes are entirely confined to the fix files (fix_pimd_langevin.h and fix_fix_pimd_langevin.cpp). No other changes to any core lammps classes have been made. It is added to the REPLICA package.

**Post Submission Checklist**


src/REPLICA/fix_pimd_langevin.cpp
```diff
@@ -54,14 +54,17 @@ using MathConst::MY_SQRT2;
...
 
-namespace {
-std::map<int, std::string> Barostats{{FixPIMDLangevin::MTTK, "MTTK"},
-                                     {FixPIMDLangevin::BZP, "BZP"}};
-std::map<int, std::string> Ensembles{{FixPIMDLangevin::NVE, "NVE"},
-                                     {FixPIMDLangevin::NVT, "NVT"},
-                                     {FixPIMDLangevin::NPH, "NPH"},
-                                     {FixPIMDLangevin::NPT, "NPT"}};
-}    // namespace
+enum { PIMD, NMPIMD };
```
Yi-FanLi: No need to add lines 57~64 as they are already in fix_pimd_langevin.h.

Yi-FanLi (commented):
I did some changes in https://github.com/Lily200202/lammps/pull/1.

akohlmey:
@Lily200202 @Yi-FanLi Can you please explain why there needs to be a Nose-Hoover chain thermostat in a *langevin* fix? What about using fix pimd/nvt? Looking at the documentation does not provide any help either. It merely lists the flag, but no explanation about the differences of the algorithms or no pointer to a paper where to get further information.

akohlmey:
@Lily200202 @Yi-FanLi due to inactivity and lack of feedback, I will close this pull request now. Please resubmit when you have addressed my concerns and can provide a convincing justification for the change.

Yi-FanLi:
@akohlmey 
Sorry not not responding sooner. I was processing the other few PRs about `fix pimd/langevin` and did not have the time to touch this one last week. 

Using pimd in the Nose-Hoover chain thermostat has been a long-standing issue with us. Several attempts have been made to test the `fix pimd/nvt` but suspicious results have been found by many people -- we were not able to generate the correct nvt distribution from the `fix pimd/nvt` implementation. That was the starting point of having `fix pimd/langevin`. 

The original `fix pimd/langevin` supports correct sampling with the `PILE_L` thermostat. We benchmarked it with `i-PI` to jusitify its correctness. Since we have the keyword `thermostat` in this fix style, we do have the possibility of supporting multiple thermostats. 

In some cases people still want the Nose-Hoover chain thermostat, for example, applications related to quantum dynamics. But we still have the issue that the thermodynamic distribution sampled by `fix pimd/nvt` does not agree with that from `fix pimd/langevin` and `i-PI`. Then we chose to support Nose-Hoover chain in `fix pimd/langevin`. That is the motivation of this work.

akohlmey:
> In some cases people still want the Nose-Hoover chain thermostat, for example, applications related to quantum dynamics. But we still have the issue that the thermodynamic distribution sampled by `fix pimd/nvt` does not agree with that from `fix pimd/langevin` and `i-PI`. Then we chose to support Nose-Hoover chain in `fix pimd/langevin`. That is the motivation of this work.

Sorry, but that is not how we want to do development in a community code. If there a problems with a feature in the code, they need to be documented, reported, and - if possible - fixed.
Hacking your own code, is something you can do for an in-house code, but for a community code, you can at best use that internally to have a reference and then you need to work in resolving the issue in the existing code. We don't want to end up with N+1 implementations of the same feature in different locations and in places where people do not expect them.

Please understand that having a contribution in a community code is a privilege and a responsibility. You cannot treat it like an in-house code.

Sorry, if that sounds a bit harsh, but I have to look out for the long-term viability of the package and not just addressing some immediate issue.

Bottom line: of course you are free to implement what you want into your local version of LAMMPS, but in the public version, the constraints are stricter.

Yi-FanLi:
Hi @akohlmey, thank you so much for sharing your concerns and requirements. That doesn't sound harsh at all! I totally agree with what you said. Definitely I don't want to treat a community code like my in-house code. I only target here is to make sure that the PIMD code in LAMMPS is robust and its features are enough for users. 

The point here is that we need to have a pimd code that works fine for the Nose-Hoover chain thermostat. If I had as much knowledge about the source code of LAMMPS dating back to when I started working on `fix pimd/langevin` as I do now, I would definitely first figure out the issue of the existing implementation -- although it has been out of maintenance since about 2010 after it was added. 

According to some discussions dating back to 2023, we decided to make `fix pimd/langevin` separate from the previous implementation. This practice was appropriate at that moment and comes out successful. I have witnessed `fix pimd/langevin` being useful for several research projects since then. 

However, having these two separate `pimd` styles is only a temporary solution and is not an optimal way of having PIMD in LAMMPS. Also, the`fix pimd/nvt` is not the most ideal naming -- the Nose-Hoover chain thermostat can definitely be used to support NpT. 

Ideally, different thermostats and barostats should share the NVE part of the velocity Verlet integrator, the code for calculating outputs, and the inter-replica communication code. Then individual thermostats can then be implemented either as derived classes or as separate functions.

With this said, I propose the following improvement:

1. Extract the parts that `fix pimd/nvt` and `fix pimd/langevin` should have in common, and make it a base class called `FixPIMDNVE`. 
2. Derive a class called `FixPIMDNHC` which implements the thermostat part of the current `fix pimd/nvt` code; derive a class called `FixPIMDLangevin` which implements the thermostat part of the current `fix pimd/langevin` code. 

Does this sound like a reasonable approach to you?

akohlmey:
> Does this sound like a reasonable approach to you?

Yes, it does.

PR #4699, opened 2025-08-18
WIP: Various C++17 updates

rbberger:
**Summary**

A bunch of cleanups due to C++17 requirement.

- make use of C++17 `<filesystem>` for various platform helpers
- remove member variables of type `std::map` in `group` and `delete_atoms` and replace them with a better suited `std::set`
- avoid declaring long `::iterator` types and instead use `auto`.

**Related Issue(s)**

**Author(s)**

@rbberger 

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**


akohlmey:
@rbberger it looks like the filesystem changes are not compatible with Windows (native with MSVC++ or cross-compiler with GCC). I've cherry-picked the other changes into PR #4711. Since we already have a platform abstraction of the filesystem functions in the platform namespace moving to the C++ functions would make the most sense, if we can get rid of the functions in platform altogether. That would be a more invasive change and something for later.

But there are other opportunities for improvement. Fix colvars, fix qmmm, and fix imd all use a custom hashtable code. This can likely be replaced with `std::unordered_map`, and there is also `hashlittle.h/.cpp` that could be replaced with `std::hash`.

rbberger:
@akohlmey for MSVC we need to make sure to use a new enough toolset 14.3+.

rbberger:
@akohlmey looks like I just fixed the MSVC build with 8cd6a60. Turns out the standard doesn't define implicit conversions from `std::filesystem::path` to `std::string`. So doing it explicitly makes MSVC happy.

akohlmey:
> So doing it explicitly makes MSVC happy.

Also Rocky Linux 8 wasn't too happy to compile it in legacy more. I had to replace it with Alma Linux 9. RHEL comes with GCC 8.5 anyway.

PR #4707, opened 2025-08-31
In compute stress, use virial instead of thermo_press. Removed repeated line.

chogene:
**Summary**

1. For the stress fluctuation term, only virial stress should be calculated

2. Removed repeated line

**Related Issue(s)**

N/A

**Author(s)**

Eugene L. Cho (Hanyang University)
Email: cho.eugenelee@gmail.com

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

Changed /examples/ELASTIC_T/BORN_MATRIX/Silicon/output.in:

1. Replaced compute stress all pressure thermo_temp with compute stress all pressure NULLvirial.

2. Removed repeated line: variable s33 equal v_s3*v_s3.

**Post Submission Checklist**

N/A


Bibobu:
Hi @chogene,

Thank you for the modification. Concerning your first point, both stress can be used. There is only a factor 2 correction at some point in the post-processing of the data in the diagonal term as far as I know, since the self correlation of the kinetic part should vanish (see for example [this paper from Lutsko, especially eq 33b and 37.](https://www.researchgate.net/profile/James-Lutsko/publication/224513058_Generalized_expressions_for_the_calculation_of_elastic_constants_by_computer_simulation/links/56dd9da208ae46f1e99f81fe/Generalized-expressions-for-the-calculation-of-elastic-constants-by-computer-simulation.pdf)

Both are correct but your correction might be more intuitive to users.

akohlmey:
@chogene @Bibobu we would need to update the log files with the change in the input. Also, would the same or equivalent change be needed in the Argon example?

chogene:
No changes are needed for the Argon example.

sjplimp (changes requested):
would like to look at this more closely before accepting/merging

Bibobu:
> No changes are needed for the Argon example.

@akohlmey is right. As the input file is changed it would require a change in the log file to match the change. This is for consistency.

chogene:
> @chogene @Bibobu we would need to update the log files with the change in the input. Also, would the same or equivalent change be needed in the Argon example?

Should I rename the log file from log.elastic.ortho to log.16May22.ortho.g++.4 ?

sjplimp:
I think @athomps should review this change since the ELASTIC_T script is his.  Since the T means this is calculating elastic constants at a finite temperature, possibly the kinetic energy term in the pressure (stress) should be included.

athomps (changes requested):
Only one of these options can be correct, but nobody has provided evidence for which one it is.  The post processing is hard-coded in output.in here:

`variable C11 equal v_F11+v_B[1]+v_K11`

So which one is right?

Bibobu:
As discussed with Aidan, the original example is correct. If only the virial stress is used, the files require a change of a factor 2 in the kinetic contribution.[^1] We checked that both cases give the same results (~133GPa for C11 and same for other elastic constants).

For simplicity, I'll rewrite the input file and produce the new reference log files.

[^1]: see [this paper](https://www.researchgate.net/publication/224513058_Generalized_expressions_for_the_calculation_of_elastic_constants_by_computer_simulation)

Bibobu (approved):
Should be good with changes made.

athomps (approved):
I agree with analysis and changes of  @Bibobu.  In particular, the original script was correct, but was somewhat redundant.  I made a few more tweaks. It is now better than ever.

athomps:
> @chogene @Bibobu we would need to update the log files with the change in the input. Also, would the same or equivalent change be needed in the Argon example?

I checked the Argon examples, both Analytical and Numerical, and they are already doing things properly. I will add reference log files for these.

akohlmey:
@sjplimp This is now waiting on you.

sjplimp (approved):

PR #4714, opened 2025-09-09
Set version strings for next feature release

akohlmey:
**Summary**

The next LAMMPS feature release is planned for 10 September 2025


stanmoore1 (changes requested):
Last minute bugfix coming in a bit

stanmoore1 (approved):

PR #4718, opened 2025-09-19
Gibbs Ensemble Monte Carlo

andrewykhong:
**Summary**

This adds the Gibbs Ensemble Monte Carlo (GEMC) algorithm to LAMMPS

**Related Issue(s)**

No known issues at this time

**Author(s)**

Andrew Y. K.  Hong (Sandia)

**Licensing**

**Backward Compatibility**

No known issues with backward compatibility

**Implementation Notes**

Includes all major MC moves of the GEMC algorithm (translate, exchange, and box volume). The code requires two boxes which means there must at least be a 1x1 partition. fix_gemc.h/fix_gemc.cpp include the main definitions and basic functions; fix_gemc_mcmoves.cpp has the main MC moves. The implementation was tested on different sized partitions. An LJ fluid with a cutoff of 2.5 was simulated, and the phase diagram was extracted to check for proper phase separation. Points below and  above the critical point were chosen that phase separation occurred when expected.

Only works with atoms right now (no molecules)

I am in the process of cleaning up the code and adding documentation and examples. NOT READY TO PULL (doing now since I need to pass this code to Stan)

**Post Submission Checklist**


stanmoore1:
Thanks @andrewykhong! I will take a look.

sjplimp:
I would like to review this once it is past "draft" status

src/REPLICA/fix_gemc.cpp
```diff
@@ -0,0 +1,514 @@
...
+
+   See the README file in the top-level LAMMPS directory.
+------------------------------------------------------------------------- */
+
+// Gen Questions:
+// 1) What is a tagint?
+// 2) What are all the nullptrs in the definition of a fix
+
+/* ----------------------------------------------------------------------
+   Contributing author: Paul Crozier, Aidan Thompson (SNL)
```
akohlmey: The author contribution line needs to be updated.

src/REPLICA/fix_gemc.cpp
```diff
@@ -0,0 +1,514 @@
...
+   Copyright (2003) Sandia Corporation.  Under the terms of Contract
+   DE-AC04-94AL85000 with Sandia Corporation, the U.S. Government retains
+   certain rights in this software.  This software is distributed under
+   the GNU General Public License.
+
+   See the README file in the top-level LAMMPS directory.
+------------------------------------------------------------------------- */
+
+// Gen Questions:
+// 1) What is a tagint?
```
akohlmey: The `tagint` type is defined in the `lmptypes.h` and is either a 32-bit or a 64-bit integer to store atom IDs.

src/REPLICA/fix_gemc.cpp
```diff
@@ -0,0 +1,514 @@
...
+// 2) What are all the nullptrs in the definition of a fix
```
akohlmey: For all new and refactored code we initialize all class members that are pointers to `nullptr` so you can safely use `memory->destroy()` or `delete[]` on them, even if they have not yet been allocated. When a class is allocated and the memory allocator "recycles" previously used address space, it is not easily possible to tell if the memory has been allocated or not (malloc on Linux is special, that first allocations are all zero bytes so people tend to depend on this, but that is dangerous and can lead to all kinds of problems like use after free or memory corruption or segmentation faults).  For code paths that may be passed multiple times, this simplifies the code by just deleting/destroying it always and thus memory leaks are avoided.

src/REPLICA/fix_gemc.cpp
```diff
@@ -0,0 +1,514 @@
...
+
+// not sure what all these nullptrs are for
+FixGEMC::FixGEMC(LAMMPS *lmp, int narg, char **arg) : Fix(lmp, narg, arg)
+{
+  buf = NULL;
+
+  if (narg < 12) utils::missing_cmd_args(FLERR, "fix gemc", error);
+  // must have only two boxes
+
+  if (universe->nworlds != 2) error->universe_all(FLERR, "Must use exactly two partitions");
```
akohlmey: Since this fix implements both, Monte Carlo and a multi-partition fix, it would need to be added to either the MC package or the REPLICA package. My preference is for the REPLICA package. For sure the source code cannot remain in the src folder. This is reserved for core functionality.

src/MC/fix_gemc.cpp
```diff
@@ -0,0 +1,514 @@
...
+
+  if (narg < 12) utils::missing_cmd_args(FLERR, "fix gemc", error);
+  // must have only two boxes
+
+  if (universe->nworlds != 2) error->universe_all(FLERR, "Must use exactly two partitions");
+
+  if (molecule_flag) error->universe_all(FLERR, "GEMC w/ molecules currently not supported");
+
+  // various fix flags (partial copy from gcmc)
+  time_integrate = 0; // do not time integrate (use only MC moves)
```
akohlmey, line 78: Please see `fix.h` for explanations of these flags and then read other fix commands that use them.

src/REPLICA/fix_gemc.cpp
```diff
@@ -0,0 +1,514 @@
...
+  box_temp = utils::numeric(FLERR,    arg[7], false, lmp);
+  displace = utils::numeric(FLERR,    arg[8], false, lmp);
+  max_volume = utils::numeric(FLERR,  arg[9], false, lmp);
+  max_rho = utils::numeric(FLERR,     arg[10], false, lmp);
+  seed = utils::inumeric(FLERR,       arg[11], false, lmp);
+
+  if (max_rho <= 0.0)
+    error->universe_all(FLERR, "FixGEMC: Max density must be positive number");
+
+  // DEBUG : Test if inputs correct
```
akohlmey: Please remove *all* (commented out) debug code before marking your pull request ready for review.

src/MC/fix_gemc.cpp
```diff
@@ -0,0 +1,514 @@
...
+
+  if (energy_stored > 1E6)
+    error->one(FLERR,"fix gemc: Energy of old configuration big");
+
+
+  for (int i = 0; i < nmoves; i++) {
+    imove = random_universe->uniform();
+
+    if (fabs(energy_full()) > 1E8) {
+      printf("step: %i; prev_step: %i\n", i, prev_step);
```
akohlmey: Please *never* use printf() always output to "screen" and/or "logfile". This is most conveniently done using `utils::logmesg()`.

src/REPLICA/fix_gemc.cpp
```diff
@@ -0,0 +1,514 @@
...
+    error->one(FLERR,"fix gemc: Energy of old configuration big");
+
+
+  for (int i = 0; i < nmoves; i++) {
+    imove = random_universe->uniform();
+
+    if (fabs(energy_full()) > 1E8) {
+      printf("step: %i; prev_step: %i\n", i, prev_step);
+      printf("%i - %g\n", i, energy_stored);
+      error->one(FLERR,"bad energy before");
```
akohlmey: Since you have a multi-replica run, you may need to use `Error::universe_all()` or `Error:universe_one()` in many cases. Please check out some of the other fixes in the REPLICA package.

akohlmey (changes requested):
@andrewykhong Here is a first set of comments and suggestions.

Please note that including a documentation file is a strict requirement and adding at least one small and quick example showcasing the usage (not necessarily generating converged results) highly recommended.

athomps:
I took a close look at this, and it looks promising, most of the basic functionality seems to be there.  I made a test script for a Lennard-Jones system with (rho, T) inside the vapor dome.  However, I failed to get it to run reliably, it failed intermittently in different places.  I fixed a few uninitialized variables but I suspect there are more. In order to make more progress, it will take valgrind or something similar running under MPI (because -p 2xN is required).  @stanmoore1 do you any suggestions?

akohlmey:
> it will take valgrind or something similar running under MPI

I am using valgrind with MPI all the time.
Start by configuring the suppressions from `tools/valgrind` as explained in the README.
Then you can just do something like:
```
mpirun -np 4 valgrind --log-file=valgrind-%p.log ./lmp -p 2x2 -in in.input
```
and then inspect the created log files.

src/REPLICA/fix_gemc.cpp
stanmoore1: Fixed.

src/REPLICA/fix_gemc.cpp
stanmoore1: Fixed.

src/REPLICA/fix_gemc.cpp
stanmoore1: Fixed.

src/REPLICA/fix_gemc.cpp
stanmoore1: Fixed.

akohlmey:
@athomps @stanmoore1 as discussed at the last developer meeting, we should close this pull request and resubmit it when it is further along in the development and ready for a final review.

athomps:
This is now running free of obvious code errors on 2xN partitions.  I will continue to test it, clean it, and document it.

athomps (approved):
I think this is in pretty good shape now.  I added an example with `-p 1 4` that runs pretty fast for LJ vapor/liquid coexistence. I added a doc page.  Caveats are 1) restart not tested 2) vector values output not tested.

stanmoore1 (approved):
Code looks good to me, but I'm not sure if you intended to delete Packages.rst?

akohlmey:
> I added a doc page.

I don't see it. I've updated the branch, fixed the whitespace issue, restored Packages.rst, and completed some other minor integration items.

akohlmey:
@sjplimp You wanted to be notified when it is close to being merged, this point has been reached (pending the inclusion of the missing doc file by @athomps). I have assigned the pull request to you. Please assign it back to me when you are done with it.

athomps:
Doc page added.

athomps:
I think this is ready to be merged now

doc/src/fix_gemc.rst
```diff
@@ -0,0 +1,198 @@
...
+"""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
+
+This fix writes the state of the fix to :doc:`binary restart files
+<restart>`.  This includes information about the random number generator
+seed, the next timestep for MC exchanges, the number of MC step attempts
+and successes etc.  See the :doc:`read_restart <read_restart>` command
+for info on how to re-specify a fix in an input script that reads a
+restart file, so that the operation of the fix continues in an
+uninterrupted fashion.
+
```
sjplimp, line 137: I suggest adding a line or two here about restarting a 2-partition simulation.  I presume that is required, but possibly the 2 partitions do not need to be the same sizes in the restarted run?

sjplimp (changes requested):
see my recent comments

sjplimp:
Thanks @andrewykhong for submitting this new Gibbs ensemble feature.  It's something we've talked about for a while, but never done!  Just a couple of hi-level comments.

(a) Does the command require the first partition be vapor, the second liquid?  Or it doesn't matter? The doc page isn't clear on that to me.

(b) Please add a pointer in the new doc page to the examples/mc/in.gemc.lj script so the user can see how to use this command.

(c) Please add a examples/mc/README.gemc file for in.gemc.lj.  Repeat the instructions on how to run the script with 2 partitions.  And briefly explain how to interpret the output to verify that it is running a "correct" vapor/liquid coexistence model.  The vapor partition only has 4 particles ?

(d) If it is valid to run GEMC as an MC-only model (no MD), then I think its doc page should have a "note" about how to do that efficiently.  You can steal a "note" from the pending PR #4740 which explains it for other MC fixes.

andrewykhong:
Of course! The code seems to have changed a lot recently thanks to the work by @athomps and @akohlmey (also @stanmoore1 was a huge help during the initial development) so what I intended/wrote may be different from what is in the current version.

(a) Both partitions/boxes can be the same phase. This was an important test case early on to make sure there is no phase separation when it was not expected
(b) -(c) I will add these over the next few weeks if this remains missing (currently busy at new job so hard to find time right now)
(d) I am not as familiar/an expert in MD vs MC so forgive me if I ask/say something obvious. I believe this is MC-only. The particle positions are randomly perturbed and the particle velocities aren't used to determine whether new configurations are accepted or not

sjplimp:
Re (d) - there is line in the first paragraph of the new fix gemc doc page which says is fix nvt is defined, then you can do a hybrid MC/MD simlulation.  If that should not be allowed, then the doc page should state that and the code should check and error out.  If it is allowed then I think the efficiency note I mentioned (see the other PR) should be added.

akohlmey:
> (b) -(c) I will add these over the next few weeks if this remains missing (currently busy at new job so hard to find time right now)

Perhaps @athomps can give a helping hand as he had done before. It would be nice to have this merged before the upcoming feature release next week.

athomps:
I have resolved merge conflicts and updated some of the reference files to reduce numerical jitter

athomps (approved):
I approve.

stanmoore1:
@athomps what does this PR lack? I'd really like to get it merged into the next patch release, before it goes stale.

sjplimp (approved):

akohlmey (approved):

PR #4722, opened 2025-09-28
fix pimd/langevin: support running for group

Kehan-Cai-nanako:
**Summary**

The current code of `fix ID group-ID pimd/langevin keyword value` runs for all atoms even if this fix style is set for a target group of atoms in the input file, e.g, `fix pimd/langevin groupA`.

The code of this PR is designed to support running `fix pimd/langevin` for a target group of atoms, by masking atoms appropriately in `fix_pimd_langevin.cpp`, specifically for some functions of integration and computation of physical observables. 

**Author(s)**

Kehan Cai, Princeton University

Yifan Li @Yi-FanLi, Princeton University

**Licensing**

**Backward Compatibility**

No changes in the pull request will break backward compatibility for inputs.

**Implementation Notes**

When performing simulations with `fix ID group-ID pimd/langevin keyword value`, the group-ID should be set as the target group of atoms. 

The correctness was verified by comparing results of the 1 replica/bead PIMD simulations with those of classical MD simulations. Besides, for multiple-replicas/beads situations, temperatures, pressures, virials and energies were checked if they were consistent according to formulas and mutual relations.

**Post Submission Checklist**

[example_PIMD_PR.zip](https://github.com/user-attachments/files/23173441/example_PIMD_PR.zip)


src/REPLICA/fix_pimd_langevin.cpp
```diff
@@ -514,10 +514,15 @@ void FixPIMDLangevin::init()
 void FixPIMDLangevin::setup(int vflag)
 {
   int nlocal = atom->nlocal;
+  int *mask = atom->mask;
   double **x = atom->x;
   imageint *image = atom->image;
   if (mapflag) {
-    for (int i = 0; i < nlocal; i++) domain->unmap(x[i], image[i]);
+    for (int i = 0; i < nlocal; i++) {
```
akohlmey: There is no actual change in this function, the original code should be restored and the comments removed.

src/REPLICA/fix_pimd_langevin.cpp
```diff
@@ -550,7 +555,11 @@ void FixPIMDLangevin::setup(int vflag)
       nmpimd_transform(bufbeads, x, M_xp2x[universe->iworld]);
   }
   if (mapflag) {
-    for (int i = 0; i < nlocal; i++) domain->unmap_inv(x[i], image[i]);
+    for (int i = 0; i < nlocal; i++) {
```
akohlmey: There is no actual change in this function, the original code should be restored and the comments removed.

src/REPLICA/fix_pimd_langevin.cpp
```diff
@@ -565,10 +574,15 @@ void FixPIMDLangevin::setup(int vflag)
 void FixPIMDLangevin::initial_integrate(int /*vflag*/)
 {
   int nlocal = atom->nlocal;
+  int *mask = atom->mask;
   double **x = atom->x;
   imageint *image = atom->image;
   if (mapflag) {
-    for (int i = 0; i < nlocal; i++) domain->unmap(x[i], image[i]);
+    for (int i = 0; i < nlocal; i++) {
```
akohlmey: There is no actual change in this function, the original code should be restored and the comments removed.

src/REPLICA/fix_pimd_langevin.cpp
```diff
@@ -663,7 +677,11 @@ void FixPIMDLangevin::initial_integrate(int /*vflag*/)
   }
 
   if (mapflag) {
-    for (int i = 0; i < nlocal; i++) { domain->unmap_inv(x[i], image[i]); }
+    for (int i = 0; i < nlocal; i++) {
```
akohlmey: Same as above

src/REPLICA/fix_pimd_langevin.cpp
```diff
@@ -710,17 +729,25 @@ void FixPIMDLangevin::post_force(int /*flag*/)
   if (atom->nmax > maxunwrap) reallocate_x_unwrap();
   if (atom->nmax > maxxc) reallocate_xc();
   for (int i = 0; i < nlocal; i++) {
-    x_unwrap[i][0] = x[i][0];
-    x_unwrap[i][1] = x[i][1];
-    x_unwrap[i][2] = x[i][2];
+    // if (mask[i] & groupbit) {
```
akohlmey: There is no actual change in this function, the original code should be restored and the comments removed.

akohlmey (changes requested):
I don't understand the purpose of submitting a pull request with modifications, most of which are then commented out. Why are those changes commented? This seems counter-intuitive given the original description of the pull request. Please explain.

Also, there should be a simple/small example provided that clearly demonstrates the changed functionality.

Kehan-Cai-nanako:
Thank you for reviewing the code.

Initially, I added the changes related to `mask & groupbit` in all places where I believed they were necessary. However, subsequent tests indicated that some of these changes were not needed, so I temporarily commented them out. Specifically, the changes related to mask should not be done in functions related to communications and transformations among processes. I have now removed those commented lines from the code.

I have also revised the summary section of this PR to make it more concise and easier to follow, removing unnecessary details that might have caused confusion. In short, this PR enables the use of `fix pimd/langevin` for a specific group of atoms rather than for all atoms, aligning with the functionality already established for other fix styles.

I hope this clarification is clear. Please let me know if any additional information or adjustments are needed.

akohlmey:
> I hope this clarification is clear. Please let me know if any additional information or adjustments are needed.

As I mentioned before, there should be a simple/minimal example that demonstrates the impact of the changes and that can be used for validation.

Kehan-Cai-nanako:
> > I hope this clarification is clear. Please let me know if any additional information or adjustments are needed.
> 
> As I mentioned before, there should be a simple/minimal example that demonstrates the impact of the changes and that can be used for validation.

I have added an example for the group functionality of `fix pimd/langevin` in the [examples/PACKAGES/pimd/langevin_reduecd_units_group](https://github.com/Kehan-Cai-nanako/lammps/tree/feature/fix_pimd_langevin_mask_atoms/examples/PACKAGES/pimd/langevin_reduced_units_group) folder. The example is also attached in the summary section for your convenience.

megmcca:
@Kehan-Cai-nanako thanks for the example including the run command you've used. i can confirm that the new code compiles and the example runs to completion with 4x1 processors on my local workstation. 

to complete your example, could you also upload output files similar to the other `examples/PACKAGES/pimd/langevin*` directories (the four per-proc log files)?

Kehan-Cai-nanako:
> @Kehan-Cai-nanako thanks for the example including the run command you've used. i can confirm that the new code compiles and the example runs to completion with 4x1 processors on my local workstation.
> 
> to complete your example, could you also upload output files similar to the other `examples/PACKAGES/pimd/langevin*` directories (the four per-proc log files)?

Thanks for testing the code! I have uploaded output log files to the corresponding example folder.

megmcca (approved):
updated log file names, i think it's ready to go! approved.

akohlmey:
> I hope this clarification is clear. Please let me know if any additional information or adjustments are needed.

Sorry, but no. If the fix is applied only on a specific group of atoms, why should it touch *any* data that belongs to *other* atoms except for data that is maintained internally and will not lead to any changes of the non-group atoms? The fact that it does not make a difference for your specific test example is not a sufficient argument.

Please note that those other atoms my be moved by a different fix (e.g. fix move or fix nve).

Kehan-Cai-nanako:
> > I hope this clarification is clear. Please let me know if any additional information or adjustments are needed.
> 
> Sorry, but no. If the fix is applied only on a specific group of atoms, why should it touch _any_ data that belongs to _other_ atoms except for data that is maintained internally and will not lead to any changes of the non-group atoms? The fact that it does not make a difference for your specific test example is not a sufficient argument.
> 
> Please note that those other atoms my be moved by a different fix (e.g. fix move or fix nve).

The data of the non-group atoms are accessed only within the routines responsible for communication and normal-mode transformations. The communication operations, i.e., `inter_replica_communications()`, do not alter any data of the non-group atoms. Regarding the normal-mode transformations, the operations on the non-group atoms effectively involve applying a transformation and its exact inverse transformation to their coordinates, which leaves these atoms unchanged. In summary, no changes are made to the non-group atoms, thereby preserving the correctness of the molecular dynamics simulations.

The only minor issue is that the current implementation introduces some redundant communication and transformation steps for the non-group atoms. This design choice was made to simplify the implementation and maintain code clarity. Specifically, during the communication of atomic information across replicas, atoms are sorted by their tags rather than by indices, which makes it challenging to exclude the non-group atoms efficiently. Implementing such an exclusion would require additional mapping operations, thereby increasing computational overhead and resource usage. Therefore, we chose the current approach as a balanced solution, considering the trade-off between the two routines.

akohlmey (approved):

PR #4731, opened 2025-10-15
reduce forces and torques in FixRigid::Setup using a single MPI call instead of 2 separate calls

jibril-b-coulibaly:
**Summary**

The `FixRigid::Setup()` function computes the forces and torques on rigid bodies. The is currently done by:

1. summing all atom forces from each proc into a `sum` array and reducing it into `all` array
2. summing all atom torques from each proc into a `sum` array and reducing it into `all` array

`sum` and `all` have dimensions `[nbody][6]` but the procedure above:

1. fills the forces into `sum[0..2]` and `MPI_Allreduce` on `6*nbody` `MPI_DOUBLE` (this reduces `3*nbody` `MPI_DOUBLE` that are not used)
2. zeroes `sum` , fills the torques into `sum[0..2]` and `MPI_Allreduce` on `6*nbody` `MPI_DOUBLE` (this reduces `3*nbody` `MPI_DOUBLE` that are not used)

This PR adds both forces and torques into `sum` and only performs one `MPI_Allreduce` call instead of 2

**Related Issue(s)**

N/A

**Author(s)**
Jibril B. Coulibaly (jibril.coulibaly@gmail.com)

**Licensing**

**Backward Compatibility**

Yes

**Implementation Notes**

Similar code is already in place in `FixRigid::compute_forces_and_torques()`

**Post Submission Checklist**


jtclemm (approved):
This looks like a nice simplification to me.

akohlmey:
> Similar code is already in place in `FixRigid::compute_forces_and_torques()`

Thus it should be *even* better from the code maintenance point of view to call that function and remove some redundancy, too. As far as I can tell, what would be required is to move the loop that zeroes out the `langextra` array up so that it is run before `compute_forces_and_torques()` is called.

jibril-b-coulibaly:
Hi @akohlmey,

I did not further refactor because in compute_forces_and_torque, there is the calculation of the gravity forces that is not present in Setup.

The solution you propose would add that gravity calculation to Setup.
Is that an omission in Setup?

Thank you

akohlmey:
> Is that an omission in Setup?

@jibril-b-coulibaly 
Yes, I believe it is. Here is a simple test input deck where gravity is the only force contribution. With this updated pull request both runs produce the same trajectory, without the pressure is different on continuing from step 5 and the trajectories diverge. If I add "pre no" to the second run command in the first run the trajectories become identical again, since the call to `setup()` is skipped.

```
region box block 0 10 0 10 0 10
create_box 1 box

create_atoms 1 single  2.5 5 5
create_atoms 1 single  7.5 5 5
mass 1 1.0

fix 1 all gravity 10.0 vector 0 0 1.0
fix 2 all rigid single gravity 1
dump 1 all custom 1 grav1.lammpstrj id z fz vz

thermo 1
run 5 post no
run 5 post no

clear

region box block 0 10 0 10 0 10
create_box 1 box

create_atoms 1 single  2.5 5 5
create_atoms 1 single  7.5 5 5
mass 1 1.0

fix 1 all gravity 10.0 vector 0 0 1.0
fix 2 all rigid single gravity 1
dump 1 all custom 1 grav2.lammpstrj id z fz vz

thermo 1
run 10 post no
```

jtclemm:
These changes still look good. 

Maybe this is going too far, but why not just a logic check in `compute_forces_and_torques()` before adding Langevin thermostat forces? 

i.e.
```
if (langflag) {
  for (ibody = 0; ibody < nbody; ibody++) {
      ...
  }
}
```

Maybe a small cost savings.

akohlmey:
> Maybe a small cost savings.

@jtclemm I don't think this is worth it. The savings are likely as large as the added overhead. This is a loop over the bodies, not the atoms, and with fix rigid the number of bodies is generally small or you would (or should?) be using fix rigid/small.

jtclemm:
> > Maybe a small cost savings.
> 
> @jtclemm I don't think this is worth it. The savings are likely as large as the added overhead. This is a loop over the bodies, not the atoms, and with fix rigid the number of bodies is generally small or you would (or should?) be using fix rigid/small.

That's fair.

sjplimp:
I agree with @jtclemm that the Lang force section in compute_f_and_t() should be put inside an if test on langflag.  Just like
the gravity loop that follows.  Also, the setting of langforce = 0.0 in setup should be within the same if test.  Ditto for allocation, destruction.  So that langforce is not allocated, initialized, or used if langflag is not set.  This is really for code clarity, not for efficiency.

ndtrung81 (approved):

akohlmey:
> This is really for code clarity, not for efficiency.

@sjplimp I have made the change, but I don't feel like this improves code clarity, rather it makes it more convoluted.

Issue #4726, opened 2025-10-09
[BUG] Molecule JSON vs .molecule_template molecule command parsing causes different simulation outputs


salamaaya:
**Summary**

When the same input script is used but the molecules are provided in JSON (LAMMPS molecule schema) vs the classic *.molecule_template format, the simulation produces different outputs. The two input files are identical except that one references the ```*.molecule_template``` files and the other references the equivalent ```*.json``` molecule files. I expect the two to be equivalent and produce identical simulation behavior.

**LAMMPS Version and Platform**

LAMMPS (22 Jul 2025 - Update 1)
Apple M3 Pro (18GB) with macOS 15.6.1 Sequoia

**Expected Behavior**

Converting a molecule template to the JSON molecule schema preserves atom coordinates, types, charges and connectivity (bonds/angles/dihedrals/impropers). When an input script is otherwise identical and only the molecule file format is changed, the simulation should produce identical behavior and outputs.

**Actual Behavior**

Running the same input script that references .molecule_template files produces different simulation output compared to the input that references the .json files:
  - The run that uses JSON molecules shows different reaction progress (different values in "Temp", "Press", "Density", "f_myrxns[1]", and "f_myrxns[2]" when running ```fix bond/react```).
 
Note: both JSON and molecule_template files (attached below) contain equivalent fields when opened (coords, types, molecules, charges, bonds, angles, dihedrals, impropers). The JSON files were created from the templates, yet the runtime behavior differs.

**Steps to Reproduce**

1. run a simulation with the ```*.json``` files with ```lmp -in in.tiny_nylon.stabilized``` found in the ```examples/PACKAGES/reaction/tiny_nylon_json``` folder.
2. run a simulation with the ```*.molecule_template``` files with ```lmp -in in.tiny_nylon.stabilized``` found in the ```examples/PACKAGES/reaction/tiny_nylon``` folder.
3. observe that the simulation produced different results (compare the results using a diff checker).

I have also attached 2 other examples which can be run against the current ```examples/PACKAGES/reaction/tiny_epoxy``` folder and the ```examples/PACKAGES/reaction/tiny_polystyrene``` folder.

Note: The example with ```tiny_epoxy_json``` produces an error altogether because there may be a bug in the  ```molecule``` command.

**Further Information, Files, and Links**

[tiny_polystyrene_json.zip](https://github.com/user-attachments/files/22799555/tiny_polystyrene_json.zip)

[...]

PR #4735, opened 2025-10-16
Add write_molecule command

akohlmey:
**Summary**

Add a write_molecule command which can output molecule templates in either native or JSON format.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

The write_molecule command is mostly a wrapper of calling functions in the `Molecule` class.
The output of the native format was build by extending and correcting the existing but commented out `print()` method for debugging. The JSON output is implemented by adding a `to_json()` to the molecule class which returns a JSON data structure and thus could be used elsewhere where such a data structure would be needed.

Also a reverse `find()` function was added to the `LabelMap` class.

**Post Submission Checklist**


jrgissing:
The current capability of this command is only converting between molecule file formats, right? Do you have plans to extend this functionality? If so, what are the planned extensions and how could they affect the syntax of this command?

akohlmey:
> The current capability of this command is only converting between molecule file formats, right?

Conversion was only the secondary motivation. I primarily wanted to have an option to check the internal data against the input. This is basically the long way to debug issue #4726 and similar issues in the future.

In the process, there was some functionality added to LabelMap and Molecule, that may have use beyond this command.

> Do you have plans to extend this functionality? If so, what are the planned extensions and how could they affect the syntax of this command?

One possible extension could be to also output data that was auto generated. Right now there are flags added so that only data that was input is written out.
But I am not so much thinking about adding to the `write_molecule` command but rather having other ways that would create or modify molecule templates, e.g. creating a molecule from a group. Then write_molecule would allow to output those without having to write the output over and over again. I initially looked into adding this to the `Molecule` class and the `molecule` command, but that is very quirky and so it seemed more straightforward to just do a little wrapper (like write_dump).

jrgissing:
Sounds good, I will update #4640 to use your new labelmap find() function

doc/src/write_molecule.rst
```diff
@@ -0,0 +1,55 @@
...
+
+Write the data from a :doc:`molecule template<molecule>` to a file.
+
+The file format is determined by the file name: if the file name ends
+in ``.json`` the file will be written `JSON format <https://www.json.org/>`_,
+otherwise the file is written in the native LAMMPS file format.
+
+When the molecule template contains multiple data sets, the filename
+must contain a '%' character which will be replaced by the data set number
+so that each data set can be written to a separate file.
```
sjplimp: When the molecule template contains multiple molecules as defined by the "molecule" (link) command, the filename
must contain a '%' character which will be replaced by the molecule number (1 to N ?)
so that each molecule is written to a separate file.

akohlmey: @sjplimp thanks. Made the change.

akohlmey:
> Sounds good, I will update #4640 to use your new labelmap find() function

Well, first this PR needs to be approved and merged. :wink:

jrgissing:
> Well, first this PR needs to be approved and merged. 😉

I should be able to take a closer look tonight!

jrgissing:
The order of bonds, angles, etc. seems to be changed when writing out a molecule template. I do not immediately see an issue with this, but it is more difficult to confirm that the input is the same as the output. Can the order be made consistent? The atom IDs seem to remain in the same order, but note that if that is not guaranteed either, that would mess up 'fix bond/react' because it relies on atom mapping between templates.

jrgissing:
nlohmann::json->dump is alphabetizing the keys and adding a bunch of newlines between elements of arrays, such that the output json look nothing like the examples in lammps/examples/PACKAGES/reaction/tiny_nylon_json. It looks like using   nlohmann::ordered_json might help. Putting all our 'specification description' keys at the top is especially useful for human readability.

akohlmey:
> The order of bonds, angles, etc. seems to be changed when writing out a molecule template. I do not immediately see an issue with this, but it is more difficult to confirm that the input is the same as the output. Can the order be made consistent?

This is primarily due to the JSON library stored data. We can change that by using a different storage container. However, you cannot exactly recover the original order in the file, but the output will be in the order of how data is stored internally and with the different storage container the output will be in the order of how the json class instance is populated. This is consistent with the output for the native format.

For per-atom data the input can have any ordering, on output the data will be sorted by the atom-ID. For example the following "Coords" sections are equivalent:
```
Coords

2    0.75695   0.52032   0.00000
1    0.00000  -0.06556   0.00000
3   -0.75695   0.52032   0.00000
```
and
```
Coords

 1  0 -0.06556 0
 2  0.75695 0.52032 0
 3  -0.75695 0.52032 0
```
but output will always be sorted by atom-ID like the second version.
When using the ordered container in JSON we'll get the same behavior which is an argument in favor of using that setting.

For sections defining bonded interactions, there is a similar behavior due to how the data is stored, which is different for the newton_bond setting on or off. For "Bonds" the output is sorted by the first atom in the bond, for "Angles", "Dihedrals", and "Impropers" the output is sorted by the second atom in the definition.

Finally, also the order of the sections is determined by the C++ code and not how they were ordered on input since that information is lost when filling the data structures.

> The atom IDs seem to remain in the same order, but note that if that is not guaranteed either, that would mess up 'fix bond/react' because it relies on atom mapping between templates.

The mapping of atom-ID to its data is crucial and thus not affected, but - as mentioned above - on output, the data will be sorted by the atom-ID since the internal data structures are indexed by `atom-ID - 1`.

akohlmey:
> Putting all our 'specification description' keys at the top is especially useful for human readability.

I agree with this suggestion. As mentioned in my response to your other comment, the output ordering with the ordered container is determined by when the data is inserted by the C++ code. The exact input formatting and order information is lost. I think that should be the expected behavior. A benefit of using the ordered container is, that we can have the same order of information in both JSON and native output files.

I shall add some clarifying information to the documentation about this.

It seems that the extra newlines can only be avoided by bypassing the JSON class entirely and writing a JSON format file with conventional I/O calls (like for the native format). I am not sure if that is a desired approach, since going via the JSON class has the option to output in more formats than text files including a selection of binary formats.

sjplimp (approved):
approved by me for merging, once Jake's concerns are satisfied

akohlmey:
> adding a bunch of newlines between elements of arrays, such that the output json look nothing like the examples

I would like to point out, that pure whitespace changes can be suppressed on comparison, e.g. `diff` has the `-w` flag, or my preferred graphical comparison tool, `meld` has a preference option to ignore whitespace changes. My next choice, `emacs' ediff-mode` also knows how to handle this. I am certain `vim` has something equivalent. Thus, if there are only whitespace differences and there is consistent ordering, the files can be compared even if the "look different".

jrgissing:
> It seems that the extra newlines can only be avoided by bypassing the JSON class entirely and writing a JSON format file with conventional I/O calls (like for the native format). I am not sure if that is a desired approach, since going via the JSON class has the option to output in more formats than text files including a selection of binary formats.

Looks like someone provided a solution to this, but the PR has been sitting for a while:

https://github.com/nlohmann/json/pull/3841

jrgissing (approved):

PR #4736, opened 2025-10-17
Enhancment to fix store/state to store recent history

sjplimp:
**Summary**

Add an optional "history" keyword to fix store/state so that multiple recent timesteps of per-atom data can be stored with each atom.  The motivation is to enable new kinds of computes, like a compute vacf/history, to be written which uses this data to perform analyses using recent data about each atom.

Also added support to compute property/atom for extracting history from fix store/state, so it can be dumped or accessed from other commands, e.g. variables.

So (1) and (2) in the list below are implemented, but not (3) or (4).

**Related Issue(s)**

N/A

**Author(s)**

Steve, in consultation with Aidan, who has some use cases in mind.

**Licensing**

**Backward Compatibility**

Should not alter behavior of fix store/state when not using the new optional keyword.

**Implementation Notes**

The multiple snapshots of per-atom history are stored in 3d array (i.e. a tensor).  Since the data needs to persist with atoms and migrate with them to a new owning processor, a large history could consume a lot of memory or slow down a simulation.  Still thinking about how to best expose this data to either code (e.g. a new compute) or output by a user.  See (1) and (2) below.

Here is a list of things to think about:

(1) Added an extract() method which allows caller code to access all the history params and data.  Maybe others have ideas of alternate ways to enable this.

(2) Thinking of adding support in compute property/atom for values of the form history[N][M] and an optional keyword "history fix-ID".  Fix-ID = the fix store/state ID to access the history.  N = which history frame (1 = most recent, Nframes = oldest).
M = which value in the history frame (1 to # of values).  This would allow the input script to dump history values or access them in a per-atom variable or process them for thermo output.

(3)  Could add an example or two of a compute which uses history data.  E.g. compute msd/history or compute msd/history/atom.  Or Aidan's compute vacf/history idea.  Not sure if better to do that as a new stand-alone compute, or as an added "history" option to compute msd and compute msd/atom.  This would illustrate how to use the extract() method and write a loop to process the history data.  As would the code in (2) for compute property/atom.

(4) It's possible that fix store/state with history would enable an idea we've talked about before.  Which is diffusion coeffs (via MSD) that achieve better time averaging by continually using multiple starting points (as a run proceeds) for the MSD calculations.

Test script:

[in.history.txt](https://github.com/user-attachments/files/23812064/in.history.txt)

**Post Submission Checklist**

Here is a simple input script
[in.history.txt](https://github.com/user-attachments/files/22961109/in.history.txt)
 which uses the new feature.  If a DEBUG section in the code is uncommented it will
produce screen output (in serial or parallel) which shows it is working (at least for this example).


sjplimp:
@athomps @akohlmey @stanmoore1 @jtclemm - I'm ready for any feedback you have on this PR.  See the list of implementation ideas above (1-4).  Can someone loop in Anders - I do not know his GitHub name.

jtclemm:
@anjohan

This looks neat. One immediate question, how does this handle new atoms being created? I could be mistaken, but without set_arrays() I would think this means newly created atoms could inherit old junk data.

Not sure if better to do something like zero all historical data or copy the current state everywhere for new atoms, but would be good for the documentation to clarify.

doc/src/fix_store_state.rst
```diff
@@ -94,6 +102,56 @@ steps.
...
+Nfreq.  This can be useful to limit storage of atom attributes to only
+the time windows when the attributes are needed.  For example, this
+command in the Examples section above:
+
+.. code-block:: LAMMPS
+
+   fix 2 all store/state 0 vx vy vz history 5 20 1000
+
+will store attibutes once every 5 timesteps repeated 20 times
+(including the current timestep), but only preceeding timesteps which
```
athomps: Should be "preceding."

athomps (approved):
@sjplimp This looks really great. I only noticed 1 typo in the doc page, should be preceding.

athomps (approved):
I approve

sjplimp:
@akohlmey This is ready for you to take a look at.  There are a couple NOTEs for you in compute_property_atom.cpp.  Also a couple additional notes in its init() which are more generic.  Not sure we are consistent about the latter 2 notes everywhere else.

Also, could not whitespace check this on my Mac laptop :(

sjplimp:
@akohlmey I still need to address Joel's comment about adding set_arrays() to fix store/state.  Ditto for modifying src/APIP/atom_vec_apip.cpp to work with compute property/atom like other atom styles, which I noticed while working on compute property/atom.

sjplimp:
@akohlmey Added.a simple test script to the Overview above, which you can use for testing.
I can also not test valgrind or parallel on my Mac laptop :( -- so there may still be bugs.

akohlmey:
> @akohlmey This is ready for you to take a look at. 

@sjplimp I am not planning to work on this or any other pull requests that are not ready for merge until after the feature release. And even then I plan to focus on paper writing and changes related to them.

akohlmey:
> @akohlmey This is ready for you to take a look at. 

@sjplimp I am not planning to work on this or any other pull requests that are not ready for merge until after the feature release. And even then I plan to focus on paper writing and changes related to them.

athomps:
@sjplimp Using these new history features, I created a fully inline sampling of the velocity time correlation function for particles as they pass through a shock front.  Up to this point, this required running the simulation *once* to record the shock timestep for each particle (using `fix set`), running the simulation *again* to dump out the velocity history for each particle, and then *post-processing* the dump file to generate collisional velocity correlation function CVCF(t).  The new script does everything in one simulation, averaging over all atoms at the end.  It is sufficiently non-trivial and interesting that I have added the script to this PR as `examples/PACKAGES/shock/nemd/in.nemd`.
It is still a little verbose and time resolution is hard-coded because `compute property/atom` does not allow using `history[*][1]`.  If that was added, the script could be made much shorter and the time resolution could be varied. Here is a plot comparing CVCF(t) generated using both methods.  The inline method data is taken directly from  'examples/PACKAGES/shock/nemd/cvcf.dat' and so is lower resolution, but it matches the three-step method well and will scale to larger systems.
cvcf_single

jrgissing:
@sjplimp are you willing/interested in seeing how easily 'fix store/state' (including this new history feature) can be extended to support per-atom arrays in addition to per-atom vectors?

sjplimp:
@jrgissing yes, adding support for per-atom arrays could be useful.  However, I'd prefer to do that in a future PR and merge this one as-is.  The future PR could also address Aidan's suggestion for syntax to allow easier access to the history.  Which could also be affected by adding per-atom arrays to fix store/state.

sjplimp:
@akohlmey Note that this PR is not ready to merge yet.  See my comment above about NOTEs for you in a couple places in the source code.  They should be addressed, or removed if you are OK with the code as-is.

akohlmey:
@sjplimp As promised I (finally!) had a look at the code and updated the one parsing section. But I also found a few nasty bugs and unintended side effects, which took a bit longer to resolve. They should be addressed now. Please have a look.

akohlmey (approved):

examples/PACKAGES/shock/nemd/README
```diff
@@ -0,0 +1,25 @@
+This example reproduces the 2010 study of theshockwave structure in a
```
jtclemm: space

examples/PACKAGES/shock/nemd/README
```diff
@@ -0,0 +1,25 @@
...
+fluid of Holian et al.[1]
+
+It runs a non-equilibrium molecular dynamics simulations of strong
+shockwave propagation in a fluid.  The potential is the smoothly
+truncated cubic spline Lennard-Jones proposed by Holian and REvelo
```
jtclemm: double caps on REvelo. I thin it's also Ravelo with an a.

jtclemm (approved):

sjplimp:
@akohlmey all changes look good - thanks
Ready to merge from my perspective

PR #4737, opened 2025-10-17
PKG-LDD 

Maria-Lesniewski:
**Summary**

This is an extension that enables a family of pair additive potentials that are a function of local densities and square gradients of the local density as described in DeLyser (2019) and defined in DeLyser (2021), respectively. 
It has primarily been used for coarse-grained models, specifically those constructed with BOCS (version 5, Lesniewski (2025)) but the package itself is pretty flexible. 
It dually enables a variety of local density based potentials and also a lammps trajectory format that can be used with BOCSv5+ for recovering/constructing such potentials for coarse-graining. 

Where available, links to public works defining/testing this software will be included below. 

**Related Issue(s)**

N/A 

**Author(s)**

The LDD package was written by Michael R. DeLyser @mdelyser and Maria C. Lesniewski @Maria-Lesniewski under the advising of Will G. Noid @wnoid, all from @noid-group based in The Pennsylvania State University.

**Licensing**

**Backward Compatibility**

N/A - not yet thoroughly tested with older compilers/libraries. 
We do not think anything implemented relies specifically on a particular version of e.g. gcc/intel compilers.

**Implementation Notes**

=>**src/LDD Code Summary**<=

The code itself uses an atom_style (atom_style ldd) that tracks local densities and gradients of local densities together with a pair style (pair_style ldd) that calculates local densities and gradients of local densities at each time step.
These quantities are defined based on user-specified indicator functions, which we take from the user in the pair_coeff command.
They are calculated by the pair_style at the begining of the compute function. 
Once the local densities and gradients of local densities are calculated, the corresponding pairwise contributions to the force are calculated depending on the potential/gradient args we took from pair_coeff.
We created factories for creating different kinds of indicator and potential/gradient functions that can be plugged into our pair_ldd function. 
Thus it is fairly easy to implement highly flexible/new definitions of the local densities and square gradient of the local density functions within our package. 
We use these factories to offer a variety of functional forms that have already been implemented by our group and others in past works (see our doc files for more info). 
We also created a dump style (ldd), which prints local density and gradient of local density information for each surrounding particle. 

We have checked this code for correctness in several ways. 
Some have already been published.
Others we intend to publish in an upcoming paper (Lesniewski 2025) that documents/announces the release of BOCSv5.
Here is an overview of this testing:
1) BOCS is a package that is used to construct CG force-fields using the MSCG method (using the force-matching variational principle). 
If there is no coarse-graining and the force functions simulated can be represented exactly with e.g. a linear/spline/delta basis, then BOCS can exactly recover the functional forms of the forces from just per-atom force data. 
Thus, we simulated a series of known LDD interactions in our LAMMPS package, and then used BOCSv5 to recover the simulated non-bonded local density and square gradient of the local density interactions from those simulation statistics. 
We verified that the recovered and simulated force-field functions are exact (within numerical limits). 
Atomic test cases like this were published in DeLyser (2019) and DeLyser (2021) and validate that the forces are implemented correctly.
These simulations were conducted with early versions of atom_style ldd implemented in this package.
These early test cases consisted of 1 or 2 LDD atom types and single atom molecules. 
More recently, we have also verified the correctness of this package for more complex force-fields (more LDD atom types) with molecular atom styles (more branching topologies), wherein the hybrid atom_style is called with ldd. 
These more complex molecular recoveries will be published in Lesniewski 2025. 

2) We simulated a system of 2 particles with LD and/or SG potentials in the canonical ensemble. 
For these systems, we can analytically calculate the distribution of pair distances and local densities.
Thus, we verified that our simulated distributions agree with analytic distributions for a variety of indicator options.
These results will also be published in Lesniewski 2025. 

3) We created several simple systems, for which the pair-wise forces and local densities and gradients of the local density can be calculated by hand. We verified that the code agrees with these simple cases and constructed unit tests based on similar simple cases.

=>**other src files (minor changes)**<=

Where possible we tried to avoid changing any base lammps src files, however we ended up making minor changes to 2 base src files. 
If we need to avoid these changes and the developers have suggestions for doing so, we would be glad to implement those suggestions.

*Context for these changes:* 
LDD potentials impose pair additive forces that can be defined between each pair of particle types $\alpha$ and $\beta$.  
Like other force functions implemented in LAMMPS, we define pair interactions between site types with pair_coeff $\alpha$ $\beta$ where for instance, the LAMMPS input syntax is something like pair_coeff $\alpha$=1 $\beta$=2. 
In particular we opted for syntax that looks like this:

    pair_style ldd 6.5
    pair_coeff 1 2 lucy 0.0 6.5 self no potential quadratic 1 -2 1 gradient constant 1
    pair_coeff 2 1 dpd 0.0 5.0 self no potential linear 1 -1 gradient constant 2

However, for any given pair of particles $I$ and $J$, the pair additive force generally depends on the local densities surrounding both ($\rho_I$ , and $\rho_J$ respectively). 
These local environments and the potentials that govern them are defined with respect to a central particle type, and need not be the same for different central particle types. 
So e.g. the LDD potentials governing particles of type 1 surrounded by particles of type 2 need not be the same as the LDD potentials governing particles of type 2 surrounded by particles of type 1.  
(DeLyser 2019 covers this in depth for local densities, and DeLyser 2021 does the same for gradients of the local density.) 
In order to allow for users to define these interaction types as distinct from one another, in our implementation we use a "surrounded by" convention to interpret the pair_coeff input.
Wherein, e.g., the LDD pair_coeff command with args 1 2 defines the interaction that governs a central particle of type 1 surrounded by particles of type 2.
Likewise, the LDD pair_coeff command with args 2 1 defines the interaction that governs a central particle of type 2 surrounded by particles of type 1. 
Consequently, we made changes to 2 root files such that when LDD is used, pair_coeff 1 2 is treated as distinct from pair_coeff 2 1.
We do not think these changes impact other functions. 

*Change to input.cpp*:

It seems that sometime in 2021 input.cpp changed how pair_coeff args are processed.
Specifically, the code now reorders the user arguments such that pair_coeff 1 2 and pair_coeff 2 1 are treated equivalently. 
I added a flag to check if the pair_coeff being processed is for the LDD package.
We skip this reordering if the pair_coeff command invoked corresponds to the LDD potentials. 

*Change to pair_hybrid_overlay.cpp*

This code can be used with or without regular pair potentials. 
To add LDD potentials to regular pair potentials, pair_style hybrid/overlay is used. 
In order for the ij arg loop to treat hybrid commands involving the LDD pkg potentials correctly, we added a simple flag to check if our package is used.
If the pair_coeff command is invoking the LDD package, this flag changes the default loop behavior to one where the * syntax treats the order of user args. (e.g. such that 2 * is different from * 2).

=>**unit tests added**<=

Forces: 

We have included a battery of unit tests using the force-style test .yaml framework with some simple systems for which we could calculate forces by hand.
We then used the test function to generate the reference data in each .yaml file, and used these tests to monitor the correctness of our software as we made changes to better comply with the requirements on the LAMMPS website, e.g., coding style standards. 
Because of their pair additive nature, we listed these under files with the syntax "mol-pair-ldd*.yaml" with further dashes/variants for the different args possible and sub functions tested.

atom style:

We have also modified test_atom_styles.cpp by adding a few test fixtures to verify that the members of the base atom_style ldd class are set up correctly when LDD is installed. These test fixtures are also simple systems with analytically determined solutions for the total local density/gradients of local density. 

=>**doc/src**<= 

We have added files for documenting this package, including how to use this package and more details on the force-fields implemented.
This documentation and its linked components can be primarily viewed from Howto_ldd.rst, pair_ldd.rst and dump_ldd.rst, which users will probably access from the packages page of the website. 

**Post Submission Checklist**

The user can specify local density and square gradient of the local density potential. In practice, we determine these potentials via force-matching all-atom sims with the BOCS software package. 
The package can be found at [this link](https://github.com/noid-group/BOCS/), while the original paper is [here](https://pubs.acs.org/doi/10.1021/acs.jpcb.7b09993).

Here are the referenced prior publications that define the potentials in more depth and also implemented/tested simple systems while the package was being developed:
[DeLyser (2019)](https://pubs.aip.org/aip/jcp/article/151/22/224106/198468/Analysis-of-local-density-potentials)
[DeLyser (2021)](https://pubs.aip.org/aip/jcp/article/151/22/224106/198468/Analysis-of-local-density-potentials)

Here are some additional prior works that have used this package during its development:
[DeLyser(2017)](https://pubs.aip.org/aip/jcp/article/147/13/134111/195655/Extending-pressure-matching-to-inhomogeneous)
[DeLyser(2020)](https://pubs.aip.org/aip/jcp/article/153/22/224103/973527/Bottom-up-coarse-grained-models-for-external)
[Szukalo(2023)](https://pubs.aip.org/aip/jcp/article/159/7/074104/2906863/A-temperature-dependent-length-scale-for)

We are currently preparing the following manuscript to describe and demonstrate the LDD package and the corresponding force-field parameterization software in our new version of BOCS. 
The manuscript authors and title are listed below:

Maria C. Lesniewski, Michael R. DeLyser, W.G. Noid "Progress toward a better BOCS: systematic coarse-graining with local density potentials" (2025) *Manuscript in prep*


akohlmey:
> It seems that sometime in 2021 input.cpp changed how pair_coeff args are processed.
> Specifically, the code now reorders the user arguments such that pair_coeff 1 2 and pair_coeff 2 1 are treated equivalently.

FYI, previously, a `pair_coeff 2 1` command would not be accepted at all. The change was required to facilitate the use of type labels instead of numerical types.

> Consequently, we made changes to 2 root files such that when LDD is used, pair_coeff 1 2 is treated as distinct from pair_coeff 2 1.
> We do not think these changes impact other functions.

I am sorry, but in my personal opinion, those two changes are not acceptable. We want to keep the core code paths free from any "do this different for feature X" kind of hacks as much as possible, especially if X is an optional feature.

Since your potential is pair-wise, but not symmetric, I would suggest to use an approach similar to what is done with many-body potentials (which in essence your model is, too). You could use a single `pair_coeff * * <parameter_file> <type1> <type2> <type3>` command and then have the corresponding parameters set in the file similar to your current implementation.
In the file you would use symbolic types and then have settings for all supported pairs of types. You would then store the ordered parameter to pair of types mapping internally.

Maria-Lesniewski:
Hello! 

Thank you for your suggestion. 
This new implementation I pushed uses the framework that you suggested so that we can avoid making changes to the mainline of the code and basically keeps our old implementation in tact.

Now we parse input in the in.lammps_input file like e.g. a system with 2 types: 

    atom_style ldd 2
    pair_style ldd 6.5 
    pair_coeff * * input.txt  # (an entry example with no element map)
    pair_coeff * * symput.txt A B # (an entry example with an element map)

where e.g. input.txt contains lines for setting the pair parameters with our old syntax like: 

    pair_coeff 1 2 lucy 0.0 6.5 self no potential quadratic 1 -2 1 gradient constant 1
    pair_coeff 2 1 dpd 0.0 5.0 self no potential linear 1 -1 gradient constant 2
[Here in these custom input files, I process the input syntax for the whole system with my own read_file and coeff_ldd commands so that I can leave the main line alone.]

I also added symbolic type compatability so that e.g. symput.txt can look like:

    pair_coeff A B lucy 0.0 6.5 self no potential quadratic 1 -2 1 gradient constant 1
    pair_coeff B A dpd 0.0 5.0 self no potential linear 1 -1 gradient constant 2

So far I have checked that this new implementation syntax gives me the same log file results as my old syntax does for the examples under examples/PACKAGES/ldd. 

Please let us know if this approach is acceptable. 
If so, then I will go ahead and update the rest of our contribution accordingly (e.g. docs/unit tests/clean up white spaces etc.). 

Thank you for your time and help, 

Maria

akohlmey:
> Please let us know if this approach is acceptable.

I cannot tell immediately. This is too big a package with too many changes.
From a quick glance, it looks like you have not fully understood what I was suggesting.

But I am also seeing other things that will likely need changes or at least clarification and discussion and I see code that looks like it was "inspired" by code that we have by now modernized and refactored.

There are two different strategies to proceed.

1. You let me do part of the work and then I ask you to complete it
2. I add only comments and requests for changes and you will have to work your way through it.

Option 1. is less work for both of us, but you have to trust me to not mess things up.
Option 2. is more "educational" but a lot more work for both of us, especially for me.

I had asked the same kind of question in another pull request review recently and then my first pass reviewing the pull request had something like 40 individual comments. https://github.com/lammps/lammps/pull/4694#pullrequestreview-3347522982
There will be more iterations needed since at this point I have not even attempted to compile LAMMPS with the package included and run some examples. Checking for memory leaks and fixing them can become a lot of work sometimes.

Maria-Lesniewski:
Hello, 

Thank you for your quick response and your willingness to help us improve our package. 
We understand that this is a large package that will require extensive review and improvements. 

Our last comment “Please let us know if this approach is acceptable” was aimed at addressing the initial concern regarding parsing input in the syntax listed above. Since we had to change the way the input is parsed from the user to avoid any changes to main src files, this means that the input shown in our manual pages and .yaml files for the unit tests will need to be adjusted. There are a lot of these, so if there was something fundamentally wrong with the syntax, we were looking to get a heads up before redoing it all. In hindsight, we realize that this is probably not a reasonable thing to ask since the package itself is large and probably makes more sense to review as a whole. So we are happy to update the input in these files even if it needs to be changed later. 

As far as how to proceed, we would be happy to take option 1 if it generates less work for everyone. However, before we pursue this option, please let me update the input for the unit tests I wrote, the examples, and the documentation. I checked that these were correct with the old input, and I suspect it will be useful to start off changes from a base copy of this package that can run / pass these tests with the current input. 

Thank you and please let us know if this is an acceptable path forward,

Maria

akohlmey:
> Our last comment “Please let us know if this approach is acceptable” was aimed at addressing the initial concern regarding parsing input in the syntax listed above. Since we had to change the way the input is parsed from the user to avoid any changes to main src files, this means that the input shown in our manual pages and .yaml files for the unit tests will need to be adjusted. There are a lot of these, so if there was something fundamentally wrong with the syntax, we were looking to get a heads up before redoing it all. In hindsight, we realize that this is probably not a reasonable thing to ask since the package itself is large and probably makes more sense to review as a whole. So we are happy to update the input in these files even if it needs to be changed later.

I will likely want to change that to be more along the lines of what LAMMPS users expect.
We can worry about the tests later. For now I have two items that you can work on while I am busy with other tasks.

doc/src/ldd_indicator_dpd.rst
```diff
@@ -0,0 +1,43 @@
+.. index:: pair_coeff ldd indicator dpd
```
akohlmey, line 1: Having all these different ldd_*_*.rst files as separate files and with indexing etc seems like overkill. 

Please have a look at pair style granular (both implementation and documentation).
It has a somewhat similar architecture where you have classes derived from `GranSubMod` that implement different model variants that can be selected via keywords. Those are all documented in the page for pair style granular. I think something similar would be more suitable here and make the documentation more compact and accessible (instead of being scattered).

src/BOCS/dump_ldd.cpp
```diff
@@ -0,0 +1,649 @@
+/* ----------------------------------------------------------------------
```
akohlmey: Why do we need an ldd dump style?

You could make the custom per-atom properties from atom style ldd accessible through compute property/atom (there probably needs to be a -D LMP_LDD define so the code is only included when the package is installed, which in turn will need some tweaking of the Install.sh script for GNU Make builds) and then you could use dump style custom or cfg or yaml

src/BOCS/pair_ldd.h
```diff
@@ -0,0 +1,109 @@
...
+  void settings(int, char **) override;
+  void coeff(int, char **) override;
+  void init_style() override;
+  double init_one(int, int) override;
+  void write_restart_settings(FILE *) override;
+  void read_restart_settings(FILE *) override;
+  double single(int, int, int, int, double, double, double, double &) override;
+  void coeff_ldd(int narg, char **arg);
+
+  /* I do this the same way it's done in force.h */
```
akohlmey, line 54: > /* I do this the same way it's done in force.h */

Not anymore. We have dropped `typedef` (which is a C feature) and replaced it with `using` (which is a modern C++ feature).

But putting this all in the header is overkill. I think most of this could be put into the pair_ldd.cpp file into an anonymous namespace and using a static initializer for the factory maps. This is not something like pair styles or bond styles where there is a large number of optional ones and there will always be new variants added and including them would need to be automated as much as possible.

As written in the programmer's guide, we want to keep header files lean and avoid potential namespace collisions as much as possible. This is sometimes difficult to achieve for legacy code, but with new additions, this is certainly a goal. If some
simple static declaration does not suffice, we can always go back to the PIMPL programming pattern.

This is probably the second change I would like to make myself after refactoring the processing of the `pair_coeff` command.

akohlmey (changes requested):
Here are three topics that stood out from just looking at the pull request.

doc/src/ldd_indicator_dpd.rst
Maria-Lesniewski: Okay, I took a look at this doc page and  some source files. I will make an attempt to compress the information in the indicator/potential documents into one pair_ldd page this coming week and update this branch with that commit when I have it. As a placeholder for the lammps input file syntax in the examples I will put the old syntax from the original doc together for now and I can update it as we go.

src/BOCS/dump_ldd.cpp
Maria-Lesniewski: We have developed a package, BOCS, for parameterizing CG potentials and analyzing simulations. 
BOCS uses the information in the ldd dump file and is designed to read this specific format. 
So the purpose of dump ldd style was to automatically format this info a file with fixed form so that it'd be easier for BOCS users who are newer to LAMMPS to generate a trajectory with LDD info pre-formatted such that it can be correctly read by BOCS with limited room for user error. 

 I think what you're saying is that we could basically get this format using dump custom id x y z vx vy vz fx fy fz c_ldd_local_density_id[\*] c_ldd_square_gradient_id[\*] etc. if I write 5 different compute functions for accessing the 5 different per atom LDD fields. I think I can see how this would work.
 If you guys would prefer not to have another dump style I can work at getting this approach to work and if necessary rework our BOCS reader and some directions for our users.
 I need to learn more about implementing computes to do this, so it may take me some time to work out and doubly verify in BOCS, but I can get on this.

doc/src/ldd_indicator_dpd.rst
akohlmey: Perhaps you should hold off with the documentation updates after I had a closer look at the pair_coeff processing. That way you could to the reorganization and updates in one go.

src/BOCS/pair_ldd.h
Maria-Lesniewski: Thank you for this explanation, I think I understand what you're saying and this sounds good to me.
If it is easier for you to make these changes we would be thankful for them.

src/BOCS/dump_ldd.cpp
akohlmey: > if I write 5 different compute functions for accessing the 5 different per atom LDD fields. I think I can see how this would work.

That is *not* what I am saying. I think you can just add access to the properties in the atom style to the *existing* compute property/atom compute. There is a small uncertainty since this is by default not set up to process optional atom styles. I have to discuss with @sjplimp whether we should move the atom style to `src` or use a define line `-D LMP_LDD` to make all references to the ldd atom style conditionally compiled.

In addition, please note that in custom dump styles, you can modify the text used in the column headers with dump_modify (so you have something more meaningful than `c_ID[1]` and `c_ID[2]` etc.). If this is well explained in the LDD package howto this should work also for new users and as a bonus you gain support for different dump style formats, especially easy to parse ones in general formats like YAML or NetCDF or HDF5.

Maria-Lesniewski: Okay thank you for your explanation, this sounds good to me. 
Thank you also for clarifying, I am going to get more familiar with compute_property_atom.h/.cpp and go back through the atom class doc so I can make this happen.

Maria-Lesniewski: Hello again!  I reviewed the atom_vec documentation again and also looked at compute_property_atom. 
It seemed like the way to let this function access my atom_style properties was to write property_atom / pack_property_atom functions for our ldd atom_style, so thats what I have tried here.
Similar to other atom styles that I saw do this, I wrote a property_atom function that takes different keywords from the user and determines an index for the pack_property_atom function.
Like other functions, I use this index to determine which variables should be packed.  
Most of atom_style ldd's members are per_atom vectors though, so my indices determined by property_atom also store the info about which column I want to print based on the keyword phrase passed. e.g. I expect commands like  

    compute ldds all property/atom ldd_local_density1 ldd_local_density2 
    compute gradzs all property/atom ldd_grad_densityz1 ldd_grad_densityz2
where e.g. keyword ldd_local_density1 will pack each atom i's atom->ldd_local_density[i][1] data, ldd_local_density2 will pack each atom i's atom->ldd_local_density[i][2] data etc. for each member variable of atom_style ldd. (based on the surrounding atom type passed)

If I need to try a different approach for either accessing the data or the functions I wrote themselves, please let me know and I am happy to try to improve these.

akohlmey:
@Maria-Lesniewski I am sorry, but I don't expect to be able to work on this pull request this week due to other priority items. For any such pull requests (the same applies to the addition of the MBX package pull request) I will need some larger chunk of available time to work on it efficiently.

Maria-Lesniewski:
Hello! 

Thank you again for your time/help working on this package so far.  
I recently realized that I forgot to consider MPI when I adjusted the ldd coeff input to be more like the other manybody pair styles (commit 27147ba).
Specifically, I embarrassingly wrote an MPI hang into the code and forgot to send my input read to the other cores for setting up the forces.
I fixed this by replacing some of my code with utils::read_lines_from_file and then parsing the broadcasted strings on each core. This seems to fix the hanging, and I verified that I could reproduce my multi-core output from the original code contributed at  commit 55b14d4.

I remembered you had said you would like to take a look at the input section of the code and make changes yourself whenever you had a larger section of time, so I didn't want to push changes that affect that to this branch without checking in. Is it alright if I pushed my MPI fix here? 

Thank you, 

Maria

akohlmey:
> Is it alright if I pushed my MPI fix here?

Certainly! At the moment, I don't have large enough chunks of time available to work consistently on either of those two contributed packages. At the moment it looks like it could be until Christmas break or even after that.

akohlmey:
@Maria-Lesniewski I just requested an AI code review for your pull request. Even though none of the LAMMPS developers seems to have much time available right now, this could help spotting some issues. But please be careful about the results. I have recently seen many incorrect suggestions like asking to use an obsolete internal API instead of the modern one.

Maria-Lesniewski:
Hello! Thank you for requesting the AI review and giving me a heads up. I will look closely at its suggestions and follow back here as soon as I can. 
Kind regards,
Maria

akohlmey:
@Maria-Lesniewski I have decided to let my other projects rest for a bit and work on pull requests like yours that have been neglected for quite some time. Another one was just merged yesterday and your contribution is next in line.

Before looking at the details, I have one high-level question: since this is related to BOCS, wouldn't it be better to add this to the BOCS package rather than starting a new package? or move the content of the BOCS package to LDD?

Maria-Lesniewski:
Hello! Thank you very much for taking the time to work on this package.

I talked to Will, and we think it makes sense to add PKG-LDD into PKG-BOCS if that would be easiest. Both of these LAMMPS packages implement potentials that can be parameterized with the [Bottom-up Open-source Coarse-graining Software (BOCS)](https://github.com/noid-group/BOCS) that we distribute, so it makes sense to put them together. 

We had not done this originally since they make such different modifications to LAMMPS. e.g. PKG-BOCS employs a fix to modify the barostat with a "pressure correction” that accounts for errors in the internal pressure of CG models. The ``pressure correction'' implemented in PKG-BOCS does not generate any interparticle forces and only works for homogeneous systems. In contrast, PKG-LDD implements local density and square gradient potentials that generate forces and improve thermodynamic properties and transferability within CG models for both homogeneous and inhomogeneous systems.

I am happy to assist in any way I can for making these and any other changes we might need to make things better along the way, though currently I am in the last ~2 weeks before I defend my dissertation. While I may be slower to respond within this period, please let me know if I can clarify anything or if there are other things you would like me to work on for this. 

Thank you again and kind regards, 
Maria

akohlmey:
@Maria-Lesniewski Thanks for getting back to me so quickly.

> I talked to Will, and we think it makes sense to add PKG-LDD into PKG-BOCS if that would be easiest. Both of these LAMMPS packages implement potentials that can be parameterized with the [Bottom-up Open-source Coarse-graining Software (BOCS)](https://github.com/noid-group/BOCS) that we distribute, so it makes sense to put them together.

Ok. Yes, that would be my preferred option. We are close to having 100 packages in LAMMPS which can make it sometimes difficult for users to identify the package they need to install. Thus we are trying to combine things where possible, e.g. single style contributions can be added to one of the EXTRA-* packages and similar.

> though currently I am in the last ~2 weeks before I defend my dissertation

Ok. I will see how much of the changes that are just "technical" I can make myself, and I will only bother you with questions to the "science". I don't know yet how fast this will be, since there is already a lot of information in this pull request discussion and I will have to re-familiarize myself with it and confirm the corresponding updates and so on.

Would it make a difference for you, if the pull request is merged before you are defending?

Maria-Lesniewski:
Okay that all makes sense. There is no difference to me if the pull request is merged before or after I defend, so feel free to take your time with it and to offload any code you think I should be fixing (probably post-defense) as you look into the package details some more.

Maria-Lesniewski:
Hello! 

I hope you are doing well today and all is well. 
Thank you again for all your help with this, I am now post-defense and should be able to help too.
If there is something that I can take care of, feel free to let me know and I will work at it. 

Thank you again, 

Maria

akohlmey:
@Maria-Lesniewski 

> I hope you are doing well today and all is well. Thank you again for all your help with this, I am now post-defense and should be able to help too.

I assume the defense went well for you.

> If there is something that I can take care of, feel free to let me know and I will work at it.

You could update to the latest upstream code and make the force-style test inputs work again. I won't be able to have a serious look until some time next week or the week after that. There are currently too many other projects requiring my attention. Sorry.

akohlmey:
@Maria-Lesniewski I had a little time to do some cleanup work on your branch.
Most was just cosmetic like removing unneeded includes (there are likely more that can be removed) and applying clang-format.
There were some inconsistencies with "virtual" and "override".
In the LddPotential and LddIndicator classes I made the functions that *must* be overridden virtual.
I removed some unused class members and modernized the calls to the Error class. Also I changed the fopen() for the table file to `utils::open_potential()` so they are found automatically in the potentials folder.
I replaced calls to `pow()` with integer exponent to use `MathSpecial::powint()` which is *much* faster **and** more accurate. Some modern C++ compilers do a similar thing if you use std::pow(), but using our own optimized function makes it platform neutral. This changed some numerics, so I had to reset the unit test YAML files.
I had to change the use of MPI_CXX_BOOL to use and communicate an MPI_INT since the MSMPI library is based on an old MPICH version (which has only MPI-2) and thus fails to compile without the change. This does not show in the unit test, because that compiles LAMMPS without MPI and I had already added support for MPI_CXX_BOOL to STUBS.

Some observations:
- reading of potential and table files should be only done on one processor. I have not checked properly, but it looks in some cases they are opened on all processors.
- related to that is the use of `Error::all()` versus `Error::one()`. I am not certain what is currently done is correct.
- When I have more time, we need to talk about the type mapping and parsing. It looks like we still need to make some changes there. For example, I would prefer we use the PotentialFileReader class and the Tokenizer/ValueTokenizer classes.
- The file names in the potentials folder should be changed to remove the `pair_ldd_input_` prefix and rename the `.txt` suffix to `.ldd` and document what those files are for in `potentials/README`
- The potential files should have a comment line like this in the first line:
`# DATE: 2007-06-11 UNITS: metal CONTRIBUTOR: Aidan Thompson, athomps@sandia.gov CITATION: Stillinger and Weber,  Phys Rev B, 31, 5262, (1985)
`
- It doesn't make much sense to me to keep the `pair_coeff` text in them.
- I noticed some MPI_Barrier() calls that are probably not necessary. If parsing is done on the MPI rank 0 only, you then in a second step use MPI_Bcast() to send the data around. That has an implicit barrier.

This was a quick session with only some rather obvious changes, so there is more work left. 

Please make some checks that I have not accidentally broken something. The change from pow() to powint() will result in some numeric changes (pow() does an implicit log() followed by an exp() and that has numerical accuracy compared to plain multiplication).

P.S.: there are still some failures on Windows, but their number is significantly reduced. I'm a bit confused about this because on my Windows virtual machine with Visual Studio all tests pass. Thus I am not sure where the failures on GitHub come from.

Maria-Lesniewski:
Hello! I hope you are doing well today. Thank you for these notes.

>@Maria-Lesniewski I had a little time to do some cleanup work on your branch.
Most was just cosmetic like removing unneeded includes (there are likely more that can be removed) and applying clang-format.
There were some inconsistencies with "virtual" and "override".
In the LddPotential and LddIndicator classes I made the functions that must be overridden virtual.
I removed some unused class members and modernized the calls to the Error class. Also I changed the fopen() for the table file to utils::open_potential() so they are found automatically in the potentials folder.
I replaced calls to pow() with integer exponent to use MathSpecial::powint() which is much faster and more accurate. Some modern C++ compilers do a similar thing if you use std::pow(), but using our own optimized function makes it platform neutral. This changed some numerics, so I had to reset the unit test YAML files.
I had to change the use of MPI_CXX_BOOL to use and communicate an MPI_INT since the MSMPI library is based on an old MPICH version (which has only MPI-2) and thus fails to compile without the change. This does not show in the unit test, because that compiles LAMMPS without MPI and I had already added support for MPI_CXX_BOOL to STUBS.

Thank you for these changes, I reviewed them and I think I learned a lot.

 >some observations:
>reading of potential and table files should be only done on one processor. I have not checked properly, but it looks in some cases they are opened on all processors.
 > related to that is the use of Error::all() versus Error::one(). I am not certain what is currently done is correct.
 >I noticed some MPI_Barrier() calls that are probably not necessary. If parsing is done on the MPI rank 0 only, you then in a second step use MPI_Bcast() to send the data around. That has an implicit barrier.
 
 Thank you for catching these MPI related errors and telling me. 
 I think I see the suspicious error->all within the table reader function you were alluding to. I will look for similar such inconsistencies / sanity check where I think one processor is working vs. multiple sometime this week.   
 
 >When I have more time, we need to talk about the type mapping and parsing. It looks like we still need to make some changes there. For example, I would prefer we use the PotentialFileReader class and the Tokenizer/ValueTokenizer classes.
>It doesn't make much sense to me to keep the pair_coeff text in them.
 
 All this related to the input is okay by me. If you would like ahead of time I could try to set up the current input to work with the PotentialFileReader/Tokenizer classes. I see the example in PairTersoff and I can look at the other styles in the MANYBODY package, then I can try to mimic what is done over there with these classes more closely. When I do this I can also rework the input to remove the "pair_coeff" text from the input files / get that section of code set up better for after there is time to think about the type mapping/parsing. (I also understand if you would prefer I work this up in one go after there is time to talk about it) 
 
 >Please make some checks that I have not accidentally broken something. The change from pow() to powint() will result in some numeric changes (pow() does an implicit log() followed by an exp() and that has numerical accuracy compared to plain multiplication). 

I am planning to re-run some of my larger tests to just sanity check that my force recoveries still work with the trajectory data generated. So far the log files from the examples and the differences in the .yaml files look numerical so I am optimistic but will verify anyway. 

>The potential files should have a comment line like this in the first line:
> \# DATE: 2007-06-11 UNITS: metal CONTRIBUTOR: Aidan Thompson, athomps@sandia.gov CITATION: Stillinger and Weber,  Phys Rev B, 31, 5262, (1985) 
 >The file names in the potentials folder should be changed to remove the pair_ldd_input_ prefix and rename the .txt suffix to .ldd and document what those files are for in potentials/README

I will take care of this today 


>P.S.: there are still some failures on Windows, but their number is significantly reduced. I'm a bit confused about this because on my Windows virtual machine with Visual Studio all tests pass. Thus I am not sure where the failures on GitHub come from.

Thank you for cleaning things up and getting us this far. 
I currently do not have Windows based programming work flow so I am going to see if I can get one and help out with these failures. The tests currently failing all pass through the mdpd style ldd potential. This potential should be pretty close to the quadratic style that Windows is currently happy with, the only difference I can think of is that it has a dependence on the indicator function cutoff r_c (causing more work to parse args into coeffs in this ldd potential style's setup_potl function). I am thinking I will start looking around where that processing happens for errors once I have the ability to check a Windows compiled version of code. 

Thank you again for your time and help, 

Maria

akohlmey:
@Maria-Lesniewski I am terribly sorry, but we (i.e. the LAMMPS developers) are still overloaded (as you can see from the long list of pending pull requests). I myself am travelling during the next couple of weeks and was busy with other projects in the last weeks. So your pull request will have to stay in "purgatory" a bit longer. At the moment, I have to prioritize smaller changes and bugfixes.

akohlmey:
@mdelyser @Maria-Lesniewski @wnoid
This now works to my satisfaction. There were a few minor remaining issues:
- input files had tabs: "tabs are evil (TM)"
- I added "loop geom" to velocity create so the initial velocities are the same regardless of the number of processors and the log files match
- I changed the input for the molecules to create a data file first, again so that the logs from 1 and 4 processors match
- I reduced the neighbor list skin which also made the neigh_modify one option obsolete
- to avoid segfaults or data corruption from insufficient communication buffers the pair style now enforces that it is defined *after* the system box is created.

akohlmey (approved):

sjplimp (approved):

PR #4740, opened 2025-10-18
Add a doc page note on how to run Monte-Carlo-only simulations efficiently

sjplimp:
**Summary**

For the fix atom/swap command doc page, clarify how to most efficiently run MC-only simulations (no MD).
Added similar note to fix neighbor/swap and fix gcmc.

**Related Issue(s)**

N/A

**Author(s)**

Steve, after discussion with Aidan

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

**Post Submission Checklist**


sjplimp:
@athomps Please see what you think of this added note.  If you agree, should something similar be added to other MC fixes, e.g. fix gcmc, widom, sgcmc, neighbor/swap, etc ?

akohlmey:
@sjplimp there is no question your suggestion improves the efficiency, but how can we monitor the progress of such a simulation, e.g. various energies, a "trajectory", or number of MC successes vs. attempts?
Should we perhaps have a different run style ("mc")? That would omit calling general force computation (and time integration?).

sjplimp:
@akohlmey good point on monitoring progress of a run - I think it should be sufficient to do something like this (for 5000 swaps):
fix atom/swap 1 500 ...; run 10    or
fix atom/swap 1 50 ...; run 100
which would have minimal inefficiency.  We should add that to the doc NOTE.
So a new run_style mc is an interesting idea, but not needed for this case, IMO.
Let's see what @athomps has to say ...

sjplimp:
@akohlmey Added a sentence about monitoring MC progress to the doc page note.

akohlmey:
> @akohlmey Added a sentence about monitoring MC progress to the doc page note.

Thanks. How about making this a separate file and the description sufficiently generic, so that we can add it to all relevant doc files with an ".. include:: mc_only_note.rst"?
Same as we do for accelerated styles.

athomps (changes requested):
This is a useful thing to capture for fix atom/swap, but it may or may not be true for other MC fixes in LAMMPS.  I tested the idea out on examples/mc/in.gcmc.lj and found that it worked as expected, but there was no time saving, because *X* is 100. Making *X* even bigger does not make any difference to performance.  For any MC method, the worst thing to do is set *X*=1, because that maximizes the cost of overhead.  I would shorten the note and keep it general: set `thermo 1`, set `*N* = 1, make *X* small enough to see evolution of system, but if it is too small, the overhead at the start and end of the MC sequence will start to slow down the simulation rate.

sjplimp:
@athomps Followed your advice for 3 MC fixes: fix atom/swap, fix gcmc, and fix neighbor/swap.

@akohlmey I think this is ready to merge after Aidan approves.

akohlmey:
> @akohlmey I think this is ready to merge after Aidan approves.

@sjplimp I needed to resolve a merge conflict and fix some whitespace issues.

PR #4742, opened 2025-10-20
Collected small changes and fixes

akohlmey:
**Summary**

This pull request combines multiple small changes and bug fixes that do not warrant pull requests on their own.

**Related Issue(s)**

Fixes #4741 
Includes and thus closes #4743 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Artificial Intelligence (AI) Tools Usage**

By submitting this pull request, I confirm that I did NOT use any AI tools to generate
all or parts of the code and modifications in this pull request.

**Backward Compatibility**

N/A

**Implementation Notes**

The following individual changes are included:
- bugfix for read_dump to not allow creating atoms with out-of-range atom type
- correct some minor issues in the recently merged fix rigid and rigid/small refactor
- small refactoring changes suggested by clang-tidy
- add missing includes in a couple tester programs
- document with more detail build prerequisites and add note about problems with some IntelLLVM compilers
- update kspace grid when using fix pimd/langevin with variable cell imported from PR #4743 

**Post Submission Checklist**

- [ ] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [x] The source code follows the LAMMPS formatting guidelines
- [x] Suitable new documentation files and/or updates to the existing docs are included
- [x] The added/updated documentation is integrated and tested with the documentation build system
- [x] The feature has been verified to work with the conventional build system
- [x] The feature has been verified to work with the CMake based build system
- [ ] Suitable tests have been added to the unittest tree.

PR #4743, opened 2025-10-20
fix pimd/langevin: support kspace for npt

Yi-FanLi:
**Summary**

The added code executes kspace->setup everytime when the volume is changed, similar to what is done at https://github.com/lammps/lammps/blob/dbc06d45e1e25800e7354b5487ff07aed4c920da/src/fix_nh.cpp#L872C5-L872C45. It ensures the correct kspace mesh when the volume changes and is necessary when kspace is involved in the force evaluation. 

**Related Issue(s)**

**Author(s)**

Yifan Li
Princeton University
yifanl0716@gmail.com

**Licensing**

**Backward Compatibility**

No breaking change.

**Implementation Notes**

**Post Submission Checklist**

### DIFF
--- a/src/REPLICA/fix_pimd_langevin.cpp
+++ b/src/REPLICA/fix_pimd_langevin.cpp
@@ -878,6 +878,7 @@ void FixPIMDLangevin::qc_step()
     MPI_Bcast(&domain->boxhi[0], 3, MPI_DOUBLE, 0, universe->uworld);
     domain->set_global_box();
     domain->set_local_box();
+    if (kspace_flag) force->kspace->setup();
   }
 }
 
--- a/src/REPLICA/fix_pimd_langevin.cpp
+++ b/src/REPLICA/fix_pimd_langevin.cpp
@@ -476,9 +476,6 @@ void FixPIMDLangevin::init()
     error->universe_all(FLERR, fmt::format("Unknown integrator parameter for fix {}", style));
   }
 
-  if (force->kspace) kspace_flag = 1;
-  else kspace_flag = 0;
-
   comm_init();
 
   mass = new double[atom->ntypes + 1];
@@ -882,7 +879,7 @@ void FixPIMDLangevin::qc_step()
     MPI_Bcast(&domain->boxhi[0], 3, MPI_DOUBLE, 0, universe->uworld);
     domain->set_global_box();
     domain->set_local_box();
-    if (kspace_flag) force->kspace->setup();
+    if (force->kspace) force->kspace->setup();
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/REPLICA/fix_pimd_langevin.cpp
```diff
@@ -475,6 +476,9 @@ void FixPIMDLangevin::init()
     error->universe_all(FLERR, fmt::format("Unknown integrator parameter for fix {}", style));
   }
 
+  if (force->kspace) kspace_flag = 1;
```
akohlmey: There doesn't seem to be a significant benefit to adding a kspace_flag member to the class over just testing for `force->kspace` being null or non-null. The flag is only used once and not inside a loop.

Yi-FanLi: Yes, I totally agree with you. I did this because the similar thing is done in https://github.com/lammps/lammps/blob/bb9dbbfdc454a3e6d092cbad0a4500e6f16e7442/src/fix_nh.cpp#L872, and I think it's better to keep consistent with the core code of LAMMPS.

akohlmey: > I did this because the similar thing is done in

There is a significant difference: there is more than one use of kspace_flag within that source file and the flag is used in derived classes. This is not the case here.

Yi-FanLi: Yes, it is true that `kspace_flag` is used twice in `fix_nh.cpp`, one in `initial_integrate` and the other in `pre_force_respa`. But it is only used once in one MD step -- either using the regular integrator or RESPA. This `kspace_flag` is only used when the volume is changed. In this sense the case is totally the same in classical MD and PIMD. If we are going to support RESPA for PIMD in the future, we are going to do exactly the same thing as classical MD does. Therefore it seems to me that keeping consistent with `fix_nh.cpp` would benefit readability of the code.

akohlmey: There are other issues to consider:
- we want classes to be lean and thus every added flag has to have a good reason. There is a lot of legacy code in LAMMPS that does not follow this rule, but things are changing. For example, in many cases class members "me" and "nprocs" have been removed and substituted with `comm->me` and `comm->nprocs`.
- please also note that any line of code *not* written is automatically correct.
- there is not much similarity between pimd and nh so consistency is not really expected

The primary exception in favor of keeping such class members is when those are used in performance critical places, e.g. in loops over all atoms or nested loops. This does not apply here.

Yi-FanLi: Thank you for your clarification. I have removed the `kspace_flag` from the code.

akohlmey:
@Yi-FanLi thanks for making the change. To reduce the review burden on the other LAMMPS developers, I've merged this pull request in my current "collection PR" #4742

PR #4744, opened 2025-10-21
Fully parallelize pack/unpack in ML-IAP Kokkos.

wcwitt:
**Summary**

The packing and unpacking routines in ML-IAP Kokkos are not fully parallelized, which can become a bottleneck for message-passing potentials when the number of atoms is small. The figure shows speedups achieved for the [NequIP](https://nequip.readthedocs.io/en/latest/integrations/lammps/mliap.html) and [MACE](https://mace-docs.readthedocs.io/en/latest/guide/lammps_mliap.html) specializations of the pair style. The effect is more pronounced for NequIP, probably because it requires more and larger messages.

pr

**Author(s)**

Chuck Witt, Harvard University

**Licensing**

**Backward Compatibility**

As far as I know, this change is backwards compatible.

**Implementation Notes**

I added hierarchical parallelism to the relevant routines.

**Post Submission Checklist**

### DIFF
--- a/src/KOKKOS/pair_mliap_kokkos.cpp
+++ b/src/KOKKOS/pair_mliap_kokkos.cpp
@@ -425,18 +425,11 @@ int PairMLIAPKokkos<DeviceType>::pack_forward_comm_kokkos(
   auto val=fill.view<DeviceType>();
   int nf=vec_len;
   auto to=copy_to;
-  Kokkos::parallel_for(
-    Kokkos::TeamPolicy<>(nv, Kokkos::AUTO, Kokkos::AUTO),
-    KOKKOS_LAMBDA (Kokkos::TeamPolicy<>::member_type team_member) {
-      const int i = team_member.league_rank();
+  Kokkos::parallel_for(nv*nf, KOKKOS_LAMBDA (int start) {
+      const int i = start/nf;
       const int gstart=idx(i)*nf;
-      const int start=i*nf;
-      Kokkos::parallel_for(
-        Kokkos::TeamVectorRange(team_member, nf),
-        [=] (const int j) {
-          val(start+j) = static_cast<double>(to[gstart+j]);
-        }
-      );
+      const int j = start%nf;
+      val(start+j) = static_cast<double>(to[gstart+j]);
     }
   );
   return nv*nf;
@@ -503,18 +496,11 @@ void PairMLIAPKokkos<DeviceType>::unpack_forward_comm_kokkos(
   auto val=fill.view<DeviceType>();
   int nf=vec_len;
   auto to=copy_to;
-  Kokkos::parallel_for(
-    Kokkos::TeamPolicy<>(nv, Kokkos::AUTO, Kokkos::AUTO),
-    KOKKOS_LAMBDA (Kokkos::TeamPolicy<>::member_type team_member) {
-      const int i = team_member.league_rank();
+  Kokkos::parallel_for(nv*nf, KOKKOS_LAMBDA (int start) {
+      const int i=start/nf;
       const int gstart=(first_up+i)*nf;
-      const int start=i*nf;
-      Kokkos::parallel_for(
-        Kokkos::TeamVectorRange(team_member, nf),
-        [=] (const int j) {
-          to[gstart+j] = static_cast<CommType>(val(start+j));
-        }
-      );
+      const int j=start%nf;
+      to[gstart+j] = static_cast<CommType>(val(start+j));
     }
   );
 }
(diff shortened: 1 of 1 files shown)


bathmatt:
I think this is a very positive change.  Thanks!!  I can't review it officially, but unofficially I think it moves in the right direction

wcwitt:
One thing I'd appreciate from whoever reviews is to double check whether there are any missing Kokkos::fences, etc. As far as I can tell everything is correct, but I did see one additional crash from this version in my testing. So far I'm attributing it to a quirk of the foundation model I am using, which is not fine-tuned for this task, but I am not 100% sure.

bathmatt:
One thing to try maybe, going to a MD Range policy so you don't have a full warp when the inner loop is small.

as for the fence, I believe there is a fence prior to comm, if not a fence at the end of the packing is needed.

athomps (approved):
I approve these changes. However, if the new code is crashing and the old code is not, that does suggest the need for additional fences or other checks. Perhaps @stanmoore1 can comment.

stanmoore1:
`CommKokkos` has fences after every pack and unpack kernel, see:
https://github.com/lammps/lammps/blob/develop/src/KOKKOS/comm_kokkos.cpp#L195
https://github.com/lammps/lammps/blob/develop/src/KOKKOS/comm_kokkos.cpp#L220

The issue is that GPU-aware MPI calls can be asynchronous and you don't want them to overlap with Kokkos kernels, i.e. don't want to start MPI sends before you packed all the data, or start unpacking the data before MPI is done receiving data.

stanmoore1:
Honestly I'm not 100% sure both fences are needed.

wcwitt:
> One thing to try maybe, going to a MD Range policy so you don't have a full warp when the inner loop is small.

I tried this first, actually, with similar effect (from what I remember). Then I switched to the hierarchical version because, while the inner loop always accesses contiguous data, the outer loop does not. But either approach is okay with me.

> However, if the new code is crashing and the old code is not, that does suggest the need for additional fences or other checks.

Thanks everyone for looking at the fences. I still don't _think_ there is a problem, but trying to be vigilant because if there is a subtle race or similar I worry it would be hard to find later.

stanmoore1:
@wcwitt have you measured performance for larger systems? For Lennard Jones we find that a `TeamPolicy` threading over neighbors is actually slower at around 16k atoms. If you have enough work then a `RangePolicy` has less overhead.

wcwitt:
This is basically all I've done, aside from some early experimenting. We were mainly trying to understand why ML-IAP was so much slower than `pair_style nequip` (on single GPU), and profiling revealed a lot of time spent packing and unpacking.

It is probably worth re-running these tests with RangePolicy - if nothing else, that code is easier to read.

fglines-nv:
This is  a good change, I'm a little disappointed in ourselves that we didn't catch this when we first wrote it.

Chiming in about the `RangePolicy` vs. `MDRangePolicy` vs. `TeamPolicy` as currently written -- with the `TeamPolicy`, you'll end up with half empty warps on GPUs (groups of 32 threads on NVIDIA GPUs) since each team internal will run with a multiple of the warp size. For example with a vector-length/# of features of 80, only 80/(3x32) = 80/96 = 83% of threads will do real work. Tiling with `MDRangePolicy` I think will lead to the same issue.

An easy solution would be to collapse the 2 dimensions into a 1D range policy with the teams effectively decoupled from the 2 dimensions. Only a single warp in the whole kernel will potentially end up half-empty. (I can help with that if need be tomorrow)

bathmatt:
@wcwitt I'm going to push a fork branch which does the unrolling, based of what you've done, it would be good to test. think it will help some in the 100s of atoms of ranges with smaller num_var counts

bathmatt:
Here's a format-patch to try, should be close to optimal. It's what @fglines-nv suggested

[0001-unrolled-pack-unpack-loop.patch](https://github.com/user-attachments/files/23054315/0001-unrolled-pack-unpack-loop.patch)

bathmatt:
i couldn't figure out how to make a pull request to your branch, so that's why the patch.  @wcwitt

wcwitt:
Thanks, both! As for catching this, it helped that I wrote a [similar pair style (for mace)](https://github.com/wcwitt/symmetrix/blob/720a91ce0d5752d36e9e9fda4dfb21ab749bc5fc/pair_symmetrix/pair_symmetrix_mace_kokkos.cpp#L239-L257).

Today was busy, but I will try to test the various options by the end of the week.

I do have a question about the full flattening. The outer loops have some random access - won't that mess with the warps in a different way? Most MLIPs have messages that are multiples of 32 by design, so I'm not too worried overall.

bathmatt:
So, in 
```
  Kokkos::parallel_for(nv*nf, KOKKOS_LAMBDA (int start) {
      const int i = start/nf;
      const int gstart=idx(i)*nf;
      const int j = start%nf;
      val(start+j) = static_cast<double>(to[gstart+j]);
    }
  );
  ```
you stride contiguous in j, however, you do have jumps in to[gstart] so you lose that stride 1 access at jumps of nf. 

This is all for fp32

In the case where nf isn't a multiple of 32, this will require most likely 4 lines of data (we don't really expect ever val[start] to be page aligned) and we will have to stall while we wait for this data, where on a full warp dedicated to a single gstart value, you'd only need to wait for 2.  IIUC, this is your concern.

This can result in more over fetching.  For example if your vector len is 31, then you'd split the second value of start across 2 warps and maybe 2 SMs.  so there are areas where this will hurt performance.  This is a trade-off of lowering effective occupancy by 31/32.  With a value like 31, this won't hurt, but with something like 17 you will probably see performance drop off.  

I think on average this will be better., but you have the test setup.

wcwitt:
Fair enough, thanks for explaining.

bathmatt:
To be fair, there are times where I'd make a big bet on better performance and find it go backwards, and vice versa.  so the proof is in the timings.

akohlmey:
@wcwitt and @stanmoore1 please let me know when I can process this by assigning me to the pull request and move it in the project to "ready for review". Thanks.

stanmoore1 (approved):
I approve

wcwitt:
Thanks @stanmoore1 - reminds me that I did re-run the benchmark, comparing the original version with the unrolled version. I also swapped the MACE model for something more stable (no crashes this time) and ran for larger sizes. I'm happy with either version.

pr

PR #4747, opened 2025-10-24
Fix bug in Kokkos neigh list with FP32 positions + newton on

stanmoore1:
**Summary**

Fix bug in Kokkos half neigh list with atom positions in FP32 (true for either `single` or `mixed` precision) and `newton on` by falling back to triclinic neighborlist build (even when the box is orthogonal). The issue is that FP32 leads to a large amount of roundoff error in the coordinates that can put particles into an adjacent neighborlist bin leading to an incorrect number of neighbors and missed interactions. This is similar to the triclinic case where:

```
// for triclinic, bin stencil is full in all 3 dims
        // must use itag/jtag to eliminate half the I/J interactions
        // cannot use I/J exact coord comparision
        //   b/c transforming orthog -> lambda -> orthog for ghost atoms
        //   with an added PBC offset can shift all 3 coords by epsilon
```

```
  // For half stencils, only the upper plane is needed
  // for triclinic, need to use full stencil in all dims
  //   not a half stencil in y
  // b/c transforming orthog -> lambda -> orthog for ghost atoms
  //   with an added PBC offset can shift both coords by epsilon
  // thus for an I/J owned/ghost pair, the xy coords
  //   and bin assignments can be different on I proc vs J proc
```

**Related Issue(s)**

**Author(s)**

Stan Moore (SNL), reported by @weinbe2 

**Licensing**

**Backward Compatibility**

Yes

### DIFF
--- a/src/KOKKOS/kokkos.cpp
+++ b/src/KOKKOS/kokkos.cpp
@@ -107,6 +107,10 @@ KokkosLMP::KokkosLMP(LAMMPS *lmp, int narg, char **arg) : Pointers(lmp)
   int device = 0;
   nthreads = 1;
 
+  kk_fp32 = 0;
+  if (sizeof(KK_FLOAT) != sizeof(double))
+    kk_fp32 = 1;
+
   threads_per_atom = 1;
   threads_per_atom_set = 0;
   pair_team_size = 128;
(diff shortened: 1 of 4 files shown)


jtclemm (approved):
Changes look good to me. 

Might be a reason to rethink whether TRI/ORTHO are the most accurate labels, but maybe this is just the one exception...

PR #4749, opened 2025-10-25
Fix exchange communication buffer size bug

### LINKED ISSUE #4706: [BUG] _Segmentation Fault and Memory Corruption with Large Number of Dihedrals_
**Summary**

I'm encountering a segmentation fault and memory corruption when running LAMMPS on a system with a large number of dihedrals. The simulation runs fine for smaller domains (upto 2x2x2 replication), but crashes when scaled up. The error includes double free or corruption (!prev) and corrupted double-linked list, followed by an abort signal.

LAMMPS version: 22Jul2025  (also tested on 29Sep2021 and 2Aug2023)
Build: foss-2024a with Kokkos (OpenMP backend)
Hardware: CPU-only nodes (also tested on GPU nodes)
MPI launcher: mpirun -np 64

**Expected Behavior**

LAMMPS should handle large systems gracefully, even if memory allocation fails. If the system exceeds memory limits or encounters internal issues, it should report a clear error message and exit cleanly — not crash with corrupted memory or segmentation faults.

**Actual Behavior**

When simulating a system with a large number of dihedrals (141,024 in the unit cell, scaled up to 50–200 unit cells), LAMMPS crashes after 10,000–20,000 timesteps with memory corruption errors. Removing dihedrals allows the simulation to run, but alters the physical behavior, so they are necessary for accurate modeling.

**Steps to Reproduce**

1. Use the attached input script and data file for a 5×2×2 supercell (approx. 2.8 million dihedrals).
2. Run the input.txt file
3. Observe crash after ~10,000–20,000 timesteps. 

**Further Information**

This issue was discussed with the HPC administrator, who confirmed that this behavior likely indicates a bug in LAMMPS. He mentioned that even with high memory usage, the code should not crash and should report the error appropriately.

**Attached input script and data files for reproducing the error**

[lammps_bug_files.tar.gz](https://github.com/user-attachments/files/22055323/lammps_bug_files.tar.gz)

akohlmey:
**Summary**

This pull request addresses a communication buffer overflow when a system has extremely many bonds, angles, dihedrals, or impropers.

**Related Issue(s)**

Fixes #4706 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

This re-purposes `AtomVec::maxexchange`. The comment in `atom_vec.h` suggests it needs to be
set to a nonzero value in case more per-atom data than the BUFEXTRA constant in `Comm` is
needed. But it was never set to anything but 0 and there are no checks, which leads to the
overflow reported in issue #4706 as the example input deck would require a BUFEXTRA value
of about 8000. In the existing code `AtomVec::maxexchange` is used *in addition* to BUFEXTRA.

The modifications in this pull request now re-purpose this member to represent the number of
items required by per-atom properties that are dynamically determined and not from fixes or
bonus data, that is bonds, angles, dihedrals, and impropers. This means that 
`AtomVec::maxexchange` is now set during `read_data` and `create_box`. Its value is stored
in binary restart files and restored during `read_restart`.

There is no more crash, no memory corruption (and malloc()/free() complaining about it) and the provided example run is completely valgrind clean.

**Post Submission Checklist**


src/read_data.cpp
```diff
@@ -710,6 +710,8 @@ void ReadData::command(int narg, char **arg)
     if (firstpass) {
       delete lmap;
       lmap = new LabelMap(lmp, ntypes, nbondtypes, nangletypes, ndihedraltypes, nimpropertypes);
+      // reset maxexchange
+      atom->avec->maxexchange = 0;
```
jtclemm: Does this work if a user uses multiple read data commands with the *add* option?

Could read_data.cpp calculate a maxexchange value for that instance and then replace it if needed?
`atom->avec->maxexchange = MAX(atom->avec->maxexchange, maxexchange_here);`

akohlmey: *Very* good question. I think this should only be set on the first file (`addflag == NONE`) or during `create_box`. I will also have to check what happens with the LabelMap instance. This looks like some updates to the pull request are needed, but I need to test and probably should check the available tests and add some updates there, too.

akohlmey: Prefixing this statement with `if (addflag == NONE)` does the trick.
> Could read_data.cpp calculate a maxexchange value for that instance and then replace it if needed?
No. The current limit may only be set when the box is created. That is why we have the `extra/XXX/per/atom`. Later this cannot be changed.

Or rather, cannot be changed *yet*. Steve was mentioning to try and lift this restriction.

jtclemm: Makes sense.

One other case, what happens if the atom vec style actually defines a larger value of `maxexchange`? For instance, `AtomVecBody` calculates the max exchange size from what I assume is the number of nodes. 

Should it then be `atom->avec->maxexchange += maxexchange` in the create_box and read_data classes? Or equivalently initialize those temporary counter `maxexchange` to  `atom->avec->maxexchange` instead of zero...

akohlmey: The latter. The line in read_data is not needed because maxexchange is already set to zero already in its constructor and will be added to only once for the first data file and during the first pass of reading the data file.

jtclemm (approved):
Looks good to me.

PR #4753, opened 2025-10-28
adding energy calculation

snikolov50:
**Summary**

Adding energy tracking to fix_langevin_spin - to replicate similar behavior by fix_langevin

**Related Issue(s)**

N/A

**Author(s)**

Svetoslav Nikolov
Julien Tranchida

**Licensing**

**Backward Compatibility**

Should be backward compatible

**Implementation Notes**

Correctness verified by carrying out magnon thermal conductivity calculations for iron. See https://doi.org/10.1007/s10853-021-06865-3

**Post Submission Checklist**

### DIFF
--- a/src/SPIN/fix_langevin_spin.cpp
+++ b/src/SPIN/fix_langevin_spin.cpp
@@ -193,12 +193,11 @@ void FixLangevinSpin::compute_single_langevin(int i, double spi[3], double fmi[3
 
 double FixLangevinSpin::compute_vector(int n)
 {
-  double energy_allS;
-  MPI_Allreduce(&energyS,&energy_allS,1,MPI_DOUBLE,MPI_SUM,world);
-  double energy_allD;
-  MPI_Allreduce(&energyD,&energy_allD,1,MPI_DOUBLE,MPI_SUM,world);
-  energy_vec[0] = energy_allS / output->thermo_every;
-  energy_vec[1] = energy_allD / output->thermo_every;
+  double energy_one[2] = {energyS, energyD}, energy_all[2];
+  MPI_Allreduce(energy_one, energy_all, 2, MPI_DOUBLE, MPI_SUM, world);
+  energy_vec[0] = energy_all[0] / output->thermo_every;
+  energy_vec[1] = energy_all[1] / output->thermo_every;
+
   if (n==0) energyS = 0;
   if (n==1) energyD = 0;
   if (n == 0) return energy_vec[0];
--- a/src/SPIN/fix_langevin_spin.cpp
+++ b/src/SPIN/fix_langevin_spin.cpp
@@ -33,6 +33,8 @@
 #include "random_mars.h"
 #include "respa.h"
 #include "update.h"
+#include "memory.h"
+#include "output.h"
 
 #include <cmath>
 #include <cstring>
@@ -51,6 +53,11 @@ FixLangevinSpin::FixLangevinSpin(LAMMPS *lmp, int narg, char **arg) :
   temp = utils::numeric(FLERR,arg[3],false,lmp);
   alpha_t = utils::numeric(FLERR,arg[4],false,lmp);
   seed = utils::inumeric(FLERR,arg[5],false,lmp);
+
+  nevery = 1;
+  vector_flag = 1;
+  size_vector = 2;
+  extvector = 0;
 
   if (alpha_t < 0.0) {
     error->all(FLERR,"Illegal langevin/spin command");
@@ -72,6 +79,7 @@ FixLangevinSpin::FixLangevinSpin(LAMMPS *lmp, int narg, char **arg) :
 
   // random = new RanPark(lmp,seed + comm->me);
   random = new RanMars(lmp,seed + comm->me);
+  allocate();
 }
 
 /* ---------------------------------------------------------------------- */
@@ -79,6 +87,7 @@ FixLangevinSpin::FixLangevinSpin(LAMMPS *lmp, int narg, char **arg) :
 FixLangevinSpin::~FixLangevinSpin()
 {
   delete random;
+  memory->destroy(energy_vec);
 }
 
 /* ---------------------------------------------------------------------- */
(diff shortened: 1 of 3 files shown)


src/SPIN/fix_langevin_spin.cpp
```diff
@@ -168,6 +185,29 @@ void FixLangevinSpin::compute_single_langevin(int i, double spi[3], double fmi[3
...
-    if (temp_flag) add_temperature(fmi);
+    if (temp_flag) add_temperature(spi,fmi);
   }
 }
+
+/* ---------------------------------------------------------------------- */
+
+double FixLangevinSpin::compute_vector(int n)
+{
+  double energy_allS;
```
akohlmey: If you use something like this (untested) for lines 196-201:
```C++
double energy_one[2] = {energyS, energyD}, energy_all[2];
MPI_Allreduce(energy_one, energy_all, 2, MPI_DOUBLE, MPI_SUM, world);
energy_vec[0] = energy_all[0] / output->thermo_every;
energy_vec[1] = energy_all[1] / output->thermo_every;
```
You can combine two costly MPI_Allreduce calls into one.

snikolov50: Thanks I'll take a look later today. Julien should be reviewing this btw - not sure what the rest of the people you added are going to contribute.

akohlmey: @snikolov50 To get *any* change merged into the LAMMPS source code, two of the *core LAMMPS developers* (https://www.lammps.org/authors.html) have to agree with the change.
As the person doing the merge, I would be one (and I see no technical problem) but I need an approval from a second person or else the merge will be blocked.
I can add Julien as a reviewer and his feedback can inform others, but since he is not on the core team, his approval will not be sufficient.

snikolov50: Ah okay, that makes sense. From a practical standpoint though Julien is the developer of the SPIN package and the rest of the people you listed aren't involved with that work. So definitely worth adding him on here. Thanks!

snikolov50: Added the improvements you suggested on lines 196-201 and retested the compiled version. The produced results were identical to the ones before.

julient31:
Hello @akohlmey   and @snikolov50 

I reviewed the code, it all looks good to me. Thank you very much for asking for my review.

For anyone’s information, the main purpose of the PR is to add the possibility of extracting the added and subtracted energies from a spin Langevin thermostat. Those quantities corresponds to the ‘energyD’ and ‘energyS’ variables, respectively. 

A derivation of the corresponding quantities can be found in the two following references :
- Ma, P. W., Dudarev, S. L., Semenov, A. A., & Woo, C. H. (2010). Temperature for a dynamic spin ensemble. Physical Review E—Statistical, Nonlinear, and Soft Matter Physics, 82(3), 031111.
- Nurdin, W. B., & Schotte, K. D. (2000). Dynamical temperature for spin systems. Physical Review E, 61(4), 3579.

Svetoslav and I tested this implementation in combination with a SNAP+Spin potential, with correcponding results published in this reference :
- Nikolov, S., Tranchida, J., Ramakrishna, K., Lokamani, M., Cangi, A., & Wood, M. A. (2022). Dissociating the phononic, magnetic and electronic contributions to thermal conductivity: a computational study in alpha-iron. Journal of Materials Science, 57(23), 10535-10548.

Do not hesitate if you have any additional question!
Best,
Julien.

src/SPIN/fix_langevin_spin.cpp
julient31: @akohlmey do not hesitate to always add me as a reviewer if some changes concern the SPIN package. I perfectly understand that my approval is not sufficient, but I will to my best to give some feedback to the other reviewers.

Actually, I would feel a little anxious to know that something has been modified in the SPIN package, without me (or Svetoslav) verifying it. The implemented physics is rather subtle and sometimes difficult to follow for non-experts in those type of models. 
Thanks Axel!

julient31 (approved):

src/SPIN/fix_langevin_spin.cpp
```diff
@@ -168,6 +185,28 @@ void FixLangevinSpin::compute_single_langevin(int i, double spi[3], double fmi[3
...
+{
+  double energy_one[2] = {energyS, energyD}, energy_all[2];
+  MPI_Allreduce(energy_one, energy_all, 2, MPI_DOUBLE, MPI_SUM, world);
+  energy_vec[0] = energy_all[0] / output->thermo_every;
+  energy_vec[1] = energy_all[1] / output->thermo_every;
+
+  if (n==0) energyS = 0;
+  if (n==1) energyD = 0;
+  if (n == 0) return energy_vec[0];
+  if (n == 1) return energy_vec[1];
```
jtclemm: Just for clarity, is there a reason to define a class variable for `energy_vec` and allocate/deallocate its memory instead of just using a local variable, like:
```
double energy_result;
if (n == 0) energy_result = energy_all[0] / output->thermo_ever;
if (n == 1) energy_result = energy_all[1] / output->thermo_ever;
return energy_result;
```
Maybe future functionality?

jtclemm:
One other comment, would be good to note the new global vector variables in the doc page.

akohlmey:
@julient31 You should have been automatically added to the list of reviewers because we added you to the CODEOWNERS file for all SPIN folder files, but it seems like GitHub has changed who is eligible for that and seems to be requiring accounts with write access to the repository (which is restricted to the core LAMMPS team).

akohlmey:
@snikolov50 @jtclemm I just committed some further simplification of the code.

I am a bit worried with resetting the data without having some check on the timestep. If the fix data is read multiple times or used in a multiple variables or similar, only the first access would give a result. For now my changes should preserve the results. They just simplify the code. There should not be a call to allocate for a 2 element array.

akohlmey (changes requested):
As per the LAMMPS manual:
We do not accept changes in functionality without the proper documentation. 

The added feature should be a new paragraph in the "Restart, fix_modify, output, run start/stop, minimize info" and ideally that paragraph should be prefixed with:
```
.. versionadded:: TBD
```
So that users reading the online documentation can see in which release this was added. The TBD will be replaced as part of the release process.

athomps (changes requested):
@snikolov50 Thanks for submitting this change.  I noticed a couple of outstanding issues:

1. The lack of documentation. At a minimum, the basic definitions above by @julient31 should be added to the doc page in the section "Restart, fix_modify, output, run start/stop, minimize info".  
2. Even better (but optional to you) would be to follow the examples of `fix langevin` and `fix langevin/eff` and add a `tally` keyword to turn on and off this functionality, again with documentation similar to that in `fix_langevin.rst`. It is hard to overstate the value, to future users and future developers, of copying keyword patterns from similar commands and code patterns from similar classes.  In fact, to fully replicate the functionality of `fix langevin` it would probably make sense to optionally compute a global scalar value in addition to the global vector i.e. `energy = D + sigma`

akohlmey:
@snikolov50 This is sitting idle now and waiting for changes for 3 weeks now. Any updates?
If you can't work on this soon, I suggest to close the pull request and resubmit when it is ready for a final review. Thanks.

akohlmey:
@snikolov50 @julient31 I am closing this now due to inactivity. Please re-open or re-submit when the requested changes are implemented and the code is updated to upstream.

Milenaw13:
Hi @snikolov50 , I am using the energy tracking feature for my thesis (I have had contact with you over email via @julient31). The goal is to compute the thermal conductivities for FeRh using the same method as in https://doi.org/10.1007/s10853-021-06865-3. I have some questions about the computed quantities 'energyD' and 'energyA'. For example, how do they compare to the output of fix langevin with the tally argument? Are they cumulative like in the fix langevin case? Are 'energyD' and 'energyA' the added and substracted energies from the thermostat to the spin system or the other way around? Any advice on how to correctly interpret these quantities would be very helpful. Thank you!

PR #4608, opened 2025-05-29
Add support for single precision (FP32) and mixed precision to KOKKOS…

stanmoore1:
… package

**Summary**

Add support for single precision (FP32) and mixed precision to KOKKOS package

**Related Issue(s)**

Closes https://github.com/lammps/lammps/issues/4056.

**Author(s)**

Stan Moore (SNL). Helpful macros for view types in `KOKKOS/kokkos_type.h` contributed by Mitch Murphy (@alphataubio).

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

Yes

PR #4754, opened 2025-10-28
Removing silent fp32/fp64 narrowing/widening conversions from many Kokkos package benchmarks + mixed precision SNAP

weinbe2:
**Summary**
This PR primarily removes silent, unnecessary fp32 <-> fp64 conversions for single and mixed precision in a handful of common LAMMPS/Kokkos benchmarks, covering LJ, EAM, Chute, Chain, Tersoff, Rhodopsin, and the most expensive kernels in ReaxFF. Example sources of silent conversions, and systematic ways to (syntactically) address them, are given in the “Implementation Notes” section below.

It exposes best-faith versions of single and mixed precision in SNAP, though it's untested for long timescale science workflows (aka, nve for tens of thousands+ of iterations).

This PR touches many different parts of many different benchmarks, and I did my best to clearly partition my changes into separate commits. This should hopefully make the review easier (or at least easier to break up into multiple sittings/cups of coffee), and (ideally) make it easy to make changes/updates without creating a cascade of breakage.

ReaxFF is not fully ported because some kernels run into numerical stability errors due to cancellations in the denominators of some bond order expressions, as first noticed by @stanmoore1 in #4608. That said, I patched up the most time-consuming kernels, which gives a nice end-to-end speed-up relative to double precision. I also improved the numerical stability of the iterative solve in fix/qeq.

For PPPM workflows, I made the decision to convert the computation of momentum-space convolution factors (phases for derivatives, Green’s functions) back to pure double before storing them in single. This is because they are (in theory) only computed once at initialization (…though I know that may not be true in np[t/e] simulations where the volume changes aggressively). I’m not married to this decision and will revert it back to a mixed-precision calculation if that’s preferred.
For fix/shake, I made the decision to set `1e30f` to be the overflow threshold in single precision, since the threshold of `1e150` for double doesn’t fit in `fp32`. I’m not married to this decision, it was “somewhat” arbitrary (similar to the reasoning given in code comments for `1e150`), and I’m open to other values.

Tangentially related: this PR promotes “torque” Views to the accumulation precision to match “force” views.

While I would have liked to go through and address silent conversions throughout the LAMMPS/Kokkos package… it would just take more time than I (or perhaps any one person) has. The main challenge is it doesn’t lend itself to being fully automated. Some calculations become numerically unstable in reduced precision, and there may or may not be methods to mitigate these instabilities. ReaxFF is an example of both (seemingly) unavoidable numerical stabilities _and_ instabilities that can be mitigated via targeted investigations. 

All this said: hopefully this PR serves as a useful template for other LAMMPS/Kokkos users who want to similarly optimize the components of their simulation that are most relevant to them. Folks are welcome to reach out to me on the LAMMPS Slack or by email (given below) if they have any questions while doing their own porting and I’ll be happy to support them to the best of my ability!

**Related Issue(s)**

Iterative improvement on #4608 

**Author(s)**

Evan Weinberg (NVIDIA). eweinberg [at] nvidia.com preferred, evansweinberg [at] gmail.com for "long-lived"; some decisions were informed by conversations with Stan Moore (thank you!)

**Licensing**

**Backward Compatibility**

No changes to backwards compatibility.

**Implementation Notes**

An example of the type of silent conversions this PR addresses are:
```
float val = [...];
float output = 1.4 + log(sqrt(1.0 + val));
```

Because the C++ standard permits silent float -> double conversions and, more importantly, silent double -> float narrowing conversions, the majority of this expression (all of the transcendentals) are evaluated in fp64 because of the initial 1.0 + val. The version of this that avoids this issue for an arbitrary floating-point type is:

```
using real_t = float; // or double
real_t output = static_cast<real_t>(1.4) + log(sqrt(static_cast<real_t>(1.0) + val));
```
Where the `static_cast<real_t>` are ideally evaluated at compile time.

These types of conversions were systematically hunted down via warning flags passed to `clang` as a CPU compiler, taking advantage of Kokkos' performance portability. The relevant warning flags to expose widening and narrowing conversions are `-Wall -Wextra -Wimplicit-float-conversion -Wdouble-promotion -pedantic` (possibly just a subset of these, I erred on the side of being overcomplete). When building via the `cmake` build system, these can be added to the `CMAKE_CXX_FLAGS`configure variable.

Silent fp32 <-> fp64 conversions were tracked down by first compiling LAMMPS using clang with the flags noted above (g++ also has analogous flags)---generating an absurd number of warnings along the way---and then deleting individual *.o files for a given *.cpp file I wanted to work through, taking advantage of how (most of the time) only modified cpp files/missing *o files will be recompiled.

The *.o files for `cmake` builds get buried in the `build` directory. As a representative example for how to find them, here’s the command I use to find the `pair_reaxff_kokkos.o` file from the build directory:
```
find . -name "*pair_reaxff_kokkos.*"
```
Where the asterisks on each side are intentional. After deleting a *.o file, I’ll re-run make to get a set of warnings for just that file.

More details on what the relevant warning flags find:

`-Wdouble-promotion` warnings uncover one of three scenarios.
1.	In some cases, it uncovered expressions like `1.0 + var` when `var` was already a KK_FLOAT, like the example given above, which is solved by replacing the `1.0` with `static_cast<KK_FLOAT>(1.0)`. 
2.	In other cases, it uncovered expressions where a scalar variable was a double that needed a `KK_FLOAT` version. This came up, for ex, when a double precision scalar variable from a non-Kokkos parent class gets used in a Kokkos kernel. An example of this, and my convention for addressing it, is given further below.
3.	Last, it could be an intentional and necessary cast from float to double, either due to accumulating into a `KK_ACC_FLOAT` for mixed precision compiles or for accumulating into a ` double` from a non-Kokkos parent class. In this case I added an explicit `static_cast<KK_ACC_FLOAT/double>` for the sake of being explicit (and to silence the warning).

`-Wimplicit-float-conversion` warnings generally uncover one of two scenarios.
1.	In some cases, it uncovers silent conversions from double back to float at assignment. These can normally be traced to something on the same line throwing a `-Wdouble-promotion` warning, so once you’ve addressed the silent promotions *to* double, you’ve addressed the silent narrowing conversions back to float.
2.	In other cases, this uncovers conversions from (unsigned) `int` or `long` to float, or (unsigned) `long` to `double`. These are relevant because floats and doubles cannot represent all ints and longs exactly, with doubles storing 32-bit integers exactly being the exception. Whether or not this is an issue depends on the context… but I don’t think I saw a case where it was an issue. For example, it came up in scaling factors for adding extra padding to buffers before allocating them, for ex factors of 1.1 or 1.2, where being off by a smidge wouldn’t lead to any under-allocations or silent correctness issues. These warnings are silenced via introducing explicit `static_cast` calls.

Example of scalar doubles in non-Kokkos parent classes: In EAM, there is a variable `double rhomax`. I created an analogous variable `KK_FLOAT rhomax_kk`---`_kk` being the convention I settled on---in `PairEAMKokkos` and set it to `static_cast<KK_FLOAT>(rhomax);` at the start of each call to `compute`. The decision to set it in `compute` on' each iteration is informed by the possibility, however unlikely, that `rhomax` was changed elsewhere in the MD simulation.

*Specific notes on fix/qeq*:

When compiling in single or mixed precision, the conjugate gradient solve within fix/qeq is unstable/meaningless for relative residual tolerances below the order of single precision machine epsilon, ~1e-7.

If there’s a need, it’s certainly possible to engineer support for both single and double precision solves (toggled via input file/command line arg) independent of if LAMMPS/Kokkos is compiled in single/mixed/double.

I made single precision solves more stable by promoting intermediate calculations within some of the BLAS-like calls to full double precision, for ex:
```
const auto alpha_0 = alpha[0]; 
d_st(i,0) += alpha_0 * d_d(i,0); 
d_r(i,0) += -alpha_0 * d_o(i,0); 
d_p(i,0) = d_r(i,0) * d_Hdia_inv_i; 
out.v[0] += d_r(i,0) * d_p(i,0); 
```
->
```
const double alpha_0 = alpha[0]; 
d_st(i,0) += static_cast<KK_FLOAT>(alpha_0 * static_cast<double>(d_d(i,0))); 
double r = static_cast<double>(d_r(i, 0)) - alpha_0 * static_cast<double>(d_o(i,0)); 
double p = r * static_cast<double>(d_Hdia_inv_i); 
out.v[0] += r * p; 
d_p(i,0) = static_cast<KK_FLOAT>(p); 
d_r(i,0) = static_cast<KK_FLOAT>(r);
```
Without these changes, solves to looser tolerances such as 1e-6 could still become unstable/non-convergent when combined with other reduced-precision fixes.

These changes did simultaneously optimize some of these blas-like calls because they avoid redundant stores then loads (which was also a source of silent narrowing and widening conversions).

*Specific notes on SNAP*:

Mixed precision was exposed in SNAP by adding a new template parameter `accum_t` along with the existing `real_t`. These are set to `KK_ACC_FLOAT` and `KK_FLOAT`, respectively. The addition of the `accum_t` template parameter is the source of a lot of lines of code changes.

`accum_t` was used for the accumulation of:

- B_j^j1,j2 triple products (sums of appropriate products of components of U * Z)
- dedr forces (sums of appropriate products of dU and Y)

As reasonable sources of cancellations of positive and negative scalars. This is an “educated guess” and not something explicitly numerically proven to cure all ills relative to pure single, as discussed further below.

*General comments on correctness/stability*

Correctness was verified by comparing the results of pure double precision runs from this PR with runs from develop, with `fix langevin` disabled as appropriate.

Mixed and single precision was verified to the best of my ability by checking the numerical stability of the benchmarks I touched for a few hundred timesteps. Mixed precision, and especially pure single precision, is “guilty until proven innocent”, and there’s no guarantee that all of these implementations will be numerically stable for (for ex) an nve ensemble for tens of thousands of MD iterations. With this in mind, all I can do is help incrementally enable these investigations…

*Performance measurements: RTX Blackwell Pro 6000 Server Edition*

The performance measurements below come from an RTX Blackwell Pro 6000 Server Edition using CUDA 13.0 update 1 set to maximum compute clocks. This serves to be a representative example for GPU with strong fp32 and relatively poor f64, emphasizing the benefit of single/mixed precision ASSUMING numerical stability, which as discussed above isn’t proven for each of these potentials. The atom counts and the number of timesteps per run are given in a data table below.

I’ll note this data predates the tricyclic fix in #4747 (but includes everything else in develop). I didn’t want to hold up opening the PR on re-collecting that data. I should be able to provide new data tomorrow. I still expect the trends to hold.
 
rtx-blackwell-pro-6000-server-edition

Further run data below; all numbers are reported in Matom-timestep/sec.

| benchmark | neigh | newton | atoms   | timesteps | double    | mixed     | single    |
|-----------|-------|--------|---------|-----------|-----------|-----------|-----------|
| LJ        | full  | off    | 8192000 | 500       | 312.104   | 548.119   | 556.283   |
| EAM       | half  | off    | 8192000 | 500       | 129.616   | 237.320   | 237.058   |
| Tersoff   | half  | on     | 8192000 | 500       | 155.986   | 551.189   | 827.789   |
| Chain     | half  | on     | 8192000 | 500       | 251.712   | 245.487   | 248.361   |
| Chute     | half  | on     | 8192000 | 500       | 764.091   | 754.636   | 761.252   |
| Rhodopsin | half  | on     | 1024000 | 500       | 13.965    | 22.133    | 23.688    |
| ReaxFF (HNS)  | half (qeq full) | on     | 933888    | 500       | 2.269     | 12.264    | 12.711   |
| SNAP 2J8 (chunksize 262144) | half  | on     | 1024000 | 500       | 0.969     | 10.948    | 12.303    

The FFTs in Rhodopsin were done with cuFFT; single and mixed used single precision FFTs, double used double precision FFTs as toggleable via the `-DFFT_SINGLE=[ON/OFF]` cmake flag.

It’s interesting that chain and chute have rough performance parity across precisions, which suggests they’re very latency bound---a point of future investigation. Based on the results for H100 below, mixed and single may become slower with proper tricyclic neighbor builds.

I note that these benchmarks have not necessarily been run at the atom counts that saturate GPU performance. If there’s interest in a particular system configuration, just ask and I can try to collect data for it.

*Performance measurements: NVIDIA H100-HBM3-SXM*

These performance numbers come from an NVIDIA H100-HBM3-SXM using CUDA 12.9 set to maximum compute clocks.
 
nvidia-h100-hbm3-sxm

Further run data below; all numbers are reported in Matom-timestep/sec.

| benchmark | neigh | newton | atoms   | timesteps | double    | mixed     | single    |
|-----------|-------|--------|---------|-----------|-----------|-----------|-----------|
| LJ        | full  | off    | 8192000 | 500       | 840.299   | 948.171   | 922.251   |
| EAM       | half  | off    | 8192000 | 500       | 290.754   | 302.296   | 303.252   |
| Tersoff   | half  | on     | 8192000 | 500       | 832.600   | 924.056   | 1041.000  |
| Chain     | half  | on     | 8192000 | 500       | 334.034   | 309.069   | 311.412   |
| Chute     | half  | on     | 8192000 | 500       | 1140.000  | 1102.000  | 1096.000  |
| Rhodopsin | half  | on     | 1024000 | 500       | 30.261    | 32.861    | 33.633    |
| ReaxFF (HNS)   | half (qeq full) | on     | 933888    | 500       | 10.749    | 15.524    | 15.491   |
| SNAP 2J8 (chunksize 262144) | half  | on     | 1024000 | 500       | 6.387     | 10.459    | 10.420    |

I’ll note that in the case of chain and chute where mixed and single are slightly slower than double. A cursory investigation suggests the difference is connected to using the tricyclic neighbor build, which wasn’t necessary for double.

Otherwise, the benefits of mixed or single for LJ, EAM, Tersoff, and Rhodopsin are not very pronounced, suggesting they’re relatively latency bound in cache, but further investigations would be interesting.

The FFTs in Rhodopsin were done with cuFFT; single and mixed used single precision FFTs, double used double precision FFTs as toggleable via the `-DFFT_SINGLE=[ON/OFF]` cmake flag.

Again, these benchmarks have not necessarily been run at the atom counts that saturate GPU performance. If there’s interest in a particular system configuration, just ask and I can try to collect data for it.

**Post Submission Checklist**

None.


stanmoore1:
@weinbe2 there are a couple of PPPM unit tests failing:

```
projectroot.unittest.force-styles.KSpaceStyle:pppm_tilted
[ RUN      ] PairStyle.kokkos_omp
/home/builder/workspace/dev/pull_requests/fedora/unit_tests/cmake_serial_kokkos_smallbig_shared/unittest/force-styles/test_pair_style.cpp:678: Failure
Failed
ERROR on proc 0: Out of range atoms - cannot compute PPPM
For more information see https://docs.lammps.org/err0004 (src/KOKKOS/pppm_kokkos.cpp:1198)
```

```
projectroot.unittest.force-styles.KSpaceStyle:pppm_tri
[ RUN      ] PairStyle.kokkos_omp
/home/builder/workspace/dev/pull_requests/fedora/unit_tests/cmake_serial_kokkos_smallbig_shared/unittest/force-styles/test_pair_style.cpp:678: Failure
Failed
ERROR on proc 0: Out of range atoms - cannot compute PPPM
For more information see https://docs.lammps.org/err0004 (src/KOKKOS/pppm_kokkos.cpp:1198)
```

weinbe2:
Fixed, the issue was a bug in updating the `boxlo_kk[...]` values for tricyclic/tilted systems here:

```
  // convert atoms from box to lamda coords
  if (triclinic == 0) {
    boxlo[0] = domain->boxlo[0];
    boxlo[1] = domain->boxlo[1];
    boxlo[2] = domain->boxlo[2];
  } else {
    boxlo[0] = domain->boxlo_lamda[0];
    boxlo[1] = domain->boxlo_lamda[1];
    boxlo[2] = domain->boxlo_lamda[2];

    domain->x2lamda(atomKK->nlocal);
  }

+  boxlo_kk[0] = static_cast<KK_FLOAT>(domain->boxlo[0]);
+  boxlo_kk[1] = static_cast<KK_FLOAT>(domain->boxlo[1]);
+  boxlo_kk[2] = static_cast<KK_FLOAT>(domain->boxlo[2]);
```

Where at the bottom I accidentally/erroneously set `boxlo_kk[...] = static_cast<KK_FLOAT>(domain->boxlo[...]);`, completely ignoring if the system is tricyclic/tiled or not. The fix was setting it to `boxlo_kk[...] = static_cast<KK_FLOAT>(boxlo_kk[...]);` so, by construction, it is guaranteed to respect the system type.

Along with this, I added extra updates to `*_kk` variables s.t. they've been reassigned to their non-`*_kk` values before they get used in any kernels.

These tests are now passing on my system.

stanmoore1:
I'm running through through internal Kokkos regression testing now.

stanmoore1 (approved):
Regression tests passed. I approve. Ready to merge from my POV.

PR #4761, opened 2025-11-01

### LINKED ISSUE #4512: [BUG] Zeroed pulling forces when restarting SMD simulation
**Summary**

It seems that the accumulated pulling forces gets zeroed when restarting an [smd](https://docs.lammps.org/fix_smd.html) simulation. And, perhaps more importantly, the restarting the trajectory does not produce an identical trajectory as to one that was never interrupted

Other people seems to have had similar trouble previously:
* https://matsci.org/t/how-to-restart-a-smd-simulation/14740/2
* https://matsci.org/t/force-resetting-when-restart-fix-smd/32108

It can be that I'm erroneously restarting this type of simulation, but I was able to restart a simple NVE simulation with the same parameters and input, just without the smd fix.

LAMMPS (4 Feb 2025 - Development - patch_4Feb2025-446-g6f0a59372c)
LAMMPS has been compiled with EXTRA-FIX, BASIC, and MANYBODY

```bash
lsb_release -a
Distributor ID:	Ubuntu
Description:	Ubuntu 24.10
Release:	24.10
Codename:	oracular
```

**Expected Behavior**
I would expect the accumulated pulling forces from a (correctly) restarted simulation to keep on accumulating. From reading the docs I assumed that the simulation would continue uninterrupted although it's not explicitly mentioned that the pulling forces would keep on accumulating. [From the documentation](https://docs.lammps.org/fix_smd.html):

> The fix stores the direction of the spring, current pulling target distance and the running PMF to [binary restart files](https://docs.lammps.org/restart.html). See the [read_restart](https://docs.lammps.org/read_restart.html) command for info on how to re-specify a fix in an input script that reads a restart file, so that the operation of the fix continues in an uninterrupted fashion.

Perhaps more importantly, I expected the positions of atoms from a simulation run for 100 steps to be identical to the positions of atoms from a simulation restarted from step 50 and run for another 50 steps (to reach the same total steps). **It is not**

**Actual Behavior**
When restarting a calculation with the fix smd, all other quantities (equilibrium distance of the spring, distance between the two reference points, accumulated PMF) gets reloaded _except_ the pulling forces (x, y, and z-component of the pulling force as well as the total force).

**Steps to Reproduce**
1. Run a simulation with `in.simple` that pulls the end of a simple 6 atom gold chain for 100 steps
```bash
./lmp -in in.txt -log first.log
```
2. Run a restarted simulation from step 50 for 50 step (for a total of 100 steps)
```bash
./lmp -in restart.txt -log restarted.log
```
3. Compare the pulling forces in column `v_fx` in `first.log` and `restarted.log`. The forces in `restarted.log` will start from 0 instead of whatever force is at step 50 in `first.log`.

I've attached input files, molecular structure, and parameter file here.

[restarted.log](https://github.com/user-attachments/files/19396871/restarted.log)
[restart.txt](https://github.com/user-attachments/files/19396872/restart.txt)
[in.txt](https://github.com/user-attachments/files/19396870/in.txt)
[first.log](https://github.com/user-attachments/files/19396873/first.log)
[au.lammps.txt](https://github.com/user-attachments/files/19396874/au.lammps.txt) (the file ending should be renamed to `au.lammps.eam` for the input files to work)

[gold_chain.txt](https://github.com/user-attachments/files/19627475/gold_chain.txt) (the file ending should be renamed to `gold_chain.dat` for the input files to work)

akohlmey:
**Summary**

This pull request combines multiple small changes and fixes that do not warrant a pull request of their own.

**Related Issue(s)**

Closes #4512
Includes and thus closes #4762

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- during custom dumps, re-check during `Dump::init_style()` if computes and fixes provide per-atom data. If they would have been replaced with different settings, using the dump could result in crashes.
- add warning to `fix smd` that it is unmaintained and that `fix colvars` or `fix plumed` should be used instead
- implement improvements and modernization for fix bond/react suggested by static code analysis tools
- don't provide bounding box information for variable shape or dynamic regions
- fix copy and modify bug in recent fix langevin refactor that rendered the "omega" keyword unusuable
- add (mostly hidden) dummy pages for removed pages and commands, so web searches for them don't end with a 404 page not found error.

**Post Submission Checklist**


akohlmey:
@jrgissing Coverity Scan is complaining a lot about the "unsafe" reading and parsing of files in fix bond/react.  We should try and convert this to use the TextFileReader and ValueTokenizer classes.  That would also avoid some unexpected behavior: using sscanf("%d") on a floating point number (say 2.5) will return the leading integer part (i.e. 2) instead of raising an error flagging that an unsupported value was entered.

doc/src/atc_add_molecule.rst
jtclemm, line 1: Not confident I understand this, does it just redirect to the fix_atc.cpp page to display the deprecation message?

doc/src/fix_gcmc.rst
```diff
@@ -68,82 +68,82 @@ Examples
...
-determined by the optional *mcmoves* keyword (see below).  The default
-behavior is as follows. If the *mol* keyword is used, only molecule
-translations and molecule rotations are performed with equal probability.
-Conversely, if the *mol* keyword is not used, only atom translations are
-performed.  M should typically be chosen to be approximately equal to the
-expected number of gas atoms or molecules of the given type within the
-simulation cell or region, which will result in roughly one MC move per
-atom or molecule per MC cycle.
+This fix performs grand canonical Monte Carlo (GCMC) exchanges of atoms
+or molecules with an imaginary ideal gas reservoir at the specified T
```
jtclemm: Here's another instance of T if your intention is to italicize (re: line 142).

doc/src/atc_add_molecule.rst
akohlmey: The fix_atc.rst is a dummy page that redirects to https://docs.lammps.org/Commands_removed.html
The intention is to have websites and search indices land on this (removed) page so people don't get a 404 page not found error but instead get the information that the command and package was removed.
The symlinks have the same purpose and save us copying the same page many times.

doc/src/fix_gcmc.rst
akohlmey: thanks. yes, the intention is to make the formatting match the math expressions closely. using an inline math directive `:math:`T`` seems overkill in this case.

doc/src/atc_add_molecule.rst
jtclemm: I'm guessing there's no easy way to create a list of redirects that all point to the one dummy page other than creating all of these separate files?

Seems unfortunate that these 50 or so files need to stay in /doc/src/ and clutter it up.

jtclemm (approved):

doc/src/atc_add_molecule.rst
akohlmey: > I'm guessing there's no easy way to create a list of redirects that all point to the one dummy page other than creating all of these separate files?

I have thought about it, but it is not as simple as it looks, because the webserver is service multiple versions of the documentation and for the stable branch those files are still present. In general, I prefer to have workarounds and so on in the doc sources, so that when we would have to move to a different web hosting service, there is less legacy configuration to transfer. With over 1000 rst files adding some dummy files is not too bad, especially since you can tell them apart in the source as symlinks.

sjplimp (approved):

jrgissing:
> @jrgissing Coverity Scan is complaining a lot about the "unsafe" reading and parsing of files in fix bond/react. We should try and convert this to use the TextFileReader and ValueTokenizer classes. That would also avoid some unexpected behavior: using sscanf("%d") on a floating point number (say 2.5) will return the leading integer part (i.e. 2) instead of raising an error flagging that an unsupported value was entered.

@akohlmey thanks, I will add these changes to the next bond/react update

akohlmey:
> > @jrgissing Coverity Scan is complaining a lot about the "unsafe" reading and parsing of files in fix bond/react. We should try and convert this to use the TextFileReader and ValueTokenizer classes. That would also avoid some unexpected behavior: using sscanf("%d") on a floating point number (say 2.5) will return the leading integer part (i.e. 2) instead of raising an error flagging that an unsupported value was entered.
> 
> @akohlmey thanks, I will add these changes to the next bond/react update

Feel free to reach out in case you need assistance (and the programmer's guide needs improvements :wink:).

PR #4769, opened 2025-11-08
Improve support for dynamic and variable regions

akohlmey:
**Summary**

This pull request improves support for dynamic (i.e. translating and rotating) regions and regions with variables defining the shape. In particular, it adds support for visualizing
those with dump image and for updating the bounding box and thus improving on the bugfix in commit 51c2d126a739001f68df5b182e2b0ebe66164c56

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

This makes extensive use of `Region::forward_transform()` to apply the effect of rotation and translation.

Fix styles `deposit`, `gcmc`, and `widom` are currently not compatible with dynamic regions, but could be made compatible by repeating the bounding box initialization from the constructor in whatever function is called before inserting particles.

**Post Submission Checklist**

### DIFF
--- a/src/region_block.cpp
+++ b/src/region_block.cpp
@@ -19,8 +19,8 @@
 #include "math_extra.h"
 #include "variable.h"
 
-#include <cstring>
 #include <algorithm>
+#include <cstring>
 
 using namespace LAMMPS_NS;
 
@@ -169,7 +169,7 @@ RegBlock::RegBlock(LAMMPS *lmp, int narg, char **arg) :
       extent_zlo = zlo;
       extent_zhi = zhi;
     }
-  } else bboxflag = 0;
+  }
 
   // particle could be close to all 6 planes
   // particle can only touch 3 planes
--- a/src/region_block.cpp
+++ b/src/region_block.cpp
@@ -20,10 +20,11 @@
 #include "variable.h"
 
 #include <cstring>
+#include <algorithm>
 
 using namespace LAMMPS_NS;
 
-static constexpr double BIG = 1.0e20;
+static constexpr double BIG = 1.0e200;
 
 /* ---------------------------------------------------------------------- */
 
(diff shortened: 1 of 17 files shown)


src/region_block.cpp
```diff
@@ -508,6 +512,53 @@ void RegBlock::shape_update()    // addition
...
+      double zmin = BIG;
+      double zmax = -BIG;
+
+      // the corners of face[0] and face[1] cover the full extent of the region
+      // transform and get min/max in x-, y-, and z-direction for each corner
+
+      for (int i = 0; i < 4; ++i) {
+        MathExtra::copy3(corners[0][i], pos);
+        forward_transform(pos[0], pos[1], pos[2]);
+        xmin = std::min(xmin, pos[0]);
```
jtclemm, line 535: I'm curious, is there any relative advantage/disadvantage to `std::min` vs the existing `MIN` macro?

src/region_block.cpp
```diff
@@ -154,18 +155,21 @@ RegBlock::RegBlock(LAMMPS *lmp, int narg, char **arg) :
...
-    extent_xhi = xhi;
-    extent_ylo = ylo;
-    extent_yhi = yhi;
-    extent_zlo = zlo;
-    extent_zhi = zhi;
-  } else
-    bboxflag = 0;
+  // set extent of block
+
+  if (interior) {
```
jtclemm, line 160: Is it true that `bboxflag` is always 1 if the surface is interior? Why set it equal to 1 separately in each branch of the if statement?

src/region_block.cpp
akohlmey: Not sure if this is an advantage or disadvantage, but since they are templated inline functions `std::min()` and `std::max()` give a compile error, if the two arguments don't have the same data type. That is the main difference that I am aware of. I was using it mostly as part of the overall attempt to modernize the code from C to C++.

This happened later, but using `std::max()` helped finding the issue fixed in  a4310ccb7254380f0e7329a06b3357f58274ce9a

src/region_block.cpp
akohlmey: > Is it true that bboxflag is always 1 if the surface is interior?

For the block region after the changes in this pull request, yes.

> Why set it equal to 1 separately in each branch of the if statement?

Copy-n-modify and then diff-n-merge, or in other words: laziness.

jtclemm (approved):

PR #4772, opened 2025-11-11
Improve Kokkos support for hybrid `atom_vec` styles

stanmoore1:
**Summary**

This PR adds support for MPI buffer pack/unpack on device for Kokkos `atom_vec_hybrid` which was previously disabled. It greatly reduces duplicated code in the Kokkos `atom_vec` styles by combining all pack/unpack the methods into a single method in the parent class. It fixes a race condition with Kokkos `reverse comm` on GPUs when using GPU-aware MPI and removes view initialization from `MemoryKokkos::kokkos_realloc()`. It also removes fences when only 1 MPI rank is used, which gives a few % speedup in some benchmarks. It changes the `atom_mask` from a 32 bit integer to a 64 bit integer so that up to 64 masks can be used.

**Related Issue(s)**

None

**Author(s)**

Stan Moore (SNL)

**Licensing**

**Backward Compatibility**

Yes


weinbe2:
I gave this a look-through and everything looks solid. I love the massive reduction in the number of lines of code. Per an offline conversation there are still a few things Stan wants to double-check, but this receives a provisional approval "on the merits". Awesome work @stanmoore1

stanmoore1:
This has a few failing regression tests that I need to debug.

stanmoore1:
It turns out I had some false positives with the regression tests. I reran everything and it looks good, so this PR is ready to merge from my POV.

src/KOKKOS/atom_vec_kokkos.cpp
```diff
@@ -845,14 +1375,1530 @@ void AtomVecKokkos::unpack_reverse_kokkos(const int &n,
...
+  set_size_exchange();
+
+  if (nsend > (int) (k_buf.view_host().extent(0)*
+              k_buf.view_host().extent(1))/size_exchange) {
+    int newsize = nsend*size_exchange/k_buf.view_host().extent(1)+1;
+    k_buf.resize(newsize,k_buf.view_host().extent(1));
+  }
+
+  if (space == HostKK) {
+    if (size_exchange == 11) {
```
jtclemm: Maybe I'm missing something obvious, but why 11?

stanmoore1: @jtclemm Good question. It is the size of the "default" exchange, which is 3 velocity + 3 positions + 1 id + 1 mask + 1 type + 1 image + 1 extra for buffer = 11, see https://github.com/lammps/lammps/blob/develop/src/atom_vec.cpp#L41-L42. However this could change so I removed the hardcoded value in the last commit.

jtclemm: Ah, I didn't get the +1 buffer. Makes sense not to hardcode though.

jtclemm (approved):

PR #4775, opened 2025-11-16
fix bond/react: JSON output

jrgissing:
**Summary**

Adds *file* keyword to allow printing of reaction information (atom IDs, types, and coordinates)

**Related Issue(s)**

none

**Author(s)**

Jake Gissinger (Stevens)

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

none

**Post Submission Checklist**


src/output.h
```diff
@@ -72,16 +72,24 @@ class Output : protected Pointers {
 
   MPI_Datatype createParticleStructType();
 
+  struct JSON_Metadata {
```
akohlmey: @jrgissing I would prefer if we had only a forward declaration here and put it outside the Output class, but still inside the LAMMPS_NS namespace.

The name would be more consistent with the rest of LAMMPS if it was either `json_metadata` or `JSONMetadata`. I personally prefer the former, but both are consistent.

The actual definition of the struct would be put into a `json_metadata.h` header file which is included where needed.

src/output.h
```diff
@@ -72,16 +72,24 @@ class Output : protected Pointers {
...
-  void write_dump(bigint);                                 // force output of dump snapshots
-  void write_restart(bigint);                              // force output of a restart file
-  void write_molecule_json(FILE *, int, int, int *);       // output molecule JSON objects to file
-  void reset_timestep(bigint);                             // reset output which depends on timestep
-  void reset_dt();                                         // reset output which depends on timestep size
+  void setup(int memflag = 1);                        // initial output before run/min
+  void write(bigint);                                 // output for current timestep
+  void write_dump(bigint);                            // force output of dump snapshots
+  void write_restart(bigint);                         // force output of a restart file
+  void write_molecule_json(FILE *, int, int, int *, JSON_Metadata = JSON_Metadata{}); // JSON dump molecules
```
akohlmey: Following my suggestion from above, this would become something like:

```suggestion
  void write_molecule_json(FILE *, int, int, int *, json_metadata *mtd = nullptr); // JSON dump molecules
```

src/REACTION/fix_bond_react.h
```diff
@@ -67,6 +68,10 @@ class FixBondReact : public Fix {
   int newton_bond;
   FILE *fp;
   tagint lastcheck;
+  FILE *fpout;
+  bool outflag;
+  int json_init;
+  Output::JSON_Metadata rxn_metadata;
```
akohlmey: There should be only a forward declaration of the metadata struct here as well.
I can then be allocated in the constructor and deallocated in the destructor of the fix.

akohlmey (changes requested):
For maintenance reasons we should try to keep the number of declarations inside classes small, especially if those headers may be included in many places or are core LAMMPS classes.

I'm suggesting one way (most straightforward to me) to achieve this, but am open for discussing alternatives.

src/REACTION/fix_bond_react.h
```diff
@@ -67,6 +69,10 @@ class FixBondReact : public Fix {
   int newton_bond;
   FILE *fp;
   tagint lastcheck;
+  FILE *fpout;
+  bool outflag;
+  int json_init;
+  std::unique_ptr<json_metadata> rxn_metadata;
```
akohlmey, line 76: following the IWYU principle, you should add `#include <memory>` to this header.

akohlmey (approved):

PR #4791, opened 2025-11-25

### LINKED ISSUE #4797: file src/INTEL/TEST/mW.sw units inconsistency
The parameters in this file are mislabeled.
It cites Stillinger & Weber (1985), but the actual values correspond to the mW water model from:

Moore & Molinero, “Water Modeled as an Intermediate Element between Carbon and Silicon,” J. Phys. Chem. B 113, 4008 (2009).

Additionally, the parameters are expressed in real units, not metal units. For instance, ε = 6.189 is not in eV as stated, but in kcal/mol.

akohlmey:
**Summary**

**Related Issue(s)**

Fixes #4797 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

The GitHub Copilot coding instructions were mostly updated by GitHub Copilot.

**Backward Compatibility**

N/A

**Implementation Notes**

- remove references to https://ci.lammps.org and update information on automated testing
- update GitHub checkout action for testing workflows to version 6
- update GitHub coding instructions to follow best practices
- include LAMMPS version date only for 'release' and 'stable' branch when creating the manual. otherwise use 'Maintenance' (for 'maintenance 'branch) or 'Development' (for everything else)
- add deprecation messages for recently removed styles
- add more doxygen comments for LAMMPS class and Pointers class
- some more improvements for binary packaging on Linux
- remove some debug code
- reduce compiler warnings
- improve error messages and discussion of error details
- fix logic bug from last minute change in PR #4772
- update metadata in mW potential file in INTEL package tests

**Post Submission Checklist**


src/lammps.cpp
```diff
@@ -965,10 +996,13 @@ void LAMMPS::post_create()
   }
 }
 
-/* ----------------------------------------------------------------------
-   initialize top-level classes
-   do not initialize Timer class, other classes like Run() do that explicitly
-------------------------------------------------------------------------- */
+/** Initialize the LAMMPS top-level classes
```
stanmoore1: I think this type of comment style is harder to read and doesn't match the rest of LAMMPS.

src/pointers.h
```diff
@@ -98,41 +100,40 @@ class Pointers {
...
-
-  MPI_Comm &world;
-  FILE *&infile;
-  FILE *&screen;
-  FILE *&logfile;
-
-  class AtomKokkos *&atomKK;
-  class MemoryKokkos *&memoryKK;
-  class Python *&python;
+  LAMMPS *lmp;            ///< LAMMPS instance
```
stanmoore1: I think `///<` is ugly.

src/lammps.cpp
```diff
@@ -891,14 +913,23 @@ void LAMMPS::create()
...
-   check suffix consistency with installed packages
-   invoke package-specific default package commands
-     only invoke if suffix is set and enabled
-     also check if suffix2 is set
-   called from LAMMPS constructor and after clear() command
-     so that package-specific core classes have been instantiated
-------------------------------------------------------------------------- */
+/** Perform LAMMPS class tasks that require all base classes being set up
+ *
+\verbatim embed:rst
```
stanmoore1: Why are there `rst` commands in the comment? I think this makes the comment block harder to read.

stanmoore1 (commented):
Most of the changes look good but I'm concerned that adding doxygen and rst commands into comment blocks makes them ugly, harder to read, and doesn't match the rest of LAMMPS.

src/lammps.cpp
akohlmey: Show me a different way to transfer that kind of information into the manual without redundancies.  Some things take some getting used to, but the payoff in the long run is worth it.

We've transformed the C-language and Python-language interface so have only one documentation and set of comments so it is easier to keep it all consistent and up-to-date. The C++ programming interface is woefully missing; mostly because it is a lot of work. Since I found that GitHub Copilot is quite good at producing a usable draft for those comments, I am now working to have at least the generally usable members of the base classes documented.

src/pointers.h
akohlmey: Please check out https://doxygen.nl/manual/docblocks.html#memberdoc for alternatives. Happy to apply a different one.

src/lammps.cpp
akohlmey: This is the only way to insert references to other parts of the manual and write embedded documentation that is consistent with the rest of the manual.

While looking at the docs for the latest version of breathe, I've noticed an alternative tag where the asterisks can remain. Need to check how recent this is and if it works for all existing cases that we already have. https://breathe.readthedocs.io/en/latest/markups.html#handling-leading-asterisks

src/lammps.cpp
stanmoore1: I think we need to get input from other developers on this, e.g. @athomps and/or @sjplimp.

akohlmey: > I think we need to get input from other developers on this, e.g. @athomps and/or @sjplimp.

I understand the desire to keep things the way they are, but we are already too far down the road of adding Doxygen comments to reverse this process. The benefits, especially in the long run, outweigh the discomfort, IMO.  Like with other "modernization" changes (even though Doxygen is hardly modern), it takes some time to get used to it, embrace it, and use it to the advantage of LAMMPS users and developers, especially future ones.  Happy to continue discussing this in private.

akohlmey: > I think we need to get input from other developers on this, e.g. @athomps and/or @sjplimp.

@stanmoore1 FYI, I've moved the doxygen related changes out of this pull request into #4793 to have them all in one place and not block merging this PR due to this discussion.

jtclemm (approved):

jtclemm (approved):

src/lammps.cpp
stanmoore1: Thank you @akohlmey

PR #4795, opened 2025-11-28
Adding per-atom torque controls

jtclemm:
**Summary**

Mirroring fix addforce and setforce, this PR includes two new GRANULAR fixes for adding and setting peratom torques. To reduce the risk of confusion, the existing fix addtorque was renamed fix addtorque/group and the new fixes were named fix add(set)torque/atom.

I also reduced the number of references to `gm` in several granular submodel classes because I found them a bit annoying.

**Related Issue(s)**

NA

**Author(s)**

Joel Clemmer (SNL)

**Licensing**

**Backward Compatibility**

An alias of fix addtorque for fix addtorque/group was added to preserve backwards compatibility. @akohlmey, please let me know if this isn't the best way to do this.

**Implementation Notes**

NA

**Post Submission Checklist**


doc/src/fix_addtorque_group.rst
```diff
@@ -20,8 +20,8 @@ Examples
 
 .. code-block:: LAMMPS
 
-   fix kick bead addtorque 2.0 3.0 5.0
-   fix kick bead addtorque 0.0 0.0 v_oscillate
+   fix kick bead addtorque/group 2.0 3.0 5.0
+   fix kick bead addtorque/group 0.0 0.0 v_oscillate
 
 Description
 """""""""""
```
akohlmey, line 27:
```suggestion
"""""""""""

.. versionchanged:: TBD

   Fix *addtorque* was renamed to fix *addtorque/group*
```

We should document when this fix was renamed for people reading the online docs while using older versions of LAMMPS.

src/EXTRA-FIX/fix_addtorque_group.cpp
```diff
@@ -36,10 +36,10 @@ enum{NONE,CONSTANT,EQUAL,ATOM};
 
 /* ---------------------------------------------------------------------- */
 
-FixAddTorque::FixAddTorque(LAMMPS *lmp, int narg, char **arg) :
+FixAddTorqueGroup::FixAddTorqueGroup(LAMMPS *lmp, int narg, char **arg) :
   Fix(lmp, narg, arg)
 {
-  if (narg != 6) error->all(FLERR,"Illegal fix addtorque command");
+  if (narg != 6) error->all(FLERR,"Illegal fix addtorque/group command");
```
akohlmey, line 44:
```suggestion

  if ((strcmp(style, "addtorque") == 0) && (comm->me == 0))
    error->warning(FLERR,"Fix addtorque was renamed to fix addtorque/group. Please update your input");
```

We should inform people with legacy inputs of the rename.

akohlmey (changes requested):
In addition to the suggested changes, also `fix_addtorque.cpp` and  `fix_addtorque.h` need to be added to the file `src/Purge.list`

Furthermore, for the renamed and the added fixes the files `doc/src/fix.rst` and `doc/src/Commands_fix.rst` need to be updated.

Finally, the example input in `examples/PACKAGES/addtorque/` needs to be updated and new logs created to replace the old ones.

src/GRANULAR/fix_settorque_atom.cpp
```diff
@@ -0,0 +1,359 @@
...
+  if (utils::strmatch(update->integrate_style, "^respa")) {
+    nlevels_respa = (dynamic_cast<Respa *>(update->integrate))->nlevels;
+    if (respa_level >= 0)
+      ilevel_respa = MIN(respa_level, nlevels_respa - 1);
+    else
+      ilevel_respa = nlevels_respa - 1;
+  }
+
+  // cannot use non-zero forces for a minimization since no energy is integrated
+  // use fix addforce instead
```
jibril-b-coulibaly: This should be use "fix addtorque/atom" ?
Wider question: does minimization work with torques/rotational degrees of freedom for finite-size particles? If not, do we need the code below for whichflag ==2 ? Or should we provide a more general warning/error about minimization with rotational DOFs?

akohlmey:
> An alias of fix addtorque for fix addtorque/group was added to preserve backwards compatibility. @akohlmey, please let me know if this isn't the best way to do this.

No, this is a good way to make such a change. A few more additional steps are still needed, see my review comments and suggestions.

Once the alias has been around for long enough, we move it to `fix deprecated` (see PR #4791) and update `doc/src/Commands_removed.rst`, so that people will get a hard error with a proper explanation.

src/GRANULAR/fix_settorque_atom.cpp
```diff
@@ -0,0 +1,359 @@
...
+          torque[i][2] = zvalue;
+      }
+  }
+}
+
+/* ---------------------------------------------------------------------- */
+
+void FixSetTorqueAtom::post_force_respa(int vflag, int ilevel, int /*iloop*/)
+{
+  // set force to desired value on requested level, 0.0 on other levels
```
jibril-b-coulibaly: this should be "set torque to desired value"

jibril-b-coulibaly (approved):
I only have minor form comments, everything looks good to me otherwise.

On naming, I find the `/group` suffix to be possibly confused with the LAMMPS `group` command and concept, e.g., `fix settorque/atom` still applies to a `group`.
What is your opinion on using a word that refers to a collection of atoms, but does not already have a specific LAMMPS meaning (i.e., not `group`, `chunk`, `cluster`, `aggregate`, ...) instead of `group`, e.g.:

- `fix settorque/collection`
- `fix settorque/assembly`
- `fix settorque/set`

Thank you

akohlmey:
> On naming, I find the `/group` suffix to be possibly confused with the LAMMPS `group` command and concept, e.g., `fix settorque/atom` still applies to a `group`.

@jibril-b-coulibaly I disagree on this. The difference is that fix addtorque/group applies a toque to the whole group (by adding forces) while fix addtorque/atom applies it to the individual atoms in the group. So the naming choice seems very adequate to me.

The property that fixes and computes have fix/compute groups is a general property and thus need not be reflected in the name of the style. I don't think that will confuse anybody familiar with LAMMPS and beginners need to read the documentation well, anyway. Most of the suggested alternatives have no existing use in LAMMPS except for collection (for multi-style neighbor lists) and thus are much more confusing, IMO.

P.S.: the current naming scheme would also work well with a potential future fix addtorque/chunk which the net torque would be applied to groups of atoms defined by compute chunk/atom.

doc/src/fix_addtorque_group.rst
jtclemm: Yes, I missed those details. I believe I got them all now.

src/EXTRA-FIX/fix_addtorque_group.cpp
jtclemm: Ditto here

src/GRANULAR/fix_settorque_atom.cpp
jtclemm: That's a good question, I actually do not know if/how minimize works with torques. I would have to look that up.

src/GRANULAR/fix_settorque_atom.cpp
jtclemm: Good catch. There were lots of 'forces' hidden in the comments. I think I caught them all now.

src/GRANULAR/fix_settorque_atom.cpp
jtclemm: I think you are right that it doesn't work with torques (particularly since friction history is built up from integrating angular velocity). I removed that code block and mentions of minimization in the doc page.

akohlmey:
@jtclemm did some more corrections and updates to the documentation.

dsbolin (approved):

akohlmey (approved):

sjplimp:
@jtclemm I like the style names of the new and old addtorque commands.  I think they are clear.
One suggestion is to add a "..note" comment to the new torque/atom doc pages that it should not be
used to attempt to add/set torque for rigid bodies, i.e. by adding torque to particles in the rigid body.

In principle this could work, but would require the user knows what they are doing.  I.e. all atoms in
each body are included, the sum of per-atom added torques is the desired added torque for the body, etc.

If we want to allow this, possibly the code could check if atoms in rigid bodies are specified and give a warning.

sjplimp (changes requested):
see my comment about rigid bodies

jtclemm:
> One suggestion is to add a "..note" comment to the new torque/atom doc pages that it should not be
> used to attempt to add/set torque for rigid bodies, i.e. by adding torque to particles in the rigid body.

Good point. I added a note to both fixes. The one exception (actually one of the driving reasons for this PR) is if one simply wants to zero torques (and forces) to freeze rigid particles. Currently that could only be done by excluding those atoms from fix rigid (and therefore not creating a rigid body), but this won't work with some newer granular rigid pair styles being written that need information of the rigid body to calculate geometric terms. 

In the future, I may try to revisit how fix set/add force/torque and freeze interact with fix rigid. Maybe do something like what is done for fix gravity, but that's an issue for later. 

> If we want to allow this, possibly the code could check if atoms in rigid bodies are specified and give a warning.

Is there an easy way to check if an atom is in a rigid body?

akohlmey:
> Is there an easy way to check if an atom is in a rigid body?

Please check out `Modify::check_rigid_group_overlap()`.
```console
$ grep -l check_rigid_group_overlap fix*.cpp */fix*.cpp
fix_temp_berendsen.cpp
EXTRA-FIX/fix_temp_csld.cpp
EXTRA-FIX/fix_temp_csvr.cpp
```

sjplimp:
@jtclemm Conversely, the renamed addtorque/group commands could be used to change the torque on a rigid body if the group matched the body.  But only a single body, so probably not so useful.  But again, the use of those 2 commands with rigid bodies should probably be documented.

A variant that might be useful would be addtorque/chunk, which could be used to adjust the torque on each of a collection of rigid bodies.  E.g. make them all spin.

Another (future) idea would be to enhance the velocity command to allow the initial translational or rotational state of rigid bodies (or molecules) to be set.

jtclemm:
> @jtclemm Conversely, the renamed addtorque/group commands could be used to change the torque on a rigid body if the group matched the body. But only a single body, so probably not so useful. But again, the use of those 2 commands with rigid bodies should probably be documented.
> 
> A variant that might be useful would be addtorque/chunk, which could be used to adjust the torque on each of a collection of rigid bodies. E.g. make them all spin.
> 
> Another (future) idea would be to enhance the velocity command to allow the initial translational or rotational state of rigid bodies (or molecules) to be set.

I agree. The more I've looked at RIGID the more I've thought about how it could better integrate with many existing fixes/computes. Not a current task I want to take on, but maybe down the road.

sjplimp (approved):

PR #4804, opened 2025-12-09
Add ESP: a new FFT-based k-space solver (kspace_style esp) and matching Coulomb pair styles (coul/esp and lj/cut/coul/esp)

LiangJiuyang:
**Summary**

This pull request adds a prolate spheroidal wave function (PSWF)-based fast Ewald summation variant (“ESP”) to LAMMPS via a new KSpace style and two matching Coulomb pair styles:

New ```kspace_style esp``` that uses PSWF for the Ewald split and for particle–mesh spreading/interpolation, with separate “splitting accuracy” and “spreading accuracy” inputs. 

New ```pair_style coul/esp``` intended to be used together with ```kspace_style esp``` (analogous to coul/long + pppm). 

New ```pair_style lj/cut/coul/esp``` intended to be used together with ```kspace_style esp``` (analogous to lj/cut/coul/long + pppm). 

We follow the standard LAMMPS user interface patterns (e.g., ```kspace_modify mesh ... order ...```) in our implementation, so the new ESP solver is compatible with most existing PPPM functionality.

Background and motivation: ESP replaces both the Gaussian used in classical Ewald splitting and the B-splines used in particle–mesh operations (spreading and interpolation) with PSWF-based kernels, reducing FFT mesh size requirements and the spreading/interpolation order needed to reach a given accuracy. For a given error tolerance and spreading/interpolation ```order```, the new ```coul/esp``` + ```kspace/esp``` can require up to an order of magnitude fewer FFT grids than the ```coul/long``` + ```kspace/pppm``` solver. Please refer to our paper published on [Nature Communications](https://www.nature.com/articles/s41467-026-73232-8) for more details.

**Author(s)**

Jiuyang Liang — Flatiron Institute, Simons Foundation

Libin Lu — Flatiron Institute, Simons Foundation 

Shidong Jiang — Flatiron Institute, Simons Foundation 

co-developers: Alex Barnett, Leslie Greengard (Flatiron Institute, Simons Foundation)

Please contact  jliang@flatironinstitute.org, llu@flatironinstitute.org, sjiang@flatironinstitute.org for any questions.

**Licensing**

**Backward Compatibility**

Backward compatible. Existing inputs are unaffected; this PR adds new styles (```esp```, ```coul/esp```, ```lj/cut/coul/esp```) without changing the behavior of existing styles such as ```pppm``` and ```lj/cut/coul/long```.

**Implementation Notes**

Algorithmic change: PSWF kernel splitting (instead of Gaussian) reduces the required Fourier grid for a target accuracy; PSWF-based particle–mesh operations (instead of B-splines) reduce the needed spreading/interpolation order. 

User-facing usage is intended to mirror PPPM workflows. Example (from the provided quick-start pattern):

```pair_style coul/esp 9```

```pair_style lj/cut/coul/esp 9 9 (optional)```

```kspace_style esp 1e-4 1e-4 (optional)```

```kspace_modify mesh 100 100 100 order 5```

Verification/validation: We compared forces and energies against standard PPPM using matched accuracy tolerances. We also added unit tests in the ```unittest``` tree to check stability and accuracy. Performance benchmarks are reported in our [paper](https://www.nature.com/articles/s41467-026-73232-8); they show that the ESP solver is typically between 2.5× and 6× faster in both single-core and multi-core runs.

**Post Submission Checklist**

Reference describing the ESP method and its integration into MD codes: https://www.nature.com/articles/s41467-026-73232-8.

### CODE REVIEW

File: .github/workflows/unittest-linux.yml [hunk 1]
```diff
@@ -19,7 +19,7 @@ concurrency:
 jobs:
   build:
     name: Linux Unit Test
-    if: ${{ github.repository == 'lammps/lammps' }}
+    if: ${{ github.repository == 'LiangJiuyang/lammps' }}
```
akohlmey: Please make sure you revert this change when marking this pull request ready for review. Thanks.

doc/src/Commands_kspace.rst
```diff
@@ -18,6 +18,7 @@ OPT.
    * :doc:`msm (o) <kspace_style>`
    * :doc:`msm/cg (o) <kspace_style>`
    * :doc:`msm/dielectric <kspace_style>`
+   * :doc:`esp <kspace_style>`
```
akohlmey: The entries in this list are supposed to be sorted alphabetically

doc/src/kspace_style.rst
```diff
@@ -164,7 +169,7 @@ matching keyword to the name of the KSpace style, as in this table:
 +----------------------+-----------------------+
 | Pair style           | KSpace style          |
 +----------------------+-----------------------+
-| coul/long            | ewald or pppm         |
+| coul/long            | ewald or esp or pppm  |
```
akohlmey: I am confused by this change. Does kspace style esp require a /esp pair style or is also compatible with /long?

src/pair.h
```diff
@@ -228,6 +229,8 @@ class Pair : protected Pointers {
   virtual void min_xf_get(int) {}
   virtual void min_x_set(int) {}
   virtual void transfer_history(double *, double *, int, int) {}
+  virtual double atom2cut(int) { return 0.0; }
```
akohlmey: This looks like you are re-introducing code that had been removed from LAMMPS.

.github/workflows/unittest-linux.yml
LiangJiuyang: Thanks for the reminder. I have reverted this change.

doc/src/Commands_kspace.rst
LiangJiuyang: Thanks for pointing that out. I have updated the list so the entries are now sorted alphabetically.

doc/src/kspace_style.rst
LiangJiuyang: It was a mistake in the previous version. kspace_style esp is not compatible with /long; it requires the matching /esp pair styles. I’ve updated the PR accordingly.

src/pair.h
LiangJiuyang: Thanks for pointing this out. I have removed them from the PR.

LiangJiuyang:
@akohlmey @stanmoore1 Thanks. Since this PR is a change to KSPACE, we would really appreciate any early feedback so we can iterate and improve. Any suggestions would be very helpful.

akohlmey:
@LiangJiuyang I just requested an AI code review for your pull request. Even though none of the LAMMPS developers seems to have much time available right now, this could help spotting some issues. But please be careful about the results. I have recently seen many incorrect suggestions like asking to use an obsolete internal API instead of the modern one.

LiangJiuyang:
> @LiangJiuyang I just requested an AI code review for your pull request. Even though none of the LAMMPS developers seems to have much time available right now, this could help spotting some issues. But please be careful about the results. I have recently seen many incorrect suggestions like asking to use an obsolete internal API instead of the modern one.

No worries, I will keep up with the develop branch and improve the code. Happy holidays!

src/kspace.cpp
```diff
@@ -92,9 +93,10 @@ KSpace::KSpace(LAMMPS *lmp) :
   mixflag = 0;
 
   splittol = 1.0e-6;
-  scale = 1.0;
 
   maxeatom = maxvatom = 0;
+  eatom = nullptr;
+  vatom = nullptr;
```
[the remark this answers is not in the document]
akohlmey: In addition the lines initializing `eatom` and `vatom` are redundant since the two pointers are already set to null in the initializer list of the constructor.

src/kspace.cpp
```diff
@@ -33,15 +33,16 @@ static constexpr double SMALL = 0.00001;
 
 /* ---------------------------------------------------------------------- */
 
-KSpace::KSpace(LAMMPS *lmp) :
-    Pointers(lmp), eatom(nullptr), vatom(nullptr), gcons(nullptr), dgcons(nullptr)
+KSpace::KSpace(LAMMPS *lmp) : 
+    Pointers(lmp), eatom(nullptr), vatom(nullptr), gcons(nullptr), dgcons(nullptr),
+    force_poly_coeff(nullptr), energy_poly_coeff(nullptr), Fourier_poly_coeff(nullptr), Fourier_spreading_coeff(nullptr)
```
akohlmey: Please keep a consistent naming scheme of the class members and make the "Fourier" lower case.

src/pair.cpp
```diff
@@ -431,17 +436,39 @@ void Pair::init_tables(double cut_coul, double *cut_respa)
       egamma = 1.0 - (r/cut_coul)*force->kspace->gamma(r/cut_coul);
       fgamma = 1.0 + ((double)rsq_lookup.f/cut_coulsq)*
         force->kspace->dgamma(r/cut_coul);
-    } else {
+    }
+    else if (espflag)
```
akohlmey: Please follow the LAMMPS programming style and place the opening brace of a code block on the same line as the `if`, `else`,`else if`, `for`, or `while` statements. And also put the `else` or `else if` on the same line of the closing brace of the previous block.

src/KSPACE/esp.h
```diff
@@ -0,0 +1,245 @@
...
+  virtual void allocate_groups();
+  virtual void deallocate_groups();
+  virtual void make_rho_groups(int, int, int);
+  virtual void poisson_groups(int);
+  virtual void slabcorr_groups(int, int, int);
+
+  // =========================================
+  // Horner's method for polynomial evaluation
+  // =========================================
+
```
akohlmey: All of these methods don't need to be in the class header and should be moved into an anonymous namespace in the corresponding .cpp file. 

Please also note that class member functions are automatically considered as `inline`.

src/KSPACE/esp.h
```diff
@@ -0,0 +1,245 @@
...
+#ifdef KSPACE_CLASS
+// clang-format off
+KSpaceStyle(esp,ESP);
+// clang-format on
+#else
+
+#ifndef LMP_ESP_H
+#define LMP_ESP_H
+
+#include "error.h"
```
akohlmey: Please follow the IWYU principle and only include headers that are used in the header.
I don't see any use for the `Error` class or `MathConst` namespace here.

src/pair.h
```diff
@@ -229,6 +230,7 @@ class Pair : protected Pointers {
   virtual void min_xf_get(int) {}
   virtual void min_x_set(int) {}
   virtual void transfer_history(double *, double *, int, int) {}
+  virtual double radii2cut(double, double) { return 0.0; }
```
akohlmey: This seems to be re-introducing code that was already removed from LAMMPS.

src/kspace.cpp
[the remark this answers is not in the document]
LiangJiuyang: Thanks! I removed the initialization of `eatom` and `vatom`. It looks like `scale = 1.0;` is set in both lines 89 and 95 in the development version of LAMMPS.

src/pair.h
LiangJiuyang: Thanks! I removed this line.

src/KSPACE/esp.h
LiangJiuyang: Thanks! I have removed the use of `Error` class. Since I use `MY_PI` in `esp.h`, I now use `MathConst::MY_PI;` instead.

src/KSPACE/esp.h
LiangJiuyang: Thank you for these comments. I have removed this class header and removed all `inline` keywords.

src/pair.cpp
LiangJiuyang: Thank you. The style conflict has been resolved.

src/kspace.cpp
LiangJiuyang: Thanks! I have renamed `Fourier_poly_coeff` to `splitting_poly_coeff` and `Fourier_spreading_coeff` to `spreading_poly_coeff` to keep the naming consistent.

stanmoore1:
I tested this using the following script below. It is working, but the error estimator seems a little off for some configurations. For example, with cutoff=10.0 and requested relative accuracy = 1e-5, I get an actual relative error of 0.0003, **which is 30x higher than requested**. In contrast PPPM gives 6.8e-06 which is pretty close.

```
log             tmp.log
variable        n equal 100                # number of particles (choose an even integer)
variable        box_length equal 10        # size of the box in real units
variable        rmin equal 2               # minimum separation between particles

units           real
atom_style      charge
atom_modify     map array

lattice         sc 1.0
region          box block 0 ${box_length} 0 ${box_length} 0 ${box_length}
create_box      1 box

variable        q equal 1
variable        x equal random(xlo,xhi,123456)
variable        y equal random(ylo,yhi,123456)
variable        z equal random(zlo,zhi,123456)
label           loopa
variable        a loop $n
  label           loopb
  label           try_again
  variable        xtmp equal $x
  variable        ytmp equal $y
  variable        ztmp equal $z
  variable        b loop $a
    if              '$a == $b' then "jump SELF break"
    variable        xb equal x[$b]
    variable        yb equal y[$b]
    variable        zb equal z[$b]
    variable        r equal sqrt((${xtmp}-${xb})^2+(${ytmp}-${yb})^2+(${ztmp}-${zb})^2)
    if              '$r < ${rmin}' then "jump SELF try_again"
    label           break
  next            b
  jump            SELF loopb
  create_atoms    1 single $x $y $z units box
  variable        q equal -$q
  set             atom $a charge $q
  print           "Atom number $a created."
next            a
jump            SELF loopa

mass            1 1.0
pair_style      lj/cut/coul/long 12.0
pair_coeff      1 1 0.0 0.0
pair_modify     table 0 compute yes

mass    1 1.0

kspace_style    ewald 1e-20
kspace_modify   compute yes
run             0
fix             ref all store/state 0 fx fy fz

pair_style      lj/cut/coul/esp 10.0
#pair_style      lj/cut/coul/long 10.0
pair_coeff      1 1 0.0 0.0
pair_modify     table 0 compute yes
kspace_style    esp 1e-5
#kspace_style    pppm 1e-5
thermo_style    one
run             0
fix             test all store/state 0 fx fy fz
variable        ferrsq atom ((f_ref[1]-f_test[1])^2+(f_ref[2]-f_test[2])^2+(f_ref[3]-f_test[3])^2)
compute         avefersq all reduce ave v_ferrsq
thermo_style    custom step c_avefersq cpu
log             test12.log
run             100
variable        rms_force_error equal sqrt(c_avefersq)
variable        rms_rel_force_error equal sqrt(c_avefersq)/332.063
print           "computed absolute RMS force accuracy = ${rms_force_error}"
print           "computed relative RMS force accuracy = ${rms_rel_force_error}"
unfix           test
uncompute       avefersq
```

stanmoore1:
If I increase to cutoff=12 with the same requested relative accuracy = 1e-5, I get an actual relative error of 4.35455938826383e-06, similar to PPPM. And if I decrease the cutoff=4, I get 1.8e-05 which is also close. So the issue seems to be in the error estimator, not the actual method itself.

stanmoore1 (changes requested):
@LiangJiuyang The code looks good to me, and I think this is almost ready to merge, except can you please address my comment about the error estimator having poor accuracy?

stanmoore1:
@LiangJiuyang do the changes from #4939 need to be applied here for ESP as well? E.g. enabled here: https://github.com/lammps/lammps/blob/51560228a38e414901e06a3202d1975aa3d96ddb/src/kspace.cpp#L510

LiangJiuyang:
> @LiangJiuyang The code looks good to me, and I think this is almost ready to merge, except can you please address my comment about the error estimator having poor accuracy?

Thanks, @stanmoore1. We tightened the tolerance to improve the accuracy of the error estimator. We reran your script, and the updated results are shown below.  Could you please check whether this addresses your concern? Please let us know if any further changes are needed.

| cutoff | ESP grid | ESP rel RMS | PPPM grid | PPPM rel RMS |
|---:|:---:|---:|:---:|---:|
| 4  | 12^3 | 9.62e-6 | 30^3 | 1.08e-5 |
| 5  | 10^3 | 7.43e-6 | 24^3 | 8.50e-6 |
| 6  | 8^3 | 4.54e-6 | 20^3 | 9.15e-6 |
| 8  | 6^3 | 2.58e-6 | 15^3 | 7.22e-6 |
| 10 | 5^3 | 2.21e-6 | 12^3 | 6.78e-6 |
| 12 | 4^3 | 1.41e-6 | 10^3 | 5.04e-6 |

stanmoore1 (approved):

LiangJiuyang:
> @LiangJiuyang do the changes from #4939 need to be applied here for ESP as well? E.g. enabled here:
> 
> https://github.com/lammps/lammps/blob/51560228a38e414901e06a3202d1975aa3d96ddb/src/kspace.cpp#L510

Yes, in principle, this could be applied to ESP as well. However, the ESP case requires additional mathematical derivation and validation before enabling it. I think it would be better to handle this in a separate PR. For now, I have updated the code to disable this option when ESP is enabled.

stanmoore1:
Sounds good, thanks. @akohlmey This is ready to merge from my POV.

PR #4808, opened 2025-12-11
ML-IAP bugfix

wcwitt:
Follow-on PR to #4744 that fixes bugs introduced by the unrolled version. Apologies that I didn't catch them sooner - I'm frankly amazed the profiling didn't crash. However, looking back, the trajectories were stable but indeed wrong.

@bathmatt @fglines-nv


bathmatt:
sorry, not sure how this didn't get caught either, thanks!!

stanmoore1 (approved):
Thanks @wcwitt. Some of the mliap examples are used as regression tests which run on GPUs with Kokkos CUDA and those tests are all still passing too, weird.

wcwitt:
I did wonder about tests. Is it possible that none of the regression tests use message passing?

stanmoore1:
> I did wonder about tests. Is it possible that none of the regression tests use message passing?

Yes that must be the case.

PR #4809, opened 2025-12-12
Fix infinite loop crash in BondHybrid and PairHybrid copy_svector

poyais001:
**Summary**

This PR fixes a critical bug in `BondHybrid::copy_svector` and `PairHybrid::copy_svector` that caused an infinite loop and subsequent segmentation fault (crash) due to out-of-bounds memory access.

The issue was caused by an incorrect `for` loop termination condition:
```cpp
// Old (Incorrect)
for (int l = 0; this_style->single_extra; ++l) 

// New (Correct)
for (int l = 0; l < this_style->single_extra; ++l)
```

The original code checked the constant value this_style->single_extra instead of comparing it against the loop iterator l. This caused the program to crash on the first invocation of these functions when single_extra was non-zero.

This bug appears to be specific to the base BondHybrid and PairHybrid classes. I verified that other hybrid styles (e.g., hybrid/overlay) correctly implement this loop, which likely explains why this issue persisted unnoticed for some time.

**Related Issue(s)**

**Author(s)**

**Licensing**

**Backward Compatibility**

This change is a bug fix and does not break backward compatibility. It ensures the code behaves as originally intended.

**Implementation Notes**

The fix involves changing the loop condition from `this_style->single_extra` to `l < this_style->single_extra` in both `bond_hybrid.cpp` and `pair_hybrid.cpp`. This ensures the loop iterates exactly single_extra times.

**Post Submission Checklist**

### DIFF
--- a/src/bond_hybrid.cpp
+++ b/src/bond_hybrid.cpp
@@ -469,7 +469,7 @@ void BondHybrid::copy_svector(int type)
   // there is only one style in bond style hybrid for a bond type
   Bond *this_style = styles[map[type]];
 
-  for (int l = 0; this_style->single_extra; ++l) { svector[l] = this_style->svector[l]; }
+  for (int l = 0; l < this_style->single_extra; ++l) { svector[l] = this_style->svector[l]; }
 }
 
 /* ----------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


sjplimp (approved):
@poyais001 thanks for catching this !

PR #4812, opened 2025-12-14
Support rendering graphics objects from fixes, add a dedicate graphics objects fix, and add and enable transparency handling in dump image

akohlmey:
**Summary**

This pull request adds support for passing information of graphical objects from fixes to dump image (or dump movie).

Specifically, most wall fixes (except those for regions), fix indent, fix reaxff/bonds, and fix smd/wall_surface are supported with the changes in this pull request.

Also, a new fix, `fix graphics` was added which allows to add individual graphic objects like spheres, cylinders to the image, but also more complex items like arrows or a progress bar. These may change or move during a simulation, since many parameters of the objects can be entered as variable references.

Finally, a so-called 'screen-door transparency' effect was added to the `Image` class that does the low-level rendering for `dump image`. Unlike real transparency, this works by not writing all pixels of an object.  This implementation uses a 16x16 pixel threshold matrix for 257 different opacity levels. This can be utilized for atoms, bonds, regions, walls and surface meshes either through dump image arguments or dump_modify commands.

Changing the dump image command in the indent example to:
```
dump           2 all image 1000 image.*.jpg type type &
               zoom 1.6 adiam 1.5 fix 4 const 0.0 2.0
dump_modify    2 pad 5 fcolor 4 purple
```

produces the following kind of image with the indenter rendered in purple:
image 26000

Adding the following lines to the peptide example:
```
variable prog equal step/1000
fix graphics all graphics 25 progbar 1 3 z $(xhi) $(ylo-7.0) $(zlo+0.5*(zhi-zlo)) $(lz) 1.0 v_prog 10

dump           3 peptide movie 25 movie.mkv element type zoom 1.2 fsaa yes  &
               axes no 0.8 0.02 view 80 -10 bond atom 0.5 fix graphics element 0 0 size 800 800
dump_modify    3 pad 3 element C C O H N C C C O H H S O H acolor 10 silver
```
produces the following animation:
peptide

The following image shows the pseudo-transparency effect:
funnel

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

The `zcylinder` option was removed from `fix wall/gran` since it is superseded by `fix wall/gran/region'.

**Post Submission Checklist**


doc/src/fix_graphics.rst
```diff
@@ -0,0 +1,174 @@
...
+command keywords for the simulation box parameters and timestep and
+elapsed time.  Thus it is easy to specify graphics object properties
+like position, orientation, radius or more that change as a function of
+time or span consecutive runs in a continuous fashion.  For the latter,
+see the *start* and *stop* keywords of the :doc:`run <run>` command and
+the *elaplong* keyword of :doc:`thermo_style custom <thermo_style>` for
+details.
+
+For example, if a sphere's x-position is specified as v_x, then this
+variable definition will keep it's center at a relative position in the
```
jtclemm: its center

doc/src/fix_indent.rst
```diff
@@ -201,6 +201,35 @@ contains *xlat*, *ylat*, *zlat* keywords of the :doc:`thermo_style
...
+
+The *fflag1* setting of *dump image fix* has no impact on rendering a
+spherical indenter or a planar indenter in 3d systems.  For a
+cylindrical indenter and a planar indenter in 2d systems it determines
+whether the cylinder is capped with a sphere at the ends: 0 means no
+caps, 1 means the lower end is capped, 2 means the upper end is capped,
+and 3 means both ends are capped.
+
+The *fflag2* setting allows you to adjust the radius of the rendered
+object for spherical indenters, cylindrical indenters, and planar
```
jtclemm, line 225: I'm a bit confused by this. Is this option specifying a change in the rendered radius? So < 0 => the object looks smaller and >0 => the object looks larger? This is not directly specifying a radius where < 0 is meaningless?

doc/src/fix_graphics.rst
```diff
@@ -0,0 +1,174 @@
...
+
+Fix graphics is designed to be used with the *fix* keyword of :doc:`dump
+image <dump_image>`.  The fix will pass geometry information about the
+objects listed on the command line to *dump image* so that they are
+included in the rendered image.
+
+The *fflag1* setting of *dump image fix* determines whether cylinder
+elements are capped with spheres:  0 means no caps, 1 means the lower
+end is capped, 2 means the upper end is capped, and 3 means both ends
+are capped.  This applies to the *cylinder* object and also to the
```
jtclemm, line 169: So 3 is a spherocylinder?

doc/src/fix_graphics.rst
akohlmey: fixed. the text was copied with small adjustments from fix indent, so it fixed it there as well.

doc/src/fix_indent.rst
akohlmey: Exactly. The `Fix::image()` API does not allow to pass any settings to the fix (contrato what the docs in the develop branch say) and the fix returns the radius (or rather diameter) that is used in the indent force computation. For now, I stuck without changing the API (I don't want to break the granular surfaces branch more than I already did), but I am open for changes. I would not like to make changes to the individual fixes, though. That would result in a lot of replicated work.

doc/src/fix_graphics.rst
akohlmey: According to Wikipedia, yes. https://en.wikipedia.org/wiki/Capsule_(geometry)
I don't recall having heard or read that term (or using "capsule" for that matter) for a capped cylinder. It would be nice, if the LAMMPS Image class had an option to put a circle shape on top of that (or render a cone).

src/fix_graphics.cpp
```diff
@@ -0,0 +1,677 @@
...
+  nevery = utils::inumeric(FLERR, arg[3], false, lmp);
+  if (nevery <= 0) error->all(FLERR, 3, "Illegal fix graphics nevery value {}", nevery);
+  global_freq = nevery;
+
+  numobjs = 0;
+  varflag = 0;
+
+  int iarg = 4;
+  while (iarg < narg) {
+    if (strcmp(arg[iarg], "sphere") == 0) {
```
jtclemm, line 56: I understand arrows/progress bars, but can't one just draw a sphere/cylinder using the region command? Those two options seem redundant.

doc/src/dump_image.rst
```diff
@@ -504,27 +507,44 @@ change this via the dump_modify command.
...
+* :doc:`fix wall/lj126 <fix_wall>`
+* :doc:`fix wall/lj1043 <fix_wall>`
+* :doc:`fix wall/colloid <fix_wall>`
+* :doc:`fix wall/harmonic <fix_wall>`
+* :doc:`fix wall/harmonic/outside <fix_wall>`
+* :doc:`fix wall/lepton <fix_wall>`
+* :doc:`fix wall/morse <fix_wall>`
+* :doc:`fix wall/reflect <fix_wall_reflect>`
+* :doc:`fix wall/reflect/stochastic <fix_wall_reflect_stochastic>`
+* :doc:`fix wall/table <fix_wall>`
```
jtclemm, line 549: Fix wall/gran might be another

src/fix_graphics.cpp
akohlmey: Using regions for just visualization seem a bit overkill to me. The benefit of fix graphics is that you can draw multiple objects in just a single command and a single addition to dump image/movie. For that reason, I am thinking about adding a triangle object as well.

doc/src/dump_image.rst
akohlmey: I'll have a look.

akohlmey: @jtclemm It is done for planes. Do you have an example for *zcylinder*?
I am not going to implement this for `fix wall/gran/region since there the existing render option for regions is the better way.

The following was created by adding the following lines to the `pour/in.pour` example:
```
dump viz all image 1000 myimage-*.jpg type diameter size 600 600 zoom 1.331 &
     shiny 0.2 fsaa yes view 50 20 ssao yes 453983 0.75 &
     box yes 0.025 axes no 0.0 0.0 center s 0.5 0.5 0.5 &
     fix zlower const 0 0
dump_modify viz pad 9 boxcolor yellow backcolor silver fcolor zlower yellow
```
myimage-000004000

Similarly for 2d systems here is a command for `pour/in.pour.2d`:
```
dump viz all image 500 myimage-*.ppm type diameter size 800 400 zoom 3.0 &
     shiny 0.2 fsaa yes box no 0.005 axes no  1.0 0.1 center s 0.5 0.5 0.5 &
     fix xwalls const 3 0.2 fix ywalls const 3 0.3
dump_modify viz pad 9 backcolor black fcolor xwalls white fcolor ywalls white
```
pour2d

src/image.cpp
```diff
@@ -44,20 +44,170 @@ using MathConst::DEG2RAD;
...
+constexpr int NCOLORS = 140;
+constexpr int NELEMENTS = 109;
+constexpr double EPSILON = 1.0e-6;
+
+enum { NUMERIC, MINVALUE, MAXVALUE };
+enum { CONTINUOUS, DISCRETE, SEQUENTIAL };
+enum { ABSOLUTE, FRACTIONAL };
+enum { NO, YES };
+
+// #define USE_BAYER_MATRIX
```
akohlmey: @jtclemm @sjplimp By commenting out this define, you can switch between two different dithering patterns and it would be nice, if you could give some feedback which one looks better. The default version has a bit more "randomness", while the Bayer matrix is more regular. Since the latter is generated with a simple recursive formula, it would be possible to generate a larger matrix (32x32 or 64x64 and so on) with correspondingly more levels of effective transparency and a larger "pattern" due to the larger tiles. For me either has benefits and shortcomings.

src/fix_graphics.cpp
jtclemm: Overkill in which manner? Is it clunky to add them to the dump image command (does a union work)? Or as in it's slow? Or something else?

doc/src/dump_image.rst
jtclemm: Oh nice! I assumed that'd be an easy ctrl+c ctrl+v from fix wall, but I actually never knew fix gran/wall had a _zcylinder_ option. 

I don't think there's any example of it being used and, to be honest, I would argue it could be deprecated. But if you want an example, the heat conduction system could be converted from fix wall/gran/region to fix wall/gran _zcylinder_ with a line change. That might also be the one scenario where a wall has a time varying property, the temperature.

src/fix_graphics.cpp
akohlmey: You have to define a new region for each object and it has a lot of unwanted data and possible replicated input commands associated with it, if you want multiple things. For just a single object, it would be about the same. For cylinder also the interface is different (you provide two endpoint centers which in a region you have to choose a direction and then translate and rotate the region), so it offers an alternate way that may or may not be more convenient. With fix graphics, you can easily define several objects or some more complex compound (e.g. from a pending future interface to triangles). Plus from the implementation side, the fix version is far simpler to implement and maintain. From my perspective, the two features complement each other.

One important thing to note is that cylinder region visualizations are assembled from triangles, since the cylinder graphics primitive is optimized for use with a sphere at the end and does not render the backside (which is not visible if you have spheres at the ends). So a cylinder region is a bit different from a cylinder primitive (which does not offer a flat cap).

doc/src/dump_image.rst
akohlmey: > Oh nice! I assumed that'd be an easy ctrl+c ctrl+v from fix wall, but I actually never knew fix gran/wall had a _zcylinder_ option.

It was not as simple as that, since the selection and interface is quite different. In fix wall/harmonic and derived classes as well as in fix wall/reflect and derived classes, you can have up to six walls at the same time. In fix wall/gran you must choose one direction and can have one or two walls for that. I had to do some extra programming to map it to the FixWall scheme and then use the same function to update the list of graphics objects for all three families of wall fixes.

src/image.cpp
jtclemm: Personally, I've never really liked the repetition that ordered dithering can create. But that's just a general preference. Do you happen to have examples of what an atomistic system looks like with either option? I can also run just run it this evening.

Are there particular implementation/performance differences between the two?

doc/src/dump_image.rst
akohlmey: > to be honest, I would argue it could be deprecated

I agree with that. Especially, since it only supports the z direction and no rotation.
I would assume its implementation predates fix wall/gran/region. Removing it also removes maintenance effort since it is superseded by fix wall/gran/region.

src/image.cpp
akohlmey: > Personally, I've never really liked the repetition that ordered dithering can create. But that's just a general preference.

I agree with that, but it was dead simple to implement and *very* fast. For me personally, it is also a reminder of my early days of using a computer where this kind of technique was commonplace since computers had graphics with 4-bit or 8-bit color spaces (unlike the 24-bit common now) and far less powerful CPUs. I prefer real transparency as produced by a ray tracer or even the approximation that OpenGL offers,
but that is much slower. I have been thinking for a long time, that we could have a dump style opengl, povray, or tachyon and would allow a more potent rendering. It may be as simple as replacing the Image class with an abstract class and then use polymorphism. But that is another project...

The appeal for using a simple and fast approximate method is for me the option to interface to it in LAMMPS-GUI to get a quick look at how things are set up and having some items in the foreground (like walls or regions) not obscure things behind them.
For regions I have implemented a wireframe option for that purpose, since transparency was not available at the time. Had it been, I would have avoided the extra effort.

> Do you happen to have examples of what an atomistic system looks like with either option? 

Not right now, but I can create some in a few hours.

doc/src/dump_image.rst
jtclemm: Let's cut it then. I can do that next time I pull together a set of updates (unless you want to remove it before then).

src/image.cpp
akohlmey: > Are there particular implementation/performance differences between the two?

No. They are just different patterns and the values are all pre-computed. This is what makes this pseudo-transparency so fast. With the Bayer scheme, I could generate a larger matrix and thus the repetition would be more coarse, but for that there could be a performance impact since the matrix would occupy more space in the CPU cache or not fix at all. A rank 16 matrix takes 2kB a rank 32 matrix 8kB a rank 64 matrix 32kB.

jtclemm: By simple and fast, do you mean relative to the more typical error diffusion dithering algorithms? Or compared to ray tracing/opengl?

akohlmey: Fast compared to (software) OpenGL (as implemented in the Mesa library, for example) and ray tracing (*much* faster than that).
And I cannot imagine anything simpler to implement that the screen-door method, which consists of adding a single if statement where the pixel position is compared to a value in the threshold matrix with the pattern.

doc/src/dump_image.rst
akohlmey: It is easy to remove it now since I am already messing with the source files and the documentation. I don't know what it is, but some pull requests have the property of blowing up in an almost uncontrolled fashion and what started as a small change to add one simple feature end as a significant rewrite with a bunch of new features. It is as if each feature inspires new ones (and then they breed like rabbits :wink:)

src/image.cpp
akohlmey: > do you mean relative to the more typical error diffusion dithering algorithms

Please keep in mind that I was trained in college as a (wet lab) chemist and then later graduated over classical MD and HPC. So my training in physics and especially applied math is lacking quite a bit. I am dependent on what information readily available to me via web searches and in an accessible form. I basically know about the methodology as much as I learned in the last 24 hours. :disappointed:
So I cannot really give you an in-depth discussion.

That said, I think what is implemented is very adequate for the intended purpose and assuming that over 99% of the use of dump image will not require transparency at all.
This fits very well with the "good enough" paradigm in LAMMPS-GUI where I see transparency most likely used.  For anything else, it is not worth investing much effort in the internal visualization but to revert to external tools that specialize in visualization and are much more capable.

akohlmey: > Do you happen to have examples of what an atomistic system looks like with either option?

Here we go with three images of the rhodo benchmark. One opaque, two with 30% opacity for the water molecules with the (more regular) Bayer matrix and one with the other matrix (in that order).
rhodo-opaque

rhodo-non-bayer

rhodo-bayer

jtclemm: I like the 3rd one best. So the "other" matrix.

akohlmey: > I like the 3rd one best. So the "other" matrix.

Actually, I mislabeled them. But here is another reason to favor it. I just found out through experimentation that at opacity levels of exactly 25%, 50%, and 75% you can't see a pattern if you also turn on FSAA.
rhodo-fsaa

So this is the decider for me. I also added a note to the documentation about these settings.

ndtrung81: Very nice improvement for rendering images!

ndtrung81 (approved):

jtclemm (approved):

PR #4817, opened 2025-12-19
Visualize body particle faces and ellipsoid particles

akohlmey:
**Summary**

This pull request adds the missing code to enable drawing body particle faces for body styles `rounded/polyhedron` and `rounded polygon`. The existing code is somewhat refactored to be simpler and more like that for tri/line particles. Also some minor issues from previous `dump image` related pull requests are addressed. In addition, this pull request adds support for rendering ellipsoid particles by approximating the ellipsoid as a mesh of triangles.
Furthermore, also an arrow graphics object (constructed from triangles and a cylinder) was implemented and added to fix graphics.

image 001000
myimage-000030000

ellipsoid-optimized

arrows-test

**Related Issue(s)**

The changes in this pull request are further extended in PR #4819 and parts of it refactored in PR #4820. Thus any changes about the details of the implementation are better done in PR #4820 and comments added there.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Artificial Intelligence (AI) Tools Usage**

Some of the low-level graphics code was prototyped by Claude Sonnet (through Perplexity.ai) and Google Gemini. It has been significantly refactored, though.

**Backward Compatibility**

N/A

**Implementation Notes**

The body style `image()` function now increments the number of graphics elements it computed.
Some invariant calculations were moved outside the loops.

**Post Submission Checklist**

- [x] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [x] The source code follows the LAMMPS formatting guidelines
- [x] Suitable new documentation files and/or updates to the existing docs are included
- [x] The added/updated documentation is integrated and tested with the documentation build system
- [x] The feature has been verified to work with the conventional build system
- [x] The feature has been verified to work with the CMake based build system

PR #4819, opened 2025-12-23
Add fix graphics/arrows command to add arrows for vector properties to dump image

akohlmey:
**Summary**

This pull request implements a new `fix graphics/arrows` command that can display per-atom arrows, e.g. for per-atom forces, velocities, or dipoles. Also a "wrap" keyword was added to `compute com/chunk` to optionally wrap the resulting COM position back into the simulation cell.

<hr/>

water-velocity

(per-atom velocity vectors)

<hr/>

myimage-9900

(atom_style dipole vectors)

<hr/>

water-dipoles

(per-chunk dipole moments)

<hr/>

**Related Issue(s)**

This uses functionality from PR #4817 and thus may be merged only after that pull request has been approved and merged (otherwise merging this pull request with merge both at the same time). It is further extended and parts are refactored in PR #4820. Thus any refactoring changes to the `DumpImage` class should be done on PR #4820 and not here.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

This depends on the high-level graphics object class for drawing arrows added in PR #4817 

**Post Submission Checklist**


doc/src/compute_com_chunk.rst
```diff
@@ -8,54 +8,73 @@ Syntax
...
 
 * ID, group-ID are documented in :doc:`compute <compute>` command
 * com/chunk = style name of this compute command
 * chunkID = ID of :doc:`compute chunk/atom <compute_chunk_atom>` command
+* zero or more keyword/value pairs may be appended
+* keyword = *wrap*
+
+  .. parsed-literal::
+
+       *wrap* logical = wrap center of mass back to cell if argument is *yes*, *true*, *on* or *1*, otherwise do not (default)
```
jrgissing, line 21: should we just choose one? perhaps on some more general docs page, we can state that other arguments can typically be substituted for 'yes'

doc/src/compute_com_chunk.rst
```diff
@@ -8,54 +8,73 @@ Syntax
...
-This compute calculates the :math:`(x,y,z)` coordinates of the center of mass
-for each chunk, which includes all effects due to atoms passing through
-periodic boundaries.
+This compute calculates the :math:`(x,y,z)` coordinates of the center of
+mass for each chunk, which includes all effects due to atoms passing
+through periodic boundaries.
+
+.. versionadded:: TBD
+
+If the *wrap* flag is used with an argument of *yes*, *true*, *on* or
```
jrgissing, line 54: also here could be simplified to *yes* or *no*

doc/src/compute_com_chunk.rst
akohlmey: I am not opposing this, but this should be done consistently throughout the *entire* code and it is thus outside the scope of this pull request.  This applies to all cases where `utils::logical()` is used on arguments which does not apply everywhere. There are some cases where you have logicals as arguments with additional choices, which are handled manually. I suggest submitting this as a feature request issue and - if you feel strongly about it - implement it yourself or recruit a volunteer. Otherwise just flag it as a task looking for a volunteer.

P.S.: This kind of pull request would also be an opportunity to update some legacy code to *use* `utils::logical()` and similarly update the error message handling according to the current conventions.

doc/src/compute_com_chunk.rst
akohlmey: please see my comment on the other case.

doc/src/dump_image.rst
```diff
@@ -49,7 +49,12 @@ Syntax
...
-         width = numeric value for tringle edge width (distance units)
+         width = numeric value for triangle edge width (distance units)
+       *ellipsoid* = color eflag level width
+         color = *type*
+         eflag = 1 for triangles, 2 for wireframe, 3 for both
+         level = mesh refinement level, value between 1 (low resolution) and 6 (ultra high resolution)
+         width = diameter of wireframe edges (distance units) (ignored for triangles)
        *body* = color bflag1 bflag2
          color = *type*
          bflag1,bflag2 = 2 numeric flags to affect how bodies are drawn
```
jrgissing, line 60: can a brief description be added here? something like:
bflag1 = diameter of body particle
bflag2 = 1 for triangles, 2 for wireframe, 3 for both

akohlmey: This can be different for different body styles, e.g. for nparticle the visualization is just a collection of spheres.

jrgissing: So, these descriptions are not incorrect, just sometimes the keyword is ignored, right? I still think it will help users to have the brief description here.

akohlmey: They *can* be incorrect.

jrgissing: OK. How so?

doc/src/fix_graphics_arrows.rst
```diff
@@ -0,0 +1,189 @@
...
+This fix allows to add arrows to images rendered with :doc:`dump image
+<dump_image>` using the *fix* keyword to represent vector properties
+with arrows for either all atoms in the fix group or for :doc:`chunks
+<Howto_chunk>`.
+
+The *group-ID* sets the group ID of the atoms selected to have the selected
+property represented.  This may be a dynamic group.
+
+The *Nevery* keyword determines how often the arrows graphics data is
+updated.  This should be the same value as the corresponding *N*
```
jrgissing, line 71: can the nevery keyword just be extracted from 'dump image', and removed from this fix?

akohlmey: Hmmm... tricky, since the fix needs to be defined before the dump. Not sure if you can change it after the fact. Also, there is the idea to have the fix output VMD script commands to a file (not yet implemented) so the same kind of graphics (but at higher quality) can be created in VMD. I've already implemented a command for exporting region visualizations. I'm currently not making VMD a priority since the overall future of VMD is a bit in limbo. They have not had a proper release in over 9 years now.

doc/src/Howto_body.rst
```diff
@@ -500,10 +500,11 @@ sphere is determined by the *bflag1* parameter for the *body* keyword.
 The *bflag2* argument is ignored.
 
 For the *rounded/polygon* body style, each body is drawn as a polygon
-with N line segments.  For the *rounded/polyhedron* body style, each
-face of each body is drawn as a polygon with N line segments.  The
-drawn diameter of each line segment is determined by the *bflag1*
-parameter for the *body* keyword.  The *bflag2* argument is ignored.
+and for *rounded/polyhedron* as a polyhedron.  The *bflag2* argument
```
jrgissing, line 503: I do not understand why the default value is needed. But, I suggest rephrasing:
The *bflag2* argument affects how body particles are rendered. For a value of 1, only faces are drawn. For a value of 2, only the wireframe mesh is drawn. For a value of 3, both the faces and the wireframe mesh is drawn. A value of 0 selects the default, which is the same as a value of 3.  The diameter of each wireframe line is determined by the *bflag1* parameter for the *body* keyword.

doc/src/fix_graphics_arrows.rst
```diff
@@ -0,0 +1,189 @@
...
+Most commonly this will be either :doc:`compute com/chunk
+<compute_com_chunk>` for "mobile" chunks or compute :doc:`compute
+property/chunk <compute_property_chunk>` for binning based chunks.  The
+*vec-ID* is the ID of a per-chunk :doc:`compute command <compute>`.
+Either per-chunk compute must return a global array with at least 3
+columns and *only* the first three columns are used for the arrows.  For
+computes that compute a tensor only the trace of the tensor is used.
+Currently the following computes are compatible:
+
+   * :doc:`angmom/chunk <compute_angmom_chunk>`
```
jrgissing, line 110: should the ability to visualize vector values be mentioned on each of these supported fixes' doc pages?

doc/src/fix_graphics_arrows.rst
akohlmey: I should add that in a way the fixes creating graphics data would fit into the LAMMPS logic better if they were computes. Computes are "on-demand" but there is currently no interface for them in dump image. But then again, from some computes and variables, you need to have the addstep/clearstep commands and those are just simpler to make work in a fix with a global period parameter.

doc/src/fix_graphics_arrows.rst
akohlmey: Not sure. This fix is a bit of a gimmick at the moment, so I don't want to oversell it and overload the documentation with many cross-references (although I am a stickler for adding those normally). We'll have to see which of those graphics related new features will be useful to people and then improve the documentation accordingly.

jrgissing (approved):
cool features!

PR #4822, opened 2025-12-31
Fix bond/react: wildcard atom types

jrgissing:
**Summary**

Add support for wildcard atom types in 'fix bond/react' reaction templates, and autotyping of higher-order interactions that contain wildcard atoms.

Additional description here:

J. R. Gissinger, I. Nikiforov, Y. Afshar, B. Waters, M. Choi, D. S. Karls, A. Stukowski, W. Im, H. Heinz, A. Kohlmeyer, and E. B. Tadmor, J Phys Chem B, 128, 3282-3297 (2024).
https://doi.org/10.1021/acs.jpcb.3c08419

**Related Issue(s)**

none

**Author(s)**

Jake Gissinger (Stevens)

**Licensing**

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**


jtclemm (approved):

akohlmey:
@jrgissing What do you think about adding Doxygen style comments and having some description of how to use them in contributed code in the "Developer Information" part of the Programmer's Guide in the LAMMPS manual?

Probably best done in a separate pull request, though.

jrgissing:
Sounds like a good idea. Are you referring to the type label source, fix bond/react, or both?

akohlmey:
> Sounds like a good idea. Are you referring to the type label source, fix bond/react, or both?

`src/label_map.h` and `src/label_map.cpp`.

For examples see `src/tokenizer.h`, `src/tokenizer.cpp`, `src/arg_info.h`, and `src/arg_info.cpp` and `doc/src/Developer_utils.rst`

akohlmey:
My long-term goal is to have programmer's documentation for all core classes. See also https://github.com/lammps/lammps/pull/4793

At some point this could then become the "official LAMMPS C++ API"

This would serve needs that go between the very high-level info in the `Modify` pages and the extra detailed information in the examples discussing implementing a pair style or a fix.
By using Doxygen we have the best chance of having this consistent with the source.
This step pretty much complete for the C-language library interface and the python module.

As for `fix bond/react`, when discussing this initially with @sjplimp he explicitly requested to not do this for derived classes and only document class members that people *should* use.
This is also in line with feedback I have from @stanmoore1 who feels that doxygen comments reduce readability, so we should restrict them to core functionality.

Personally, I believe there is a need for this kind of documentation in the manual instead of just pointing people to the source code since the way how people program (or get trained or train themselves to program) is changing and the ability to read and understand source code and comments (at least the style of comments that are present in large parts of LAMMPS) is fading away.

PR #4573, opened 2025-04-28
Support setting threads per atom with neigh/thread on with the KOKKOS package

ndtrung81:
**Summary**

This PR allows users to set the number of threads per atom when threading over neighbors in force calculations the from command line e.g. "-pk kokkos neigh/thread on threads/per/atom 8".

**Related Issue(s)**

N/A

**Author(s)**

Trung Nguyen (U Chicago)

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

Should maintain backward compatibility.

**Implementation Notes**

**Post Submission Checklist**

- [x] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [x] The source code follows the LAMMPS formatting guidelines
- [x] Suitable new documentation files and/or updates to the existing docs are included
- [ ] The added/updated documentation is integrated and tested with the documentation build system
- [ ] The feature has been verified to work with the conventional build system
- [x] The feature has been verified to work with the CMake based build system
- [ ] Suitable tests have been added to the unittest tree.
- [ ] A package specific README file has been included or updated
- [ ] One or more example input decks are included

**Further Information, Files, and Links**

PR #4830, opened 2026-01-08
Adding support for Kokkos kernel params autotuner

ndtrung81:
**Summary**

This PR adds a performance tuner for the GPU kernel launch parameters for the KOKKOS package. The tuner, if enabled, will scan through the possible values of the team sizes and vector sizes, measures the overall performance within a specified interval, and find the combination that gives the best performance. The optimal combination will be used for the `pair_compute_neighlist` kernel` (defined in `pair_kokkos.h`) from that point on.

This PR is built upon the features provided by #4573. Based on the findings of that PR, the most influential kernel parameters are `pair/team/size` and `threads/per/atom`, that is, teamsize and vectorsize, respectively, of the pair style kernels. That's the motivation for scanning over these 2 parameters in the /kk pair styles in this PR.  Bond styles have a single parameter to tune (`bond_block_size`).

**Related Issue(s)**

N/A

**Author(s)**

Trung Nguyen (U Chicago)

**Licensing**

**Backward Compatibility**

Should maintain backward compatiblity

**Implementation Notes**

It would be more efficient if the tuner can be refactored out of individual pair styles, to avoid the repetition for each /kk pair style.

**Post Submission Checklist**


src/KOKKOS/tune_kokkos.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+  if (firststep == 1) {
+    new_cpu = 0.0;
+    dvalue = 0.0;
+    firststep = 0;
+  } else {
+    new_cpu = timer->elapsed(Timer::TOTAL);
+    double cpu_diff = new_cpu - last_spcpu;
+    int step_diff = new_step - last_step;
+    if (step_diff > 0.0) dvalue = cpu_diff/step_diff;
```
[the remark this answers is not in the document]
ndtrung81: thanks. I also switched to using `bigint` for `new_step` and `step_diff`.

src/KOKKOS/tune_kokkos.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+  for (int vs = 64; vs <= 512; vs += 32)
+    team_sizes.push_back(vs);
+
+  for (int ts = 1; ts <= 32; ts *= 2)
+    vector_sizes.push_back(ts);
+
+  // compute the total number of parameter combinations
+  // cols = team sizes (vector sizes: 64, 96, ..., 512)
+  // rows = vector sizes (threads per atom: 1, 2, 4, 8, 16, 32)
```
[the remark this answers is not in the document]
ndtrung81: Renamed the variables to be consistent.

src/KOKKOS/pair_lj_cut_coul_long_kokkos.cpp
```diff
@@ -118,6 +122,10 @@ void PairLJCutCoulLongKokkos<DeviceType>::compute(int eflag_in, int vflag_in)
 
   // loop over neighbors of my atoms
 
+  copymode = 1;
+
+  if (lmp->kokkos->autotuning) tuner->tuning_kernel_params();
```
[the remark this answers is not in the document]
ndtrung81: `tuner` is allocated only when `lmp->kokkos->autotuning` is non-zero in `init_style()`. Extra check for a valid `tuner` is added.

doc/src/package.rst
```diff
@@ -665,6 +667,15 @@ The *bond/block/size* keyword sets the number of GPU threads per block
 used for launching the bond force kernel on the GPU.  The default value
 of this parameter is determined based on the GPU architecture at runtime.
 
+The *auto/tuning* keyword enables the auto-tuning feature of
+the KOKKOS package.  When enabled, the KOKKOS pair style in use will scan
+through the possible values of the *pair/team/size* and *threads/per/atom*
+parameters and choose the combination that gives the best performance during the first
+stage of the run.  The *nevery* keyword sets the interval used to estimate
+the performance of a combination of these two parameters.  This feature is
+disabled by default.
```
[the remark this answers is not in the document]
ndtrung81: Done. This is a good point referring to the LAMMPS documentation guidelines.

src/KOKKOS/pair_lj_cut_kokkos.cpp
```diff
@@ -102,6 +106,8 @@ void PairLJCutKokkos<DeviceType>::compute(int eflag_in, int vflag_in)
 
   copymode = 1;
 
+  if (lmp->kokkos->autotuning) tuner->tuning_kernel_params();
```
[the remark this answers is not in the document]
ndtrung81: Fixed.

ndtrung81:
Preliminary results from the autotuner:
* `in.lj`: 1.5X speedup vs. using default values:
[perf-lj.pdf](https://github.com/user-attachments/files/24574532/perf-lj.pdf)

* `in.rhodo`: 1.3X speedup vs. using default values:
[perf-rhodo.pdf](https://github.com/user-attachments/files/24574534/perf-rhodo.pdf)

src/KOKKOS/pair_dpd_kokkos.cpp
```diff
@@ -310,6 +338,116 @@ void PairDPDKokkos<DeviceType>::operator() (TagDPDKokkos<NEIGHFLAG,EVFLAG>, cons
...
+  jnum = d_numneigh[i];
+  rand_type rand_gen = rand_pool.get_state();
+
+  KK_FLOAT fxtmp = 0.0;
+  KK_FLOAT fytmp = 0.0;
+  KK_FLOAT fztmp = 0.0;
+
+  Kokkos::parallel_reduce(Kokkos::TeamThreadRange(team, jnum),
+    [&] (const int &jj, KK_FLOAT &fx, KK_FLOAT &fy, KK_FLOAT &fz) {
+
```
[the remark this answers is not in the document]
ndtrung81: This is a very good point. In commit 64dbab4 I replace `rand_gen` with SARU (from `lib/gpu/lal_dpd.cu`) which is state-less for the random number generator.

akohlmey:
@ndtrung81 Can you please give us an update on the status of this pull request and whether you expect it to be ready for merging within the next couple of weeks?

ndtrung81:
@akohlmey the PR is ready for merging within the next couple of weeks, pending the reviewers' feedback. My goal for this PR is to create a framework for tuning the kernel team policies on the fly, and add the tuner to certain pair styles that benefit the most from tuning.

ndtrung81:
@stanmoore1 @sjplimp could you please advise if there is anything you expect to be addressed in this PR?

doc/src/package.rst
```diff
@@ -133,10 +133,15 @@ Syntax
...
-        *nbor/block/size = blocksize
-          blocksize = # of GPU threads per block for the flat neighbor build method
-        *bond/block/size = blocksize
-          blocksize = # of GPU threads per block for the bond force computation
+        *nbor/chunk/size = chunksize
+          chunksize = # of iterations each thread will perform for the flat neighbor build method
+        *bond/chunk/size = blocksize
+          chunksize = # of iterations each thread will perform for the bond force computation
+        *auto/tuning = nevery nsamples mode reltol
+          nevery = # timesteps between autotuning adjustments (default = 0, no autotuning)
```
sjplimp: 0 = no autotuing
Normally the defaults should be listed at the bottom of the doc page. I.e. 0 in this case would mean no autotuning is done

If 0 is used for nevery, do the other 3 params need to be specified?  Or are they just ignored?  This should be clarified in the autotuning section below

doc/src/package.rst
```diff
@@ -133,10 +133,15 @@ Syntax
...
+          nsamples = # samples the tuner(s) collects per parameter set (default = 5)
```
sjplimp: Since this has to be set (if nevery > 0), it seems like there is no default ?

doc/src/package.rst
```diff
@@ -133,10 +133,15 @@ Syntax
...
+          mode = how to pick a representative value from the samples to a parameter set, i.e. maximum, average or median value
```
sjplimp: seems like awkward grammar, not sure what "to a parameter set" means

doc/src/package.rst
```diff
@@ -133,10 +133,15 @@ Syntax
...
+          reltol = relative tolerance for performance degradation that triggers a re-scan of parameter space (default = 0.2)
```
sjplimp: maybe: re-tuning of parameter values

doc/src/package.rst
```diff
@@ -666,15 +671,54 @@ The *nbin/atoms/per/bin* keyword sets the number of atoms per bin
...
-the GPU threads are assigned to the bins.
+The *nbor/chunk/size* keyword sets the number of iterations that a work item
+is scheduled for the neighbor list builds on the GPU using the flat method (i.e.,
+each thread finds the neighbor list of an atom).  If not specified, the default value
+of this parameter is determined based on the GPU architecture at runtime.
 
-The *bond/block/size* keyword sets the number of GPU threads per block
-used for launching the bond force kernel on the GPU.  The default value
+The *bond/chunk/size* keyword sets the number of iterations that a work item
+is scheduled used for launching the bond force kernel on the GPU.  The default value
```
sjplimp: grammar:  is scheduled used ?

doc/src/package.rst
```diff
@@ -666,15 +671,54 @@ The *nbin/atoms/per/bin* keyword sets the number of atoms per bin
...
 of this parameter is determined based on the GPU architecture at runtime.
 
+.. versionadded:: TBD
+
+The *auto/tuning* keyword enables the auto-tuning feature of
+the KOKKOS package.  When enabled, the KOKKOS styles in use will scan
```
sjplimp: Is tuning only provided for pair and bond styles?  If so, that should be stated directly.

doc/src/package.rst
```diff
@@ -666,15 +671,54 @@ The *nbin/atoms/per/bin* keyword sets the number of atoms per bin
...
+through the possible values of the kernel launch parameters (that is, execution policies)
```
sjplimp: remove "the"

doc/src/package.rst
```diff
@@ -666,15 +671,54 @@ The *nbin/atoms/per/bin* keyword sets the number of atoms per bin
...
+such as *pair/team/size* and *threads/per/atom* for pair styles and *bond/chunk/size*
```
sjplimp: replace "such as" with "which are", assuming that is all the tuning that is attempted

doc/src/package.rst
```diff
@@ -666,15 +671,54 @@ The *nbin/atoms/per/bin* keyword sets the number of atoms per bin
...
+for bond styles, and find the combination that gives the best simulation
```
sjplimp: to find the value (bond style) or values (pair style) that give the best ...

doc/src/package.rst
```diff
@@ -666,15 +671,54 @@ The *nbin/atoms/per/bin* keyword sets the number of atoms per bin
...
+performance in terms of the number of timesteps per second.
+The optimal combination of the kernel launch parameters are then fixed
```
sjplimp: The optimal kennel launch parameters ...

doc/src/package.rst
```diff
@@ -666,15 +671,54 @@ The *nbin/atoms/per/bin* keyword sets the number of atoms per bin
...
+for the remaining part of the run, unless the performance drops below
+a certain relative tolerance compared to the found optimal performance.
```
sjplimp: drop "found"

doc/src/package.rst
```diff
@@ -666,15 +671,54 @@ The *nbin/atoms/per/bin* keyword sets the number of atoms per bin
...
+
+The following parameters are needed for the auto-tuning process.
+
+*nevery* controls the interval used to estimate
```
sjplimp: *Nevery* sets the timestep interval for calculating performance of a single sample with specific kernel launch values. It needs to be large enough... with a sufficiently large number of kernel calls.  It also needs to be small enough to quickly scan over all possible values.  A value of 100 is usually reasonable.

doc/src/package.rst
```diff
@@ -666,15 +671,54 @@ The *nbin/atoms/per/bin* keyword sets the number of atoms per bin
...
+the overall performance for a combination of these two parameters.
+*nevery* needs to be large enough to have a stable estimate
+of the performance, to achieve a sufficiently large number of kernel calls,
+while small enough to reduce the time required for scanning
+over all the combinations. *nevery* = 100 is usually a reasonable value.
+
+*nsamples* indicates the number of samples the tuners will
```
sjplimp: *Nsamples" sets ...

doc/src/package.rst
```diff
@@ -666,15 +671,54 @@ The *nbin/atoms/per/bin* keyword sets the number of atoms per bin
...
+collect for each parameter set.  *nsamples* = 5 is usually a reasonable value.
+
+*mode* controls how the representative performance of a paramter set
```
sjplimp: *Mode" determines how the performance of a collection of samples is determined:

doc/src/package.rst
```diff
@@ -666,15 +671,54 @@ The *nbin/atoms/per/bin* keyword sets the number of atoms per bin
...
+is selected from the collected samples: *max* means the maximum value,
+*ave* means the arithmetic average value, and *median* means the median
+value.
+
+*reltol* sets the relative tolerance for performance degradation
+compared to the last optimal performance, which will trigger a re-scan
```
sjplimp: which may trigger

doc/src/package.rst
```diff
@@ -666,15 +671,54 @@ The *nbin/atoms/per/bin* keyword sets the number of atoms per bin
...
+of the parameter space.  A value of 0.2 means that if the current performance
+is more than 20% below the last optimal performance for *nsamples* times,
+a re-scan will be triggered.  Setting *reltol* to be equal or greater than 1.0
+will disable the re-scanning.
```
sjplimp: For *reltol* I don't understand if this means the performance is being continually monitored over a long run?
E.g. if nevery = 100 and nsamples = 5, does this mean after the first parameter scan, that perf continues to be tallied every timestep for all kernels, and the overall perf calculated every 500 steps, to test if reltol is exceeded?  I.e. the every-step tallying is never disabled, even for a long run?

doc/src/package.rst
```diff
@@ -666,15 +671,54 @@ The *nbin/atoms/per/bin* keyword sets the number of atoms per bin
...
+each thread finds the neighbor list of an atom).  If not specified, the default value
+of this parameter is determined based on the GPU architecture at runtime.
 
-The *bond/block/size* keyword sets the number of GPU threads per block
-used for launching the bond force kernel on the GPU.  The default value
+The *bond/chunk/size* keyword sets the number of iterations that a work item
+is scheduled used for launching the bond force kernel on the GPU.  The default value
 of this parameter is determined based on the GPU architecture at runtime.
 
+.. versionadded:: TBD
```
sjplimp, line 628: In this section, my comments are aimed at clarifying for users how this new feature works.

doc/src/package.rst
```diff
@@ -666,15 +671,54 @@ The *nbin/atoms/per/bin* keyword sets the number of atoms per bin
...
+value.
+
+*reltol* sets the relative tolerance for performance degradation
+compared to the last optimal performance, which will trigger a re-scan
+of the parameter space.  A value of 0.2 means that if the current performance
+is more than 20% below the last optimal performance for *nsamples* times,
+a re-scan will be triggered.  Setting *reltol* to be equal or greater than 1.0
+will disable the re-scanning.
+
+The *auto/tuning* feature is supported by a limited number of the KOKKOS styles
```
sjplimp: I think this sentence should be above at the start of the autotuning section.  And be more specific about which Kokkos styles support it.  E.g. is it all pair and bond styles?  Also, I think a couple sentences about possible performance gains should be added to motivate users to try it.  E.g. tell them what kinds of problems could achieve what percentage perf gain.

sjplimp:
@ndtrung81 I really like this new feature - thanks for this PR.  Made several comments on the doc info for the autotuning option.  Am ready to approve for merging once you respond.

It would also be very good to add a couple of your perf plots for autotuning to the Benchmark web page.

ndtrung81:
@sjplimp thank you for the feedback! I have updated the doc page with your suggestion to clarify the case when nevery is 0 and mention the defaults at the bottom of the page.

Attached is the plot showing the performance with and without auto/tuning for `in.lj` and `in.rhodo` with `nevery` = 100, `nsamples` = 5, `mode` = median with 1 MPI on 1 A100 PCIe GPU, the KOKKOS package was built with single-precision.

Do you think they are suitable for the Benchmark page?

perf-autotuning

sjplimp:
@ndtrung81 I think these should be added to the package command doc page, as well as the Benchmark page.  Talk to @stanmoore1 about the Benchmark page.

sjplimp:
@stanmoore1 I also think the Benchmark page could be updated to report Benchmark timings that reflect the increased autotuning timesteps rates, where they make a difference,

sjplimp:
@ndtrung81 Have my other comments on the doc page been addressed?  If so, then I will approve this PR for merging,

doc/src/package.rst
ndtrung81: @sjplimp I have addressed these comments in commit [040ba01](https://github.com/lammps/lammps/pull/4830/commits/040ba017edd42ee0c5cb02f9ba78af8b8632603b): put the default (no autotuning at the bottom of the page) and clarify that when nevery is 0, other 3 params need to be specified but have no effect because no autotuning is done.

sjplimp (changes requested):

sjplimp (approved):

sjplimp:
@akohlmey per email, you can merge this when Trung completes his final (small) edits

doc/src/package.rst
ndtrung81: this is a mistake, I removed "used",

doc/src/package.rst
ndtrung81: yes, I used "re-tuning of parameter values".

doc/src/package.rst
ndtrung81: right, I removed (default = 5).

doc/src/package.rst
ndtrung81: I renamed "parameter set" to "parameter combination".

doc/src/package.rst
ndtrung81: I list all the styles that support auto-tuning currently and suggest how to support the feature for others.

doc/src/package.rst
ndtrung81: thanks, I made this change.

doc/src/package.rst
ndtrung81: [checked]

doc/src/package.rst
ndtrung81: I have rewritten the text to clarify these parameters with an example. After the first scan, the tuner will continue to monitor the performance every nevery = 100 steps. If the perf drops below 20% pf the last stored best perf by nsamples = 5 times, then the tuner will re-scan.

doc/src/package.rst
ndtrung81: thanks, I added as much detail as possible to explain how this feature works. If it needs further improvement, please advise.

doc/src/package.rst
ndtrung81: I moved this sentence toward to beginning, including the possible performance gains and the plots for in.lj and in.rhodo. Also included are the list of the currently supported pair and bond styles, and some suggestions to enable this feature in other styles, or refactor the tuners to avoid duplicated everywhere. Please let me know if the changes make more sense.

ndtrung81:
@akohlmey I have tried to address Steve's comments and edits. The doc page is good from my point of view. I will assign it to you, unless you need me to make more changes. Thanks!

akohlmey:
> I will assign it to you, unless you need me to make more changes. Thanks!

@ndtrung81 I made a few cosmetic changes to package.rst and am now running this one more through code review. A few minor findings:

1. get_current_vector_size() indexes team_sizes instead of vector_sizes — live for the GENERIC dpd/dipole integrations that call it, yielding an invalid vector length.
2. utils::logmesg in kokkos.cpp has 3 {} placeholders but 2 arguments (and the first should be the interval, not perf_nsamples) — throws when autotuning is enabled.
3. `if (tuner) delete tuner;` is redundant. You can just do `delete tuner;`
4. package.rst default-settings line still lists auto/tuning = disabled.
5. Double blank line around the #else // LMP_KOKKOS_GPU not defined block.
6. Constructor accepts interval == 0 / nsamples == 0 (< 0 checks); not reachable as a live div-by-zero since a tuner is only created when autotuning > 0, but <= 0 would be clearer defensively.
7. Side-effecting getter: get_optimal_combination_idx() mutates opt_perf and, in MEDIAN mode, std::sorts the performance rows in place — yet it's called from regular_performance_check() purely for logging. Make it const/non-mutating, or compute the median on a copy.
8. Repetition inside the tuner: allocate(), set_team_size_values(), set_vector_size_values() each duplicate the ncombinations/memory->create block.

Please check if these are valid concerns.

ndtrung81:
@akohlmey thanks for the code review. These are valid concerns.

1. This is a bug, I need to index vector_sizes for get_current_vector_size() from the current combination index.
2. Missing autotuning for the first placeholder.
3. Will simplify this.
4. This is intentional because auto/tuning is by default disabled. What is your suggestion?
5. Will remove the double blank lines.
6. Will check of <= 0 for interval and nsamples.
7. Will make this change.
8. The duplicated memory->create() blocks (followed by destroy()) in  allocate(), set_team_size_values(), set_vector_size_values() are because the two set functions may be called after allocate() is already called in the tuner's constructor.  A Pair class may want to override the scanned values of team sizes or vector sizes.

akohlmey:
> @akohlmey thanks for the code review. These are valid concerns.

[...]

> 4. This is intentional because auto/tuning is by default disabled. What is your suggestion?

This was the result of the AI review. I was too tired to double check everything.
I think that disabling auto-tuning by default is the right call. We had a similar
thing for KSPACE (benchmarking the FFT time percentage and auto-tuning it) that
is now both disabled by default.

ndtrung81:
@akohlmey I am done with addressing the concerns from code review.

akohlmey:
> @akohlmey I am done with addressing the concerns from code review.

Then let me quickly merge it before somebody get any more ideas for changes...

PR #4832, opened 2026-01-10
delete_atoms condense yes for molecular systems

alphataubio:
**Summary**

This pull request extends the `delete_atoms ... condense yes` functionality to work with **molecular systems** (e.g., `atom_style bond`, `angle`, `molecular`, `full`), whereas it was previously restricted to `atom_style atomic`.

To achieve this robustly for complex topologies, this PR:
1.  Implements a scalable, order-preserving **Parallel Sample Sort** algorithm to reassign atom IDs contiguously while preserving global ordering.
2.  Refactors `DeleteAtoms::condense_tags` to correctly update molecular topology (bonds, angles, dihedrals, impropers) and internal fix data (e.g., CMAP) when IDs change.
3. Delegates `update_ids()` to fixes with `stores_ids=1`:
    - [x] `fix cmap`
    - [x] `fix shake`
    - [x] `fix rigid/small`

**Related Issue(s)**

**Author(s)**

@alphataubio

**Licensing**

Portions of the code refactoring for `DeleteAtoms::condense_tags` (specifically the implementation of the Parallel Sample Sort algorithm) and the debugging of `Comm/Neighbor` list updates were developed with the assistance of Google Gemini (Large Language Model). The AI tools were used to generate the MPI communication patterns for the sample sort and to identify the root cause of segmentation faults related to stale neighbor bins. All generated code has been manually reviewed, tested, and integrated by the author. 

PR description was also generated by gemini pro 3 based on the chat contents working on this PR.

Gemini Pro 3 generated `DeleteAtomsTest.CondenseMolecular` and `DeleteAtomsTest.CondenseFixes` unit tests.

**Backward Compatibility**

This PR modifies the `Fix` base class by adding a new virtual method `virtual void update_ids(double **newIDs)`. This is an ABI change; external plugins or user-customized fixes inheriting from `Fix` may need recompilation, though source compatibility is preserved (the default implementation is empty).

**Implementation Notes**

1.  **Algorithmic Overhaul (Ring vs. Sample Sort):**
    * **Old Implementation (Atom::ATOMIC only):** The previous code relied on a `comm->ring` collective. This approach rotated data tokens through *every* processor in the simulation ring ($O(P)$ complexity steps). For large processor counts, the latency of passing tokens sequentially around the ring makes this algorithm extremely inefficient and unscalable. Furthermore, it required allocating local buffers proportional to the global tag range divided by processor count, which can hit memory limits for sparse but high-ID systems.
    * **New Implementation (Universal):** This PR replaces the ring heuristic with a **Parallel Sample Sort**.
        * **Scalability:** The new algorithm is $O(N/P \log(N/P))$ and uses optimized `MPI_Alltoallv` patterns instead of point-to-point ring shifting. It scales gracefully to large processor counts.
        * **Memory Efficiency:** It processes only active atoms, avoiding the allocation of buffers based on the potentially massive global ID range.
        * **Correctness:** It guarantees strict global order preservation ($OldID_A < OldID_B \implies NewID_A < NewID_B$). This is critical for molecular systems where users often rely on ID clustering in dump files (e.g., keeping atoms of a specific molecule sequential).

2.  **Enablement for Molecular Systems:**
    * The restriction preventing `condense yes` for molecular systems has been removed.
    * The topology update logic now iterates through bonds, angles, dihedrals, and impropers, mapping their constituent atom IDs to the new contiguous set using a `double**` buffer pattern (compatible with `forward_comm_array` and bit-exact `ubuf` casting).

5.  **Fixing Segmentation Faults:**
    * Compacting the atom array changes `nlocal`, which previously left `Comm` and `Neighbor` lists pointing to invalid memory indices/holes.
    * Added explicit `neighbor->setup_bins()` and `comm->setup()` calls *before* any communication occurs in `condense_tags` to ensure ghost lists match the new memory layout, resolving segmentation faults and bus errors.

6.  **Fix Interface Extension (CMAP Support):**
    * Added `virtual void update_ids(double **newIDs)` to `src/fix.h`.
    * Implemented this method in `FixCMAP` and `FixCMAPKokkos` to update their internal 5-body interaction terms when IDs change. This resolves "missing atom" errors during deletions in CHARMM-based simulations.

**Post Submission Checklist**


sjplimp:
@alphataubio Thanks for pointing out the scaling inefficiencies in delete_atoms due to "ring" comm.  And for submitting this PR to address that and add other extensions to the delete_atoms "condense" option.

The LAMMPS way of avoiding ring comm in other places where it was once used (inefficiently), has been to use "rendezvous" communication, which can use either MPI_All2all or what we term *irregular* comm internally.  In the case of "condense" I think it should be possible to use rendezvous plus a local sort, instead of the global parallel sort you have implemented here.

But I need to think about this a bit, and check if rendezvous is also compatible with the other changes you are proposing.

PR #4841, opened 2026-01-16
Added option to use final_integrate in brownian time step.

samueljmcameron:
**Summary**

Currently, the BROWNIAN package time-integrators all call initial_integrate. This PR adds the option to have them call final_integrate. This option is important if e.g. atoms are created in a different fix's post_integrate function.

**Author(s)**

Sam Cameron - The Open University
samuel.j.m.cameron@gmail.com

**Licensing**

**Backward Compatibility**

It is backwards compatible.

**Implementation Notes**

Changes were verified through short runs testing a two particle system.

**Post Submission Checklist**

This PR adds a change which is necessary for a different package that I will be submitting shortly, where atoms can die or divide (mimicking biological population dynamics).

### DIFF
--- a/doc/src/fix_brownian.rst
+++ b/doc/src/fix_brownian.rst
@@ -23,7 +23,7 @@ Syntax
 * temp = temperature
 * seed = random number generator seed
 * one or more keyword/value pairs may be appended
-* keyword = *rng* or *dipole* or *gamma_r_eigen* or *gamma_t_eigen* or *gamma_r* or *gamma_t* or *rotation_temp* or *planar_rotation*
+* keyword = *rng* or *dipole* or *gamma_r_eigen* or *gamma_t_eigen* or *gamma_r* or *gamma_t* or *rotation_temp* or *planar_rotation* or *final_integrate*
 
   .. parsed-literal::
 
@@ -44,6 +44,7 @@ Syntax
      *rotation_temp* values = *T* for *brownian/sphere* and *brownian/asphere*
         *T* = rotation temperature, which can be different then *temp* when out of equilibrium
      *planar_rotation* values = none (constrains rotational diffusion to be in xy plane if in 3D)
+     *final_integrate* values = none (executes the time integration at the end of the step)
 
 Examples
 """"""""
@@ -51,7 +52,7 @@ Examples
 .. code-block:: LAMMPS
 
    fix 1 all brownian 1.0 12908410 gamma_t 1.0
-   fix 1 all brownian 1.0 12908410 gamma_t 3.0 rng gaussian
+   fix 1 all brownian 1.0 12908410 gamma_t 3.0 rng gaussian final_integrate
    fix 1 all brownian/sphere 1.0 1294019 gamma_t 3.0 gamma_r 1.0
    fix 1 all brownian/sphere 1.0 19581092 gamma_t 1.0 gamma_r 0.3  rng none
    fix 1 all brownian/asphere 1.0 1294019 gamma_t_eigen 1.0 2.0 3.0 gamma_r_eigen 4.0 7.0 8.0 rng gaussian
@@ -200,6 +201,10 @@ compatible with *brownian/sphere* and *brownian/asphere*.
 If the *planar_rotation* keyword is used, then rotation is constrained
 to the *x*\ -- *y* plane in a 3D simulation. Only compatible with
 *brownian/sphere* and *brownian/asphere* in 3D.
+
+If the *final_integrate* keyword is used, then the time integration
+is done at the end of the step. Otherwise, it is done at the beginning
+of the step (using previous step's forces).
 
 ----------
 
--- a/src/BROWNIAN/fix_brownian.cpp
+++ b/src/BROWNIAN/fix_brownian.cpp
@@ -49,32 +49,33 @@ void FixBrownian::init()
   g2 *= sqrt(temp / gamma_t);
 }
 
-/* ---------------------------------------------------------------------- */
 
-void FixBrownian::initial_integrate(int /*vflag */)
+
+void FixBrownian::call_integrate()
 {
   if (domain->dimension == 2) {
     if (!noise_flag) {
-      initial_integrate_templated<0, 0, 1>();
+      integrate_templated<0, 0, 1>();
     } else if (gaussian_noise_flag) {
-      initial_integrate_templated<0, 1, 1>();
+      integrate_templated<0, 1, 1>();
     } else {
-      initial_integrate_templated<1, 0, 1>();
+      integrate_templated<1, 0, 1>();
     }
   } else {
     if (!noise_flag) {
-      initial_integrate_templated<0, 0, 0>();
+      integrate_templated<0, 0, 0>();
     } else if (gaussian_noise_flag) {
-      initial_integrate_templated<0, 1, 0>();
+      integrate_templated<0, 1, 0>();
     } else {
-      initial_integrate_templated<1, 0, 0>();
+      integrate_templated<1, 0, 0>();
     }
   }
-}
+} 
+
 
 /* ---------------------------------------------------------------------- */
 
-template <int Tp_UNIFORM, int Tp_GAUSS, int Tp_2D> void FixBrownian::initial_integrate_templated()
+template <int Tp_UNIFORM, int Tp_GAUSS, int Tp_2D> void FixBrownian::integrate_templated()
 {
   double **x = atom->x;
   double **v = atom->v;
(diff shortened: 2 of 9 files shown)


akohlmey:
> This option is important if e.g. atoms are created in a different fix's post_integrate function.

Not sure that this is a good idea. As far as I recall, such changes should be done in `Fix::pre_exchange()` or earlier. This applies to 11 of the 12 fixes in the MC package, for example.

samueljmcameron:
> > This option is important if e.g. atoms are created in a different fix's post_integrate function.
> 
> Not sure that this is a good idea. As far as I recall, such changes should be done in `Fix::pre_exchange()` or earlier. This applies to 11 of the 12 fixes in the MC package, for example.

Okay that is good to know. I suppose this isn't the right place to discuss further why it should be pre_exchange, but I am curious...

akohlmey:
> Okay that is good to know. I suppose this isn't the right place to discuss further why it should be pre_exchange, but I am curious...

I've been thinking about this some more and now believe that everything else is a mistake.
When you make *any* changes to the system composition it *must* be done *before* the neighbor list build and it must be enforced that the neighbor list is updated. Not only would the neighbor list be invalid after a change to the system, the same applies to the status of ghost atoms. Individual properties can be updated with an explicit forward communication but not the system composition, i.e. removing or adding particles.

Another problem is that making changes to the geometry *after* the force computation will make positions and forces inconsistent and prevent correct restarting (forces are not stored but reconstructed from positions during setup, and thus a restart would start with different forces than when continuing a run. But these inconsistency would also leak into dump output and more. All properties that are computed with data computed or accumulated during the force computation would become inconsistent and output to dump files is *after* end_of_step().

The logical consequence is that the changes in this pull request are promoting incorrect applications that would lead to inconsistencies and thus they should not be made.

samueljmcameron:
Thanks for the explanation! I have a very specific use case in mind (which I was thinking about contributing to LAMMPS if it was of interest) which _does_ add atoms in the post_integrate() step and seems to work okay. That isn't to say it is okay more generally I know.

For context, the current PR is anticipating the new package I'm talking about above (let's call it POPULATION), where over-damped particles can divide and die. Perhaps the package isn't suitable for wide-spread sharing in LAMMPS though, as it can require rebuilding neighbour lists quite frequently depending on division/death rates. But the key point is that for POPULATION, no time integration can be performed until the new particles are created/destroyed.

> I've been thinking about this some more and now believe that everything else is a mistake. When you make _any_ changes to the system composition it _must_ be done _before_ the neighbor list build and it must be enforced that the neighbor list is updated. Not only would the neighbor list be invalid after a change to the system, the same applies to the status of ghost atoms. Individual properties can be updated with an explicit forward communication but not the system composition, i.e. removing or adding particles.

For POPULATION, the idea is that at each step, any particle can divide (i.e. a second particle is created and placed on top/near the parent particle). I've implemented this in the post_integrate step which does (for verlet.cpp) get called prior to neighbor->decide() from what I can see? If any particle dies or divides, it triggers a neighbour list update. If no particle divides/dies, then it doesn't trigger a neighbour list update. I suppose I can just use post_integrate to trigger that a neighbour list should be built, and then actually create new atoms in the pre_exchange as you're suggesting. Regardless, it does come back to the initial issue which I've said at the end of my previous paragraph (restating here to be clear): no time integration can be performed until the new particles are created/destroyed.

> Another problem is that making changes to the geometry _after_ the force computation will make positions and forces inconsistent and prevent correct restarting (forces are not stored but reconstructed from positions during setup, and thus a restart would start with different forces than when continuing a run. But these inconsistency would also leak into dump output and more. All properties that are computed with data computed or accumulated during the force computation would become inconsistent and output to dump files is _after_ end_of_step().

Yes I agree that this is true and I've noticed this when running my simulations. I hadn't thought carefully about a work-around (if one exists) since it didn't affect my work and honestly forgot about the issue until you've just raised it now. 

> The logical consequence is that the changes in this pull request are promoting incorrect applications that would lead to inconsistencies and thus they should not be made.

I think you are right about this if there is no real value in including the POPULATION package as I've described above into LAMMPS official repo. If you think there might be some value in something like the POPULATION package (which I've already developed and used parts of in a peer-reviewed article), then maybe it's worth me trying to think of a work-around for the issues you've raised. I hope I've understood your comments correctly, apologies if not.

akohlmey:
> Regardless, it does come back to the initial issue which I've said at the end of my previous paragraph (restating here to be clear): no time integration can be performed until the new particles are created/destroyed.

I don't think that is a problem even if the time integration only happens in initial_integrate(). While it won't happen on the current step, it *will* happen on the next step and so on.
What would be the harm in that?

jtclemm:
> For POPULATION, the idea is that at each step, any particle can divide (i.e. a second particle is created and placed on top/near the parent particle).

That sounds like a pretty cool application. I'm always a fan of seeing different non-MD simulations being implemented in LAMMPS. I for one think that would make a neat video example.

> I've implemented this in the post_integrate step which does (for verlet.cpp) get called prior to neighbor->decide() from what I can see? If any particle dies or divides, it triggers a neighbour list update. If no particle divides/dies, then it doesn't trigger a neighbour list update. I suppose I can just use post_integrate to trigger that a neighbour list should be built, and then actually create new atoms in the pre_exchange as you're suggesting.

If post_integrate forces a neighbor list rebuild upon creating an atom, then I agree it shouldn't make a practical difference where atoms are created/destroyed. But keeping the atom creation/deletion portion of the code in pre_exchange is more explicit. Hypothetically, if there were a bug, it'd be a lot easier to detect that atoms aren't being created (pre_exchange not called) vs. newly divided atoms having no interactions for a few timesteps (post integrate doesn't correctly trigger an nlist build). 

> Regardless, it does come back to the initial issue which I've said at the end of my previous paragraph (restating here to be clear): no time integration can be performed until the new particles are created/destroyed.

I'm trying to understand the distinction between initial/final integrate and I might be missing something. Am I correct in thinking fix brownian does not follow the typical velocity-verlet/leapfrog integration scheme but rather is more like Eulerian integration with no half step?

If so, then shouldn't the distinction between initial/final integration be mathematically artificial? Seems like this setup:

- Initial integrate: do nothing
- Post integrate/pre-exchange: create new atoms using data from timestep n
- Compute forces at timestep n , F_n
- Final integrate: calculate velocities/positions at time n+1 using F_n

is equivalent to this setup:

- Initial integrate: calculate velocity/positions at timestep n, using F_{n-1}
- Post integrate/pre-exchange: create new atoms using data from timestep n
- Compute forces at timestep n , F_n
- Final integrate: do nothing

outside of how you choose to label each quantity with a +/- 1 shift.

akohlmey:
> If so, then shouldn't the distinction between initial/final integration be mathematically artificial? Seems like this setup:
> 
> * Initial integrate: do nothing
> * Post integrate/pre-exchange: create new atoms using data from timestep n
> * Compute forces at timestep n , F_n
> * Final integrate: calculate velocities/positions at time n+1 using F_n

I think that would be worse than the current setup since the forces and positions would be inconsistent.

As indicated in my other comment. The time integration would just happen in the next timestep. That is also compatible with restarting where forces are reconstructed from positions. The initial_integrate() step is where the position update from time integration is expected to happen.

jtclemm:
> I think that would be worse than the current setup since the forces and positions would be inconsistent.

I'm not saying it's not different computationally, I'm just saying mathematically it's arbitrary. I'm just trying to clarify what could be a potential mix up from shoe horning a 1-step integration fix into a 2-step integrator class.

In which case the best course of action is to follow the LAMMPS dogma and keep position updates in initial integrate.

akohlmey:
> I'm just saying mathematically it's arbitrary.

Apologies, I misunderstood what you were trying to say.

athomps:
@samueljmcameron The consensus seems to be that it should be possible to make your POPULATION package work with the existing `initial_integrate` mode of `fix brownian`.  Also, it is probably a good idea to create/delete particles in `pre_exchange` rather than `post_integrate`.  You can find more information here: https://docs.lammps.org/Developer_flow.html

athomps (commented):
Based on previous reviewer comments, it sounds like this option should *not* be added.

samueljmcameron:
Hi everyone! Thanks for the feedback and very sorry for the mistake on my end. When I initially wrote the POPULATION package, I thought I had a good reason for needing the final_integrate option, but now that I'm reading all of your comments I realise I was clearly wrong. I'm not sure what was going through my head (developed it around this time last year so can't recall any good reasons)... So yes please close this PR it is not useful as far as I can tell. I will aim to submit a PR for the POPULATION package later in the week and prior to that do some testing to be sure that final_integrate isn't required.

akohlmey:
closed as requested

PR #4842, opened 2026-01-17
BROWNIAN: Geometric rotational integrator for fix brownian/sphere + rotation_style keyword

ArtStrau:
**Summary**

 - Adds a geometric integrator for overdamped rotational Brownian motion (RBM) of axisymmetric particles in `fix brownian/sphere` (2D, planar rotation, and 3D)
 - The geometric scheme preserves the normalization constraint by construction, greatly reducing the numerical error
 - It permits for up to a factor of ten larger timesteps, compared to the existing integration scheme based on projection 
 - For a validation within LAMMPS (comparisons LAMMPS numerics vs analytical theory), see: https://github.com/ArtStrau/lammps-rbm
 - An accompanying peer-reviewed publication (theory and geometric integration scheme):  
  F. Höfling & A. V. Straube, Phys. Rev. Research 7, 043034 (2025). DOI: [10.1103/wzdn-29p4](https://doi.org/10.1103/wzdn-29p4)

**Related Issue(s)**
none

**Author(s)**

Arthur Straube (Zuse Institute Berlin), long-lived email: arthur.straube@gmail.com

**Licensing**

**Backward Compatibility**

- Introduced keyword: `rotation_style {geometric|projection}`
- Default is `rotation_style geometric`, original behavior is available via `rotation_style projection`

Usage (new keyword):
-  fix ... brownian/sphere                                              (default, geometric)
-  fix ... brownian/sphere ... rotation_style geometric   (default, geometric)
-  fix ... brownian/sphere ... rotation_style projection  (original, projection)

**Implementation Notes**

- The new keyword `rotation_style {geometric|projection}` is parsed in `FixBrownianBase` and stored in the new member `rot_style` (enum `RotStyle {ROT_PROJECTION, ROT_GEOMETRIC}`). The default is `ROT_GEOMETRIC`
- In `FixBrownianSphere`, `rot_style` is mapped to a new template parameter `Tp_ROTGEOM` in the existing templated integration kernel: `Tp_ROTGEOM=1` selects the geometric integrator and `Tp_ROTGEOM=0` selects the original projection integrator
- This **PR affects** the **rotational** update in `fix brownian/sphere` only; the **translational** Brownian update is **unchanged**
- Validation is performed in an external repo by comparing probability density functions (PDFs) from LAMMPS against analytic PDFs (exact/projection/geometric) for torque-free overdamped RBM in 2D and 3D, as follows from the peer-reviewed publication, see below
  

**Post Submission Checklist**

- LAMMPS implementation: Arthur Straube @ArtStrau, method paper coauthor: Felix Höfling @fhoefling
- Validation (comparisons probability density functions from LAMMPS vs analytical theory for torque-free overdamped rotational Brownian motion in 2D and 3D): https://github.com/ArtStrau/lammps-rbm
- Peer-reviewed publication (theory and geometric integration scheme):  
  F. Höfling & A. V. Straube, Phys. Rev. Research 7, 043034 (2025). DOI: [10.1103/wzdn-29p4](https://doi.org/10.1103/wzdn-29p4)


akohlmey:
@ArtStrau @samueljmcameron There are going to be merge conflicts resulting from your overlapping changes. Perhaps you can try and work out how to make things compatible and in which order it will be simpler to merge the pull requests. ... or you pool your efforts and submit a combined pull request that supersedes the individual ones, where the merge conflicts are already resolved?

doc/src/fix_brownian.rst
```diff
@@ -41,8 +41,10 @@ Syntax
        *gt1*, *gt2*, and *gt3* = diagonal entries of body frame translational friction tensor
      *gamma_t* values = *gt* for *brownian* and *brownian/sphere*
         *gt* = magnitude of the (isotropic) translational friction tensor
+     *rotation_style* values = *geometric* or *projection* for *brownian/sphere*
```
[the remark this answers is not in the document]
ArtStrau: Added `.. versionadded:: TBD` for the new `rotation_style` keyword in `fix_brownian.rst`.

doc/src/fix_brownian.rst
```diff
@@ -193,8 +212,14 @@ If the *dipole* keyword is used, then the dipole moments of the particles
 are updated as described above. Only compatible with *brownian/asphere*
 (as *brownian/sphere* updates dipoles automatically).
 
+If the *rotation_style* keyword is used with the *geometric* value, then the
+geometric, rotation-based integration scheme (:ref:`(Hoefling) <Hoefling1>`)
+is used. If the keyword is used with the *projection* value, the linearized,
+projection-based scheme (:ref:`(Ilie) <Ilie1>`) is used. Only
+compatible with *brownian/sphere*.
```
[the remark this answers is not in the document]
ArtStrau: Added `.. versionchanged:: TBD` in the Default section to document the changed default (behavior change).

doc/src/fix_brownian.rst
```diff
@@ -41,8 +41,10 @@ Syntax
        *gt1*, *gt2*, and *gt3* = diagonal entries of body frame translational friction tensor
      *gamma_t* values = *gt* for *brownian* and *brownian/sphere*
         *gt* = magnitude of the (isotropic) translational friction tensor
+     *rotation_style* values = *geometric* or *projection* for *brownian/sphere*
+
```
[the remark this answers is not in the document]
ArtStrau: Added indented descriptions for `geometric` and `projection` values in the parsed-literal block.

ArtStrau:
@akohlmey @samueljmcameron
Thanks — I think it’s easier to keep the PRs separate (mine is self-contained). I’m happy to coordinate the merge order: if Sam’s PR lands first, I’ll rebase mine onto updated develop and resolve conflicts on my side; if mine lands first, Sam can rebase accordingly. Let me know which order you prefer.

ArtStrau:
@akohlmey CI is still blocked with “This workflow is awaiting approval from a maintainer.” Could you please approve the workflow run so the checks can execute? Thanks.

akohlmey:
> @akohlmey CI is still blocked with “This workflow is awaiting approval from a maintainer.” 

Yup, this is GitHub feature to prevent people from abusing the computing resources.
It will be automatically approved after your first pull request is merged.

> Could you please approve the workflow run so the checks can execute? Thanks.

Done.

ArtStrau:
@akohlmey CI is green now. Thanks!

samueljmcameron:
Hello, sorry just catching up. This is a great improvement to the original BROWNIAN package! I took a quick scan of the code/paper which is a very nice read. The only thing I want to check is that the geometric scheme doesn't take much longer than the original projection scheme for a given time discretisation choice? In my opinion (and from practical experience simulating e.g. Active Brownian Particles with any sort of strongly repulsive, LJ type potential) the time discretisation for which accuracy becomes an issue for the old (projection) scheme are larger than the translational time step can handle without the simulation blowing up anyway. So if the geometric scheme were to take significantly longer per time step, I might suggest having the default remain as the projection scheme to keep things consistent with previous releases of the BROWNIAN package (and then users can specify the geometric scheme if they want higher accuracy). From looking at the code I would guess the two schemes are similar in computational time taken so the above point is probably moot, but just wanted to check.

samueljmcameron:
Also - to address the original question of merging, if @ArtStrau wants to keep them separate that's fine with me. I can resubmit my PR later after this one is merged since mine is much less significant a change.

athomps (approved):
This looks very nice. I approve.

athomps:
@ArtStrau Can this be used also with `fix brownian/asphere`. If not, is an ERROR or WARNING issued?

akohlmey:
> * Default is `rotation_style geometric`, original behavior is available via `rotation_style projection`

@ArtStrau Can you please explain why you don't retain the original behavior?
We generally try to be backward compatible as much as possible unless there are good reasons not to.

src/BROWNIAN/fix_brownian_sphere.cpp
```diff
@@ -26,6 +27,7 @@
 #include "random_mars.h"
 
 #include <cmath>
+#include <cassert>
```
akohlmey: It looks like this header is not needed (anymore).

src/BROWNIAN/fix_brownian_base.h
```diff
@@ -25,6 +25,7 @@ class FixBrownianBase : public Fix {
   void init() override;
   int setmask() override;
   void reset_dt() override;
+  enum RotStyle { ROT_PROJECTION, ROT_GEOMETRIC };
```
akohlmey: Since these constants are used only internally, can you move them into the `protected:` section?

src/BROWNIAN/fix_brownian_base.cpp
```diff
@@ -177,6 +179,14 @@ FixBrownianBase::FixBrownianBase(LAMMPS *lmp, int narg, char **arg) :
         error->all(FLERR, "The planar_rotation keyword is not allowed for 2D simulations");
       iarg = iarg + 1;
 
+    } else if (strcmp(arg[iarg], "rotation_style") == 0) {
+      if (narg == iarg + 1) { error->all(FLERR, "Fix brownian rotation_style requires a value."); }
+
+      if (strcmp(arg[iarg + 1], "projection") == 0) rot_style = ROT_PROJECTION;
+      else if (strcmp(arg[iarg + 1], "geometric") == 0) rot_style = ROT_GEOMETRIC;
+      else error->all(FLERR, "Fix brownian rotation_style must be projection or geometric.");
```
akohlmey:
```suggestion
      else error->all(FLERR, iarg + 1, "Fix {} rotation_style must be 'projection' or 'geometric'", style);
```

This adds an indicator to the problematic word of the command line and uses the style name of the derived class.

src/BROWNIAN/fix_brownian_base.cpp
```diff
@@ -177,6 +179,14 @@ FixBrownianBase::FixBrownianBase(LAMMPS *lmp, int narg, char **arg) :
         error->all(FLERR, "The planar_rotation keyword is not allowed for 2D simulations");
       iarg = iarg + 1;
 
+    } else if (strcmp(arg[iarg], "rotation_style") == 0) {
+      if (narg == iarg + 1) { error->all(FLERR, "Fix brownian rotation_style requires a value."); }
```
akohlmey:
```suggestion
      if (narg == iarg + 1) utils::missing_cmd_args(FLERR, "fix brownian rotation_style", error);
```
For missing arguments we prefer using this utility function.

akohlmey (changes requested):
@ArtStrau I have requested some changes to have this pull request more aligned with the current LAMMPS programming style and our ongoing project of improving error messages. There are more improvements needed outside your changes and I think some changes that were added earlier need testing an corrections.

We can include them into this pull request or I can add them to a separate pull request. Also, if you want to make the changes, I can tell you what to do and what to look for, or I can try to add them myself. Please let me know what you prefer.

ArtStrau:
> @ArtStrau Can this be used also with `fix brownian/asphere`. If not, is an ERROR or WARNING issued?

@athomps Thanks for your assessment.

I tested `rotation_style` with `fix brownian/asphere`. With valid asphere syntax (`gamma_t_eigen`/`gamma_r_eigen`), both `rotation_style geometric` and `rotation_style projection` currently parse and `run 0` completes. If `rotation_style` is given without a value, the common option parser errors (“Fix brownian rotation_style requires a value”).

Since `rotation_style` is intended only for `fix brownian/sphere`, I agree it should not be silently accepted for `brownian/asphere`. I will update the option parsing in `fix_brownian_base.cpp` so that specifying `rotation_style` with any fix style other than `brownian/sphere` triggers a clear ERROR (rather than a warning), e.g. “rotation_style is only supported for fix brownian/sphere”. If you’d prefer a WARNING instead, or different wording for the error message, please let me know.

ArtStrau:
@samueljmcameron Thanks for your feedback — I’ll run focused per-step benchmarks (geometric vs projection) and report back with results.

@akohlmey On the default choice, I propose we revisit after I share those benchmark results.

@akohlmey About your follow-up suggestion: could you point me to the exact change you have in mind (file/lines or a brief bullet list of the intended edits)? If it’s small and belongs in this PR, I can implement it here; if it’s more invasive, it may be cleaner as a separate PR — I just want to avoid duplicating effort.

src/BROWNIAN/fix_brownian_sphere.cpp
ArtStrau: @akohlmey Thanks — this comment refers to the added `#include <cassert>`, right? The file still contains an `assert(...)` (and my build fails if I remove `<cassert>`), so I added it explicitly.
If you’d prefer to avoid `assert` here (since it’s compiled out under `-DNDEBUG`), I can replace it with a proper runtime check via `error->all(FLERR, ...)` and then drop `<cassert>`. Do you have a preference (keep as debug-only `assert`, or convert to runtime error)?

akohlmey:
> @akohlmey About your follow-up suggestion: could you point me to the exact change you have in mind (file/lines or a brief bullet list of the intended edits)? If it’s small and belongs in this PR, I can implement it here; if it’s more invasive, it may be cleaner as a separate PR — I just want to avoid duplicating effort.

@ArtStrau I have made the changes to the current code here: https://github.com/lammps/lammps/pull/4845/changes#diff-cbed5f313d1136ab422ca9f4f52233164c9f455da4d7299e4e69e356f48edfbd
I also noticed that there were multiple indexing bugs which could lead to segfaults when arguments were missing, so changes would only be required for any options you have added.
They should be made in whichever of the two pull requests is merged later.

P.S.: I forgot to mention that under the hood the Error class now throws an exception, which allows us to intercept errors in the library interface and then, e.g. re-throw them as Python exceptions in the Python module, or have graphical applications like LAMMPS-GUI or OVITO that link to the LAMMPS library and use the library interface, catch them and not instantly terminate when an error in LAMMPS happens.

src/BROWNIAN/fix_brownian_sphere.cpp
akohlmey: > The file still contains an assert(...) 

My bad. I didn't notice it.

> If you’d prefer to avoid `assert` here (since it’s compiled out under `-DNDEBUG`), I can replace it with a proper runtime check via `error->all(FLERR, ...)` and then drop `<cassert>`. Do you have a preference (keep as debug-only `assert`, or convert to runtime error)?

We don't use assertions because they are incompatible with parallel execution. That is why I assumed the include was left over from some debugging.
You have to be careful about using `error->all()`. You can only use it if the same condition *always* happens on *all* MPI ranks. Otherwise, you need to use `error->one()` to avoid a deadlock. For details, please see: https://docs.lammps.org/latest/Developer_notes.html#errors-warnings-and-informational-messages

ArtStrau: done

src/BROWNIAN/fix_brownian_base.h
ArtStrau: done

src/BROWNIAN/fix_brownian_base.cpp
ArtStrau: done

src/BROWNIAN/fix_brownian_base.cpp
ArtStrau: done

ArtStrau:
@athomps @akohlmey Implemented the requested changes

@akohlmey Could you please approve the pending workflow run so the CI checks can run? Thanks!

akohlmey:
@ArtStrau I have made some small changes to make certain that rot_style is consistent with what algorithm is used even though it is not checked for in fix brownian/asphere. Also I applied clang-format. Unfortunately, I cannot push into your branch. Can you please merge with upstream and then apply the attached patch?
[fix_brownian_update.diff.gz](https://github.com/user-attachments/files/24898179/fix_brownian_update.diff.gz)

jtclemm (approved):
Changes look good to me. My only comment is that the doc should clearly indicate that *projection* was the former behavior, particularly if the default is changing to the new integrator.

ArtStrau:
@samueljmcameron @akohlmey

Thanks for the careful feedback. I ran targeted performance and accuracy tests to address (i) per-step overhead and (ii) cost-to-accuracy. Here is a summary.

**(A) Per-step performance (rotation-dominated microbenchmark)** 
I benchmarked “projection vs geometric” in a setup where translation is effectively inert (no interactions, no external forces, minimal I/O), so wall/loop time is dominated by the rotational update. With a Release build (`-O3 -DNDEBUG -march=native -fno-fast-math`), geometric is slower per step: ≈ +20-24% (3D), ≈ +18-21% (2D). Callgrind indicates the dominant (98-99%) extra cost is `sin/cos` in the rotation kernel; `sqrt`/norm terms contribute ~O(1%).

*Note:* As a separate test, switching only trig evaluation to single precision (`sinf/cosf`) reduces the per-step overhead to about +10% (3D) and +5% (2D). With end-of-step normalization enabled, dipole-length roundoff behaves similarly for `sinf/cosf` vs `sin/cos` because normalization dominates length preservation (without normalization, the expected larger length drift with `sinf/cosf` is observed).

**(B) Cost-to-accuracy (torque-free rotational Brownian motion)** 
For torque-free rotational Brownian motion, analytic PDFs are available in 2D and 3D; I quantify error via a TV-distance metric, 
`TV = (1/2) ∫ |p_num(θ) - p_exact(θ)| dθ`. For the same tolerance (e.g., 3D, `0.001`: `TV_geom ≈ 1.23e-4` vs `TV_proj ≈ 1.36e-3`, i.e. ~11× larger error for projection), geometric allows ~10× larger `dt` than projection in these tests. With the measured per-step overhead, this corresponds to an overall time-to-solution improvement of roughly `10/1.2 ≈ 8×` in rotation-dominated regimes. This is why I think per-step overhead alone is not always the decisive metric when rotational statistics/accuracy drive the timestep choice.

**(C) Translation-limited interacting systems** 
I agree that for stiff repulsive interactions the stable translational timestep can dominate. In that regime the rotation integrator choice typically becomes a smaller fraction of total runtime, and both schemes are accurate at sufficiently small `dt`. Geometric remains somewhat more expensive per step, but its impact on end-to-end runtime can be small when translation dominates.

**Default behavior / backward compatibility (Axel’s question)** 
I set the default to `geometric` because `projection` shows clear large-`dt` artifacts against analytics in torque-free tests, whereas `geometric` remains accurate at substantially larger `dt`. Since `fix brownian/sphere` is often used as a “once set-and-run forever” component, I believe a more robust default reduces the risk of silent misinterpretation in rotation-dominated regimes. Legacy behavior remains available via `rotation_style projection`. This is my personal view; if the project prefers that backward compatibility is more important, I can switch the default to the original behavior.

akohlmey:
> This is my personal view; if the project prefers that backward compatibility is more important, I can switch the default to the original behavior.

I think you have compelling arguments in favor of changing the default, but - as @jtclemm mentioned - this needs to be visibly and clearly documented. It also means that all relevant example log file need to be updated accordingly. It is probably easier and resulting in more consistent output, if I do that in my next "collection" pull request *after* this pull request is merged. Other than that, I am waiting for the patch that I attached to an earlier comment to be included and then we should be good to merge this.

ArtStrau:
> Changes look good to me. My only comment is that the doc should clearly indicate that _projection_ was the former behavior, particularly if the default is changing to the new integrator.

@jtclemm: Thanks, sure, I can adust accordingly.

> I am waiting for the patch that I attached to an earlier comment to be included and then we should be good to merge this.

@akohlmey Yes, thanks -- I saw your comment and will fix this. Was just busy with all the tests.

akohlmey (approved):

ArtStrau:
@akohlmey: merged upstream, applied patch

@jtclemm, @akohlmey: added doc note “projection = former default / legacy”.

samueljmcameron:
Thanks for improving the package @ArtStrau. If I understand your paper correctly, the integrator can also be used in uniaxial objects too, is that right? In that case perhaps the geometric method could be applied to the fix brownian/asphere package as well (where the real difference is just that the body-framed diffusion tensor has 2 or 3 unique eigenvalues) or some variant of it. May be something worth considering in the future if you think it's possible without new theory needed.

ArtStrau:
@samueljmcameron: Thanks, Sam -- agreed this is an interesting point and I also thought about it. The geometric update implemented here is a manifold integrator for a single unit director ($\mathbf{u} \in S^2$), which maps naturally onto ``brownian/sphere`` dipole orientations, so that part was unambiguous and straightforward.

For ``brownian/asphere``, the scope is broader: it advances **full rigid-body orientation** ($SO(3)$, typically via quaternions) and supports general anisotropic diffusion (up to three distinct eigenvalues). The $S^2$ scheme from the PRR paper would carry over cleanly only for a **uniaxial "director-only" subset** where the physics is invariant under spin about $\mathbf{u}$; representing that subset with quaternions is fine. However, once body-fixed features depend on spin (e.g., a dipole tilted relative to the symmetry axis) or for fully triaxial diffusion (*three distinct* eigenvalues), this goes clearly beyond the PRR director-only $S^2$ scheme.

So I think an axisymmetric subset could be a reasonable and practically relevant follow-up, but it’s not obvious to me yet how to integrate it into ``brownian/asphere`` in a way that stays fully transparent and maintainable; e.g., as ``atom_style ellipsoid`` (ellipsoids of revolution) + uniaxial director-only physics.

samueljmcameron:
@ArtStrau yes good point, since the current theory doesn't work for quaternions it's probably a big task to sort that. Maybe someday!

Issue #4844, opened 2026-01-20
[Feature Request] Future additions to the GRAPHICS package


akohlmey:
**Summary**

Here are some ideas for future additions to the GRAPHICS package

**Detailed Description**

- [x] add support for computes that generate image info to `dump image`
- [x] add compute style to compute and visualize hydrogen bonds
- [x] add support for highlighting atoms involved in a reactions modeled by `fix bond/react`
- [x] add support for highlighting atoms involved in Monte Carlo steps of MC package fixes
   - [x] `fix atom/swap`
   - [x] `fix bond/break`
   - [x] `fix bond/create`
   - [x] `fix bond/create/angle`
   - [x] `fix mol/swap`
   - [x] `fix neighbor/swap`
- [x] add support for showing the shape of rigid bodies in `fix rigid` and especially `fix rigid/small`. This could use a similar approach than `fix graphics/isosurface` i.e. define a (local) grid, apply a Gaussian distribution of the particle (number or mass) "density" to the grid points and then use the marching cubes algorithm to triangulate a surface for a given isovalue. Another option would be to define a convex hull using Delauney triangulation. The same could be applied to body style body/nparticle which is in essence the same as `fix rigid/small`.
- [x] add support for creating a color scale bar to `fix graphics/labels`
- [x] add support to position labels in `fix graphics/labels` in image coordinates rather than MD system coordinates
- [x] add support for Delauney triangulation to `fix graphics/isosurface` and rename it back to `fix graphics/surface`
- [ ] add support for creating isosurfaces from general 3d grids and 3d-binning data to `fix graphics/(iso)surface`
- [ ] add support for coloring isosurfaces by a gridded property with interpolation to `fix graphics/(iso)surface`
- [x] add a `fix graphics/lines` command that can create trajectory lines like OVITO, this could use data averaging similar to fix ave/atom (or just set up and use fix ave/atom internally) for smoother lines. Also we should use a limit for the maximum length and have an option to fade the line to transparency. Another possible added feature could be to make the diameter of the lines dependent on some property, e.g. velocity.
- [ ] add support for importing STL objects and support them as global objects in `fix graphics/objects` or for body particles or atoms for atom styles with orientation
[...]

PR #4848, opened 2026-01-22
GRAPHICS package updates

akohlmey:
**Summary**

This pull request includes multiple small GRAPHICS package updates and improvements.

**Related Issue(s)**

Addresses TODO items from #4844

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- split the `Image::draw_pixmap()` method in two variants: one using box coordinates and the other pixmap coordinates (with the former calling the latter). Pixmap coordinates have by default a distance of zero to the camera.
- set up `dump image fix` so that for displaying pixmaps the fflag1 value select whether box coordinates (=0) or image coordinates (=1) are used.
- add support for creating vertical text labels in addition to horizontal ones
- add support for drawing a colormap legend of the atom color map to fix graphics/labels
- add support for reading TGA format files (compressed and uncompressed 8-bit RGB).
- add examples for different color maps with images to the `dump image` documentation

**Post Submission Checklist**


akohlmey:
@sjplimp FYI, with the changes in this pull request it is now possible to use the image coordinate system instead of the box coordinate system to insert images or text. Image coordinates start at (0,0) as the lower left corner and extend to (width, height) as the upper right corner.

sjplimp:
@akohlmey Thanks for adding the option to specify labels in the image plane rather than the simulation box coords.
From the Implementation Notes bullet above and the doc pages it is a little unclear to me how this works.  Here is what I think is the most likely use case, so I'd like to understand if this is the default or at least easy to specify:

(a) there are one or more graphics objects (title, label, plot, colorbar, etc) which the user wants to specify in the image plane and wants them to appear on "top" of everything else in the simulation box image
(b) thus they should be drawn after all the simulation box objects, and effectively not use a depth value, meaning each pixel in the image plane object overwrites any previous pixel value
(c) if two labels (for example) overlap each other (one or more of the same pixels), then the one drawn last (order of fixes?)  should be fully visible, and thus portions of the one drawn first will not be

Note that for simulation box objects, like an atom sphere, I think it is possible to zoom so much that part of the sphere is behind the camera, so if you are relying on depth values to be in some bounded range, that may not be the case.  I only mention this b/c I think the image plane objects should still be visible regardless, i.e. effectively ignore any depth setting of a simulation box image pixel.

Finally, I read the edits to the 2 doc pages in the PR: dump image and fix graphics/lables -- but did not see some of these points clearly explained for the user.  Possibly I missed them.  Also is everything I listed above as an image plane object (e.g. a plot) part of fix graphics/labels?  Or possibly some of these suggestions need to also be applied to other graphics objects?  Which may be your intent with future work?

sjplimp (approved):

akohlmey:
@sjplimp when zooming in very much, the image coordinate label stays on top of the atoms cutting the camera plane. The atom turns into a circle as expected. So we have the desired behavior. See the attached image.

melt-000700

sjplimp:
good to know - thanks for checking

PR #4859, opened 2026-01-28
Improve error messages

aaronsch0:
**Summary**

Refactored error handling in multiple source files to improve user experience.

**Changes:**
* Refactored legacy "Illegal command" error messages in `src/pair_zbl.cpp`, `src/fix_momentum.cpp`, and various styles in `src/EXTRA-PAIR/`.
* Replaced generic error strings with descriptive messages indicating specific causes (e.g., incorrect argument counts, invalid cutoff values).
* Added the `Error::NOLASTLINE` flag to suppress redundant output line printing.

**Related Issue(s)**

Addresses the error message refactoring task suggested by @akohlmey.

**Author(s)**

Aaron Schöne, University of Stuttgart
st183829@stud.uni-stuttgart.de

**Licensing**

**Backward Compatibility**

No backward compatibility issues. The input script syntax remains unchanged; only the error message output is improved.

**Implementation Notes**

The changes strictly follow the pattern of checking `narg` and cutoff values, providing specific error strings, and using `Error::NOLASTLINE` for syntax/parameter errors in the setup phase.

**Post Submission Checklist**


akohlmey:
@aaronsch0 Bad news: almost all of your changes are not correct.
Have you actually tried whether your changes do what you expect?
You should compare to other sources that were already updates.

akohlmey:
@aaronsch0 forget about copilot. Most of its suggestions are garbage.

aaronsch0:
@akohlmey Thank you for the clear warning about Copilot and the pointer to the developer notes.

I have gone through the changes again, using `fix_brownian.cpp` and the documentation as a reference as suggested. I removed `Error::NOLASTLINE` from the setup functions (and used `iarg` in loops) to ensure the input line is printed correctly. I also dismissed the incorrect Copilot suggestions.

Could you please let me know if this now matches the expected style? I want to make sure I haven't overlooked anything else or misunderstood part of the refactoring pattern.

akohlmey:
> Could you please let me know if this now matches the expected style? I want to make sure I haven't overlooked anything else or misunderstood part of the refactoring pattern.

Sorry, but no. I just updated the compute styles and fix the way I would be doing it.

You can put the indicator index as the second argument everywhere where the last line is the command in the input. For fixes and computes that is usually the constructor or functions like `options()` that are sometimes called from it. Inside an options function, you usually first have to figure out the offset in the global argument vector that you are starting from.

I recommend to stay away from pair styles. The situation is rather tricky there.

aaronsch0:
@akohlmey Reverted pair styles as advised. I also applied the utils::missing_cmd_args pattern from your recent commits to src/imbalance_store.cpp.

Does this look correct to you now?

src/imbalance_store.cpp
```diff
@@ -33,7 +33,7 @@ ImbalanceStore::~ImbalanceStore()
 
 int ImbalanceStore::options(int narg, char **arg)
 {
-  if (narg < 1) error->all(FLERR, "Illegal balance weight command");
+  if (narg < 1) utils::missing_cmd_args(FLERR, "balance weight", error);
```
akohlmey:
```suggestion
  if (narg < 1) utils::missing_cmd_args(FLERR, "balance weight store", error);
```
This change would be better.

akohlmey:
> Does this look correct to you now?

close. see my suggestion for improvement.

PR #4863, opened 2026-01-28
Fix F90 and F95 Makefile vars for cmake QUIP-ML compilation

### LINKED ISSUE #4862: [BUG] ML-QUIP makefile defines wrong make var names for latest QUIP version
**Summary**

Latest version of QUIP has changed some Makefile variable names, and LAMMPS cmake compilation no longer sets correct ones

lammps develop branch commit 81453f

**Expected Behavior**

compilation with cmake and `-D PKG_ML-QUIP=ON` compiles LAMMPS

**Actual Behavior**

Compilation fails on QUIP `src/libAtoms/System.F90` with error about line 238 being truncated at column 132. Running `make` with `VERBOSE=1` shows that the expected fortran flags (specifically `--free-line-length-none`), set in `quip_build-prefix/src/quip_build/arch/Makefile.lammps`, are not being passed to fortran compiler.  This is because lammps currently sets those in `F95FLAGS`, while QUIP now expects them in `F90FLAGS`.

**Steps to Reproduce**

Compile lammps with cmake `-D PKG_ML-QUIP=ON`, have it auto-download current QUIP version (as of 2026 Jan 28) and use a compiler that defaults to 132 columnes for fortran source, e.g. gfortran 12.

bernstei:
**Summary**

Update the QUIP-ML cmake file to set the needed fortran 90/95 flags in both F90FLAGS and F95FLAGS make vars, to work with a wider range of QUIP version

**Related Issue(s)**

closes #4862 

**Author(s)**

Noam Bernstein, U S Naval Research Laboratory

**Licensing**

**Backward Compatibility**

no

**Implementation Notes**

Simple change to Makefile was tested by compiling on our local systems with scripts that worked before the QUIP update which broke the lammps compilation

**Post Submission Checklist**

N/A


akohlmey (approved):
Thanks for submitting this. I had noticed that Quip compilation was broken, but didn't have the time to figure out why.

bernstei:
~~Note that there may also be a link-time issue - I'm currently investigating in my local build.  I'll follow up when I'm sure it works all the way through for me.~~ Never mind - I just did a clean compile and it's fine.

However, I also had to fix something on the QUIP side, namely https://github.com/libAtoms/QUIP/pull/708, so it might be worth waiting until that's merged before concluding that this one is complete.

akohlmey:
> However, I also had to fix something on the QUIP side, namely [libAtoms/QUIP#708](https://github.com/libAtoms/QUIP/pull/708), so it might be worth waiting until that's merged before concluding that this one is complete.

This can be worked around with CMake. There is also an issue with the md5.c code with recent C compilers. Ultimately, the people maintaining QUIP should be "read the riot act". They keep including obsolete code that is broken with recent compilers. At some point we have to drop ML-QUIP from LAMMPS because it cannot be compiled anymore.

```
$ git diff
  diff --git a/cmake/Modules/Packages/ML-QUIP.cmake b/cmake/Modules/Packages/ML-QUIP.cmake
  index ccd0b64643..df3f66d744 100644
  --- a/cmake/Modules/Packages/ML-QUIP.cmake
  +++ b/cmake/Modules/Packages/ML-QUIP.cmake
  @@ -9,7 +9,7 @@ endif()
   option(DOWNLOAD_QUIP "Download the QUIP library instead of using an already installed one" ${DOWNLOAD_QUIP_DEFAULT})
   if(DOWNLOAD_QUIP)
     string(TOUPPER "${CMAKE_BUILD_TYPE}" BTYPE)
  -  set(temp "F77 = ${CMAKE_Fortran_COMPILER}\nF90 = ${CMAKE_Fortran_COMPILER}\nF95 = ${CMAKE_Fortran_COMPILER}\n")
  +  set(temp "FC = ${CMAKE_Fortran_COMPILER}\nF77 = ${CMAKE_Fortran_COMPILER}\nF90 = ${CMAKE_Fortran_COMPILER}\nF95 = ${CMAKE_Fortran_COMPILER}\n")
     set(temp "${temp}CC=${CMAKE_C_COMPILER}\nCPLUSPLUS=${CMAKE_CXX_COMPILER}\nLINKER=${CMAKE_Fortran_COMPILER}\n")
     if(CMAKE_Fortran_COMPILER_ID STREQUAL Intel)
       set(temp "${temp}FPP=${CMAKE_Fortran_COMPILER} -E\nOPTIM=${CMAKE_Fortran_FLAGS_${BTYPE}}\n")
  @@ -29,7 +29,7 @@ if(DOWNLOAD_QUIP)
     else()
       message(FATAL_ERROR "The ${CMAKE_Fortran_COMPILER_ID} Fortran compiler is not (yet) supported for building QUIP")
     endif()
  -  set(temp "${temp}CFLAGS += -fPIC -Wno-return-mismatch \nCPLUSPLUSFLAGS += -fPIC -Wno-return-mismatch\nAR_ADD=src\n")
  +  set(temp "${temp}CFLAGS += -fPIC -Wno-return-mismatch -DPROTOTYPES=1\nCPLUSPLUSFLAGS += -fPIC -Wno-return-mismatch\nAR_ADD=src\n")
     set(temp "${temp}MATH_LINKOPTS=")
     foreach(flag ${BLAS_LIBRARIES})
       set(temp "${temp} ${flag}")
```

akohlmey:
FYI, I am including your PR and my commit https://github.com/akohlmey/lammps/commit/6e84a73d5e256a32c21450ec919873618a5aff29 in a future maintenance pull request of mine (so there only have be one review and we are short on reviewers these days).

bernstei:
Out of curiosity, what's wrong with `md5.c`? I just checked with gcc 14, which is fairly recent and it compiles without any special flags. I guess I do get a bunch of warnings from `icx`, that have to do with prototypes.  Looks like they're putting together a big update PR, so maybe I can get that fix added.

bernstei:
> This can be worked around with CMake.

True, and that's what I did locally for testing purposes, but https://github.com/libAtoms/QUIP/pull/708 has been merged, so there should be no need.  

Since you mentioned it, I'm also checking a fix https://github.com/libAtoms/QUIP/pull/710 for the `md5.[ch]` issue.

akohlmey:
> Out of curiosity, what's wrong with `md5.c`? I just checked with gcc 14, which is fairly recent and it compiles without any special flags.

The default setting is that it doesn't use prototypes, i.e. conforms only to K&R C and not ANSI C,  which is considered *very* outdated. These fail with GCC 15. I've run across similar problems with other packages that keep copies of very old code around.

akohlmey:
> so there should be no need.

But it doesn't hurt either.

stanmoore1 (approved):

akohlmey:
Changes were superseded by PR #4864

PR #4866, opened 2026-01-30
add global array and  thermo_modify colname auto to fix colvars

alphataubio:
**Summary**

ADDED TO `fix colvars`:

- [x] access to global array of the current values of each variable already computed and stored in the colvars library
- [x] automatic thermo headers based on the names of each colvar from configuration when using `thermo_modify colname auto`

**Related Issue(s)**

Fixes https://matsci.org/t/how-to-get-current-value-colvar-into-lammps-variable/66820

**Author(s)**

@alphataubio 

**Licensing**

**Backward Compatibility**

n/a

**Implementation Notes**

- [x] tested with `examples/PACKAGES/colvars/in.peptide-colvars`

**Post Submission Checklist**


jrgissing:
Please use the same convention as existing fixes for auto thermo column names. E.g., the fix ID should be included. See fix nvt as an example.

alphataubio:
> same convention as existing fixes

@jrgissing done ✅

src/COLVARS/fix_colvars.cpp
```diff
@@ -456,7 +470,10 @@ void FixColvars::setup(int vflag)
   if (me == 0) {
     setup_io();
     proxy->parse_module_config();
+    size_vector = proxy->colvars->num_variables();
```
giacomofiorin: Should we probably make this optional, since some use cases may involve a lot of collective variables?

src/COLVARS/fix_colvars.cpp
```diff
@@ -921,6 +938,40 @@ double FixColvars::compute_scalar()
...
 }
 
+/* ---------------------------------------------------------------------- */
+
+double FixColvars::compute_vector(int n)
+{
+  double value;
+  if (comm->me == 0) {
+    auto *variables = proxy->colvars->variables();
+    value = (*variables)[n]->value();
```
giacomofiorin: There's a caveat here: some of the variables computed by Colvars have vector values themselves. Therefore, calling `value()` as done here would work for most variables (which are scalar), but raise an error otherwise that could not be worked around.

At a minimum, for this PR to work in all Colvars use cases, only the scalar variables should be processed here (I can add that check), and this restriction documented in the LAMMPS doc.

giacomofiorin (changes requested):
Looks good overall, but there are two caveats that I'm commenting on specifically. Below are more general considerations (not necessarily for this PR).

Automatically setting `thermo` variables is useful, but I was also wondering about using `fix_modify` to define and set LAMMPS variables. This would give access to more than just the current values of the `colvar` objects: for instance, you could get the generalized forces, or the restraint centers. But the LAMMPS variables would have to be created by the `fix_modify` command itself, and I didn't know how to do that.

For example, in the current code the following command would simply print the value of "xi" to the log:
```
fix_modify Colvars colvar "xi" value
```
But what we may want is saving its value into a LAMMPS variable; following your example in the forum, this could be:
```
run 1000000 every 1000 &
    "fix_modify Colvars colvar "xi" value dist" &
    "if ${dist} < 3 then  …"
```

If you have a good idea about setting the LAMMPS variables through `fix_modify`, I can help with corresponding changes from the Colvars library. This could be a separate PR in the Colvars repo, to which it's straightforward to give you push access.

doc/src/fix_colvars.rst
```diff
@@ -168,6 +168,14 @@ Aside from the state information in a ".colvars.state" file, other
 are produced by Colvars depending on the type of simulation.  For this
 reason, the "output" keyword is required for fix colvars.
 
+.. versionadded:: ???
+
+This fix computes a global vector of the current values of each colvar. This
+fix supports automatically generated thermo column names when using
+:doc:`thermo_modify colname auto <thermo_modify>`. The thermo column names
+are “f_”, followed by the fix ID, followed by a colon, followed by the colvar
+name as defined by the Colvars configuration, followed by the global vector index.
```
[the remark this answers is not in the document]
alphataubio: done

doc/src/fix_colvars.rst
```diff
@@ -168,6 +168,14 @@ Aside from the state information in a ".colvars.state" file, other
 are produced by Colvars depending on the type of simulation.  For this
 reason, the "output" keyword is required for fix colvars.
 
+.. versionadded:: ???
+
+This fix computes a global vector of the current values of each colvar. This
+fix supports automatically generated thermo column names when using
+:doc:`thermo_modify colname auto <thermo_modify>`. The thermo column names
+are “f_”, followed by the fix ID, followed by a colon, followed by the colvar
```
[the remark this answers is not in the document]
alphataubio: done

src/COLVARS/fix_colvars.cpp
alphataubio: size_vector needs to be known, all this does is gets the .size() of a std::vector which is trivial. the compute_vector is "on-demand", ie. even if there are 20 colvars but only 1 or 2 are requested as variables or thermo then only those 2 get called

src/COLVARS/fix_colvars.cpp
alphataubio: good catch ! ill make it `compute_array()` instead of `compute_vector()`, then each colvar can be one row for the array. only column 0 available for a scalar colvar and columns 0,...,n-1 for a vector colvar

giacomofiorin: Isnt't the number of columns fixed in an array variable?

alphataubio: ill use the maximum vector length of all colvars as the number of array columns, and return 0.0 for empty columns:

[ 1.0 0.0 0.0 0.0 ] <- scalar colvar
[ 2.0 3.0 4.0 0.0 ] <- 3d colvar
[ 5.0 6.0 7.0 8.0 ] <- quaternion colvar
...

this way the scalar colvar is `f_id[1][1]`, the 3d colvar is `f_id[2][1], f_id[2][2], f_id[2][3]`, ...

akohlmey: From fix.h
```
  int scalar_flag;                 // 0/1 if compute_scalar() function exists
  int vector_flag;                 // 0/1 if compute_vector() function exists
  int array_flag;                  // 0/1 if compute_array() function exists
  int size_vector;                 // length of global vector
  int size_array_rows;             // rows in global array
  int size_array_cols;             // columns in global array
  int size_vector_variable;        // 1 if vec length is unknown in advance
  int size_array_rows_variable;    // 1 if array rows is unknown in advance
  int global_freq;                 // frequency s/v data is available at
```

giacomofiorin: > this way the scalar colvar is `f_id[1][1]`, the 3d colvar is `f_id[2][1], f_id[2][2], f_id[2][3]`, ...

Ok, sorry to make this complicated, but there are variables that compute a vector with more than 4 components. One example is [`cartesian`](https://colvars.github.io/master/colvars-refman-lammps.html#sec:cvc_cartesian), which is used for custom or scripted functions. One would normally disable the output of these rather complex variables, so it would be also good to not have to allocate a large 2D array.

alphataubio: arrays are not actually allocated in lammps, there are no computation done in lammps and no additional memory used because the values are already computed and stored in colvars library.

`compute_array(int m, int n)` just returns on-demand the value directly proxy pointer to the colvar.

giacomofiorin: Ok, then. You could just check that the dimensionality is correct at runtime. The `value()` method of the `colvar` class returns an object with a `size()` method

alphataubio: @giacomofiorin done... `double FixColvars::compute_array(int m, int n)`

akohlmey:
> If you have a good idea about setting the LAMMPS variables through `fix_modify`,

@giacomofiorin 
Not sure if this is what you need, but LAMMPS has "internal" style variables where the value is set by code in LAMMPS and not some variable evaluation. From the documentation:

> internal-style variables require their values be set by code elsewhere in LAMMPS. When a LAMMPS input script or command evaluates an internal-style variable, it must have a current value set (internally) via that mechanism.

So you could have a table or map that registers an (existing) internal style variable with the fix and then the fix will update it.

alphataubio:
> > If you have a good idea about setting the LAMMPS variables through `fix_modify`,
> 
> @giacomofiorin Not sure if this is what you need, but LAMMPS has "internal" style variables where the value is set by code in LAMMPS and not some variable evaluation. From the documentation:
> 
> > internal-style variables require their values be set by code elsewhere in LAMMPS. When a LAMMPS input script or command evaluates an internal-style variable, it must have a current value set (internally) via that mechanism.
> 
> So you could have a table or map that registers an (existing) internal style variable with the fix and then the fix will update it.

@akohlmey you can disregard that earlier discussion. the legacy fix_modify in colvars just sent the `fix_modify get` command to colvars script to print the value of the variable in the log. 

my new version properly creates the plumbing between f_ID[I][J] and pointers into the colvars lib memory for any arbitrary type of colvars no kludge, no reverse scraping of strings, no need to reupdate the value of variables at every step because compute_array() returns live value directly from colvars module data structures.

_ps: compute_array() doesnt actually re-compute anything, it just exposes what was already computed in the colvars module._

athomps (approved):
I approve.

src/COLVARS/fix_colvars.cpp
```diff
@@ -305,142 +264,106 @@ void FixColvars::init_taglist()
...
   return 0;
 }
 
-
 void FixColvars::setup_io()
 {
   if (comm->me == 0) {
     proxy->set_input_prefix(std::string(inp_name ? inp_name : ""));
-    if (proxy->input_prefix().size() > 0) {
+    if (proxy->input_prefix().size() > 0)
```
giacomofiorin: Unlike the one-line if statements, this cosmetic change makes the code less readable IMO.

doc/src/fix_colvars.rst
```diff
@@ -168,6 +168,39 @@ Aside from the state information in a ".colvars.state" file, other
...
+.. versionadded:: TBD
+
+This fix accesses a global array of the current values of each collective variable computed in the Colvars module. The number of rows of the global array is equal to the number of collective variables and the columns are:
+
+* `f_ID[I][1]` for a scalar-valued collective variable
+* `f_ID[I][1], ..., f_ID[I][N]` for a vector-valued collective variable
+
+Almost all collective variables are scalars:
+
+* distances (`distance, distanceZ, distanceXY, distanceInv <https://colvars.github.io/colvars-refman-lammps.html#sec:cvc_distances>`_)
```
giacomofiorin: This list is both quite detailed, but also incomplete. I suggest simplifying so that we list at most 1-2 *examples* for scalar vs. non-scalar variables.  Duplicating the full list of variable types here would only make it outdated in the future.

src/COLVARS/fix_colvars.cpp
```diff
@@ -921,6 +755,66 @@ double FixColvars::compute_scalar()
   return energy;
 }
 
+/* ---------------------------------------------------------------------- */
+
+void FixColvars::update_colvars()
```
giacomofiorin: This being a setup function, which is not being called during a time step, could be named differently. How about `reinit_colvars()`?

giacomofiorin (approved):
Thanks for addressing the dimensionality issue, and for your patience. I just resolved conflicts that arose with the code in the Colvars repo (due to the extensive reformatting of this PR), and opened a [PR](https://github.com/Colvars/colvars/pull/916) there.

Overall, this looks good to me now save for very minor issues with code style and documentation.

1. Removing lots of empty lines without replacing them with comments makes the code much denser. However, for files under `src`, the LAMMPS core devs have the last word.

2. I'll add a quick commit to simplify how the dimensionality of the variables are discussed. If you really would want to list all possible variable types when discussing the thermo variables, please revert that commit and then add the missing types.

src/COLVARS/fix_colvars.cpp
```diff
@@ -82,49 +86,39 @@ int FixColvars::instances = 0;
 FixColvars::FixColvars(LAMMPS *lmp, int narg, char **arg) :
   Fix(lmp, narg, arg)
 {
-  if (narg < 4)
-    error->all(FLERR,"Illegal fix colvars command: too few arguments");
-
-  if (instances > 0)
-    error->all(FLERR,"Only one colvars fix can be active at a time");
+  if (narg < 4) error->all(FLERR,"Illegal fix colvars command: too few arguments");
```
akohlmey: This should use `utils::missing_cmd_args()` like it is done elsewhere in LAMMPS

src/COLVARS/fix_colvars.cpp
```diff
@@ -82,49 +86,39 @@ int FixColvars::instances = 0;
...
+  if (instances > 0) error->all(FLERR,"Only one colvars fix can be active at a time");
```
akohlmey:
```suggestion
  if (instances > 0)
    error->all(FLERR, Error::COMMAND, "Only one fix colvars instance may be active at a time");
```
This should be better.

src/COLVARS/fix_colvars.cpp
```diff
@@ -82,49 +86,39 @@ int FixColvars::instances = 0;
...
   root2root = MPI_COMM_NULL;
   proxy = nullptr;
-
-  if (strcmp(arg[3], "none") == 0) {
-    conf_file = nullptr;
-  } else {
-    conf_file = utils::strdup(arg[3]);
-  }
-
+  if (strcmp(arg[3], "none") == 0) conf_file = nullptr;
```
akohlmey: Please revert for readability. You can leave the curly braces away if you want, but it should be at least 4 lines instead of 2. The same line version is only ok for a single line "if" without "else".

src/COLVARS/fix_colvars.cpp
```diff
@@ -82,49 +86,39 @@ int FixColvars::instances = 0;
...
+  else conf_file = utils::strdup(arg[3]);
   rng_seed = 1966;
   unwrap_flag = 1;
-
   inp_name = nullptr;
   out_name = nullptr;
   tfix_name = nullptr;
 
-  /* initialize various state variables. */
+  // initialize various state variables
```
akohlmey: Please leave non-functional changes out of your pull requests. Comments, empty lines, and conditionals should remain at the same style as before.
Any larger style changes we *only* accept for the maintainers of a package
and usually they should start with enabling and using clang-format.

src/COLVARS/fix_colvars.cpp
```diff
@@ -243,57 +228,31 @@ int FixColvars::setmask()
   return mask;
 }
 
+/* ---------------------------------------------------------------------- */
```
akohlmey, line 249: This is one is an exception to what I wrote about style changes. Adding this comment separator is correcting an omission and consistent with the LAMMPS programming conventions.

src/COLVARS/fix_colvars.cpp
```diff
@@ -449,60 +371,42 @@ void FixColvars::setup(int vflag)
...
   if (me == 0) {
-
     std::vector<int>           &tp = *(proxy->modify_atom_types());
     std::vector<cvm::atom_pos> &cd = *(proxy->modify_atom_positions());
     std::vector<cvm::rvector>  &of = *(proxy->modify_atom_total_forces());
     std::vector<cvm::real>     &m  = *(proxy->modify_atom_masses());
     std::vector<cvm::real>     &q  = *(proxy->modify_atom_charges());
-
     // store coordinate data in holding array, clear old forces
-
```
akohlmey, line 493: Combining two empty lines into one would also be ok, but the removing the empty lines completely is not a desired change and in conflict with our conventions.

src/COLVARS/fix_colvars.cpp
```diff
@@ -575,29 +462,18 @@ void FixColvars::setup(int vflag)
...
+    // blocking receive to wait until it is our turn to send data
     MPI_Recv(&tmp, 0, MPI_INT, 0, 0, world, MPI_STATUS_IGNORE);
     MPI_Rsend(comm_buf, nme*size_one, MPI_BYTE, 0, 0, world);
   }
-
   // run pre-run setup in colvarproxy
-  if (me == 0)
-    proxy->setup();
-
+  if (me == 0) proxy->setup();
```
akohlmey, line 532: This is a case of an exception that I mentioned earlier where joining lines is acceptable. The same would be done by clang-format.

src/COLVARS/fix_colvars.cpp
```diff
@@ -614,60 +490,47 @@ void FixColvars::setup(int vflag)
 void FixColvars::post_force(int /*vflag*/)
 {
   const auto me = comm->me;
-
-  // some housekeeping: update status of the proxy as needed.
-  if (me == 0) {
+  // some housekeeping: update status of the proxy as needed
+  if (me == 0)
```
akohlmey: For nested conditionals or loops, all but the innermost curly braces should remain.

src/COLVARS/fix_colvars.cpp
```diff
@@ -614,60 +490,47 @@ void FixColvars::setup(int vflag)
...
     if (proxy->want_exit())
       error->one(FLERR,"Run aborted on request from colvars module.\n");
```
akohlmey:
```suggestion
      error->one(FLERR, Error::NOLASTLINE, "Run aborted on request from colvars module.");
```
This avoids the pointless last line was "run ####" message.

src/COLVARS/fix_colvars.cpp
```diff
@@ -921,6 +754,66 @@ double FixColvars::compute_scalar()
...
+{
+  int sizes_array[2];
+  if (comm->me == 0) {
+    proxy->parse_module_config();
+    const auto& variables = *proxy->colvars->variables();
+    size_array_rows = variables.size();
+    size_array_cols = 0;
+    for ( int m=0 ; m<size_array_rows; m++ ) {
+      int variable_size = static_cast<int>(variables[m]->value().size());
+      size_array_cols = std::max<int>(size_array_cols, variable_size);
```
akohlmey: Why not use the existing and pre-defined MAX() macro?

src/COLVARS/fix_colvars.cpp
```diff
@@ -921,6 +754,66 @@ double FixColvars::compute_scalar()
...
+  output->thermo->colname_auto();
+}
+
+double FixColvars::compute_array(int m, int n)
+{
+  double value = 0.0;
+  if (comm->me == 0) {
+    const auto& variables = *proxy->colvars->variables();
+    if (m >= variables.size())
+      error->all(FLERR, "f_{}[{}][{}] out-of-bounds: {} collective variables available.",
```
akohlmey:
```suggestion
      error->all(FLERR, Error::NOLASTLINE, "f_{}[{}][{}] out-of-bounds: {} collective variables available.",
```

src/COLVARS/fix_colvars.cpp
```diff
@@ -921,6 +754,66 @@ double FixColvars::compute_scalar()
...
+                 id, m+1, n+1, variables.size());
+    const auto& variable = variables[m]->value();
+    if (n >= variable.size())
+      error->all(FLERR, "f_{}[{}][{}] out-of-bounds: collective variable {} has size {}.",
```
akohlmey:
```suggestion
      error->all(FLERR, Error::NOLASTLINE, "f_{}[{}][{}] out-of-bounds: collective variable {} has size {}.",
```

akohlmey (changes requested):
I agree with Giacomo, that most of the "cosmetic" changes collide with the readability and we generally prefer changes to the coding conventions and source formatting to be separated from functional changes. There are a small number of exceptions that I flagged where the change is making the code *more* consistent with the LAMMPS conventions, but most like the change of C style comments to C++ style or the removal of empty lines are not.

src/COLVARS/fix_colvars.cpp
alphataubio: update_colvars() -> setup_colvars()

src/COLVARS/fix_colvars.cpp
alphataubio: done

alphataubio:
> 2. I'll add a quick commit to simplify how the dimensionality of the variables are discussed. If you really would want to list all possible variable types when discussing the thermo variables, please revert that commit and then add the missing types.

of all the changes requested by @giacomofiorin and @akohlmey, this is the only one i disagree with. i reverted to the more comprehensive doumentation that explains in more clear details the structure of the global array under different possible types of colvars.

src/COLVARS/fix_colvars.cpp
alphataubio: done

src/COLVARS/fix_colvars.cpp
alphataubio: done

src/COLVARS/fix_colvars.cpp
alphataubio: done

akohlmey:
> If you have a good idea about setting the LAMMPS variables through `fix_modify`

@giacomofiorin here is a simple example to create an internal style variable and change it.
```
        auto *variable = input->variable;
        char *varname = (char *) "mynewvar";

        std::vector<char*> args(4,nullptr);
        args[0] = (char *) varname;
        args[1] = (char *) "internal";
        args[2] = (char *) "0.0";

        // create and initialize the internal variable
        variable->set(args.size()-1,args.data());

        // check if set correctly

        printf("new value of variable is %s\n", variable->retrieve(varname));

        // get index of variable and change its internal value directly
        int idx = variable->find(varname);
        if (variable->internalstyle(idx)) {
            variable->internal_set(idx, 2.0);
        }
        printf("new value of variable is %s\n", variable->retrieve(varname));
```
After this you can use the variable anywhere where scalar variables are allowed:
```
new value of variable is 0
new value of variable is 2
print "${mynewvar}"
2
info variable

Printed on Mon Feb  9 23:10:07 2026

Variable information:
Variable[  0]: mynewvar,         style = internal,         def =2

```

src/COLVARS/fix_colvars.cpp
alphataubio: done

src/COLVARS/fix_colvars.cpp
alphataubio: done

alphataubio:
@akohlmey i made all your requested changes. im done for today, have a nice evening. 💤

akohlmey:
> > 2. I'll add a quick commit to simplify how the dimensionality of the variables are discussed. If you really would want to list all possible variable types when discussing the thermo variables, please revert that commit and then add the missing types.
> 
> of all the changes requested by @giacomofiorin and @akohlmey, this is the only one i disagree with. i reverted to the more comprehensive doumentation that explains in more clear details the structure of the global array under different possible types of colvars.

I disagree strongly. This is not information that is relevant to LAMMPS but about colvars, so that information should be kept in the colvars documentation not the LAMMPS manual. At best it should point to the colvars manual, but that should be a suitably "permanent" URL that will not change, if the colvars manual is updated.

Having this in the LAMMPS documentation almost guarantees that it will be outdated and incorrect soon, and what is worse that terse documentation is incorrect documentation.
I have already a hard time chasing down external links and fighting bit rot. Having to
make certain that the LAMMPS manual is consistent with the colvars documentation and implementation is a *very* big ask.

When you have that in the LAMMPS-related part of the colvars documentation, it is visible and searchable by the colvars developers and much easier to keep it in sync with the colvars functionality.

akohlmey:
> @akohlmey i made all your requested changes. im done for today, have a nice evening. 💤

Sorry, but it looks like you only changed the cases that I explicitly pointed out. All other, similar cases are unchanged. I was expecting that making a general statement and pointing out a few examples explains sufficiently what I am looking for. Do you really expect me to point out each case???

giacomofiorin:
> > 2. I'll add a quick commit to simplify how the dimensionality of the variables are discussed. If you really would want to list all possible variable types when discussing the thermo variables, please revert that commit and then add the missing types.
> 
> of all the changes requested by @giacomofiorin and @akohlmey, this is the only one i disagree with. i reverted to the more comprehensive doumentation that explains in more clear details the structure of the global array under different possible types of colvars.

Okay with me to revert it (as I also mentioned above), but the second part of my request is also important.

This section is long to read (for a human at least) because it lists many of the variable types. But it doesn't list all of them. What is your suggestion to address that?

alphataubio:
@akohlmey @giacomofiorin sorry im putting PR back to draft until ill have time to finish it after feb 27. i have what i need for now.

giacomofiorin:
> @akohlmey @giacomofiorin sorry im putting PR back to draft until ill have time to finish it after feb 27. i have what i need for now.

No worries, and thank you. Overall I think this is a useful change, and I'd want to have it, independently from creating named variables via `fix_modify` ([link](https://github.com/lammps/lammps/pull/4866#issuecomment-3875215069)).

sjplimp:
@jrgissing Is the goal of this PR related to your PR (can't recall if merged on still in progress) of generally enabling more descriptive column names for thermo and (possibly) dump output?  In which case, the only changes required should be to fix colvars to provide the info that the Thermo class needs.  But this PR is also modifying thermo.cpp and making FixColvars a "friend" class of Thermo -- both of which do not seem like good things to do ?

jrgissing:
@sjplimp that feature has already been merged. It was applied to just a few fixes and computes at that time. This PR is applying it to fix colvars. @alphataubio why is it necessary to make FixColvars a friend class? 
Also, can you update the log file of the example to correspond to the updated input script?

alphataubio:
> But this PR is also modifying thermo.cpp 

@jrgissing's PR added `colname auto` for _scalar_ and _vector_. i added support to `thermo.cpp` for fix _array_ from `fix colvars` and compute array at same time for others in the future.

```
+         if ( argi.get_dim() == 2 )
+          keyword_user[ifield] += fmt::format("[{}]", argindex2[ifield]);
```

```
+        if ( argi.get_dim() == 2 )
+          keyword_user[ifield] += fmt::format("[{}]", argindex2[ifield]);
```

> making FixColvars a "friend" class of Thermo -- both of which do not seem like good things to do ?

@sjplimp FixColvars needed to call _private_ `Thermo::colname_auto()` when updating its columns with `fix_modify`. My first draft was to have FixColvars a "friend" class of Thermo, but now instead i made `Thermo::colname_auto()` a _public_ method which is less invasive than having access to all the internals as a friend class.

alphataubio:
> @alphataubio why is it necessary to make FixColvars a friend class? 

see previous comment. `Thermo::colname_auto()` is now a public method so that FixColvars can call it.

> Also, can you update the log file of the example to correspond to the updated input script?

done ✅

alphataubio:
@akohlmey did i get all the cosmetic changes you wanted ?

sjplimp:
Re: Thermo and FixColvars classes - I thought one of the goals of Joel's earlier PR was that his changes to the Thermo class meant that adding support for auto column names to a new fix, required no further changes to Thermo, but only adding a method(s) to the fix itself.  Since this PR is changing the Thermo class, is that not the case?

sjplimp:
Sorry - Jake not Joel ...

alphataubio:
i can remove the colname auto feature if it bothers people. i was showboating and i dont really need it

jrgissing:
@sjplimp he is adding colname auto to a higher-dimensional output than the first couple examples (array, vs the previously supported scalar or vector output). future extensions to other arrays will then not need to modify Thermo

jrgissing:
> i can remove the colname auto feature if it bothers people. i was showboating and i dont really need it

it doesn't bother me, I am happy you are using this feature! and it will make it easier to for future users to add arrays with automatic thermo column names. A couple minor points:

1) It slightly undermines this demo when the collective variable names are so generic ('1' going to 'one'). Can they be updated to reflect of the type of collective variable, like distance1?
2) The indexing of automatic thermo column names seem redundant in a couple ways. For example, I would expect 'f_colvars:distance1' for a scalar output, rather than f_colvars:distance1[1][1], since the colvar name is replacing the first index, and the second index can only be '1'.

jrgissing:
also, links to the colvars manual in the docs seem to be broken

alphataubio:
> also, links to the colvars manual in the docs seem to be broken

links removed from rst  ✅

sjplimp:
@jrgissing Please remind me of the column heading syntax for scalars or vector elements, for the previous enabled fixes?  And what is proposed syntax in this newly enabled fix for array elements?

I understand extending Thermo here to enable printing of column headers for global array elements, but otherwise I think there should be zero communication from the fix to Thermo, i.e. a push.  This feature should be 100% pull, where Thermo queries the fix for anything it needs.  I.e. no friend class, no making a Thermo method or data public so a fix can call it, set it, etc.  If either of you think otherwise, please explain ?

jrgissing:
@sjplimp vector output examples include f_1:eta[1] and f_1:eta_dot[1] for 'fix nvt/npt/nph', and a scalar output examples is f_rxns:diels_alder for 'fix bond/react'. Upon closer look at this PR, I believe that no 'array-style' thermo columns are necessary because there is only vector output from fix colvars (see my previous comment), so that should remove all changes to Thermo. @alphataubio let us know if you agree

sjplimp:
Good - I like that syntax for vector elements.  I'm not opposed to augmenting Thermo to include a syntax for array elements.  Even if not needed for FixColvars, maybe it would be useful for other fixes in the future which generate global arrays.  I just want to ensure the handshake between Thermo and any fix is initiated and performed by Thermo (which didn't appear to be the case by this PR), w/out any fix needing to access Thermo.  Unless I am missing something, that seems like the right "API" for column header labeling.

alphataubio:
> I just want to ensure the handshake between Thermo and any fix is initiated and performed by Thermo (which didn't appear to be the case by this PR), w/out any fix needing to access Thermo. Unless I am missing something

`fix colvars` is an edge case because it can `fix_modify` its computed arrays during a run so it needs to tell Thermo to refresh `colname_auto()`. im not aware of other fixes that can change their output scalar/vectors/arrays now but this is the most flexible future api for dynamic output fixes.

sjplimp:
Interesting.  I normally think of fix_modify acting between runs.  Are you saying fix collars may want to change column names during a single run ?

Either way, I think this should work as follows.  At one-time setup between Thermo and a fix, Thermo should query whether the fix needs static or dynamic col names.  If dynamic, the fix can also return a flag for "start-of-each-run" or "every-thermo-output".  Then Thermo will know when it needs to re-query the fix for new col names.

sjplimp:
Oops - col headers are only printed once per run, so ignore my "every-thermo-output" option ...

alphataubio:
**@sjplimp problem solved. it was an order of execution issue.`fix colvars` no longer needs to call `Thermo::colname_auto()` during `FixColvars::setup()` so i reverted `Thermo::colname_auto()` back to private.**

_details below..._


You’re right: this is an **order-of-execution** issue. Here’s the exact sequence and why `colname_auto()` is too early in one case and correct in the other.

---

## Execution order

**1. While the input script is being parsed (before any `run`):**

- `fix 2 all colvars peptide.colvars` → fix is **created** (constructor only).
- `thermo_style custom ... f_2 ...` → Thermo object is **created** and keywords are parsed.
- `thermo_modify colname auto` → **`colname_auto()` runs here.**

At this moment **no fix has run `setup()` yet**. In particular, `FixColvars::setup()` (and thus `setup_colvars()`) has not been called, so the fix does not yet know its array shape or column layout. So any `get_thermo_colname()` calls from `colname_auto()` happen **too early**.

**2. When `run` starts (e.g. `run 100`):**

From `Verlet::setup()` (and similarly in other integrators):

```159:161:lammps-alphataubio/src/verlet.cpp
  modify->setup(vflag);
  output->setup(flag);
  update->setupflag = 0;
```

So:

1. **`modify->setup(vflag)`** runs first → every fix’s `setup()` runs, including **`FixColvars::setup()`** (line 373), which calls **`setup_colvars()`** (line 384). Only after this does the colvars fix know its columns and array dimensions.
2. **`output->setup(flag)`** runs next → it calls **`thermo->header()`** (and then `thermo->compute(0)`).

So the first time the thermo header is printed, **all fixes have already been set up**. If `colname_auto()` runs there, it is **after** fix colvars (and any other fix) knows its columns.

---

## Conclusion

- **`thermo_modify colname auto`** in the script runs during **parsing**, before any `run` → **too early**; fix colvars has not run `setup()` yet, so column names can’t be correct.
- **Calling `colname_auto()` at the start of `Thermo::header()`** runs when the header is first printed, which is from **`output->setup()`**, i.e. **after** `modify->setup()` → **correct order**; fix colvars has already run `setup()` and `setup_colvars()`.

So your reading of the order is correct: the problem is that `colname_auto()` is being run too early when it’s only triggered by `thermo_modify colname auto` in the script. Moving the “refresh column names” behavior to the moment when the thermo header is actually printed (e.g. by calling `colname_auto()` at the beginning of `Thermo::header()` in `thermo.cpp`) fixes the order: it runs once per run when the header is printed, and only after all fixes have been set up.

sjplimp:
Good summary - just one more Q.

It looks like you have changed the call to colname_auto() from a one-time eval when fix_modify enables auto,
to an every-run eval in run setup() via the Thermo::header() method.  I understand why that is needed by fix colvars, but it is not needed by any other fix or compute which provides column headers (currently)?

Is there a simple way to keep it a one-time eval unless some fix like colvars is flagged as "dynamic" and needs it once per run?
I ask b/c there are various use cases where a large number of tiny runs (e.g. run 0) are triggered to run as fast as possible.  I think we should avoid unnecessary overhead in the thermo header construction in iight of those cases.

sjplimp:
And I should not that some use cases involve thermo output with 100s or even 1000s of columns.

doc/src/fix_colvars.rst
```diff
@@ -168,6 +168,39 @@ Aside from the state information in a ".colvars.state" file, other
...
+
+with the exception of:
+
+* 3-dimensional vector (``distanceVec`` and ``distanceDir``)
+
+* 4-dimensional unit quaternion (``orientation``)
+
+* flat vector of :math:`N \!\times\! 3` coordinates (``cartesian``)
+
+* flat vector of :math:`N_1 \!\times\! N_2` distances (``distancePairs``)
```
giacomofiorin: [@alphataubio](https://github.com/alphataubio) The expression "link rot" that you used in an earlier commit implies that the links went bad at the source, but what really happened here is that the links were broken to begin with.

On a careful look, both you and I made mistakes in entering those URLs: however, because you reverted my earlier commit, only your broken links remained, until you finally removed them altogether.

My earlier concern that the doc page for fix colvars is getting more complicated, while at the same time being less complete, was not really addressed:
https://github.com/lammps/lammps/pull/4866#issuecomment-3875145984
I didn't think much of it, because the feature you added here is very useful, and the doc can always be improved afterwards. 

Now, without the links, can you please reconsider if it is really required to include a long list of very specific Colvars keywords in this context?  There is absolutely no way that a new user could get those features working without reading the Colvars doc, and until then, no way that they could use them as `thermo` columns.

giacomofiorin (dismissed):
See below for doc issues.

src/thermo.cpp
```diff
@@ -1218,13 +1220,19 @@ void Thermo::colname_auto()
     ArgInfo argi(word);
     if (argi.get_type() == ArgInfo::COMPUTE) {
       auto *icompute = modify->get_compute_by_id(argi.get_name());
-      if (icompute->thermo_modify_colname)
+      if (icompute != nullptr && icompute->thermo_modify_colname) {
```
akohlmey: If `icompute` is nil, there should be an error here instead of ignoring it. Same for `ifix` below.

src/thermo.cpp
```diff
@@ -1218,13 +1220,19 @@ void Thermo::colname_auto()
...
         keyword_user[ifield] = icompute->get_thermo_colname(argindex1[ifield]-1);
+        if ( argi.get_dim() == 2 )
```
akohlmey:
```suggestion
        if (argi.get_dim() == 2)
```

src/thermo.cpp
```diff
@@ -1218,13 +1220,19 @@ void Thermo::colname_auto()
...
+          keyword_user[ifield] += fmt::format("[{}]", argindex2[ifield]);
+      }
     }
     if (argi.get_type() == ArgInfo::FIX) {
       auto *ifix = modify->get_fix_by_id(argi.get_name());
-      if (ifix->thermo_modify_colname)
+      if (ifix != nullptr && ifix->thermo_modify_colname) {
         keyword_user[ifield] = ifix->get_thermo_colname(argindex1[ifield]-1);
+        if ( argi.get_dim() == 2 )
```
akohlmey:
```suggestion
        if (argi.get_dim() == 2)
```

src/COLVARS/fix_colvars.cpp
```diff
@@ -172,22 +175,15 @@ int FixColvars::parse_fix_arguments(int narg, char **arg, bool fix_constructor)
...
       if (iarg + 1 == narg) {
-        if (fix_constructor) {
-          error->all(FLERR, ("Missing argument to keyword \""+
-                             std::string(arg[iarg]) +"\""));
-        } else {
-          // Error code consistent with Fix::modify_param()
-          return 0;
-        }
+        if (fix_constructor)
+          error->all(FLERR, ("Missing argument to keyword \""+std::string(arg[iarg]) +"\""));
```
akohlmey:
```suggestion
          error->all(FLERR, "Missing argument to keyword {}", arg[iarg]);
```
`Error::all()` is now overloaded to take a std::format style format string and variable arguments, so that hard to read hack from before is no longer needed.

src/thermo.cpp
alphataubio: done. for stuff like this (spaces, tabs, brackets, ...) could there be a makefile target (similar to make check-whitespace) auto clang formatting based on a standard lammps code formattting template ?

akohlmey: Typing `clang-format -i *.cpp` will do just as well.
Even easier is to configure your editor to run `clang-format` on its buffer, either from a hotkey or automatically before saving.

alphataubio:
i removed `colname auto` form everywhere and reverted thermo.[h,cpp] to develop to get this PR approved

src/thermo.cpp
alphataubio: i reverted all my changes to thermo.cpp back to develop

doc/src/fix_colvars.rst
alphataubio: done ✅

alphataubio:
all done

akohlmey:
@giacomofiorin and @sjplimp can you please have a look before I do a last pass and merge it, if there are no new objections.

sjplimp (approved):

akohlmey (approved):

giacomofiorin:
> @giacomofiorin and @sjplimp can you please have a look before I do a last pass and merge it, if there are no new objections.

I'm okay with the latest version (although I didn't get a chance to act on this message from you in time before merging).

@alphataubio Thank you for this very useful new feature!

PR #4867, opened 2026-02-01
Add a compute hbond/local and dump image support for computes

akohlmey:
**Summary**

This adds a new compute style hbond/local to compute hydrogen bond info and extends the dump image support for graphics info from fixes to also support computes with the corresponding keywords.  Except for calling a compute function instead of a fix function, the internal processing is identical.

**Related Issue(s)**

Inspired by https://matsci.org/t/computing-hydrogen-bonds-in-lammps/66826/

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

Some changes were required to the processing of arrow objects that have been reflected in the documentation of the affected fix styles. Minor issues in graphics related code are also addressed. The default color map uses the same color twice under two different names ("aqua" and "cyan"), which can lead to the confusing situation that atom types differing by 1 have the same color.  This default was changed to use "cyan" and "magenta" instead.

To simplify selecting whole molecules with dynamic groups, an "include molecule" option was added to the `group dynamic` keyword.

**Post Submission Checklist**

-----

Some questions for reviewers:
- ~~should the sign of the "hydrogen bond strength" be reversed? Currently, the typical strength of an SPC/E water hydrogen bond is about -5.5 kcal/mol. The (few) publications I looked at use a positive value (so does Wikipedia).~~  Done in commit acb677b9c52c339014cc4ee7a3562c707cdb9a18
- ~~should there be an (optional!) energy criterion? E.g. only count a hydrogen bond if the strength is above a certain value? e.g. 1.5 kcal/mol for water-water?~~ Done in commit acb677b9c52c339014cc4ee7a3562c707cdb9a18
- would it be worth to also support systems without explicit bonds? e.g. by using a distance cutoff (half the hydrogen bond cutoff?) and walking the neighbor list a second time looking for atoms in the hydrogen selection? This would probably make the most sense with ReaxFF, but for that we could use the actual HBondList info and produce a separate `compute hbond/reaxff` specifically for that purpose similar to `fix reaxff/bonds`. Are there other cases?
- is it enough to output the atom IDs and not also coordinates (could be easily looked up from the regular trajectory)?

-----
hbond-000300hbonds-02000

https://youtube.com/shorts/1QEjIITapwQ


doc/src/compute_hbond_local.rst
```diff
@@ -0,0 +1,247 @@
...
+  donor and acceptor atom in distance units
+- The *angle* value is the current hydrogen-donor-acceptor angle in degrees
+- The *hdist* value is the current distance between the hydrogen atom
+  and hydrogen bond acceptor atom in distance units
+- The *ehb* value is the hydrogen bond strength computed as the sum of
+  the pairwise potential energy between a) the donor atom and the
+  acceptor atom, and b) the hydrogen atom and the acceptor atom.  This
+  is typically a positive value for an attractive interaction.
+
+If the *ecut* keyword is used, and additional energy cutoff is applied.
```
athomps:
```suggestion
If the *ecut* keyword is used, an additional energy cutoff is applied.
```

doc/src/compute_hbond_local.rst
```diff
@@ -0,0 +1,247 @@
...
+The computed hydrogen bond strength must be larger than the *ecut* value
```
athomps, line 82: In a hydrogen bond, the acceptor-hydrogen energy should be attractive, should it not? In that case, it would make sense to only count bonds where pair energy is *less than ecut*.  In other words, this test in the code should be:
`                if (epot < ehbcutoff)) { 
`
The default behavior would be do not test.

athomps (dismissed):
I have a concern about the definition of *ecut* that should be addressed.  Also, has any testing been done to compare with other tools like VMD?

doc/src/compute_hbond_local.rst
akohlmey: Please see one of the (now crossed out) "questions to reviewers". The convention seems to be to use a "hydrogen bond strength" which is positive. Papers I've found typically mention about 5 kcal/mol and not -5 kcal/mol. Thus the sign of the energy from `Pair::single()` is reversed and the sign of the test correct. This should be mentioned in the documentation, if I remember correctly.

akohlmey:
> Also, has any testing been done to compare with other tools like VMD?

You cannot compare the bulk SPC/E water data with VMD directly, since it uses a slightly different definition of the angle (D-H-A and not the more common H-D-A) and it does not properly handle PBC.

*But* the results for the same cutoffs match published results (nHB/molecule, eHB, length and angle distributions and maxima) which apply only geometric conditions.

doc/src/compute_hbond_local.rst
akohlmey: FYI, reference output and distributions are in the example folder. I have added this to the rdf-adf folder since it uses the same data file and does a related computation. In fact, the distance cutoff should be taken from the first minimum of the oxygen-oxygen g(r).

src/EXTRA-COMPUTE/compute_hbond_local.cpp
```diff
@@ -0,0 +1,397 @@
...
+static constexpr double EPSILON = 1.0e-10;
+
+enum { HYDROGEN = 0, DONOR, ACCEPTOR, DIST, ANGLE, HDIST, ENGPOT, MAXVAL };
+
+/* ---------------------------------------------------------------------- */
+
+ComputeHBondLocal::ComputeHBondLocal(LAMMPS *lmp, int narg, char **arg) :
+    Compute(lmp, narg, arg), list(nullptr), alocal(nullptr), imgobjs(nullptr), imgparms(nullptr)
+{
+  if (atom->molecular == Atom::ATOMIC)
```
jtclemm, line 51: Will this compute work withe Atom::TEMPLATE? Is the special list guaranteed to be built in that case?

akohlmey: It has to. Otherwise you could not apply exclusions correctly to the non-bonded interactions.

akohlmey: It has to. Otherwise you could not apply exclusions correctly to the non-bonded interactions.

akohlmey: It has to. Otherwise you could not apply exclusions correctly to the non-bonded interactions.

jtclemm (approved):
Can't comment on the physical choices, but the implementation looks good to me.

doc/src/compute_hbond_local.rst
akohlmey: @athomps The default value of `ehbcutoff = -1.0` is safe because, if it is set, you would use a positive value and then also epot is computed from `Pair::single()`, otherwise epot is set to 0.0. Thus the test for `((ehbcutoff < 0.0) || (epot > ehbcutoff))` will be always true if `epot` is not computed and ehbcutoff is negative and otherwise only true if the hydrogen bond strength is larger than the required minimum strength, i.e. the cutoff.

I've updated the argument parsing to refuse any negative cutoff values to enforce this logic and made the documentation more explicit.

PR #4869, opened 2026-02-02
Conservative APIP

d-immel:
**Summary**

This enhancement of the APIP package allows the computation of conservative AP potentials according to Immel, Drautz and Sutmann arXiv:2512.07693.

**Related Issue(s)**

None.

**Author(s)**

David Immel¹³, Ralf Drautz², Godehard Sutmann¹²

1: Jülich Supercomputing Centre (JSC), Forschungszentrum Jülich, Jülich, Germany
2: Interdisciplinary Centre for Advanced Materials Simulations (ICAMS), Ruhr Universität Bochum, Bochum, Germany
3: [d.immel@fz-juelich.de](mailto:d.immel@fz-juelich.de)

**Licensing**

**Backward Compatibility**

This changes do not break backward compatibility.

**Implementation Notes**

The conservative potential is computed using existing APIP pair styles, but a new fix to compute a differentiable switching parameter and the force caused by the differentiation of this switching parameter. The switching parameter is calculated using a local average (la). It has been verified that a conservative AP potential conserves the energy just as well as a constant precision EAM or ACE potential (cf. examples/PACKAGES/apip/in.surface.la.cons).

**Post Submission Checklist**

Immel, Drautz and Sutmann arXiv:2512.07693.
https://doi.org/10.48550/arXiv.2512.07693


akohlmey:
@d-immel I have one high-level question. Why do we need a new atom style? Why can't you just extend the existing one or use fix property/atom for the additional per-atom properties?

We would like to keep the addition of new atom styles to the minimum, since they require changes (and maintenance work!!) in the core code and there are already too many of those.

d-immel:
> @d-immel I have one high-level question. Why do we need a new atom style? Why can't you just extend the existing one or use fix property/atom for the additional per-atom properties?
> 
> We would like to keep the addition of new atom styles to the minimum, since they require changes (and maintenance work!!) in the core code and there are already too many of those.

@akohlmey I can extend the existing one. I wasn't aware that one can pass arguments to an atom style.

doc/src/fix_lambda_la_csp_apip.rst
```diff
@@ -0,0 +1,198 @@
...
+parameters.
+
+.. note::
+
+   The fast potential and the precise potential are combined via
+   :doc:`pair_style hybrid/overlay <pair_hybrid>` as shown in the code
+   example below.
+
+The original CSP :ref:`(Kelchner) <Kelchner_3>`
+is used instead of the differentiable CSP with the option *csp_mode dynamic*.
```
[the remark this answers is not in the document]
d-immel: The copilot suggestion is wrong.
csp_mode static = differentiable CSP = default
csp_mode dynamic = non-differentiable CSP

src/APIP/fix_lambda_la_csp_apip.cpp
```diff
@@ -0,0 +1,1344 @@
...
+      if (jj == -1) {
+        printf("own atom i % x %f %f %f\n", i, xtmp, ytmp, ztmp);
+        printf("nlocal %i nghost %i nall %i\n", atom->nlocal, atom->nghost,
+               atom->nlocal + atom->nghost);
+        for (jj = atom->map((tagint) stored_tags[i][j]); jj >= 0; jj = atom->sametag[jj])
+          printf("possible pair ngh jj %i x %f %f %f\n", jj, x[jj][0], x[jj][1], x[jj][2]);
+        for (jj = atom->nlocal; jj < atom->nlocal + atom->nghost; jj++) {
+          if (atom->tag[jj] == (tagint) stored_tags[i][j])
+            printf("atom with same tag jj %i x %f %f %f\n", jj, x[jj][0], x[jj][1], x[jj][2]);
+        }
```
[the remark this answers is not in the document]
d-immel: I've removed the debug output.

src/APIP/fix_lambda_la_csp_apip.cpp
```diff
@@ -0,0 +1,1344 @@
...
+            fabs(delz_j) < domain->prd_half[2]) {
+          // correct periodic image
+          break;
+        } else {
+          // next periodic image
+          jj = atom->sametag[jj];
+        }
+      }
+      if (jj == -1) {
+        printf("own atom i % x %f %f %f\n", i, xtmp, ytmp, ztmp);
```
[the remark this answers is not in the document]
d-immel: I've removed the debug output, including this printf.

src/APIP/fix_lambda_la_csp_apip.cpp
```diff
@@ -0,0 +1,1344 @@
...
+      if (kk == -1) {
+        printf("own atom i %i x %f %f %f\n", i, xtmp, ytmp, ztmp);
+        printf("nlocal %i nghost %i nall %i\n", atom->nlocal, atom->nghost,
+               atom->nlocal + atom->nghost);
+        for (kk = atom->map((tagint) stored_tags[i][j + nhalf]); kk >= 0; kk = atom->sametag[kk])
+          printf("possible pair ngh kk %i x %f %f %f\n", kk, x[kk][0], x[kk][1], x[kk][2]);
+        for (kk = atom->nlocal; kk < atom->nlocal + atom->nghost; kk++) {
+          if (atom->tag[kk] == (tagint) stored_tags[i][j + nhalf])
+            printf("atom with same tag kk %i x %f %f %f\n", kk, x[kk][0], x[kk][1], x[kk][2]);
+        }
```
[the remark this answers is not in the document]
d-immel: I've removed the debugging code.

doc/src/compute_property_atom.rst
```diff
@@ -78,6 +79,9 @@ Syntax
            *apip_lambda* = switching parameter
            *apip_lambda_input* = input used to calculate the switching parameter
            *apip_e_fast,apip_e_precise* = potential energies mixed by the adaptive-precision potential
+           *apip_la_inp* = input used for the local averaging of the descriptor
+           *apip_la_norm* = locally averaged radial weighting function used for normalization
+           *apip_la_avg* = locally averaged descriptor used to calculate the switching parameter
```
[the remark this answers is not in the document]
d-immel: I've added ".. versionchanged:: TBD" to the documentation.

doc/src/atom_style.rst
```diff
@@ -121,6 +121,10 @@ the Additional Information section below.
      - *atomic* + apip_lambda, apip_lambda_required, apip_lambda_input, apip_lambda_const, apip_lambda_input_ta, apip_e_fast, apip_e_precise, apip_f_const_lambda, apip_f_dyn_lambda
      - :ref:`APIP <PKG-APIP>`
      - adaptive-precision interatomic potentials(APIP), see :doc:`APIP howto <Howto_apip>`
+   * - *apip/la*
+     - *atomic* + apip_lambda, apip_lambda_required, apip_la_inp, apip_la_avg, apip_la_norm, apip_e_fast, apip_e_precise
+     - :ref:`APIP <PKG-APIP>`
+     - conservative adaptive-precision interatomic potentials(APIP), see :doc:`APIP howto <Howto_apip>`
```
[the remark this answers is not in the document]
d-immel: I've added ".. versionchanged:: TBD" to the documentation of the extended apip pair style.

doc/src/fix_lambda_la_csp_apip.rst
```diff
@@ -0,0 +1,198 @@
...
+
+**(Immel2025)** Immel, Drautz and Sutmann, J Chem Phys, 162, 114119 (2025)
+
+.. _Immel2026_1:
+
+**(Immel2026)** Immel, Drautz and Sutmann, arXiv:2512.07693
+
+.. _Kelchner_3:
+
+**(Kelchner)** Kelchner, Plimpton, Hamilton, Phys Rev B, 58, 11085 (1998)
```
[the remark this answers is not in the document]
d-immel: I've added ".. versionadded:: TBD" to the documentation.

d-immel:
@akohlmey I've extended the existing apip pair style and addressed the issues found by copilot.

akohlmey:
> @akohlmey I've extended the existing apip pair style and addressed the issues found by copilot.

@d-immel Thanks, now we need to wait for some other LAMMPS developer doing a proper code review. I've flagged a few candidates. I currently don't have the capacity for anything but high-level questions and Copilot is only good for simple things and gets easily confused by the LAMMPS code architecture and programming style.

doc/src/atom_style.rst
jmgoff, line 1: following [Axel's comment](https://github.com/lammps/lammps/pull/4869#issuecomment-3835643088), I am concerned about changes to atom style. If extra per atom properties are needed, developers should consider `fix pair` before adding atom styles.

doc/src/Howto_apip.rst
```diff
@@ -65,14 +66,41 @@ potential is explained in :ref:`here <implementing_new_apip_styles>`.
...
+        #. A fast potential: :doc:`eam/apip <pair_eam_apip>` or :doc:`pace/fast/apip <pair_pace_apip>`.
+        #. A precise potential: :doc:`pace/precise/apip <pair_pace_apip>`.
+        #. :doc:`fix lambda/la/csp/apip <fix_lambda_la_csp_apip>` to calculate a differentiable switching parameter :math:`\lambda_i`.
+        #. :doc:`pair_style hybrid/overlay <pair_hybrid>` to combine the previously mentioned pair_styles.
+        #. :doc:`fix atom_weight/apip <fix_atom_weight_apip>` to approximate the load caused by every atom, as the computations of the pair_styles are only required for a subset of atoms.
+        #. :doc:`fix balance <fix_balance>` to perform dynamic load balancing with the calculated load.
+
+   .. tab:: dynamic switching parameter with correction
 
         #. :doc:`atom_style apip <atom_style>` so that the switching parameter :math:`\lambda_i` can be stored.
```
jmgoff, line 105: Can this switching parameter be handled in `fix pair` instead? An extra atom style, like Axel said, bloats ongoing maintenance.

src/atom.cpp
jmgoff, line 1: I would suggest that all per-atom apip variables be handled by `fix pair`. This will require a bit of restructuring of the APIP package.

src/compute_property_atom.cpp
jmgoff, line 339: I would also consider handling these old properties in `fix pair`.

src/fix.cpp
```diff
@@ -292,6 +292,7 @@ void Fix::v_setup(int vflag)
   int i,n;
 
   evflag = 1;
+  vflag_either = vflag;
```
jmgoff: Was `vflag_either` incorrect before? I would double check with @akohlmey or @athomps about whether or not changes to `fix.cpp` are necessary.

src/APIP/pair_eam_apip.cpp
jmgoff, line 1: The pair_style looks good from the technical perspective, but I think the implementation could be done without relying on a new `atom_style`.

src/APIP/fix_lambda_la_csp_apip.cpp
```diff
@@ -0,0 +1,1327 @@
...
+  if (count > 1) error->all(FLERR, "More than one fix lambda/la/csp/apip.");
+
+  if (force->pair->cutforce < cut_hi)
+    error->all(FLERR, "cutoff of potential ({}) smaller than cutoff of weighting function ({})",
+               force->pair->cutforce, cut_hi);
+  if (force->pair->cutforce * force->pair->cutforce < csp_cutsq)
+    error->all(FLERR, "cutoff of potential ({}) smaller than cutoff of the CSP ({})",
+               force->pair->cutforce, sqrt(csp_cutsq));
+
+  if (strcmp(atom->atom_style, "apip"))
```
jmgoff, line 278: Atom style to remove.

src/fix.cpp
akohlmey: @jmgoff @d-immel This change is *definitely* incorrect. Fixes may only contribute to the virial if so configured. Typically this is turned off and must be enabled with `fix_modify <fix-ID> virial yes`. This default can be changed in the individual fix by setting the default value of `thermo_virial` to 1, as is done for example for `fix rigid`, `fix shake`, and others.

The same goes for energy and `thermo_energy`. Changing the global default will break existing functionality.

src/APIP/fix_lambda_la_csp_apip.cpp
```diff
@@ -0,0 +1,1327 @@
...
+  double xtmp, ytmp, ztmp, delx, dely, delz, rsq, value, weight;
+  double *lambda, *csp, *csp_avg, *csp_norm;
+
+  tagint *tag = atom->tag;
+
+  mask = atom->mask;
+  lambda = atom->apip_lambda;
+  csp = atom->apip_la_inp;
+  csp_avg = atom->apip_la_avg;
+  csp_norm = atom->apip_la_norm;
```
jmgoff: find `FixPair`  by fix index and extract per atom properties from there instead.
``

```suggestion
   int ifix = modify->find_fix(fix_pair_id);
   if (ifix < 0) error->all(FLERR, "Target fix pair not found");
   FixPair *fix_lmb = (FixPair *)modify->fix[ifix]
   lambda = ...
   csp = ...
```
see `FixPair` for how to extract per-atom property from pair style within `fix pair`.

src/atom.cpp
```diff
@@ -581,6 +582,9 @@ void Atom::peratom_create()
   add_peratom("apip_lambda_const",&apip_lambda_const,DOUBLE,0);
   add_peratom("apip_f_const_lambda",&apip_f_const_lambda,DOUBLE,3,1);
   add_peratom("apip_f_dyn_lambda",&apip_f_dyn_lambda,DOUBLE,3,1);
+  add_peratom("apip_la_inp",&apip_la_inp,DOUBLE,0);
+  add_peratom("apip_la_avg",&apip_la_avg,DOUBLE,0);
+  add_peratom("apip_la_norm",&apip_la_norm,DOUBLE,0);
```
jmgoff:
```suggestion
```
As much as possible, I would avoid adding new per-atom properties this high up in LAMMPS.

src/APIP/fix_lambda_la_csp_apip.cpp
akohlmey: @jmgoff This is a deprecated use of an outdated API. You should get a pointer to a fix by fix-ID and then cast to a specific fix class rather with with something like this:
```
auto *myfix = dynamic_cast<FixPair *>(modify->get_fix_by_id(fix_pair_id);
if (!myfix)
  error->all(FLERR, "Fix with fix-ID {} not found or incorrect style", fix_pair_id);
```

doc/src/atom_style.rst
akohlmey: I fear there is some misunderstanding here. The purpose of fix pair is to make per-atom properties computed *inside* a pair style and stored in memory that is managed by the pair style accessible to custom dump styles or atom style variables etc.
This kind of data is typically cleared and reset every time the `Pair::compute()` function is called and does not need to be restarted or carried over between timesteps.

Other data can be either stored by a custom fix, fix property/atom, or an atom style. We already *had* an atom style "apip" so adding to it is not a big deal and the functionality of `fix pair` is in that case taken over by `compute property/atom`. The difference between using `fix property/atom` and an atom style is in the different complexity of reading per-atom data from data files and restart files. This is much easier and cleaner for an atom style, thus `fix property/atom` is best used for properties that do not need to be read from data files. It saves the effort of managing the data over having the data managed by the pair style.

Of course the, final word on these kinds of fundamental decisions lies always with @sjplimp
From my perspective, however, the worst issue, i.e. adding a *second* APIP atom style is already avoided and thus adding properties to the existing one is only a minor concern.

src/fix.cpp
jmgoff:
```suggestion
```
As Axel mentioned, not correct. If the code relies on this change, then something is off. The virial calculations probably need to be revisited.

jmgoff (changes requested):
I've left sporadic comments and suggested changes, but pretty much all of them relate to the same thing. The per-atom properties used by apip should not be treated with an atom style.

I understand that accessing the extra per-atom properties through atom styles can be convenient, but they do face extra scrutiny for many reasons including those Axel mentioned. From my end, I am concerned that the atom style is mostly tailored for apip and a small number of adjacent methods. As-is, this atom style won't be useful or strictly necessary for many other fixes and pair_styles unlike other atom styles (e.g.  atom_style charge). 

The two ways i could see this being resolved would be:
1. Restructuring the apip code and accessing the necessary per-atom properties through `fix pair` or `fix property/atom` (recommended)
2. Develop and use a more general atom style that is useful for apip as well as many other methods and fixes. (not recommended but _maybe_ possible)

I mention this with the caveat that other reviewers may just say no to this, but the latter approach _may_ be possible.  It would still face heavy scrutiny because of changes to high-level LAMMPS code. I think it may be possible possible if there are compelling arguments for the proposed atom style being useful/necessary for apip, many existing methods, and new methods. There would be changes elsewhere in LAMMPS that would need to be made in this PR if this approach were taken. This would likely take a lot of work and discussion with other reviewers and devs before it would be considered. Before I would try this myself, I would try to talk a lot more with reviewers and other devs to see if something like a new general atom style is useful or necessary. Given the time, effort, and buy-in from others needed for this approach, I would suggest modifying apip and associated methods to use `fix pair` and/or `fix property/atom`.

Other than this, there are a couple other changes that will likely need to be made to things virial calculations, but the way these problems are showing up now may change a bit after removing dependence on the atom style. Otherwise, I think the models and updates present some important advances. Documentation and examples are thorough. I'm excited to see conservative APIP in LAMMPS after the changes are made.

src/APIP/fix_lambda_la_csp_apip.cpp
jmgoff: Good point. For @d-immel, dont use the deprecated API like I did in my example. You should be able to access `FixPair` or `FixPropertyAtom` using the updated syntax Axel has here.

src/fix.cpp
d-immel: @jmgoff @akohlmey I'm aware of the default behaviour and change `thermo_virial` to 1 in the new fix.

To my understanding, my suggested change does not affect the global default.

I've changed `Fix::v_setup(int vflag)` in fix.cpp, which is only called under certain conditions by `v_init(int vflag)` in fix.h:
`if (vflag && thermo_virial) v_setup(vflag);`
`else veflag = vflag_either = [...] = 0;`
Currently, `Fix::v_setup(int vflag)` does not affect `vflag_either`.
I thought that `vflag_either` should be initialised independent of the condition.
I suggest the initialisation of `vflag_either` to `vflag` like it is done in `Fix::ev_setup(int eflag, int vflag)`.

d-immel: The code does not rely on this change. I use `vflag_either`, but I could use `evflag` instead.

doc/src/atom_style.rst
d-immel: I probably do not need to add more properties to the existing APIP atom style.
I am working on using FixStoreAtom within the new APIP fix.

jmgoff: I did think that some of these properties were handled/computed within the pair_style (or at least that some of these properties could be with some refactoring of APIP code). 

I understand that the `apip` atom style is an existing atom style, but my concern about it being tailored specifically for apip remains. Auxiliary per-atom properties like this may be needed in other ML methods. I would be interested in helping with a refactor in the future (not this PR) that replaces `atom_style apip` with something that could be used by other methods with auxiliary per-atom properties (something like `atom style ml`). This way in the future, if another method needs per-atom properties in the same way, they can all make it work with a general atom style rather than modifying `atom_style apip` or adding a new atom style.

This is not a problem for @d-immel and the developers on this PR though, I will remove my requested changes related to this.

src/APIP/pair_eam_apip.cpp
jmgoff: [see comment other comment](https://github.com/lammps/lammps/pull/4869#discussion_r2793639886)

src/APIP/fix_lambda_la_csp_apip.cpp
jmgoff: [See other comment.](https://github.com/lammps/lammps/pull/4869#discussion_r2793639886)

src/APIP/fix_lambda_la_csp_apip.cpp
jmgoff: [See other comment.](https://github.com/lammps/lammps/pull/4869#discussion_r2793639886)

src/atom.cpp
jmgoff: [See other comment.](https://github.com/lammps/lammps/pull/4869#discussion_r2793639886)

src/fix.cpp
d-immel: I've used `evflag` instead.

d-immel:
I've revised the new `fix lambda/la/csp/apip` so there is no need for new global per-atom variables in `atom_style apip`.

Furthermore, I've reverted the change in `Fix::v_setup(int vflag)` and use `evflag` instead of the potentially uninitialised `vflag_either` in the new fix.

jmgoff (approved):
Thank you for the updates and cleaning up the additions to the high-level code. It is nice to have fewer global per-atom variables in atom_style apip.
I can look into a refactor later for a more general atom type as discussed in other comments.

PR #4879, opened 2026-02-09
Bond/react: wildcard and infer interaction updates

jrgissing:
**Summary**

Fixes some issues related to non-symmetric higher-order interactions when inferring interaction types and using wildcards with 'fix bond/react'

Also adds an option to check the self-consistency of types labels read in from molecule template, if using the standard hyphen-delimited format

TODO:

- [x] Docs
- [x] Fix broken unit tests (or reimplement in non-breaking way)
- [ ] Option to test self-consistency of simulation type labels (?)

**Related Issue(s)**

None

**Author(s)**

Jake

**Licensing**

**Backward Compatibility**

NOTE: modifying behavior of infer_* routines output is a silent breaking change. should be fine since this function was committed two weeks ago

**Implementation Notes**

**Post Submission Checklist**


jrgissing:
@akohlmey Do we already have a mechanism to perform runtime (just before a 'run' command) error checks from a command? If not, would it be reasonable to add a virtual routine to do so in the command class? I am trying to add a check for self-consistent type labels (new labelmap 'check_labels' keyword in this PR), which needs ghost atoms to have been communicated.

akohlmey:
> @akohlmey Do we already have a mechanism to perform runtime (just before a 'run' command) error checks from a command? If not, would it be reasonable to add a virtual routine to do so in the command class? I am trying to add a check for self-consistent type labels (new labelmap 'check_labels' keyword in this PR), which needs ghost atoms to have been communicated.

This is non-trivial and conceptually problematic. Commands are "run-and-done" and not all commands are even command styles, but functions called from the `Input` class.
How about you create an internal fix (fix style name all upper case) that is added if it doesn't exist already. Then the fix would be called during `Modify::init()`. if you give it a unique fix-ID, e.g. LABELMAP_PRE_RUN_CHECK, you can use `Modify::get_fix_by_id()` to see, if the fix already exists and create it (or recreate it), only if necessary. That seems to me to have the same effect without having to change anything. It may be a bit challenging to make this work with restarting or data files, but probably, the read_restart and read_data commands can be adjusted to check for it or the whole function for creating the fix can be delegated to be a function in the `Input` class that is called from either command.

jrgissing:
Would you be opposed to me adding one line in Modify::init(), at the bottom where other runtime checks are performed, to avoid creating another fix?

akohlmey:
> Would you be opposed to me adding one line in Modify::init(), at the bottom where other runtime checks are performed, to avoid creating another fix?

That is a question for @sjplimp. He has the final word on these kinds of changes.

athomps (approved):
I approve.

akohlmey (approved):

PR #4887, opened 2026-02-15

### LINKED ISSUE #4855: Inconsistent behavior of the GPU-accelerated CORE-SHELL model
**Summary**

The bug fixes of #4764 produce only a partial solution to my initial complaints #4712 in the sense that the GPU accelerated version now produces "acceptable looking" results. However, a closer examination reveals that these results are not consistent with the stable non-GPU version and that they strongly depend on the cutoff specified in the `pair_style` declaration, which is not the expected behavior.

LAMMPS (22 Jul 2025 - Update 1)
Apple M4 Pro (24Gb) Mac OS 15.6.1 Sequoia

**Expected Behavior**

Consistent results between the non-GPU and GPU variants, reasonably independent of the `pair_style` cutoff.

**Actual Behavior**
The present version of the `pair_style born/coul/long/cs/gpu` resulting from #4764 radically eliminates the same-atom contributions from the Coulomb sum, which permits to avoid issues with single precision arithmetics. This step is reproduced in my (more explicit) way in the attached versions of the corresponding code, leading to numerically identical results. 

It appears, however, that in this way the overall force balance gets shifted/perturbed, which shows up in numerical disagreement with the original `core-shell` implementation as well as in altered and most probably incorrect physics results like lattice parameter variation and phase transitions temperatures. The phase transitions temperatures decreasing with the cutoff value may hint towards the pair interaction not being properly combined with the long range `kspace` component.

While differences in atomic coordinates, velocities and forces are visible from the very beginning of the MD trajectories, the altered lattice parameters and phase transition temperatures only manifest themselves after a few `ps` runtime on a large enough box. Thus the choice of "long runs" in the present attachments.

Cf. attachments.

**Steps to Reproduce**

Cf. attachment.

[GPU_long_run_report.zip](https://github.com/user-attachments/files/24864084/GPU_long_run_report.zip)

### LINKED ISSUE #4890: [BUG] Error in fix_pair when trying to use sub pair styles from hybrid/scaled
**Summary**
I was running a simulation to see how "pair_style hybrid/scaled" worked with "fix pair" using a simple Al eam potential. The simulation crashed with an error message when trying to execute the run command. I also tested this with an Al eam/alloy potential, and the same thing occurred. 

LAMMPS version: 22 Jul 2025 
Platform: Chicoma IC Cluster, OS: HPE Cray Shasta, Hardware: AMD Rome EPYC 7H12

**Expected Behavior**
The expected behavior was that when I use the "pair_style hybrid" command was that "it can be a sub-style used in a [pair_style hybrid](https://docs.lammps.org/pair_hybrid.html) command. If there are multiple sub-styles using the same pair style, then pstyle should be specified as “style:N”, where N is the number of the instance of the pair style you wish monitor (e.g., the first or second). For example, pstyle could be specified as “pace/extrapolation” or “amoeba” or “eam:1” or “eam:2”." 

**Actual Behavior**

The actual behavior was that when I used the following commands:
- pair_style hybrid/scaled 0.7 eam 0.3 eam
- pair_coeff * * eam 1 Al_jnp.eam
- pair_coeff * * eam 2 Al_jnp.eam
- fix rho_peratom1 all pair 1 eam:1 rho 0
- fix rho_peratom2 all pair 1 eam:2 rho 0

I got the following error message when LAMMPS tried to execute the run command:
- ERROR: Pair style eam for fix pair not found (src/fix_pair.cpp:199)

**Steps to Reproduce**
I have attached the file "in.fix_pair_hybrid_scaled_bug_eam" which is the minimal input deck I used to generate this error. I have also attached the "Al_jnp.eam" file which is the specific Al eam potential I used in the minimal input deck. 

Once you remove the ".txt" from both file names, it should be able to run with little effort and minimal computational resources. MPI and Open_MP are not required to test this bug that I am seeing, but I did compile my version of LAMMPS with MPI and Open_MP.

[in.fix_pair_hybrid_scaled_bug_eam.txt](https://github.com/user-attachments/files/25396214/in.fix_pair_hybrid_scaled_bug_eam.txt)
[Al_jnp.eam.txt](https://github.com/user-attachments/files/25396251/Al_jnp.eam.txt)

akohlmey:
**Summary**

**Related Issue(s)**

Fixes #4890
Closes #4855 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

This pull request contains mostly changes to address issues reported by static code analysis from Coverity Scan and bug reports.  Coverity had been unavailable for several weeks and thus we have to catch up right now.

In addition, there are some programming style updates to make code more consistent with the rest of LAMMPS and the usual error message and error handling improvements.

Also, pair style `born/coul/long/cs/gpu` stops with an error when the GPU package is compiled in single precision mode due to unreliable results.

Finally, throughout the documentation there are changes that replace the term "processors" with "MPI processes" or similar where applicable to avoid ambiguities when running with multiple threads per MPI rank.

**Post Submission Checklist**


akohlmey:
@jtclemm FYI, any additional fixes or improvements to the recent fix nvt/sllod changes are best added to this pull request. In case you are running any tests, it is probably best to start with this branch as well.

jtclemm (approved):

ndtrung81 (approved):

PR #4889, opened 2026-02-18
MBX package small changes 3

Miniland1333:
**Summary**

 - Implemented small changes recommended previously by @jtclemm 
 - Support for compiling MBX package for Windows with Linux-to-Windows cross compiler including support for creating a plugin installer package

**Author(s)**

Henry Agnew, UC San Diego and Axel Kohlmeyer, Temple U

**Licensing**

**Implementation Notes**

For the small changes suggested by @jtclemm, I implemented them and then ensured that the example output is the same as the expected logfiles.

**Post Submission Checklist**

Continuation of https://github.com/lammps/lammps/pull/4694


Miniland1333:
> I would like to revisit the portability to Windows. At least with the Linux2Windows cross-compiler it should work. For native compilation with MSVC++ there is next to no chance because you are using autoconf. Cygwin or MinGW might work, but we don't test it anymore.

For Windows, this is what was previously mentioned by @akohlmey.

akohlmey:
> For Windows, this is what was previously mentioned by @akohlmey.

@Miniland1333 I have indeed been able to compile LAMMPS with the MBX package included for Windows using the MinGW64 Linux-to-Windows cross-compiler toolkit available on Fedora Linux. But it needs some hacks. In part these are required to overcome limitations of the Fedora toolkit customizations, in part these are due to limitations of the MBX package itself.

As for the latter, there is the use of the `M_PI` macro, which is non-portable and not covered by any standard. C++20 offers these https://cppreference.com/w/cpp/numeric/constants.html, but LAMMPS only recently moved up from requiring C++11 to C++17, so this is a long way away from us.
This is why for LAMMPS we decided a long time ago to have our own pool of mathematical constants in the `math_const.h` header, which has the added benefit that those are consistent across platforms and compilers. After some experimentation, I opted to have this definition added to CPPFLAGS with a hack to the `configure` script, that is only triggered when using the cross-compiler.

Another issue is the use of `sysconf()` which is a POSIX feature and thus also not directly available on Windows platforms. Here I opted to use conditional compilation in the code and just hardcoded a suitable constant.

The final issue is the compilation of the `I-pi` plugin where I hacked the Makefile to not include that subdirectory. It would be nice to have a option to `configure` to disable this.

It still won't work to compile MBX natively on Windows, especially not with MSVC++ because of the use of the GNU autotools, which require a Unix-like shell environment.

I probably have a few more (CMake?) changes to make to be able to compile a Windows installer for a plugin version of the package so it can be distributed without violating any of the licenses of MBX or LAMMPS.

Independent of this, I replaced the one use of `std::fill()` with a call to `memset()` to silence some nasty and misleading compiler warnings.

Update: I have submitted a pull request upstream to have what is currently a patch become part of the MBX distribution and we don't have to patch it after downloading anymore.

src/MBX/fix_mbx.cpp
```diff
@@ -35,6 +35,10 @@
 #include <cstring>
 #include <mpi.h>
 
+#if !defined(M_PI)
+#define M_PI 3.14159265358979323846
```
jtclemm: Why not use MY_PI in math_const.h?

akohlmey: It is needed for code included from the MBX library.

jtclemm: Got it

akohlmey: Otherwise it does not compile with MinGW64. A similar hack is needed for MBX itself, hence the same hack is in the patch file for compiling MBX https://github.com/lammps/lammps/pull/4889/changes#diff-4b6b0f12198cfe9ea20e12000c71df61bbc90f1ee40c290a1e2b49882463ecc1R27

I've also submitted a pull request upstream, but that won't help for LAMMPS. :disappointed: 
https://github.com/paesanilab/MBX/pull/104/changes

jtclemm (approved):

stanmoore1 (approved):

akohlmey:
@Miniland1333 can you give me a progress report and possibly an ETA of the changes in the MBX repo related to the patch in this pull request?
Basically, I would like to figure out whether we should merge this pull request or wait until the next update of the MBX library.

Please let me know at your convenience.

Miniland1333:
@akohlmey let's probably draft this PR for now and wait for the next minor release of MBX. I have an internal MBX PR that will remove dependency on `M_PI`, and I want to strengthen some of our tests to confirm my changes are good.

akohlmey:
> @akohlmey let's probably draft this PR for now and wait for the next minor release of MBX. I have an internal MBX PR that will remove dependency on `M_PI`, and I want to strengthen some of our tests to confirm my changes are good.

@Miniland1333 can you please provide a status update? I am considering making a new feature release next week and would like to include the porting to Windows so I can prepare plugin packages for Windows.

Miniland1333:
Hello everyone! I have finally released a new public version of MBX v1.3.12, which includes the replacement of the `M_PI` macro with a constant. I have updated the tarball and removed the parts of the PR relating to `M_PI`.

While for the most part we should be good to go, please let me know if you happen to get an out-of-memory error while compiling the MBX plugin. The new `1b_A1B1C1D3E1F1G1H3` (nma) potential is large, and thus may cause compilation memory issues if you have limited memory.

akohlmey:
> and thus may cause compilation memory issues if you have limited memory.

No problems so far. I needed to update the CMake scripting and get rid of the patch because of the updated configure script. Will make more compilation tests in the next days while preparing for the next feature release. Any required updates will be in a separate pull request.

PR #4897, opened 2026-02-26
Add granular superellipsoids DEM support

JBil8:
**Summary**

The ASPHERE package enables simulation of non-spherical particles, but these particle shapes cannot be used in granular simulation because of the lack of a contact detection algorithm and contact forces for ellipsoidal particles in the GRANULAR package.
This PR adds such contact capabilities to the GRANULAR package, and enables granular simulation of super-ellipsoidal particles.
The super-ellipsoids particles use the `atom_style ellipsoid`, and generalize it by adding 2 "blockiness" exponents (variable `block`) in the `Bonus` struct.
These exponents default to a value of 2 (ellipsoid) and slight modifications of the `set`, `read_data`, `write_data`, make this feature backward compatible. Incidentally, code duplication in the calculation of ellipsoid properties (e.g., inertia) has been eliminated from the ASPHERE, CG-DNA, RIGID, and SRD packages.
The atom_style now defines the `radius` pointer and `radius_flag`, and the encompassing radius of the particle is used to build the neighbor list from existing LAMMPS machinery. We discussed this with @jtclemm and this should solve issue #3999.
To minimize disruption to the `pair_style granular` which is currently designed for spheres, we implemented a Hooke and Hertz contact law in the old style (standalone) of `pair/gran`. Future work will involve refactoring `granular` and build adequate subclasses to integrate this into the modern `granular` styles, with possibility to include more complex contact models.

Contact detection for ellipsoids requires multivariate minimization. Newton's method is used, which involves building residual vectors, Jacobian matrices and solving many 4x4 systems of equations. There is a lot of math defined in a `MathExtraSuperellipsoid` utility. This can use LAPACK/linalg, but for a 4x4 system, we've experienced very large overhead and could get much better performance with a hand-written dedicated 4x4 solver. Please let us know what is your preferred way of dealing with small-system linear algebra in LAMMPS. 
In the case the minimization does not reach the desired tolerance, the current behaviour is lammps throwing a warning rather than an error. For most tests with reasonable shapes and initial condition, this seems to be the case very rarely and it does not lead to divergence in the simulation.
Two formulations are implemented for the minimization problem, on with and one without preconditioning.
The user can decide which one to use at runtime. A bounding box check can be used before solving the newton-rapson loop to reduce computational time.

A few examples on how to run the code (with ASPHERE and GRANULAR packages enabled) can be found in examples/ASPHERE/superellipsoid_gran, the documentation of the pair style in doc/src/pair_gran_ellipsoid.rst, some tests in unittest/utils/test_math_extra_superellipsoids.cpp.

**Related Issue(s)**

Related to #3999 (https://github.com/lammps/lammps/issues/3999)

**Author(s)**

Jibril B. Coulibaly 
Jacopo Bilotto (EPFL) jacopo.bilotto@epfl.ch

**Licensing**

JB used "gemini 3.0 pro" to clean and refactor parts of the code added with this PR.
JB used "gemini 3.0 pro" was used to generate some of the input scripts made available with this package, later edited by the author.

**Backward Compatibility**

Compatible

**Post Submission Checklist**


doc/src/Build_settings.rst
```diff
@@ -12,6 +12,7 @@ explains how to do this for building both with CMake and make.
 * `FFT library`_ for use with the :doc:`kspace_style pppm <kspace_style>` command
 * `Size of LAMMPS integer types and size limits`_
 * `Read or write compressed files`_
+* `Support for JPEG, PNG, and movie files`_ in the :doc:`dump image <dump_image>` or :doc:`dump movie <dump_image>` or :doc:`fix graphics/labels <fix_graphics_labels>` commands
```
akohlmey: This is re-introducing text that was moved to a different part of the documentation, since this is only relevant for features of the (new) GRAPHICS package.

doc/src/Build_settings.rst
```diff
@@ -383,6 +384,77 @@ in whichever ``lib/gpu/Makefile`` is used must be the same as above.
 
 ----------
 
+.. _graphics:
```
akohlmey: This is in `doc/src/Build_extras.rst` now.

doc/src/Commands_fix.rst
```diff
@@ -84,9 +84,7 @@ OPT.
    * :doc:`gld <fix_gld>`
    * :doc:`gle <fix_gle>`
    * :doc:`graphics/arrows <fix_graphics_arrows>`
-   * :doc:`graphics/isosurface <fix_graphics_isosurface>`
```
akohlmey, line 87: This is incorrectly removing lines.

doc/src/dump_image.rst
```diff
@@ -56,7 +56,7 @@ Syntax
          level = mesh refinement level, value between 1 (low resolution) and 6 (ultra high resolution)
          width = diameter of wireframe edges (distance units) (ignored for triangles)
        *body* = color bflag1 bflag2
-         color = *type* or *index*
```
akohlmey: This is incorrectly reverting changes.

doc/src/dump_image.rst
```diff
@@ -222,13 +222,13 @@ Examples
 Description
 """""""""""
 
-Dump a high-quality rendered image of the atom configuration every
```
akohlmey, line 227: This pull request does not concern dump image and thus there should be no changes in parts of LAMMPS that have nothing to do with the actual changes.

doc/src/fix.rst
```diff
@@ -263,9 +263,7 @@ accelerated styles exist.
 * :doc:`gld <fix_gld>` - generalized Langevin dynamics integrator
 * :doc:`gle <fix_gle>` - generalized Langevin equation thermostat
 * :doc:`graphics/arrows <fix_graphics_arrows>` - add arrow graphics objects to :doc:`dump image <dump_image>` output
-* :doc:`graphics/isosurface <fix_graphics_isosurface>` - add an isosurface for a group of atoms to :doc:`dump image <dump_image>` output
```
akohlmey, line 266: ditto

doc/src/fix_graphics_labels.rst
```diff
@@ -35,7 +35,7 @@ Syntax
         x, y, z  = position where the center of the text is located in the visualization
         any of x, y, or z can be a variable (see below)
 
-        keyword = *fontcolor* or *framecolor* or *backcolor* or *transcolor* or *size* or *horizontal* or *vertical*
```
akohlmey, line 38: ... and again

doc/src/fix_graphics_objects.rst
```diff
@@ -14,7 +14,7 @@ Syntax
 * graphics/objects = style name of this fix command
 * Nevery = update graphics information every this many time steps
 * one or more keyword/args pairs may be appended
-* keyword = *sphere* or *cylinder* or *arrow* or *cone* or *progbar*
```
akohlmey, line 17: ditto

doc/src/fix_graphics_periodic.rst
```diff
@@ -8,17 +8,17 @@ Syntax
 
 .. code-block:: LAMMPS
 
-   fix ID group-ID graphics/periodic Nevery keyword args ...
```
akohlmey, line 11: ditto

doc/src/Howto_viz.rst
```diff
@@ -5,7 +5,7 @@ Snapshots from LAMMPS simulations can be viewed, visualized, and
 analyzed in a variety of ways.
 
 LAMMPS snapshots are created by the :doc:`dump <dump>` command, which
-can create files in several formats.  The native LAMMPS dump format is a
```
akohlmey, line 8: ditto

doc/src/read_data.rst
```diff
@@ -1327,19 +1327,20 @@ and a general discussion of how type labels can be used.
 *Ellipsoids* section:
 
 * one line per ellipsoid
-* line syntax: atom-ID shapex shapey shapez quatw quati quatj quatk
+* line syntax: atom-ID shapex shapey shapez quatw quati quatj quatk block1 block2
```
akohlmey: This seems to be breaking backward compatibility or if not, it should be **clearly** indicated in the documentation that this is optional and examples be given for the case with and without

src/GRAPHICS/dump_image.cpp
```diff
@@ -294,11 +294,7 @@ DumpImage::DumpImage(LAMMPS *lmp, int narg, char **arg) :
       if (iarg+4 > narg) utils::missing_cmd_args(FLERR,"dump image body", error);
       bodyflag = YES;
       if (strcmp(arg[iarg+1],"type") == 0) bodycolor = TYPE;
-      else if (strcmp(arg[iarg+1],"index") == 0) bodycolor = INDEX;
```
akohlmey, line 338: more changes that are incorrectly reverting code from upstream

unittest/formats/test_atom_styles.cpp
```diff
@@ -1311,6 +1313,46 @@ TEST_F(AtomStyleTest, ellipsoid)
     EXPECT_NEAR(bonus[3].quat[1], sqrt(5.0 / 30.0), EPSILON);
     EXPECT_NEAR(bonus[3].quat[2], sqrt(5.0 / 30.0), EPSILON);
     EXPECT_NEAR(bonus[3].quat[3], sqrt(5.0 / 30.0), EPSILON);
+    EXPECT_NEAR(bonus[0].block[0], 2.0, EPSILON);
```
akohlmey: See my note from above about backward compatibility. This only seems to be testing new case *with* additional properties. It should also check for the backward compatible case.

akohlmey (changes requested):
It seems that this pull request has not been correctly merged with upstream. This *must* be corrected.
Also, see my comments about backward compatibility and being more explicit in documenting optional properties. Whatever is contained in the pull request **must** keep full backward compatibility at the input file and data file level.

akohlmey:
@JBil8 Please also note that we recently merged changes that make the Bonus data accessible in the KOKKOS package. Thus there are some related changes required here as well. @stanmoore1 is more experienced in these matters to give advice, though.

src/atom_vec_ellipsoid.cpp
```diff
@@ -41,29 +42,32 @@ AtomVecEllipsoid::AtomVecEllipsoid(LAMMPS *lmp) :
   bonus_flag = 1;
 
   size_forward_bonus = 4;
-  size_border_bonus = 8;
-  size_restart_bonus_one = 8;
-  size_data_bonus = 8;
+  size_border_bonus = 13;
```
akohlmey: I don't agree with this kind of change that makes the code allocate and communicate additional data for the normal case, even though it is only needed for the special case.

One way to handle this cleanly would be to implement an optional keyword to the atom style that will "activate" the support for superellipsoids and then allocations and communication buffers can be adjusted accordingly.

Clean backward compatibility and not wasting memory (anything that is per-atom is a problem here) and communication bandwidth are development goals for LAMMPS.

JBil8: Hello, thank you for all the comments.

I see I made a mistake in a previous rebase that dragged along some unintended changes (e.g. to dump_image.rst). I am in the process of reverting that, by starting from the upstream develop branch and only cherry-pick necessary commits from this fork. I will then force push that branch here so that we do not loose the PR history.

Regarding backward compatibility, your feedback made me realize we were not testing for a case involving reading a data file from a previous ellipsoid simulation. This is a straightforward fix, but how we tackle it depends on the following design decision.

I agree with the idea of making "superellipsoid" an optional keyword so that standard ellipsoid simulations do not incur unnecessary memory and communication costs. This will require some code refactoring, which I have already begun.
However, in the current PR, @jibril-b-coulibaly also modified code for the aspherical compute and time integrators, to read inertia from the bonus struct rather than recomputing it. To move forward, I see two main options:
1 Keep the standard bonus struct as-is and add a second, optional bonus struct for superellipsoids (initialized only if the keyword is passed). We would then revert the changes to the computes and integrators so they compute inertia on the fly for standard ellipsoids. We would add an if-statement so that superellipsoids can still read cached inertia, avoiding the expensive beta function calculations.
2 Keep inertia in the standard bonus struct (accepting the memory cost) to save on computation time. Much of the implementation would remain as described above, but we would keep the current integrator/compute refactor. An additional bonus struct would still be needed for the blockiness exponents if the keyword is passed.

Avoiding block to be defined for standard ellipsoids, would also prevent much of the backward compatibility issue one could face.

Which approach do you recommend? @jibril-b-coulibaly @akohlmey @jibril-b-coulibaly @akohlmey

akohlmey: I suggest looking into making the superellipsoid bonus struct derived from the regular ellipsoid one. There should be a superellipsoid_flag added to atom.h to indicate which one to use. Any code not requiring the new data can remain unchanged.

doc/src/read_data.rst
jibril-b-coulibaly: I am responsible for these changes.
Some of the code processing ellipsoids was hardcoded. I had to make it modular to treat super ellipsoids. The values of `block1` and `block2` are only read if they are present in the data file. They default to 2 if they are absent so that the previous behavior is recovered, which should be backward compatible.
We will better document that these are optional

unittest/formats/test_atom_styles.cpp
jibril-b-coulibaly: You are correct, this is not exercising the backward compatibility w.r.t the `set` and `read` commands since the data files L.1198 writes the `block` attribute. I will modify this test.

src/atom_vec_ellipsoid.cpp
stanmoore1: > I suggest looking into making the superellipsoid bonus struct derived from the regular ellipsoid one. There should be a superellipsoid_flag added to atom.h to indicate which one to use. Any code not requiring the new data can remain unchanged.

I strongly agree with this.

jibril-b-coulibaly: @akohlmey thank you for the guidance in helping us keep the code lean.

I had made this design decision based on the fact that inertia is recomputed at every step in `fix_nh_asphere` and `fix_nve_asphere`, and there was a lot of code duplication. For regular ellipsoids, it is relatively cheap but for superellipsoids, the inertia requires calls to the `std::beta` function so I made a space/time tradeoff to store `inertia` in `Bonus` since the values never change. This choice was based on an assumption: I did not profile the code to see if storing/fetching/communicating was actually slower than recomputing at every step.
Was the code duplication and on-the-fly calculation an intentional memory-saving approach?
If we revert to recomputing inertia on the fly for all cases, we could remove `inertia` from the Bonus struct.

> I suggest looking into making the superellipsoid bonus struct derived from the regular ellipsoid one

If we have a derived structure (let's call it `BonusSuperellipsoid`) and point to it with a parent `Bonus*` pointer, this might involve a lot of casting to access `inertia` and `block` when super-ellipsoids are used. Would it be preferred to make an additional pointer to a `BonusSuperellipsoid` struct which only contains `inertia` and `block` attributes?

Thank you

akohlmey: @jibril-b-coulibaly It is difficult to make a meaningful assessment while the pull request branch is in the current rather messy state with many changes that don't belong.
Since the choice between the two data structures is a global one, you could just do (dynamic!) casts to both possible versions and have two pointers where the second will be NULL if only the ellipsoid version is used. You can then just use some if statements to avoid code duplication and you can eliminate the cost of the if statements, by using a templated function. We have similar code constructs in multiple places, e.g. the OPT package or in fix langevin. There is also precedent for having pointers to multiple bonus structs, for example in compute property/atom.

JBil8: @akohlmey thank you.

Now the history and PR should look better, since there are only the relevant files we intended to add/edit.

We could indeed template the functions in the integrators and computes to save if statement checks in critical loops.

For the bonus struct, we can add another `super_bonus` which inherits from the ellipsoid one. We can also add a `superellipsoid_flag` to `atom.h`.
By adding a `process_args` call which scans for a `superellipsoid` keyword in `atom_vec_ellipsoid` we can then set the flag and the buffer size.

However, if we already have the flag, would it be truly useful to have a null pointer to `super_bonus`?
Could we simply have one `bonus` pointer and then downcast it to `super_bonus` if the flag is on, when needed?   
Or maybe we could also drop the flag and keep the two pointers, one of them being null?

sjplimp:
Are you also changing to define a new (derived) AtomVec class for superellipsoids, along with its own Bonus data struct?
I'm not clear that it is a good idea to generalize AtomVecEllipsoid to include superellipsoids.

jibril-b-coulibaly:
> Are you also changing to define a new (derived) AtomVec class for superellipsoids, along with its own Bonus data struct? I'm not clear that it is a good idea to generalize AtomVecEllipsoid to include superellipsoids.

@sjplimp thank you for the insight: I think that is a great idea.

My initial plan was to re-use as much existing features as possible, and I thought I remembered from talks with Joel that adding atom_style should be done with caution to avoid having too many, that is why I used the `ellipsoid` atom style .

We could certainly derive a new `superellipsoid` AtomVec. This would solve most of the issues discussed above.

We would still want to allow users wanting to run granular simulations with "regular" ellipsoids being able to do so with the `ellipsoid` atom style. So there are few changes we would need to make to the flag checking logic in the pair_style, but I don't anticipate those being large.

If we get your green light, we will implement it that way and leave the Bonus struct of ellipsoids as is.

Thank you!

JBil8:
> > Are you also changing to define a new (derived) AtomVec class for superellipsoids, along with its own Bonus data struct? I'm not clear that it is a good idea to generalize AtomVecEllipsoid to include superellipsoids.
> 

Thank you for this suggestion. I have started to implement it as an add-on to `atom_vec_ellipsoid` as Axel was suggesting. But we could still pivot to a separate atom_style. 
How much support would be needed for superellipsoids outside the granular context? Could we start with only one time integrator, and then if there is demand, expand it over time? Also what other hurdles would there be in creating a new atom-style? Anything we should be particularly aware of?

Thank you

sjplimp:
@JBil8 @jibril-b-coulibaly 

(1) I think a new AtomVec class, like AVSuperEllipsoid, derived from AVEllipsoid, is good if it means you can avoid changing so many existing (ellipsoid-related) files.  I.e. just add your new AV class and new granular/superellipsoid pair styles.
But maybe there are nuances I don't understand yet, as well as @akohlmey does.

(2) For granular interactions with (super)ellipsiods or other possible uses of superellipsoids, @jtclemm is the best person to weigh in.  Note there is also an AtomVecBody class and BODY package, which implemented pair polygon and polyhedron styles for "body" particles in 2d/3d.   These have rounded corners and look similar to some pics I found of superellipsoids, but are more general since they can be arbitrary polygons/polyhedra.  It is a good example of a self-contained package which implements a new kind of aspherical particle with several support styles, touching little else in LAMMPS.  @ndtrung81 implemented most of this package.  The pair styles are contact potentials but do not (yet) have granular friction, history, etc.  So he might be interested in chiming in.

(3) I think your new doc pages definitely need some pics of superellipsoids for users.  I couldn't find a definitive set of them, so am confused myself as to what shapes they can be.

akohlmey:
> (1) I think a new AtomVec class, like AVSuperEllipsoid, derived from AVEllipsoid, is good if it means you can avoid changing so many existing (ellipsoid-related) files. I.e. just add your new AV class and new granular/superellipsoid pair styles.
> But maybe there are nuances I don't understand yet, as well as @akohlmey does.

I think in this specific case it would be sufficient to just have a derived class/struct for the Bonus component. Adding the enclosing radius to the ellipsoid atom style is a useful addition and all other differences are restricted to Bonus data. This way, any code that does not require accessing the superellipsoid bonus data can remain unchanged for both cases. I think that would reduce code redundancy and improve clarity. Of course, the problem with such predictions is that you don't really know whether this is the case until *after* it is done.

src/atom_vec_ellipsoid.h
```diff
@@ -26,13 +26,25 @@ namespace LAMMPS_NS {
...
     int ilocal;
   };
   struct Bonus *bonus;
 
+  struct BonusSuper : public Bonus {
+    double block[2];
+    double inertia[3];
+    BlockType type;
+  };
+  struct BonusSuper *bonus_super;
```
JBil8, line 46: Hi @akohlmey and @jibril-b-coulibaly,

I just pushed the updated architecture. As Axel suggested, superellipsoids are now implemented as a struct add-on directly within AtomVecEllipsoid.

To keep the PR limited, I've restricted the superellipsoid physics support strictly to fix nve/asphere and the granular pair styles. I reverted the other aspherical time integrators/ and some computes to upstream and added safety traps to throw an error if they are called with superellipsoids.

Tests and documentation are not yet updated to reflect this new memory layout. Before moving on to those, I wanted to know if this is a satisfactory layout, or if we should try something else.

sjplimp:
Can you explain how your updated arch affects the user, both in the input scrip and data files (Atoms and Ellipsoids sections)?  Is there now (still) only a single AtomVecEllipsoid?  No flag on the atom_style command?  Can the data file mix ellipsoids and superellipsoids?  Presumably different styles are going to check that only ellipsoids of the correct sub-style are defined ?

I found pics of superellipsoids here: https://en.wikipedia.org/wiki/Superellipsoid#/media/File:Superellipsoid_collection.png
Are you saying all these (some pointy) shapes will now work with granular contacts?  I'm skeptical of that :)

Also, the original purpose of Bonus data structs was to enable models with mostly point particles and a few ellipsoids to work with a single atom-style.  So pair gayberne and reqsuared encode soft point-particle/ellipsoid interactions as well.  Is that your intent with granular models as well, so that your new pair styles also encode spheroid/superellipsoid interactions?

Can there only be a single contact between any pair of superellipsoids in the pic or between a sphere and superellipsoid?
Doesn't seem possible if the two particles are different sizes.  If not, then detecting granular contacts and maintaining appropriate contact history seems complex.

JBil8:
> Can you explain how your updated arch affects the user, both in the input scrip and data files (Atoms and Ellipsoids sections)? Is there now (still) only a single AtomVecEllipsoid? No flag on the atom_style command?

Hello @sjplimp, 
yes surely. The new architecture keeps one unique `AtomoVecEllipsoid` atom style for both ellipsoids and superellipsoids.
To enable superellipsoids, the user adds a flag in the input script:
```
atom_style      ellipsoid superellipsoid
```
If this flag is missing, the code defaults to the legacy Bonus struct, ensuring no changes (and minimal performance regression limited to a few if-statements which branch predictors should take care of on modern cpus) for legacy users. If the flag is present, the code allocates the `BonusSuper` struct for extended particles.

> Can the data file mix ellipsoids and superellipsoids? Presumably different styles are going to check that only ellipsoids of the correct sub-style are defined ?
>  
Because `Bonus` and `BonusSuper` are mutually exclusive in memory per simulation, one cannot mix standard ellipsoids and superellipsoids in the same data file. If `superellipsoid` is active, the Atoms/Ellipsoids data file sections must include the blockiness parameters.
 
> I found pics of superellipsoids here: https://en.wikipedia.org/wiki/Superellipsoid#/media/File:Superellipsoid_collection.png Are you saying all these (some pointy) shapes will now work with granular contacts? I'm skeptical of that :)
> 

Yes, these is exactly how superellipsoids look like! We borrow much of the mathematical formulation from https://www.k1-met.com/fileadmin/user_upload/Publications/Journal_articles_open_access/Podlozhnyuk_2017.pdf ,  essentially a potential particle formulation.
In our code, and in the paper, the block parameters are defined as the inverse of those on Wikipedia (n_1 = 1 / epsilon_1).
Since the mathematical formulation relies on the hessian matrix, we cannot do pointy particles, but anything else from smoothed bricks to disks and cylinders can be done. The current code will throw an error for pointy particles (n1 or n2 < 2).
We document some of the mathematical details in `pair_gran_ellipsoid.rst` (which we could rename to pair_gran_superellipsoid.rst, TBD).
We also plan to release a short arXiv note detailing the exact solver inside `math_extra_superellipsoids` (which is essentially a careful Newton-Raphson with a line search).

> Also, the original purpose of Bonus data structs was to enable models with mostly point particles and a few ellipsoids to work with a single atom-style. So pair gayberne and reqsuared encode soft point-particle/ellipsoid interactions as well. Is that your intent with granular models as well, so that your new pair styles also encode spheroid/superellipsoid interactions?
> 

As of right now, no, we are not supporting interactions with pure point particles.
The main reason is that in DEM, even perfect spheres need a defined physical radius and angular momentum to handle friction and torques correctly. 
Currently, if a user wants to mix spheres and superellipsoids, they must declare the spheres as superellipsoids (by setting $shapex=shapey=shapez$ and $n_1=n_2=2$).
We recognize this means paying the computational penalty of storing and communicating the extra `BonusSuper` data for simple spheres.
However, keeping the contact solver unified makes this initial implementation much more robust. 
If there is significant community demand for running dilute superellipsoids in a bath of standard DEM spheres, adding a specialized point/sphere-to-superellipsoid analytical bypass could be a great future optimization.

> Can there only be a single contact between any pair of superellipsoids in the pic or between a sphere and superellipsoid? Doesn't seem possible if the two particles are different sizes. If not, then detecting granular contacts and maintaining appropriate contact history seems complex.

As long as the particles are not perfect cubes (i.e., they retain some finite curvature at the "flat" faces), the continuous overlap potential still yields a unique, mathematically rigorous point of midway penetration.
This proxy contact point allows us to use standard Hertzian or Hookean history tracking. We are drafting an ArXiv note detailing exactly how our Newton-Raphson solver preconditions and resolves this for polydisperse systems.

sjplimp:
@JBil8 - thanks for the clarifications

> As of right now, no, we are not supporting interactions with pure point particles.

For superellipsoids, I was asking about spheroids, not point particles.  If you are treating a spheroid as a particular shape of superellipsoid, does that mean you are changing the read_data file syntax to not have an ellipsoidflag = 0/1 ?  See the Atoms section of the read_data doc page.  Or if you are keeping that, are you requiring it equal 1 ?  Another option would be to allow the user to specify 0 (as for all other atom styles with bonus structures I think), but internally make it an SE with the settings for a spheroid, so that your pair style can treat everything as SE/SE interactions.   The other cost tradeoffs you mention for treating everything as SE/SE seem acceptable.

Re: contacts - are you allowing polydispersity in the definition of a collection of superellipsoids, e.g. some large, some small, similar to granular spheroids?  If so, you are still saying it is impossible for a single pair of superellipsoids to have multiple contact points, for all the SE shapes you support ?  Just want to check if this is the correct conceptual model.

Are the doc pages essentially complete?  I so, I will read thru them and make any suggestions I have.  Otherwise I am ready to approve this PR.  But I think @jtclemm should have the final approval.

JBil8:
> For superellipsoids, I was asking about spheroids, not point particles. If you are treating a spheroid as a particular shape of superellipsoid, does that mean you are changing the read_data file syntax to not have an ellipsoidflag = 0/1 ? See the Atoms section of the read_data doc page. Or if you are keeping that, are you requiring it equal 1 ? Another option would be to allow the user to specify 0 (as for all other atom styles with bonus structures I think), but internally make it an SE with the settings for a spheroid, so that your pair style can treat everything as SE/SE interactions. The other cost tradeoffs you mention for treating everything as SE/SE seem acceptable.
> 
Hi again and thanks for the support and review of this PR.
I am not 100% sure I understand this point, would you mind explaining better?
So far we are still keeping the ellipsoid flag. We require the user to set it to 1 since superellipsoids are extended particles. If the flag is initially zero and the user calls either set_shape or set_block the flag is changed and the additional bonus struct is allocated. 

> Re: contacts - are you allowing polydispersity in the definition of a collection of superellipsoids, e.g. some large, some small, similar to granular spheroids? If so, you are still saying it is impossible for a single pair of superellipsoids to have multiple contact points, for all the SE shapes you support ? Just want to check if this is the correct conceptual model.
> 
I am not sure I understand your point about polydispersity, but I can see that weird things might happen if a particle gets completely inside another one. In general this would violate many of the assumptions of DEM and it could be an edge case that we need to guard against. 
On the contrary you are right that the uniqueness of the contact point is not guaranteed for truly flat surfaces. While our code does a good job at finding one contact point (especially with the geometric formulation turned on), it does not guarantee it is the only one (for a flat on flat contact there would be infinite of them). Usually I would advise users to keep block values below 10. 
As an example see a snapshot from a simulation with small particles (yellow) with semiaxes 5 times smaller than the big particles (blue) which run without problems on 4 cores.
polydisperse_blocki_simulation


> Are the doc pages essentially complete? I so, I will read thru them and make any suggestions I have. Otherwise I am ready to approve this PR. But I think @jtclemm should have the final approval.

I have just updated the doc to reflect recent changes.
And if there is anything else to improve, or any other typos we might have left, please let us know.

jtclemm:
> Are the doc pages essentially complete? I so, I will read thru them and make any suggestions I have. Otherwise I am ready to approve this PR. But I think @jtclemm should have the final approval.

@sjplimp, are you content with the current use of atom vec classes and ellipsoid vs. super ellipsoid bonus structures now? If so, I will start going through and reviewing. 

> For superellipsoids, I was asking about spheroids, not point particles. If you are treating a spheroid as a particular shape of superellipsoid, does that mean you are changing the read_data file syntax to not have an ellipsoidflag = 0/1 ?

The doc page explicitly states that an ellipsoid flag of zero is for point particles. Since this is DEM, I would argue any particles with an ellipsoid flag of zero should therefore trigger an error in the granular pairstyle. Allowing an ellipsoid flag of zero also introduces the problem we had with lines/tris - what is the default radius since the data file can't read it? What happens when you use hybrid? Might introduce a lot of gotchas. 

@JBil8 for context, this is what Steve is referring to: https://github.com/lammps/lammps/blob/128499e0fa3b828d986b48b28e0fd112fb24f823/doc/src/atom_style.rst?plain=1#L351-L355

> Re: contacts - are you allowing polydispersity in the definition of a collection of superellipsoids, e.g. some large, some small, similar to granular spheroids? If so, you are still saying it is impossible for a single pair of superellipsoids to have multiple contact points, for all the SE shapes you support ? Just want to check if this is the correct conceptual model.

If particles have no concavities (their shape is equivalent to their convex hull, no pointiness) and no perfectly flat features, then I would think two particles in contact would always have one unique contact point regardless of their relative sizes. Is this correct @JBil8?

jibril-b-coulibaly:
I would agree with Joel that it would be best to error out if the superellipsoid flag is set in the atom style but not all atoms have the ellipsoid flag set to 1.
I did not enforce this when I first designed this on top of the ellipsoid atom style, but I think it is implicitly assumed in the pair style so we should definitely guard against it. 

Regarding the contact points, superellipsoids with exponents larger than or equal to 2 are convex and have strictly positive curvature everywhere so it is guaranteed there is a unique contact points, regardless of size. This contact point might be difficult to find for high exponents where the surfaces are near-flat, but there is only always one.

It might be possible to extend this to exponents strictly larger than 1, where particles remain convex with positive curvature everywhere, but I am not sure the guarantees remain around the ridges (we will need to do the math).

src/GRANULAR/pair_gran_hooke_history_ellipsoid.cpp
```diff
@@ -597,6 +597,18 @@ void PairGranHookeHistoryEllipsoid::init_style()
   if (comm->ghost_velocity == 0)
     error->all(FLERR, "Pair gran/h/ellipsoid* requires ghost atoms store velocity");
 
+  // ensure all atoms have an allocated ellipsoid bonus structure (ellipsoidflag > 0)
```
JBil8: How about this error check in the init_style of the granular pair_style?

sjplimp:
I agree with the consensus above that when using atom_style ellipsoid with the SE flag set, to require all the particles in the datafile have ellipsoidflag = 1.  I think you are also saying that the only way to simulate a small number of odd-shaped SEs in a bath of spheroids, is to treat all the spheroids as SEs as well.  Which seems fine, even if not as efficient as it could be for spheroid/spheroid interactions.

For my Q about polydispersity, it seems your pics (like the one above) seem to be with one style of SE, all the same size.  
So I am asking 2 things:
(a) Can a simulation model a mixture of different shaped SEs?
(b) For a SE rounded cube (or any other shape), can you run a simulation with a mixture of large cubes and small cubes?
The same as non-SE granular for spheroids can run a simulation with every spheroid being a different size if desired.

@jtclemm yes, I am content now with the new AtomVecEllipsoid class with options for the old Bonus (ellipsoid) and new Bonus (SE).

jibril-b-coulibaly:
> I think you are also saying that the only way to simulate a small number of odd-shaped SEs in a bath of spheroids, is to treat all the spheroids as SEs as well.

Yes this is correct.

> (a) Can a simulation model a mixture of different shaped SEs?

Yes. The `shape` and `block` are defined per-atom and can be `set` to any acceptable value.
Jacopo has played with this more than me but you can see an oblate particle hitting rounded cylinders in the `in.bowling` example (video below).

https://github.com/user-attachments/assets/2e8387a5-61a2-496a-b7d4-1a76720841e3

> (b) For a SE rounded cube (or any other shape), can you run a simulation with a mixture of large cubes and small cubes?
The same as non-SE granular for spheroids can run a simulation with every spheroid being a different size if desired.

Yes. The size of the SE is determined by the `shape` and can be different for each atom. The contact detection and resolution can handle contact between any 2 particles with any 2 `shape` and `block` vectors. The limits lies in the numerics of the Newton's method which requires more iteration / is more fragile for blocky particles (large `block`) and contact between very small and very large particles. @JBil8's research addresses these limitations and he'll be best positioned to go through the details. His "geometric" re-formulation of the minimization problem mitigates some of these difficulties pretty well.

JBil8:
> For my Q about polydispersity, it seems your pics (like the one above) seem to be with one style of SE, all the same size.

Ah I see the image was too small.
To clarify the visual: the yellow particles are actually parallelepipeds with semi-axes five times smaller than the blue particles. Additionally, the red floor consists of SE cubes ($n_1=n_2=8$), showing that the solver handles different SE shapes interacting in the same simulation. 
polydisperse_blocki_simulation_zoomed

I have also added a minimal standalone unit test for the superellipsoid atom_style.

src/ASPHERE/math_extra_superellipsoids.cpp
```diff
@@ -0,0 +1,835 @@
...
+};
+
+/* ----------------------------------------------------------------------
+   shape function computations for superellipsoids
+------------------------------------------------------------------------- */
+
+double shape_and_derivatives_local(const double* xlocal, const double* shape, const double* block, const int flag, double* grad, double hess[3][3])
+{
+  double shapefunc;
+  // TODO: Not sure how to make flag values more clear
```
jtclemm, line 166: Would it work to break the flag into two? Instead of passing an enum flag in `PairGranEllipsoid`, pass two ints:
  `int is_ellipsoid = flag == AtomVecEllipsoid::ELLIPSOID`
  `int is_n1_eq_n2 = flag == AtomVecEllipsoid::N1_EQUAL_N2`
Then all these functions have `int is_ellipsoid` and `int is_n1_eq_n2` as arguments.

An extra argument, but at least the variables aren't ambiguous.

src/ASPHERE/math_extra_superellipsoids.cpp
```diff
@@ -0,0 +1,835 @@
...
+      X0[2] = X_line[2];
+      X0[3] = X_line[3];
+    }
+
+    if (converged) break;
+  }
+
+  if (!converged){
+    if (shapefunc[0] > 0.0 || shapefunc[1] > 0.0) return 1;
+    std::cout << "Current residual norm: " << norm << std::endl;
```
jtclemm: Are these debug commands that can be commented out? Probably not good to call cout in parallel. If not, can the error message be displayed in PairGranEllipsoid?

src/ASPHERE/math_extra_superellipsoids.h
```diff
@@ -0,0 +1,841 @@
...
+void global2local_vector(const double v[3], const double *quat, double local_v[3]);
+
+inline double det4_M44_zero(const double m[4][4]);
+
+// 4 by 4 sytems solvers, they all overwrite b with the solution
+inline bool solve_4x4_manual(double A[16], double b[4]);
+inline bool solve_4x4_robust(double A[16], double b[4]);
+inline bool solve_4x4_robust_unrolled(double A[16], double b[4]);
+
+// ADD CONTACT DETECTION HERE
```
jtclemm: An old comment?

jtclemm:
Hi @JBil8, I made a few minor changes in a commit. Good to check, but it's mostly whitespace and comments (if you didn't already know, appending `?w=1` to the URL will hide whitespace changes in github). I also left a few comments. I saw a few other TODOs, are there any in particular you want to address in this PR vs. comments for later work? 

I'm thinking about the gran pair styles and will get back to reviewing shortly.

src/ASPHERE/math_extra_superellipsoids.cpp
JBil8: I think it could work, I do not have a strong opinion on this. 
The functions traces might become a bit longer though. 
However, could there be a case where by accident both ints are equal to 1? Maybe it is better to keep it as is, so it is clear there is only one "state". @jibril-b-coulibaly shall I remove this comment?

src/ASPHERE/math_extra_superellipsoids.cpp
JBil8: Thank you for that.
Yes I used them to debug, but they are deleted in the latest commit.
We already had warning in the Pair style for these cases.

src/ASPHERE/math_extra_superellipsoids.h
JBil8: Yes indeed. Removed in the latest commit

JBil8:
Hi @jtclemm, many thanks, it looks great to me!
 
>  I also left a few comments. I saw a few other TODOs, are there any in particular you want to address in this PR vs. comments for later work?

I have gone through and cleaned up the some TODOs that were already resolved.
Maybe the TODO for the touch function is something to defer to when this will be ported to the granular submodules. But feel free to edit. 
For the math_extra, the main remaining structural question is the gradient-only evaluation function used in the line search. In order not to duplicate code, we have a function that computes only the gradient and hessian, but  for speed one could have a function that computes both and a function that computes the gradient only. What would you prefer?

src/ASPHERE/math_extra_superellipsoids.cpp
jbcoulibaly: I think we should remove that comment.

We could possibly template this function on the `enum` / `int`, as opposed to passing the flag as an argument, that could save switching at every step on the statement below? Not sure if that's more convenient / performant than my current approach below

src/GRANULAR/pair_gran_hooke_history_ellipsoid.cpp
jbcoulibaly: The check line 603 below is what we discussed, @jtclemm is this enough, or did you want to check outside of the pair_style too?

src/ASPHERE/math_extra_superellipsoids.cpp
jtclemm: Agreed with just removing comment. Considering the surrounding math, I would think this performance might not be worth the effort.

Might be good to put a note on the definitions of the block types highlighting that their current 1, 2, 3 values are used in this method?

jtclemm:
> For the math_extra, the main remaining structural question is the gradient-only evaluation function used in the line search. In order not to duplicate code, we have a function that computes only the gradient and hessian, but for speed one could have a function that computes both and a function that computes the gradient only. What would you prefer?

I do not have a preference, so I'd say it's your decision. Could also leave it as is and always look for performance gains in a subsequent PR?

For the pair style, I thought about it and I believe it would be advantageous to move away from the pair gran/ellipsoid/X naming convention. This is because (a) the gran/X pair styles are outdated and has a connotation of being less accurate (b) specifying different contact models for different atom types seems like a good use case that will be easy to support. 

A better option, I think, is to combine the two pair styles into a single pair granular/ellipsoid (as an aside, why gran/ellipsoid instead of grain/superellipsoid? the latter seems more general). This would not need to use any of the subclass setup as pair granular, but could just be a copy of one of your existing pair style with a simple if/else branch in the compute() method for hooke/hertz and type-dependent coefficients. 

In the future, this pair style could then be easily added to. If enough contact models are added, then this could eventually be converted (while preserving input syntax) to the subclass setup in pair granular.

If you'd like, I'd be happy to draft the pair style. Then you can see what you think.

JBil8:
> I do not have a preference, so I'd say it's your decision. Could also leave it as is and always look for performance gains in a subsequent PR?
> 
Agreed!
> A better option, I think, is to combine the two pair styles into a single pair granular/ellipsoid (as an aside, why gran/ellipsoid instead of grain/superellipsoid? the latter seems more general).

I thought about this in the past and I am open to change the name of the pair style to superellipsoids. I thought many users could look for ellipsoids, and then maybe not find it, but probably is not an issue.

> If you'd like, I'd be happy to draft the pair style. Then you can see what you think.

Of course, so far there is a lot of code redundancy between hertz and hooke. I will keep an eye out for that.

Just so you now, you may notice there are currently both and `ALGEBRAIC` and a `GEOMETRIC` formulation of the contact problem. I am still testing the latter, but in the end it might not provide any performance gain compared to the former. I will keep you updated

jtclemm:
@JBil8 I just added a possible version of pair_granular_superellipsoids.cpp/h that combines the hooke and hertz variant. Please let me know if you have any thoughts one way or another.

Here's what I see:
Pros:
* one can specify different coefficients (or hertz vs hooke models) for different pairs of atom types
* it removes any stigma from the gran naming convention
* it is in a state that'd make it much easier to rewrite with the GranularModel class if there more contact models are added in the future (do we think that could ever happen?)
* it consolidates code for maintenance (the calculation of forces is located in one place instead of 4)

Cons:
* there are lots of new class variables that normally are held by the GranularModel class
* there is probably a performance drop due to copying variables and calling functions
* it'll need to be tested to ensure I didn't goof in copying + pasting

If this new approach is preferred, there are some other differences between pair granular and gran/X that can be ported over. This includes more usage of MathExtra to simplify math, changes in how the linear tangential model is calculated (better rotation?), (possibly) differences in how stiffnesses are normalized (I added a TODO to check), and better organization of history.

sjplimp:
I like the name pair granular/superellipsoids.  Also if you are adding lots of complexity to math_extra, specific to granular systems, I would suggest putting it in a math_granular.cpp/h file instead and putting that file in the GRANULAR package.

JBil8:
> @JBil8 I just added a possible version of pair_granular_superellipsoids.cpp/h that combines the hooke and hertz variant. Please let me know if you have any thoughts one way or another.

Hi @jtclemm, to me this looks great. I went over the .cpp and found a few minor domain-specific things that got lost in translation. I pushed these fixes directly to the branch. I left a TODO for the history transfer, since I realized even in the previous pair/gran I was probably doing something wrong with the shear transfer. Shall it be reversed when particles move across processors?

> it is in a state that'd make it much easier to rewrite with the GranularModel class if there more contact models are added in the future (do we think that could ever happen?)

That would be awesome. Maybe I am naive, but I do not see any computational hurdle in implementing twisting, rolling (might not be needed since the shapes are not spheres?) and even cohesion, since we should already have access to all the required data. However, most of these models rely on the exact contact area of the contact patch. We have an approximation of this area for the hertz model (those curvature computations), but it is an approximation, that degrades the more the aspect ratio becomes extreme or the block parameters become high. We could have the code and then the user decides if it is worth for their specific use case. 

> If this new approach is preferred, there are some other differences between pair granular and gran/X that can be ported over. This includes more usage of MathExtra to simplify math, changes in how the linear tangential model is calculated (better rotation?), (possibly) differences in how stiffnesses are normalized (I added a TODO to check), and better organization of history.

Green light from me, but I let @jbcoulibaly have the final say.
If this brings us closer to having wall contacts (we already have the function to determine the contact point with a wall ready to go), I am 100% on board.

jibril-b-coulibaly:
You have my green light.

I like the pair granular/superellipsoids, and we need to reduce code duplication when feasible.

There were subtleties that made `gran/hertz` different form `gran/hooke`, i.e., the `kn` and `kt` units
>"Thus Kn has units of force per area and is thus specified in units of (pressure)."

and the conversion we do in `PairGranHertzHistoryEllipsoid::settings()`
```cpp
kn /= force->nktv2p;
kt /= force->nktv2p;
```

I think these were changed in the modern pair granular, and we'd have to make sure to update the documentation and code of pair granular/superellipsoid accordingly.

Regarding the duplication of functions that compute shape functions, gradients, etc, the goal was indeed performance. Many expensive calculations are common to computing, e.g., the shape function and the gradient.
The trick was (under the assumption that`block>=2`), to compute the highest (`M`-th) derivative of `f(x)~x^N` first (which requires `std::pow(x, N-M)`), and then multiply it by `x` to get the lower derivatives. This avoids calling multiple `pow` or performing divisions. Depending on what the need is in the code (e.g., gradient vs hessian), the highest derivative needed is different and not all intermediate variables need to be calculated. I duplicated all this code to have one fast function for each specific case.
This was the immediate space/time tradeoff I chose. There is likely a better way and I'm open to change this

akohlmey:
Please note that for integer exponents you should use MathSpecial::powint()
which is *much* faster and more accurate than the regular pow() function.

Axel Kohlmeyer  ***@***.*** https://sites.google.com/site/akohlmey
Institute for Computational Molecular Science and HPC, https://icmshpc.com
College of Science & Technology, Temple University, Philadelphia A, USA

jibril-b-coulibaly:
@akohlmey thatnks for pointing me to MathSpecial::powint. Although the examples we’ve shown all use integer exponents, the exponents can be any real number.

It might be beneficial to have one of the Type flag indicate integer exponents, I will look into this. Thank you!

jtclemm:
@JBil8 good catch on those bugs! Particularly those negative signs... no idea how you saw that. Hopefully I didn't introduce any other bugs hiding in the code.

I took a 3rd stab at the history. I swapped it such that the required values (X0, separating axis) are first in the array and then contact model values (which may become optional) are appended to the end. Also fixed another bug where I forgot some checks on `update_history`. Lastly, I was unsure about passing a *double history value to MathExtraSuperellipsoids::check_oriented_bounding_boxes() since I didn't know if it was correct for it to always be updated. I swapped this to passing an int and having the method return the new axis so it's easier/clearer to update the cached axis. Please let me know if I misunderstood. 

Unfortunately that last change broke pair_gran_h*_ellipsoid.cpp, but I'm guessing those might eventually be removed if we are all in agreement with the consolidated approach?

Regarding twisting and other modes, I think you are correct that it isn't technically hard to add. I still want to revisit the new PairGranularSE::calculate_forces() method and more cleanly separate/notate the distinct normal, damping, and tangential sections (and update intermediate variables to the same naming convention in pair_granular.cpp). I think this will make it even clearer where new contact modes can be added. 

Thanks @jbcoulibaly for the clarification on kn and kt. I didn't know that was a difference! I removed the normalizations in the new pair style. Am I correct in understanding there are also differences in how shear forces are rotated? Is there any reason not to update this to match what's in pair granular? New documentation is still needed for pair granular/SE.

src/GRANULAR/pair_granular_superellipsoid.cpp
```diff
@@ -895,14 +895,16 @@ void PairGranularSuperellipsoid::unpack_forward_comm(int n, int first, double *b
...
+void PairGranularSuperellipsoid::transfer_history(double *source, double *target, int itype, int jtype)
 {
   // copy of all history variables (shear, contact point, axis)
-  // TODO: only shear needs to be reversed?
+
   for (int i = 0; i < size_history; i++) {
-    if (i < 3) target[i] = -source[i]; //shear
-    target[i] = source[i];
+    if (i > default_hist_size && tangential_model[itype][jtype] == LINEAR_HISTORY) {
+      target[i] = -source[i]; //shear
```
JBil8: Shall it be i>= default_hist_size here?
Otherwise maybe the x component of the shear does not get reverse?

jtclemm: yes! i can patch that.

jtclemm: I also took a stab at reorganizing the calculate forces method to align with pair granular, I can push that too

src/GRANULAR/pair_granular_superellipsoid.cpp
```diff
@@ -533,6 +529,13 @@ void PairGranularSuperellipsoid::init_style()
     if (!fix_history) error->all(FLERR,"Could not find pair fix neigh history ID");
   }
 
+  // grow history for contact models, right now this is superfluous and is just a placeholder
+
+  for (int itype = 1; itype <= atom->ntypes; itype++)
+    for (int jtype = 1; jtype <= atom->ntypes; jtype++)
+      if (tangential_model[itype][jtype] == LINEAR_HISTORY)
+        size_history += 3;
```
JBil8: Just so I understand, is size_history local or global? Does it add shear values in a different way for each pair "type" of contacts?

src/ASPHERE/pair_granular_superellipsoid.cpp
```diff
@@ -1023,10 +1025,14 @@ void PairGranularSuperellipsoid::calculate_forces()
...
-  X0_prev[2] = X0[2] - xref[2];
-  X0_prev[3] = X0[3];
+
+  if (history_update) {
+    double *X0_prev = history_data;
+    X0_prev[0] = X0[0] - xref[0];
+    X0_prev[1] = X0[1] - xref[1];
+    X0_prev[2] = X0[2] - xref[2];
+    X0_prev[3] = X0[3];
+  }
```
JBil8, line 1091: Shall we not wrap this with the if-statement?
Because I think X0 shall always be updated, so that we always start close enough to the solution.
Or is there some situation where one would not want to update it?

src/GRANULAR/pair_granular_superellipsoid.cpp
JBil8: Yes no problem for me. I can have a second look

src/GRANULAR/pair_granular_superellipsoid.cpp
jtclemm: In pair granular, size_history stores the maximum # of history values for any type-type combination. There's a separate variable that then stores the size for each specific type-type combinations. In this scenario, you always have +3 since there's only one tangential model. So I didn't bother defining the other variable yet.

src/ASPHERE/pair_granular_superellipsoid.cpp
```diff
@@ -940,9 +940,14 @@ bool PairGranularSuperellipsoid::check_contact()
...
-      double *separating_axis = &history_data[4];
-      skip_contact_detection = MathExtraSuperellipsoids::check_oriented_bounding_boxes(
+      int separating_axis = (int) (history_data[4]);
+      int new_axis = MathExtraSuperellipsoids::check_oriented_bounding_boxes(
           xi, Ri, shapei, xj, Rj, shapej, separating_axis);
+      if (new_axis != -1) {
+        skip_contact_detection = true;
+        if (history_update)
+          history_data[4] = (double) new_axis;
+      }
```
JBil8, line 997: Same here, shall we always update the previous axis for the oriented bounding box?

src/ASPHERE/pair_granular_superellipsoid.cpp
jtclemm: I'm not sure. Practically, I think you are right. There is no reason not to save the latest solution. Generally the single() method is only called at the end of a timestep to calculate properties for dump files.

Hypothetically, however, I could imagine someone writing a compute that temporarily  modifies an atom's position and calls single() to get a force. Maybe someone wants to calculate forces for a potential grain-grain monte carlo swaps? This doesn't seem like a good idea to me, but it could be safer to strictly forbid the single() method from modifying data.

sjplimp: Not sure I understand this comment thread - but I think single() should not modify data internal to the pair style.  A caller can manipulate data (e.g. coords in the case of a MC energy test, and restore the old coords if necessary) but the PairStyle cannot assume any ordering of single() calls, and single() should not modify the "state" of the PairStyle.  For pair models with granular history, this means the eval of single() can include history effects, but it should not update the history (like compute() does).  I assume that is already the case ?

akohlmey: > For pair models with granular history, this means the eval of single() can include history effects, but it should not update the history (like compute() does).

If anything, we should probably do the opposite and check in commands that do manipulate positions to get updated energy or forces (e.g. `fix numdiff` or `dyanamical_matrix` or the MC fixes) whether history is maintained by the pair style and then either print a warning or (probably safer) error out.

sjplimp: This is a good point, b/c when comparing 2 evals of single() with 2 different particle positions (as in the use cases Axel lists), I think the "history" effect is already broken.  B/c the 2 different positions may change a pair from being in-contact to out-of-contact or vice versa.

jtclemm: Given those fixes, good to keep any updates to X0 or that axis under this if statement. So no history values will be updated by single().

jtclemm:
Just pushed another large set of changes. I added in the actual linear_tangential model from pair granular and the viscoelastic damping model (which is the default, mirroring pair granular). The original model implemented in pair_gran_\*_ellipsoid.\* is still there under "classic". Unless there's an actual benefit to this original implementation, this could just be a hidden undocumented option.

While adding these extra models, I added large if/else blocks that designate the normal/tangential/damping force calculations. Also reduced some code with MathExtra functions. Some global variables and intermediate calculations are, again, unnecessary at this point, but would be useful down the road.

Please let me know @JBil8 if you have any concerns with these changes. If not, I think this PR should be good to merge after these changes are double checked for any bugs I may have introduced + documented. There are always other changes that could further align the pair style with pair granular (adding type-type history sizes, more contact models, etc.) but I think this is sufficient for now.

JBil8:
> Just pushed another large set of changes. I added in the actual linear_tangential model from pair granular and the viscoelastic damping model (which is the default, mirroring pair granular). The original model implemented in pair_gran_*_ellipsoid.* is still there under "classic". Unless there's an actual benefit to this original implementation, this could just be a hidden undocumented option.
> 

@jtclemm thanks for the refactoring and bringing this inline with the granular submodels. I kept the legacy styles working for debugging purposes; we can leave them undocumented.

> Please let me know @JBil8 if you have any concerns with these changes. If not, I think this PR should be good to merge after these changes are double checked for any bugs I may have introduced + documented. There are always other changes that could further align the pair style with pair granular (adding type-type history sizes, more contact models, etc.) but I think this is sufficient for now.

I started debugging, and caught a few things while doing so. However there is still a nasty bug when running the example `in.drop_test` with the new pair_style.
Starting from time 0.92 a lot of failed (old) contacts warnings appear and this is not the case with the old pair_gran. I looked for the bug but did not find the cause yet. Do you have any idea? 

Finally, I just noticed that because Hertz divides by local curvature, and superellipsoids can have perfectly flat faces (curvature = 0, along the principal axes) a face-to-face contacts result in infinite forces and NaN explosions. We could add a limit_curvature parameter to clamp the local curvature and prevent the solver from blowing up.

Otherwise, the rest looks quite good to me. Thanks for all the refactoring :)

jtclemm:
> However there is still a nasty bug when running the example `in.drop_test` with the new pair_style.
> Starting from time 0.92 a lot of failed (old) contacts warnings appear and this is not the case with the old pair_gran. I looked for the bug but did not find the cause yet. Do you have any idea?

Thanks for checking the examples, I can take a look at that issue and trace the logic where it fails. I'm not surprised I added a bunch of bugs. Speaking of which, good catch on those in the tangential model (vt -> vtr, missing LINEAR_HISTORY checks, etc)!

Regarding `cutoff_global`, this setting is typically only used if grains are growing using fix adapt. Do you see this as a possible future use case here? Or would modifying all the superellipsoid data structure be too complicated? If that's out of scope, we could remove that optional setting.

JBil8:
> Regarding `cutoff_global`, this setting is typically only used if grains are growing using fix adapt. Do you see this as a possible future use case here? Or would modifying all the superellipsoid data structure be too complicated? If that's out of scope, we could remove that optional setting.

I do not know if it would be very complex, but it seems no bonus_struct is supported by the current `fix_adapt`. But I think it could be useful at some point if people want to model particles wear (or some sort of thermal expansion/moisture absorption?)

jtclemm:
> But I think it could be useful at some point if people want to model particles wear

Okay we can leave it in in case someone ever desires to add those methods to fix adapt.

I fixed a significant bug that was preventing X0 (or any variable) from updating correctly. However, the code can still fail in determine_contact_point() (for instance, on step 9193 running serially). I checked all values passed to the function and they all are defined and look reasonable. Not sure what an error of status == 2 implies.

JBil8:
> I fixed a significant bug that was preventing X0 (or any variable) from updating correctly. However, the code can still fail in determine_contact_point() (for instance, on step 9193 running serially). I checked all values passed to the function and they all are defined and look reasonable. Not sure what an error of status == 2 implies.

Great, thanks for catching that with the `const` flag not being constant!

The simulation now works well, I simply swapped the order of the size history determination and the `replace_fix`.
I am still not sure this is correct, since when running for tangential models with standard shear, the size history should be 8 in our case and the debugger shows 17 (when there are two grain types).
I believe this comes from the fact the actual history size should be the sum of max of all the submodels, and not the sum over particle types. Looking at pair_granular it seems the logic there does a first pass to add the max across submodels and then sets the final size_history.

By the way, thank you for the rescaling of tangential stiffness from the Thornton paper. I had heard about it but never read it. I saw you also modified the main pair_granular, which confirmed my doubt that the tangential stiffness should be rescaled by the contact_radius if it is used.

jtclemm:
> The simulation now works well, I simply swapped the order of the size history determination and the `replace_fix`.

Ah! That might have been causing that residual error. That was a bad oversight on my part. Edit: I just checked and it looks like the bad contact error in the drop example went away now, but now I see a particle is lost at 20k timesteps. Looks like a particle gets pushed through a wall? Not obvious to me if this is a bug or a physical issue.

> I am still not sure this is correct, since when running for tangential models with standard shear, the size history should be 8 in our case and the debugger shows 17 (when there are two grain types).

Should be patched now.

I also rewrote all of the tangential rescaling calculations. I think they are much clearer (and a bit faster) now and make it more obvious why there should be a factor of contact_radius in the classic hertz model. 

I made these changes both for spheres (gran_sub_mod_tangential.cpp) and superellipsoids. @jibril-b-coulibaly, if you are also able to check these changes that'd be much appreciated. I'll run some tests of these changes too.

@JBil8 one question, why is the tangential damping for superellipsoids in terms of the relative translational tangential velocities (`vt`) instead of the relative tangential velocity including the rotational velocity (`vtr`)? Does this have to do with the non-spherical geometry?

JBil8:
> Should be patched now.

Wondeful!
 
> @JBil8 one question, why is the tangential damping for superellipsoids in terms of the relative translational tangential velocities (`vt`) instead of the relative tangential velocity including the rotational velocity (`vtr`)? Does this have to do with the non-spherical geometry?

I see the confusion. I renamed vt to vtr to match the convention for spherical particles. In the old version vt was already the total relative tangential velocity at the contact point, but I think now it is clearer.

akohlmey:
@JBil8 @jibril-b-coulibaly @jtclemm I just implemented (or rather made copilot do most of it) an option to render rounded triangles in LAMMPS and applied it to multiple cases of triangulated surfaces in LAMMPS, including ellipsoids (see PR #4925). This made me think if we should look into having an option to also add support for visualizing superellipsoids. The ellipsoid visualization is created by taking an icosahedron and iteratively refining it by replacing all triangles with four triangles and shifting the corner positions so they are located on the sphere. For the rounded triangles, you also need surface normals at the corners, but those are simply the position vector (since the original geometry is a unit sphere). This is then deformed, rotated, and translated as needed before rendering. This all works very similar to what OpenGL does if you don't use specific GLSL shaders.

With the new rounded triangles, only a moderate number of triangles is needed for a reasonably smooth visualization. So the question is: is there a reasonably simple way to obtain a suitable triangulation with normals for the superellipsoid objects?

JBil8:
> With the new rounded triangles, only a moderate number of triangles is needed for a reasonably smooth visualization. So the question is: is there a reasonably simple way to obtain a suitable triangulation with normals for the superellipsoid objects?

I tried to skim over the other PR, but I am not a graphics expert at all.
For the points to be mappd to the superquadric surface one simply needs to multiply by 

R = ( (|x/a|^n2 + |y/b|^n2)^(n1/n2) + |z/c|^n1 )^(-1/n1)
where a,b,c are shape[0], shape[1], shape[2]
n1 and n2 are block[0], block[1].

for the normals, one could use `shape_and_derivatives_local` in `math_extra_superellipsoid` and then normalize the gradient to obtain the unit normal. The function also computes the hessian though, so it might not be the most performant.

jibril-b-coulibaly:
@akohlmey I do not know about triangulation techniques. For the normals, Section 2.1 of [this paper](https://sudodem.github.io/docs/pdfs/ZhaoandZhao2019.pdf) has a closed-form expression that might be helpful (note that the exponents e1 and e2 in teh paper are different from the `block` we use, i.e., block[0] = n1 = 2/e1, block[1] = n2 = 2/e2)
I hope this helps

akohlmey:
@jbcoulibaly @JBil8 Thanks. If it is just a different deformation, than it should be straightforward to extend the existing triangulation accordingly. The normals don't need to be accurate. For the regular ellipsoids, I just re-use the normals of a sphere, so I can do the same for the superellipsoids. The "rounding" of the triangles is an illusion, anyway. It is just shading the surface as if it was rounded. But that is quite effective if you look at the updated images from the pull request here: https://github.com/lammps/lammps/pull/4925/changes
The noticeable edges from the flat triangles are gone even with a rather low resolution.

jtclemm (approved):
@akohlmey @sjplimp Do you have any opinions on the location of these new superellipsoid classes? They were split between ASPHERE and GRANULAR, but now combined into ASPHERE which simplifies compiling as both depend on ASPHERE/math_extra_superellipsoid.cpp.

akohlmey:
> Do you have any opinions on the location of these new superellipsoid classes?

I think ASPHERE is more appropriate. Do we need to track any dependencies on classes/files in GRANULAR?

akohlmey:
Does this look ok to you?
image image

or this?

image

This is still with flat triangles since PR #4925 is not yet merged.
But I have experimented with those changes as well, and it looks as if for the superellipsoids, the differences in the normals are too large and thus they need
to be adjusted.

sjplimp:
@jtclemm Does it make sense to have a new SUPERELLIPSOID package?  If not, ASPHERE is fine.

jtclemm:
> I think ASPHERE is more appropriate. Do we need to track any dependencies on classes/files in GRANULAR?

No, but it could possibly in the future if pair interactions grow and require splintering off subclasses like pair granular. Or if a wall style is ever introduced. (not sure what's planned on either front) 

> @jtclemm Does it make sense to have a new SUPERELLIPSOID package? If not, ASPHERE is fine.

A separate package might be more future-proof if capabilities ever expand. I don't know how easy/hard it might be to pull out changes from ASPHERE though, @JBil8?

But it sounds like there might be agreement on keeping everything in ASPHERE for now. Certainly the easiest option at the present.

akohlmey (approved):

akohlmey:
@JBil8 @jibril-b-coulibaly @jtclemm 

Compilation failed on macOS since the test machines are missing std::beta() despite claiming C++17 compatibility. I've added a wrapper around std::tgamma() to MathExtra which seems to be compiling. Please test if this is correct.

Also, for some reason fix rigid is broken for many test cases. This usually means some logic was changed inconsistently or there are pointers that are not properly initialized to nullptr in the initializer list or something similar. Please have a look.

src/ASPHERE/math_extra_superellipsoids.cpp
```diff
@@ -0,0 +1,820 @@
...
+    // TODO: might use a simpler function to simply compute the gradient, to
+    // avoid computing quantities already computed in compute_residual_and_jacobian
+    if (flagi <= 1)
+      val_dummy = shape_and_gradient_local_n1equaln2_surfacesearch(xilocal, shapei, blocki[0], gradi);
+    else
+      val_dummy = shape_and_gradient_local_superquad_surfacesearch(xilocal, shapei, blocki, gradi);
+
+    // Rotate gradient back to global frame to get normal
+    MathExtra::matvec(Ri, tmp_v, gradi);
+    MathExtra::normalize3(gradi, nij);
```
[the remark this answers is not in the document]
JBil8: This was a bad bug indeed. It is now fixed

src/ASPHERE/math_extra_superellipsoids.h
```diff
@@ -0,0 +1,834 @@
...
+    // If not separated, track the overlap depth
+    overlap = (R1_rad + R2_rad) - dist;
+
+    // Bias: Penalize edge axes slightly to prefer stable face contacts
+    if (i >= 6) overlap *= edge_bias;
+
+    if (overlap < min_overlap) {
+      min_overlap = overlap;
+      best_axis = i;
+    }
```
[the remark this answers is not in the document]
JBil8: This code is not used in this current implementation, but the bug has been fixed, in case someone wants to use it in the future

src/compute_property_atom.cpp
```diff
@@ -1317,50 +1336,166 @@ void ComputePropertyAtom::pack_angmomz(int n)
...
+  int *ellipsoid = atom->ellipsoid;
+  int *mask = atom->mask;
+  int nlocal = atom->nlocal;
+
+  for (int i = 0; i < nlocal; i++) {
+    if ((mask[i] & groupbit) && ellipsoid[i] >= 0)
+      buf[n] = bonus[ellipsoid[i]].block[1];
+    else buf[n] = 1.0;
+    n += nvalues;
+  }
```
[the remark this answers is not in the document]
JBil8: Fixed

JBil8:
> Compilation failed on macOS since the test machines are missing std::beta() despite claiming C++17 compatibility. I've added a wrapper around std::tgamma() to MathExtra which seems to be compiling. Please test if this is correct.
> 
Thank you, yes it does the job! I switched to the log formulation, which is what the std::beta does under the hood, to avoid overflow. 

> Also, for some reason fix rigid is broken for many test cases. This usually means some logic was changed inconsistently or there are pointers that are not properly initialized to nullptr in the initializer list or something similar. Please have a look.

I see, I forgot to initialize the `superellipsoid_flag` in the atom style. I should have fixed it, and the tests pass on my machine now.

Regarding the doc, I have seen copilot asks to add a version changed note to some of the functionalities. Is this needed in this case?

Finally, the visualizations look great with the flat triangles.
For smoothing, the normals of "blocki" particles can diverge significantly from the sphere case though.

akohlmey:
> Regarding the doc, I have seen copilot asks to add a version changed note to some of the functionalities. Is this needed in this case?

Actually, if a new feature was added, there should be a `.. versionadded:: TBD` before the paragraph describing that new feature. If some behavior was *changed* there should be a `.. versionchanged:: TBD` followed by an empty line and a properly indented single-line summary of what was changed.

> For smoothing, the normals of "blocki" particles can diverge significantly from the sphere case though.

Even for the regular ellipsoids, updated normals are needed https://github.com/lammps/lammps/pull/4925/changes/4c444a0a6d80c3249b5cb0272281b0842b684415
So once either this PR or PR #4925 are merged, code needs to be added to recompute the normals also for the superellipsoid case. Or, in case it doesn't look convincing, the code should fall back to flat triangles for superellipsoids.

akohlmey:
Apropos docs, doc/src/Packages_details.rst should be updated for the new contents of the ASPHERE package.

src/GRAPHICS/dump_image.cpp
```diff
@@ -1415,11 +1415,14 @@ void DumpImage::create_image()
         saved = reset_lighting(image, 0.3, 0.8, 0.45, 0.8);
       }
       EllipsoidObj e(elevel);
```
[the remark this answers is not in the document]
akohlmey: fixed

akohlmey:
@sjplimp Please have a final look before we merge this. Thanks.

sjplimp (approved):
I still think we need some pictures somewhere, perhaps on the Howto spherical page, of the different kinds of aspherical particles the code can model, including superellipsoids.  But that can be added in a follow-on PR.  @akohlmey Perhaps these can be generated with the recent dump image enhancements?

akohlmey:
> @akohlmey Perhaps these can be generated with the recent dump image enhancements?

Yes, I've been waiting for either this or PR #4925 to be merged so that I can combine the two code paths and implement the missing calculation of the surface normals for superellipsoids in addition to the normals for regular ellipsoids.

PR #4903, opened 2026-02-27
Fix NVE Spin hang

drjbarker:
**Summary**

When running on more than one processor and when atoms belong to different fix groups the nve/spin fix would hang and not integrate.

The issue is caused only in the case that sectoring is true (e.g. MPI with more than one processor) and a sector contains some atoms with the nve/spin fix and some without. There are while loops that become infinite because `i` never gets changed because `mask[i] & groupbit` keeps evaluating as false.

This is fixed by advancing `i` even if we don't need to integrate some spins.

**Author(s)**

Joseph Barker, School of Physics and Astronomy, University of Leeds, UK

**Licensing**

I identified the location of the hang using gdb the old fashioned way. It was clear there was the possibility for an infinite loop. I described the the situation in which the bug is triggered (>1 processes, a mixed spin/non-spin system) to OpenAI Codex (version 26.226.940 (720)) and asked it to inspect the lines I had identified in gdb and identify the root cause of the bug. The root cause is fairly obvious and simple to understand, so I accepted the proposed code changes from Codex. The code changes in this pull request are those suggested by Codex.

**Backward Compatibility**

No.

**Implementation Notes**

Tested by running a simulation with 2 mpi processes in a system of hybrid atom styles `full spin' where some atoms have spin length zero.

**Post Submission Checklist**


akohlmey:
@drjbarker Thank you for your submission. Can you provide us with a simple input deck that reproduces the hang so that we can confirm the bugfix since @julient31 seems to be too busy to confirm it at the moment?

drjbarker:
@akohlmey Below is a very simple input that triggers the bug. If you run with 1 MPI process it should run. If you run with 2 or more then it should hang. After the fix, both should work. I included a zeeman B-field for the spins to precess around so it's possible to confirm the integration is still working after the fix.

Please double check I'm not doing anything silly in the input, I'm a novice lammps user.

```
units           metal
dimension       3
boundary        p p p
atom_style      hybrid atomic spin
atom_modify     map array

lattice         sc 2.8
create_box      2 box
create_atoms    1 box

region          spin_slab block 0 10 0 10 0 4
set             region spin_slab type 2

mass            1 55.845
mass            2 55.845

group           nonspin type 1
group           spins   type 2

pair_style  hybrid/overlay lj/cut 8.0 spin/exchange 0.5
pair_coeff  * * lj/cut 0.1 2.5 8.0

pair_coeff  2 2 spin/exchange exchange 0.5 -0.01 0.0 3.0

neighbor        0.3 bin

velocity        all set 0 0 0

set             group spins spin/atom 1 1 0 1

fix             int_nonspin nonspin nve
fix             int_spins   spins   nve/spin lattice moving
fix             b_field spins precession/spin zeeman 0.1 0 0 1

compute         spcomp all property/atom spx spy spz
thermo          1
thermo_modify   flush yes
thermo_style    custom step temp pe ke press

dump            1 all custom 10 traj.dump id type x y z c_spcomp[1] c_spcomp[2] c_spcomp[3]
dump_modify     1 flush yes

timestep        0.001
run             100
```

akohlmey:
> Below is a very simple input that triggers the bug.

@drjbarker Thanks. I could reproduce the issue with it and the fix. I think we can include it, since it doesn't make matters worse (for now). :wink:.



> Please double check I'm not doing anything silly in the input, I'm a novice lammps user.

The following hybrid style is not needed, since *all* atom styles *include* "atomic".

> atom_style      hybrid atomic spin

I've played around with the recent GRAPHICS package additions and was able to create the following visualization of the spins (with total disregard of the physics otherwise, I should add). [in.spin.txt](https://github.com/user-attachments/files/25754733/in.spin.txt)

https://github.com/user-attachments/assets/88c698ab-82d4-4fa9-adba-04c3f8eea0f7

drjbarker:
Oh that's great. We had been wondering how to visualise the spins without too much hacking/processing of the output.

sjplimp (approved):

PR #4905, opened 2026-03-06
Implement coloring styles "index" and "atom" for lines, triangles, and ellipsoids

akohlmey:
**Summary**

This updates the color style selections for lines, tris, and ellipsoids to the same level as was implemented for bodies earlier. Most importantly, with coloring style "atom", the visualization can use the same colors as if they were drawn at atoms.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

Yes.

**Implementation Notes**

This follows closely the changes for bodies from PR #4898 

**Post Submission Checklist**

-----

This is showing the "line" input example colored by molecule-ID using sequential color map
image

### DIFF
--- a/doc/src/dump_image.rst
+++ b/doc/src/dump_image.rst
@@ -44,14 +44,14 @@ Syntax
            f_ID = per-grid vector calculated by a fix with ID
            f_ID[I] = Ith column of per-grid array calculated by a fix with ID
        *line* = color width
-         color = *type*
+         color = *type* or *index* or *atom*
          width = numeric value for line width (distance units)
        *tri* = color tflag width
-         color = *type*
+         color = *type* or *index* or *atom*
          tflag = 1 for just triangle, 2 for just tri edges, 3 for both
          width = numeric value for triangle edge width (distance units)
        *ellipsoid* = color eflag level width
-         color = *type*
+         color = *type* or *index* or *atom*
          eflag = 1 for triangles, 2 for wireframe, 3 for both
          level = mesh refinement level, value between 1 (low resolution) and 6 (ultra high resolution)
          width = diameter of wireframe edges (distance units) (ignored for triangles)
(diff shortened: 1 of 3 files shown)


doc/src/dump_image.rst
```diff
@@ -494,10 +494,21 @@ as described below.
...
-currently allowed for the *color* value is *type*, which will color
-the lines according to the atom type of the particle.  By default the
-mapping of types to colors is as follows:
+spheres, the same as if they were regular atoms.
+
+.. versionchanged:: TBD
+
+   added *index* and *atom* color styles
+
+The there are currently three supported settings for the *color* value:
```
sjplimp, line 503: There are currently ...

doc/src/dump_image.rst
```diff
@@ -494,10 +494,21 @@ as described below.
...
+*type*, *index*, or *atom*.  With the *type* setting the line particles
+will be colored according to the atom type of the particle.  With the
+*index* setting the coloring follows the line index instead.  With the
```
sjplimp, line 506: What is the line or tri index ?

src/GRAPHICS/dump_image.cpp
```diff
@@ -1230,12 +1248,30 @@ void DumpImage::create_image()
...
       j = clist[i];
       if (line[j] < 0) continue;
 
+      itype = type[j];
+      double opacity = aopacity[itype];
       if (lcolor == TYPE) {
-        color = colortype[type[j]];
+        color = colortype[itype];
+      } else if (lcolor == INDEX) {
+        itype = (line[j] % atom->ntypes) + 1;
```
sjplimp, line 1261: Here it looks what you are calling the line "index" is which entry it is in the list of local lines owned by each proc?  Which will change as particles (and lines) migrate from one proc to another ?  If so, I don't think the user will have any sense of what "index" means, since its an internal opaque setting.  Maybe you just want to say in the doc page that INDEX assigns colors to lines in round-robin order on each proc, and that the color of a particular line may not be static.

sjplimp (changes requested):
If you address my comments, am ready to approve.

doc/src/dump_image.rst
akohlmey: fixed, also 3 more times below.

src/GRAPHICS/dump_image.cpp
akohlmey: I have changed it to:
```
With the *index* setting, colors from
the list of available per-atom type colors are assigned to the body
particles in a non-deterministic round-robin fashion. 
```
I wanted to have something similar to having a per-molecule color.
The original implementation was for body particles and predates the implementation of the "atom" coloring style. With that, there is actually a better way to do what was intended, by adding molecule-IDs and then coloring atoms by "id" and assigning a sequential color map. I was considering to remove the "index" style because of that, but the other option is quite convoluted and requires multiple non-trivial steps, while for "index" you may just need to increase the number of atom types and use either the 6 default colors or add more colors corresponding to the number of types with `dump_modify acolor`.

sjplimp (approved):

Issue #4032, opened 2024-01-04
Creating a new compute command for the virial-velocity correlation function 


luizfcpe:
**Summary**

I would like to suggest the creation of a new compute command which calculates the virial-velocity correlation function. It should be possible to implement it by modifying the compute vacf command.

**Detailed Description**

The virial-velocity correlation function can be used to obtain the spectral decomposition of the heat current under non-equilibrium conditions, and thus obtain the spectral decomposition of the thermal conductivity. With the spectrally decomposed thermal conductivity it is possible to identify the contribution of different frequency vibrations to the thermal conductivity of systems under non-equilibrium conditions. 

The virial-velocity correlation function takes the general form K(t)=\sum_{i=atoms in relevant group} <W_i(0) v_i(t)>, where i represents the atom index, W_i(0) is the virial part of the atomic stress tensor at the moment the compute is defined, and v_i(t) is the atomic velocity vector at instant t. The spectral heat current is proportional to the Fourier transform of the virial-velocity correlation function J(omega)~FT{K(t)}, and the spectral thermal conductivity is proportional to the spectral heat current: \kappa(omega)~J(omega)/(\Delta T/L), where (\Delta T/L) is the temperature gradient along the direction of heat transport.

The theoretical background is described in the following publications:
Role of anharmonic phonon scattering in the spectrally decomposed thermal conductance at planar interfaces. K. Sääskilahti et al, Phys. Rev. B 90, 134312 (2014) - Sec. II A
Vibrational mean free paths and thermal conductivity of amorphous silicon from non-equilibrium molecular dynamics simulations. K. Sääskilahti et al, AIP Advances 6, 121904 (2016) - Sec II
Mechanisms of temperature-dependent thermal transport in amorphous silica from machine-learning molecular dynamics. T. Liang et al, Phys. Rev. B 108, 184203 (2023) - Sec. IV - A, B, C

In principle, a compute virial-velocity correlation function could be created by modifying the already implemented compute vacf command. I could help with the implementation, but I do require assistance from an experienced LAMMPS developer.

akohlmey:
> In principle, a compute virial-velocity correlation function could be created by modifying the already implemented compute vacf command. I could help with the implementation, but I do require assistance from an experienced LAMMPS developer.

The majority of features in LAMMPS are contributed, especially those that are not in mainstream use (but even some of those). The logic behind this is the following: the person who wants a feature is also the one interested in making certain it is working correctly and has an application to test it on.

[...]

PR #4911, opened 2026-03-11
Centroid stress tensor implementation for many-body pair styles + virial heat flux through the TALLY package

MarkPoulos:
**Summary**

In a recent publication [[1]](https://doi.org/10.1103/mtkk-kyyy) with @donatas-surblys , we have implemented in LAMMPS an extension of the centroid formulation of the heat flux to general many-body potentials, again based on the theoretical work by Torii et al [[2]](https://doi.org/10.1063/1.2821963). 

This formulation conveniently allows at the same time the computation of both the centroid atomic stress tensor and of the exact virial heat flux for many-body pair styles (the latter through the callback mechanism by the heat/flux/virial/tally compute).

An important new feature that this brings to LAMMPS is the ability to calculate the spectral decomposition of the heat flux from the Fourier Transform of the atomic velocities and many-body pseudo-forces output by the updated heat/flux/virial/tally compute, as described in [[1]](https://doi.org/10.1103/mtkk-kyyy). So far, this was only available for pairwise forces, or through a harmonic approximation of the many-body forces [[3]]( https://doi.org/10.1103/PhysRevB.90.134312).

**Related Issue(s)**

This is a continuation of PR #1704, while it also relates to issue #4032 concerning the spectral decomposition of the heat flux.

**Author(s)**

Markos Poulos (INSA Lyon, France) markos_poulos@yahoo.gr
Donatas Surblys (Tohoku University, Japan)

**Licensing**

**Backward Compatibility**

These changes should not affect backwards compatibility. 

**Implementation Notes**

As mentioned in the summary, the previous implementation of the centroid stress tensor was restricted to bond/angle/dihedral/improper styles due to the fact that these styles have the same way to attribute the many-body potential energy terms to each atom in the same way (i.e. equally among all contributing atoms). This is not the case for certain many-body pair styles (e.g in Tersoff, although it technically contains 2-body and 3-body energy terms, in the 3-body ones the entire potential energy is equally shared between the two first atoms i,j while the 3rd atom k gets nothing). Therefore a general implementation is needed,  taking into account the weights p_i (for the theoretical details see [[1]](https://doi.org/10.1103/mtkk-kyyy). This was done by creating new virial tally functions within the Pair class (one for general 3-body terms and one for general 4-body terms), which are called during force /energy computations and update the centroid atomic stress tensor. 

At the same time, through the compute callback mechanism of the TALLY package, these functions enable the calculation of many-body "pseudo-forces" F2-->1 across the two groups through the heat/flux/virial/tally compute, which then enables the exact computation of the virial heat flux for many-body forces, as well as its spectral decomposition by post-processing. For theoretical details see [[1]](https://doi.org/10.1103/mtkk-kyyy).

So far these functions are only called by the Tersoff and Stillinger-Weber potentials. Simulations  have been performed and verified for a variety of materials using the above potentials (solid 3D and 2D) [[1]](https://doi.org/10.1103/mtkk-kyyy).

I have added an explicit computation of the centroid stress tensor in Pair::ev_tally() function, in order to account for many-body potentials containing 2-body terms, since for those cases we cannot use CENTROID_SAME and just copy the "old" stress tensor to the centroid one.

I have also added the citation to our publication through the CiteMe class and updated the documentation files for the affected commands.

There is also possibility to extend the callback mechanism of the heat/flux/virial/tally compute to the bond/angle/dihedral/improper classes, in order to enable the computation of the virial heat flux and its spectral decomposition for these styles as well.

**Post Submission Checklist**

[1] https://doi.org/10.1103/mtkk-kyyy  (Theoretical details and results of new implementation)
[2] https://doi.org/10.1063/1.2821963 (Original theoretical work by Torii et al.)
[3] https://doi.org/10.1103/PhysRevB.90.134312 (Previous work on spectral decomposition of the HF)


MarkPoulos:
Dear @akohlmey @sjplimp @athomps, I just wanted to say that I mostly agree with all of Copilot's suggestions and I will commit them and rebase to the latest changes in the develop branch once I have your feedback. Until then, just 2 small questions:

1. Is the `pairflag` check needed on top of the `force->pair` check?
2. Should I indeed add the `.. versionchanged:: TBD` to the compute tally documentation?

akohlmey:
> Until then, just 2 small questions:
> 
> 1. Is the `pairflag` check needed on top of the `force->pair` check?

It is done with centroidstessflag above, so just for consistency it should be done.
But also "pairflag" has a different purpose than checking if force->pair is non-null.

> 2. Should I indeed add the `.. versionchanged:: TBD` to the compute tally documentation?

Actually, it should be `.. versionadded:: TBD` at the beginning of a paragraph describing a new feature, but otherwise yes.

doc/src/compute_heat_flux.rst
```diff
@@ -75,6 +75,10 @@ equation for :math:`\mathbf{J}` is **not** included in the calculation
 performed by these computes; you need to add it for a volume appropriate to the
 atoms included in the calculation.
 
+.. note::
+
+   The heat flux vector is formally defined on a cross-section of the system. The compute *heat/flux* calculates a spatially averaged heat flux vector across the volume of the group specified in the command. For the calculation of the heat flux on a control surface, see the :doc:`compute heat/flux/atom <compute_heat_flux_atom>` command.
+
 .. note::
```
athomps, line 82: This `compute heat/flux/atom` is not in LAMMPS.

athomps (changes requested):
@MarkPoulos thanks for this impressive pull request.  Because of the complexity, I need a little bit more time to review this in detail.  I did notice one item in `compute_heat_flux.rst` that you can probably resolve quite easily, in addition to the co-pilot issues.

doc/src/compute_heat_flux.rst
MarkPoulos: It was a mistake, I meant to point to the updated _heat/flux/virial/tally_ compute. Fixed it, thanks!

MarkPoulos:
> @MarkPoulos thanks for this impressive pull request. Because of the complexity, I need a little bit more time to review this in detail. I did notice one item in `compute_heat_flux.rst` that you can probably resolve quite easily, in addition to the co-pilot issues.

@athomps thank you for the kind words. I have integrated your suggestion as well as the Copilot ones. A remaining issue is: do we want to extend the new TALLY formulation through the callback mechanism to the bond/angle/dihedral/improper styles as well? If so, should this be done in this pull request or open a new one?

akohlmey:
> do we want to extend the new TALLY formulation through the callback mechanism to the bond/angle/dihedral/improper styles as well?

like with so many things in LAMMPS it is more a question of "who is taking care of it after the initial implementation?" than anything else. LAMMPS has grown pretty big and the LAMMPS developers are stretched rather thin these days to keep everything clean and working. So, if you think it is useful, willing to implement and test it **and** willing to maintain it for the foreseeable future, then I would say you should go ahead.

> If so, should this be done in this pull request or open a new one?

A new pull request. This will need to be evaluated independently.

akohlmey:
This triggers the following compile error with C++20 and later
```
/home/runner/work/lammps/lammps/src/TALLY/compute_heat_flux_virial_tally.cpp
/home/runner/work/lammps/lammps/src/TALLY/compute_heat_flux_virial_tally.cpp: In constructor ‘LAMMPS_NS::ComputeHeatFluxVirialTally::ComputeHeatFluxVirialTally(LAMMPS_NS::LAMMPS*, int, char**)’:
/home/runner/work/lammps/lammps/src/TALLY/compute_heat_flux_virial_tally.cpp:80:11: error: ‘strcmp’ was not declared in this scope
   80 |       if (strcmp(arg[iarg], "two_body") == 0)
      |           ^~~~~~
/home/runner/work/lammps/lammps/src/TALLY/compute_heat_flux_virial_tally.cpp:25:1: note: ‘strcmp’ is defined in header ‘<cstring>’; did you forget to ‘#include <cstring>’?
   24 | #include "update.h"
  +++ |+#include <cstring>
   25 | 
```

This is because using `std::format` instead of `fmt::format` implicitly includes fewer C style header files. We following the IWYU (include what you use) principle anyway, so that change is required.

akohlmey:
@athomps This is waiting on you currently.

PR #4914, opened 2026-03-12
Add ENERGY_ONLY bit to eflag parameter for force computation

### LINKED ISSUE #4847: [Feature Request] Energy-only calculation flag for pair style to speed up Monte Carlo moves evaluation
**Summary**

During the different types of MonteCarlo simulations (i.e. atom/swap, gcmc) there is no need for pair style to evaluate forces/stresses. This can save ~20-40% of computation time, depending on the model complexity, especially for MLIPs.

**Detailed Description**
Try to extract and set the "compute_energy_only" flag for the pair style if available (like [here](https://github.com/yury-lysogorskiy/lammps/blob/feature/mc-energy-only-compute/src/MC/fix_atom_swap.cpp#L613))
```
    int* flag_compute_energy_only = (int *)force->pair->extract("compute_energy_only",dim);
    if (flag_compute_energy_only) *flag_compute_energy_only = 1;
    force->pair->compute(eflag, vflag);
    if (flag_compute_energy_only) *flag_compute_energy_only = 0;
```

Pair style must be able to
* accept this flag ([link](https://github.com/yury-lysogorskiy/lammps/blob/feature/mc-energy-only-compute/src/ML-PACE/pair_grace_fs.cpp#L417))
* deal with flag in internal implementation, like [here](https://github.com/yury-lysogorskiy/lammps/blob/feature/mc-energy-only-compute/src/ML-PACE/pair_pace.cpp#L166) or [here](https://github.com/yury-lysogorskiy/lammps/blob/feature/mc-energy-only-compute/src/ML-PACE/pair_grace_fs.cpp#L153)

Computation tests show ~30% speed up for MC moves, when only energy eval is required.

I'd like to ask if there is already similar mechanism for doing this, and, if not, if the proposed solution could be accepted into LAMMPS after modifications, if needed.

akohlmey:
**Summary**

This adds a new bit, ENERGY_ONLY, to the eflag variable which indicates that only the (global) energy is required. This can be used by force styles to skip over steps only required for computing forces and may result in significant savings for computations like MC fixes when they need to do a full energy computation.

**Related Issue(s)**

This addresses the feature request and thus closes #4847 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

Yes.

**Implementation Notes**

Rather than using "magic" numbers we now use the symbolic constants to set eflag and vflag to make the desired intent more obvious.

**Post Submission Checklist**

### DIFF
--- a/src/DIELECTRIC/fix_polarize_bem_gmres.cpp
+++ b/src/DIELECTRIC/fix_polarize_bem_gmres.cpp
@@ -350,8 +350,10 @@ void FixPolarizeBEMGMRES::compute_induced_charges()
   double *em = atom->em;
   double *epsilon = atom->epsilon;
   int nlocal = atom->nlocal;
-  int eflag = 1;
-  int vflag = 0;
+
+  // flag that we only need to compute the global energy
+  int eflag = ENERGY_GLOBAL | ENERGY_ONLY;
+  int vflag = VIRIAL_NONE;
 
   // compute the right hand side (vector b) of Eq. (40) according to Eq. (42)
   // keep the scaled real charges intact here to compute efield for the right hand side (b)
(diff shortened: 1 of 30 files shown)


athomps (approved):
I approve.

ndtrung81 (approved):

Issue #2, opened 2015-02-09
examples/KAPPA/log.heat.1Feb14 & /log.heatflux.1Feb14


ee10bt:
As far as I can tell, these two log files have come from the same calculation, run on separate occasions. I noticed the problem in the header of log.heat, it is supposed to relate to "add/subtract energy to 2 regions via fix heat", not the Green-Kubo method of log.heatflux. As such, the output of running in.heat does not match up with the provided log.heat.

TL;DR, log.heat in examples/KAPPA does not match up with in.heat.

junghans:
Sorry, this a mirror repository, we don't take bug report here, instead please report the issue on lammps' mailing list: lammps-users@lists.sourceforge.net

PR #4915, opened 2026-03-15
`region grid(/kk)` and other misc improvements / bugfixes 

alphataubio:
**Summary**

- [ ] `src/region_grid.h`
`src/region_grid.cpp`
`src/grid3d.h`
`src/grid3d.cpp`
`src/compute.h`
`doc/src/region_grid.rst`
`examples/grid/in.grid.3d`
`examples/ttm/in.ttm.grid`
region grid (define detailed region, instead of simple polygon, based on a 3d grid from a fix or compute, eg. fix ttm/grid)

- [ ] `src/KOKKOS/region_grid_kokkos.h`
`src/KOKKOS/region_grid_kokkos.cpp`
KOKKOS version region grid/kk

- [ ] `examples/KAPPA/in.heat.region_grid`
compute_grid / region_grid Adaptive Thermal Channel example relevant to sandia scientists

- [x] `src/atom.cpp` 
remove error "_Label maps are currently not supported with Kokkos_" since labelmaps work great in KOKKOS as comprehensively tested by me

- [x] `src/dump_local.h`
`src/dump_local.cpp` 
add step field to dump local

- [x] `src/EXTRA-COMMAND/group2ndx.cpp`
deleted group segfault bugfix

- [x] `src/compute_property_local.h`
`src/compute_property_local.cpp`
`doc/src/compute_property_local.rst`
add pmolecule1 pmolecule2 

- [x] `src/EXTRA-FIX/fix_ave_moments.cpp`
`doc/src/fix_ave_moments.rst`
add coefficient of variation (cv) 

- [x] `src/KOKKOS/atom_vec_ellipsoid_kokkos.h`
`src/KOKKOS/atom_vec_ellipsoid_kokkos.cpp`
`src/KOKKOS/fix_langevin_kokkos.h`
`src/KOKKOS/fix_shake_kokkos.cpp`
fix compile errors for KOKKOS_PREC mixed

- [x] `src/ADIOS/*`
modernize to ADIOS 2.11.0 (BP5) / C++20 and segfault bugfixes

**Related Issue(s)**

i didnt bother making issues, i just fixed the bugs 

**Author(s)**

@alphataubio 

**Licensing**

Claude Sonnet and Gemini Pro generated first draft of some source files, and assisted in analyzing code during debugging.

**Backward Compatibility**

n/a

**Implementation Notes**

**Post Submission Checklist**


sjplimp:
Hi @alphataubio - thanks for the submission - I scanned this new PR briefly and it has 2 problems.

(1) We (meaning the developers) don't like PRs which combine a potpourri of unrelated new features, bug fixes, code changes, etc because it is harder for us to wade thru.  That's especially true if you are making changes (even minor ones) to the core of LAMMPS.  So I'm going to ask you limit this PR to just the new "region grid" feature.  Other proposed changes or bug fixes can be posted in different PRs.  Again, one set of changes per PR for one issue or bug-fix.  That will speed up the review process for you and us.

(2) This PR also has no description at the top for the new feature.  Why is it useful, by who and for what could it be used, how is it implemented, etc.  Those kind of details can't simply be buried in new code, or people like me have no context for what to look for, or why.  For example, I want to know (before I look at code) what a grid-based region even means, since it seems like it could be collection of disjoint grid cells, which change continuously.  Why is it useful to call that a region?  Why is it not implemented as simply a dynamic group of particles?  So please post a Summary which aims to convince us this is useful and implemented correctly.

Thanks!

alphataubio:
> (1) We (meaning the developers) don't like PRs which combine a potpourri of unrelated new features, bug fixes, code changes, etc because it is harder for us to wade thru. That's especially true if you are making changes (even minor ones) to the core of LAMMPS. So I'm going to ask you limit this PR to just the new "region grid" feature. Other proposed changes or bug fixes can be posted in different PRs. Again, one set of changes per PR for one issue or bug-fix. That will speed up the review process for you and us.

sorry @sjplimp its too inconvenient and time consuming to work on mutliple small parallel branches and PRs, git switch/merge/..., resolve conflicts, ... , wait for months/years for changes to get merged into develop, port my changes from PRs ready to merge into my latest PR ro run production simulations from my current research that depends on changes from 5+ PRs waiting to get merged , ...

so for now ill keep all my work into _one_ draft PR and if/when i have time in the summer after school ill split all my contributions into 10+ PRs of 2-5 commits/each but for now i simply dont have the spare time to spend hours with git

alphataubio:
## ADIOS2 dump: C++20 modernization + bug fixes

### Bugs fixed

#### 1. Heap corruption in `dump_atom_adios.cpp` — `init_style()`

The original code assigned string literals directly to `columns`, a `char*`
member that the `Dump` base class manages with `new[]`/`delete[]`:

```cpp
// BEFORE (bug): assigns a pointer into the read-only data segment
columns = (char *) "id type x y z";
```

When the base class later calls `delete[] columns` on a string-literal address
it writes into the allocator's free-list metadata, producing:

```
malloc_consolidate(): invalid chunk size
free(): corrupted unsorted chunks
```

The crash manifests far from the write site — typically inside PMIx's
background event loop during MPI finalization — because the allocator only
detects the damage the next time it inspects that memory region.

Fix: always place a heap-allocated copy via `utils::strdup()`:

```cpp
// AFTER (fixed)
delete[] columns;
columns = nullptr;
columns = utils::strdup("id type x y z");
```

`dump_custom_adios.cpp` already did this correctly; the fix brings
`dump_atom_adios.cpp` into line.

---

#### 2. Wrong shape used for atom table write in `dump_custom_adios.cpp` — `write()`

The original called `SetShape()`/`SetSelection()` on the `Variable` object but
then wrote using a string name:

```cpp
// BEFORE (bug): string lookup ignores the updated shape/selection
internal->varAtoms.SetShape({nAtomsGlobal, nColumns});
internal->varAtoms.SetSelection({{startRow, 0}, {nAtomsLocal, nColumns}});
// ...
internal->fh.Put<double>("atoms", buf);   // resolves to placeholder {1, nColumns}
```

`SetShape()`/`SetSelection()` update the internal state of the `Variable`
object. A string-name lookup resolves the variable to its original placeholder
dimensions from `DefineVariable({1, nColumns}, ...)`, completely ignoring those
updates. The engine calculates the wrong buffer size, reads past the end of
`buf`, and corrupts the heap.

`dump_atom_adios.cpp` already used the variable object correctly; the fix
brings `dump_custom_adios.cpp` into line:

```cpp
// AFTER (fixed): use the object, not the string name
internal->fh.Put<double>(internal->varAtoms, buf);
```

---

### `adios_common.h`

**`static const char default_config[]` → `inline constexpr std::string_view`**
`static` in a header gives each translation unit its own copy of the data,
violating the one-definition rule for constants. `inline constexpr` shares a
single definition across all TUs. `std::string_view` is the C++20-idiomatic
type for referring to compile-time string data without heap allocation.

**BP4 → BP5 engine**
BP5 is the recommended default engine since ADIOS2 2.9. It has lower memory
footprint, faster metadata handling, and better multi-step append performance.

**`substreams` → `NumAggregators` parameter**
`substreams` is a BP4-only parameter and is silently ignored by the BP5 engine.
`NumAggregators` is the correct BP5 key controlling how many MPI ranks write
directly to storage.

---

### Both `.h` files

**`DumpAtomADIOSInternal *internal` → `std::unique_ptr<DumpAtomADIOSInternal>`**
Expresses ownership unambiguously, eliminates the manual `delete internal` in
every destructor body, and makes double-free impossible. Works with the
incomplete-type pimpl pattern because the destructor is defined out-of-line in
the `.cpp` where the type is complete.

**`~DumpAtomADIOS() override` declared, defined as `= default` in `.cpp`**
`unique_ptr<IncompleteType>` requires the destructor to be defined at a point
where the pointed-to type is complete. Declaring it in the header and
defaulting it in the `.cpp` satisfies this requirement.

**Copy constructor and copy-assignment `= delete`**
The pimpl struct holds ADIOS2 handle objects (`IO`, `Engine`, `Variable`) that
carry internal reference-counted state and must not be inadvertently copied.
Making this explicit prevents accidental copying that would compile silently but
corrupt engine state at runtime.

**`#include <memory>` added**
Required for `std::unique_ptr`.

---

### Both `.cpp` files — pimpl struct

**`class` → `struct`**
All members need to be accessible from within the same translation unit.
`struct` is the idiomatic choice for pimpl implementation types.

**Custom destructor calling `fh.Close()` before `ad` is destroyed**
`adios2::Engine` is a lightweight reference wrapper; it does NOT call `Close()`
when it goes out of scope. Without an explicit close, the `ADIOS` object (`ad`)
can be destroyed while the engine still holds open file handles, causing
undefined behaviour. The destructor body runs before member destructors, so
`fh` and `ad` are both still valid when `fh.Close()` executes. Destruction then
proceeds in reverse-declaration order: `fh` first, then `io`, then `ad`.

**`adios2::ADIOS *ad` → `std::unique_ptr<adios2::ADIOS> ad`**
Replaces manual `delete internal->ad` in every exit path. Destruction order
(reverse declaration order) ensures `fh` is already closed before `ad` is
released.

**Copy and move operations `= delete`**
Same rationale as in the header: ADIOS2 handle types carry internal state that
must not be naively duplicated.

**`static constexpr std::string_view ioName`**
A compile-time constant string does not need a per-instance `std::string` heap
allocation. `string_view` with `static constexpr` storage is the C++20
idiomatic choice.

---

### Both `.cpp` files — constructor

**`new DumpAtomADIOSInternal()` → `std::make_unique<DumpAtomADIOSInternal>()`
in the member-initializer list**
Exception-safe: if the ADIOS constructor throws, the pimpl object is
automatically destroyed. The initializer-list placement ensures the pimpl is
valid before the constructor body runs.

**`fopen`/`fclose` → `std::filesystem::exists` + `std::ofstream`**
Replaces raw C I/O with C++17/20 standard-library facilities. `std::ofstream`
is RAII so the file is always closed even if an exception is thrown. No need
to check the return value of `fclose`.

**`catch (std::ios_base::failure &e)` → `catch (const std::ios_base::failure &e)`**
Catching by const reference is correct when the exception object is not
modified in the handler.

---

### Both `.cpp` files — `write()`

**`bigint atomOffset;` → `bigint atomOffset{}`**
Value-initializes `atomOffset` to zero. Without initialization the variable
holds an indeterminate value between declaration and `MPI_Scan`, which is
undefined behaviour if `MPI_Scan` somehow fails silently before writing it.

**`atomOffset -= nme` → `atomOffset -= bnme` with clarifying comment**
`MPI_Scan` returns an *inclusive* prefix sum (the sum of values on ranks
0..me). Subtracting the local value converts it to an *exclusive* prefix sum
(the offset of this rank's data in the global array). The variable renamed to
`bnme` makes clear it is the `bigint` form of `nme` that was actually passed to
`MPI_Scan`.

**`static_cast<uint64_t>` on all integer `Put` calls**
`update->ntimestep`, `ntotal`, `bnme`, and `atomOffset` are LAMMPS `bigint`
(typically `int64_t`) or `int`. Passing them directly to `Put<uint64_t>` relies
on an implicit narrowing conversion. Explicit casts make the intention clear
and silence compiler warnings.

**`internal->fh.Put<double>(internal->varAtoms, buf)` in `dump_custom_adios`**
See bug fix #2 above.

---

### Both `.cpp` files — `init_style()`

**`strchr` / manual byte-shift loop → `std::shift_left` (C++20)**
The original hand-rolled loop shifted bytes one position left to delete the
`%` character in-place. `std::shift_left` from `<algorithm>` (C++20) expresses
the same operation on a range in a single readable call:

```cpp
const std::size_t tail = std::strlen(pct) + 1;  // include '\0'
std::shift_left(pct, pct + tail, 1);
```

**`SetEngine("BPFile")` → `SetEngine("BP5")`**
`BPFile` is an alias that historically resolved to BP4. Naming BP5 explicitly
is unambiguous and forward-compatible.

**`{"substreams", nstreams}` → `{"NumAggregators", nstreams}`**
See `adios_common.h` rationale above.

**`int num_aggregators = multiproc; if (num_aggregators == 0) num_aggregators = 1;`
→ `const int num_aggregators = (multiproc > 0) ? multiproc : 1;`**
Combines declaration, initialization, and the fallback into one expression.
`const` documents that the value does not change after initialization.

**`reinterpret_cast<int *>` → `reinterpret_cast<const int *>`**
`domain->boundary` is not modified by `DefineAttribute`. Using a pointer-to-const
is correct and prevents the compiler from generating a spurious warning about
casting away const-ness if the underlying type ever changes.

**Comment on `DefineAttribute<int>("boundary", boundaryptr, 6)`**
Added explanation that `domain->boundary` is `int[3][2]` — three spatial
dimensions each with a lo and hi value — laid out contiguously in memory, giving
exactly 6 integers.

---

### `dump_atom_adios.cpp` — `init_style()` specifically

**`delete[] columns; columns = nullptr;` before the `utils::strdup` assignments**
See bug fix #1 above. The `columns = nullptr` guard ensures that if
`utils::strdup` were to throw (it won't, but defensively), the pointer is not
left dangling.

---

### `dump_custom_adios.cpp` — `init_style()` specifically

**`combined.size()` → `!combined.empty()`**
`empty()` is the idiomatic C++20 way to test whether a string has any content.
Avoids an implicit conversion from `size_t` to `bool`.

**Validation loop comment rewritten**
The original comment said the compute/fix/variable loops do nothing and it is
"unclear why we need them here." This is incorrect. The loops:
1. Cache the `compute[]`, `fix[]`, and `variable[]` pointers that `count()` and
   `pack()` will dereference on every timestep.
2. Emit clear early errors for missing IDs or incompatible fix output
   frequencies rather than crashing silently during the first write.
They are not dead code and must not be removed.

**`id_variable[i]` added to the variable-not-found error message**
The original message was `"Could not find dump custom/adios variable name"` with
no indication of *which* variable was missing. The name is now included for
diagnosability.

alphataubio:
> detailed analysis of comm tiled and comm brick in kokkos to see if one is vulnerable to this mixed precision bug while the other comm style is not and WHY ?
TransformView::modify ERROR: Concurrent modification of legacy host and Kokkos host views in TransformView "atom:v"

Now I have everything needed for a thorough analysis. Let me synthesize this.This is a rich bug — let me lay out the full mechanistic explanation with a diagram.

---

## Root cause: CommTiledKokkos has no device-native exchange or borders path

The error `TransformView::modify ERROR: Concurrent modification of legacy host and Kokkos host views in TransformView "atom:v"` is about the **TransformView** mechanism — the LAMMPS/Kokkos abstraction used for mixed-precision storage. For `atom:v`, it maintains two conceptually separate host representations:

- **Legacy host view** — the raw `double** v` pointer accessed by non-Kokkos code
- **Kokkos host view** — the `Kokkos::View<float**>` (or double) on the CPU side, managed by Kokkos

The TransformView prohibits both being dirty simultaneously, because it can't know which one is authoritative — there is no safe sync direction to resolve the conflict.

---

## How the conflict arises: the `auto_sync + modified(Host)` pattern

The unsafe sequence that triggers it is:

```
atomKK->sync(Host, V_MASK)         // 1. Sync to host — clean state
lmp->kokkos->auto_sync = 1         // 2. Enable auto-sync mode
legacy_comm->exchange_or_borders() // 3. Accesses atom->v (double**)
                                   //    auto_sync intercepts → marks LEGACY HOST dirty
lmp->kokkos->auto_sync = 0         // 4. Restore
atomKK->modified(Host, V_MASK)     // 5. Tries to mark KOKKOS HOST dirty
                                   //    But legacy host already dirty → BOOM
```

Step 5 is the detonator. The TransformView sees two concurrent modifications from two different host-side owners with no intervening sync to reset state.

---

## Why CommKokkos avoids this (when `exchange_comm_legacy = false`)

`CommKokkos` has a complete **device-native** exchange and borders path — `exchange_device<DeviceType>()` and `borders_device<DeviceType>()`. When `exchange_comm_legacy = false` (the default), both functions return early after calling these Kokkos-native implementations:

```cpp
// CommKokkos::exchange() — line 858
if (!exchange_comm_legacy) {
    if (exchange_comm_on_host) exchange_device<LMPHostType>();
    else exchange_device<LMPDeviceType>();
    return;  // ← exits before the legacy host path
}
// Only reaches here if explicitly forced to legacy
atomKK->sync(Host, datamask_exchange);
lmp->kokkos->auto_sync = 1;
CommBrick::exchange();
lmp->kokkos->auto_sync = prev_auto_sync;
atomKK->modified(Host, datamask_exchange);  // ← only reached if exchange_comm_legacy=true
```

`exchange_device()` uses Kokkos kernels (`pack_exchange_kokkos`, `unpack_exchange_kokkos`) that work exclusively on Kokkos views. It calls `atomKK->sync(ExecutionSpace, X_MASK)` once to build the sendlist, but it **never touches the legacy `double**` pointer for `atom:v`**, so `auto_sync` never marks the legacy host view dirty. There's no `atomKK->modified(Host, ...)` at the end. The TransformView never sees the conflict.

Same logic applies to `borders_device()` in `CommKokkos::borders()`.

---

## Why CommTiledKokkos is always vulnerable

Now look at `CommTiledKokkos::exchange()` — **the entire function**:

```cpp
// CommTiledKokkos::exchange() — line 362
void CommTiledKokkos::exchange()
{
  atomKK->sync(Host, atomKK->avecKK->datamask_exchange);   // ← sync to host

  int prev_auto_sync = lmp->kokkos->auto_sync;
  lmp->kokkos->auto_sync = 1;                              // ← auto_sync ON
  CommTiled::exchange();                                    // ← legacy access to atom->v
                                                           //    auto_sync marks LEGACY HOST dirty
  lmp->kokkos->auto_sync = prev_auto_sync;

  atomKK->modified(Host, atomKK->avecKK->datamask_exchange); // ← marks KOKKOS HOST dirty → CONFLICT
}
```

There is no `exchange_comm_legacy` check. There is no `exchange_device()`. There is no early exit. **Every single call to `CommTiledKokkos::exchange()` runs this pattern unconditionally.** The `exchange_comm_legacy` flag is stored in the class (line 74 of `comm_tiled_kokkos.cpp`), but is never consulted in `exchange()`.

`CommTiledKokkos::borders()` has the exact same structure — it always calls the legacy host path with `auto_sync=1`, then calls `atomKK->modified(Host, datamask_border_vel)` which includes `V_MASK` when `ghost_velocity=true`.

---

## Why `atom:v` specifically, not `atom:x`

`atom:x` has `NEED_TRANSFORM` checks scattered throughout both comm styles. The fast-path shortcut `if (comm_x_only && !atomKK->k_x.NEED_TRANSFORM)` lets positions bypass the pack/unpack entirely when no precision transform is needed — and when a transform IS needed, the code is structured to avoid the conflict.

`atom:v` only enters the picture when `ghost_velocity = true`. When that is set:

- It's included in `datamask_border_vel`, `datamask_comm_vel`, and `datamask_exchange`
- `CommTiledKokkos::borders()` explicitly syncs and modifies it via `datamask_border_vel`
- `CommTiledKokkos::exchange()` includes it via `datamask_exchange`

The conflict becomes detectable precisely because velocity is touched during the legacy exchange/borders path (via `CommTiled::exchange()` reading `atom->v` internally with `auto_sync=1`) and then marked modified again via `atomKK->modified(Host, ...)`.

For `atom:x`, `BuildExchangeListFunctor` in `CommKokkos::exchange_device()` takes `atomKK->k_x` directly as a `ttransform_kkfloat_1d_3_lr`, bypassing the legacy pointer entirely. CommTiledKokkos has no equivalent functor path for exchange.

Here's the structural picture:---

## Detailed breakdown by function

### `exchange()` — the primary trigger

`CommKokkos::exchange()` guards its entire logic behind `if (!exchange_comm_legacy)`, which defaults to `false`. When false, it calls `exchange_device<DeviceType>()` and returns. That function calls `atomKK->sync(ExecutionSpace, X_MASK)` — positions only, for building the sendlist — then calls `pack_exchange_kokkos` / `unpack_exchange_kokkos` which are Kokkos-only kernels that never dereference `atom->v` via the legacy `double**` pointer. The legacy host view of `v` is never accessed, `auto_sync` never fires for it, and `atomKK->modified(Host, ...)` is never called. The TransformView remains in a single-owner state.

`CommTiledKokkos::exchange()` is four lines without any guard. It unconditionally does `sync(Host) → auto_sync=1 → CommTiled::exchange() → modified(Host)`. The `exchange_comm_legacy` member variable is populated in `init()` but is **never consulted** in `exchange()`. There is no `exchange_device()` template defined in `CommTiledKokkos` at all — it simply doesn't exist. Every call runs the conflicting pattern.

### `borders()` with `ghost_velocity=true`

Same structural asymmetry. `CommKokkos::borders()` routes to `borders_device<DeviceType>()` when `!exchange_comm_legacy`. That function builds the sendlists with `BuildBorderListFunctor` on the device, calls `pack_border_kokkos` / `unpack_border_kokkos`, and never calls `modified(Host, V_MASK)`. `CommTiledKokkos::borders()` always goes through `CommTiled::borders()` with `auto_sync=1`, then calls `atomKK->modified(Host, datamask_border_vel)` — which includes `V_MASK` — producing the same concurrent dirty conflict.

### `forward_comm()` — symmetric, not the source

Both classes have `forward_comm_device()` and guard their legacy path with `if (forward_comm_legacy)`. The conflict isn't sourced here — it's that once `exchange()` or `borders()` has already put the TransformView into a bad state, any subsequent access that tries to resolve it may amplify the error.

---

## The `NEED_TRANSFORM` connection

The bug only manifests when `NEED_TRANSFORM = true` on `atomKK->k_v` — i.e., when the Kokkos device view stores a different precision than the legacy `double**` host view (typically single-precision GPU storage with double-precision CPU). You can see this conditioning in the fast-path guards throughout `forward_comm_device()`:

```cpp
if (comm_x_only && !atomKK->k_x.NEED_TRANSFORM) {
    // direct DMA into device x buffer — no pack/unpack needed
}
```

When `NEED_TRANSFORM = false` (pure double precision), the legacy and Kokkos host views are the same memory and the TransformView is essentially a pass-through, so "concurrent modification" of two views that alias to the same allocation is harmless. When `NEED_TRANSFORM = true`, they are genuinely separate buffers with a precision conversion between them, and the TransformView enforces the invariant that only one can be authoritative at a time.

---

## Fix path for CommTiledKokkos

The correct fix is to implement `exchange_device()` and `borders_device()` for the tiled comm style — mirroring the `BuildExchangeListFunctor` / `BuildBorderListFunctor` patterns from `CommKokkos`, adapted for the recursive bisection decomposition. The shorter workaround is to add a guard in `CommTiledKokkos::exchange()` and `borders()` that restructures the legacy path to not double-mark the velocity:

```cpp
void CommTiledKokkos::exchange()
{
  atomKK->sync(Host, atomKK->avecKK->datamask_exchange);
  int prev_auto_sync = lmp->kokkos->auto_sync;
  lmp->kokkos->auto_sync = 1;
  CommTiled::exchange();
  lmp->kokkos->auto_sync = prev_auto_sync;
  // Don't call atomKK->modified(Host,...) here if auto_sync already
  // marked the legacy host dirty — the sync at the top already established
  // a clean baseline; auto_sync handles ongoing tracking.
}
```
Now I have everything needed for a thorough analysis. Let me synthesize this.This is a rich bug — let me lay out the full mechanistic explanation with a diagram.

---

## Root cause: CommTiledKokkos has no device-native exchange or borders path

The error `TransformView::modify ERROR: Concurrent modification of legacy host and Kokkos host views in TransformView "atom:v"` is about the **TransformView** mechanism — the LAMMPS/Kokkos abstraction used for mixed-precision storage. For `atom:v`, it maintains two conceptually separate host representations:

- **Legacy host view** — the raw `double** v` pointer accessed by non-Kokkos code
- **Kokkos host view** — the `Kokkos::View<float**>` (or double) on the CPU side, managed by Kokkos

The TransformView prohibits both being dirty simultaneously, because it can't know which one is authoritative — there is no safe sync direction to resolve the conflict.

---

## How the conflict arises: the `auto_sync + modified(Host)` pattern

The unsafe sequence that triggers it is:

```
atomKK->sync(Host, V_MASK)         // 1. Sync to host — clean state
lmp->kokkos->auto_sync = 1         // 2. Enable auto-sync mode
legacy_comm->exchange_or_borders() // 3. Accesses atom->v (double**)
                                   //    auto_sync intercepts → marks LEGACY HOST dirty
lmp->kokkos->auto_sync = 0         // 4. Restore
atomKK->modified(Host, V_MASK)     // 5. Tries to mark KOKKOS HOST dirty
                                   //    But legacy host already dirty → BOOM
```

Step 5 is the detonator. The TransformView sees two concurrent modifications from two different host-side owners with no intervening sync to reset state.

---

## Why CommKokkos avoids this (when `exchange_comm_legacy = false`)

`CommKokkos` has a complete **device-native** exchange and borders path — `exchange_device<DeviceType>()` and `borders_device<DeviceType>()`. When `exchange_comm_legacy = false` (the default), both functions return early after calling these Kokkos-native implementations:

```cpp
// CommKokkos::exchange() — line 858
if (!exchange_comm_legacy) {
    if (exchange_comm_on_host) exchange_device<LMPHostType>();
    else exchange_device<LMPDeviceType>();
    return;  // ← exits before the legacy host path
}
// Only reaches here if explicitly forced to legacy
atomKK->sync(Host, datamask_exchange);
lmp->kokkos->auto_sync = 1;
CommBrick::exchange();
lmp->kokkos->auto_sync = prev_auto_sync;
atomKK->modified(Host, datamask_exchange);  // ← only reached if exchange_comm_legacy=true
```

`exchange_device()` uses Kokkos kernels (`pack_exchange_kokkos`, `unpack_exchange_kokkos`) that work exclusively on Kokkos views. It calls `atomKK->sync(ExecutionSpace, X_MASK)` once to build the sendlist, but it **never touches the legacy `double**` pointer for `atom:v`**, so `auto_sync` never marks the legacy host view dirty. There's no `atomKK->modified(Host, ...)` at the end. The TransformView never sees the conflict.

Same logic applies to `borders_device()` in `CommKokkos::borders()`.

---

## Why CommTiledKokkos is always vulnerable

Now look at `CommTiledKokkos::exchange()` — **the entire function**:

```cpp
// CommTiledKokkos::exchange() — line 362
void CommTiledKokkos::exchange()
{
  atomKK->sync(Host, atomKK->avecKK->datamask_exchange);   // ← sync to host

  int prev_auto_sync = lmp->kokkos->auto_sync;
  lmp->kokkos->auto_sync = 1;                              // ← auto_sync ON
  CommTiled::exchange();                                    // ← legacy access to atom->v
                                                           //    auto_sync marks LEGACY HOST dirty
  lmp->kokkos->auto_sync = prev_auto_sync;

  atomKK->modified(Host, atomKK->avecKK->datamask_exchange); // ← marks KOKKOS HOST dirty → CONFLICT
}
```

There is no `exchange_comm_legacy` check. There is no `exchange_device()`. There is no early exit. **Every single call to `CommTiledKokkos::exchange()` runs this pattern unconditionally.** The `exchange_comm_legacy` flag is stored in the class (line 74 of `comm_tiled_kokkos.cpp`), but is never consulted in `exchange()`.

`CommTiledKokkos::borders()` has the exact same structure — it always calls the legacy host path with `auto_sync=1`, then calls `atomKK->modified(Host, datamask_border_vel)` which includes `V_MASK` when `ghost_velocity=true`.

---

## Why `atom:v` specifically, not `atom:x`

`atom:x` has `NEED_TRANSFORM` checks scattered throughout both comm styles. The fast-path shortcut `if (comm_x_only && !atomKK->k_x.NEED_TRANSFORM)` lets positions bypass the pack/unpack entirely when no precision transform is needed — and when a transform IS needed, the code is structured to avoid the conflict.

`atom:v` only enters the picture when `ghost_velocity = true`. When that is set:

- It's included in `datamask_border_vel`, `datamask_comm_vel`, and `datamask_exchange`
- `CommTiledKokkos::borders()` explicitly syncs and modifies it via `datamask_border_vel`
- `CommTiledKokkos::exchange()` includes it via `datamask_exchange`

The conflict becomes detectable precisely because velocity is touched during the legacy exchange/borders path (via `CommTiled::exchange()` reading `atom->v` internally with `auto_sync=1`) and then marked modified again via `atomKK->modified(Host, ...)`.

For `atom:x`, `BuildExchangeListFunctor` in `CommKokkos::exchange_device()` takes `atomKK->k_x` directly as a `ttransform_kkfloat_1d_3_lr`, bypassing the legacy pointer entirely. CommTiledKokkos has no equivalent functor path for exchange.

Here's the structural picture:---

## Detailed breakdown by function

### `exchange()` — the primary trigger

`CommKokkos::exchange()` guards its entire logic behind `if (!exchange_comm_legacy)`, which defaults to `false`. When false, it calls `exchange_device<DeviceType>()` and returns. That function calls `atomKK->sync(ExecutionSpace, X_MASK)` — positions only, for building the sendlist — then calls `pack_exchange_kokkos` / `unpack_exchange_kokkos` which are Kokkos-only kernels that never dereference `atom->v` via the legacy `double**` pointer. The legacy host view of `v` is never accessed, `auto_sync` never fires for it, and `atomKK->modified(Host, ...)` is never called. The TransformView remains in a single-owner state.

`CommTiledKokkos::exchange()` is four lines without any guard. It unconditionally does `sync(Host) → auto_sync=1 → CommTiled::exchange() → modified(Host)`. The `exchange_comm_legacy` member variable is populated in `init()` but is **never consulted** in `exchange()`. There is no `exchange_device()` template defined in `CommTiledKokkos` at all — it simply doesn't exist. Every call runs the conflicting pattern.

### `borders()` with `ghost_velocity=true`

Same structural asymmetry. `CommKokkos::borders()` routes to `borders_device<DeviceType>()` when `!exchange_comm_legacy`. That function builds the sendlists with `BuildBorderListFunctor` on the device, calls `pack_border_kokkos` / `unpack_border_kokkos`, and never calls `modified(Host, V_MASK)`. `CommTiledKokkos::borders()` always goes through `CommTiled::borders()` with `auto_sync=1`, then calls `atomKK->modified(Host, datamask_border_vel)` — which includes `V_MASK` — producing the same concurrent dirty conflict.

### `forward_comm()` — symmetric, not the source

Both classes have `forward_comm_device()` and guard their legacy path with `if (forward_comm_legacy)`. The conflict isn't sourced here — it's that once `exchange()` or `borders()` has already put the TransformView into a bad state, any subsequent access that tries to resolve it may amplify the error.

---

## The `NEED_TRANSFORM` connection

The bug only manifests when `NEED_TRANSFORM = true` on `atomKK->k_v` — i.e., when the Kokkos device view stores a different precision than the legacy `double**` host view (typically single-precision GPU storage with double-precision CPU). You can see this conditioning in the fast-path guards throughout `forward_comm_device()`:

```cpp
if (comm_x_only && !atomKK->k_x.NEED_TRANSFORM) {
    // direct DMA into device x buffer — no pack/unpack needed
}
```

When `NEED_TRANSFORM = false` (pure double precision), the legacy and Kokkos host views are the same memory and the TransformView is essentially a pass-through, so "concurrent modification" of two views that alias to the same allocation is harmless. When `NEED_TRANSFORM = true`, they are genuinely separate buffers with a precision conversion between them, and the TransformView enforces the invariant that only one can be authoritative at a time.

---

## Fix path for CommTiledKokkos

The correct fix is to implement `exchange_device()` and `borders_device()` for the tiled comm style — mirroring the `BuildExchangeListFunctor` / `BuildBorderListFunctor` patterns from `CommKokkos`, adapted for the recursive bisection decomposition. The shorter workaround is to add a guard in `CommTiledKokkos::exchange()` and `borders()` that restructures the legacy path to not double-mark the velocity:

```cpp
void CommTiledKokkos::exchange()
{
  atomKK->sync(Host, atomKK->avecKK->datamask_exchange);
  int prev_auto_sync = lmp->kokkos->auto_sync;
  lmp->kokkos->auto_sync = 1;
  CommTiled::exchange();
  lmp->kokkos->auto_sync = prev_auto_sync;
  // Don't call atomKK->modified(Host,...) here if auto_sync already
  // marked the legacy host dirty — the sync at the top already established
  // a clean baseline; auto_sync handles ongoing tracking.
}
```
Now I have everything needed for a thorough analysis. Let me synthesize this.This is a rich bug — let me lay out the full mechanistic explanation with a diagram.

---

## Root cause: CommTiledKokkos has no device-native exchange or borders path

The error `TransformView::modify ERROR: Concurrent modification of legacy host and Kokkos host views in TransformView "atom:v"` is about the **TransformView** mechanism — the LAMMPS/Kokkos abstraction used for mixed-precision storage. For `atom:v`, it maintains two conceptually separate host representations:

- **Legacy host view** — the raw `double** v` pointer accessed by non-Kokkos code
- **Kokkos host view** — the `Kokkos::View<float**>` (or double) on the CPU side, managed by Kokkos

The TransformView prohibits both being dirty simultaneously, because it can't know which one is authoritative — there is no safe sync direction to resolve the conflict.

---

## How the conflict arises: the `auto_sync + modified(Host)` pattern

The unsafe sequence that triggers it is:

```
atomKK->sync(Host, V_MASK)         // 1. Sync to host — clean state
lmp->kokkos->auto_sync = 1         // 2. Enable auto-sync mode
legacy_comm->exchange_or_borders() // 3. Accesses atom->v (double**)
                                   //    auto_sync intercepts → marks LEGACY HOST dirty
lmp->kokkos->auto_sync = 0         // 4. Restore
atomKK->modified(Host, V_MASK)     // 5. Tries to mark KOKKOS HOST dirty
                                   //    But legacy host already dirty → BOOM
```

Step 5 is the detonator. The TransformView sees two concurrent modifications from two different host-side owners with no intervening sync to reset state.

---

## Why CommKokkos avoids this (when `exchange_comm_legacy = false`)

`CommKokkos` has a complete **device-native** exchange and borders path — `exchange_device<DeviceType>()` and `borders_device<DeviceType>()`. When `exchange_comm_legacy = false` (the default), both functions return early after calling these Kokkos-native implementations:

```cpp
// CommKokkos::exchange() — line 858
if (!exchange_comm_legacy) {
    if (exchange_comm_on_host) exchange_device<LMPHostType>();
    else exchange_device<LMPDeviceType>();
    return;  // ← exits before the legacy host path
}
// Only reaches here if explicitly forced to legacy
atomKK->sync(Host, datamask_exchange);
lmp->kokkos->auto_sync = 1;
CommBrick::exchange();
lmp->kokkos->auto_sync = prev_auto_sync;
atomKK->modified(Host, datamask_exchange);  // ← only reached if exchange_comm_legacy=true
```

`exchange_device()` uses Kokkos kernels (`pack_exchange_kokkos`, `unpack_exchange_kokkos`) that work exclusively on Kokkos views. It calls `atomKK->sync(ExecutionSpace, X_MASK)` once to build the sendlist, but it **never touches the legacy `double**` pointer for `atom:v`**, so `auto_sync` never marks the legacy host view dirty. There's no `atomKK->modified(Host, ...)` at the end. The TransformView never sees the conflict.

Same logic applies to `borders_device()` in `CommKokkos::borders()`.

---

## Why CommTiledKokkos is always vulnerable

Now look at `CommTiledKokkos::exchange()` — **the entire function**:

```cpp
// CommTiledKokkos::exchange() — line 362
void CommTiledKokkos::exchange()
{
  atomKK->sync(Host, atomKK->avecKK->datamask_exchange);   // ← sync to host

  int prev_auto_sync = lmp->kokkos->auto_sync;
  lmp->kokkos->auto_sync = 1;                              // ← auto_sync ON
  CommTiled::exchange();                                    // ← legacy access to atom->v
                                                           //    auto_sync marks LEGACY HOST dirty
  lmp->kokkos->auto_sync = prev_auto_sync;

  atomKK->modified(Host, atomKK->avecKK->datamask_exchange); // ← marks KOKKOS HOST dirty → CONFLICT
}
```

There is no `exchange_comm_legacy` check. There is no `exchange_device()`. There is no early exit. **Every single call to `CommTiledKokkos::exchange()` runs this pattern unconditionally.** The `exchange_comm_legacy` flag is stored in the class (line 74 of `comm_tiled_kokkos.cpp`), but is never consulted in `exchange()`.

`CommTiledKokkos::borders()` has the exact same structure — it always calls the legacy host path with `auto_sync=1`, then calls `atomKK->modified(Host, datamask_border_vel)` which includes `V_MASK` when `ghost_velocity=true`.

---

## Why `atom:v` specifically, not `atom:x`

`atom:x` has `NEED_TRANSFORM` checks scattered throughout both comm styles. The fast-path shortcut `if (comm_x_only && !atomKK->k_x.NEED_TRANSFORM)` lets positions bypass the pack/unpack entirely when no precision transform is needed — and when a transform IS needed, the code is structured to avoid the conflict.

`atom:v` only enters the picture when `ghost_velocity = true`. When that is set:

- It's included in `datamask_border_vel`, `datamask_comm_vel`, and `datamask_exchange`
- `CommTiledKokkos::borders()` explicitly syncs and modifies it via `datamask_border_vel`
- `CommTiledKokkos::exchange()` includes it via `datamask_exchange`

The conflict becomes detectable precisely because velocity is touched during the legacy exchange/borders path (via `CommTiled::exchange()` reading `atom->v` internally with `auto_sync=1`) and then marked modified again via `atomKK->modified(Host, ...)`.

For `atom:x`, `BuildExchangeListFunctor` in `CommKokkos::exchange_device()` takes `atomKK->k_x` directly as a `ttransform_kkfloat_1d_3_lr`, bypassing the legacy pointer entirely. CommTiledKokkos has no equivalent functor path for exchange.

Here's the structural picture:

comm_kokkos_vs_tiled_bug_analysis


---

## Detailed breakdown by function

### `exchange()` — the primary trigger

`CommKokkos::exchange()` guards its entire logic behind `if (!exchange_comm_legacy)`, which defaults to `false`. When false, it calls `exchange_device<DeviceType>()` and returns. That function calls `atomKK->sync(ExecutionSpace, X_MASK)` — positions only, for building the sendlist — then calls `pack_exchange_kokkos` / `unpack_exchange_kokkos` which are Kokkos-only kernels that never dereference `atom->v` via the legacy `double**` pointer. The legacy host view of `v` is never accessed, `auto_sync` never fires for it, and `atomKK->modified(Host, ...)` is never called. The TransformView remains in a single-owner state.

`CommTiledKokkos::exchange()` is four lines without any guard. It unconditionally does `sync(Host) → auto_sync=1 → CommTiled::exchange() → modified(Host)`. The `exchange_comm_legacy` member variable is populated in `init()` but is **never consulted** in `exchange()`. There is no `exchange_device()` template defined in `CommTiledKokkos` at all — it simply doesn't exist. Every call runs the conflicting pattern.

### `borders()` with `ghost_velocity=true`

Same structural asymmetry. `CommKokkos::borders()` routes to `borders_device<DeviceType>()` when `!exchange_comm_legacy`. That function builds the sendlists with `BuildBorderListFunctor` on the device, calls `pack_border_kokkos` / `unpack_border_kokkos`, and never calls `modified(Host, V_MASK)`. `CommTiledKokkos::borders()` always goes through `CommTiled::borders()` with `auto_sync=1`, then calls `atomKK->modified(Host, datamask_border_vel)` — which includes `V_MASK` — producing the same concurrent dirty conflict.

### `forward_comm()` — symmetric, not the source

Both classes have `forward_comm_device()` and guard their legacy path with `if (forward_comm_legacy)`. The conflict isn't sourced here — it's that once `exchange()` or `borders()` has already put the TransformView into a bad state, any subsequent access that tries to resolve it may amplify the error.

---

## The `NEED_TRANSFORM` connection

The bug only manifests when `NEED_TRANSFORM = true` on `atomKK->k_v` — i.e., when the Kokkos device view stores a different precision than the legacy `double**` host view (typically single-precision GPU storage with double-precision CPU). You can see this conditioning in the fast-path guards throughout `forward_comm_device()`:

```cpp
if (comm_x_only && !atomKK->k_x.NEED_TRANSFORM) {
    // direct DMA into device x buffer — no pack/unpack needed
}
```

When `NEED_TRANSFORM = false` (pure double precision), the legacy and Kokkos host views are the same memory and the TransformView is essentially a pass-through, so "concurrent modification" of two views that alias to the same allocation is harmless. When `NEED_TRANSFORM = true`, they are genuinely separate buffers with a precision conversion between them, and the TransformView enforces the invariant that only one can be authoritative at a time.

---

## Fix path for CommTiledKokkos

The correct fix is to implement `exchange_device()` and `borders_device()` for the tiled comm style — mirroring the `BuildExchangeListFunctor` / `BuildBorderListFunctor` patterns from `CommKokkos`, adapted for the recursive bisection decomposition. The shorter workaround is to add a guard in `CommTiledKokkos::exchange()` and `borders()` that restructures the legacy path to not double-mark the velocity:

```cpp
void CommTiledKokkos::exchange()
{
  atomKK->sync(Host, atomKK->avecKK->datamask_exchange);
  int prev_auto_sync = lmp->kokkos->auto_sync;
  lmp->kokkos->auto_sync = 1;
  CommTiled::exchange();
  lmp->kokkos->auto_sync = prev_auto_sync;
  // Don't call atomKK->modified(Host,...) here if auto_sync already
  // marked the legacy host dirty — the sync at the top already established
  // a clean baseline; auto_sync handles ongoing tracking.
}
```

But the real long-term fix is the device-native path — `CommTiledKokkos` was simply never fully ported to have the Kokkos exchange/borders implementations that `CommKokkos` got, making mixed-precision use with `ghost_velocity=true` inherently broken on the tiled comm style.

sjplimp:
Hi @alphataubio - thanks for the additional details.

(1) The ADIOS support in LAMMPS was contributed by the ADIOS team.  So they will have to look at that part
(2) Stan is the one to look at the Kokkos comm part
(3) As I said, I would prefer the new feature region grid be its own PR.  And I won't look at that until there is a good description in its Summary.

It's fine if you want to hold off on these or other parts until you can reorganize the PR in the summer - thanks for understanding.

alphataubio:
## CommTiledKokkos: device-native `exchange_device` / `borders_device`

### Root cause of the bug

`CommTiledKokkos::exchange()` and `borders()` unconditionally ran the legacy host path:

```cpp
atomKK->sync(Host, datamask)        // marks Kokkos host clean
lmp->kokkos->auto_sync = 1
CommTiled::exchange/borders()       // touches atom->v via double** ptr;
                                    // auto_sync fires -> marks LEGACY host dirty
atomKK->modified(Host, datamask)    // also marks KOKKOS host dirty
                                    // => both dirty => TransformView CRASH
```

The `exchange_comm_legacy` flag was stored but **never consulted** in `exchange()` or `borders()`, and no `exchange_device()` / `borders_device()` existed for the tiled style. The crash only manifests when `NEED_TRANSFORM=true` (mixed float/double precision: float on GPU / double on CPU) **and** `ghost_velocity=true` (velocity included in `datamask_border_vel` / `datamask_exchange`), because that is the only case where `atom:v` is accessed through a TransformView.

`CommKokkos` (brick style) is immune because it guards with `if (!exchange_comm_legacy)` and routes to `exchange_device` / `borders_device`, which use only Kokkos kernels, never touch `atom->v` via `double**`, and never call `modified(Host, ...)`.

---

### Key structural difference: tiled vs brick exchange topology

| | CommKokkos (brick) | CommTiledKokkos (RCB) |
|---|---|---|
| Exchange neighbours per dim | Exactly 2 (`procneigh[dim][0/1]`) | `nexchproc[dim]` peers (`exchproc[dim][0..n-1]`) |
| Send pattern | `Sendrecv` to left + right | Broadcast to all `nexchproc[dim]` procs |
| Destination tag in buffer | Yes (proc id prepended) | No – `unpack_exchange_kokkos` filters by coordinate range |

---

### New functor: `BuildExchangeListFunctorTiled<DeviceType, BONUS_FLAG>`

Parallel scan of all `nlocal` atoms. Any atom with `x[dim] < lo || x[dim] >= hi` is added to `k_exchange_sendlist` via `Kokkos::atomic_fetch_add`. With `BONUS_FLAG=1`, simultaneously collects departing ellipsoid-bonus indices into `k_exchange_sendlist_bonus`. Semantically identical to `CommKokkos::BuildExchangeListFunctor`; duplicated here to avoid cross-TU linkage of a file-scope struct.

---

### New functor: `BuildBorderListFunctorTiled<DeviceType>`

Team-parallel 3-D box membership test for `borders_device`. Each league rank processes a slice of `[nfirst, nlast)`, counts atoms inside `sendbox[iswap][m]`, then runs `dev.team_scan()` to compute a contiguous store offset into `k_sendlist[iswap][iproc][...]`. This avoids per-atom atomics on the write path. The per-call atomic scalar `k_total_send` accumulates the total count and lets the caller detect overflow and retry with a grown list. Mirrors `CommKokkos::BuildBorderListFunctor` but adapted for the 3-D `tdual_int_3d_lr` sendlist layout used by tiled.

---

### `exchange()` dispatcher changes

- Checks `exchange_comm_legacy` (was ignored before).
- If not legacy: iterates over `atom->nextra_grow` fixes and checks each for `exchange_comm_device` support. If any fix is incompatible, emits a warning, sets `exchange_comm_legacy = true`, and falls back. Matches the same guard used in `CommKokkos::exchange()`.
- Routes to `exchange_device<LMPHostType>()` or `exchange_device<LMPDeviceType>()` depending on `exchange_comm_on_host`.
- Legacy path preserved exactly: `sync(Host)` -> `auto_sync=1` -> `CommTiled::exchange()` -> `modified(Host)`.

---

### `exchange_device<DeviceType>` — step by step

1. **Clear map / ghost state**: `atom->map_clear()` (if `atom_map_legacy`), `atom->nghost = 0`, `atom->avec->clear_bonus()`.
2. **Set sub-box pointers**: `sublo/subhi` or `sublo_lamda/subhi_lamda` for triclinic. `dimension` refreshed from domain.
3. **Sync on device**: `atomKK->sync(exec_space, X_MASK)` plus `ELLIPSOID_MASK` if ellipsoid atoms are present. Never syncs Host view.
4. **Per-dimension sendlist build** (`BuildExchangeListFunctorTiled`): retry loop doubles capacity until `k_count(0) < list.extent(0)`. Bonus sendlist grown in parallel if `ellipsoid_flag`.
5. **Sort sendlist on device**: `Kokkos::sort` on the active subview so the compaction logic is deterministic across runs.
6. **Build copylist on host**: backfill compaction — `copylist[i]` is the local atom index to copy into the hole left by departing atom `sendlist[i]`, or `-1` if that atom was already at the end of the array. Synced to device after construction.
7. **Build bonus copylist** (ellipsoid only): same backfill logic over the bonus index space. Uses a temporary `i2recv[]` map (local atom index -> position in the exchange sendlist) to align bonus copylist entries with atom copylist entries.
8. **`pack_exchange_kokkos`**: compacts the local atom array in-place and fills `k_buf_send`. May resize `k_buf_send`. `atom->nlocal` and `nlocal_bonus` set to the post-compaction counts.
9. **All-to-all size handshake** across all `nexchproc[dim]` exchange procs: each proc Irecvs `exchnum[dim][m]` from all peers, then sends its own `nsend` to all peers. `MPI_Waitall` before proceeding.
10. **Post all Irecvs** into contiguous slots of `k_buf_recv`, then **broadcast** the packed send buffer to every exchange proc (`MPI_Send` loop). `MPI_Waitall` + `DeviceType().fence()`.
11. **`unpack_exchange_kokkos`** with `(dim, lo, hi)`: accepts only atoms whose coordinate in `dim` falls in `[lo, hi)`. Atoms outside continue to the next dimension's exchange, replicating the semantics of `CommTiled::exchange()` without the legacy-host path.
12. **`nextra_grow` fix loop**: for each fix, repeats the size handshake + broadcast with the fix's `pack_exchange_kokkos` / `unpack_exchange_kokkos`. Treats all received data as a single contiguous block (no two-sided split, because tiled has no brick-style left/right distinction).
13. **`firstgroupname` fallback**: if `atom->firstgroupname` is set, syncs to host, calls `atom->first_reorder()`, marks host modified. Same fallback as `CommKokkos`.

**Critical invariant**: neither `modified(Host, ...)` nor `auto_sync` is set at any point in the device path. TransformView never sees two concurrent dirty owners.

---

### `borders()` dispatcher changes

- Checks `exchange_comm_legacy` (was ignored before).
- If not legacy: checks `atom->nextra_border || mode != Comm::SINGLE || bordergroup`. If any is true, emits a warning and falls back to legacy. Matches `CommKokkos::borders()` fallback policy exactly.
- Routes to `borders_device<LMPHostType>()` or `borders_device<LMPDeviceType>()`.
- Legacy path: `k_sendlist.sync_host()` -> `sync(Host, datamask)` -> `auto_sync=1` -> `CommTiled::borders()` -> `k_sendlist.modify_host()` -> `modified(Host, datamask)`.

---

### `borders_device<DeviceType>` — step by step

1. **Rebuild collection array** if `mode == MULTI` (already gated out by dispatcher fallback; kept for safety).
2. **Sync atom data on device**: `atomKK->sync(exec_space, datamask_border)` or `datamask_border_vel` when `ghost_velocity`. Never touches Host view.
3. **`k_sendlist.sync<DeviceType>()`** before any functor writes.
4. **Per-swap loop** over `iswap = 0..nswap-1`:
   - `nlast` updated on even swaps (new dimension start) to `nlocal + nghost` — same logic as `CommTiled::borders()`.
   - **Per send-proc inner loop** over `m = 0..nsendproc[iswap]-1`:
     - Extracts `sendbox[iswap][m]` bounds (`xlo/ylo/zlo/xhi/yhi/zhi`).
     - Runs `BuildBorderListFunctorTiled` via `TeamPolicy` with retry-on-overflow: if `k_total_send > maxsendlist[iswap][m]-1`, calls `grow_list(iswap, m, nfound)` and retries.
     - Sets `sendnum[iswap][m]`, updates `smaxone` / `ncountall`.
   - **`sendnum` exchange**: `MPI_Irecv` + `MPI_Send` + `MPI_Waitall` for all non-self recv procs; self-copy for `sendself[iswap]`.
   - **Accounting arrays**: fills `size_reverse_recv`, `reverse_recv_offset`, `size_forward_recv`, `size_reverse_send`, `firstrecv`, `forward_recv_offset` for every recv proc.
   - **Buffer sizing**: grows `k_buf_send` / `k_buf_recv` if current capacity insufficient.
   - **Post all Irecvs** into `k_buf_recv` at the correct per-proc offset.
   - **Send loop** (`sendother`): `pack_border_kokkos` or `pack_border_vel_kokkos` per proc, then `MPI_Send`.
   - **Self copy** (`sendself`): pack directly into `k_buf_send`, then unpack without MPI.
   - **`Waitany` unpack loop** (`recvother`): unpacks each message as it arrives via `unpack_border_kokkos` / `unpack_border_vel_kokkos`.
   - **`atom->nghost` increment** by the total atoms received this swap.
5. **Post-loop buffer resize** for forward/reverse comm headroom (`maxforward * smaxone`, `maxreverse * rmaxone`, etc.).
6. **`k_sendlist.modify<DeviceType>()`** to mark sendlists dirty on device so a later `sync_host()` in legacy pass-throughs is correct.
7. **`atomKK->modified(exec_space, datamask_border[_vel])`** — marks atom data dirty on the **device** execution space, **not** on Host. This is the fix: TransformView stays single-owner.
8. **`atom->map_set()`** if map style is not NONE.

---

### New header declarations (`comm_tiled_kokkos.h`)

```cpp
template<class DeviceType> void exchange_device();
template<class DeviceType> void borders_device();
```

New protected data members:

```cpp
DAT::tdual_int_1d     k_exchange_sendlist;        // atom indices leaving box this dim
DAT::tdual_int_1d     k_exchange_copylist;        // backfill copy targets
DAT::tdual_int_1d     k_exchange_sendlist_bonus;  // ellipsoid bonus indices leaving
DAT::tdual_int_1d     k_exchange_copylist_bonus;  // ellipsoid bonus backfill targets
DAT::tdual_int_1d     k_count;                    // (0)=atoms to send, (1)=bonus to send
DAT::tdual_int_scalar k_total_send;               // atomic counter for BuildBorderListFunctorTiled
DAT::tdual_int_1d     k_indices;                  // atom-index map for fix unpack_exchange_kokkos
class AtomKokkos     *atomKK;                     // set in init()
```

All new Kokkos views allocated with initial capacity 100 / `BUFMIN` in both constructors.

---

### `init()` fix

`atomKK = (AtomKokkos *) atom` is now set unconditionally in `init()`. Previously this assignment was missing entirely — `atomKK` was always null and would have segfaulted on any device-path call.

---

### Known limitations / explicit fallbacks

| Condition | Behaviour |
|---|---|
| `line_flag \|\| tri_flag \|\| body_flag` | `error->all` in `exchange_device` — same limitation as CommKokkos brick |
| `nextra_border \|\| mode != SINGLE \|\| bordergroup` | `borders()` falls back to legacy with warning |
| `nextra_grow` fix without `exchange_comm_device` | `exchange()` falls back to legacy with warning |
| `atom->firstgroupname` | `exchange_device` syncs to host, calls `first_reorder()`, marks host modified — same as CommKokkos |

akohlmey:
> It's fine if you want to hold off on these or other parts until you can reorganize the PR in the summer - thanks for understanding.

In that case, I suggest to close the pull request, since it is not mergeable in this form anyway, and then resubmit the (related groups of) changes as multiple pull requests when they are ready. This helps to keep our rather clogged pull request queue leaner and easier to review and process. Thanks.

akohlmey:
> C++20 modernization

Our currently required minimum C++ standard level is C++17 and we would like to keep it like that as much as possible. Currently KOKKOS is the exception, since the minimum standard is required by the Kokkos library, but changes that raise the C++ standard level without need and thus make a feature unavailable to LAMMPS users stuck with older compilers are not welcome.

src/KOKKOS/min_fire_kokkos.cpp
```diff
@@ -129,7 +128,11 @@ int MinFireKokkos::run_iterate(int maxiter) {
   int nlocal = atom->nlocal;
 
   if constexpr (INTEGRATOR == LEAPFROG) {
-    energy_force(0);
+    atomKK->k_x.sync_host();
```
stanmoore1: @alphataubio this doesn't look quite right:
- You shouldn't call `atomKK->k_x.sync_host()` and similar directly, you need to use `atomKK->sync(Host, X_MASK)` since it goes through some wrappers.
- You should only sync data that you are actually using. So in this case you would sync to device, not host. Comm/tiled/kk or whatever that is actually using the data on host should do the sync there.

Issue #4899, opened 2026-02-26
[Feature Request] (write/read)_restart flexibility


Matthew-Whitlock:
## Summary

The largest supercomputers are increasingly fault-prone as time goes on, and they often experience incredibly high failure rates immediately after being built. LAMMPS's direct-to-file checkpoint/restart can become a large performance drain when running on these systems. Modern HPC libraries can manage checkpoint/restart **much** more efficiently. However, the various read_restart and write_restart functions for LAMMPS's styles are currently too rigid to conveniently implement checkpoint/restart libraries as LAMMPS modules.

My suggestion is to begin migrating to new (optional) write_restart and read_restart functions that:
1. Write to LAMMPS's FileWriter virtual class
2. Avoid communication and decision-making in the individual styles during checkpoint/restart

## Detailed Description

Using FileWriter instead of directly passing a file pointer allows customizing the checkpoint mechanism without touching the individual styles. Examples:
* The [Magistrate](https://github.com/DARMA-tasking/magistrate) serialization library avoids buffer overflows and reallocations with a two-pass approach . A simple implementation would be having one FileWriter implementation that tallies up the size of writes but ignores the data and a second that writes to a pre-allocated buffer sized according to the first pass.
* [VeloC](https://github.com/ECP-VeloC/VELOC) is a modern checkpointing library designed for supercomputing and datacenters. A Filewriter implementation could be used to write directly into VeloC's memory buffers.
* [Fenix](https://github.com/sandialabs/fenix) enables distributed in-memory checkpoints that greatly outscale filesystem checkpoints. The two-pass implementation above would allow LAMMPS to use Fenix for checkpoint/restart. Note: this is my motivation, see #4652 with ~800 LOC and some hacky code to implement a similar feature in `FENIX/local_serializer.(cpp|h)`

Avoiding communication in the checkpoint read/write steps makes checkpoint/restart even more flexible. It is also a QOL improvement for style developers who will no longer need to manually manage gathering/broadcasting.

The primary decision-making to be taken out of individual styles is the typical `if (rank == 0)` check. Since communication doesn't happen in the write/read functions, each checkpointing module can choose which rank(s) write each type of data for themselves. For instance, the Fenix module in the PR linked above has every rank checkpoint its global data to improve recovery scalability. This is, of course, a decision with tradeoffs depending on the size of a given run's global data.

Suggested API demonstrated on fixes:
```cpp
[...]

PR #4918, opened 2026-03-16
(write|read)_restart flexibility

Matthew-Whitlock:
**Summary**

Add an optional FileWriter-based write_restart and read_restart API to Fixes, 

**Related Issue(s)**

Demo for #4899

**Author(s)**

Matthew Whitlock, Sandia National Laboratories, matthew@mwhitlo.dev

**Licensing**

**Backward Compatibility**

This should maintain backward compatibility.

**Implementation Notes**

There is a new Restartable class that restartable LAMMPS objects should extend. The Restartable class defines the global vs local restart functions, leaving the peratom functions as-is. Restartable also defines the default shims for callers using the old all-in-one FILE* api on objects that implement the new global/local restart functions.

There are four new "FileWriter" types: 
 1. A FileWriterSizer that just counts up the size of things "written" to it
 2. A FileWriterBuffer that serialized to a provided buffer
 3. A FileWriterWrapper that writes to a provided FILE* without taking owership
 4. A FileWriterRootWrapper that works just a FileWriterWrapper but ignores writes from ranks other than rank 0, as a convenience.

FileWriter gained a few new functions. Two (very simple) template functions to simplify writing typed data, and two functions for writing the size of a given 'Restartable' object. Making virtual functions for writing the size of the objects is not entirely necessary (callers could just make a FileWriterSizer themselves), but it simplifies things for the typical caller and will help avoid problems with sizing the same object three times if sizing-then-writing an object with nested objects.

There are new "BufferReader" types:
1. BufferReader: Simply takes an input 'untyped' char* data array and offers convenience functions for reading consecutive typed data values from it
2. BufferReaderFile: Uses a file as the input data buffer, instead of an array.
3. BufferReaderRootFile: Rank 0 uses a file as the input data buffer and broadcasts each read to remote ranks.

**Post Submission Checklist**


src/fix.cpp
```diff
@@ -143,6 +147,95 @@ void Fix::init_flags()
...
+    pos += sizeof(size_t);
+    pos += local_size*comm->me;
+
+    // This rank may be higher than nprocs when data was written
+    if (comm->me >= nprocs) local_size = 0;
+
+    this->read_restart_local(local_size, buf + pos);
+  }
+}
+
```
Matthew-Whitlock: This section showcases the usage of the new FileWriter types and functions. This is how we ensure new contributors don't have to implement the old restart API if they don't want to.

src/fix_adapt.cpp
Matthew-Whitlock, line 1: Here's a demo of how a fix would update to the new API, if/when desired.

Matthew-Whitlock:
This now looks like it requires a lot of changes, but most of them are not hard requirements. The majority of the changed lines of existing code are in read/write_restart - I rewrote their functions in terms of FileWriter and BufferReader to allow checkpoint modules to use those functions directly. That way checkpoint modules don't all need to be updated individually if new flags or features are added. I did the same for the restart functions for LabelMap and Group, for the same reason.

But if this is looking like too much code change all at once, I am happy to revert those changes and leave it to checkpoint modules to manage directly.

akohlmey:
@athomps @sjplimp I think we should have a serious look at this together when I come over in a couple of weeks.

@Matthew-Whitlock Thanks for creating this branch to demonstrate what you have in mind. This will make it *much* easier to discuss. I like the concept and where this could be heading, but I am also concerned that this may be one step too far for us given that at the moment we are lacking workforce more than before. I don't want to have to get into a situation similar to what we have with supporting two build systems and having a very, very slow transition that makes life harder for everybody involved. I would rather prefer to have one big disruptive change and then move on.  But then again, I am only one of many and we are always looking for as large a majority as possible when making big decisions.

akohlmey:
@Matthew-Whitlock we discussed this at the meeting yesterday and have converged to the view that we currently are not interested in this kind of change. There are too many other refactoring tasks we are looking into that have more immediate benefits and we don't have the capacity, not even to review this properly. We may reconsider in a few years again.

PR #4924, opened 2026-03-20
Fix mliap/pytorch.py and move model to the GPU device

canoalberto:
**Summary**

Fixes error of model living on the CPU side rather than GPU side, which triggered error:

RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0! (when checking argument for argument mat1 in method wrapper_CUDA_addmm)

reproducible using:

mpirun -np 1 lmp -k on g 1 -sf kk -pk kokkos newton on neigh half -in in.mliap.pytorch.Ta06A

**Author(s)**

Alberto Cano, Virginia Tech Advanced Research Computing

**Licensing**

**Backward Compatibility**

No known issues.

**Implementation Notes**

Added an explicit instruction to move the model to the device when the GPU is in use.

**Post Submission Checklist**

- [X] Suitable new documentation files and/or updates to the existing docs are included
- [X] The added/updated documentation is integrated and tested with the documentation build system

### DIFF
--- a/python/lammps/mliap/pytorch.py
+++ b/python/lammps/mliap/pytorch.py
@@ -135,6 +135,7 @@ class TorchWrapper(torch.nn.Module):
         device = self.device
         if (use_gpu_data and (device is None) and (str(beta.device).find('CUDA') == 1)):
             device = 'cuda' #Override device as it wasn't defined in the model
+            self.model = self.model.to(device)
         with torch.autograd.enable_grad():
 
             if (use_gpu_data):


akohlmey:
@bathmatt I think you changed this chunk of code last. Can you please have a look and let us know if this is something we should merge?

bathmatt:
I'm on vacation but will look end of the month. Sorry for the delay

stanmoore1 (approved):
It looks OK to me, but would be good for @bathmatt to review.

bathmatt:
THis looks like a good fix to prevent things on CPU.

PR #4938, opened 2026-04-02
More useful bin output for multi

jtclemm:
**Summary**

Previously, LAMMPS would print the # of bins in the largest collection for neighbor style multi. Now it prints the number and size of bins for all collections. Also fixed some whitespace.

**Author(s)**

Joel Clemmer(SNL)

**Licensing**

### DIFF
--- a/src/neighbor.cpp
+++ b/src/neighbor.cpp
@@ -1791,7 +1791,7 @@ void Neighbor::print_pairwise_info()
   out += fmt::format("  master list distance cutoff = {:.8g}\n", cutneighmax);
   out += fmt::format("  ghost atom cutoff = {:.8g}\n", cutghost);
   if (style != Neighbor::NSQ && style != Neighbor::MULTI) {
-    if (domain->dimension == 2) nbinz = 1
+    if (domain->dimension == 2) nbinz = 1;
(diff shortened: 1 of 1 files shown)


sjplimp:
@jtclemm I can approve this PR, but the Copilot comment about 2d vs 3d seems legit ?

jtclemm:
> @jtclemm I can approve this PR, but the Copilot comment about 2d vs 3d seems legit ?

Agreed, I will make that swap for 2D.

jtclemm:
> @jtclemm I can approve this PR, but the Copilot comment about 2d vs 3d seems legit ?

Done.

On a related note, this PR emerged from trying to go to even larger size distributions (diameter ratios > 1:100) where the number of bins starts to become the limiting factor (namely the `NBin->binhead` array becomes huge, but it is extremely sparse). 

Do you have any thoughts on adding an option to replace this array and the linked list with a hash table to save memory? If so, I was thinking this is only needed when you have very low particle densities (e.g. when you have large polydisperisty) so probably only needs to be implemented in *multi*.

Thoughts/concerns with me pursuing this?

sjplimp (approved):

PR #4951, opened 2026-04-17
Add fix graphics chunk command to show convex hulls of chunks

akohlmey:
**Summary**

This pull request implements a new fix graphics/chunk command that will compute a triangulated hulls for chunks using a Delaunay triangulation with alpha shape extraction.  The hull can be rendered as either a triangle mesh with rounded or flat triangles or a wireframe mesh. How strongly the alpha shape extraction is applied can be controlled with an optional "alpha" argument. The default is to use an automatic estimate. In addition flags exist to control how many positions are used in the point cloud, to restrict the point cloud to a region, and to clip of all points outside the box.

The same hull triangulation code is also used to approximate region styles "union" and "intersect" for draw styles "frame", "filled", and "transparent". In the image below the white object is the intersection between a plane (green) and a cylinder (blue).

Possible improvements:
- [x] store the per-atom/per-type radius, if available and use that for adjusting the outside of the hull

**Related Issue(s)**

Implements features from issues #4844 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

The core implementation of the convex hull based triangulation was created by Claude Opus 4.6. The generation of the point cloud and the application of the hull triangulation to regions was added manually. The improved triangulation was implemented by Claude Opus 4.6 again.

**Backward Compatibility**

N/A

**Implementation Notes**

N/A

**Post Submission Checklist**

-----

https://github.com/user-attachments/assets/8bcdeb6b-3da9-410a-bd3b-47a8bc9c7652

https://github.com/user-attachments/assets/ab66e2f0-7dc9-434b-8503-93e4e5f4ff14

https://github.com/user-attachments/assets/64b666cc-10a5-4c67-85ca-25a485d38651

image

image


akohlmey:
@jtclemm if you have any ideas where we should take this pull requests and/or have some example case to use this on (and possibly improve it), please let me know. Either here or via slack.

sjplimp:
@akohlmey What does this look like if chunks are spatial bins ?

Also how do these kinds of operations work in parallel?  Is each proc creating the portion of the image that is for the portion of the chunk or region which is within its sub-domain?  And the images naturally combine into one coherent image, like is the case for distributed atoms?  Or is this more of a serial operation?

akohlmey:
> @akohlmey What does this look like if chunks are spatial bins ?

Depends on the amount and kind of atoms in each bin. It will look a bit like the example of the half cylinder region  from the intersection of cylinder and plane. However, in that case, I can choose how many (random) points are used to fill the space to get a smoother surface. With the fix you also have options to tweak how the hulls are constructed. I plan to add more options to address needs when people ask about it.

> Also how do these kinds of operations work in parallel? Is each proc creating the portion of the image that is for the portion of the chunk or region which is within its sub-domain? And the images naturally combine into one coherent image, like is the case for distributed atoms? Or is this more of a serial operation?

I have experimented a bit and not found a single choice that gives good results all the time. I actually had to ask the AI to implement an improved algorithm since the original, strictly convex hull would include too much space not containing atoms for irregular shapes. There is a lot of flexibility in how you can control the shape and some of that is already exposed to keywords.

When running in parallel, each subdomain will compute its own hulls and then they will be merged with the usual z-buffer based merge of images. That should join the objects in the final image, but how well that works depends on the choice of the alpha parameter.

There is one idea for an alternate approach that I have not yet implemented:
- allocate an integer array for local and ghost atoms
- copy the chunk-ID to that array
- use a forward communication to forward the chunk-ID data to ghosts
- construct the hull, but discard all triangles that are fully outside the subdomain

This is how graphics/isosurface handles the grid points and it works ok for some cases, but when running in serial and the chunk straddles a box boundary, the chunk would have to be split into two and I have not yet found a good way to implement this. In case I do (or perhaps I should ask the AI?), then I would add another keyword so that you can choose between the strategy for handling parallel execution and periodicity.

akohlmey:
> @akohlmey What does this look like if chunks are spatial bins ?

@sjplimp would you like to have a visualization for the bins themselves? E.g. as transparent walls? 

That would be something that could be added as a Compute::image() function in compute chunk/atom. I can see that that could be useful for debugging an input, similar to visualizing regions.

sjplimp:
@akohlmey Thanks for the explanations.  For spatial bin chunks, the transparent planes would probably be more useful.  But maybe others will like to see the chunks of atoms in each bin.

I will approve this PR.

sjplimp (approved):

PR #4952, opened 2026-04-17
KSPACE: Fix LAMMPS/Kokkos/Host threaded FFTs; reintroducing self-copy in remap collectives (sometimes)

weinbe2:
**Summary**

This PR addresses a few performance dings in the KSPACE package affecting single-node multi-threaded configurations with MPI collectives enabled.

1. Host-side FFTs (FFTW3 and NVPL) driven through LAMMPS/Kokkos now utilize threading when available (small oversight in #4007).
2. Self-copy has been re-introduced as a new `kspace_modify collective/self/copy` option that's only relevant when the collective path has been introduced. This was added because self-copy is still empirically beneficial in the single-rank case.

The new `kspace_modify` options are:

```
kspace_modify collective/self/copy no       # develop behavior: self-data in MPI_Alltoallv
kspace_modify collective/self/copy yes      # always use direct pack+unpack for self
kspace_modify collective/self/copy onerank  # (default) self-copy only when nranks == 1
```

Details/motivations are given below; please note that I'm not married to any specific naming convention.

**Related Issue(s)**

Iterative improvements on top of #4007 and #4143.

**Author(s)**

Chris Daley @cdaley (NVIDIA, chrisdaley82@gmail.com), Chris Parrott @cparrott73 (NVIDIA, chris@parrotts.net), Evan Weinberg (NVIDIA, evansweinberg@gmail.com)

The real investigative work and the suggested changes come from Chris & Chris, I just made some implementation decisions.

**Licensing**

By submitting this pull request, I confirm that I did NOT use any AI tools to generate all or parts of the code and modifications in this pull request.

(Fwiw, AI was used to orchestrate a lot of _additional_ testing.)

**Backward Compatibility**

The default behavior for 1x MPI rank runs when using `kspace_modify collective` is to do a self-copy (which was the case before #4143). For runs with >1 MPI ranks, it includes "self" data in `MPI_Alltoallv` (as was identified to be better at scale in #4143). I'm not married to this convention, this was just a "best of both worlds" idea.

**Implementation Notes**

The various self-copy modes are now encoded into the existing `collective_flag` integer:

 * 0 -> point-to-point (consistent with old behavior)
 * 1 -> collective w/out selfcopy (consistent with old behavior, including self in `MPI_Alltoallv`
 * 2 -> collective w/selfcopy (consistent with old-old behavior, always having the self copy independent of the number of MPI ranks)
 * 3 -> collective w/selfcopy for 1x rank runs only (the new default, only using self copy for 1x rank runs)

This maintains the behavior where 0 -> false for no collectives, 1 to 3 -> true for use of collectives, which avoided changes across multiple files that check `collective_flag`. This isn't a convention I'm married to... it just works.

Correctness was checked by looking at the thermodynamic output of the rhodo benchmark replicated to 64,000 atoms + 10 timesteps + `kspace_modify collective yes`. This was done over the outer product of the 3 `collective/self/copy` modes with one or multiple MPI ranks. In all cases we got consistent results:

| Step | TotEng | KinEng | Temp |
|------|--------|--------|------|
| 0 | -50713.7719 | 42889.6626 | 299.0335 |
| 5 | -50717.1479 | 42997.9710 | 299.7886 |
| 10 | -50712.0493 | 42896.7753 | 299.0831 |

**Post Submission Checklist**

For reference, this is performance on a single-socket AMD EPYC 9684X (Genoa), 96 OpenMP threads, 1 MPI rank; GNU 15.1.0, FFTW 3.3.8 (double precision, OpenMP threading). Numbers are an average over two runs.

| Self-copy | Multithreaded FFT | Performance |
|-----------|-------------------|-------------|
| No  | No  | 702  |
| No  | Yes | 1071 |
| Yes | No  | 1041 |
| Yes | Yes | 2951 |

Both fixes together yield a ~4.2× improvement over the baseline.


stanmoore1:
Thanks @weinbe2! I will take a look.

stanmoore1 (approved):
Looks good to me, thanks @weinbe2!

stanmoore1:
@hagertnl FYI, do you want to review this?

hagertnl:
Sure! I can take a deeper look in about 2 hours, but an initial question I have is, if the self/copy is implemented primarily for benefitting the single-rank case, is there a reason to enable the collectives vs the default pt2pt, which does still do self-copy? Is it just so that one consistent input script can be used? I had done some performance analysis when implementing this initially, and at scale, separate self-copy seemed to have no benefit, though I only had access to Frontier (cray-mpich on AMD GPUs) for that.

src/kspace.cpp
```diff
@@ -56,11 +56,12 @@ KSpace::KSpace(LAMMPS *lmp) :
   // default to using MPI collectives for FFT/remap only on IBM BlueGene
 
 #ifdef __bg__
-  collective_flag = 1;
```
hagertnl, line 59: I think this is changing the default only for IBM BlueGene. The real default is 0 (pt2pt), 2 lines below.

src/kspace.cpp
```diff
@@ -56,11 +56,12 @@ KSpace::KSpace(LAMMPS *lmp) :
...
+  collective_flag = 3;
```
hagertnl: I think this is changing the default only for IBM BlueGene. The real default is 0 (pt2pt), 2 lines below.

src/kspace.cpp
hagertnl: This was accidentally commented on the old file, not new. Dis-regard, see next comment.

hagertnl:
I can performance-check this tomorrow, but I don't see anything technically incorrect with it (minus the default setting to 3 for blue gene instead of regular default)s. I think my initial question surrounding why the default pt2pt self-send backend isn't sufficient still stands.

Very happy about the multi-threaded FFT work though, apologies if that got broken during the FFT_KOKKOS break-out.

src/kspace.cpp
weinbe2: I had to stare at this a bit, but this is technically correct... in reality, it's not even necessary, due to the line:

```
  // Encode the self-copy mode into `collective_flag`
  if (collective_flag) collective_flag = selfcopy_flag + 1;
```

That line basically says _if_ we're doing collectives at all, _then_ the default is the `onerank`.

In that case, everything is still consistent: the default in BlueGene is still collectives (since `if (collective_flag)` w/`collective_flag == 3` evaluates to true), and the default otherwise is pt2pt (and `selfcopy_flag`) is ignored.

weinbe2:
> [...] if the self/copy is implemented primarily for benefitting the single-rank case, is there a reason to enable the collectives vs the default pt2pt, which does still do self-copy? Is it just so that one consistent input script can be used?

One consistent input script + out-of-the-box improvement, yes.

src/kspace.cpp
weinbe2: All that said---I'll change it back to `1`, exactly because it makes no difference if it's `1` or `3`!

hagertnl:
For single-rank cases, should we just universally force a self/copy behavior (possibly by forcing pt2pt implementation for nranks=1), instead of add a tunable for it? KSPACE has a high concentration of tunables (for good reason, lots of opportunity for improvement), but this case especially seems borderline common-sense and able to just implement without a command-line arg?

src/kspace.cpp
akohlmey: I actually think we should completely remove the special case code for BlueGene. According to Wikipedia, those have been discontinued over 10 years ago.

weinbe2:
> For single-rank cases, should we just universally force a self/copy behavior (possibly by forcing pt2pt implementation for nranks=1), instead of add a tunable for it? KSPACE has a high concentration of tunables (for good reason, lots of opportunity for improvement), but this case especially seems borderline common-sense and able to just implement without a command-line arg?

I think our combined experiments show that being able to toggle `self/copy` at runtime is useful. Our experiments have shown that, at one rank, using `self/copy` is beneficial. Your experiments have shown that, at scale, not using `self/copy` is beneficial. I doubt the general cross-over is going from one rank to two ranks. Power users will tune it. Others won't.

That said: `onerank` isn't perfect as a concept, I won't pretend this is the cleanest thing in the world. I just wanted to introduce a default behavior that out of the box preserves what's been experimentally verified.

hagertnl (approved):

hagertnl:
Sounds fine by me! Just wanted to gauge your interest in making it default for one rank without a tunable. Understood about 2 ranks likely isn't the crossover point. I believe the docs say that collectives aren't generally beneficial at smaller scales, that's my recollection over a long time of staring at KSPACE's docs page/tunables.

@stanmoore1 I'm good if you'd like to merge this.

PR #4965, opened 2026-04-28

### LINKED ISSUE #4963: [BUG] Fix print broken when using file-style variable along with other equal-style variables.
**Summary**

I have the file “steps.dat” which lists timesteps:

```
# time steps to print
1
6
8
12
91
```

The following input file:

```
units           lj
atom_style      atomic
dimension       2

lattice         sq 0.4
region          box block -30 30 -30 30 -0.2 0.2
create_box      1 box
create_atoms    1 box
mass            * 1.0

variable tfile file steps.dat
variable step equal next(tfile)
fix stepper all print v_step "HI FROM STEP"

variable        s equal logfreq(10,3,10)
fix             extra all print v_s "HI there"

run             90
```

yields the error:

`ERROR: Fix print timestep variable returned a bad timestep: 0 (src/fix_print.cpp:174)`

using the February 11th, 2026 patch of LAMMPS. Upon further inspection, this error is happening because the variable `tfile` is being deleted, meaning that the order of the variables stored internally by lammps changes. Since fix_print.cpp doesn’t internally update the order of the variables at each step (it is stored in `ivar_print` in the `init()` method but never updated later), it calls the wrong variable at time steps after tfile is deleted.I think I would read this as being a bug. Thoughts? There is a simple way to correct this, which is to add the line

LAMMPS (11 Feb 2026) on MacOS

**Expected Behavior**

I think that there should not be an error here. Instead, the script should be able to continue to print.

**Actual Behavior**

Error given above.

**Steps to Reproduce**

run the input script above (with the "steps.dat" file in the same location).

I think this bug will be very easy to fix. In `fix_print.cpp`, add the line `ivar_print = input->variable->find(var_print);` in between lines 171 and 172.

### LINKED ISSUE #4976: [BUG] LAMMPS dies with "Failed to allocate" error if "fix vector" is between two "run" commands and "write_restart" command is issued after the second run
**Summary**

If there is a "fix vector" command in between two "run" commands in the LAMMPS input file, and there is a "write_restart" command after the second "run" command, then the second "run" command will fail with the error "Failed to allocate" followed by a negative number of bytes.

This happens even if the "run" commands have arguments that satisfy the following from the "fix vector" documentation: 

> Note that if you perform multiple runs, using the “pre no” option of the [run](https://docs.lammps.org/run.html) command to avoid initialization on subsequent runs, then you need to use the stop keyword with the first [run](https://docs.lammps.org/run.html) command with a timestep value that encompasses all the runs. This is so that the vector or array stored by this fix can be allocated to a sufficient size.

This happens on the latest git version of LAMMPS as of May 5 (commit a793f279563d0a06e538bb72189ae5106a5f8a9c on the "develop" branch), but I've seen this behavior on LAMMPS versions at least as far back as commit a7af0fc0782390030163c5785f19e44cc8cf47ad from January 20, 2025.

**Steps to Reproduce**

Here's an example input file that triggers the issue:
```
# 3d Lennard-Jones melt

units           lj
atom_style      atomic

lattice         fcc 0.8442
create_box      1 box
create_atoms    1 box
mass            1 1.0

velocity        all create 3.0 87287 loop geom

pair_style      lj/cut 2.5

neighbor        0.3 bin
neigh_modify    every 20 delay 0 check no

fix             1 all nve

thermo          50
run             125 stop 250

compute         myTemp all temp
fix             2 all vector 1 c_myTemp

run             125 pre no stop 250
write_restart   test.restart
```

Put the contents of this into a file named "in.melt" and then run `/path/to/lammps/bin/lmp -in in.melt`. This will produce the following error message: "ERROR on proc 0: Failed to allocate -992 bytes for array vector:vector (src/src/memory.cpp:64)"

I've attached the logbook from a DDT run: [logbook.html](https://github.com/user-attachments/files/27522040/logbook.html)

I notice that from the results of the tracepoints in that DDT run that at the end of `Run::command()`, `update->endstep` is reset to 0. When `write_restart` causes `FixVector::init()` to be executed, the calculation of `ncountmax` in that member function -- which depends on `update->endstep` -- apparently does not account for this resetting, causing it to become negative.

akohlmey:
**Summary**

Collected small changes and fixes.

**Related Issue(s)**

Fixes #4963 and fixes #4976

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

Yes.

**Implementation Notes**

- Add deprecation warnings to `src/Makefile` and the docs about retiring the legacy build system
- Deleting a variable will no longer compress the list of variables. Instead a new variable type UNASSIGNED is introduced that flag slots in the variable arrays that can be re-used
- Update various outdated references to the LAMMPS manual in README files in the source and examples folders.
- Change regression test workflows to be only run manually to reduce the chance our workflows are throttled
- Address issues reported by Coverity Scan static code analysis
- Ensure that `ncountmax` in fix vector is > 0 so there are no memory corruption, division by zero or out-of-memory issues 
- Remove some special purpose code for IBM BlueGene style machines
- Add support for Sphinx up to version 9.1.x and upgrade MathJAX to version 4.1.2
- Small documentation fixes 

**Post Submission Checklist**


doc/src/pair_mgpt.rst
```diff
@@ -140,12 +136,17 @@ before proceeding to more complex simulations.
 
 .. note::
 
-   For good performance, LAMMPS should be built with the compiler
-   flags "-O3 -msse3 -funroll-loops" when including this pair style.  The
-   src/MAKE/OPTIONS/Makefile.mgptfast is an example machine Makefile with
-   these options included as part of a standard MPI build.  Note that it
-   as provided, it will build with whatever low-level compiler (g++, icc,
-   etc) is the default for your MPI installation.
+   For good performance on x86 architecture machines , LAMMPS should be
```
jtclemm: extra space: "machines ,"

doc/src/pair_mgpt.rst
```diff
@@ -140,12 +136,17 @@ before proceeding to more complex simulations.
...
+   built with the compiler flags "-O3 -msse3 -funroll-loops" when
+   including this pair style.  The src/MAKE/OPTIONS/Makefile.mgptfast is
+   an example machine Makefile with these options included as part of a
+   standard MPI build.  Note that it as provided, it will build with
```
jtclemm: "it is not (?) provided" ?

src/Makefile
```diff
@@ -297,7 +297,19 @@ gitversion:
...
         else mv ${TMPNAME}.lmpgitversion lmpgitversion.h ; fi
 
-# Build LAMMPS in one of 2 modes
+# deprecation warning
+
+warning:
+	@echo 'WARNING-WARNING-WARNING-WARNING-WARNING-WARNING-WARNING-WARNING-WARNING-WARNING'
+	@echo 'WARNING:'
+	@echo 'WARNING: The LAMMPS developers plan to retire the legacy build system'
+	@echo 'WARNING: Please use CMake instead. See https://docs.lammps.org/Build_cmake.html'
```
jtclemm, line 306: Might it be useful to temporarily have a quick reference table in the documentation somewhere (Howto swap from make to cmake?) that can be directly linked to that describes how to switch common make commands to cmake? Could alleviate some frustration for those making the transition. E.g. from AI:

Image

jtclemm (approved):

src/Makefile
akohlmey: Yes, this is a good idea. I've added something along those lines.

jtclemm: Great, that looks very helpful. I'll probably refer to it myself too. Some of the cmake versions of commands I can just never remember.

PR #4972, opened 2026-05-06
molecule: add force field coefficients support to molecule templates

salamaaya:
**Summary**

Continuing with our objective to turn molecule templates into standalone input files, we add capabilities to molecule templates for reading in force field coefficients.

For both legacy and JSON file formats, we add sections to define per-atom masses and coefficients for atom types, bond types, angle types, dihedral types, and improper types, including all class2 coefficients. This was tested on a `fix/bond/react` reaction input deck to confirm that we get the same results. We tested both the legacy and JSON file formats and updated the JSON example to make use of this feature. 

**Related Issue(s)**

**Author(s)**

Aya Salama ([asalama0204@gmail.com](mailto:asalama0204@gmail.com), Stevens Institute of Technology)
Jacob Gissinger ([jgissing@stevens.edu](mailto:jgissing@stevens.edu), Stevens Institute of Technology)

**Licensing**

**Backward Compatibility**

**Implementation Notes**

- The molecule reader now accepts coefficient/mass sections, not just topology/atom sections. In particular:

  - `Atom Type Masses`
  - `Pair Coeffs`, `Bond Coeffs`, `Angle Coeffs`, `Dihedral Coeffs`, `Improper Coeffs`
  - Class2-related sections: `BondBond Coeffs`, `BondAngle Coeffs`, `MiddleBondTorsion Coeffs`, `EndBondTorsion Coeffs`, `AngleTorsion Coeffs`, `AngleAngleTorsion Coeffs`, `BondBond13 Coeffs`, `AngleAngle Coeffs`

- When reading molecule templates, atom/bond/angle/dihedral/improper type labels can now be created (not only looked up) via `find_or_create(...)` instead of failing on unknown labels.

**Post Submission Checklist**


akohlmey:
> Continuing with our objective to turn molecule templates into standalone input files, we add capabilities to molecule templates for reading in force field coefficients.

@salamaaya @jrgissing 

I think this is a **very** bad idea. Force field coefficients are very specific and not only depend on the units used.  It will encourage people mix and match parameters that are not designed to be mixed.

This kind of abstraction is what is supposed to be handled by type labels where the labels provide a way to associate parameters with specific atom or similar interaction type. Embedding force field specific parameters with molecular geometries subverts the design choice for molecule *templates*.

I think you are moving into the wrong direction here and need to find a different approach.

akohlmey (changes requested):
From my perspective the more logical approach to handle automatic generation of type label and coefficient assignments would be to have a tool that can read and parse molecule files (and data files?) and extract the list of type labels, then query a database for a given force field (or use the required algorithm to generate the parameters, e.g. when an increment system is required) and then create some file(s) that can be used with the "include" keyword.

This is what I understand as the approach that a tool like moltemplate is using and that puts the main problem (matching parameters with types and topology) outside of LAMMPS.

I would also suggest to have a look at the "Simulator models" of the OpenKIM project.

jrgissing:
I disagree. Types labels are enabling this feature, but themselves do not provide enough provenance about a given molecule template. Another way to say it is, we need molecule templates to carry around information about the atom types that are already in the molecule template. In the context of database entries involving molecule templates, which will have other metadata associated with them that were tested against the force field parameters, this is meant to prevent misuse of those entries.

akohlmey:
> I disagree. Types labels are enabling this feature, but themselves do not provide enough provenance about a given molecule template. Another way to say it is, we need molecule templates to carry around information about the atom types that are already in the molecule template. In the context of database entries involving molecule templates, which will have other metadata associated with them that were tested against the force field parameters, this is meant to prevent misuse of those entries.

Sorry, this is not convincing me. Rather than scattering force field information into even more places and having more parsing of the same things and more points of failure, we should work in the opposite direction of having a central database with the force field data. But most of that can be done in pre-processing the data and does not belong into the LAMMPS code itself.

jrgissing:
Our immediate application is a database of reaction templates. These templates must be fully described, from a physics point of view, to be useful. I agree that hopefully the template parameters were derived from a centralized, verified interatomic potential database. But, some parameters may be custom, and even within a given force field, there is no one set of parameters. We are trying to enable a library of tested and user-ready templates for use in reactive MD simulations.

akohlmey:
> We are trying to enable a library of tested and user-ready templates for use in reactive MD simulations.

But you need to find a different approach. Sorry, I am quite adamant about it. Embedding this into molecule files is bad design in my opinion. I already mentioned OpenKiM where they put the logic for all of that outside of LAMMPS and used existing commands in a transparent fashion. This not ideal either, but much cleaner and keeps the cruft for specific applications out of the core LAMMPS code. I am not suggesting that you should
replicate their interface, I only mention that this is a way to add support for an
advanced abstraction and automation outside of the core LAMMPS code.

jrgissing:
Yes, I am aware of how OpenKIM deals with a similar issue. But, it is a different problem. We cannot automate the parameter generation, it is a part of the metadata associated with a given reaction. More generally, libraries of small molecules should be able to encode their physics for specific MD problems in quantitative way. We went this direction because it seems like the most elegant solution.

akohlmey:
> We went this direction because it seems like the most elegant solution.

Obviously, I don't consider it elegant at all. All my mental alarm bells are ringing that this is opening a door that should better remain closed. Already with the inferred type labels for bonded interactions you have pushed things to the boundary.

akohlmey:
> be able to encode their physics for specific MD problems

The type label or the molecule ID or the molecule file name are the facilities to encode any additional information. Since these are arbitrary strings you can make them as complex as you want them and encode multiple properties, e.g. a reaction identifier or whatever else you want, into the string.

akohlmey:
@jrgissing here is another possible approach: Since you can call ``Molecule::from_json()`` to create a molecule template, you could delegate the extra processing to fix bond/react and just read some more complex JSON files where the molecule templates are some objects and the force field parameters some other objects and you can then do all the processing from there.

jrgissing:
> @jrgissing here is another possible approach: Since you can call `Molecule::from_json()` to create a molecule template, you could delegate the extra processing to fix bond/react and just read some more complex JSON files where the molecule templates are some objects and the force field parameters some other objects and you can then do all the processing from there.

We are going in this direction for some other aspects of fix bond/react. But, I do not see it solving the general problem of encoding physics into database entries of molecules and fragments (including application-specific databases of monomers and functional groups). I am still a little confused about your objection from a conceptual standpoint, since it is the same info as a type label, just made explicit/quantitative. I think it makes the most sense to have the option to store physics in the same file/object as the types.

akohlmey:
> I am still a little confused about your objection from a conceptual standpoint, since it is the same info as a type label, just made explicit/quantitative.

No it is not. The type label can stand for anything, e.g. you can have the same type label, but use different parameters or have them in different units. Water is a good example. You can use the same molecule template for many different water models. Even more if you overwrite the assigned charges. And if you use fix shake, you can even have a different bond length and angle, it will be corrected within the first couple of steps.

You are arguing as if a specific type label implies what the force field and the corresponding parameters are, but that is not the case.

The other issue is that you look at this from how things are useful for your application while I am looking at this from how many undesired things people can do with it, especially when they lack training in force fields. We already have lots of people use ReaxFF not because ReaxFF is the best force field (it can be quite bad), but because they think they don't have to go through the laborious and unfamiliar task of atom typing and learning how a specific force field works. I want to avoid creating more such opportunities with something that I am not convinced it cannot be differently.

Furthermore i look at this from the perspective of maintenance and code complexity. The code in the molecule.cpp file is already at the limit of what I can tolerate, especially since everything is duplicated for the two supported file formats. I am very worried about adding *any* additional functionality to that code that is not generally useful. So far, all you are arguing is convenience for your specific application.

Here is yet another idea: implement a new command molecule/react that does what you want. Since this is new functionality, you don't need to bother with backward compatibility and only support the JSON format and (as in a previous suggestion) put the coefficients in a separate object so you can re-use the current molecule file processing and then do whatever else you want to do in a command that is only part of the REACTION package and solely your responsibility to keep working and deal with users (ab-)using it.

jrgissing:
> No it is not. The type label can stand for anything, e.g. you can have the same type label, but use different parameters or have them in different units. Water is a good example. You can use the same molecule template for many different water models. Even more if you overwrite the assigned charges. And if you use fix shake, you can even have a different bond length and angle, it will be corrected within the first couple of steps.
> 
> You are arguing as if a specific type label implies what the force field and the corresponding parameters are, but that is not the case.

Right, I agree with this, and that is exactly the issue we are trying to address, but in the context of interfacing databases with LAMMPS. The type label is not enough, we need to to carry the actual parameters tested against other metadata.

> The other issue is that you look at this from how things are useful for your application while I am looking at this from how many undesired things people can do with it, especially when they lack training in force fields. We already have lots of people use ReaxFF not because ReaxFF is the best force field (it can be quite bad), but because they think they don't have to go through the laborious and unfamiliar task of atom typing and learning how a specific force field works. I want to avoid creating more such opportunities with something that I am not convinced it cannot be differently.

I hear your worry about misuse, but I do not see this feature being any more problematic than other features that make LAMMPS more flexible, of which there are many. In fact, it originally came up to prevent ambiguity within a molecule template.

> Furthermore i look at this from the perspective of maintenance and code complexity. The code in the molecule.cpp file is already at the limit of what I can tolerate, especially since everything is duplicated for the two supported file formats. I am very worried about adding _any_ additional functionality to that code that is not generally useful. So far, all you are arguing is convenience for your specific application.

I believe this will be a generically useful feature likely to be utilized by other database-related tools, not specific to my application, which is why we are proposing to put it into the molecule templates. Big picture, I am trying to make it easier for LAMMPS to interface with parameterized molecular databases. 

> Here is yet another idea: implement a new command molecule/react that does what you want. Since this is new functionality, you don't need to bother with backward compatibility and only support the JSON format and (as in a previous suggestion) put the coefficients in a separate object so you can re-use the current molecule file processing and then do whatever else you want to do in a command that is only part of the REACTION package and solely your responsibility to keep working and deal with users (ab-)using it.

My above thoughts address why I think it will be useful outside of fix bond/react.

akohlmey:
> but I do not see this feature being any more problematic than other features that make LAMMPS more flexible

This is worse than anything I can think of. The more I think about this, the more I am opposed to the change. What just came to me is that when you can change coeffs in a molecule file, then you may change coeffs for atoms *outside* the molecule template where you have set the parameters with corresponding coeff command or from within the data file. This is a *very* unexpected behavior and thus makes this an absolute no-go for me.  This whole concept and method would *only* make sense to me if the coeff values would be per-atom and not per-type.

In the many years of doing software maintenance, I have rarely come across something where all my instincts are warning me so much, and the closer I look the worse it gets. I understand that it is hard to deal with this kind of strong opposition after you have worked on it hard and have become fond of your idea. I have been there myself quite a few times. But my experience is that going back and redesigning what was rejected, has predominantly resulted in something better, sometimes much better.

On the other hand, it has been a tradition in LAMMPS to let people do what they like for as long is it is contained in "their" package, does not affect portability, does not conflict with documented required policies, and does not affect anything else.

I suggest we let this discussion rest for a week or so to let emotions cool down a bit.

jrgissing:
> What just came to me is that when you can change coeffs in a molecule file, then you may change coeffs for atoms _outside_ the molecule template where you have set the parameters with corresponding coeff command or from within the data file. 

Yes, this is one of the main things I wanted to take a look at after Aya's PR. Note there is the same issue when reading in multiple data files or subsequent coeff commands. Perhaps in all cases, a warning or error should be thrown if the new coeffs are different than the old.

akohlmey:
> Note there is the same issue when reading in multiple data files or subsequent coeff commands

Don't get me started on data files. I have a long list of grievances about them and would very much prefer that we had an alternative without those deficiencies.

The difference is that back in the day, it either happened before I got involved in LAMMPS or I had no say when features like adding data files to an existing system were added.

The fact that what I personally consider as bad choices were made in the past does not justify making them again.

salamaaya:
I’d be happy to explore moving the functionality into the REACTION package instead, possibly through a separate JSON-based workflow, if that’s something you’d be interested in @jrgissing

jrgissing:
@salamaaya thanks for the offer. Let's wait on any modifications until we finish up this discussion. Steve mentioned he would weigh in on this topic; he has been busy recently.

PR #4975, opened 2026-05-07
Fixing region bugs

jtclemm:
**Summary**

In dynamic unions/intersections of regions, there are several bugs due to the fact that relevant displacement/rotation variables were either never defined (*point*, *axis*) or are defaulted to zero (*dx*, *dy*, *dz*). This produces memory/physical errors when calling Region::inverse_transform() (through match() or surface()) and Region::velocity_contact(). 

This was patched by requiring all subregions have the same dynamic settings and copying data to RegionUnion. This restriction is noted in the documentation. This fixes the inverse_transform() issues but not velocity_contact() which is a bit more complicated. Thus, I just passed that call to the appropriate subregion.

**Alternatively** (!!!!!!!!!!!), the match() and surface() methods could also be passed to the appropriate subregion and region union/intersect don't need to worry about what the dynamic settings are. Can do this if preferred.

Secondly, the contact radius for the cylindrical region style was only defined for fully closed regions. Definitions were added to open regions. 

**Author(s)**

Joel Clemmer (SNL)
Gabriel Alkuino (Syracuse) 

**Licensing**

**Backward Compatibility**

Using different dynamic settings for subregions of union/intersect will now error. However, if anyone was doing this, chances are results were just bogus.

### DIFF
--- a/doc/src/region.rst
+++ b/doc/src/region.rst
@@ -258,6 +258,11 @@ consisting of the volume that is common to all the listed regions.
    from their list of sub-regions.  Thus you cannot delete the
    sub-regions after defining a *union* or *intersection* region.
 
+.. note::
+   The *union* and *intersect* regions require all subregions have
+   the same *move* and *rotation* settings to ensure all regions
+   move in unison
+
 ----------
 
 The *side* keyword determines whether the region is considered to be
--- a/doc/src/region.rst
+++ b/doc/src/region.rst
@@ -258,11 +258,6 @@ consisting of the volume that is common to all the listed regions.
    from their list of sub-regions.  Thus you cannot delete the
    sub-regions after defining a *union* or *intersection* region.
 
-.. note::
-   The *union* and *intersect* regions require all subregions have
-   the same *move* and *rotation* settings to ensure all regions
-   move in unison
-
 ----------
 
 The *side* keyword determines whether the region is considered to be
@@ -318,9 +313,9 @@ If the *move* or *rotate* keywords are used, the region is "dynamic",
 meaning its location or orientation changes with time.  These keywords
 cannot be used with a *union* or *intersect* style region.  Instead, the
 keywords should be used to make the individual sub-regions of the
-*union* or *intersect* region dynamic.  Normally, each sub-region should
-be "dynamic" in the same manner (e.g. rotate around the same point),
-though this is not a requirement.
+*union* or *intersect* region dynamic.  Each sub-region must be "dynamic"
+in the same manner (e.g. rotate around the same point) or else LAMMPS
+will return an error.
 
 The *move* keyword allows one or more :doc:`equal-style variables
 <variable>` to be used to specify the x,y,z displacement of the region,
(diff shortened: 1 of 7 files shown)


stanmoore1 (approved):
I approve, though Steve should review as well.

doc/src/region.rst
```diff
@@ -313,9 +313,9 @@ If the *move* or *rotate* keywords are used, the region is "dynamic",
 meaning its location or orientation changes with time.  These keywords
 cannot be used with a *union* or *intersect* style region.  Instead, the
 keywords should be used to make the individual sub-regions of the
-*union* or *intersect* region dynamic.  Normally, each sub-region should
-be "dynamic" in the same manner (e.g. rotate around the same point),
-though this is not a requirement.
+*union* or *intersect* region dynamic.  Each sub-region must be "dynamic"
+in the same manner (e.g. rotate around the same point) or else LAMMPS
```
sjplimp: Does "dynamic in the same manner" only apply for rotation points (and rates), or does it also apply to the "move" option?  E.g. can a dynamic region be the overlap (intersection) of 2 oscillating squares, with different params in a variable for the oscillation.  Not sure if that is useful, but if not allowed, please clarify here in the doc page.

sjplimp:
@jtclemm @gsalkuin @akohlmey Re: your "alternative" suggestion.  I don't have a strong opinion, but the alternative does sound cleaner, code-wise.  Are there use cases for allowing a union/intersection region to have components with different dynamics properties (or to be static)?  If so, I think you are saying the alternative approach would allow that, where the current PR would not.  @akohlmey may have opinions on this as well.

akohlmey:
@jtclemm @sjplimp What is certainly needed here is a clarification. I've made a minimal attempt here https://github.com/akohlmey/lammps/blob/c6165e88ffd81fd708cf707ccc35cdcaee04cbb4/doc/src/region.rst?plain=1#L118

This was prompted by a post in the MatSci forum: https://matsci.org/t/automatically-scale-region-when-cell-rescales/67185

Internally, the code uses the terms "variable shape" and "dynamic", but to a user that is confusing because at the user level *both* are 'dynamic' regions. So this needs to be properly clarified.

I think for move and rotate enforcing that all parts of a union/intersection move the same way is reasonable (even though I can imagine people wanting to do more complex things), but to a large degree, those could still be done with variables, so the restrictions are less than what they may appear.

doc/src/region.rst
jtclemm: It means all settings for the *move* and *rotate* options must be identical. If *move* was specified with an oscillating variable, that variable would need to be the same for both to create a union/intersect in the current code in this PR.

jtclemm:
> If so, I think you are saying the alternative approach would allow that, where the current PR would not.

That would then allow the oscillatory case you mentioned above. Let me look into what it'd take to swap approaches then. My recollection is that it was a bit more hairy but not terribly so.

> What is certainly needed here is a clarification. I've made a minimal attempt here

Those clarifications do seem helpful Axel. I didn't look too deeply in the variable shape code, but I should also double check that to make sure there are no obvious issues with unions/intersections too (but I don't think there would be).

jtclemm:
Swapped to alternate, required more changes in region.h/cpp to override default parent classes in union/intersection.

PR #4977, opened 2026-05-08
Expanding limits on number of bins

jtclemm:
**Summary**

If the size of the simulation box relative to a neighbor bin grows, LAMMPS will eventually fail when either:

1. Memory runs out from allocating the binhead array (length is # of bins on a processor)
2. The # of bins on a processor exceeds the smallint limit

Throwing more processors at the problem to reduce the size of a subdomain can work to an extent, but this is a terrible waste of resources and only works a little before you suffer parallel efficiency issues and/or run out of machines.

In traditional simulations, this probably isn't an issue unless you have a very low density gas. However, when you have extreme polydispersity (say > 1 to 100), then you can easily have a very small particle density (most volume is taken up by massive single particles) coupled with a very small minimum bin size. Practically, this creates a new upper limit on achievable size ratios despite the improvements to neighbor style *multi*. 

To fix this, I created a new neighbor_modify setting to toggle whether binned atoms are stored in a linked list (the default) or a hash table. This option is currently only compatible with neighbor style *multi* since I can't think of reasonable usecases that don't have polydispersity (*multi* could also be used for monodisperse systems if needed). When using the hash option, the maximum # of bins is adjusted to bigint. 

Much bigger boxes can now be run (see example below) and early tests also demonstrate a nice speed up in highly polydisperse systems. 

**Author(s)**

Joel Clemmer (SNL)

**Licensing**

**Backward Compatibility**

NA

Here's an example script that uses the new option to create a huge box with 1e15 bins in serial:

```
atom_modify     sort 1000 100
region          box block 0 100000 0 100000 0 100000
#region          box block 0 2000 0 2000 0 2000   # even this size fails without bin/hash
create_box      1 box
mass            1 1.0

neighbor        1 multi
neigh_modify    bin/hash yes

pair_style      harmonic/cut
pair_coeff      * * 1.0 1.0

run 0
```

The old method caps out at ~1e9 bins.


stanmoore1 (approved):

sjplimp:
@jtclemm I like this idea a lot - nice work.   Just 2 comments:

(1) I expected this to be 2 new styles: for Nbin and Npair, rather than added logic in the existing styles.  Is there a reason to prefer the added logic?

(2) The comment in the doc page about atom_modify sort seems important.  Since I presume the default atom sorting would fail in the same manner as atom binning for neighbors, if not changed. Could you add a pointer/link to an example input script which uses a suitable atom_modify sort command, like the one listed above?   

For a future PR, do you think we need to re-think how atom sorting is done for these kinds of problems?  Just using huge sorting bins will not be efficient for the pairwise interactions between small particles once atoms become randomly disordered -- that is the issue atom sorting is trying to address.  Also we should probably ask Stan @stanmoore1 to comment since atom sorting is done differently in Kokkos.

jtclemm:
> (1) I expected this to be 2 new styles: for Nbin and Npair, rather than added logic in the existing styles. Is there a reason to prefer the added logic?

There was no *good* reason for for this approach. However, the actual reason I didn't create new Nbin/Npair styles is simply that I did not want to create an entirely new template variable since that would involve editing many more files and more intrusive modifications.

If there was ever a reason to use the hash table with something other than *multi*, I think creating new Nbin/Npair classes would be better. But I think this code could be rewritten to create another set of templates without affecting the user. 

> (2) The comment in the doc page about atom_modify sort seems important. Since I presume the default atom sorting would fail in the same manner as atom binning for neighbors, if not changed. Could you add a pointer/link to an example input script which uses a suitable atom_modify sort command, like the one listed above?

Currently there isn't an example script that would need this setting since its only really a large-scale problem. I could concoct something arbitrary though if you think it would be good to add a new example (maybe two particles in a box with one particle being 1000x wider?).

> For a future PR, do you think we need to re-think how atom sorting is done for these kinds of problems? Just using huge sorting bins will not be efficient for the pairwise interactions between small particles once atoms become randomly disordered -- that is the issue atom sorting is trying to address. Also we should probably ask Stan @stanmoore1 to comment since atom sorting is done differently in Kokkos.

From the doc page, I understood sorting can be up to a 4x speed up. I guess I would wonder how much of that is lost by using a less efficient sort with large bins? 

Currently, the only use cases I know of are for granular packings (solid-like) since strain rates would probably need to be prohibitively slow to do shear. But if there was a reason to study liquid like flow, I think you're right that this would be something to revisit.

sjplimp:
Since there is no example script to point to, can you include a sentence in the doc page where it refers to atom_modify sort.
Which explains that you need to use larger bins for atom sorting, and maybe a rule-of-thumb for how large the bins need to be.  The atom_modify doc page is unclear on this, b/c it says some fraction of the neighbor cutoff, but what does that mean for neighbor multi ?

Once you clarify that, I will approve the PR.

Re: future PR -- I think the 4x will be lost if sorting bins are too large.  In the limit of one sorting bin (size of simulation box), it is lost by definition.  The small optimal bins means most of the atoms a central atom interacts with are nearby in the proc's dense per-atom vector, so it's a cache effect.  What I think would be needed for the poly-disperse case is some kind of hierarchy of bin sizes, so each bin has a smallish # of atoms (whether small or large atoms), and then re-order the atoms according to those bins.  So not a simple idea to implement.

sjplimp:
E.g. a kd-tree might work well for large poly-dispersity

sjplimp:
And actually one sorting bin per proc would lose the 4x advantage.

jtclemm:
> Since there is no example script to point to, can you include a sentence in the doc page where it refers to atom_modify sort.
> Which explains that you need to use larger bins for atom sorting, and maybe a rule-of-thumb for how large the bins need to be. The atom_modify doc page is unclear on this, b/c it says some fraction of the neighbor cutoff, but what does that mean for neighbor multi ?

There is a sentence on neigh_modify.rst pointing to the sort page, but let me emphasize in it that it's important to use a small size. Since we are only just hitting the need for this hash table, setting the atom sorting just 2x larger is sufficient for simulations (until computational resources grow that is). 

However, testing the examples in multi, I actually do not see a big difference between the default sorting (sort bin = max neigh cutoff), turning sorting off, or adjusting the sorting bin size (bin = min neigh cutoff). Testing with shear, it's no more than 5-10% difference in total runtime.

If a large particle is interacting with particles far away, does that negate the benefits of sorting?

sjplimp (approved):

PR #4982, opened 2026-05-13
Clean up input file parsing in DPD-REACT package

### LINKED ISSUE #348: USER-DPD fix_rx segfaults when input file is invalid
USER-DPD fix_rx segfaults when its input file (such as kinetics.dpdrx) is invalid (for example, if there is a stray line). Ideally it should error out gracefully and notify the user the file is invalid without segfaulting.

<pre>
==129793== Invalid read of size 1
==129793==    at 0x4088098: strlen (vg_replace_strmem.c:454)
==129793==    by 0x1017C00B: LAMMPS_NS::FixRX::post_constructor() (fix_rx.cpp:319)
==129793==    by 0x10449587: LAMMPS_NS::Modify::add_fix(int, char**, int) (modify.cpp:900)
==129793==    by 0x10153083: fix (input.cpp:1559)
==129793==    by 0x10153083: LAMMPS_NS::Input::execute_command() (input.cpp:797)
==129793==    by 0x1015373B: LAMMPS_NS::Input::file() (input.cpp:244)
==129793==    by 0x100233B3: main (main.cpp:47)
==129793==  Address 0x0 is not stack'd, malloc'd or (recently) free'd
<pre>

jjramsey:
**Summary**

This cleans up the input file parsing in the DPD-REACT, replacing the old C-style parsing involving raw use of `strtok()`, `fopen()`, etc., with the use of LAMMPS' file parsing classes and utilities.

This also streamlines the handling of the parameters file for the `PairExp6rxKokkos` class, so that all of the actual file parsing is handled by the base `PairExp6rx` class instead of having almost identical file parsing code in both the base and derived class. Documentation for DPD-REACT has been slightly revised (1) to clarify that spaces are mandatory around the `+` and `=` operators in the kinetics equations in the input file for the `FixRX` class and (2) to account for how comments are supposed to be made in the input files.

**Related Issue(s)**

Fixes #348.

**Author(s)**

James J. Ramsey, currently affiliated with the Physical Modeling and Simulation Branch of DEVCOM Army Research Laboratory. Current work email is james.j.ramsey14.civ@army.mil, and my personal email is jjramsey@pobox.com

**Licensing**

By submitting this pull request, I confirm that I did NOT use any AI tools to generate all or parts of the code and modifications in this pull request.

**Backward Compatibility**

Previously, the parsing of certain files happened to allow any trailing characters after intended inputs, allowing for those trailing characters to be used as comments. Since this was more of an accident rather than an intentional choice, comments are now all required to be prefixed with a "#" character.

**Implementation Notes**

There is a new class introduced, `RxTableFileReader`, which wraps the existing LAMMPS `TableFileReader` and provides some additional functionality in order to avoid repeating similar code across the classes in the DPD-REACT package. This class should be formatted according to the `.clang-format` file provided by the LAMMPS project. In pre-existing classes whose implementations start with the comment `// clang-format off`, formatting and indenting was largely left alone, so that it would be easier to identify what changes in this pull request are substantive, rather than mixing actual code changes with changes in whitespace.

Correctness was verified by checking that several test cases yielded the same outputs as from before the code changes.

LAMMPS functionality unrelated to DPD-REACT should be unchanged.

**Post Submission Checklist**

### DIFF
--- a/doc/src/fix_rx.rst
+++ b/doc/src/fix_rx.rst
@@ -111,13 +111,15 @@ this file is described below.
 
 There is no restriction on the total number or reaction equations that
 are specified.  The species names are arbitrary string names that are
-associated with the species concentrations.  Each species in a given
-reaction must be preceded by its stoichiometric coefficient.  The
-only delimiters that are recognized between the species are either a
-*+* or *=* character.  The *=* character corresponds to an
-irreversible reaction.  After specifying the reaction, the reaction
-rate constant is determined through the temperature dependent
-Arrhenius equation:
+associated with the species concentrations, though the string names
+should be limited to what would be valid values for the names of
+:doc:`property/atom fixes <fix_property_atom>`. Each species in a
+given reaction must be preceded by its stoichiometric coefficient.
+The only delimiters that are recognized between the species are either
+a *+* or *=* character. Spaces around these delimiters are
+**required**.  The *=* character corresponds to an irreversible
+reaction.  After specifying the reaction, the reaction rate constant
+is determined through the temperature dependent Arrhenius equation:
 
 .. math::
 
@@ -167,7 +169,7 @@ parenthesized comments):
 
    # Rxn equations and parameters                                               (one or more comment or blank lines)
 
-   1.0 hcn + 1.0 no2 = 1.0  no + 0.5 n2  + 0.5 h2 + 1.0 co   2.49E+01 0.0 1.34  (rxn equation, A, n, Ea)
+   1.0 hcn + 1.0 no2 = 1.0  no + 0.5 n2  + 0.5 h2 + 1.0 co   2.49E+01 0.0 1.34 # (rxn equation, A, n, Ea)
    1.0 hcn + 1.0  no = 1.0  co + 1.0 n2  + 0.5 h2            2.16E+00 0.0 1.52
    ...
    1.0  no + 1.0  co = 0.5  n2 + 1.0 co2                     1.66E+06 0.0 0.69
--- a/doc/src/pair_exp6_rx.rst
+++ b/doc/src/pair_exp6_rx.rst
@@ -102,7 +102,7 @@ parenthesized comments):
 
    # exponential-6 parameters for various species      (one or more comment or blank lines)
 
-   h2o  exp6  11.00 0.02 3.50                          (species, exp6, alpha, Rm, epsilon)
+   h2o  exp6  11.00 0.02 3.50                        # (species, exp6, alpha, Rm, epsilon)
    no2  exp6  13.60 0.01 3.70
    ...
    co2  exp6  13.00 0.03 3.20
--- a/examples/PACKAGES/dpd-react/dpdrx-shardlow/kinetics.dpdrx
+++ b/examples/PACKAGES/dpd-react/dpdrx-shardlow/kinetics.dpdrx
@@ -4,7 +4,7 @@
 #   Arr = 1/picosec or Angstroms^3/mol/picosec; Ern = eV
 
 # format of a single entry (one or more lines):
-#   rxn equation, Arn, Ern
+#   rxn equation, Arn, n, Ern
 
 1.0 rdx           = 3.0 hcn + 1.5 no2 + 1.5 no + 1.5 h2o  2.51000000000000E+04 0.0 1.912325921964536
 1.0 hcn + 1.0 no2 = 1.0  no + 0.5 n2  + 0.5 h2 + 1.0 co   2.49081028001241E+01 0.0 1.344265387322009
--- a/src/DPD-REACT/fix_eos_table.cpp
+++ b/src/DPD-REACT/fix_eos_table.cpp
@@ -204,11 +204,11 @@ void FixEOStable::read_table(Table *tb, Table *tb2, char *file, char *keyword)
   memory->create(tb2->efile,tb2->ninput,"eos:efile2");
 
   reader.read_in_table_data([&](RxTableFileReader::TableIndex_t i,
-				ValueTokenizer & values) {
-			      values.next_int(); // throw away the initial index
-			      tb->rfile[i] = tb2->efile[i] = values.next_double();
-			      tb->efile[i] = tb2->rfile[i] = values.next_double();
-			    });
+                                ValueTokenizer & values) {
+                              values.next_int(); // throw away the initial index
+                              tb->rfile[i] = tb2->efile[i] = values.next_double();
+                              tb->efile[i] = tb2->rfile[i] = values.next_double();
+                            });
 }
 
 /* ----------------------------------------------------------------------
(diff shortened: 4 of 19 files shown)


jjramsey:
I noticed that the vast bulk of the tests failed on MPI2 constructs, such as `MPI_Type_create_struct()`. I did see that `src/output.cpp` uses `MPI_Type_create_struct()` and related functions and types, but bracketed by `#if !defined(MPI_STUBS)` and `#endif`. Is that the canonical approach for dealing with MPI constructs not covered by the STUBS library?

akohlmey:
> I noticed that the vast bulk of the tests failed on MPI2 constructs, such as `MPI_Type_create_struct()`.

LAMMPS uses mostly MPI-1 with very few MPI-2 constructs. This is mostly due to the age of the code base and for increased portability.

> I did see that `src/output.cpp` uses `MPI_Type_create_struct()` and related functions and types, but bracketed by `#if !defined(MPI_STUBS)` and `#endif`.

That function was added very recently for a specific purpose of a package and is not representative of LAMMPS' MPI programming.

> Is that the canonical approach for dealing with MPI constructs not covered by the STUBS library?

See src/STUBS/mpi.h for what is covered. The approach in LAMMPS has been that it only contains wrappers for what is used by LAMMPS. If you try to use a facility not covered by STUBS, you either have to implement the corresponding wrappers in STUBS or avoid using it.

jjramsey:
> The approach in LAMMPS has been that it only contains wrappers for what is used by LAMMPS. If you try to use a facility not covered by STUBS, you either have to implement the corresponding wrappers in STUBS or avoid using it.

Which would you prefer? I can do it either way, and I can see the tradeoffs in each approach. Personally, I'd prefer updating STUBS, since doing without `MPI_Type_create_struct()` would be less clean, and I'm doubtful that the change to STUBS would meaningfully impact LAMMPS' portability, since it seems unlikely that a system would both have an old MPI implementation on the one hand but have compilers recent enough to support C++17 on the other. However, I can see why you might be leery of introducing something new outside of a LAMMPS package, where it could potentially affect LAMMPS more broadly.

If I were to go the route of updating the STUBS library, I'm not sure if that should be done in a separate pull request (since it can affect more than just DPD-REACT) or added to this one (since the changes to STUBS would directly affect DPD-REACT).

akohlmey:
> Which would you prefer?

This is something where not only my personal opinion matters, but that of a larger part of the LAMMPS core developers. @sjplimp @stanmoore1 @athomps what do *you* think?

Please note that the C++17 requirement was added rather recently and we are generally slow in adopting such a change since there are many HPC clusters around with old OS versions. Please also note that it is generally, much simpler to install a new compiler to an old machine (especially as a user without system administrator priveledge) than an MPI library.

The most restrictive supported environment for parallelization is currently Windows since MS-MPI is currently the only usable MPI library (both MPICH and OpenMPI are desperately looking for developers willing to support and package their recent releases for Windows) and based on a rather old version of MPICH-2 and the MSVC++ compilers only support the OpenMP 2.0 standard.

> If I were to go the route of updating the STUBS library, I'm not sure if that should be done in a separate pull request (since it can affect more than just DPD-REACT) or added to this one (since the changes to STUBS would directly affect DPD-REACT).

I don't understand this statement. If you are *adding* features to the STUBS library that are not used anywhere else, how can the rest of the code be affected?

jjramsey:
> I don't understand this statement. If you are adding features to the STUBS library that are not used anywhere else, how can the rest of the code be affected?

Because there's a non-zero possibility that bugs might be introduced into the STUBS library when adding features, even to parts not originally intended to be related to those features. Those could affect the rest of the code.

> Please also note that it is generally, much simpler to install a new compiler to an old machine (especially as a user without system administrator priveledge) than an MPI library.

Point taken.

Thanks also for letting me know about MS-MPI and its limitations. It does look like it supports `MPI_Type_create_struct()`, though that obviously doesn't speak to the case for old HPC systems where users added new compilers on their own.

stanmoore1:
@jjramsey can you please remove the added tabs in the modified files and replace them with spaces? You can see a list of source lines here: https://github.com/lammps/lammps/actions/runs/25817473020/job/75910544770?pr=4982.

STUBS compile tests are failing with:
```
/home/runner/work/lammps/lammps/src/DPD-REACT/pair_exp6_rx.cpp: In member function ‘void LAMMPS_NS::PairExp6rx::genParamMpiDatatype()’:
/home/runner/work/lammps/lammps/src/DPD-REACT/pair_exp6_rx.cpp:714:3: error: ‘MPI_Aint’ was not declared in this scope; did you mean ‘MPI_Fint’?
  714 |   MPI_Aint param_displacements[NUM_PARAMS];
      |   ^~~~~~~~
      |   MPI_Fint
```

jjramsey:
> @jjramsey can you please remove the added tabs in the modified files and replace them with spaces?

I already did that locally but didn't push it yet because that would automatically trigger CI tests that would just fail due to the issues around `MPI_Type_create_struct()`. I plan on pushing the whitespace fixes along with the changes to address the aforementioned issues.

I've already drafted two possible source code changes, one that amends the STUBS library to use `MPI_Type_create_struct()` and related functions like `MPI_Aint_diff()`, and one that just avoids those functions. Just let me know which one you want to be pushed along with the whitespace fixes. [ETA: yes, I know that making changes to the STUBS library may need input from other LAMMPS developers, per [this comment](https://github.com/lammps/lammps/pull/4982#issuecomment-4451663279) from @akohlmey.]

akohlmey:
> I know that making changes to the STUBS library may need input from other LAMMPS developers, per [this comment](https://github.com/lammps/lammps/pull/4982#issuecomment-4451663279) from @akohlmey

I wanted to state my own preference only after others have commented. But since nobody seemes sufficiently motivated, I state it now so that we can move forward on integrating the pull request. My personal preference is to keep changes to STUBS to the minimum and use only a small subset of MPI to keep the code easier to understand since most contributors are not experienced (MPI) programmers. This is in line with keeping the implementation of the base styles - especially styles computing forces - simple and have extensive optimizations or additional parallelization confined to add-on packages like OPT, INTEL, KOKKOS, or OPENMP.

stanmoore1 (approved):
This looks fine to me. @jjramsey did you check if any of the changes need to be ported to the KOKKOS version?

jjramsey:
> This looks fine to me. @jjramsey did you check if any of the changes need to be ported to the KOKKOS version?

Yes. See the summary of the pull request where I noted, "This also streamlines the handling of the parameters file for the PairExp6rxKokkos class, so that all of the actual file parsing is handled by the base PairExp6rx class instead of having almost identical file parsing code in both the base and derived class." There is a virtual function `grow_exp6_params_array()` in the `PairExp6rx` class that is overridden in the `PairExp6rxKokkos` class, which allows the de-duplication of file parsing code to work.

PR #4995, opened 2026-05-26
Update references to examples in the manual and add python tool to maintain them

### LINKED ISSUE #2680: [Feature Request] Adding references to example input scripts in the doc pages
**Summary**

It would be helpful and instructive if the doc page of each command (fixes, pair styles, commands) contains pointers (or references) to some, if not all, input scripts in the examples/ sub-folders that illustrate the use of the command.

**Detailed Description**

In many cases, users would search for a command (on the web or in the documentation) and try to find an example input script for it. With some pointers (or references) to the corresponding input scripts (i.e. under which sub-folder in examples) in the doc page, users can find what they need faster. Also, when some developers contribute new codes and some example scripts, they can point users to these new examples, too.

For example, in the doc page of fix colvars, please include a section: 

Example input scripts available: examples/USER/colvars

or in doc page of fix ave/correlate 

Example input scripts available: examples/VISCOSITY; examples/KAPPA

The current workaround is to use the grep command in the examples folder, but we can do better than this, right?

akohlmey:
**Summary**

This pull request adds or corrects references to the examples tree in the LAMMPS manual. It also adds a tool written in Python to update the references specifically in the doc pages for individual commands.

**Related Issue(s)**

This implements and thus fixes #2680

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

The Python tool to report and implement possible updates was created by Claude Opus 4.7

**Backward Compatibility**

N/A

**Implementation Notes**

N/A

**Post Submission Checklist**


jtclemm:
Nice to have a formal tool to do this. I'm still parsing what the python script is doing and how it chooses the # of examples and which to print. 

I know the discussion originally suggested putting the list of examples at the top, but what about instead putting it at the very bottom near defaults? My thinking is that, say, someone is on the doc page reading about what *compute bond* does and how to use it. But before getting to the description they see:

"Example input scripts available: examples/PACKAGES/cgdna/examples/lj_units/oxDNA3/duplex2/in.duplex2, examples/PACKAGES/cgdna/examples/real_units/oxDNA3/duplex2/in.duplex2, examples/PACKAGES/cgdna/examples/lj_units/oxDNA3/unique_bp/in.dsring2"

It seems a bit jarring to immediately see a series of links with cryptic acronyms pointing you to go look somewhere else before you even know what compute bond does. Or at best it only wastes a split second glancing past a long line of text. But repeat that for every single time you read a doc page...

Thoughts?

akohlmey:
> Nice to have a formal tool to do this. I'm still parsing what the python script is doing and how it chooses the # of examples and which to print.

I think it lists up to three explicitly and if there are more in a folder then the folder
 
> I know the discussion originally suggested putting the list of examples at the top, but what about instead putting it at the very bottom near defaults? My thinking is that, say, someone is on the doc page reading about what _compute bond_ does and how to use it. But before getting to the description they see:
> 
> "Example input scripts available: examples/PACKAGES/cgdna/examples/lj_units/oxDNA3/duplex2/in.duplex2, examples/PACKAGES/cgdna/examples/real_units/oxDNA3/duplex2/in.duplex2, examples/PACKAGES/cgdna/examples/lj_units/oxDNA3/unique_bp/in.dsring2"
> 
> It seems a bit jarring to immediately see a series of links with cryptic acronyms pointing you to go look somewhere else before you even know what compute bond does. Or at best it only wastes a split second glancing past a long line of text. But repeat that for every single time you read a doc page...

You picked a particularly extreme and more disruptive example with very long paths, but I personally feel that those are at the exact right place next to the single line command examples. Also, this looks less extreme when translated to HTML, too.

> Thoughts?

Three things:

1. we should probably try to simplify the directory tree for the cgdna package so that the path names are shorter and encode what is now encoded in a directory name just as extension to the input file
2. I don't think there is much impact. My reasoning for that is that in my observations the impatient people barely look beyond the Examples section and mostly read the Syntax summary (and miss a lot of the details that they then ask for help about). For those there is no difference and the more careful people scroll to the "Description" header to read the documentation properly. So there is also little impact. On top of that, you can usually jump to individual sections of a doc page on the navigation bar on the left.
3. the lower you have information on a doc page, the less likely people will look at it. Putting the links to examples at the bottom reduces the impact of linking them in the first place.

doc/utils/find_example_refs.py
```diff
@@ -0,0 +1,560 @@
...
+
+def cmd_report(repo_root, output, list_cap=5):
+    doc_dir = repo_root / 'doc' / 'src'
+    ex_dir = repo_root / 'examples'
+
+    mapping, pages = build_mapping(doc_dir)
+    refs = scan_examples(ex_dir, mapping)
+
+    out = []
+    out.append('# Phase 2 candidate references for issue #2680.')
```
jtclemm: May not need to note when merged?

akohlmey: > May not need to note when merged?

I was thinking about re-checking and revising the script after this pull request is merged *and* a few others so that it would be worth to run again. That could be delegated to one of the next "collection" pull requests. The number of changes in this PR are too many to want to debug and revise the script.

After all, this is not (yet) going to be run automatically. It would be simpler to tell copilot to request links to examples like it already asks for index and versionadded and similar.

jtclemm:
> You picked a particularly extreme and more disruptive example

I won't deny some cherry picking! But I also didn't scroll far before I saw it.

> 2\. I don't think there is much impact. My reasoning for that is that in my observations the impatient people barely look beyond the Examples section and mostly read the Syntax summary (and miss a lot of the details that they then ask for help about). For those there is no difference and the more careful people scroll to the "Description" header to read the documentation properly. So there is also little impact. On top of that, you can usually jump to individual sections of a doc page on the navigation bar on the left.

I still contend it's more important to read a description of what the command does than to just see a list of even more examples. You could also equivalently jump to the example section in the navigation bar (or press the *end* key if it's at the bottom!). But this is subjective so probably best to just see what others think.  

Certainly not a critical difference either way.

akohlmey:
> I still contend it's more important to read a description of what the command does

With that reasoning, the whole Examples section should be moved to the bottom.
My most important point is that I don't want to have this split into two parts.
So we should either keep it or move the entire Examples block.

doc/utils/find_example_refs.py
```diff
@@ -0,0 +1,560 @@
...
+    ('command', 'fix_modify'), ('command', 'compute_modify'),
+    ('command', 'min_modify'), ('command', 'min_style'),
+    # Style-declaration verbs themselves (not the style values)
+    ('command', 'pair_style'), ('command', 'bond_style'),
+    ('command', 'angle_style'), ('command', 'dihedral_style'),
+    ('command', 'improper_style'), ('command', 'kspace_style'),
+    ('command', 'atom_style'),
+    ('command', 'pair_coeff'), ('command', 'bond_coeff'),
+    ('command', 'angle_coeff'), ('command', 'dihedral_coeff'),
+    ('command', 'improper_coeff'),
```
jtclemm, line 163: Is this list curated? Or did the AI just generate some candidates?

Some other that might be considered: *newton*, *set*, *molecule*, *minimize*, *create_atoms/bonds*, *fix enforce2d*..

jtclemm:
> With that reasoning, the whole Examples section should be moved to the bottom.

Oh I see, I didn't follow that this was the concern. This is just my personal opinion of course, but I see a clear distinction between readily readable examples (which are important and should be at the top) vs. examples that need to be navigated to outside of the doc pages (less critical and can be at the bottom). So I would still argue to split them apart to prioritize information that is self-contained within the doc page itself. 

Although do acknowledge there is something to be said about keeping all examples together. I'm happy to agree with whatever preference others have.

doc/utils/find_example_refs.py
jtclemm: That's fair.

stanmoore1 (approved):
It looks OK to me, but the comments from Joel need to be resolved before merging.

sjplimp:
@akohlmey I would also like to review this one before merging

sjplimp (changes requested):
still want to review

sjplimp:
@akohlmey @jtclemm I can see both sides of this debate.  What if we keep the list of 3 instances at the top
of each doc page (in the Examples section), but make them brief links to the GitHub webpage for that script.  The one-liner would always be concise and the user could view each of them with a click:

Usage in example input scripts:  micelle/in.melt, duplex2/in.duplex2, ...

Where the first entry links to https://github.com/lammps/lammps/blob/develop/examples/melt/in.melt
Second to https://github.com/lammps/lammps/blob/develop/examples/PACKAGES/cgdna/examples/lj_units/oxDNA3/duplex2/in.duplex2   (NOTE: this link was broken, now fixed)

The user can see the full path to the example script at the top of the GH page.

Or we could add the examples dir (or just its scripts) to the LAMMPS website as text files if that were somehow better.

Bonus points if you can come up with a URL which highlights the correct command in the input script.  Looks like you might need a PDF version of the input script for that.

sjplimp:
Also, why is the list of new RST files incomplete, e.g. no fix nve.  Seems like this should only be the case if no input script uses that command ?

akohlmey:
> e.g. no fix nve.

I explicitly asked for skipping commands that are ubiquitous, have no options, and do not benefit from examples. Fix nve is a strong contender in that field. What can you learn from an example file using fix nve that you cannot learn from the example lines quoted in the documentation?

sjplimp:
>>> What can you learn from an example file using fix nve that you cannot learn from the example lines quoted in the documentation?

Where to put it, in an input script.

akohlmey:
> Seems like this should only be the case if no input script uses that command ?

This can be *very* misleading, if the purpose of the example is not related to what the command stands for.

sjplimp:
What I meant was, there should be no one-liner added to the command doc page (no new RST file), if we have no example input script which uses that command.

akohlmey:
> Where to put it, in an input script.

Sorry, I cannot comment on this in public any further without offending your sense of what should be said in public or not.

sjplimp:
All I'm asking is to not exclude any commands from the auto-generated one-liners.  I think there is value in seeing how the command is used properly in the context of an entire working input script.  Even for fix nve.

sjplimp:
Another idea: if there are more than 3 example scripts which use the command, display the links to the shortest 3 files (by line count).  To make them easier for the user to grok.

akohlmey:
> All I'm asking is to not exclude any commands from the auto-generated one-liners. I think there is value in seeing how the command is used properly in the context of an entire working input script. Even for fix nve.

Let's talk about this offline, *please*.

akohlmey:
@jtclemm @sjplimp @stanmoore1 I am now revisiting this topic and I don't want to go over all the detailed comments because many make assumptions about the added script and what its intended purpose is that don't apply. So let me first summarize:
- the primary purpose is to be run after `make html` or `make pdf` or as part of `make check` in the doc folder where it should report a) broken references where the file was renamed or the folder moved or both, b) list .rst files that have no suggested examples but where suitable files exist. This will help us to keep this info current
- there will be an --update flag to correct broken references when the information can be deduced from the git history
- inserting links will be on a per-file basis only, i.e. the edit is automatic, but a user has to invoke the command manually and review and commit the changes manually

As mentioned before, adding references to the repository is too fragile (it is complex to make work reliably with the online manual, but will break with local copies), so if we really want to have a link, we have to add script code to convert input files into .rst files that we can include into the html/PDF documentation as an Examples appendix.

This pull request will include one sweep across the examples tree to have a starting point. I will add some heuristics to have the script omit embedding long links. This applies for example the CG-DNA package. I have already asked the CG-DNA developers to flatten their rather deep examples tree.

I maintain that adding links for "ubiquitous" command through this script is not helpful. It should focus on suggesting topical examples, so the internal lists to skip such commons will remain. Instead, such examples should be added manually, e.g. examples/melt/in.melt for fix nve and a bunch of other "trivial" commands. I personally still think that it is not really helping, because people should have learned how to use such commands when they learned how to write LAMMPS inputs in the first place, e.g. the canonical LAMMPS tutorials, but I am willing to compromise on this, but these entries should be added sparingly and manually.

doc/src/bond_class2.rst
```diff
@@ -22,6 +22,8 @@ Examples
    bond_style class2
    bond_coeff 1 1.0 100.0 80.0 80.0
 
+Example input scripts available: examples/replicate/in.replicate.cnt, examples/amoeba/in.water_box.amoeba, examples/amoeba/in.water_box.hippo
```
jrgissing: Why do some examples list directories, while other point to input files? It seems reasonable to ask the user to find the relevant input files within the directory. Then, a greater variety of example folders could be listed. This simplification could probably be extended to excluding subfolders in many cases, but that might get complicated.

akohlmey: This is to avoid overflowing the documentation with long lists. The current heuristic is to list explicit examples for up to 3 examples. Beyond that a folder is offered.
There is no intention for this list to be complete same as the command lines above do not list all permutations of commands. This is meant as some additional information without making it too distracting.

I would rather put a paragraph with a shell command for finding examples somewhere (or a python script), e.g. `find ./examples -name in.\* -print | xargs grep -l mycommand`. We want users to become "better users" not us becoming their nannies. ;-)

jrgissing: I am not suggesting to remove the cap at three examples. Just to stick with folders. So, this one might become:

examples/amoeba, examples/replicate, examples/PACKAGES/fep/C7inEthanol/fep01

In my opinion, we could even stop at examples/PACKAGES/fep. More coverage, less hand-holding.

akohlmey: > I am not suggesting to remove the cap at three examples. Just to stick with folders.

That is too coarse. Many folders have different examples and only some match.
That does not mean that other choices cannot be applied on a case-by-case basis.

Let me repeat that this pull request does two things: 1. add a script that can help us maintaining a consistent list of references to examples, 2. an initial sweep across the documentation to add such links.

I don't want to go through the impossible task of making the script do all edits and updates for everything automatically. This is by its very nature a task that requires human judgement. Providing a reference to a folder may sometimes be the better choice and sometimes not. A script cannot tell. It should by default just flag potential candidates and detect references that are no longer valid.

akohlmey: FYI, for the FEP package we have now:
```
$ grep examples/PACKAGES/fep src/*.rst
src/compute_fep.rst:Example input scripts available: examples/PACKAGES/fep
src/compute_fep_ta.rst:Example input scripts available: examples/PACKAGES/fep/ta/in.spce.lmp
src/fix_adapt_fep.rst:Example input scripts available: examples/PACKAGES/fep
src/Packages_details.rst:* ``examples/PACKAGES/fep``
src/pair_fep_soft.rst:Example input scripts available: examples/PACKAGES/fep
src/pair_fep_soft.rst:   provided in the LAMMPS source directory tree, under examples/PACKAGES/fep.
```
i.e. provide the folder except for one case where a specific command is only used in one specific example. This is what I am aiming for.

sjplimp:
@akohlmey Thanks for re-opening the discussion.  These are my 3 suggestions; all are about how the command doc pages appear for users.

(1) Syntax 

Looks like we have roughly 750 doc/src/*.rst files with an Examples section at the top, which are doc files for input script commands.  For each, I think a newline plus a single line should be added at the bottom of that section with a syntax like this:

Input scripts in examples dir: melt/in.melt, shear/in.shear.void, PACKAGES/pedone/in.pedone.melt

Details:
not examples/melt/in.melt, it's redundant
always a file, not a dir
up to 3 files if they exist, use "none" if no files exist

if doc page is for multiple commands, one line for each, like this:
    Input scripts for fix rigid in examples dir: rigid/in.rigid, ...
    Input scripts for fix rigid/small in examples dir: rigid/in.rigid.small, ...

ordering of the files:
    lower-case example dirs first, then upper case (1st priority)
    shorter files (by line count) before longer files (2nd)

(2) Scope

I think this should be done for all 750 files.  I don't see a reason to exclude this info from "common" commands.

It's easy to do for all command files.  It's just augmented info we now provide in the Examples section of each page.  Saying "people should have learned" might be true for nearly all folks reading the doc page of a common command, but why exclude the few it might help.

(3) Links

I still think it would be useful to have each file be a link to display the file in the browser.  I understand it may not be workable
to do from the manual with a link to the repo.  Though this is what we do for scripts linked to on the webpages? Like
https://www.lammps.org/examples/mixer_drum

If not workable from the repo, how about making a copy of the examples dir on www.lammps.org, and linking to those static files?  Updating the webserver examples dir infrequently should be fine.

akohlmey:
First off, you seem to be misunderstanding the scope and purpose of both the script and the pull request. I have already explained it to Jake, but perhaps I need to be more explicit:

- The purpose of the script is to check **existing** references for correctness and to provide **suggestions** for adding references.  I.e. to add new references you run the script and capture the output. Then you edit that file as needed, drop examples, reorder, add, etc. and then run the script again with --apply to insert the references in the file. Furthermore, it is run with --check as part of `make check`, `make html`, and `make pdf`. It is a deliberate choice to skip over cases with common commands as the choice of which are good examples should be made by a human. I cannot think of any heuristics that can result in meaningful suggestions for all cases, so I am not even attempting it.
- The pull request also includes a first run with suggestions for example links and a few corrections. It is not *meant* to be complete. I've added a few cases manually, too, but there are more to be done. We will need further pull requests by individuals adding examples where they are meaningful.

> (1) Syntax
 
> Input scripts in examples dir: melt/in.melt, shear/in.shear.void, PACKAGES/pedone/in.pedone.melt

"dir" is a "slang" term. Something like "tree" would be more appropriate.
 
> Details: not examples/melt/in.melt, it's redundant always a file, not a dir up to 3 files if they exist, use "none" if no files exist

I disagree with the "none". That has a high risk to become outdated and incorrect.
It is much harder to write a tool that confirms this.

> if doc page is for multiple commands, one line for each, like this: Input scripts for fix rigid in examples dir: rigid/in.rigid, ... Input scripts for fix rigid/small in examples dir: rigid/in.rigid.small, ...

This is another deliberate omission. I don't want to invest a lot more time in the script to make it smarter for a few cases. This is better handled manually. Also, in the case of fix rigid, I think this level of detail is not warranted. We point to the folder with many examples. People can search for the inputs that match.

> ordering of the files: lower-case example dirs first, then upper case (1st priority) shorter files (by line count) before longer files (2nd)

I agree with the first heuristic, but not the second. The length of input files does not have any correlation with relevance for a specific command.

> (2) Scope
> 
> I think this should be done for all 750 files. I don't see a reason to exclude this info from "common" commands.

The exclusion in the script is deliberate because there are no heuristics that can detect meaningful suggestions across hundreds of files.  As explained at the top, this must be done manually and that is outside the scope of this pull request. It is simply not worth the effort to add much more complexity to the script and I currently don't have the time or interest to go over all files manually.

The choice is simple. Either we accept this as a partial solution, or I'll trim down the pull request to only contain the check for valid references and skip the entire addition of example links.

> It's easy to do for all command files. It's just augmented info we now provide in the Examples section of each page. Saying "people should have learned" might be true for nearly all folks reading the doc page of a common command, but why exclude the few it might help.

Then I challenge you to provide a better working solution.

> (3) Links
> 
> I still think it would be useful to have each file be a link to display the file in the browser. I understand it may not be workable to do from the manual with a link to the repo. Though this is what we do for scripts linked to on the webpages? Like https://www.lammps.org/examples/mixer_drum

Yes, but 1) this is for a very small number of examples, 2) those pages can choose to point to either of the three versions of the online docs and don't have to be compatible with the offline manual version or a PDF, and 3) we run an automatic link checker that creates lists of broken or redirected links on a webpage accessible from the the internal part of the LAMMPS homepage. There still is the problem of the scripts being out of sync with the exact copy used to create the images and movies shown, whether the images and movies should be updated with the changes in the scripts.
 
> If not workable from the repo, how about making a copy of the examples dir on [www.lammps.org](http://www.lammps.org), and linking to those static files? Updating the webserver examples dir infrequently should be fine.

I think this is a bad idea. We just got rid of a ton of horribly outdated copies of example files when converting the LAMMPS homepage. I am not interested to make the same mistake again. There should not be multiple copies in multiple repositories (website source and LAMMPS source) that need to be kept in sync. Note that a simple link check is not sufficient since the contents may be changed due to syntax changes.

akohlmey:
@sjplimp the topic has come up before, but perhaps this needs to be discussed again at a future developer meeting: another option would be to unbundle the examples tree from the LAMMPS sources and have a separate lammps-examples repository and tarball.

However, that has its own drawback, since that will interfere with the new regression test scripts by @ndtrung81 and will require people submitting new features to submit two pull requests, one for each repository, that need to be reviewed and tested in sync.

sjplimp:
@akohlmey you are correct that I did not understand how the Python script was intended to work.

But more fundamentally, I don't understand:

a) why does this need to be done incrementally (preserving or checking existing doc page refs to example scripts),
    and why does any of it need to be done manually?
    it seems like it could be done from scratch every time (e.g. once per feature release) so it's correct by construction
    we're talking roughly 1000 commands (in 750 doc page RST files) and roughly 1000 in.* input scripts in examples
    so that's a double loop with a million checks for whether a script contains a command
    is that check way more expensive than I'm imagining ?
    seems like each script could be one-time pre-processed (line by line) into a dictionary of commands it contains, 
       then the check is trivial

b) what do you mean by "meaningful suggestions" of scripts to use as refs for a specific command?
    I would have thought any script which uses the command is adequate for this purpose

Probably better to talk about this by phone.  I'm curious if you (or others) think my earlier (1) and (2) are good ideas for how the ref-containing doc pages should look - assuming it could be easily implemented?

akohlmey:
> a) why does this need to be done incrementally (preserving or checking existing doc page refs to example scripts), and why does any of it need to be done manually? it seems like it could be done from scratch every time

Because I don't see how this can be done without significant human intervention.
Even the bulk additions here will likely need refinements and improvements. It is only meant to have a first step at something that has been acknowledged as useful but also dormant for a long time. My intention was to get us at least part of the way.

**I** am not going to attempt to write such a script.

We have had a similar situation during the transition from txt2html to rst where initially all documentation was auto-converted. There were plenty of issues with that and we still have some of those left in the current documentation, especially in some of the very oldest files.

> (e.g. once per feature release) so it's correct by construction we're talking roughly 1000 > commands (in 750 doc page RST files) and roughly 1000 in.* input scripts in examples so that's a double loop with a million checks for whether a script contains a command is that check way more expensive than I'm imagining ? seems like each script could be one-time pre-processed (line by line) into a dictionary of commands it contains, then the check is trivial

Please feel free to write such a script.

> b) what do you mean by "meaningful suggestions" of scripts to use as refs for a specific command? I would have thought any script which uses the command is adequate for this purpose

I disagree on this one.  For "simple" or "trivial" commands you want to point people to simple example inputs not any random file.

> Probably better to talk about this by phone. I'm curious if you (or others) think my earlier (1) and (2) are good ideas for how the ref-containing doc pages should look - assuming it could be easily implemented?

Sorry, but I don't want this to be dragged out even more over a long time with many requests for changes, since we are obviously not converging.  I have had too much of that recently and am concerned I am unable to handle any more of this gracefully and patiently.  I will now proceed to trim this pull request down to just the added script and the checks during the build of the manual and a few minor corrections.  Anything else has to be separate pull requests which may be based on the current script or a better one written by somebody else.

akohlmey:
A reduced version of this pull request that only contains the reference checking and resulting corrections to the manual was submitted as PR #5180.

I am withdrawing this one.

PR #4997, opened 2026-05-27
Add fix symmetry command

### LINKED ISSUE #3373: [Feature Request]  Symmetric Molecular Dynamics option
**Summary**

Sam Cox and Andrew D. White in their recent work [1] developed a molecular dynamics that keeps the system symmetric throughout whole simulation (for all space groups). 

**Detailed Description**

The authors developed a simple md package [2]. As written in the description:
"""
This is a simple MD engine written in C that can be constrained to obey given symmetry groups at every frame. It can be compiled for any dimensions. It supports the following features:
    1. NVE, NVT (BAOAB, CSVR thermostats), NPT
    2. LJ, harmonic, cosine potentials
    3. Neighbor lists (use nlj instead of lj for potential)
    4. Symmetry constraints and all Bravais Lattices
 """
It would be useful to utilize constrained space group option not only with these provided functions but with different LAMMPS features (for instance, OPLS force field or PLUMED). 
It could be implemented as a new Fix, which can be requested by a keyword "fix symmetry [point group]"

[1] https://pubs.acs.org/doi/10.1021/acs.jctc.2c00401
[2] https://github.com/whitead/symd

akohlmey:
**Summary**

This implements a fix symmetry command based on the reference implementation at https://github.com/whitead/symd

**Related Issue(s)**

This implements and thus closes #3373

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

The bulk of the implementation was created using Claude Code Opus 4.7. Documentation revisions on use cases, performance, and illustrative figures were created by Claude
Code Fable 5 under close supervision and with final, manual revisions.

**Backward Compatibility**

N/A

**Implementation Notes**

There are descriptions and helper scripts in the examples folder. They might be adapted into a Howto document, preferably by someone that has used it for some real world applications.

**Post Submission Checklist**

Reference publication: https://doi.org/10.1021/acs.jctc.2c00401
Reference implementation: https://github.com/whitead/symd


stanmoore1 (approved):

sjplimp:
@akohlmey I would still like to review this before it is merged

sjplimp (changes requested):
still want to review

sjplimp:
@akohlmey This looks like an interesting new command - here is what I think the PR lacks, on the doc page, to motivate and guide users.

(1) What is the purpose of this new fix, why and when should users think about using it, e.g. for what kinds of materials questions?
(2) Is this intended for tiny systems, e.g. a few dozen atoms?  Or does it make sense for larger, or much larger systems?
(3) Does this work in both serial and parallel?  It seems like in parallel, info would need to be communicated for atoms which affect constraints on other atoms, e.g. the set of atoms in an "orbit".  If large systems can be modeled, that might mean long-range communciation.  If so, does all this work scalably in parallel ?

akohlmey:
> @akohlmey This looks like an interesting new command - here is what I think the PR lacks, on the doc page, to motivate and guide users.

@sjplimp I've updated the documentation and for further illustration created some example figures to demonstrate how the symmetry operations are applied. I also noticed that a warning about the conflict with RIGID package fixes is in place and a warning was added to the code.

doc/src/fix_symmetry.rst
```diff
@@ -0,0 +1,361 @@
...
+representative* and one image atom per non-identity operator of the
+declared space group. Per step the fix:
+
+* averages the forces and velocities of the atoms in each orbit (mapped
+  back to the asymmetric unit through the operator inverses) and
+  redistributes the symmetrized values to every image, and
+* re-folds the image positions from the asymmetric representative using
+  the affine operators to cancel numerical drift.
+
+The user is responsible for setting up the data file so that the full
```
sjplimp, line 54: Where is the "full unit cell" defined?  Is this in the JSON file ? It doesn't say that in the Syntax section above or in the JSON file section below.

doc/src/fix_symmetry.rst
```diff
@@ -0,0 +1,361 @@
...
+
+.. versionadded:: TBD
+
+This fix constrains a molecular dynamics simulation to obey a chosen
+crystallographic space-group symmetry at every timestep. The algorithm
+follows :ref:`(Cox) <Cox2022>`, and the LAMMPS implementation is derived
+from the `symd <https://github.com/whitead/symd>`_ reference
+implementation.
+
+Atoms are organized into *orbits*. Each orbit has one *asymmetric
```
sjplimp, line 44: This sentence isn't clear to me.  I think you are saying (later) that the number of orbits is small and a function of the JSON file settings.  Which would mean for a large system (many unit cells?) there are many atoms assigned to each orbit?  But the sentence doesn't seem to say that.

sjplimp:
@akohlmey The new sections of the doc page which discuss system size and parallel operation address by earlier comments - thanks.  But I still the think the doc page description is opaque to non-experts (like me) who are not familiar with some of the nomenclature.  Maybe that is fine -- anyone who wants to use this will know the nomenclature.

But I think you should get someone else who understands this better to read the doc page and approve this PR.
Possibly that is Aidan, but not sure.

akohlmey:
> Maybe that is fine -- anyone who wants to use this will know the nomenclature.

@sjplimp I think that this is a given, since using this fix requires creating a suitable JSON file with the applied symmetry data, where space group and applied symmetry operations are required inputs. I don't think that anybody can/should create that without first becoming familiar with space groups and similar crystallographic terms.

sjplimp:
ok - but my suggestion to have a more expert reviewer stands.  I don't anyone on the current list, unless Aidan, fits the bill.

akohlmey:
> ok - but my suggestion to have a more expert reviewer stands. I don't anyone on the current list, unless Aidan, fits the bill.

@sjplimp On advice of @athomps I have posted a message to the feature request issue mentioning the two people asking for this feature and the developer of the reference code. Since there has been no reaction to my previous comment, I don't expect much from that, but it is worth waiting a few days.

Beyond that we are at an impasse since it seems that my experiences from learning to use Quantum Espresso are the closest to what you are looking for :-(.

athomps (approved):
I approve

PR #4999, opened 2026-05-28
Support extended particles in inertia and angular momentum computes

### LINKED ISSUE #3710: Make inertia computations compatible with extended particles like ellipsoids
See
https://matsci.org/t/unreported-bugs-in-23-june-2022-stable-release/46310/2 and
https://matsci.org/t/unreported-bugs-in-23-june-2022-stable-release/46310/4

Compute inertia/chunk, compute inertia and inertia variable functions need to support extended particles. Corresponding code can be "borrowed" from fix rigid.

akohlmey:
**Summary**

Make the moment-of-inertia and angular-momentum diagnostics account for the
finite size of extended particles instead of treating every atom as a point
mass.  Affected: `compute inertia/chunk`, `compute angmom/chunk`,
and the `inertia()` and `angmom()` variable functions.  A new standalone
`compute inertia` command for the group inertia tensor is also added.

Each finite-size particle now adds its own moment of inertia about its center
(via the parallel-axis theorem) and its intrinsic (spin) angular momentum,
covering finite-size spheres, ellipsoids, superellipsoids, line segments,
triangles, and body particles.  This mirrors the treatment in `fix rigid`
(reusing `MathExtra::inertia_*`); body and superellipsoid particles, which
`fix rigid` does not support, use their stored per-particle inertia/angmom.

**Related Issue(s)**

closes #3710

**Author(s)**

**Licensing**

The code, documentation, and unit tests in this pull request were drafted with
the help of an AI tool (Claude Code, model Claude Opus 4.7) operating under my
direction, and were reviewed, tested, and edited by me before submission.  The
AI assistance covered the implementation in `src/`, the `.rst` documentation
updates, and the added unit tests; the design decisions, analytic test values,
and final review are human.

**Backward Compatibility**

For systems of point particles the results are unchanged.  For systems
containing finite-size particles the values reported by the affected computes
and variable functions change, because they now include the finite-size
contributions.  This is intentional and is the subject of issue #3710.  The
angular-momentum removal performed by `fix momentum`, `velocity zero angular`,
and `compute temp/rotate` is deliberately left unchanged: they continue to use
the point-mass (orbital) `Group::inertia`/`Group::angmom`, since they remove
bulk rotation from the translational velocities only.

**Implementation Notes**

The finite-size contributions are isolated in two new additive `Group` methods,
`inertia_extended()` and `angmom_extended()` (each with a region variant), that
the diagnostic outputs add on top of the orbital quantities returned by
`Group::inertia`/`Group::angmom`.  Keeping the existing point-mass methods
untouched preserves the behavior of the physics routines listed above.  The
per-chunk computes inline the same per-particle contribution.  Detection checks
the shape-defining indices (`ellipsoid`/`line`/`tri`/`body`) before the sphere
(radius) case, because finite-shape particles also carry a bounding radius.

Correctness was verified with new unit tests that compare against analytic
values for each particle type across the group, variable, standalone-compute,
and per-chunk code paths (e.g. a rotated body/superellipsoid whose stored inertia
round-trips). The existing point-particle chunk-compute regression test is unaffected.

**Post Submission Checklist**


doc/src/compute_angmom_chunk.rst
```diff
@@ -27,36 +27,47 @@ Description
...
 
 This compute calculates the 3 components of the angular momentum
 vector for each chunk, due to the velocity/momentum of the individual
 atoms in the chunk around the center-of-mass of the chunk.  The
 calculation includes all effects due to atoms passing through periodic
 boundaries.
 
+.. versionchanged:: TBD
+
+For finite-size particles the intrinsic (spin) angular momentum of each
```
jtclemm: How about ```:doc:`finite-size particles <Howto_spherical>` ```?

doc/src/compute_inertia.rst
```diff
@@ -0,0 +1,91 @@
...
+   how they are set for each atom.  You can reset the image flags (e.g., to
+   0) before invoking this compute by using the :doc:`set image <set>`
+   command.
+
+Output info
+"""""""""""
+
+This compute calculates a global vector of length 6, which can be accessed
+by indices 1--6.  The six components of the symmetric inertia tensor are
+ordered :math:`I_{xx}, I_{yy}, I_{zz}, I_{xy}, I_{yz}, I_{xz}`.  These values
```
jtclemm, line 69: I see this is consistent with inertia/chunk, so not something to change, but I'm just curious, any idea why it's yz then xz? 

In comparison to pressure computes, I would have expected xz yz.

akohlmey: This is a question for @athomps (or @sjplimp?).

doc/src/compute_angmom_chunk.rst
akohlmey: > How about `` :doc:`finite-size particles <Howto_spherical>` ``?

done in ab8c289721a9ae9f3c037e6e52e0f7c418f1c44b

doc/src/compute_inertia.rst
jtclemm: After ingraining the pressure computes in my head, I could definitely see myself using these computes and just assuming an incorrect order. Not sure there's a good way to address it though.

jtclemm (approved):

sjplimp (changes requested):
would like to still review

sjplimp:
For @akohlmey @jtclemm 
I think I see what you've done, but I am not sure it is correct for angmom and omega.

I think it is correct for moments of inertia, i.e. compute inertia and compute inertia/chunk.
Because the moment of inertia is different for a collection of non-overlapping spheriods versus a collection of point particles.

However, consider the omega of the same collection.  I don't think the omega of the collection changes if the individual spheroids happen to be randomly rotating (each with their individual omega).  Is that conceptually correct ?

I'm less sure about angmom, but I think angmom of the collection is also unchanged if individual particles are rotating
and thus each have their own angmom.

Another way to look at it, is to compare what fix rigid/small would give for omega/angmom of a rotating rigid body.  For fix rigid/small, the omega and angmom of each finite-size particle is set to match the rotation rate of the entire body.  But I think if your compute angmom/omega chunk calculated it for the same body, you would be adding an extra contribution from each component particle's rotation.  Doesn't that give a different answer for omega of the chunk?  And probably also for angmom ?

Possibly for angmom, what you need to do is subtract off the angmom for each particle due to the rotation of the collection.
But I don't think that is needed for omega.

sjplimp:
Update: after looking more at fix rigid/small, I think what I said about angmom of the chunk not depending on rotation of the individual particles is wrong.  So what this PR is doing for angmom is probably OK.  But I still think what I said about omega
is correct.  Which would mean calculating omega from angmom is incorrect.  That works for rigid bodies, because the omega of each particle is reset to match the rotation of the body.  But I think for the use case of these computes and group functions, omega should not be derived from angmom.  It should just be calculated as if each particle is a point particle (probably how it was before).

akohlmey:
@sjplimp I have removed the changes to the computation of omega.

sjplimp (approved):

PR #5000, opened 2026-05-28
Make ID nomenclature consistent in the manual

### LINKED ISSUE #4598: [Feature Request] Make use of atom-ID, molecule-ID, group-ID and others consistent

akohlmey:
**Summary**

The manual mixed several conventions for identifiers (`atom ID`, `atom-ID`, `atomID`), and the bare `ID` placeholder in command synopses (`fix ID group-ID nve`) was ambiguous (one column or two?).

This PR standardizes on the dashed form throughout `doc/src` and qualifies the synopsis placeholder by its command:

- `atom-ID`, `molecule-ID`, `group-ID`, `region-ID`, `chunk-ID`, `grid-ID`, `fix-ID`, `compute-ID`, `dump-ID`
- synopsis: `fix fix-ID group-ID style`, `compute compute-ID group-ID style`, etc.
- the per-style bullets become e.g. `* fix-ID, group-ID are documented in :doc:\`fix <fix>\` command`

To keep the convention from drifting back, it adds a coding-standard checker:

- `tools/coding_standard/idnames.py` (also auto-fixes with `-f`); it is code-block aware so the synopsis rule only applies inside `code-block`/`parsed-literal`, and it preserves capitalization/plurals while leaving code tokens like `c_ID`/`f_ID[I]` untouched
- `make check-idnames` / `make fix-idnames` targets in `src/Makefile`; `check-idnames` is now part of `make check` (so it runs in the existing style-check CI)
- the convention is documented in `Modify_requirements.rst`

Most of the ~464 changed `.rst` files were updated mechanically by the fixer; line-wrapped occurrences, a few syntax placeholders (`computeID`/`fixID` in `dump_image`, `groupID` in `read_data`, `regionID` in `create_box`/`fix_wall_gran_region`), and a pre-existing synopsis typo in `compute_ti.rst` (`group` -> `group-ID`) were corrected by hand.

**Related Issue(s)**

closes #4598 

**Author(s)**

**Licensing**

The python code and documentation updates in this pull request were drafted with
the help of an AI tool (Claude Code, model Claude Opus 4.7) operating under my
direction, and were reviewed, tested, and edited by me before submission.

**Backward Compatibility**

N/A

**Post Submission Checklist**


jrgissing:
This PR seems to be going in the right direction, but note that many (most?) of these added hyphens are grammatically incorrect, and therefore can become grating once that is noticed. For my 'review', I scrolled to one random point and noticed another issue that results in a newly ambiguous sentence, due to both incorrect grammar and now-inconsistent usage:
fix_pour:
'...determine the atom or molecule-IDs to assign to inserted particles/molecules.  In both...'
I supposed this should be 'the atom-IDs or molecule-IDs' using the new proposed convention.

sjplimp:
@akohlmey @jrgissing I agree aiming for consistency is good, but I agree with Jake that this is likely to make some awkward changes.  Maybe we should go with atomID, moleculeID, etc, and just change the hyphenated versions, which would hopefully limit changes to definitions and descriptions of arguments.  And which would avoid changing "atom IDs" or similar.  Note the doc pages use both mol-ID and molecule-ID which is another inconsistency.  Also, this simple change could be done with grep/sed ?

akohlmey:
> Also, this simple change could be done with grep/sed ?

Yes, I think so. Most of it anyway. I have avoided doing a manual check, but I have prepared to do so, once we have settled on conventions. It is a one time check and we can update the .github/copilot-instructions.md file so that (many) future inconsistencies in contributions will be flagged on its review (same as for the versionadded/versionchanged markers). I have no strong feeling about whether we go with or without hyphen. What I am aiming for are two things: a) replace a plain ID with a qualified groupID, fixID, atomID etc. and b) using a consistent set of terms throughout. If needed, we can even add it somewhere in the manual. In one of the pending pull requests, I have added added a document for writing docs from scratch to follow the documents discussing the implementation of
pair styles, fixes and more. So that would be a good place for such a list.

akohlmey:
Leaving this for somebody else to sort out.

PR #411, opened 2017-03-11
Removed triclinic restrictions on slab ewald/pppm.

frobnitzem:
Since the slab correction applies specifically to the z-direction, it is not influenced by tilt.  I have confirmed the correction by calculating it manually for a system with 2 charges (+1 at 0,0,1.5 and -1 at 0.0,-1.5).  The files are attached:

[ewtest.zip](https://github.com/lammps/lammps/files/835435/ewtest.zip)

Four cases are tested -- an orthogonal and a triclinic box with and without the slab correction.
Although changing the box tilt alters the energies and forces, the correction term is identical for both, as it should be.

PR #5002, opened 2026-05-28
Support kspace slab correction (EW3DC) for triclinic boxes

### LINKED ISSUE #410: [Feature Request] implement kspace slab correction for triclinic boxes
According to the docs, http://lammps.sandia.gov/doc/kspace_modify.html
>  The slab keyword is not currently supported by Ewald or PPPM when using a triclinic simulation cell.

... which is supported by the checks at ewald.cpp:109 "Cannot (yet) use Ewald with triclinic box and slab correction"
and pppm.cpp:192 "Cannot (yet) use PPPM with triclinic box and slab correction"

From my reading of Yeh and Berkowitz, subtracting the z-dipole should not be fundamentally more difficult in non-orthogonal box shapes, and is indeed implemented straightforwardly already on pppm.cpp:2944.  The only issue I noticed when trying it out was that the coordinates, x[i][2], refer to scaled versions and need to be transformed back into Cartesian when the dipole is summed on line 2954.

Rather than do that, the simple formula in Ballenger, JCP 140, 161102 makes it really tempting to implement corrections for arbitrary box shapes by accepting a tensor, 'J' and computing the whole dipole.

akohlmey:
**Summary**

Implements the Yeh-Berkowitz EW3DC slab correction (`kspace_modify slab <volfactor>`)
for triclinic (non-orthogonal) simulation cells in the `ewald`, `ewald/disp`, `pppm`,
and `pppm/cg` kspace styles. Previously this combination was rejected with
"Cannot (yet) use ... with triclinic box and slab correction".

The root cause was that the slab z-extension (`slab_volfactor`) was missing from the
triclinic *reciprocal-space* sum, not just from `slabcorr()`. The triclinic reciprocal
lattice vectors are built via `KSpace::x2lamdaT()` straight from `domain->h_inv`, which
uses the real box z dimension rather than the slab-extended one. Because the slab normal
must be the Cartesian z axis (so `xz == yz == 0`; an `xy` tilt is allowed), the
extended-box transform `c' = (xz, yz, zz*slab_volfactor)` reduces to scaling the z
component of the transformed reciprocal vector by `1/slab_volfactor`.
Applying this in `x2lamdaT()` centralizes the fix for both Ewald and PPPM and is a strict
no-op (`slab_volfactor == 1.0`) for all non-slab calculations.

A triclinic slab box with a non-zero `xz` or `yz` tilt is *not* supported: the
reciprocal-space sum would then stack images along the tilted `c` vector while the
Yeh-Berkowitz `slabcorr()` term still assumes perpendicular z-stacking, which is
inconsistent. Such boxes were not blocked by `domain.cpp` (a non-periodic z dimension
places no constraint on `xz`/`yz`), so the kspace styles now reject them with a clear
error rather than computing an incorrect result.

For PPPM the lamda grid z spacing (`delzinv`) is additionally extended to match
`Grid3d::set_zfactor()` used in the grid decomposition, `slabcorr()` is reordered to run
after the lamda->box coordinate conversion so it operates on Cartesian z, and the
Green's-function alias-sum bound in z accounts for the extended cell.

**Related Issue(s)**

Fixes #410

**Author(s)**

**Licensing**

This pull request was developed with the assistance of AI tools. The code changes,
documentation updates, and regression tests were generated by Claude Opus 4.8 (Anthropic,
via Claude Code) under human direction, and were reviewed and validated by the author.
The AI tool diagnosed the root cause, implemented the modifications in `src/kspace.cpp`,
`src/KSPACE/`, the accelerator-variant guards, the documentation, and the unit tests, and
ran the verification described below. The author reviewed all of the resulting changes.

**Backward Compatibility**

No backward compatibility is broken. The change only enables a previously forbidden
combination (slab correction + triclinic box). Orthogonal-box results are bit-for-bit
identical: the new reciprocal-space scaling is guarded by `slabflag == 1` and reduces to a
multiplication/division by `1.0` for all non-slab runs, and the PPPM `slabcorr()` reorder
is a no-op for orthogonal boxes (the lamda->box conversion it now follows is only invoked
for triclinic cells).

**Implementation Notes**

- `src/kspace.cpp`: `x2lamdaT()` scales the transformed reciprocal vector's z component by
  `1/slab_volfactor` when `slabflag == 1` (the core fix; shared by Ewald and PPPM).
- `src/KSPACE/ewald.cpp`, `src/KSPACE/pppm.cpp`: the existing triclinic+slab guards are
  narrowed so that only `slab nozforce` and `slab ew2d` remain blocked for triclinic cells
  (these crude/2D variants are not part of this change); the EW3DC `volfactor` path is
  enabled. The `compute_group_group()` guards are narrowed the same way.
- `src/KSPACE/ewald.cpp`, `src/KSPACE/ewald_disp.cpp`, `src/KSPACE/pppm.cpp`: a new guard
  in `init()` rejects a triclinic `slab volfactor` box with a non-zero `xz` or `yz` tilt
  (covering `pppm/cg` and the OpenMP variants by inheritance); an `xy` tilt is accepted.
- `src/KSPACE/pppm.cpp`: in `setup_triclinic()` the lamda z grid spacing is extended
  (`delzinv = nz_pppm/slab_volfactor`, with `delvolinv` keeping the full `nz_pppm`);
  `slabcorr()` is reordered after `domain->lamda2x()` so the dipole sum uses Cartesian z;
  and the alias-sum bound in `compute_gf_ik_triclinic()` uses the extended cell length.
- `src/KSPACE/pppm_cg.cpp`: `slabcorr()` reordered after the coordinate conversion (cg
  inherits the rest of the PPPM triclinic machinery).
- `src/KSPACE/pppm_tip4p.cpp`, `src/GPU/pppm_gpu.cpp`, `src/INTEL/pppm_intel.cpp`: keep
  triclinic+slab blocked with explicit guards (out of scope). `pppm/kk` already self-guards.
  `esp` and `pppm/stagger` have pre-existing triclinic limitations unrelated to the slab
  correction and remain unsupported.

Verification:
- Ewald and `ewald/disp` reproduce the orthogonal slab result to machine precision
  (~2e-15) for a zero-tilt triclinic cell (exact methods).
- PPPM and `pppm/cg` match the *exact* Ewald slab result to normal PPPM accuracy
  (rms ~4e-4, same as the orthogonal PPPM-vs-Ewald baseline) for both zero tilt and a
  finite xy tilt; `pppm/cg` is bit-identical to PPPM with matching charges.
- Identical energies and forces on 1, 2, and 4 MPI ranks (parallel grid decomposition with
  the slab zfactor).
- All existing kspace force-style regression tests still pass; orthogonal results unchanged.
- The new guard rejects a triclinic slab box with a non-zero `xz`/`yz` tilt at runtime for
  `ewald`, `ewald/disp`, and `pppm` (verified), while an `xy` tilt runs normally.

**Post Submission Checklist**

New regression tests: `unittest/force-styles/tests/kspace-{ewald,pppm,pppm_cg,ewald_disp}_tri_slab.yaml`.


akohlmey:
### Follow-up: MSM slab-correction fix + triclinic-slab test coverage

Three commits were added on top of the EW3DC triclinic work:

**1. Bug fix — MSM must ignore the slab correction (`msm.cpp`).**
Extending EW3DC to triclinic boxes put a `slab_volfactor` scaling into the shared
`KSpace::x2lamdaT()` (gated on `slabflag == 1`). MSM also supports triclinic boxes and
calls `x2lamdaT()` on its triclinic paths, but it only *warned* "slab not needed" without
disabling the correction — so `MSM + triclinic + kspace_modify slab N` silently divided the
MSM z-force by `N` (e.g. `fz` 2.5236 → 1.4879 for `N=3`). Fix: in `MSM::init()` reset
`slabflag`/`slab_volfactor` **on all ranks** (a rank-0-only reset would diverge the parallel
transform). Also covers `msm/dielectric`. With the fix, an MSM run with `slab` is bit-for-bit
identical to one without, at every MPI rank count.

**2. New test — `compute group/group` in a triclinic slab box**
(`unittest/commands/test_kspace_group_group_slab.cpp`). The force-style YAML tests never
invoke `compute group/group`, whose own slab path (`slabcorr_groups`) was separately enabled
for triclinic cells. Using the two-charged-sheet reproducer from #411, it checks for `ewald`
and `pppm` that (a) the orthogonal result matches the analytic infinite-sheet values
(`Fz = 2*pi*q^2/(Lx*Ly) = pi`, `E = 5*pi`) and (b) the triclinic xy-tilted box reproduces the
orthogonal result (Ewald to ~1e-13, PPPM within mesh tolerance).

**3. New test — MSM ignores slab in a triclinic box**
(added to `test_kspace_auto_slab.cpp`): a triclinic 100-atom MSM system, asserting the forces
are identical with and without `kspace_modify slab` (locks in fix #1).

**Checks performed**
- Reproduced the #411 verification tests: orthogonal energy `6.5933251`; zero-tilt triclinic
  now matches it (the 2017 value was `2.618`); analytic `|Fz| = pi` confirmed for orthogonal,
  zero-tilt, and genuinely sheared (xy) triclinic cells; `xz`/`yz` tilt correctly rejected.
- All four `kspace-*_tri_slab.yaml` references regenerate **byte-identical** (forces current).
- Full `KSpaceStyle` suite (51 tests) + new command tests pass; bit-identical on 1/2/4 MPI ranks;
  `make check-whitespace` clean.

Please double-check before approving. This was done by Claude Opus 4.8 under close supervision
of @akohlmey.

doc/src/kspace_modify.rst
```diff
@@ -456,11 +456,26 @@ in lieu of the volfactor.  This turns off all kspace forces in the z
...
+(non-orthogonal) simulation cells for :doc:`kspace_style <kspace_style>`
+*ewald*, *ewald/disp*, *pppm*, and *pppm/cg* (and the OpenMP variants
+that reuse the same slab-correction setup).  Because the slab normal must
+be the Cartesian z axis, only an *xy* tilt is allowed: a triclinic slab
+box with a non-zero *xz* or *yz* tilt is rejected with an error.  The z
+box length must be held fixed for the duration of the run.  Triclinic cells
+are not yet supported with the *slab* keyword for the *pppm/tip4p*,
+*pppm/stagger*, *esp*, *pppm/gpu*, *pppm/intel*, or *pppm/kk* styles, nor
+for the *slab nozforce* and *slab ew2d* options.  The slab correction
+has also been extended to point dipole interactions :ref:`(Klapp)
```
stanmoore1: This paragraph is confusing. It is hard to tell if the dipole styles support the triclinic slab or just the general slab.

doc/src/kspace_modify.rst
```diff
@@ -456,11 +456,26 @@ in lieu of the volfactor.  This turns off all kspace forces in the z
...
+approximation is not needed).  The *slab* keyword with a *volfactor*
+value (the Yeh-Berkowitz EW3DC correction) supports triclinic
+(non-orthogonal) simulation cells for :doc:`kspace_style <kspace_style>`
+*ewald*, *ewald/disp*, *pppm*, and *pppm/cg* (and the OpenMP variants
+that reuse the same slab-correction setup).  Because the slab normal must
+be the Cartesian z axis, only an *xy* tilt is allowed: a triclinic slab
+box with a non-zero *xz* or *yz* tilt is rejected with an error.  The z
+box length must be held fixed for the duration of the run.  Triclinic cells
+are not yet supported with the *slab* keyword for the *pppm/tip4p*,
+*pppm/stagger*, *esp*, *pppm/gpu*, *pppm/intel*, or *pppm/kk* styles, nor
```
stanmoore1: Is there a technical reason why the triclinic slab correction wasn't ported to pppm/kk and other styles which already support the general slab correction, or did the AI just get tired :)

stanmoore1: This really is just removing an error check in the kspace styles, so it should be trivial to implement for any style that already supports slab correction, right?

Edit: I guess it needs a small tweak to the number of lattice vectors/grid points too.

stanmoore1 (approved):

Issue #2933, opened 2021-09-09
[BUG] pair lubricate/poly+brownian/poly leads to strange attractions and a temperature increase


JitingTian:
**Description of the bug** 

When combining pair lubricate/poly and pair brownian/poly to simulate the dynamics of a binary hard-sphere (HS) mixture, I find that there are strange attractions between close particle pairs and that the system's temperature cannot be kept at the target one. 
**I already realize where is wrong and write new codes to fix the bug, see details below.**

 **LAMMPS Version and Platform** 

LAMMPS Version: 29Otc2020, the most recent stable version.
Platform: Ubuntu 20.10, 8 processors.

**Expected and Actual Behavior** 

(All scripts and results are provided in the attached package)
Using Fast Lubricate Dynamics (FLD) to simulate the equilibrium dynamics of a binary hard-sphere mixture, it's expected that the system will go normal suspension Brownian dynamics at a target temperature. For a 50:50 binary HS mixture with identical diameters equilibrated at T=1, the mono (pair lubricate + pair brownian) and poly (pair lubricate/poly and pair brownian/poly) version of FLD should give the same results of a NVE simulation from this configuration. 

However, as shown below, poly-FLD will result in a temperature increase in the system.
![thermo_different](https://user-images.githubusercontent.com/33658861/132669390-ee327b46-b8fe-4772-9a4a-49620bccfe0e.png)

Observing the atomic trajectories, I find that in the poly-FLD simulation close particles tend to bound together and increase their energies mysteriously, as shown below.
![particles](https://user-images.githubusercontent.com/33658861/132670951-0a9154bd-9154-439a-8082-f43094051f86.png)
I notice this bug is already reported several years ago (https://matsci.org/t/mysterious-attraction-between-hard-spheres-with-pair-lubricate/23325/20), but it still exists in the most recent stable version of LAMMPS.

**Fixing this bug**

I already fix this bug by writing new codes.

I analyze the source codes of poly-FLD, i.e. pair_lubricate_poly.cpp and pair_brownian_poly.cpp, finding that they do not use newton on to ensure a pairwise lubricate random force. This will result in two different lubricate random forces for a single pair, violating the fluctuation-dissipation theorem, and this is why the temperature cannot be kept. On the other hand, in the source codes of mono-FLD, i.e. pair_lubricate.cpp and pair_brownian.cpp, newton on is used. This difference is shown below.
[...]

Issue #1933, opened 2020-03-13
[BUG] _pair_lubricate_poly.cpp_issues


arneshpalanisamy:
**Summary**

pair_lubricate_poly.cpp has bugs related to force calculation using a_sq, a_sh, etc. as Ranga from Edinburgh university pointed out. 
The forces are not pairwise symmetric (i.e.lubrication force between a pair of particle i and particle j must be equal and opposite, here it is not so).

secondly the 
if(shearing) branch has a bug in line 127 Ef[1][2] = Ef[2][1] = 0.5 * h_rate[3]/domain->zprd;
It is supposed to be 
Ef[1][2] = Ef[2][1] = 0.5 * h_rate[3]/domain->xprd;

**LAMMPS Version and Platform**

22Aug2018 stable release, Ubuntu 19.0 

**Expected Behavior**

The terms need to be corrected to maintain physics.

**Actual Behavior**

Incorrect force on particles.

**Steps to Reproduce**

open the file and see it. run a lammps script with (lubrication force for polydisperse samples)poly_luburicate. 

**Further Information, Files, and Links**

akohlmey:
Thanks for reporting. BTW: the same issue also exists in pair style lubricate. This will be fixed in the next patch release.

sjplimp:
Hi - not clear on this.  I think the Ef = h_rate equations are just converting
from Voigt notation back to a full matrix form?
So Ef(0,1) is the xy term = hrate[5]
E(0,2) is the xz term = hrate[4]
E(1,2) is the yz term = hrate[3]

So isn't the divisor (xprd,yprd,zprd) set by
the 2nd letter.  I.e. yprd for xy, zprd for both xz and yz.
But you are changing one of the zprd to xprd?

Steve

akohlmey:
This was closed by accident. The patch has been reverted and the issue is not yet resolved.

akohlmey:
@arneshpalanisamy following up on the remarks by @sjplimp I am also adding the unconfirmed label, as your description is not sufficient to actually explain the details. As @sjplimp observed, the current (unchanged) implementation is consistent with the description in the comments above the statement you would like to change, so we need a more specific argument and demonstration why there would be a bug. Your proposed change looks reasonable unless one looks closer...

akohlmey:
@arneshpalanisamy do you have any updates/answers on this issue?

arneshpalanisamy:
Dear Akohlmey, 

Thank you for the updates, 
[...]

PR #5004, opened 2026-05-29
Collected Small Changes and Fixes

### LINKED ISSUE #4325: [BUG] fix ttm/mod doesn't replicate the results of fix ttm
**Summary**

`fix ttm/mod` does not produce the same result as `fix ttm` in the cases it should do so. I used the example `in.ttm` file and modified it to use `fix ttm/mod` while keeping all the parameters the same. The result is very different from what the `fix ttm` produces. It seems as if `fix ttm/mod` transfers energy from the electronic subsystem to the atomic subsystem extremely slowly.

LAMMPS Version: 17 Apr 2024 - Development - patch_17Apr2024-141-g008a8fcb11
Operating System: Linux 6.10.9

**Expected Behavior**

`fix ttm/mod` should reproduce the same results as `fix ttm` when the ttm parameters are the same.

**Actual Behavior**

`fix ttm/mod` and `fix ttm` produces different results. The energy transferred between the electronic and atomic subsystems are vastly different.

**Steps to Reproduce**

Modify the `in.ttm` file in the examples folder to use `fix ttm/mod` with the same parameters.

Two files are included in the uploaded zip. `in.ttm` is the modified example file, and `Fe.ttm_mod` is the file with the ttm parameters.

[ttm.zip](https://github.com/user-attachments/files/16973030/ttm.zip)

### LINKED ISSUE #5003: [BUG] compute msd/chunk wrongly produces a warning message
**Summary**

The command `compute msd/chunk` produces the warning message: One or more chunks do not contain all atoms in molecule, while the same chunk does not produce any warning with `compute com/chunk`.

LAMMPS (12 Jun 2025). Also verified on the current development version.
OS: Debian, kernel 7.0.10-1-liquorix-amd64

**Expected Behavior**

The same input script used with two different samples should not produce any warning.

| sample01 | sample02 |
| -------------------- | -------------------- |
| Image | Image |

Note that the order of molecules in sample01 is:
```
120 * (water, formamide) + 405 * (water)
```
while in sample02 is:
```
405 * (water) + 384 * (formamide)
```

**Actual Behavior**

When the `compute msd/chunk` is used with sample01, a WARNING message is printed. The same script applied on sample02 does not produce any warning. Also, if the command  `compute msd/chunk` is replaced with  `compute com/chunk`, no warning is produced.

**Steps to Reproduce**

Please use the attached files for debug: 
[compute_msd_chunk.zip](https://github.com/user-attachments/files/28370688/compute_msd_chunk.zip)
```bash
lmp -in sample01.in -l sample01.log # WARNING issued
lmp -in sample02.in -l sample02.log  # No warning
```

See discussion on https://matsci.org/t/one-or-more-chunks-do-not-contain-all-atoms-in-molecule-reprise/67198

### LINKED ISSUE #5026: failing  unittest/force-styles/tests/dihedral-table_spline.yaml
**Summary**

Building `LAMMPS-22Jul2025_update4-foss-2025b-kokkos` or the slightly older `LAMMPS-22Jul2025_update3-foss-2025b-kokkos` fails in the unit tests `dihedral-table_spline.yaml`

`LAMMPS-22Jul2025_update4` and `LAMMPS-22Jul2025_update3` on Debian and RHEL-8

All details can be found at the open PR # [26147](https://github.com/easybuilders/easybuild-easyconfigs/pull/26147) and PR #[25379](https://github.com/easybuilders/easybuild-easyconfigs/pull/25379)

**Details**

When building the above versions of `LAMMPS` that particular unit test is failing with the following error:

[test-failure.txt](https://github.com/user-attachments/files/28705486/test-failure.txt)

For me it looks like the error margins are a wee bit to tight, see [here](https://docs.lammps.org/Build_development.html#unit-tests-for-force-styles)

So the patch in the PR #26147 mentioned above is loosening the error margins a bit to get the test job over the finishing line. However, as I am not the expert in `LAMMPS` I am wondering if there is maybe a deeper lying problems I do not get. 
The build fails on both `Debian-bookwork (12)` and also on `RHEL-Opta (8.10)`, so I am somewhat confident it is not a problem with the OS, or the underlying CPU architecture (`11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz` for Debian, AMD Rome/Icelake for the RHEL boxes). 

I would appreciate a feedback if my approach is valid. I hope the supplied PRs give enough information about the rest of the build.

akohlmey:
**Summary**

This pull request combines multiple small changes and bug fixes that do not warrant a pull request of their own.

**Related Issue(s)**

Fixes #5026 
Fixes #5003 
Closes #4325 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

The bugfix for issue #5003, the bugfix for issue #1933 (pair style `lubricate/poly`), the bugfix for issue #2933 (pair style `brownian/poly`), and the restart/neigh_modify documentation update were created with the help of Claude Opus 4.8

**Backward Compatibility**

Yes

**Implementation Notes**

- handle the unavailability of the coverity scan download page more gracefully and don't report an error, just don't run the analysis
- fix a spurious "one or more chunks do not contain all atoms in molecule" warning from `compute chunk/atom molecule` (issue #5003): the molecule check now only warns for molecules that are actually split between a chunk and the excluded atoms, instead of for molecules that are entirely excluded by the compute group or region but happen to have an ID within the range of valid chunk IDs
- clarify docs for fix ttm/mod with respect to lsurface/rsurface and update examples
- add support for MPI_BOR operation to STUBS
- add a guide for writing and updating documentation pages
- update git tag for new release of ML-RUNNER external library
- document that writing a restart file forces a neighbor list rebuild (`restart` and `neigh_modify` docs), which can change the neighbor list build count and thus the trajectory
- add temporary patch to work around compilation failure of ML-RUNNER with git signature logging enabled

**Post Submission Checklist**


src/compute_chunk_atom.cpp
```diff
@@ -1319,9 +1323,28 @@ void ComputeChunkAtom::check_molecules()
...
+    std::vector<int> molflag(nchunk, 0);
+    std::vector<int> molflagall(nchunk, 0);
+
     for (int i = 0; i < nlocal; i++) {
-      if (molecule[i] > 0 && molecule[i] <= nchunk && ichunk[i] == 0) flag = 1;
+      if (molecule[i] <= 0 || molecule[i] > nchunk) continue;
+      molflag[molecule[i] - 1] |= (ichunk[i] == 0) ? 2 : 1;
     }
+
+    MPI_Allreduce(molflag.data(), molflagall.data(), nchunk, MPI_INT, MPI_BOR, world);
```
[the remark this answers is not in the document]
akohlmey: The <vector> include is documented as being one of multiple includes that can be assumed to be included when "pointers.h" is included, so adding it is not needed.

doc/src/Developer_write_docs.rst
```diff
@@ -0,0 +1,504 @@
...
+verified by ``make style_check`` (see :ref:`below <doc_checks>`), which
+parses the ``...Style(name,class)`` macros from the C++ headers and
+reports any style or accelerated variant whose index entry is missing.
+
+Title and command heading
+"""""""""""""""""""""""""
+
+The page title is the name of the command followed by the word
+"command", underlined with a row of ``=`` characters that must be at
+least as long as the title text.
```
jtclemm, line 68: Not exactly as long?

doc/src/Developer_write_docs.rst
```diff
@@ -0,0 +1,504 @@
...
+
+   pair_style born/gauss command
+   =============================
+
+When a page documents multiple styles, it is customary to list the
+primary style in the title (or to use a more generic title), and to make
+sure each documented style has an index entry and is reachable through
+the command tables.
+
+In reStructuredText a heading is indicated by underling it with
```
jtclemm: Underlining I presume, but I like the idea of documentation underlings.

doc/src/Developer_write_docs.rst
```diff
@@ -0,0 +1,504 @@
...
+
+* mention the directory inside the "Description" or "Examples" section,
+  for instance a line like ``Example input scripts available:
+  examples/PACKAGES/foo``;
+* reference specific input files by their path in the relevant
+  discussion.
+
+There is additional, semi-automated tooling for keeping the lists of
+example directories in the manual and the top-level ``README``
+synchronized (proposed in pull request #4995).  Once that is merged, the
```
jtclemm: is there any value in citing a github pull request in the doc page? 

Other requested features don't mention the issue #.

jtclemm: Or is this just a temporary placeholder that'll be removed down the road?

doc/src/Developer_write_docs.rst
akohlmey: Yes, it may be longer but must not be shorter. In emacs I just hit Ctrl-= and it will automatically resize and for all further Ctrl-= promote or demote the level.

src/COLLOID/pair_brownian_poly.cpp
```diff
@@ -175,135 +222,130 @@ void PairBrownianPoly::compute(int eflag, int vflag)
...
-          a_sh *= 6.0*MY_PI*mu*radi;
-          a_pu = beta0*(4.0+beta0)/10.0/beta1/beta1*log(1.0/h_sep);
-          a_pu += (32.0-33.0*beta0+83.0*beta0*beta0+43.0 *
-                   cube(beta0))/250.0/cube(beta1)*h_sep*log(1.0/h_sep);
-          a_pu *= 8.0*MY_PI*mu*cube(radi);
-
-        } else a_sq = 6.0*MY_PI*mu*radi*(beta0*beta0/beta1/beta1/h_sep);
-
-        // generate the Pairwise Brownian Force: a_sq
+            beta0*(1.0+7.0*beta0+beta0*beta0)/5.0/pow(beta1,3.0)*log(1.0/xi);
```
jtclemm: Where does this starting factor of beta0 come from? I don't see it in the old expression.

doc/src/Developer_write_docs.rst
akohlmey: Yes. Damn, I can't even blame AI for this... :disappointed:

src/COLLOID/pair_brownian_poly.cpp
```diff
@@ -175,135 +222,130 @@ void PairBrownianPoly::compute(int eflag, int vflag)
...
+          a_sq *= 6.0*MY_PI*mu*rref;
+          a_sh = 4.0*beta0*(2.0+beta0+2.0*beta0*beta0)/15.0/pow(beta1,3.0) *
+            log(1.0/xi);
+          a_sh += 4.0*(16.0-45.0*beta0+58.0*beta0*beta0-45.0*pow(beta0,3.0) +
+                       16.0*pow(beta0,4.0))/375.0/pow(beta1,4.0) *
+            h_sep*log(1.0/xi);
+          a_sh *= 6.0*MY_PI*mu*rref;
+          a_pu = 2.0*beta0/5.0/beta1*log(1.0/xi);
+          a_pu += 2.0*(8.0+6.0*beta0+33.0*beta0*beta0)/125.0/beta1/beta1*
+                   h_sep*log(1.0/xi);
```
jtclemm: I may be seeing things, but `a_pu` also looks different. For instance, where does the factor proportional to 43 beta0^3 go?

doc/src/Developer_write_docs.rst
akohlmey: Now *that* I can blame on AI. The whole paragraph is useless and inaccurate.

src/COLLOID/pair_brownian_poly.cpp
akohlmey: `beta0 = radj/radi;` computed in line 181

src/COLLOID/pair_brownian_poly.cpp
```diff
@@ -175,135 +222,130 @@ void PairBrownianPoly::compute(int eflag, int vflag)
...
 
-          torque[i][0] -= tx;
-          torque[i][1] -= ty;
-          torque[i][2] -= tz;
-
-          // torque due to a_pu
+          // pumping (rotational) Brownian torque: a_pu.  This is an
+          // antisymmetric pair torque (opposite sign on the two particles),
+          // so it carries csgn.  Note: as in the original code it is not
+          // scaled by vxmu2f.
```
jtclemm: Is the original code pair brownian? Or is this note referring to the prior version of brownian/poly before this PR?

src/COLLOID/pair_brownian_poly.cpp
akohlmey: I don't understand what you are asking here. All I know is that those changes have been implemented for lubricate/poly to resolve the reported bug and it has been confirmed against the non-poly version. The changes to brownian/poly were done the same way.

 Can you ping me on slack if you want to talk more about this?

src/COLLOID/pair_brownian_poly.cpp
akohlmey: In the code before this PR as far as I know.

src/COLLOID/pair_brownian_poly.cpp
jtclemm: So is this a change in the expression for `a_sq` to better account for polydispersity? 

Previously it looks like:
$a_{sq} = 6 \pi \mu R_i \left(\frac{\beta_0^2}{\beta_1^2 h_{sep}} + \log{\frac{1}{h_{sep}}} \frac{1 + 7 \beta_0 + \beta_0^2}{5 \beta_1^3} + (\mathrm{third term}) \right)$
and now it's
$a_{sq} = 6 \pi \mu R_\mathrm{ref} \left(\frac{\beta_0^2}{\beta_1^2 h_{sep}} + \beta_0 \log{\frac{1}{h_{sep}}} \frac{1 + 7 \beta_0 + \beta_0^2}{5 \beta_1^3} + (\mathrm{third term}) \right)$

Is that right?

Maybe I'm going crosseyed trying to read c++ math.

src/COLLOID/pair_brownian_poly.cpp
jtclemm: Previously on line 199 there's a term: $43 \beta_0^3$ in the calculation of `a_pu`. Now I don't see any term with a constant prefactor that's divisible by 43. It looks like the mathematical expression has changed.

I'll have to look at lubricate/poly I guess.

doc/src/Developer_write_docs.rst
jtclemm: glad there's still something it can't do ;)

src/COLLOID/pair_brownian_poly.cpp
jtclemm: Maybe this is the relevant comment: "Using h_sep in the log-order terms (and dropping the beta0 prefactor of the squeeze log term)" just added to pair lubricate poly. Not sure what this means. Maybe I need to go to Jeffry & Onishi 1984

src/COLLOID/pair_lubricateU_poly.cpp
```diff
@@ -763,12 +767,15 @@ void PairLubricateUPoly::compute_RU(double **x)
           beta[0][4] = beta[0][3]*beta[0][1];
           beta[1][2] = beta[1][1]*beta[1][1];
           beta[1][3] = beta[1][2]*beta[1][1];
-          double log_h_sep = log(1.0/h_sep);
-          double log_h_sep_beta13 = log(1.0/h_sep)/beta[1][3];
+          // symmetric dimensionless gap xi = 2*gap/(radi+radj) = 2*h_sep/beta1
+          // (Jeffrey & Onishi 1984).  Using the per-particle h_sep = gap/radi in
+          // the log-order terms breaks the i<->j resistance symmetry.  See #1933.
+          double xi = 2.0*h_sep/beta[1][1];
+          double log_h_sep_beta13 = log(1.0/xi)/beta[1][3];
```
jtclemm: is `log_h_sep_beta13` used?

src/COLLOID/pair_brownian_poly.cpp
akohlmey: I suggest you also look at the two regression tests that were added, they have explanations for what is being tested.

src/COLLOID/pair_brownian_poly.cpp
jtclemm: Seems more confusing than helpful, weird for a comment to just effectively say "We are still not performing a multiplication here"

jtclemm:
Outside of the new RNG (which I couldn't parse and therefore could not check) and the changes to mathematical expressions (which might need to be traced down in the cited work, i might check later), the other aspects of this PR look good to me.

Update: I don't have time to dig in now, but it looks like `a_pu` in the original code is Eq. 5.6 in https://doi.org/10.1017/S0022112084000355. Not clear what the replacement equation is.


One more comment, if you close #2933 you may want to note the final reply that the variable `rad` is never defined in pair_brownian_poly.cpp and will generate bogus results (or trigger seg faults) if the pairstyle is used in conjunction with fix deform. Not formally part of the issue, but could be solved by putting an error message on line 111 (or adding an appropriate definition for a polydisperse system).

src/COLLOID/pair_lubricateU_poly.cpp
akohlmey: yes. On lines 780 and 784.

akohlmey:
> Outside of the new RNG (which I couldn't parse and therefore could not check) and the changes to mathematical expressions (which might need to be traced down in the cited work, i might check later), the other aspects of this PR look good to me.
> 
> Update: I don't have time to dig in now, but it looks like `a_pu` in the original code is Eq. 5.6 in https://doi.org/10.1017/S0022112084000355. Not clear what the replacement equation is.
> 
> One more comment, if you close #2933 you may want to note the final reply that the variable `rad` is never defined in pair_brownian_poly.cpp and will generate bogus results (or trigger seg faults) if the pairstyle is used in conjunction with fix deform. Not formally part of the issue, but could be solved by putting an error message on line 111 (or adding an appropriate definition for a polydisperse system).

I added a check and it should error out now. I also added a note to the docs.
Overall, this whole set of code feels incomplete and like it needs a proper
audit from somebody that knows this stuff much better than me (or you?).

jtclemm:
> I added a check and it should error out now. I also added a note to the docs.
> Overall, this whole set of code feels incomplete and like it needs a proper
> audit from somebody that knows this stuff much better than me (or you?).

Thanks for adding that. 

Agreed on the audit. I made some headway in the equations. The change in `a_pu` is actually from another 1984 paper by the same authors (https://doi.org/10.1007/BF00952109, Eqs 3.2 and 2.26, for g21 vs. g11). It appears that these two terms were swapped for fluctuations (different resistances, shear vs. rotation). The new expression is symmetric, solving the issue, however, I would not be confident in saying this is the correct physical choice (though it may very well be). 

The extra factor of beta0 in `a_sq` does not seem to originate from the paper. I think it's just convenient rescaling to remove a factor of r_i/r_j. 

Ideally would be a question for the original authors.

akohlmey:
> Ideally would be a question for the original authors.

I think Steve tried to reach out to them (or him or her or ???) without success.
It is a bit of a niche application (like quite a few things in LAMMPS).
My look on these issues is that the changes introduced here don't make it worse. :wink:

> Agreed on the audit.

In my experience, a good approach would be to write a tutorial as it forces you to re-check all the basics from the relevant publications. But who has the time to do it? And asking a coding agent to do it would just lead to a viscous cycle.

akohlmey:
@jtclemm @sjplimp I moved the COLLOID package changes to #5015 so we have the AI generated fixes in one collection and the "human" changes in this one. ;-)

src/GPU/pair_lj_cut_tip4p_long_gpu.cpp
```diff
@@ -200,7 +200,11 @@ void PairLJCutTIP4PLongGPU::init_style()
   GPU_EXTRA::check_flag(success, error, world);
   if (gpu_mode == GPU_FORCE) {
     auto *req = neighbor->add_request(this, NeighConst::REQ_FULL);
-    req->set_cutoff(cut_coulplus + neighbor->skin);
+    // Neighbor::init() adds the neighbor skin to a perpetual request's cutoff when
```
jtclemm: This reads like a weird AI comment. Feels unnecessary to explain why not to do something that isn't done.

jtclemm (approved):
One of the comments generated by the AI in a recent commit was odd, but other than that these changes all look good to me.

Re lubrication, considering how janky the fix deform block of code was, maybe you're right that it should be considered for deprecation or have a big warning attached unless/until an expert can clean it up.

src/GPU/pair_lj_cut_tip4p_long_gpu.cpp
akohlmey: This is the bugfix. I think some comment here is important to avoid adding the bug again. The problem is that for some code path you make a request with skin and for others without. This one had it wrong.

@ndtrung81 You "own" this code. What is you preference.

jtclemm: I could be wrong, but isn't the rule just based on whether it's an occasional request or not? Or is this a special case which is different?

akohlmey: @jtclemm yes, it is something like this. It may also apply for a perpetual neighbor list requested from a fix but for a pair style it must be without. I'll shorten comment.

akohlmey: @jtclemm ... and yes, adding four words when one would do it is quite typical

jtclemm: I looked in neighbor.cpp and it does include more details on when you'd add skin, so a shorter note here does seem like a good solution.

PR #5010, opened 2026-06-02
Process-global style factory registry (persist plugins across clear and across instances)

akohlmey:
**Summary**

Make the per-style-category factory-function maps **process-global** instead of
per-instance, so that style registration is shared by all LAMMPS instances in a
process and persists across the `clear` command.

Previously each host class (`Force`, `Modify`, `Update`, `Domain`, `Output`,
`Input`, `Atom`, `AtomVecBody`, `ReadDump`) owned a `std::map<std::string,
Creator>` that was rebuilt in its constructor, i.e. on every `clear`
(destroy/create) and separately for every LAMMPS instance. The PLUGIN package,
however, registers factory functions at runtime via `dlopen` — a process-global
operation — into those per-instance maps. That mismatch meant plugin
registrations had to be re-applied after every `clear` (via `plugin_restore`)
and a second concurrent LAMMPS instance did not see plugins already loaded by
the first.

This PR introduces a single process-global registry per style category and
migrates all 16 name-keyed categories to it. As a result plugin styles persist
across `clear` with no restore step and are immediately visible to all instances
in the same process.

**Related Issue(s)**

None filed; this addresses the long-standing architectural mismatch between the
per-instance style maps and the process-global nature of plugin (DSO) loading.

**Author(s)**

**Licensing**

This pull request was developed with the assistance of an AI tool (Claude Code,
using Anthropic's Claude Opus model). Working interactively under my direction,
the AI generated the bulk of the code and the build-system changes: the
`CreatorRegistry` template (`src/creator_registry.{h,cpp}`), the marker-parsing
generators that emit the per-category `style_*.cpp` and the single
`package_registry.cpp`, the migration of every host class to the global
registries, the registry-driven `help()`/`info styles` output with the plugin
marker, and the plugin lifecycle changes (removal of `plugin_restore`, mutex
guarding of the plugin bookkeeping). The architecture and all design decisions
were mine, and I reviewed every change and validated it by building and running
the test suite. The individual commits carry a `Co-Authored-By: Claude` trailer.

**Backward Compatibility**

No input-script breakage for normal use. The only user-visible change is that
the `plugin restore` command is now an informative no-op (kept for backward
compatibility): because plugin registrations live in the global registry they
already survive `clear`, so there is nothing to restore. `info styles` and the
`-help` output now append a trailing `*` to styles currently provided by a
plugin. The generated build files change: per-category `style_*.cpp` and a
`package_registry.cpp` are now generated, and the now-unused name-keyed
`style_*.h` and all `packages_*.h` are no longer generated (only the array/mask
`style_{nbin,npair,nstencil,ntopo,gran_sub_mod}.h` remain).

**Implementation Notes**

- `CreatorRegistry<Creator>` (`creator_registry.h`) wraps an
  `unordered_map<string, {builtin, plugin}>` behind a `std::mutex`; `find()`
  prefers a plugin override and falls back to the built-in, so unloading a
  plugin transparently restores the built-in. Built-ins are registered once per
  process via `std::call_once` from the `LAMMPS` constructor (this also forces
  the generated registration TUs to be linked into a static library).
- The build scripts (CMake `StyleHeaderUtils.cmake`, `src/Make.sh`) parse the
  existing `XxxStyle(key,Class)` marker lines and emit explicit registration
  code; the `XXX_CLASS` macro is no longer defined at compile time (the
  `#ifdef XXX_CLASS` block becomes a pure parse marker). Preprocessor directives
  inside a marker block (e.g. the SNAP/INTEL `#ifdef`s) are copied through
  verbatim so build-config-dependent registrations stay guarded.
- `LAMMPS::init_pkg_lists()` is replaced by a single generated
  `package_registry.cpp` (style keyword -> package name, all packages whether or
  not enabled), used by `match_style()`/`check_packages_for_style()`.
- `plugin_restore` is removed; `pluginlist`/`dso_refcounter` mutations are now
  guarded by a `std::mutex` (RAII) so concurrent plugin load/unload from
  multiple instances is safe.

Verification: the CMake build is clean and the full `ctest` suite passes

**Post Submission Checklist**

The remaining array/mask registries (neighbor and `gran_sub_mod`) are intended
for a separate follow-up PR, since they are not dynamically loadable and may use
a different (non-global) representation.


src/Make.sh
```diff
@@ -74,29 +118,57 @@ packages () {
...
+fi
+
+# generate style_*.cpp registration translation units for migrated categories
+# args: CLASS prefix style XxxStyle base accessor regfunc one-arg "includes"
+
+if (test $cmd = "style") then
+
+  # build the portable style/registry generator (replaces awk for portability).
+  # compiled with "cc" like DEPEND/fastdep.c, so it adds no new build dependency.
+  cc -O -o "$stylegen" DEPEND/stylegen.c || exit 1
```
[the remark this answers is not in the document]
akohlmey: The current behavior is desired, the suggested change would be incompatible with cross-compilation.

src/PLUGIN/plugin.cpp
```diff
@@ -35,6 +36,14 @@ namespace {
   // map for counting references to dso handles
   std::map<void *, int> dso_refcounter;
 
+  // serializes mutations of pluginlist / dso_refcounter so that concurrent
+  // LAMMPS instances can load or unload plugins at the same time.  Only the
+  // top-level entry points (plugin_register, plugin_unload) take this lock; the
+  // helpers they call (plugin_find, plugin_erase, plugin_get_info) assume it is
+  // already held.  RAII (std::lock_guard) guarantees the lock is released even
+  // if an error handler throws out of the critical section.
+  std::mutex plugin_mutex;
```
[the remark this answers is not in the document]
akohlmey: addressed in https://github.com/lammps/lammps/pull/5010/commits/2e766571f5db4919bda74f38c5e7c92870db5836

sjplimp (approved):

sjplimp:
@stanmoore1 Can you also look at this PR to see if you think it is OK - or if there are any Kokkos issues ?

src/DEPEND/stylegen.c
```diff
@@ -0,0 +1,336 @@
...
+            char pkg[MAXTOK];
+            dir_name(argv[i], pkg, sizeof(pkg));
+            process_package_file(argv[i], macro, member, pkg);
+        }
+        return 0;
+    }
+
+    fprintf(stderr, "stylegen: unknown mode '%s'\n", mode);
+    return 1;
+}
```
rbberger, line 384: Why would you want to write *new* code like this?

akohlmey: What you mean with "why"?

src/DEPEND/stylegen.c
```diff
@@ -0,0 +1,336 @@
...
+ *   notice, this list of conditions and the following disclaimer in the
+ *   documentation and/or other materials provided with the distribution.
+ * * Neither the name of the <organization> nor the
+ *   names of its contributors may be used to endorse or promote products
+ *   derived from this software without specific prior written permission.
+ *
+ *   THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ *   AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ *   IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ *   ARE DISCLAIMED. IN NO EVENT SHALL <COPYRIGHT HOLDER> BE LIABLE FOR ANY
```
jtclemm: Who is \<COPYRIGHT HOLDER\> and \<organization\> ?

akohlmey: This is GPL boilerplate stuff and was copied from fastdep.c which in turn was adapted from a tool that shipped with some old Linux kernels (they have switched to a more sophisticated tool since).

akohlmey: @jtclemm should be corrected now for both files

src/info.cpp
```diff
@@ -851,152 +853,124 @@ bool Info::is_defined(const char *category, const char *name)
...
-    }
+  // collect visible style names (skip internal and redundant KOKKOS variants),
+  // sort them, and mark styles currently provided by a plugin with a "*"
+  std::vector<std::string> names;
+  for (const auto &name : styles.keys()) {
+    if (isupper(name[0]) || utils::strmatch(name,"/kk/host$")
+        || utils::strmatch(name,"/kk/device$")) continue;
+    names.push_back(name);
   }
+  std::sort(names.begin(), names.end());
```
rbberger, line 965: Since this shows up many times, maybe it's better to use `std::map` as backing storage? Or are we concerned about O(log N) lookup vs O(1) lookup?

src/Make.sh
```diff
@@ -35,33 +34,78 @@ style () {
...
+    pkgentries IntegrateStyle ""         integrate_styles
+    pkgentries KSpaceStyle    ""         kspace_styles
+    pkgentries MinimizeStyle  min_       minimize_styles
+    pkgentries PairStyle      pair_      pair_styles
+    pkgentries ReaderStyle    reader_    reader_styles
+    pkgentries RegionStyle    region_    region_styles
+    echo "}"
+    echo ""
+    echo "}    // namespace LAMMPS_NS"
+  } > package_registry.cpp.tmp
```
rbberger, line 101: If you go through all this effort, why not have the utility create the entire `package_registry.cpp` in one go instead of `Make.sh`. You're already creating a custom utility anyway, that has no other purpose.

src/info.cpp
```diff
@@ -851,152 +853,124 @@ bool Info::is_defined(const char *category, const char *name)
...
-  }
+  if (styles.contains(name)) return true;
 
   if (suffix_check && lmp->suffix_enable) {
     if (lmp->suffix) {
-      std::string name_w_suffix = name + "/" + lmp->suffix;
-      if (find_style(lmp, styles, name_w_suffix, false)) {
-        return true;
-      }
+      if (find_style(lmp, styles, name + "/" + lmp->suffix, false)) return true;
```
rbberger: `std::format`?

src/DEPEND/stylegen.c
rbberger: Why go with C, instead of C++17 for a higher abstraction. Also, see comment below. If this is so specialized, why not create the entire file instead of this piece-meal solution that has many moving parts?

akohlmey: This is only used by the conventional build and will go away when we drop it (which could be after the next stable release). I didn't want to invest much effort into this. The coding agent updated the process in Make.sh to use awk and I wanted to avoid adding this dependency. The step in the traditional build process where this is needed is before we have access to the CC make variable and thus we opted to go for the lowest common denominator and settle for a minimal C code. This was a long time ago when we dropped the very slow and fragile GCC dependency file creation for using `fastdep.c` and the coding agent just followed the precedent (even copied the license note that had not been customized verbatim as Joel noticed).

rbberger:
Fundamentally, the entire process of generating code with Make.sh comes from maintaining the Makefile build system. Otherwise, letting CMake detect new source files, adding it to targets and a static registration infrastructure in C++ that is being set up here would be sufficient.

src/info.cpp
akohlmey: Let me have a closer look. These are exactly the steps where a human review is so valuable. When you get a blob of code dumped in front of you, you mostly worry about whether it does what you asked for and if some detail was overlooked and less about how well it was done.

I also noticed in recent experiments that when you ask for a refactoring change, the coding agent will do what you ask for but rarely reports or even acts on any additional opportunities for improvement. Instead it has a tendency to produce over-engineered solutions that try to reproduce the original behavior too literal instead of adapting to the "spirit" of the change.

src/info.cpp
akohlmey: Let me check what it looks like.

rbberger (approved):

akohlmey:
> comes from maintaining the Makefile build

As mentioned elsewhere. We are close to the point of getting rid of it. Up to that point, I don't want to invest much effort in changing stuff around the build system. After we are CMake-only, we can make a plan for how we can restructure and clean up various steps and components. That would be the time and place to worry about stuff. There is probably still enough stuff around that was written with backward compatibility in mind. While we will have to stick with C++17 for some time with the non-KOKKOS stuff, we can probably afford to raise the minimum CMake version again and take advantage of new features.

For this pull requests and a few others, I am in two minds, on one hand I would like to have it merged as quickly as possible, even if a bit ugly, for as long as the overall design is sound, so we get to test it properly before the stable release. But I am also eager to have this done cleanly and improved, so waiting has its merits if there are good suggestions for significant improvements. So far, I have invested very little personal effort into this, so I don't worry much about having to spend time cleaning it up. There is also the option to give the coding agent more specific instructions and scrap this and let it do it again. It still gets us to have this refactor done much faster than if I would do it all by hand.

src/info.cpp
akohlmey: @rbberger I had a closer look and I don't think it makes a difference either way. Vibe coding has not yet reached the point where we had so many styles of all kinds that the lookup time or the memory consumption matters. Having an explicit sort makes it explicit that the returned data is sorted. So there is a tiny bonus to make it more obvious that the returned list is sorted. But this can be changed without much problem at a later point.

src/info.cpp
akohlmey: yup. confirmed. this reads better.

akohlmey:
@Maria-Lesniewski @jtclemm This simplifies the management of modular substyles by having a table instead of fragile and hard to read pre-processing code in the core styles (which is also removed by this PR). Since all substyles are shiped and always configured, there is no benefit for a dynamic registration. If you ever think of a way that people would want to load a submodule with the plugin command at runtime, we have to look into making it more flexible (and complex). Please let me know ASAP if you have any concerns.

Maria-Lesniewski:
Hello! I do not foresee any future issues with static registration for package BOCS/ I see no need for more complex code. Thank you for making me aware of the change and for adding the contribution directions within the registration file.

stanmoore1 (changes requested):
I want to test this for KOKKOS before it is merged.

akohlmey:
@stanmoore1 FYI, this works on my home desktop with KOKKOS/HIP. 
https://github.com/lammps/lammps/pull/5010/commits/d22b7af8b4abe0cd4a839d8dd983d6962bebe7f6
Were the only KOKKOS specific changes needed.

stanmoore1:
I tested KOKKOS CPU/GPU overlap with `/kk/host` suffix on H100. Everything is working as expected.

stanmoore1 (approved):

PR #5011, opened 2026-06-03
PERI: fix per-type-pair critical stretch bond breaking (#984), with cleanup and tests

### LINKED ISSUE #984: Problem with critical strain s0 in pair_style peri/pmb
## Summary

In the pair_peri_pmb.cpp source the implementation of equation 9 of www.sandia.gov/~mlparks/papers/pd-lammps-cpc.pdf is slightly different from what stated by the theory; the present implementation happens to work when all the alpha's and the s00's are the same, but fails to make the crack easier when we have a smaller s00 at the interface. 

## Type of Issue

Bug Report

## LAMMPS Version (Bug Report)

LAMMPS 30March2018 version

## Expected Behavior (Bug Report)

From our understanding of the theory as presented in www.sandia.gov/~mlparks/papers/pd-lammps-cpc.pdf (equation 9 in particular), the parameters s00 and alpha need not be the same for all the pair interactions (at least in principle). Our idea was to use a smaller s00 for the interactions between the two blocks initially at contact, so to determine an area where a crack is more likely to initiate. We would prefer this solution as opposed to the one we found suggested in the mailing list, i.e. to weaken the 'effective spring constant' of the Peridynamic bond, as this choice would affect also the short-range repulsion (eq.10a of the same article).

## Actual Behavior (Bug Report)

This fails to work in the present implementation. Specifically, in the pair_peri_pmb.cpp source we find the line: (L266)

MAX(s0_new[i],s00[itype][jtype] - (alpha[itype][jtype] * stretch))    

not to implement equation 9 mentioned above: instead of really minimizing the stretch within the horizon of particle i, in the code the total quantity s0 is maximized. 
This implementation happens to work when all the alpha's and the s00's are the same, but fails to make the crack easier when we have a smaller s00 at the interface (because the MAX will not select that smaller value). If different alpha's and s00's (depending on the particles type) appear within the horizon of the i particle, the MAX over the total quantity s0 is in general different from a combination of the appropriate s00 for that bond with the MIN over the stretch for that particle.

We tried to modify pair_peri_pmb.cpp (see attachments), but our version currently works only in the special case alpha=0.

## Steps to Reproduce (Bug Report)

This is a simple&quick input file (also attached) which illustrates the problem we are referring to:

#--------------------------------------------------------------------------
# INITIALIZATION
#--------------------------------------------------------------------------

units           si
boundary        p p p
atom_style      peri
atom_modify     map array
neighbor        0.0001 bin

#--------------------------------------------------------------------------
# SETUP
#--------------------------------------------------------------------------

# TRIALBOX
lattice         sc 0.00025
region          trialbox block 0 0.006 0 0.006 0 0.006 units box
create_box      2 trialbox

############################################

region          constraintA block 0.00275 0.00326 0.00275 0.00326 INF 0.00025 units box
region          fixedA block 0.00275 0.00326 0.00275 0.00326 0.0002501 0.00075 units box
region          movingA block 0.00275 0.00326 0.00275 0.00326 0.0007501 0.00125 units box

############################################

# CREATE ATOMS (bottom to top)
create_atoms    1 region constraintA
create_atoms    1 region fixedA
create_atoms    2 region movingA

############################################

# GROUPS (bottom to top)
group           constraint region constraintA
group           fixed region fixedA
group           moving region movingA
group           lower union constraint fixed

#--------------------------------------------------------------------------
# POTENTIAL
#--------------------------------------------------------------------------

pair_style      peri/pmb

# PMMA --> parameters from 10.1016/j.prostr.2016.06.022
pair_coeff      *   *   7.0247e+22   0.000751      0.013      0.0

#pair_coeff      1   2   7.0247e+22   0.000751      0.0065      0.0

#--------------------------------------------------------------------------
# PROPERTIES SETUP
#--------------------------------------------------------------------------

set             group all density 1180
set             group all volume 1.5625e-11
velocity        all set 0.0 0.0 0.0 sum no units box

#--------------------------------------------------------------------------
# FIXES
#--------------------------------------------------------------------------

fix             F1 lower nve
fix             F2 moving move linear NULL NULL 1.0 units box
fix             F3 constraint setforce 0.0 0.0 0.0

compute         Cdam all damage/atom

#--------------------------------------------------------------------------
# RUN
#--------------------------------------------------------------------------

timestep        5e-08
thermo_style    custom step time #...
thermo          20
dump            D1 all custom 20 dump.peri id type x y z c_Cdam

run             20000

#--------------------------------------------------------------------------

The results obtained with both LAMMPS compiled with the original pair_peri_pmb.cpp and the one compiled with our fixed version are identical because here all s00's are the same.
However, if line number 50

#pair_coeff      1   2   7.0247e+22   0.000751    0.0065   0.0

is uncommented, the difference is clear. The fixed version produces a crack at the interface between particles of type 1 and particles of type 2 as their s00 is just half of that acting between 1-1 and 2-2
interactions. The uncorrected LAMMPS instead produces exactly the same trajectory as with the commented line (all equal s00's).

However, note that this example has alpha=0. Our version would fail with any alpha != 0, e.g. the usual alpha=0.25 choice.
The reason for this failure is the need to break the bonds symmetrically as prescribed by the min in the first line of equation 9. We tried to implement this symmetrization of the bond breakage in the commented lines from L283 to L295, but the results are quite bad.

[pair_peri_pmb.cpp.gz](https://github.com/lammps/lammps/files/2151822/pair_peri_pmb.cpp.gz)
[in.pmb.peri.gz](https://github.com/lammps/lammps/files/2151823/in.pmb.peri.gz)
www.sandia.gov/~mlparks/papers/pd-lammps-cpc.pdf

akohlmey:
**Summary**

Fixes the per-type-pair critical-stretch bond-breaking criterion in the PERI
package (issue #984), plus a package-wide review with dead-code removal,
typo/error-message fixes, modernization, new unit tests, and documentation
updates.

The PMB-style breaking criterion collapsed the per-bond critical stretch
`s0 = s00 - alpha*smin` into a single per-particle scalar (the maximum of
`s00 - alpha*s` over a particle's bonds). That is only valid when `s00` and
`alpha` are identical for all type pairs; with type-dependent coefficients a
bond inherited the critical stretch of the surrounding bulk bonds, so a
deliberately weakened interface (a smaller `s00` on the 1-2 type pair) would
never fracture.

The fix stores the geometric minimum stretch in a new per-atom field
`atom->smin` and evaluates the critical stretch *per bond* at break time as
`s00[ij] - alpha[ij]*max(smin_i, smin_j)`. This equals `min(s0_i, s0_j)` for
uniform coefficients (bit-for-bit backward compatible) and honors the
per-type `s00`/`alpha` otherwise. The per-atom `s0` diagnostic is kept and
filled exactly as before. Applied to `peri/pmb`, `peri/lps`, `peri/ves`,
`peri/eps` and the `peri/pmb/omp` and `peri/lps/omp` accelerator variants.

Commit 2 is an independent PERI-wide cleanup (no functional change to the
existing examples, verified bit-for-bit):
- remove dead members `PairPeri::elastic_energy` (declared, destroyed and
  exposed via `extract()` but never allocated) and `FixPeriNeigh::r1` /
  `FixPeriNeigh::thetaValue` (declared, never used);
- fix a copy/paste warning in `compute damage/atom` (it warned about
  "dilatation/atom"), `"Illegal compute Dilatation/atom"`, and the
  "wighted"/"instanteneous" comment typos;
- modernize the bond-count report to `utils::logmesg` and accumulate the
  count as a `bigint` (the previous `int` total could overflow for large
  models);
- use `MY_PI` instead of a hardcoded `3.141592653589793` in `peri/eps` and
  `peri/ves` (bit-identical, matches `peri/lps`).
- `peri/eps` would allocate `theta` only to `atom->nlocal` but it is used
   in a forward communication to store it with ghost atoms. allocate to atom->nmax

**Related Issue(s)**

Fixes #984

**Author(s)**

**Licensing**

By submitting this pull request, I agree, that my contribution will be
included in LAMMPS and redistributed under either the GNU General Public
License version 2 (GPL v2) or the GNU Lesser General Public License version
2.1 (LGPL v2.1).

These changes were developed with the assistance of an AI tool (Claude Code,
model Claude Opus) used interactively under the direction and review of the
author. The AI assisted with the root-cause analysis, the implementation of
the bond-breaking fix across the pair styles and atom-style plumbing, the new
unit test, the documentation edits, and the dead-code/typo cleanup. All
changes were reviewed, tested, and verified by the author before submission
(consistent with the `Co-Authored-By` trailers on the commits).

**Backward Compatibility**

- For uniform `s00`/`alpha` the bond-breaking decisions are bit-for-bit
  identical to before (the `peri/pmb` shatter example reproduces the
  reference log exactly), so existing inputs behave unchanged.
- The per-atom `s0` value reported by `compute property/atom s0` keeps its
  previous meaning and values. The documentation for it is corrected: earlier
  docs incorrectly described `s0` as the maximum bond stretch; it is the
  per-particle critical stretch.
- `atom_style peri` gains a per-atom field (`smin`), so the binary restart
  layout for peri systems changes: restart files written with this version
  are not interchangeable with those from older versions. (No data-file format
  change; `smin` is not written to or read from data files.)

**Implementation Notes**

- `smin` is added as a regular per-atom field (registered in `Atom`, listed in
  `AtomVecPeri`'s grow/copy/comm/border/exchange/restart/create field lists)
  so it migrates with atoms and is ghosted, exactly like `s0`. The pair styles
  keep a transient `smin_new` scratch buffer (mirroring `s0_new`) because the
  break test reads the previous step's `smin` of neighbors while the new
  values are accumulated.
- The "no bonds / first compute" sentinel for `smin` is `-DBL_MAX` (so the
  implied critical stretch is `+infinity`), which preserves the documented
  "no bond breaking on the first timestep" behavior; the `+DBL_MAX` value used
  to initialize the per-step MIN accumulator is mapped back to `-DBL_MAX` for
  bondless atoms.
- Verification: new `unittest/force-styles/test_pair_peri.cpp` (interface
  fracture for pmb/lps that fails on the old code, below-threshold no-break,
  `s0 == s00 - alpha*smin` relation, momentum conservation, base-vs-`/omp`
  consistency, and a `/kk`-ready case that skips until a KOKKOS port exists);
  all four examples reproduce their reference logs for the bond-based case and
  are otherwise unchanged; `make check` (whitespace/permissions) is clean.

**Post Submission Checklist**

### DIFF
--- a/doc/src/Howto_peri.rst
+++ b/doc/src/Howto_peri.rst
@@ -581,11 +581,22 @@ as broken in a simulation by removing them from the bond family
 
 A naive implementation would have us first loop over all bonds and
 compute :math:`s_{min}` in :ref:`(9) <peris0>`, then loop over all bonds
-again and break bonds with a stretch :math:`s > s0` as in
+again and break bonds with a stretch :math:`s > s_0` as in
 :ref:`(8) <perimu>`, and finally loop over all particles and compute forces
 for the next step of :ref:`Algorithm 1 <algvelverlet>`. For reasons of
-computational efficiency, we will utilize the values of :math:`s_0` from
-the *previous* timestep when deciding to break a bond.
+computational efficiency, we instead store the per-particle minimum stretch
+:math:`s_{min}` from the *previous* timestep and, when deciding whether to
+break a bond, form its critical stretch :math:`s_0 = s_{00} - \alpha\,
+s_{min}` from that bond's own :math:`s_{00}` and :math:`\alpha`.  This keeps
+the criterion correct when :math:`s_{00}` and :math:`\alpha` differ between
+atom-type pairs (for example, a deliberately weakened interface).
+
+.. versionchanged:: TBD
+
+Earlier versions stored a single per-particle critical stretch :math:`s_0`
+(the maximum of :math:`s_{00} - \alpha s` over a particle's bonds), which
+reproduces :ref:`(9) <peris0>` only when :math:`s_{00}` and :math:`\alpha`
+are identical for all atom-type pairs.
 
 .. note::
 
--- a/doc/src/compute_property_atom.rst
+++ b/doc/src/compute_property_atom.rst
@@ -91,7 +91,7 @@ Syntax
 
            PERI package per-atom properties:
            vfrac = volume fraction
-           s0 = max stretch of any bond a particle is part of
+           s0 = critical bond stretch (bond-breaking threshold) of a particle
 
   .. parsed-literal::
 
--- a/doc/src/pair_peri.rst
+++ b/doc/src/pair_peri.rst
@@ -151,6 +151,42 @@ details please see the description in "(Mitchell2011a)".
 
 ----------
 
+Bond breaking criterion
+"""""""""""""""""""""""
+
+For all of these styles a peridynamic bond between particles *i* and *j*
+breaks irreversibly once its stretch :math:`s = (r - r_0)/r_0` exceeds a
+critical stretch.  Following :ref:`(Parks) <Parks>` (eq. 9), the critical
+stretch of a particle is :math:`s_0 = s_{00} - \alpha\, s_{min}`, where
+:math:`s_{min}` is the minimum (most compressive) stretch over all of the
+particle's bonds, and a bond breaks when :math:`s > \min(s_{0,i},
+s_{0,j})`, i.e. symmetrically from the point of view of both particles.
+
+The s00 and :math:`\alpha` coefficients may be chosen differently for each
+pair of atom types.  This makes it possible, for example, to assign a
+smaller s00 to the bonds across an interface between two materials so that
+a crack preferentially initiates there.  The critical stretch is evaluated
+*per bond* using that bond's own s00 and :math:`\alpha` together with the
+geometric :math:`s_{min}` of each endpoint.
+
+.. versionchanged:: TBD
+
+In previous versions the critical stretch was stored as a single
+per-particle value computed as the maximum of :math:`s_{00} - \alpha s`
+over a particle's bonds.  That is only equivalent to the criterion above
+when s00 and :math:`\alpha` are identical for all type pairs; with
+type-dependent coefficients a bond would incorrectly inherit the critical
+stretch of the surrounding bulk bonds and a weakened interface would not
+fracture.  The criterion is now evaluated per bond so that type-dependent
+s00 and :math:`\alpha` behave as intended.
+
+The :doc:`compute property/atom <compute_property_atom>` *s0* per-atom
+property reports this per-particle critical stretch.  Earlier versions of
+this documentation incorrectly described *s0* as the maximum stretch of
+any bond a particle is part of.
+
+----------
+
 .. include:: accel_styles.rst
 
 ----------
--- a/src/OPENMP/pair_peri_lps_omp.cpp
+++ b/src/OPENMP/pair_peri_lps_omp.cpp
@@ -27,6 +27,7 @@
 #include "neighbor.h"
 #include "suffix.h"
 
+#include <cfloat>
 #include <cmath>
 
 #include "omp_compat.h"
@@ -56,9 +57,11 @@ void PairPeriLPSOMP::compute(int eflag, int vflag)
 
   if (atom->nmax > nmax) {
     memory->destroy(s0_new);
+    memory->destroy(smin_new);
     memory->destroy(theta);
     nmax = atom->nmax;
     memory->create(s0_new,nmax,"pair:s0_new");
+    memory->create(smin_new,nmax,"pair:smin_new");
     memory->create(theta,nmax,"pair:theta");
   }
 
@@ -111,6 +114,7 @@ void PairPeriLPSOMP::eval(int iifrom, int iito, ThrData * const thr)
 
   double *vfrac = atom->vfrac;
   double *s0 = atom->s0;
+  double *smin = atom->smin;
   double **x0 = atom->x0;
   double **r0   = fix_peri_neigh->r0;
   tagint **partner = fix_peri_neigh->partner;
@@ -267,6 +271,7 @@ void PairPeriLPSOMP::eval(int iifrom, int iito, ThrData * const thr)
     ztmp0 = x0[i][2];
     itype = type[i];
     jnum = npartner[i];
+    smin_new[i] = DBL_MAX;
     first = true;
 
     for (jj = 0; jj < jnum; jj++) {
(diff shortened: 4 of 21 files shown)


jtclemm:
Cleaning up PERI is a good task, but I believe an even better solution is to just port the force models to bond styles via the BPM package and scrap the existing code. This is something that's has been floated to be done this summer. 

If I recall correctly, the PERI package hard codes both the solid mechanics + contact models into a single pair style so users can't mix and match. Tying the solid models to a pair style also prohibits control over the solid geometry. The pair styles automatically create pseudo-bonds from the initial neighbor list so users can't create cracks, place solid objects adjacent to each other, output connectivity, etc. 

The schema in the BPM package is more flexible and works with other LAMMPS commands since it just uses actual bonds. Using a single framework for both packages would also remove a lot of cryptic code for fitting the solid models in a pair style (PERI pair styles can just be children of BondBPM which includes all the needed infrastructure). 

I can review these changes and we can merge them, but they may ultimately be a stop gap.

akohlmey:
> I can review these changes and we can merge them, but they may ultimately be a stop gap.

I am fine with this. I only would suggest that when this gets superseded by BPM we need a "porting guide" added to the Peridynamics Howto and then add a statement in the docs and a warning in the implementation that we discourage the use of the PERI package and will eventually remove it (~ 1-2 years later). Same as we did with other, similar cases.

One more piece of advice. Don't treat any of the pull requests that I created with AI support the same way as if I wrote all the changes myself. The way they come to be requires comparatively little effort from me (I need to be very careful when writing the prompt and typically have to make some adjustments to the AI suggested plan and some polishing afterwards) and thus won't be bothered by dropping such a pull request (unlike those where I spent a lot of work on), or parts of it. Ideally, if you can provide some additional guidance, I can use that (and still open or already closed pull request) to repeat the programming task and give it another try. Also, for any issue that has been unresolved for a long time, even a partial fix is preferable over doing nothing. Or when the proposed change should be dropped we should change the documentation and explain.

As I get more practice *and* for as long as I have sufficient access to a frontier model, there should be improvements when re-doing a such fixes, since I am asking to record my personal feedback on suggested solutions.

stanmoore1 (approved):
It looks OK to me, but I am no peridynamics expert either.

sjplimp:
@akohlmey I would also like to review this one before merging

sjplimp (changes requested):
still want to review

sjplimp:
@akohlmey @jtclemm The summary says this is fixing an issue with the PMB PD style, but the change is applied to
multiple PD styles.  Possibly all 4 of them, though the code for EPS looks a bit different.  Are we confident that is the correct thing to do?

One of these 4 styles should be the original PD implementation in LAMMPS, possibly PMB, but not sure since we later changed the names as more variants were added.  The original PD was implemented with help from Mike Parks who was part of the PD group at Sandia.  So I would be surprised if the bond-breaking criteria in that version was incorrect or incomplete.  Unless it was a later refinement in PD itself (completely independent of LAMMPS), which the LAMMPS version never caught up with.

akohlmey:
> The summary says this is fixing an issue with the PMB PD style, but the change is applied to multiple PD styles. Possibly all 4 of them, though the code for EPS looks a bit different. Are we confident that is the correct thing to do?

This pull request started as a bugfix for issue #984, but upon closer inspection, the issue was present in all peri variants. For peri/eps I also discovered a memory allocation bug: the theta array was used for forward communication but only allocated to atom->nlocal. 
As mentioned in the description for the ceramic disk example, the fix here does not make a difference.

There is a bit of a risk, but on the other hand the code before this pull request is *definitely* incorrect. We *will* get a thorough check when @jtclemm's plan's to recast the PERI package into the BPM framework will happen.

sjplimp:
ok - I will approve it - I support the @jtclemm plan to support PD within BPM if that can be done staightforwardly, since it is actively developed.  It would also be nice to know if BPM/PD is as performant as the current PD, since it is a fairly simple model (at least the original PD in LAMMPS).

sjplimp (approved):

PR #5012, opened 2026-06-03
fix dump vtk parallel summary referencing wrong piece files

akohlmey:
**Summary**

Fix the parallel summary file (`.pvtp`/`.pvtu`) written by `dump vtk` so it references
the correct per-processor piece files. Previously, for any dump file name containing more
than one `.` (e.g. `dump.reference.*%.vtp`), the `<Piece Source=...>` entries pointed at
nonexistent files (e.g. `dump_0.vtp`) and a spurious extra piece file was created, so the
distributed output could not be opened in ParaView without manual post-processing.

**Related Issue(s)**

No GitHub issue. Reported on the MatSci user forum:
https://matsci.org/t/66845 ("small problem with vtk files .pvtp").

**Author(s)**

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS
and redistributed under either the GNU General Public License version 2 (GPL v2) or the
GNU Lesser General Public License version 2.1 (LGPL v2.1).

This pull request was prepared with the assistance of an AI tool (Claude Code, model
Claude Opus 4.8). The AI performed the root-cause analysis and generated the new
`DumpVTK::write_pvtk()` and `DumpVTK::pvtk_piece_filename()` methods and the edits to
`write_vtp()`/`write_vtu()` in `src/VTK/dump_vtk.cpp` (and the matching header
declarations), under human review and direction. The XML structure was cross-checked
against the VTK XML file-format specification.

**Backward Compatibility**

No input-script changes. The names of the actual per-processor piece files are unchanged.
The change only corrects the *content* of the generated `.pvtp`/`.pvtu` summary file (which
was previously wrong) and stops creating one stray piece file per snapshot. Anyone who had
worked around the bug by post-editing the `.pvtp` no longer needs to.

**Implementation Notes**

Root cause: `write_vtp()`/`write_vtu()` delegated the summary file to VTK's
`vtkXMLPPolyDataWriter` / `vtkXMLPUnstructuredGridWriter`. Those classes derive the piece
file names from the summary file name by stripping everything after the *first* `.`
(`vtksys` "complete extension" behavior), so `dump.reference.000000.pvtp` yields the piece
prefix `dump` and references `dump_0.vtp`, `dump_1.vtp`, ... — which do not match the
per-rank files LAMMPS itself writes (`dump.reference.000000_0.vtp`, ...). The parallel
writer also emitted its own (wrongly named) piece data file as a side effect.

Fix: LAMMPS already writes all per-rank piece files itself with correct names, so the two
VTK parallel-writer code paths are replaced by a hand-written summary file:
- `DumpVTK::write_pvtk(int)` emits the `PPolyData`/`PUnstructuredGrid` XML directly
  (`VTKFile` header with runtime-detected byte order, `GhostLevel="0"`, `PPointData`
  declarations built from the existing `vtype`/`vector_set`/`name` maps using the same
  iteration as `reset_vtk_data_containers()`, `PPoints`, and one `<Piece>` per rank).
  `PPoints` precision is read from the live `vtkPoints` object so it matches the piece
  files rather than being hardcoded. No `PCellData` is written because all dump-vtk fields
  are point data.
- `DumpVTK::pvtk_piece_filename(int)` builds each `<Piece>` `Source` with the same
  `%`->`_<id>` substitution and `utils::star_subst()` timestep handling used to create the
  actual files, stripped to a path relative to the summary file.

The now-unused `vtkXMLPPolyDataWriter`/`vtkXMLPUnstructuredGridWriter` includes were
removed and `<cstdio>` added.

Verification: the file-name construction was checked with a standalone reproduction and
produces `dump.reference.000000_0.vtp` ... `_3.vtp` for the reported case, handles
sub-directory paths (relative reference) and the `dump*.%.suffix.vtp` pattern. NOTE: this
was not compiled/run against the VTK library locally (VTK is not installed in the
development environment used / `PKG_VTK=OFF`), so a build with `-D PKG_VTK=on` and a
multi-rank visual check in ParaView still needs to be confirmed by CI or a reviewer.

**Post Submission Checklist**

- Forum report: https://matsci.org/t/66845
- VTK XML file format reference: https://docs.vtk.org/en/latest/vtk_file_formats/vtkxml_file_format.html

### DIFF
--- a/src/VTK/dump_vtk.cpp
+++ b/src/VTK/dump_vtk.cpp
@@ -42,6 +42,7 @@
 #include "variable.h"
 
 #include <cmath>
+#include <cstdio>
 #include <cstring>
 #include <sstream>
 #include <vector>
@@ -60,7 +61,6 @@
 #include <vtkPolyData.h>
 #include <vtkPolyDataWriter.h>
 #include <vtkXMLPolyDataWriter.h>
-#include <vtkXMLPPolyDataWriter.h>
 #include <vtkRectilinearGrid.h>
 #include <vtkRectilinearGridWriter.h>
 #include <vtkXMLRectilinearGridWriter.h>
@@ -68,7 +68,6 @@
 #include <vtkUnstructuredGrid.h>
 #include <vtkUnstructuredGridWriter.h>
 #include <vtkXMLUnstructuredGridWriter.h>
-#include <vtkXMLPUnstructuredGridWriter.h>
 
 using namespace LAMMPS_NS;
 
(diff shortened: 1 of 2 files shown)


akohlmey:
I have tested and confirmed the fix on Fedora 44 with vtk 9.5.2

sjplimp (approved):

PR #5013, opened 2026-06-03
[BUGFIX] GCMC region keyword: graceful translation reject (#66883) + faster region+mol gas-list rebuild (#225)

### LINKED ISSUE #225: [Feature Request] Slowdown for fix gcmc with regions and molecule files
Reported on lammps-users by Evan Lowry and confirmed by Aidan Thompson as follows:

```
The loss of performance occurs when combining the region and mol keywords.
After every MC move, the entire list of active GCMC atoms/molecules is
regenerated from scratch. With region and molecule, this requires calculating
the the center-of-mass of every molecule. This turns out to be a very expensive
operation.  The best solution would be replace the full list generation with a
list update operation.
```

Example input files and output are available at the links below.

[log_gcmc_region.txt](https://github.com/lammps/lammps/files/537413/log_gcmc_region.txt)
[log_gcmc.txt](https://github.com/lammps/lammps/files/537416/log_gcmc.txt)
[in_gcmc.txt](https://github.com/lammps/lammps/files/537417/in_gcmc.txt)
[in_gcmc_region.txt](https://github.com/lammps/lammps/files/537414/in_gcmc_region.txt)
[system.txt](https://github.com/lammps/lammps/files/537418/system.txt)
[mtn.txt](https://github.com/lammps/lammps/files/537415/mtn.txt)

akohlmey:
This PR groups two related changes to `fix gcmc`, both around the `region` keyword
(same source file, `src/MC/fix_gcmc.cpp`).

**Summary**

1. **Graceful translation reject (forum #66883).** A region-restricted GCMC *translation*
   trial move that could not be placed inside the region within `max_region_attempts`
   (1000) tries aborted the entire simulation via `error->one()`, unlike the insertion
   paths which skip the trial move. All four translation routines now reject the
   unplaceable move and continue, emitting a one-time warning.
2. **Faster region+mol gas-list rebuild (issue #225).** `update_gas_atoms_list()` rebuilt
   the in-region gas list by looping over every molecule ID and calling `Group::xcm()`
   (an O(natoms) scan + an `MPI_Allreduce`) for each -- O(nmolecules*natoms) work and
   nmolecules collective calls per rebuild, run after every successful exchange/move. It
   now does a single pass over local atoms plus one `MPI_Allreduce`.

**Related Issue(s)**

- Forum report https://matsci.org/t/66883 (region-restricted moves appearing to hang).
- closes #225 (`fix gcmc` slowdown with `region` + `mol`).

**Author(s)**

Axel Kohlmeyer, Temple University.

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in
LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2)
or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

The code, the example input, and the documentation changes in this pull request were
generated with the assistance of Claude Code (Claude Opus 4.8). The MPI-collective-safety
design, the gas-list refactor, and the verification strategy were directed and reviewed by
the author. Consistent with the `Co-Authored-By` trailers on the commits.

**Backward Compatibility**

No input syntax changes. Behavior change for change (1): a region-restricted translation
that cannot be placed within 1000 attempts no longer aborts; it is rejected (still counted
as an attempt) with a one-time warning. Change (2) is behavior-preserving (see below).
Documented with `versionchanged:: TBD` in `doc/src/fix_gcmc.rst`.

**Implementation Notes**

*Translation reject* -- all four translation routines reject via a shared
`warn_region_reject()` helper, MPI-collective-safe: molecule variants (synchronized RNG +
Allreduce'd COM) `return` together; `attempt_atomic_translation` falls through to its
existing `MPI_Allreduce(&success)`; `attempt_atomic_translation_full` broadcasts the
owner-only failure with a one-int `MPI_Allreduce(MIN)` so all ranks skip `energy_full()`
together. The "outside box" checks (a genuine error) remain `error->one()`.

*Gas-list rebuild (#225)* -- the per-molecule `Group::xcm()` loop is replaced by one pass
accumulating each molecule's mass-weighted unwrapped position, one `MPI_Allreduce` over all
molecules, then normalize by the gas molecule mass and remap. Per rebuild this drops from
O(nmolecules*natoms) work + nmolecules collectives to O(natoms) work + 1 collective; the
collective-count reduction is the scalability win at high MPI rank counts. The refactor is
**bit-for-bit identical** to the previous algorithm: verified by building the old
`Group::xcm` path (with the indexing fix) and the new path side by side and confirming
identical trajectories for a region+mol CO2 GCMC run at 1 and 4 MPI ranks. It also fixes a
latent one-past-the-end read (COM arrays were sized `maxmol_all` but indexed by molecule ID
up to `maxmol_all`; now indexed into a `maxmol_all+1` buffer).

A new example `examples/mc/in.gcmc.region` (with reference logs for 1 and 4 ranks) showcases
the `region` keyword and documents a commented-out variant that triggers the change-(1)
warning.

**Post Submission Checklist**


akohlmey:
@athomps my home machine is not large enough to verify the performance impact of the optimization. Please check it out and either approve and re-assign the pull request to me when confirmed or suggest what else to do (e.g. revert or apply changes)

athomps (approved):
@akohlmey Both of these improvements look good. I tested the new molecule list builder on up to 480 cores (5 nodes of Hops) using the CO2 example and a region that is the entire box.  Both the old and new codes should give the same output (on average) as the original CO2 script with no region, and hopefully the new code will run much faster than the old, and only a bit slower than `no region`.  This turned out to be true on all counts, with the bonus that the old code exhibit a lot of hangs.  I did not explore why, but it likely had some correctness issues.  
Here is a comparison of timings for `no region` and `region` with the new code.  For this small example, 1 core is always fastest, but the runtimes are consistent between both cases, which indicates that the time spent in rebuilding the molecule list is now negligible, as we hoped.  
```
nprocs  Tnoregion Tregion (new) 
  [-]    [s]     [s]
  1      2.2     2.4
  2      5.9     4.6
  4      6.2     5.8
  8      6.4     7.0
  15     7.5     8.1
  30     9.8     9.4
  60    11.0    10.5
  120   10.8    10.4
  240   16.4    17.2
  480   25.3    26.1
```

PR #5015, opened 2026-06-03
More collected small changes

### LINKED ISSUE #648: [Suggestion for an Enhancement] Introduce Double-band EAM potentials
## Summary
Introduce in LAMMPS the posibility to use the double-band EAM potentials, where the contributions from s-band and d-electrons are treated separately. An example of such a system is FeCr, which is widely used as a model alloy for many construction materials in civil and nuclear/fusion energetics.

## Type of Issue

Suggestion for an Enhancement

## Detailed Description (Enhancement Suggestion)

The EAM potentials with double-band support should be integrated  in LAMMPS from the box. 

The description of the double-band EAM potential formalism can be taken from:

- P. Olsson et al.. Two-band modeling of alpha-prime phase formation in Fe-Cr. PHYSICAL REVIEW B 72, (2005) 214119 [Link to the article](https://journals.aps.org/prb/abstract/10.1103/PhysRevB.72.214119)
- G. Bonny et al. Iron chromium potential to model high-chromium ferritic alloys. Philosophical Magazine Vol. 91, No. 12, 21 April 2011, 1724–1746  [Link to the article](http://www.tandfonline.com/doi/abs/10.1080/14786435.2010.545780)

### LINKED ISSUE #1933: [BUG] _pair_lubricate_poly.cpp_issues

### LINKED ISSUE #2933: [BUG] pair lubricate/poly+brownian/poly leads to strange attractions and a temperature increase

### LINKED ISSUE #3543: [BUG] _Unexpected qeq/point result_
**Summary**

qeq/point for two atoms behaves as expected. qeq/point on four atom system gives apparently incorrect result. qeq/dynamic is correct for that four atom system. Script will be attached below. This problem might also exist in other qeq implementations.

LAMMPS (24 Mar 2022), windows 10, wsl system.

**Expected Behavior**

Four atom chain spaced linearly with spacing of 2. The first and last atoms are assigned charge of 1 and -1. The middle two atoms are subjected to qeq/point. The middle two atoms are the same species, thus same chi (3) and eta (2). The expected result is 1, -1/6, 1/6, -1. However, qeq/point gives 1, 1/6, -1/6, -1. In fact, even with the initial correct charge on atom 2 and 3, upon qeq/point, the charge changes, with total PE INCREASES from -3.8 to -1.4. All charges are in terms of +e. All energies are eV. The interaction here is Coul/cut to simplify the verification of the qeq result (mathematica).

Qeq/dynamic behaves as expected for this four atom system.

**Actual Behavior**

However, qeq/point gives 1, 1/6, -1/6, -1.

**Steps to Reproduce**

```
lmp_serial < in.qeqbug
Charge distribution is in lata4olivia dumpfile as
=======================================================
ITEM: TIMESTEP
0
ITEM: NUMBER OF ATOMS
4
ITEM: BOX BOUNDS pp pp pp
0.0000000000000000e+00 2.0000000000000000e+02
0.0000000000000000e+00 2.0000000000000000e+02
0.0000000000000000e+00 2.0000000000000000e+02
ITEM: ATOMS id type x y z vx vy vz fx fy fz q
1 1 6 10 10 -13.8882 -10.731 1.42973 0.849979 0 0 0.166667
2 1 8 10 10 13.0414 -16.8753 24.4422 -0.849979 0 0 -0.166667
3 1 4 10 10 -3.87453 21.1627 -9.68082 -0.0499988 0 0 1
4 1 10 10 10 4.72134 6.44364 -16.1911 0.0499988 0 0 -1
=========================================================

###########qeq parameterfile  param.qeq2 ####################
===========================================================
# UNITS: metal
1     3.00000         2       0.0   0.        0.000000
2     3.00000         2        0.0   0.        0.000000
============================================================

############lammps script: in.qeqbug #################
================
units           metal
atom_style      charge

lattice		fcc 2
region		box block 0 100 0 100 0 100
region		r1 block 3 3 5 5 5 5
region		r2 block 4 4 5 5 5 5
region          r3 block 2 2 5 5 5 5
region          r4 block 5 5 5 5 5 5

create_box	2 box
create_atoms	1 region r1
create_atoms	1 region r2
create_atoms	1 region r3
create_atoms	1 region r4
mass		* 1.0

pair_style      coul/cut 20.0
pair_coeff      * *

group		gr1 region r1
group		gr2 region r2
group		gr3 region r3
group		gr4 region r4

set		group gr1 charge -0.1666
set		group gr2 charge 0.1666
set		group gr3 charge 1.0
set		group gr4 charge -1.0

neighbor        1.0 bin

run		0

group           type1 type 1
compute         charge1 type1 property/atom q
compute         q1 type1 reduce ave c_charge1
group           type2 type 2
compute         charge2 type2 property/atom q
compute         q2 type2 reduce ave c_charge2
variable        qtot equal count(type1)*c_q1+count(type2)*c_q2
variable        nqeq equal f_2

thermo_style    custom step pe c_q1 c_q2 v_qtot v_nqeq
thermo          10
thermo_modify   format line "%4d %12.9g %12.8f %12.8f %16.12f %6.0f"

timestep        0.0001

fix             1 all nve

velocity  	all create 300.0 1281937 


group		g12 union gr1 gr2
# QEQ/Point
fix        	2 g12 qeq/point    1 10 1.0e-8 500 param.qeq2

# QEQ/dynamic
#fix        	2 gr1 qeq/dynamic    1 10 1.0e-6 100 param.qeq2

dump            4 all custom 10 lata4olivia id type x y z vx vy vz fx fy fz q
run		0

================
```

### LINKED ISSUE #5027: Potential bug in  fix_rigid_small.cpp
https://github.com/lammps/lammps/blob/5af1318e9eddbb1a79c582dfa4f61adcd91962a2/src/RIGID/fix_rigid_small.cpp#L150

There is potentially a bug in this line where it should read `else bodyID[i] = 0` instead  `else bodyID[0] = 0`

akohlmey:
A collection PR for small and local changes (a follow-on to the existing large
collection PR, which is under active review). More small changes may be added to
this branch until it goes under review.

---

### 1. `change_box` / `write_data`: apply the `thermo_modify lost` policy (forum 63775)

`change_box` could silently leave atoms outside a non-periodic boundary. Its
`move_atoms` path uses `Irregular::migrate_atoms()`, which **clamps** out-of-box
atoms onto the nearest proc (`comm->coord2proc`) rather than removing them,
unlike `comm->exchange()` during a run. So its lost-atom check never fired,
`atom->natoms` was left inconsistent, and a later `write_data` wrote a data file
whose header atom count disagreed with the number of `Atoms`-section lines when
`thermo_modify lost` was `warn` or `ignore` (https://matsci.org/t/63775).

- `change_box`: after the box change, cull atoms outside non-periodic boundaries
  (as a reneighbor would), rebuild the atom map, and apply the `thermo_modify
  lost` policy: `error` (default) aborts, `warn` warns and continues, `ignore` is
  silent. `atom->natoms` is reset so the system stays self-consistent.
- `write_data`: same policy when the recomputed count differs from
  `atom->natoms`; `warn`/`ignore` reset `atom->natoms` so the file is consistent.

Verified by reproduction (`error` aborts; `warn`/`ignore` cull to the correct
count and write a header==body file; lossless `change_box` unchanged), serial and
4 MPI ranks.

### 2. QEQ: fix wrong-sign charges for a subset fix group (issue #3543)

`fix qeq/point` (and the other s/t-method variants `qeq/shielded`, `qeq/slater`,
`qeq/ctip`, which share the base-class code) gave **wrong-sign** equilibrated
charges, with the energy *increasing*, when applied to a subset of the atoms
while the rest carry fixed charges. `FixQEq::init_storage()` initialized both the
`s` and `t` auxiliary vectors to the current charge for all atoms. Initializing
`s` to the fixed charge of out-of-group atoms is intentional (it supplies their
charge as the external field via the matrix-vector product), but doing the same
for `t` -- which only encodes the group's charge-neutrality response -- pollutes
the neutralization (`u = sum_s/sum_t`) and flips the sign. Fix: initialize `t` to
0 for atoms outside the fix group.

Verified: `qeq/point` now matches `qeq/dynamic` (and minimizes the energy) on the
4-atom reproducer from #3543; the common full-group case is a no-op
(`examples/qeq` output unchanged). The fix is in the shared base class, so it
corrects the same defect in `qeq/shielded`/`slater`/`ctip`; `qeq/dynamic` and
`qeq/fire` use a different method and are unaffected.

### 3. DOC+TOOLS: two-band (s+d) EAM via `hybrid/overlay` (refs #648)

Two-band ("double-band") EAM models -- e.g. the Fe-Cr potentials of Olsson et al.
(Phys. Rev. B 72, 214119 (2005)) and Bonny et al. (Phil. Mag. 91, 1724 (2011)),
requested in #648 -- add an extra "s-band" embedding term to the standard EAM
energy. Grouping the terms shows this is *exactly* a `hybrid/overlay` of two
`eam/fs` potentials, so it already runs with existing styles and needs no source
changes:

```
E_i = [1/2 sum_j V + F_d(rho_d)]  +  [F_s(rho_s)]
      \___ a standard eam/fs ___/     \__ eam/fs with V == 0 __/
```

The cross-only s-band density (zero for like pairs, nonzero only for unlike
pairs) requires the Finnis-Sinclair (`eam/fs`) form -- it cannot be expressed as
`eam/alloy`; the s-band file zeroes the pair (`r*phi`) section.

- `doc/src/Howto_eam_overlay.rst`: new Howto with the derivation, why the s-band
  must be `eam/fs`, the `pair_style hybrid/overlay` recipe, how to build the
  s-band file, and a `compute pair` energy-decomposition check. Linked from
  `Howto.rst` and `pair_eam.rst`.
- `tools/eam_generate/two_band_fecr.py`: writes the s-band `eam/fs` file
  (cross-only Slater density, `F_s = c1*sqrt(rho)+c2*rho^2+c3*rho^4`, zero pair);
  literature presets plus a self-contained, runnable `--demo` mode.

Verified by running the generated demo: the overlay reproduces
`E_total = E_dband + E_sband` exactly; a pure element has zero s-band energy; an
alloyed site has a nonzero, localized s-band term. The new Howto page builds
cleanly with no Sphinx warnings.

### 4. CMake: find `omp.h` for Clang toolchains that omit it (ROCm `hipcc`)

Some Clang-based compilers -- most notably `hipcc` from ROCm -- do not ship
`omp.h` in the compiler's own resource directory, so building with
`-D BUILD_OMP=on` (e.g. the GPU package with `GPU_API=hip` and a `hipcc` host
compiler) failed to compile host code that includes `<omp.h>`, even though the
`-fopenmp` flag is accepted. When `BUILD_OMP` is on and the C++ compiler is
Clang-based, CMake now checks the compiler's resource directory for `omp.h` and,
if it is missing, adds a matching-version system Clang's `omp.h` as a fallback
include path via `-idirafter` (searched last, so it does not shadow other
headers). A `.. note::` in the GPU/HIP build docs explains the workaround.

Verified: a `BUILD_OMP=on GPU_API=hip` configure+build with `hipcc` (ROCm 6.4,
clang 19) now compiles all host OpenMP sources; previously it failed with
`'omp.h' file not found`.

### 5. GPU: make `GPU_ARCH` the canonical architecture setting for all backends

The GPU package used `GPU_ARCH` for `GPU_API=cuda` but `HIP_ARCH` for
`GPU_API=hip`, so `-D GPU_ARCH=...` was silently ignored for HIP builds.
`GPU_ARCH` is now the single canonical architecture variable for all `GPU_API`
backends, with a backend-appropriate default (`sm_75` for cuda and hip/nvcc,
`gfx906` for hip/amd, `spirv` for hip/spirv). The backend-specific `HIP_ARCH` variable is still accepted but deprecated: setting it prints
a CMake `DEPRECATION` warning and migrates its value to `GPU_ARCH` (the canonical
setting wins if both are given). This is a step toward making `GPU_ARCH` the only
documented option and deriving/retiring the others. `Build_extras.rst` updated
accordingly.

Verified (`GPU_API=hip`, ROCm 6.4): `-D GPU_ARCH=gfx1103` now sets
`--offload-arch=gfx1103` with no warning, while `-D HIP_ARCH=gfx900` still works
but prints the deprecation warning and uses `gfx900`.

### 6. GPU: fix pre-existing pair-style test failures on AMD shared-memory iGPUs

Five independent, pre-existing defects exposed by the GPU package unit tests on an
AMD Radeon 780M (gfx1103 APU, shared host/device memory); none are specific to
this branch (all reproduce on `develop`). With these fixes the OpenCL and HIP
`(Pair|KSpace)` suites both pass 352/352 at double precision (previously 10 OpenCL
and 3 HIP failures). One commit per fix; see the commit messages for details:

- **OpenCL sub-buffer misalignment (Geryon):** on shared-memory GPUs `_host_view()`
  created a sub-buffer at a data-dependent origin not aligned to
  `CL_DEVICE_MEM_BASE_ADDR_ALIGN`, failing with `CL_MISALIGNED_SUB_BUFFER_OFFSET`
  and aborting `sw`/`tersoff`/`vashishta`/`gayberne`. Fall back to the
  discrete-memory copy path (mapped host pointer) on misalignment.
- **`coul/long/cs` fast kernel:** the special-bond force term was
  `-(1 - factor_coul)` instead of `-factor_coul`, giving grossly wrong forces on
  all backends (energy was correct).
- **HIP host neighbor lists on AMD APUs:** device neighbor builds page-faulted
  (`lj/cut/dipole/cut`) or hit the ellipsoid/sphere-mix restriction (`gayberne`);
  detect integrated AMD GPUs and force host neighbor lists under HIP, matching the
  long-standing OpenCL behavior (also fixes an `std::string::find` truthiness bug
  in the OpenCL detection).
- **`lj/cut/tip4p/long`:** the neighbor-request cutoff double-counted the neighbor
  skin and exceeded the communication cutoff in the host-neighbor path.

Verified: OpenCL `(Pair|KSpace)` 352/352 on this branch; HIP 352/352 on a sibling
branch carrying byte-identical changes.

### 7. fix `pair_style lubricate/poly` not obeying Newton's third law for polydisperse (unequal radii) particles (issue #1933):

the near-field lubrication resistance functions (Jeffrey & Onishi, 1984) were nondimensionalized by the central particle's radius, so the pairwise force computed on particle *i* did not equal minus the force computed on particle *j*. The squeeze-term (`a_sq`) log coefficient was missing a `beta0 = radj/radi` factor (which the analogous shear term `a_sh` does include), and all log-order terms used the per-particle gap `h_sep = gap/radi` instead of the symmetric dimensionless gap `xi = 2*gap/(radi+radj)`. Adding the `beta0` factor and using `xi` restores exact pairwise force symmetry (verified to machine precision) and is a no-op for equal radii, so monodisperse results are unchanged. The same correction is applied to the `lubricate/poly/omp` accelerator variant and, for consistency, to `lubricateU/poly` (`compute_Fh`/`compute_RU`/`compute_RE`). A new regression test `unittest/commands/test_pair_lubricate_poly.cpp` asserts net force balance for polydisperse configurations (it fails on the previous code and passes with this fix).

### 8. fix `pair_style brownian/poly` not obeying Newton's third law for the pairwise *random* Brownian force (issue #2933):

the style runs under `newton off` on a full neighbor list and applied the random force only to the local atom *i*, drawing an *independent* random number for the `(i,j)` and `(j,i)` visits, so the forces on the two members of a pair were unrelated rather than equal and opposite. This injected net linear momentum (and kinetic energy) every step and heated the system. The random force is now drawn once per pair from a deterministic, order- and MPI-rank-independent RNG (keyed on the unordered pair of atom IDs, the timestep, the style seed, and a per-component stream index) and applied equal and opposite to both particles, so linear momentum is conserved exactly (verified to machine precision, and identical across MPI ranks and OpenMP thread counts). The near-field resistance functions are also switched to the symmetric Jeffrey & Onishi gap `xi = 2*gap/(radi+radj)` with the `beta0` squeeze prefactor, matching the issue #1933 `lubricate/poly` fix, so that by the fluctuation-dissipation theorem the Brownian variance equals the lubrication friction regardless of which particle is taken as the reference. The same fix is applied to the `brownian/poly/omp` accelerator variant, which previously also depended on the OpenMP thread count. A new regression test `unittest/commands/test_pair_brownian_poly.cpp` asserts net force balance for polydisperse configurations.

### 9. unittest: fix exit crashes with host-only KOKKOS builds

In a KOKKOS build with only host backends (serial and/or OpenMP) and testing enabled, every test program whose tests initialized Kokkos segfaulted at program exit *after* all tests had passed: the test driver mains never called `lammps_kokkos_finalize()` (unlike `main.cpp`), so Kokkos statics were destroyed in undefined order and crashed in a fence issued from `~SerialInternal`. The shared `main()` of the force-styles and c-library drivers now releases Kokkos/Python/plugins after `RUN_ALL_TESTS()`. Additionally, the kokkos tests in `test_library_external` requested 4 threads unconditionally (a fatal error with a serial-only KOKKOS library) and now fall back to one thread like the force-styles drivers. This affected 23 force-styles tests plus `LibraryExternal`; it was never seen in CI because the unittest workflow does not enable KOKKOS, and KOKKOS GPU builds skip these sub-tests. All 655 tests pass in the host-only KOKKOS configuration now.

### 10. use `__restrict` with MSVC in the `_noalias` macro

MSVC (and clang-cl) support the `__restrict` extension but fell through to the empty definition of `_noalias`, so the OPENMP and OPT package styles lost the no-aliasing promise on their hot-loop pointers when compiled with MSVC.

**Related Issue(s):**

closes #3543;
Closes #648 
Forum post: https://matsci.org/t/63775.
Fixes #1933 
Fixes #2933 
Fixes #5027 

**Author(s):** Axel Kohlmeyer, Temple University.

**Licensing**

in LAMMPS and redistributed under either the GNU General Public License version 2
(GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

The code and documentation changes in this pull request were generated with the
assistance of Claude Code (Claude Opus 4.8). The designs were specified and
reviewed by the author, and the changes were verified by reproduction. Consistent
with the `Co-Authored-By` trailers on the commits.

**Backward Compatibility**

No input syntax changes. `change_box` with the default `lost error` now aborts
(instead of silently continuing) if it pushes atoms outside a non-periodic
boundary. The QEQ change is a no-op for the standard full-group case. The two-band EAM
addition is documentation plus a standalone tool only; no code or input
syntax changes. The GPU CMake changes are backward compatible: the deprecated `HIP_ARCH` variable still works (with a warning), and the `omp.h` handling only adds a fallback include path when the compiler lacks the header. On AMD integrated GPUs (APUs) the HIP build now defaults to host-built neighbor lists (matching the existing OpenCL behavior on such devices); this is a correctness fix, not a syntax change.

**Post Submission Checklist**


akohlmey:
@sjplimp @jtclemm @ndtrung81 I would like to get this pull request included as soon as possible since it fixes a few annoying issues in the post merge tests.

If you need me to make any changes or move changes to a different branch, please let me know.

sjplimp:
@akohlmey I don't think it's a good idea to combine multiple complex bug fixes in a single PR.  Unless you are going to be sure to get specific approval for each of them before merging.  Which in this case requires several approvals instead of one.

I think 7,8 need to be approved by @jtclemm 
5,6 by @ndtrung81 
2,3 by @athomps 
9 by @stanmoore1 
1 by me -- which I approve, but want to verify that change_box is still using migrate_atoms() with an added clean-up operation afterwards to remove out-of-box atoms ?? - the (AI generated?) summary is unclear to me on this
4,10 - don't know who is best

akohlmey:
> I don't think it's a good idea to combine multiple complex bug fixes in a single PR.

I agree that these "collections" have a tendency to mushroom out of control every time we have a review backlog. Let us talk today about possible ways to deal with this and perhaps some alternate strategy that I can use.

I will use the given the division of possible reviewers to split this into multiple pull requests later and close this one for now.

PR #5021, opened 2026-06-05
Refactor long-range dispersion styles to idiomatic LAMMPS (structural, no behavior change)

akohlmey:
**Summary**

This refactors the family of long-range *dispersion* pair and k-space styles
originally written by Pieter J. in 't Veld (and collaborators) so they follow
the prevailing LAMMPS coding idiom. These files use many old-fashioned C
shortcuts -- manual pointer arithmetic, sentinel-pointer loops, function-like
macros, magic bit-flag literals, and parallel-array linear-search dispatch --
that make them hard to read and unfamiliar compared to the rest of the code base.

Styles touched:

* `pair_style lj/long/coul/long` (+ `/omp`, `/opt`)
* `pair_style buck/long/coul/long` (+ `/omp`)
* `pair_style lj/long/tip4p/long` (+ `/omp`)
* `pair_style lj/long/dipole/long`
* `kspace_style ewald/disp`

The changes are **structural only** and behavior-preserving (bit-for-bit within
the regression-test tolerances); the numerical kernel *expressions* are left
unchanged in this PR. A follow-up will modernize the inner kernel math itself.

Key changes:

* the magic `ewald_order`/`ewald_off` bit literals (`(1<<1)`, `(1<<6)`, `0x42`,
  `0x4a`, `64`, `2`) are replaced by a named `enum {EWALD_COUL, EWALD_DISP[,
  EWALD_DIPOLE]}`; `(ewald_off^-1)` is spelled `~ewald_off`
* `options()` takes an explicit flag mask and dispatches with `strcmp` instead
  of a parallel-array linear search + `switch`
* `extract()` is rewritten as explicit if-chains; the vdW/dispersion cutoff is
  now exposed under the key `cut_vdwl` consistently across the family, with
  `cut_LJ` retained as a deprecated alias. The only in-tree consumer,
  `pppm/disp` (and `pppm/disp/tip4p` by inheritance), is updated to `cut_vdwl`.
* compute kernels: manual pointer arithmetic (`x0+(i+(i<<1))`, `f0+3*i`,
  `memcpy`) and sentinel-pointer neighbor loops are replaced with standard
  indexed `ilist`/`jlist` loops and `x[i]`/`f[i]` indexing; assignment-in-
  condition is split out; the shadowing force-update scratch variable is removed
* `ewald/disp`: the `C_RMULT`/`C_CRMULT`/`C_SET`/`C_CONJ`/`C_ANGLE` function-like
  macros are converted to type-safe `static inline` functions; the
  `COMPLEX_NULL`/`EWALD_NSUMS` brace-macros are inlined; the dead
  `//#define DEBUG` toggle is removed

The intricate k-vector structure-factor loops in `ewald/disp`, and the dense
inner force/energy expressions in all kernels, are intentionally left untouched
in this PR.

**Related Issue(s)**

None.

**Author(s)**

Axel Kohlmeyer (Temple University), @akohlmey

**Licensing**

in LAMMPS and redistributed under either the GNU General Public License version 2
(GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

The structural refactoring in this pull request was carried out with the
assistance of an AI coding tool (Claude Code, Claude Opus) driven and reviewed by
the human author. The AI applied the mechanical, behavior-preserving
transformations described in the summary (flag enums, `options()`/`extract()`
rewrites, loop and pointer-arithmetic modernization, and macro-to-inline-function
conversion). The numerical kernel expressions were carried over verbatim from the
original code. Every change was verified by the human author against the existing
force-style and k-space regression tests.

**Backward Compatibility**

No user-facing behavior changes. The internal `Pair::extract()` key `cut_LJ` is
retained as a (now deprecated) alias of the new `cut_vdwl` key, so any external
consumers of that key continue to work.

**Implementation Notes**

Verified with the CMake build (KSPACE, OPENMP, OPT, INTEL, DIELECTRIC packages
enabled) and the `unittest/force-styles` regression suite: all `lj_long*`,
`buck_long*`, `buck_table*`, `lj_long*tip4p*`, `lj_long*dipole*`, `ewald_disp*`,
and `pppm_disp*` tests pass, including each accelerated variant exercised by the
test driver under `-sf omp -pk omp 4` and `-sf opt`. `make
check-whitespace/permissions/homepage/errordocs` are clean. No behavior change is
intended or observed.

The deeper rewrite of the numerical kernel expressions (expanding the packed
Horner forms, renaming single-letter scratch variables, splitting compound
assignments) is deliberately deferred to a follow-up so this PR stays easy to
review as behavior-preserving.

**Second pass (in response to review)**

Following the review comment that `ewald_disp.cpp` still reads very differently
from `ewald.cpp`/`ewald_dipole.cpp`, a second, deeper pass was applied to that
file (still strictly behavior-preserving):

* every pointer-sentinel atom loop (`while (x<xn)` with `*(x++)`) is now an
  indexed loop over `nlocal` using `x[i]`/`f[i]`/`q[i]`/`type[i]`/`mu[i]`
* the running output pointers (`*(ke++)`, `(cek++)->im`, `(--bi)[0]`) are
  replaced by explicit running indices into `kenergy`/`kvirial`/`sfac_all`
* the `zx/zy/zz` pointer dance that builds the per-atom `e^(ikr)` table is an
  indexed recursion written directly into the table (also removes a scratch
  `new[]`+`memcpy`+`delete[]` per atom per step)
* raw `new[]`/`delete[]` plus a hand-maintained byte counter are replaced by
  `memory->create()/destroy()` and a `memory_usage()` computed from sizes
* members/methods are renamed to the `ewald.cpp`/`ewald_dipole.cpp` vocabulary:
  `compute_ek()` -> `eik_dot_r()`, `coefficients()` -> `coeffs()`,
  `compute_slabcorr()` -> `slabcorr()`, `nbox` -> `kmax`, `nkvec` -> `kcount`,
  `unit` -> `unitk`, `cek_local/cek_global` -> `sfac/sfac_all`
* the confusingly named `function[]` term-enable array is now `termflag[]`,
  indexed by named constants (`TERM_COUL`, `TERM_DISP_GEOM`, `TERM_DISP_ARITH`,
  `TERM_DIPOLE`/`TERM_DISP_NONE`) shared via `ewald_const.h`; the same rename is
  applied consistently to `pppm/disp` and its OPENMP/INTEL/DIELECTRIC variants
* dead code is removed (the never-called `compute_surface()`, commented-out
  blocks, stale "should this be different?" questions in comments), and the
  k-vector ordering invariant that the `kx`/`ky` caching in all the k-loops
  relies on is now documented where the list is built
* new regression test `kspace-ewald_disp_arith.yaml` covers the previously
  untested seven-channel arithmetic-mixing path

Verification of the second pass: forces, torques, global and per-atom
energy/stress are **bit-for-bit identical** to the previous code over 4-step
runs covering geometric/arithmetic/dipole mixing, triclinic boxes, the slab
correction, per-atom output, `pppm/disp`, and a 4-rank MPI run; all
`ewald_disp*`/`pppm_disp*`/`lj_long*`/`buck_long*` force-style tests pass.

What is *not* attempted here: reorganizing `ewald/disp` onto the `cs/sn`
recursion-array template of `ewald.cpp` (k-outer/atom-inner loops, per-term
structure-factor arrays).  That changes summation order, so results would only
match within tolerance instead of bit-for-bit, and it changes the performance
profile; it belongs in the follow-up PR together with the kernel-math
modernization (with its own performance comparison).  The requested
before/after performance table for the changes in *this* PR will be posted as
a separate comment.

**Post Submission Checklist**


stanmoore1 (commented):
I looked at `ewald_disp.cpp` and it still doesn't feel like normal LAMMPS code, e.g. it is still vastly different than `ewald.cpp` and `ewald_dipole.cpp`.

stanmoore1 (changes requested):
@akohlmey I think we need assurance that there is no significant performance regression, e.g. a performance table of before and after on a few different compilers

akohlmey:
> I looked at `ewald_disp.cpp` and it still doesn't feel like normal LAMMPS code, e.g. it is still vastly different than `ewald.cpp` and `ewald_dipole.cpp`.

OK, I will dig into this again and ask for more changes. If you can find there time, can you point out some of the hardest to read parts so I can specifically ask to have those rewritten?

akohlmey:
Some performance numbers for `lj/long/coul/long cut long 12.0 ` using GCC 15.2 `-O2 -g ` 1500 bulk water molecules serial with pppm. Loop times in s, average over 3 runs 1 CPU:

| compiler | before | after | non-disp |
| --- | --- | --- | --- |
| GCC 15.2 | 31.0  | 30.7 | 33.8 |
| Clang 19.0 | 31.5 | 31.7 | 32.7 |

stanmoore1 (approved):
This is much improved.

PR #5025, opened 2026-06-06
GPU package: avoid NaN energies with chipStar (HIP SPIR-V)

pvelesko:
The `lib/gpu` block-wide energy/virial reduction assumes warp-synchronous execution, which the SPIR-V/OpenCL model does not guarantee — producing NaN energies/stresses on chipStar/Intel GPUs (forces are unaffected). Enable the existing `LAL_NO_BLOCK_REDUCE` flag for `HIP_ARCH=spirv` to accumulate energy/virial on the host instead.

Verified on Intel PVC (Aurora) and Intel UHD 770: GPU pair-style unit tests go from 28 NaN failures to 0 (297/297 pass).


ndtrung81 (approved):
@pvelesko Thanks very much for the fix!

PR #5028, opened 2026-06-10
Add compute frenkel for on-the-fly Frenkel-pair defect analysis

### LINKED ISSUE #3452: [Feature Request] Add compute/dump functionality to identify defects
**Summary**

Add compute/dump functionality to identify, count, and output point defects (see [1,2]).

**Detailed Description**

When simulating a primary knock-on atom (PKA) coming from a collision by a high energy particle, the resulting damage cascade is typically composed of point defects, such as vacancies and interstitials. The large simulation boxes required can lead to large data dumps that need to be post-processed in order to extract the point defects, a process that can be expensive. This problem has been addressed in [1,2] through a new compute/dump style, and the question here is whether that approach can added to LAMMPS, rather than used as an add-on.

Greetings @hammondkd! Would it be possible to incorporate your add-on directly into LAMMPS?

[1] https://doi.org/10.1016/j.cpc.2019.106862
[2] https://data.mendeley.com/datasets/p39kb4fshj/1

akohlmey:
## Summary

Adds `compute frenkel` to the EXTRA-COMPUTE package: an on-the-fly Wigner-Seitz analysis that counts vacancies and interstitials (Frenkel pairs) during a run, without post-processing. Useful for following radiation-damage cascades, thermal defect formation, or any process that creates/annihilates point defects.

It is a port of Karl Hammond's published compute/dump style (Comput. Phys. Commun. 247, 106862 (2020)), updated to current LAMMPS conventions.

**Related Issue(s)**

Implements and closes #3452 

**Author(s)**

Karl Hammond, University of Missouri
Axel Kohlmeyer, Temple U

**Licensing**

Adapting and refactoring the compute was mostly done by Claude Opus 4.8
under the supervision of Axel Kohlmeyer.

**Implementation Notes**

- **Reference sites** come from the most recent `lattice` and are generated the way `create_atoms` would, but without adding atoms or creating groups/running commands as side effects. An explicit site file and a region/group restriction are optional.
- **Outputs**: a global vector `[n_vac, n_int, n_irregular]` (so `c_ID[1]` is the Frenkel-pair count), a 2x20 cluster-size histogram, a per-atom distance-to-nearest-site, and a local array of defect clusters (id, signed size, x, y, z) for `dump local`.
- **Visualization reuses `dump image ... compute`** (one sphere per defect cluster) instead of adding a separate dump style; vacancies/interstitials get color index 1/2 and use the existing `ccolor`/`acolor` mechanisms.
- **Modernization**: std::vector/std::string instead of new/delete, `TextFileReader` + `platform::file_is_readable()`, `memory->grow()`, format-style error messages with `Error::NOLASTLINE`, `modify_param`, no `using namespace std`.
- Adds `MPI_PROC_NULL`/`MPI_REQUEST_NULL`/`MPI_STATUSES_IGNORE` to the serial MPI STUBS.
- **Parallel design note**: ghost lattice sites are intentionally *not* binned, so the atom-to-site assignment always lands on a *local* site (occupancy is tallied by the owning rank); cross-subdomain defect clusters instead connect through each ghost site's neighbor list (built from its clamped edge bin), which `find_clusters` merges and `MPI_Allreduce` propagates -- rebuilding the bins to include ghosts would route occupancy through ghost indices and is deliberately avoided.
- **Post-review refactoring** (results unchanged; checked on 1/2/4 ranks): `tagint`-safe cluster IDs (LAMMPS_BIGBIG); `std::sort`/`std::unique` de-duplication; `std::vector` (not `std::list`) neighbor lists; an `unordered_map` for O(1) ghost-tag lookup; small fixed MPI tags (below `MPI_TAG_UB`); `compute_modify drvac`/`drint` now propagate to the derived search/bin cutoff, with a warning if it exceeds the ghost communication cutoff; the four-array ghost exchange collapsed into a single `exchange_one()` helper.

Validated against an independent Python Wigner-Seitz reference analysis and on a 2 keV bcc-Fe cascade (peak ~143 Frenkel pairs, a few surviving).
Tested to produce the same results in serial and in parallel on a systems where the domain decomposition splits the cluster(s) (and fixed a hang that appeared after refactoring).

**Post Submission Checklist**

defects
peak
cascade


src/EXTRA-COMPUTE/compute_frenkel.h
```diff
@@ -0,0 +1,112 @@
...
+#else
+
+#ifndef LMP_COMPUTE_FRENKEL_H
+#define LMP_COMPUTE_FRENKEL_H
+
+#include "compute.h"
+#include "region.h"
+
+#include <list>
+
```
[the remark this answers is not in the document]
akohlmey: This is not required. The LAMMPS documentation states explicitly that all classes derived from Pointers can depend on a list of headers already included (to avoid redundancies with commonly used headers) '<string>' and '<vector>' are in this list.

hammondkd:
Thanks for working on this! I admit that I had forgotten about it for the last few years. I am glad you were able to eliminate some of the more hack-ish things I did in there.

I don't know that getting rid of the dump style makes sense: the trouble with cascade damage, for example, is that you often do not know what the "best" angle is to view it at until after the fact, so having a dump file with the defects in it (not just an image) is helpful. In fact, I wrote the dump style long before I wrote the compute. Ideally, I guess, it would dump the defects in any style the user wanted, but I don't know how to do that easily (or whether it's possible) for the sites, which are effectively atoms that do not exist.

One of the reasons I had not initiated a pull request for this feature years ago is because there was a difficult-to-trace bug in it that I wasn't able to figure out. I suspect it has something to do with the damage cascade spanning two or more processor domains, but I never found the bug (and then forgot about it entirely). My relatively thorough testing did not trip it, but a user or two has run into the problem, and I have as yet been unable to help them. Perhaps the changes you made fixed the problem (unwittingly)?

I will work on compiling this and checking that it runs, and I'll try to reproduce the bug if I can.

sjplimp (changes requested):
would like to review this

akohlmey:
> I don't know that getting rid of the dump style makes sense

With the dump style and with the corresponding hacks, this would never make it into the distribution. Sorry.

But things are not as glum as you might think:
- you can use dump style local to output the data. 
- when using a recent version of LAMMPS-GUI you can check the view quasi-interactively.

If you need a demonstration, contact me at my personal email and a) I can give you the (unfinished and raw) tutorial material I would like to use it for and you can check it out, and b) can give you a demonstration of how to check things out interactively.

hammondkd (approved):
I ran a few tests on this, and it occurred to me that the bug I mentioned was probably confined to the *dump* style, not the compute, so given that and the tests I have already run on the code you put in there, I approve (pending any edits from @sjplimp et al.).

sjplimp:
A few Qs about this. which are not clear to me from the doc page or the AI summary.

a) does this work with periodic and non-periodic dimensions ?
b) what is the definition of a "cluster" (one of the compute outputs)
c) in the 2d diagram of a simple cubic lattice with one vacancy and one interstitial ...
    it seems like for a thermalized lattice, the interstitial could be assigned (randomly) to any of 
      4 lattice sites since it on a corner points of 4 unit cells
    if it is assigned to a site that is not adjacent to the vacancy site, does that mean it is not in a cluster with it ?
    which seems like it might not allow it to be classified as a Frenkel pair ?
d) how are drvac and drint actually used?  The description of a vacancy and interstitial just talks about counting atoms in
    in each unit cell - which doesn't seem like a cutoff is needed
    if a cutoff distance from a reference site is used, it talks about counting 0 or 2 for vac/int sounds like a single cutoff
e) why is the frenkelgroup not always the compute group ?  doc page doesn't explain how the compute group is used ?
f) does the rescale keyword allow for tricllinic boxes ?
    e.g. a sheared system which produces Frenkel pair defects as the shear increases
g) does this compute work in a restarted simulation?
   the read/write restart commands don't talk about lattice info persisting - if not, does that affect this command?
h) is the site file option a one-time output?  When is it done ?
    The doc page doesn't explain the format of this file or what sites are output
    if it is like a dump file, is some kind of parallel dump operation encoded within this compute ?

Also a couple Qs about the internal workings of the compute.

1) is there any info within this compute which persists during a running simulation?  e.g. reference sites
    or are the site coords recalculated at every compute invocation?
2) what is the purpose of the MPI_Allreduce which is mentioned in the summary?
     does that mean this compute will not scale to very large simulations ?

akohlmey:
> A few Qs about this. which are not clear to me from the doc page or the AI summary.

@sjplimp you should explicitly direct these questions to @hammondkd who wrote/designed the compute originally. I mostly adapted it for current LAMMPS, dropped the parts that we would not want to include by recasting it to use dump style local, and then added the doc changes that you requested previously. What you are asking now goes *far* deeper.

All I can say is that I have used it on a periodic system at finite temperature with great success and obtained consistent results and that I don't see why it should not work for non-periodic systems.

sjplimp:
@hammondkd is following this PR, so my comments are to either of you.

My (a-h) are about the doc page.  It needs to clarify what the command and code does or does not do.
My (1-2) are about the code.  I'm not asking for changes (yet) but just want to understand how it works.

akohlmey:
@sjplimp In an attempty of speeding things up, I've asked the coding agent to address your questions by evaluating the source code and comparing the documentation with what is actually implemented. I've edited the (detailed and verbose) responses down to the essentials. If you want to see it all, just let me know and I'll email it to you.

@hammondkd please correct me where I misstate the original design intent.

Doc questions:

a) periodicity

It works in any combination. There is one yet undocumented caveat: a system with a vacuum area will result in a lot of false positives for vacancies. The documentation should advise to use the "region" keyword for those cases to restrict where the compute is active.

b) definition of a cluster

A cluster is a connected component of defect sites: every site with occupancy != 1
is a defect site, and two defect sites of any kind are connected when they are closer
than *drvac* (if the neighbor site is empty) or *drint* (if the neighbor site is
multiply occupied).   The reported cluster "size" is the net defect content, i.e.
(number of multiply-occupied sites) - (number of empty sites);  The cluster ID is the smallest reference-site atom-ID in the cluster. It should be documented that clusters with a net 0 size are not included in the histogram, but its interstitials and vacancies are counted toward their respective totals.

c) thermalized lattice

The designation of Frenkel pairs is independent from the clustering. If the distance between the interstitial and vacancy is beyond the threshold, they will be listed as single defect clusters. The assignment to individual cell may fluctuate between snapshots.
To account for thermal expansion, the documentation already recommends either using the "rescale yes" keyword or recording the trajectory frames and then quenching them with a minimization before analysis in case thermal fluctuations are large.

d) how drvac and drint are used

They are used for the clustering (as mentioned in b) not the assignment to cells.
Internally, MAX(drvac,drint) is used for binning to identify each atoms nearest site
and as ghost site communication. The documentation is misleading because it referred
this internal property without the reader knowing about it.

e) frenkelgroup usage

This is a leftover from the original implementation that added particles to represent the defects. This version has this feature removed. The defect locations are available to dump image for direct visualization and as a local vector for output with dump local.
The code currently does not use the compute group, but probably should. That would be helpful in multi-species systems, e.g. He in W where you want the analysis only performed for the W atoms.

f) rescale with triclinic

*rescale yes* currently does not handle shear. Adding this would imply that the reference cells are deformed the same way across the system. There are two options: error out in case that is detected or implement support for shearing with the given caveat.

g) restart support

Like with most other computes, there is no restart support. lattice, compute, and compute_modify settings must be repeated. This should be transparent unless *rescale yes* is used. In that case the lattice command has to use the rescaled lattice constant.

h) site_file

It is not an output but an input that would replace the lattice generated reference sites with data from the file. This is to handle the use case where the reference is not a perfect crystal. The documentation needs to be expanded to spell this out.


Code questions:

1) persistent state

Only the list of reference sites is built once per run, everything else is regenerated.
There is a limitation that needs to be addressed: this is incompatible with fix balance and comm_style tiled.

2) collective communications

There are a few, all rather small and similar to what thermo output requires.
During code analysis, one parallel scaling issue was found: the handshake
that sizes the ghost-site exchange currently posts a point-to-point count message
to *every* other rank, although only the up-to-26 neighboring subdomains can have
nonzero counts. This can be improved in the code.

The following documentation changes are suggested:
* rewrite the drvac/drint description (cluster connection radii, not detection
  cutoffs) and document the defaults with their rationale
* document the cluster definition, net-size convention, and the size-0 case
* add the vacuum/free-surface caveat for non-periodic dimensions
* clarify site_file as an input, with file format
* add the restart paragraph incl. the rescale caveat
* document group semantics after we settle (e); code: drop *frenkelgroup*, honor
  the compute group in the occupancy analysis

The following code changes are suggested:
* restrict the ghost-count handshake to neighboring subdomains
* guard (or support?) changing tilt with *rescale*, mid-run rebalancing,
  and comm_style tiled

sjplimp:
@akohlmey - I agree with are your doc/code suggestions.

One more code Q - how is the list of sites persisted in normal operation,  since they are not per-atom data.
Not asking about load balancing (another important issue you mention), but when a proc sub-domain
dynamically grows/shrinks on reneighboring steps.  That seems different than ghost site exchange.
I would call it site migration.

Also, point-to-point with all other ranks sounds bad.

akohlmey:
> One more code Q - how is the list of sites persisted in normal operation, since they are not per-atom data.
> Not asking about load balancing (another important issue you mention), but when a proc sub-domain
> dynamically grows/shrinks on reneighboring steps. That seems different than ghost site exchange.
> I would call it site migration.

@sjplimp there should be no need for migrating these since this is like a grid applied to the subbox and because the processing happens in fractional coordinates, it is also immune to rescaling during fix deform or fix npt provided the rescale yes option is used. But this is also the reason why load balancing and comm style tiled are not supported, those would change how the sites are distributed.

Without rescale yes, things are a bit more complicated: small fluctuations are still tolerated similar to letting atoms move outside the box a little bit between neighbor list updates. For larger box changes the results are incorrect because the sites grid is not adjusted but stays at its original layout. It breaks down when the communication of the ghost sites become incomplete.

I will add a test for simulations where the box size or shape changes and print a warning if the `rescale` keyword is not set to `yes`.

akohlmey:
@sjplimp are there any open problems left that need an answer on this one?
I would really like to merge this ASAP, since I am using it in the tutorials that I am currently working on.

sjplimp:
@akohlmey Thanks for the updates - here are a few more issues.

(1) I think it would be better to not use compute_modify, but rather
add optional keyword/values to the command itself for drvac, drint,
region, rescale, site_file, etc.  And document them at the top of
compute_frenkel.rst.  Are there other computes which use
compute_modify for arguments unique to that compute?  I can't think of
any, and the compute_modify doc page doesn't say so.

(2) The units of drvac and drint are not clear to me.  The doc page
says both "distance units" and "lattice spacings" which are not the
same.  And it says the defaults of 1.01 and 1.42 are for 1st and 2nd
neighbors which I don't think is correct.  What LAMMPS calls a lattice
spacing is typically for a cubic cell which contains multiple
lattice sites.  So for bcc and fcc, the 1st neighbors are at sqrt(3)/2
and sqrt(2)/2 respectively.  Also, if the unit cell is not cubic
(e.g. hex or custom) then there are xyz lattice spacings, so which one
is this compute referring to?

So maybe you want drvac and drint to be unitless and be scale factors
on the 1st neighbor distance?  Or since 1st/2nd/3rd/etc neighbor
distances depend on the lattice style, maybe you want to have them be
required args so that the user thinks about what makes sense for their
lattice and uses distance units (e.g. Angstroms) or some more clearly
defined definition of a lattice spacing.

(3) The doc page says "region" is sometimes required.  I think you
mean "recommended", unless an error check is performed?  Likewise it
says a warning may be given when "rescale" is enabled.  Is an error
given when the box shape changes (shear) or when comm tiled is used?
I'm thinking it should be.

(4) Please explain for the user what a Wigner-Seitz cell is or that
Wigner-Seitz analysis simply means each atom is assigned to its
nearest lattice site.  How is that done efficiently, especially when
an arbitrary site_file is used?  E.g. is some kind of binning of
lattice sites performed, similar to what a neighbor list build would
use?  Ditto for the clustering analysis for a site_file lattice (see
next (5)).

(5) The definition of a "cluster" is unclear to me.  I don't think you
mean a partitioning of all the vac/int lattice sites into unique
clusters.  I think what you mean is that each vac/int site is treated
as the center of a cluster, and any nearby vac/int sites are part of
that cluster.  So for two nearby sites A,B -- the cluster at A could
contain B, and the cluster at B could also contain A.  Is that
correct?

If that's the case, I suggest you clarify that a "cluster" is simply a
way to label each lattice site based on its neighborhood, and that the
"center" of a cluster in the local dump file is simply the coords of
the lattice site at its center.  ("coords of cluster center" sounds
like an averaging over several nearby defective lattice sites)

akohlmey:
@hammondkd can you help me with any of the requested changes?

akohlmey:
@sjplimp I tried to address your concerns and add some 2d/3d figures for illustration of Wigner-Seitz cells and their relation to lattice sites plus a pointer to the more detailed explanation on Wikipedia. The code now uses unitless parameters for drvac and drint and modify_params() is gone. For good measure, there is also a section explaining the differences between this version and what is in the referenced publication.

doc/src/compute_frenkel.rst
```diff
@@ -0,0 +1,426 @@
...
+processor are sorted into a regular grid of bins, in the same way a
+neighbor list is built.  An atom is mapped to its bin, the closest site
+in that bin is picked, and the list of neighboring sites of that site,
+which is built from the same bins, is followed until no closer site is
+found.  Whether the atom is really inside the Wigner-Seitz cell of that
+site is then confirmed with the bisecting planes described above.  The
+bins and the lists of neighboring sites are set up once at the beginning
+of a run and rebuilt only when the simulation box changes.
+
+Nearby defects are further grouped into clusters.  Two defective sites
```
sjplimp: Nearby defective sites (0 or 2+ atoms) are further ...

doc/src/compute_frenkel.rst
```diff
@@ -0,0 +1,426 @@
...
+are connected when they are closer to each other than *drvac* if the
+second site is empty, or closer than *drint* if it holds more than one
+atom.  A cluster is a group of defective sites that are connected in this
```
sjplimp: This definition of a cluster is now clear to me (thanks), but it means a single cluster could have a long extent across multiple procs (e.g. along a slip plane or for a large vacany cluster).  Does the code figure this out - similar logic to compute cluster/atom or compute fragment/atom, looping over comm->forward ?

doc/src/compute_frenkel.rst
```diff
@@ -0,0 +1,426 @@
...
+from vacancy-type clusters, and a cluster with as many vacancies as
+interstitials has a size of zero.  A site holding more than two atoms
+counts as one interstitial, not as one per extra atom.  Note that the
+identification and counting of the vacancies and interstitials themselves
+depends only on the number of atoms at each site, not on this clustering
+or the two distances.
+
+The optional keywords listed above adjust the settings of the analysis.
+
+The *drvac* and *drint* distances are given as multiples of the nearest
```
sjplimp: are specified as unitless multiples of

doc/src/compute_frenkel.rst
```diff
@@ -0,0 +1,426 @@
...
+   compute ID group-ID frenkel keyword value ...
+
+* ID, group-ID are documented in :doc:`compute <compute>` command
+* frenkel = style name of this compute command
+* zero or more keyword/value pairs may be appended
+* keyword = *drvac* or *drint* or *region* or *rescale* or *site_file*
+
+  .. parsed-literal::
+
+       *drvac* value = distance for including vacancies in a cluster (nearest neighbor distances)
```
sjplimp: (unitless multiple of nearest neighbor distance)

doc/src/compute_frenkel.rst
```diff
@@ -0,0 +1,426 @@
...
+       *drint* value = distance for including interstitials in a cluster (nearest neighbor distances)
```
sjplimp: ditto

doc/src/compute_frenkel.rst
```diff
@@ -0,0 +1,426 @@
...
+neighbor distance of the reference lattice, not in distance units, so
+that the same setting has the same meaning for different lattices and
+different lattice constants.  They default to 1.5 and 1.82,
+respectively.  With those values a vacancy is connected to defects in
+the first two neighbor shells and a multiply occupied site to defects in
+the first three, and this is the case for simple cubic, bcc, and fcc
+lattices alike; the larger value for interstitials accounts for their
+spatially more extended configurations such as dumbbells.  The nearest
+neighbor distance and the resulting two distances are printed at the
+beginning of each run.  With *rescale yes* they follow the box, so that
```
sjplimp, line 142: Do you mean printed once when the compute is defined?  That would make more sense to me.  It doesn't change from run to run, correct ?  Or maybe this is intended to track "rescale yes" ?  Are there any other computes/fixes which print something at the start of each run ?

doc/src/compute_frenkel.rst
```diff
@@ -0,0 +1,426 @@
...
+rescaled reference lattice.
+
+Use the *region* keyword to exclude reference sites where no atoms are
+expected.  This is strongly recommended when parts of the simulation box
+are empty, for example the vacuum above a free surface in a non-periodic
+dimension, since otherwise every site in the empty space is counted as a
+vacancy.  Such a mistake is not detected as such, but a warning is
+printed when more than 20 percent of the reference sites are found to be
+empty, which also catches a reference lattice that does not match the
+crystal.
```
sjplimp: does not match the crystalline structure of the atoms.

doc/src/compute_frenkel.rst
```diff
@@ -0,0 +1,426 @@
...
+
+The reference sites never follow a change of the box *shape*, so shearing
+the box is not supported.  A warning is printed when the box shape will
+change during a run, and combining a shearing box with *rescale yes* is
+an error.
+
+With the *site_file* keyword the reference sites are not generated from
+the lattice but read from the given text file, which must contain one
+site per line as three coordinates "x y z" (anything following a "#"
+character is ignored).  This allows using a reference structure that is
```
sjplimp: Does this mean blank lines or comment lines starting with # are allowed ?

sjplimp:
@akohlmey Doc page is much more clear to me now - thanks for the updates.
Just added a few suggestions on wording and a clarification Q or 2.
The Q about large cluster sizes is not for the doc page, but rather to ensure the
code accounts for this possibility.
Will approve once the comments are addressed.

sjplimp:
Another conceptual Q I just thought of.  The doc page says the "owner" of a cluster is the proc which owns the averaged center point.  For a multi-proc spanning cluster is the center point average computed consistently on each proc?  Otherwise, a round-off error could result in two procs (or none) thinking they own the center point.

doc/src/compute_frenkel.rst
akohlmey: > Does this mean blank lines or comment lines starting with # are allowed ?

Both.

It re-uses the class that is also used for reading potential files which mimics Fortran formatted I/O with the added feature to truncate strings at the '#' character.

https://docs.lammps.org/latest/Developer_utils.html#_CPPv4N9LAMMPS_NS14TextFileReader9next_lineEi

doc/src/compute_frenkel.rst
akohlmey: This is done during `Compute::init()` and must be done there since the lattice may have changed since the compute command was issued.

> Are there any other computes/fixes which print something at the start of each run ?

Plenty. Some examples: compute fep, fix srd, fix msst, fix pour, fix sgcmc, fix pimd/\*, fix nvt/sllod, or fix polarize/\*

doc/src/compute_frenkel.rst
akohlmey: @sjplimp Yes, the code handles this with the same iterate-until-no-global-change logic as compute cluster/atom, but it cannot use `comm->forward_comm()` because the cluster members are lattice sites, not atoms.

In find_clusters() each defect site gets the site tag as initial cluster ID, when two sites belong to a cluster the lower ID is used. this is first iterated until no change locally, then with ghost atoms after `exchange_lattice_ghosts()`.  `push_ghost_labels_to_owners()` is the equivalent of the forward communication and an MPI_Allreduce() on "did any owned label decrease" is used to check for convergence.

A cluster spanning many procs converges in roughly as many passes as its extent in subdomains. Per-cluster size/center are then Allreduce sums, so no rank ever holds a whole cluster. An automated test run with on 1/2/4/8 ranks in six different grid shapes resulted in identical clusters.

akohlmey:
> Another conceptual Q I just thought of. The doc page says the "owner" of a cluster is the proc which owns the averaged center point. For a multi-proc spanning cluster is the center point average computed consistently on each proc? Otherwise, a round-off error could result in two procs (or none) thinking they own the center point.

@sjplimp Yes, consistent by construction — no rank ever computes its own version of the average. Each rank accumulates partial sums over only its owned sites and the global center is computed with `MPI_Allreduce()`, so cluster_center[] is bitwise identical on every rank. The compute does not allow balancing and comm_style tiled which would make this difficult.

doc/src/compute_frenkel.rst
```diff
@@ -0,0 +1,426 @@
...
```
sjplimp: ... following a "#" comment character is ignored).  Blank lines and comment-only lines are skipped.

sjplimp (changes requested):
Last comment to clarify site_file syntax.

doc/src/compute_frenkel.rst
akohlmey: @sjplimp done

sjplimp (approved):

PR #5030, opened 2026-06-11
Remove too conservative error check in dump file code

stanmoore1:
**Summary**

Remove too conservative error check in dump file code that prevented sorting and balancing dump files when `nfile` or `fileper` options were set.

**Related Issue(s)**

None

**Author(s)**

Stan Moore (SNL) and Github Copilot with Claude Sonnet 4.8

**Licensing**

I used Github Copilot with Claude Sonnet 4.8

**Backward Compatibility**

Yes

**Implementation Notes**

I tested this on 12 MPI ranks.


stanmoore1:
Description from Copilot:

`dump_modify sort` was unconditionally rejected when `nfile` or `fileper` set `multiproc > 1`. The restriction was overly conservative — the existing sort/balance logic already works correctly for the multi-file case.

**Changes**

- **`src/dump.cpp`** — Remove the 3-line error check in `Dump::init()` that blocked `sort` when `multiproc > 1`.
- **`doc/src/dump_modify.rst`** — Update the `sort` keyword description to explain that sorting works with `nfile`/`fileper`: atoms are sorted globally across all procs, then forwarded in order to each file-writer proc so each file piece contains a consecutive sorted range.

**Why it already works**

`sort()` distributes atoms to procs `0..nprocs-1` in sorted order via `Irregular` on `world`. Because `nfile`/`fileper` always assigns **consecutive** proc IDs to each cluster, every cluster inherits a consecutive sorted range. `write()` already pings cluster members in increasing `iproc` order, so data arrives at the file-writer in sorted order — no re-sorting at the file-writer is needed.

`balance()` (called after `sort()`) preserves global sorted order by using offset-based positioning, and distributes exactly `ntotal/nprocs` atoms per proc regardless of `nfile`.

**Backward Compatibility**

Previously `dump_modify sort id nfile N` was a hard error. It now works. Existing inputs that do not combine `sort` with `nfile`/`fileper` are unaffected.

stanmoore1:
The one artifact I'm seeing is that if `fileper` or `nfile` does not divide `nprocs` evenly, the balancing doesn't work perfectly since balancing is performed before accumulation, and some filewriter ranks accumulate from different numbers of ranks. But I think this is fine and I put a note in the docs.

doc/src/dump_modify.rst
```diff
@@ -751,6 +752,19 @@ snapshot are balanced to be nearly the same. A balance value of *no*
...
 run, this option is ignored since the output is already balanced.
+When multiple processors write separate dump file
+pieces via the "%" wildcard together with the *nfile* or *fileper*
+keywords, balancing is performed globally across all processors first
+and the balanced data are then forwarded to a smaller subset of
+processors who accumulate the data.
+
+.. note::
+
+   Because balancing is performed before data accumulation, *nfile* or
```
sjplimp: I would say this differently, something like

The balancing is performed on a per-processor basis, not per-file when using *nfile* or *fileper*.
If different numbers of processors contribute data to different dump files, the files will naturally be of different lengths.

sjplimp (changes requested):
small change to doc file

doc/src/dump_modify.rst
stanmoore1: Done, this is ready to merge from my POV.

akohlmey:
@sjplimp this is now waiting on you

sjplimp (approved):

PR #5022, opened 2026-06-05
GPU package: remove host/device particle-split load balancer

akohlmey:
## Summary

This pull request removes the legacy host/device particle-split load-balancing mechanism from the GPU package — a feature that was effectively disabled since CUDA 5/Fermi-era hardware and has carried dead weight in ~250 files ever since.

Specifically, the following changes are applied:

- Deprecate the `split` keyword in `package gpu` (warn, value validated but ignored); remove the `_particle_split` member from `FixGPU`
- Introduce `lib/gpu/include/lammps_gpu.h` as the single consolidated public C++ API between `src/GPU/` and `lib/gpu/`, replacing the previous, error-prone pattern where each `*_ext.cpp` file re-declared its own subset of functions
- Delete `lal_balance.h` (207-line `Balance<>` class) and all `hd_balancer` / `gpu_split` machinery from the 8 base classes (`base_atomic`, `base_charge`, `base_dipole`, `base_sph`, `base_three`, `base_dpd`, `base_ellipsoid`, `base_amoeba`) and from `lal_device`; simplify `init_nbor()` call sites to pass `host_nlocal=0, gpu_host=0`
- Remove `host_start` / `cpu_time` parameters from all 69 GPU pair-style `init()` / `compute()` / `compute_n()` function signatures in `lib/gpu/` (headers, ext files, implementations) and in `src/GPU/` (pair cpp call sites); remove the `cpu_compute()` host fallback entirely from the 53 pair styles that implemented one (the remaining 16 never had a host code path)
- Update `doc/src/package.rst` — mark `split` as deprecated with `.. versionchanged:: TBD`, remove examples, remove newton-pair restriction paragraph
- Move header from `src/GPU/` to `lib/gpu/include/`; add tagint prerequisite guard; remove leaked SARU RNG macros and duplicate EPSILON constants; add `extern` to all declarations; add style comments mapping abbreviated function prefixes to their LAMMPS pair-style names
- remove the unused `Answer::_newton` member and the uncalled `dpd_gpu_update_coeff()` / `dpd_coul_slater_long_gpu_update_coeff()` exports together with the thereby-unreachable `DPDCoulSlaterLong::update_coeff()` (`DPD::update_coeff()` stays; the *dpd/tstat* wrapper calls it)

Net change: 372 files, −9971 / +3660 lines. The `gpu` library loses ~6500 lines of CPU fallback code that was never exercised with `split 1.0` (the only supported value since GPU-force mode became default).

## Test plan

[...]

PR #5031, opened 2026-06-11
Refactor EAM pair style family to eliminate duplicated code

akohlmey:
**Summary**

Refactoring of the EAM pair style family (`eam`, `eam/alloy`, `eam/fs`, `eam/he`, `eam/cd`) and its accelerated variants in the OPT, OPENMP, INTEL, GPU, and KOKKOS packages. The potential file readers move into the `PairEAM` base class behind a file format selector and the `eam/he` embedding-table variation is folded into the base class compute path, so every format/engine combination becomes a constructor-only class. This removes about 5700 lines of duplicated code with no change in functionality or results, and prepares for adding accelerated `eam/he` variants in a follow-up pull request.

**Related Issue(s)**

None.

**Author(s)**

**Licensing**

The changes in this pull request were generated with the assistance of Anthropic's Claude Code, directed and reviewed by Axel Kohlmeyer. The refactoring is predominantly verbatim motion of existing, human-written code; the new glue code (file format dispatch, embedding-index helper, GPU function-pointer bindings, constructor-only subclasses) was AI-generated and human-reviewed.

**Backward Compatibility**

No changes to input syntax, file formats, or simulation results. For code deriving from these classes: `PairEAMAlloy`, `PairEAMFS`, and `PairEAMHE` remain (now constructor-only), their reader implementations moved to protected `PairEAM` methods (`read_setfl()`, `read_fs()`, `file2array_setfl()`, `file2array_fs()`), `read_file()`/`file2array()` stay virtual, and the virtual inheritance workaround for the OPT diamond is gone. External subclasses need a recompile and possibly trivial renames. One intentional behavior change (bug fix): `PairEAM::single()` and `compute_atomic_energy()` previously ignored the `rhomin` table offset of `eam/he` and returned wrong embedding contributions (e.g. through `compute group/group`, which changed from a nonsensical +1785 eV to -291.6 eV for a PdHHe test system); they now use the correct evaluation.

**Implementation Notes**

- `PairEAM::compute()`, `single()`, and `compute_atomic_energy()` share an inline `embedding_index()` helper; the `eam/he` variation (rhomin offset, two-sided clamping and linear extrapolation) is selected by `he_flag`, now a `PairEAM` member. `PairEAMHE::compute()`, a verbatim copy of `PairEAM::compute()` except for ~15 lines, is deleted.
- `PairEAM` gains a `fileformat` selector (FUNCFL/SETFL/FS); `coeff()`, `read_file()`, and `file2array()` dispatch on it. The setfl/fs readers moved verbatim from `PairEAMAlloy`/`PairEAMFS` into the base class; their two line-identical `coeff()` bodies are merged into one `coeff_mapped()`; repeated cleanup code is factored into `delete_setfl()`/`delete_fs()`.
- MANYBODY/OPT: `PairEAMAlloy`, `PairEAMFS`, `PairEAMHE` are constructor-only; virtual inheritance is removed; `eam/alloy/opt` and `eam/fs/opt` derive from `PairEAMOpt` alone (diamond gone); `eam/cd` no longer initializes a virtual base.
- OPENMP/INTEL: `eam/alloy/omp`, `eam/fs/omp`, `eam/alloy/intel`, `eam/fs/intel` drop their verbatim reader copies (about 320 lines each) and become constructor-only.
- GPU: `PairEAMGPU` now binds the per-style device-instance functions from the gpu library via function pointers set in the constructor; `eam/alloy/gpu` and `eam/fs/gpu` become constructor-only subclasses, removing their full copies of both the GPU wrapper and the readers.
- KOKKOS: `eam/alloy/kk` and `eam/fs/kk` were ~1500-line near-verbatim copies of `pair_eam_kokkos.cpp`; both are now constructor-only template subclasses of `PairEAMKokkos<DeviceType>`. The copies had drifted behind the base class (the KK_FLOAT mixed-precision modernization was never applied to them); they now inherit the maintained implementation, which is identical for default double-precision builds.
- Verification: thermo output is byte-identical before/after every commit for `eam`, `eam/alloy`, `eam/fs`, `eam/he`, `eam/cd` and with `-sf opt`, `-sf omp` (2 threads), `-sf intel` (mode double), `-sf gpu` (OpenCL), and `-k on t 1 -sf kk` (serial backend), each serial and with 4 MPI ranks, newton on and off, including `bench/in.eam` and both `examples/shear` inputs. The full ctest suite passes in all tested configurations.
- Pre-existing issue observed but NOT addressed here: the kk-suffix sub-tests of `test_pair_style` segfault in a host-only KOKKOS serial build, equally before and after this PR and also for untouched styles (`born/kk`, `sw/kk`, `tersoff/kk`, ...); to be investigated separately.
- Bug fix (last commit): the unit-conversion loops in `read_setfl()` skipped the last tabulated point of the frho and z2r tables (`j < nrho` / `k < nr` instead of `<=`), inconsistent with the funcfl and fs readers. The effect is limited to the table end points and stays within the tolerances of the `eam_alloy_real` and `eam_cd_real` unit tests, which pass unchanged.

**Performance**

Best-of-3 loop times (32000 atoms, 100 steps, `units metal`, AMD Ryzen desktop, GCC -O2). Classic styles produce byte-identical trajectories before/after, so the deltas are machine noise (interleaved re-runs scatter +-1.5% in both directions):

| style | serial develop | serial this PR | omp 4 threads develop | omp 4 threads this PR |
|---|---|---|---|---|
| eam (bench/in.eam) | 2.76 s | 2.78 s | 0.72 s | 0.74 s |
| eam/alloy (AlCu) | 3.89 s | 3.93 s | 0.98 s | 0.97 s |
| eam/fs (CuZr) | 10.76 s | 10.81 s | 2.66 s | 2.57 s |
| eam/he (PdHHe) | 6.98 s | 6.82 s | - | - |

**Post Submission Checklist**

A follow-up pull request based on this branch will add `eam/he/omp`, `eam/he/kk`, and `eam/he/gpu` variants, which reduce to constructor-only classes plus small embedding-kernel updates after this refactoring.


sjplimp:
Due to changes to compute(), I would like benchmarking on all the hardware we support (CPU, GPU, KOKKOS, etc) for a range of problem sizes to see if there is any perf impact.

akohlmey:
> Due to changes to compute(), I would like benchmarking on all the hardware we support (CPU, GPU, KOKKOS, etc) for a range of problem sizes to see if there is any perf impact.

@sjplimp Please let me complete this before adding any review comments. This is a draft and I have not yet flagged it ready for review or requested any reviews.

akohlmey:
@ndtrung81 This overlaps with PR #5022, please let me know which you would prefer to process first. I'll then set draft and ready-for-review status accordingly.

akohlmey:
@stanmoore1 Before we take this under serious review I would like to ask you to check if the general code layout change is acceptable. KOKKOS has different requirements and could not adopt the complex inheritance tree of OPT, GPU, and OPENMP. This means there is a larger potential that the coding agent messed things up at a fundamental level, especially on GPUs (I can currently only test on my AMD APU which is supported by KOKKOS but has restrictions and is slow).  Please also check that all single/mixed/double required changes have survived the refactor. Please assign it to @ndtrung81 afterwards for the same kind of check. Thanks. Ping me on slack if you have any questions or need clarifications.

ndtrung81:
@akohlmey please process this PR first. Thanks!

akohlmey:
> @akohlmey please process this PR first. Thanks!

OK, then you should wait until @stanmoore1 assigns this to you after he is done with his preliminary checks and I have implemented any requested changes.

ndtrung81:
@akohlmey thanks for refactoring the `eam/*/gpu` styles. The build with CUDA and OpenCL are successful on my end. The performance with these refactored styles remain the same as the original version.

akohlmey:
> Due to changes to compute(), I would like benchmarking on all the hardware we support (CPU, GPU, KOKKOS, etc) for a range of problem sizes to see if there is any perf impact.

There are no changes to compute() (in the base EAM class). This pull request merely removes redundant implementations resulting from the overly complex inheritance layout resulting from not having the various file format readers in the base class.

akohlmey:
> @akohlmey thanks for refactoring the `eam/*/gpu` styles. The build with CUDA and OpenCL are successful on my end. The performance with these refactored styles remain the same as the original version.

I didn't expect anything else.

ndtrung81 (approved):
The changes to the /gpu styles don't touch the computation on the device. I don't see changes in the performance on the GPU with `bench/in.eam` (32K atoms).

PR #5035, opened 2026-06-12
Improve force style tests for multi-precision KOKKOS support with GPU, Serial, and OpenMP backend

akohlmey:
**Summary**

This pull request extends the YAML-based force-style testers so they can validate KOKKOS builds in all host *and* GPU configurations and adds CI coverage for them, which previously did not exist at all:

- a new `kokkos_serial` test fixture that runs when the KOKKOS package is compiled with only the Serial backend; the existing `kokkos_omp` fixture now strictly requires the OpenMP backend instead of silently falling back to one thread
- a new `kokkos_gpu` test fixture (the complement of the two host fixtures) that runs the suite under a GPU back end of the KOKKOS package, together with a runtime probe `Info::has_kokkos_gpu_device()` (mirroring `Info::has_gpu_device()` of the GPU package) so the suite skips transparently on machines without a compatible GPU and can be run unchanged everywhere
- support for reduced precision KOKKOS builds (`-D KOKKOS_PREC=mixed/single`): a precision-dependent tolerance ladder (analogous to the GPU package fixture) and `skip_tests` entries that can be qualified by precision (e.g. `kokkos_serial_single`)
- a new CI workflow `unittest-kokkos.yml` that builds a {serial-only, OpenMP} x {double, mixed, single} matrix with all KOKKOS-capable packages and runs the force-style and c-library tests
- a new `cmake/presets/kokkos-packages.cmake` preset with all packages that contain KOKKOS styles, shared between the new workflow and the KOKKOS regression test workflow (which previously enabled a dozen packages without any KOKKOS styles)
- a `tags:`-keyword mechanism (`TestConfig::has_tag()`) so the fixtures special-case tests by a descriptive tag instead of by hard-coded style names

Running the previously untested configurations immediately exposed several genuine bugs in the KOKKOS package that are invisible at double precision because the legacy and Kokkos host views alias the same data there:

1. seven fix styles (aveforce, planeforce, drag, spring, lineforce, oneway, nve/noforce) declared shadowing `execution_space` members, so the integrator's sync/modified bracketing addressed the wrong views
2. six wall fixes and angle style dipole were missing the force/torque bit in `datamask_read` although their kernels read-modify-write those arrays
3. bond style quartic never marked its kernel forces modified before host-side bond-breaking post-processing
4. bond style gaussian had a dead underflow guard (`2.3e-308` cast to float is `0.0`) and accumulated its Gaussian sums in compute precision
5. pair style reaxff computed several logistic-form exponentials in compute precision, where they overflow to `inf` and produce `0*inf = NaN`, and its tabulation spline index could go out of bounds (segfault)
6. pair styles bondval and bondval/vec did not compile at all for non-double precision (wrong view types, plus an unrelated buffer-resize ordering bug)
7. fix acks2/reaxff/kk now errors out for non-double precision KOKKOS builds at runtime since its BiCGStab solver does not reliably converge there

The KOKKOS package code fixes are grouped into separate commits so they can be cherry-picked independently.

**Additional force-style test improvements**

The same branch also writes the YAML reference files as standard YAML (no non-specific `!` tag) so they can be validated against a JSON schema (`tools/json/force-style-test-schema.json`) with `check-jsonschema`, and adds an `input_coeffs:` key so the fix-timestep tester reads its force-field/group setup from a file instead of a hardcoded block.

It also adds force-style YAML tests for a number of previously untested fix styles (`gjf`, `press/langevin`, `gld`, `addtorque/group`, several simpler forcing/integrator fixes, and the aspherical `*/asphere` integrators) and the `resquared` pair style, introduces `no_respa` / `no_reset_dt` test tags, and relaxes the shared ellipsoid test configuration. Writing these surfaced and fixed a few latent restart bugs (RNG-state and related per-fix checkpointing) in `fix gjf`, `fix press/langevin`, and `fix gld`; see the individual commit messages for the detailed justification.

**Related Issue(s)**

#5037 (the `bond_style quartic/kk` device-only bug surfaced by the new `kokkos_gpu` fixture). That same fixture also surfaced several other device-only `/kk` bugs (in various `fix`, `kspace`, and `pair` styles) that are correct on host-only KOKKOS builds but wrong on a real GPU; these are currently skipped via `skip_tests: kokkos_gpu` and reported as additional findings in #5037.

**Author(s)**

**Licensing**

The code and documentation changes in this pull request were generated with Claude Code (Anthropic Claude model) operating under the direction of and with review, testing, and corrections by Axel Kohlmeyer.

**Backward Compatibility**

No changes to user inputs. One behavior change: fix acks2/reaxff/kk now stops with an error when LAMMPS was compiled with `-D KOKKOS_PREC=mixed` or `single` instead of producing unconverged charges silently. The GPU package device-teardown deferral is test-only and a no-op in production (and a no-op when the GPU package is not built).

**Implementation Notes**

- The six force-style test drivers share the test body of the KOKKOS fixtures through a file-local `run_kokkos_test()` helper; the precision is queried at runtime via `Info::has_accelerator_feature("KOKKOS", "precision", ...)`. Exactly one of `kokkos_serial` / `kokkos_omp` / `kokkos_gpu` runs per build, selected by the active back end.
- `Info::has_kokkos_gpu_device()` is backed by `lmp_has_compatible_kokkos_gpu()` in the KOKKOS package (`cudaGetDeviceCount` / `hipGetDeviceCount` / SYCL device enumeration). Those vendor device-count queries return an error code instead of aborting, so they are safe to call *before* `Kokkos::initialize()`, which aborts on a GPU build with no usable device.
- The `kokkos_gpu` fixture runs with `-pk kokkos neigh half` so the device kernels execute with the input templates' default `newton on`. The GPU default `neigh full` requires newton off (`KokkosLMP::newton_check()`), which the force-style templates do not use; KOKKOS provides `HALF`/`HALFTHREAD` newton-on kernels for essentially all styles (including manybody, and SNAP which *requires* newton on), so no per-style newton plumbing is needed.
- GPU package <-> KOKKOS coexistence: when both packages can use a GPU in one test process, the GPU package's per-instance teardown (`~FixGPU` -> `lmp_clear_device()` -> `delete gpu` -> `UCL_Device::clear()` -> `hipDeviceReset()`) resets the whole device and invalidates the KOKKOS device context, crashing `Kokkos::finalize()`. A generic, test-only switch (`lmp_gpu_defer_device_clear()` in lib/gpu, wrapped by `Info::gpu_defer_device_clear()`) defers that reset; `test_main` enables it when both `has_package("GPU")` and `has_kokkos_gpu_device()` hold, so the `gpu` and `kokkos_gpu` fixtures run together.
- Tag-based special-casing: `TestConfig::has_tag()` checks the YAML `tags:` list. The fixtures use `single_thread` (dpd/snap/pace, which cannot run multi-threaded in the test) and `spica_pair` (the angle setup that needs `pair_style lj/spica`) instead of `strmatch` on style names.
- Tolerance ladder: mixed = 2.0e9 x epsilon, single = 1.0e10 x epsilon on top of the existing KOKKOS relaxation. The mixed factor was tuned over the complete suite (the GPU package's 5.0e8 was up to a factor 2.5 too tight for five tests).
- Precision-qualified `skip_tests` entries were added for tests that cannot pass in reduced precision for fundamental reasons (Gaussian tail underflow, Buckingham exp/r^6 cancellation, ML-PACE, the acks2 solver, and fix efield global force totals that are cancellation sums of a neutral system).
- ML-IAP requires `KOKKOS_PREC=double` (enforced by CMake), so it is excluded from the shared preset and enabled explicitly in the double precision cells of the workflow.
- Validation: the full force-style suite passes 576/576 in all six local host configurations; the new workflow was exercised on a fork three times (the first two runs caught a c-library test using `EXPECT_DOUBLE_EQ` on reduced-precision results and a missing skip entry) and the final run is green in all six matrix cells. The `kokkos_gpu` fixture was validated on a real GPU (HIP, double precision): pair `lj_cut` family 26/26, manybody `tersoff` 9/9, with the `gpu` and `kokkos_gpu` fixtures running cleanly in the same process.
- The TransformView concurrent-modification guard was the key debugging aid for the view-bookkeeping bugs; a useful gdb note for reproducing the reaxff NaNs: `call feenableexcept(...)` does not work under gdb (it restores the FP control registers after inferior calls), use `set $mxcsr = $mxcsr & ~0x680` instead.

**Post Submission Checklist**


akohlmey:
@athomps @stanmoore1 with the changes in this pull request we should now be prepared to run automated tests on GPUs now for both, KOKKOS and the GPU package.

akohlmey:
@stanmoore1 @ndtrung81 If you don't have any major concerns, can we work on getting this PR merged next. Since it expands automated KOKKOS testing and enables co-existence of KOKKO/GPU and GPU package testing, it would be a big help to make certain that any of the also pending accelerator porting and refactoring changes have no unwanted side effects.

stanmoore1 (approved):
This looks really great, thanks @akohlmey for implementing it!

ndtrung81:
@akohlmey I don't have any concern with this PR. The addition of `lmp_gpu_defer_device_clear()` for coexistence of GPU and KOKKOS packages only in the tests makes sense to me. The unittest yaml file with the 3 precision modes for the KOKKOS package build looks good to me.

ndtrung81 (approved):

PR #5043, opened 2026-06-15
Fix neighbor list circular copy dependency and re-enable trimming  (issue #4529)

### LINKED ISSUE #4529: [BUG] Neighbor list logic issue can lead to circular dependencies
**Summary**

 Neighbor list logic issue can lead to circular dependencies. Neighbor lists are copies of each other, and sometimes an error is generated. This was seen in a test using DPD-REACT.

Latest develop

**Expected Behavior**

Successful initialization of LAMMPS neighbor lists

**Actual Behavior**

`ERROR: Failed to reorder neighbor lists to satisfy constraints - Contact the LAMMPS developers for assistance (../neighbor.cpp:1146)`

Tweaking the order of the lists in the reproducer below gives lists that are all copied from each other:

```
5 neighbor lists, perpetual/occasional/extra = 5 0 0
  (1) pair lj/cut, perpetual, copy from (2)
      attributes: half, newton on, cut 6.3
      pair build: copy
      stencil: none
      bin: none
  (2) pair lj/cut, perpetual, copy from (3)
      attributes: half, newton on, cut 6.3
      pair build: copy
      stencil: none
      bin: none
  (3) pair lj/cut, perpetual, copy from (2)
      attributes: half, newton on, cut 6.3
      pair build: copy
      stencil: none
      bin: none
  (4) pair lj/cut, perpetual, copy from (1)
      attributes: half, newton on
      pair build: copy
      stencil: none
      bin: none
  (5) pair lj/cut, perpetual, copy from (1)
      attributes: half, newton on
      pair build: copy
      stencil: none
      bin: none
```

**Steps to Reproduce**

```
boundary        p p p
units           lj
atom_style      atomic

lattice         fcc 1.0
region          box block 0 10.0 0.0 10.0 0.0 10.0 units lattice
create_box      1 box
create_atoms    1 box

mass 1 1.0

pair_style      hybrid/overlay lj/cut 5.0 lj/cut 5.0 lj/cut 6.0 lj/cut 6.0 lj/cut 6.0
pair_coeff      1 1 lj/cut 1 1.0 1.0 5.0
pair_coeff      1 1 lj/cut 2 1.0 1.0 5.0
pair_coeff      1 1 lj/cut 3 1.0 1.0 6.0
pair_coeff      1 1 lj/cut 4 1.0 1.0 6.0
pair_coeff      1 1 lj/cut 5 1.0 1.0 6.0

run             0
```

N/A

stanmoore1:
**Summary**

Fix neighbor list circular copy dependency and re-enable trimming  (issue #4529)

**Related Issue(s)**

Fixes #4529 

**Author(s)**

**Licensing**

I used Claude Code (Opus 4.8)

**Backward Compatibility**

Yes

**Implementation Notes**

Description from Claude:

The neighbor list "morph" logic could create circular copy
dependencies (e.g. list 2 copies from 3 while 3 copies from 2),
causing initialization to fail with "Failed to reorder neighbor
lists to satisfy constraints" (issue #4529). The same logic also
disabled neighbor list trimming and mishandled custom-cutoff
occasional lists (issue #4437).

In morph_unique(), a list with a custom cutoff is now flagged as
unique when its cutoff differs from the master cutoff (cutneighmax)
or it is occasional, rather than comparing against cutneighmin.
Lists whose cutoff equals the master cutoff are no longer forced to
be unique, so they become plain copies of the master instead of
forming circular copy chains, while shorter-cutoff lists keep their
cutoff and are trimmed from the master.

In morph_copy_trim(), the "cutoffs too small" guard that blocked
copy/trim is now restricted to occasional lists, re-enabling
trimming of perpetual lists with shorter cutoffs.

Re-enable the previously disabled one_trim_half_list_newton and
one_trim_half_list_nonewton unit tests, which now pass since trim
detection works again.

Verified:
- #4529 hybrid/overlay 5x lj/cut reproducer initializes (no error)
- #4437 compute rdf occasional lists get correct cutoffs, g(r) full
- trimmed sublist energy matches the equivalent standalone list
- full test_neighbor_class suite: 40 passed, 0 failed


stanmoore1:
Comment from Claude:

### Why the `morph_copy_trim()` guard compares against `cutneighmin_pair` rather than `cutneighmin`

The default master list (`cut=0`) is built with per-type cutoffs `cutneighsq[i][j]`, where each entry is the **max over all pair-style cutoffs** for that type pair. So pair-style requests always raise the master to cover themselves, but `fix`/`compute` requests do not. A custom-cutoff list can therefore only safely copy/trim from the master if its cutoff doesn't exceed the smallest per-type cutoff *actually present in the master* — that value is `cutneighmin_pair` (captured before the custom request cutoffs are folded into `cutneighmin`).

**Concrete example — single atom type, three perpetual lists with cutoffs 5, 6, 7:**

- The master is built to 7 (it contains all pairs out to 7).
- `cutneighmin_pair = 7` (smallest per-type-pair cutoff of the master).
- `cutneighmin = 5` (dragged down by the cutoff-5 request at neighbor.cpp:386–389).
- The **cutoff-6** list:
  - With `cutneighmin`: `6 > 5` → guard fires → it does **not** trim; it builds its own 6.0 list from scratch (redundant build + memory).
  - With `cutneighmin_pair`: `6 > 7` → false → guard skipped → it **trims** from the 7.0 master, which already contains everything it needs.

Both choices are **correct** (neither drops pairs); the difference is purely efficiency — `cutneighmin_pair` lets a middle-tier perpetual list reuse the master via a cheap trim instead of doing a redundant full build.

The reason this matters versus keying the guard on `irq->occasional`: that flag is unrelated to whether the master actually contains the needed pairs. Gating on `occasional` leaves perpetual custom-cutoff lists from fixes/computes (e.g. `group ... dynamic ... within`) able to trim from a per-type-truncated master and silently drop pairs, whereas comparing against `cutneighmin_pair` is correct for both perpetual and occasional lists without depending on that flag.

src/neighbor.cpp
```diff
@@ -1208,7 +1216,7 @@ void Neighbor::morph_unique()
       if (!irq->occasional)
         irq->cutoff += skin;
```
jtclemm, line 1219: I understand the logic of separately comparing the min/max of all requests and just the pair requests, but why also have a occasional check? Does a particular example still fail w/o it? 

Since occasional don't add a skin, I'd guess this can only be triggered coincidentally? But it'd also have to match machine precision errors to equate?

The nomenclature also seems misleading since now cutneighmin_pair and cutneighmax both refer to a pairstyle-only value. But cutneighmin is for the pairstyles + requests. May want to rename?

stanmoore1: Reply with help from Claude:

Thanks @jtclemm — these were good catches. Both addressed; details below.

**(1) Why the `|| irq->occasional` check, and does anything fail without it?**

It's required — the #4437 reproducer (`compute rdf ... cutoff 6.0`) fails without it. `compute_rdf` deliberately requests `cutoff = cutoff_user + skin`:

```c
// compute_rdf.cpp
mycutneigh = cutoff_user + skin;
// skin is included b/c Neighbor uses this value similar
//   to its cutneighmax = force cutoff + skin
req->set_cutoff(mycutneigh);
```

so the occasional cutoff is intentionally in the same units as `cutneighmax`. When the user's rdf cutoff equals the largest pair cutoff (a normal choice), `mycutneigh == cutneighmax` **exactly** — same arithmetic, not a floating-point coincidence. That's precisely #4437 (`6.0 + 2.0 == 8.0`).

I confirmed it: dropping the clause and rerunning #4437, the cutoff-6.0 rdf loses its cut flag and copies the per-type master:

```
(2) compute rdf, occasional, copy from (1)
```

Since the master is built only to each pair's per-type cutoff (`cutneighsq`), the result is truncated — g(r) decays to ~0 past ~5.4 Å and the coordination number flattens at 61.7 instead of 75. fix/compute request cutoffs don't feed `cutneighsq`, so keeping such lists `unique` is the safe behavior. I've added a short comment at that line explaining this.

**(2) "Occasional don't add skin, so equality is only coincidental / machine precision":**

That's the key subtlety — for these lists the skin *is* present (added by the requester, e.g. `compute_rdf` above), so the equality with `cutneighmax` is exact and deterministic rather than precision-dependent. The neighbor-list code itself only skips adding skin for occasional lists because the requester has already included it.

**(3) Naming:**

Agreed, and refactored. The members now have consistent, scope-explicit names:

```c
double cutneighmin;     // min neighbor cutoff over type pairs (pair styles) — parallels cutneighmax
double cutneighminall;  // min neighbor cutoff over type pairs AND custom list request cutoffs
double cutneighmax;     // max neighbor cutoff over type pairs (pair styles)
```

`cutneighmin`/`cutneighmax` are now a matched pairstyle-only pair, and the request-inclusive value is `cutneighminall`. All former `cutneighmin` uses that included request cutoffs (library `neigh_cutmin`, `delete_atoms`/`create_bonds`/`dump_image` warnings, `NBin` bin sizing) were routed to `cutneighminall`, so behavior is unchanged. The copy/trim guard now reads against `cutneighmin` (the smallest cutoff guaranteed present in the master).

stanmoore1: @jtclemm Alternatively we could rename to `cutneighmin_pair`, `cutneighmin` as the "all", and `cutneighmax`. This would require less changes to the code. Let me know what you think.

jtclemm: Okay, I think I understand what the AI is doing now. While I agree it's technically correct, I still think it's conceptually wrong. In my opinion (Steve or others may disagree), ultimately the fact that a list is occasional should not imply anything about how the cutoff is interpreted or handled with trimming. I think this solution is starting down a path that risks creating an incomprehensible web of logic where some request variables have hidden connotations.

Alternatively, I mocked up a different solution in this branch:
https://github.com/jtclemm/lammps/tree/nrequest-clarify-cutoffs
I believe this works in similar manner to this branch, but it accomplishes the goal by creating a new nrequest variable `cut_fixed` which clarifies how a custom cutoff is interpreted. Either:
a) `cut_fixed = 1` => there is a fixed cutoff applied to all atom types (needed for something like RDF where the analysis extends a fixed distance)
b) `cut_fixed = 0` => the cutoff is a maximum cutoff for some atom type (typical of pair styles like hybrid/kim)

Is it easy to ask Claude to confirm whether or not this works too? It probably is testing a wider set of scenarios than I did.

Although I didn't do it in that branch, I think adding `cut_fixed` could also let us clarify the logic of trimming lists and how to set cutoffs in nbin.cpp, nstencil.cpp, and npair.cpp to further help identify where LAMMPS can trim/copy lists. Maybe Claude could do this automatically?

Happy to be proven wrong though!

stanmoore1: @jtclemm this is a good alternative, working on it...

jtclemm: Cool, glad it might be useful. Apologies if it means extra work for you though!

sjplimp:
@stanmoore1 @jtclemm I looked over this PR and comments - it would be nice to get into the stable release since it is fixing a bug.  From your perspective is it ready to go or are there still issues to decide?  

Re: Joel's comments about occasional lists, I agree that there should be no implied restrictions on occasional lists, e.g. their cutoff.  Also these points:

(a) the most important task for the Neighbor class (relevant to this PR) is to determine the optimal set of perpetual lists to build to support the perpetual requests - optimal means the fewest and fastest set of lists - sometimes new requests for perpetual lists are generated to support this goal

(b) occasional lists should have no influence on (a) - it is a bonus if an occasional list can be derived (whenever it is needed) from a perpetual list, but the fact that the occasional list is needed (potentially very rarely) should not influence (a)

(c) the key divider here is perpetual vs occasional, not pair vs non-pair requests - there are fix styles which request perpetual lists

jtclemm:
Thanks for the input @sjplimp. To make sure we're all on the same page, I just want to clarify/repeat one detail.

> (c) the key divider here is perpetual vs occasional, not pair vs non-pair requests - there are fix styles which request perpetual lists

I do not think anyone is saying pair vs. non-pair requests should be treated differently (and maybe you were just generally stating this). 

What should be treated differently is what pair styles *typically* assume for neighbor lists and what non-pair styles *typically* assume for neighbor lists. Namely, whether `request->cutoff` applies to all atom types (say in compute RDF which wants everything included to a fixed distance) vs. whether `request->cutoff` is the maximum for some atom type (say in a pair style which may have different cutoffs for different types).

I think the solution to this circular logic is to clarify these two interpretations of `request->cutoff`. This isn't a hard distinction between pair and non-pair requests, but the pair vs. non-pair distinction is just helpful for explaining the difference between the two interpretations.

stanmoore1:
@sjplimp @jtclemm I think your feedback is addressed in https://github.com/lammps/lammps/pull/5154. Can you take a look? I'm going to close this PR in favor of https://github.com/lammps/lammps/pull/5154.

PR #5044, opened 2026-06-17
Fix store state history array

jrgissing:
**Summary**

- [x] add support for full arrays to 'fix store/state' history keyword
- [x] add wildcard support to history keyword of 'compute property/atom'
- [x] simplify shock/nemd example
  - [ ] could keep improving to the point of easily increasing resolution of output
- [x] docs

**Related Issue(s)**

none

**Author(s)**

Jake Gissinger (Stevens Institute of Technology)

**Licensing**

No, was not necessary.

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**


jrgissing:
@sjplimp I added array support for 'fix store/state'. One potential improvement is to allow the user to find out how many columns the expanded input has. Not necessary for my purposes, but I imagine this issue could come up in an input script. For example, a output vector of length Ninputs could store how many columns each user input was expanded to.

@athomps I also added the history[*][J] feature you requested, and starting simplifying your shock/nemd example. Currently, it reproduces your cvcf.dat exactly, but I a few more small modifications could allow increasing the resolution of that output. Please let me know if that would be useful.

doc/src/compute_property_atom.rst
```diff
@@ -76,6 +77,7 @@ Syntax
            *d_name* = custom floating point vector with name
            *i2_name[I]* = Ith column of custom integer array with name
            *d2_name[I]* = Ith column of custom floating-point array with name
+           *history[\*][J]* = all history frames (Nrepeat frames) for Jth attribute (1 to Nattribute)
```
sjplimp: The way other args which allow wildcards are documented, is to state here that "index I can include a wildcard (see below)" - e.g. see the list of allowed args for the dump custom command.  Then the explanation of how to use the wildcard is in the text.  One thing to decide is if you are going to support different forms of wildcard usage, e.g.:  * or 2* or *4 or 2*4, which most commands do.  Maybe it is less applicable in this use case.

doc/src/compute_property_atom.rst
```diff
@@ -33,7 +33,8 @@ Syntax
                              corner1x, corner1y, corner1z,
                              corner2x, corner2y, corner2z,
                              corner3x, corner3y, corner3z,
-                             i_name, d_name, i2_name[I], d2_name[I], history[I][J],
+                             i_name, d_name, i2_name[I], d2_name[I],
+                             history[\*][J], history[I][J],
```
sjplimp: see next comment, don't list the asterisk form here

sjplimp (changes requested):
@jrgissing - this PR looks good - just asking for doc page changes

@akohlmey you could look at the code changes to the two *.cpp files to support the wildcard usage- I don't know if there is a better way to do this

sjplimp:
@jrgissing one more question - what does this line in the Summary above mean?
>  add support for full arrays to 'fix store/state' history keyword

jrgissing:
@sjplimp can you taking a look at 8fa28ff8232e0489bc9f0a3c3f20072c1ca37671? Seems to be an indexing bug that fixes some issues for me. Not sure why it's not causing any issues with @athomps's example script

doc/src/compute_property_atom.rst
jrgissing: thanks for catching this. currently, in this PR, just a single asterisk is supported for compute property/atom's 'history', but the more complicated wildcards are supported within 'fix store/state' values due to use of utils::expand_args

jrgissing:
> @jrgissing one more question - what does this line in the Summary above mean?
> 
> > add support for full arrays to 'fix store/state' history keyword

this was meant to refer to wildcard support

sjplimp:
>@sjplimp can you taking a look at https://github.com/lammps/lammps/>commit/8fa28ff8232e0489bc9f0a3c3f20072c1ca37671? Seems to be an indexing bug that fixes some issues for me. Not >sure why it's not causing any issues with @athomps's example script

yes, I think that change is correct

sjplimp (approved):
Looks like you addressed all my comments - thanks

akohlmey:
> @jrgissing - this PR looks good - just asking for doc page changes
> 
> @akohlmey you could look at the code changes to the two *.cpp files to support the wildcard usage- I don't know if there is a better way to do this

@sjplimp this looks like it is using the canonical way of handling this.
Unfortunately, given the complexity of the command itself, only detailed testing can tell if there are any omissions, but that is not yet implemented.

PR #5045, opened 2026-06-18
BPM-based peridynamics: pair_style/bond_style bpm/peri (PERI models on the BPM framework)

akohlmey:
**Summary**

This pull request re-implements the four constitutive models of the PERI
(peridynamics) package -- prototype microelastic brittle (PMB), state-based
linear peridynamic solid (LPS), viscoelastic solid (VES), and elastic-plastic
solid (EPS) -- inside the BPM (bonded particle model) framework, as a companion
`pair_style bpm/peri` (short-range contact) and `bond_style bpm/peri` (the
constitutive law, model chosen by the first `bond_coeff` keyword), plus a
`compute bpm/peri/damage/atom` diagnostic.

Peridynamic bonds become real LAMMPS bonds, so the implementation uses the
existing `atom_style bond` (no new atom style), stores per-bond reference state
and history in restart files via `FixBondHistory`, breaks bonds individually,
and reuses BPM tooling -- while reproducing the legacy peridynamic physics.  On
a periodic 16^3 block the new styles are 3.6x-5.5x faster than the corresponding
`peri/*` pair styles (see `examples/bpm/peri/benchmark`).

The branch also carries `bond_style bpm/zero` (a force-free "zero" BPM bond
style added to support this work and mirror the other zero styles, used for
testing and debugging the BPM machinery); this pull request adds its previously
missing documentation.

**Related Issue(s)**

None.

**Author(s)**

Implementation by Claude Opus 4.8 (Anthropic) under the direction of Joel
Clemmer (SNL) and Axel Kohlmeyer (Temple University), derived from the PERI
package by Mike Parks (SNL; PMB/LPS) and Rezwanur Rahman and J. T. Foster
(UTSA; VES/EPS).  Corresponding author: Axel Kohlmeyer, akohlmey@gmail.com.

**Licensing**

By submitting this pull request, I agree, that my contribution will be included
in LAMMPS and redistributed under either the GNU General Public License version
2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Artificial Intelligence (AI) Tools Usage**

The new code in this pull request -- `bond_style bpm/peri`, `pair_style
bpm/peri`, `compute bpm/peri/damage/atom`, their documentation, the example
input decks, the `unittest/bpm` test harness, and the documentation page added
for `bond_style bpm/zero` -- was generated by Claude Opus 4.8 (Anthropic) under
the direction of Joel Clemmer and Axel Kohlmeyer, derived from the existing PERI
[...]

PR #5046, opened 2026-06-18
PERI: fix peri/eps yield norm and plastic return mapping

akohlmey:
**Summary**

Two latent bugs in the `peri/eps` (state-based elastic-plastic) peridynamic
model are corrected.

1. **Yield norm.**  `compute_DeviatoricForceStateNorm()` built the trial
   deviatoric force state with an extra dilatation factor `theta`
   (`omega*theta/wvolume`), but the force loop's deviatoric force state
   (`omega/wvolume`, line ~306) has no such factor.  The yield surface was
   therefore tested against a different, `theta`-inflated force state, so the
   material yielded much later than intended.

2. **Plastic extension update.**  The per-bond plastic deviatoric extension was
   updated as `edpNp1 + rkNew*deltalambda`, which reduces (uniform case) to a
   gain of `~2/r0` on `(e_dev - edp)` and overshoots/diverges under sustained
   plastic flow.  It now uses the dimensionally consistent radial-return update
   `edpNp1 + (deviatoric_extension - edpNp1)*(1 - sqrt(2*psi)/||t_d||)`, scaling
   the elastic part by the same radial factor applied to the force state.

**Related Issue(s)**

Found while re-implementing the PERI models on the BPM framework (see the
companion `bond_style bpm/peri` pull request #5045), where the corrected forms
are used.

**Author(s)**

Found and fixed by Claude Opus 4.8 (Anthropic) under the direction of Joel
Clemmer (SNL) and Axel Kohlmeyer (Temple University).  Corresponding author:
Axel Kohlmeyer, akohlmey@gmail.com.

**Licensing**

The analysis and the two code corrections in this pull request were produced by
Claude Opus 4.8 (Anthropic) under the direction of Joel Clemmer and Axel
Kohlmeyer, and reviewed by the directing authors.

**Backward Compatibility**

Results for `peri/eps` runs that reach the plastic regime will differ from
earlier versions (the model now yields at the intended stress and is stable);
purely elastic results are unaffected.  This is documented with a
`versionchanged` note in `doc/src/pair_peri.rst`.

**Implementation Notes**

Verified with an elastic-perfectly-plastic uniaxial test (incremental loading of
a periodic block): before the fix the deviatoric stress diverged/oscillated once
the plastic regime was entered; after the fix it yields at the intended stress
and plateaus stably.  The existing `PairPeri` unit test (`unittest/force-styles/
test_pair_peri.cpp`) still passes.

**Post Submission Checklist**


src/PERI/pair_peri_eps.cpp
```diff
@@ -310,8 +310,16 @@ void PairPeriEPS::compute(int eflag, int vflag)
...
+        // radial return of the deviatoric force state onto the yield surface,
+        // and the consistent update of the plastic deviatoric extension: the
+        // elastic part (deviatoric_extension - edpNp1) is scaled by the same
+        // radial factor.  The previous update edpNp1 + rkNew*deltalambda is
+        // dimensionally inconsistent (a ~2/r0 gain) and overshoots/diverges
+        // under sustained plastic flow.
+        double radial = sqrt(2.0*pointwiseYieldvalue) / tdnorm;
+        rkNew = radial * tdtrialValue;
+        deviatorPlasticExtTemp[i][jj] =
+          edpNp1 + (deviatoric_extension - edpNp1) * (1.0 - radial);
```
jtclemm: For what it's worth, I believe this is Eq. 45 in "A Nonlocal, Ordinary, State-Based Plasticity Model for Peridynamics" by J. A. Mitchell. Need to map out variables in code to math before I can parse the difference.

jtclemm: So Rk Is the expression for t^d_(n+1) in Eqn. 45. 
deltalambda is Eqn. 44. 
and edpNp1 is now e^dp_n

So the old expression matches the text. So if Claude found something real, I believe it's a limitation of the original formulation not in the code.

akohlmey: Is there somebody you know that we can consult about this?

jtclemm: I can see, but I think the AI is either wrong or it's fixing a unit error in the wrong place. Eq. 45 is:
image
Where psi_0 (`pointwiseYieldvalue`) has units of t^2 (`tdnorm` and `tdtrialvalue`) and t has units of alpha. The `deltalambda` variable has units of 1/alpha:
image


So if the dimensions are wrong, it must be where these factors are defined above. If you push back does it clarify?

src/PERI/pair_peri_eps.cpp
```diff
@@ -561,9 +569,12 @@ double PairPeriEPS::compute_DeviatoricForceStateNorm(int i)
...
       double omega_minus = influence_function(delx0,dely0,delz0);
 
+      // the yield-norm trial force state must match the deviatoric force state
+      // the force loop actually applies (line ~306, with no dilatation factor);
+      // the previous (omega*theta/wvolume) form carried a spurious extra theta
       tdtrial = ( 15 * shearmodulus[itype][itype]) *
-           ((omega_plus * theta[i] / wvolume[i]) +
-             ( omega_minus * theta[j] / wvolume[j] ) ) * (ed - edPNP1);
+           ((omega_plus / wvolume[i]) +
+             ( omega_minus / wvolume[j] ) ) * (ed - edPNP1);
```
jtclemm, line 581: This is equation 40:

Image
alpha is the 15 * the modulus (mu). Have to track down what the underscored omega means.

jtclemm: I think this change is right. But the paper is vague about the relevant variables like tdnorm (from my nonexpert reading) so I couldn't say for certain.

akohlmey: It was found and confirmed empirically. Without the bugfix simulations would produce unstable or unexpected results. The best I can do is to request an input deck that demonstrates the behavior in question that can then be run against the code before and after applying the changes in this branch.

jtclemm: I"m tempted to say we accept this change but revert the other one. 

I believe the other fix is inventing a new yield surface criterion. Whether its better or not I could not say. But I think it makes sense to stick to the original implementation if it matches with the intended reference.

akohlmey: > I'm tempted to say we accept this change but revert the other one.

It is your call. You can edit the pull request or give me specific instructions I should pass to the coding agent. I still think that having a test case and observing what the change tries to correct and then experimenting with it and deciding whether that is an important choice/property is probably the safest route. But I don't know enough about the physics to make a choice. I only have to point out that the same change is in PR #5045 and we have to ask to revert the parts that need to be reverted to make the legacy style and the new implementation consistent.

akohlmey:
@jtclemm I asked the coding agent to look at the discussion and summarize its arguments why both changes are needed (which it insisted). So here is the detailed response.

---

## The two corrections fix two *different* defects, and both are needed

Short version: the yield-norm change (line ~577) fixes *where* the material
yields; the plastic-update change (line ~322) fixes *whether the integration is
stable at all* once it yields. They are independent defects, so "accept one and
revert the other" leaves a model that is still wrong -- and, as shown below, still
numerically divergent. I include a small deck that demonstrates this directly.

### Notation (code <-> Mitchell 2011)

For one bond of node `i` with partner `j`:

```
e_d        = deviatoric_extension = dr - theta_i*r0/3        (deviatoric extension, a length)
edp        = deviatorPlasticextension[i][jj]                 (plastic deviatoric extension, internal variable, a length)
K          = 15*mu*(omega_i/m_i + omega_j/m_j)               (per-bond geometric factor; omega = 1/r0, m = wvolume)
t_d_trial  = tdtrialValue = K*(e_d - edp)                    (trial deviatoric force state of the bond)
||t_d||    = tdnorm = sqrt( sum_bonds (t_d_trial)^2 * V )    (per-node force-state norm)
psi0       = pointwiseYieldvalue = 25*sigma_y^2/(8*pi*delta^5)
yield:       fsurf = (1/2)||t_d||^2 - psi0 > 0
radial     = sqrt(2*psi0)/||t_d||   in (0,1]  while yielding
deltalambda= (||t_d||/sqrt(2*psi0) - 1) * m_i/(15*mu)        (Eq. 44)
```

The projected (returned) force state actually used for the force this step is the
**same in both versions**:

```
rkNew = radial * t_d_trial          (old: sqrt(2*psi0)*tdtrial/tdnorm ; new: radial*tdtrial -- identical)
```

So the change at line 322 does **not** change the yield surface and does **not**
change the force applied on the current step. It changes only the value stored in
the internal plastic variable `edp` that is carried to the *next* step. That is why
the difference is invisible at `run 0` / single-step tests and only appears under
sustained plastic loading.

---

## Correction 1 -- yield norm (line ~577): the deviatoric force state must not carry the dilatation

The deviatoric force state is, by construction, the part of the response with the
dilatation removed: the deviatoric extension itself is `e_d = dr - theta*r0/3`,
i.e. `theta` has already been subtracted. Accordingly, the **force loop**
(`tdtrialValue`, ~line 306) and the **deviatoric energy** (~line 296/eflag block)
both build the force state as `15*mu*(omega/m)*(e_d - edp)` with **no** `theta`
factor. Only `compute_DeviatoricForceStateNorm()` carried an extra
`omega*theta/wvolume` -- so the yield check was comparing `psi0` against a
`theta`-inflated norm that no other part of the model uses. Removing the `theta`
makes the norm match the force state the model actually applies. (This is the
change we already agree on; it controls the *yield onset*.)

## Correction 2 -- plastic update (line ~322): consistent return mapping vs. a divergent explicit step

This is the contested one. Two precise points.

**(a) The new form is the exact closest-point projection onto the (unchanged) yield
surface.** Write the new update as

```
edp_new = edp + (e_d - edp)*(1 - radial)    =>    (e_d - edp_new) = radial*(e_d - edp)
```

Every bond's elastic deviatoric extension is scaled by the same `radial`, so on the
next evaluation `t_d = K*(e_d - edp_new) = radial*t_d_trial` and therefore
`||t_d|| = radial*tdnorm = sqrt(2*psi0)` **exactly** -- the state lands on the yield
surface. That is the definition of a radial return. Because `radial` is in `(0,1]`
while yielding, the update moves `edp` toward `e_d` by a bounded fraction and can
never overshoot: it is unconditionally stable.

**(b) The old form differs from this by a factor that is not dimensionless, and that
factor is the source of the divergence.** Substituting Eq. 44 into the old update:

```
rkNew*deltalambda = radial*t_d_trial * (1/radial - 1)*m_i/(15*mu)
                  = (e_d - edp)*(1 - radial) * (omega_i + omega_j*m_i/m_j)
```

i.e. the old increment is the *consistent* increment multiplied by
`(omega_i + omega_j*m_i/m_j)`. With the LAMMPS peridynamic influence function
`omega = 1/r0` and `m_i = m_j` in the interior, that factor is `2/r0`. For a
correct update this multiplier would have to be the dimensionless number `1`;
instead it carries dimension `1/length`. In SI units with a sub-millimeter horizon
(`r0 ~ 1e-3 m` here) it evaluates to `~2000`, so the old update overshoots the
consistent plastic increment by roughly three orders of magnitude the moment a bond
yields. Once `(1 - radial)*(2/r0) > 1` the bond's stored elastic extension flips
sign every step, the deviatoric force state oscillates and grows, and the
integration diverges.

This is consistent with the reading that "the dimensional problem is really
upstream, in how `t_d` and `deltalambda` are non-dimensionalized in the discrete
code relative to the paper." That is true -- and it is exactly why the fix is
written purely in **extension (length) variables**: `edp += (e_d - edp)*(1-radial)`
enforces the geometric definition of the return mapping directly and is therefore
*independent* of the `t_d`/`deltalambda` unit convention. So it both removes the
instability and side-steps the upstream convention question, rather than depending
on it. It is not a new yield criterion: the yield surface
(`(1/2)||t_d||^2 = psi0`) and the projected force state (`rkNew = radial*t_d_trial`)
are identical to before; only the integration of the *same* Eq. 45 flow rule is
changed from an explicit one-step form to the consistent (closest-point) form.

---

## Demonstration: deviatoric strain-and-hold (attached deck `in.peri_eps_hold`)

A fully periodic 10x10x10 block of a single `peri/eps` material is given **one
volume-preserving deviatoric strain** (`e_xx=+0.05`, `e_yy=e_zz=-0.025`, about
2.2x the yield strain), then **held at fixed shape** and velocity-damped to rest.
Bond breaking is disabled (`s00 = 1.0`) so the response is pure plasticity, not
fracture. Volume preservation keeps `theta ~ 0`, so this deck is dominated by the
plastic-update map (Correction 2). At fixed strain the only thing still evolving is
`edp`, updated once per force call while yielding: a consistent update is a
contraction toward the yield surface (stress settles, kinetic energy damps to ~0);
a divergent update keeps re-kicking the bonds every step (stress and kinetic energy
never settle), even though the imposed strain is constant.

Same deck, three binaries:

| variant | step-0 von Mises | outcome of the hold |
|---|---|---|
| baseline `develop` (neither fix) | 1.14e9 Pa | diverges to ~1e16 Pa; KE pinned at ~5e9 J for 20000 steps |
| **norm fix only** (keep 577, revert 322) | 2.69e8 Pa | **diverges to NaN within ~200 steps** |
| **PR #5046** (both fixes) | 2.69e8 Pa | **stable plateau at 3.25e8 Pa; KE damps to ~1e-13 J** |

See `peri_eps_hold.png` (von Mises stress and kinetic energy vs. step, log scale)
for the baseline-vs-PR comparison.

(The stabilized plateau sits at ~0.33 GPa rather than the 1 GPa `m_yield` input;
that offset is coarse-lattice discretization -- this deck uses a horizon of only
`m = 3` lattice spacings, where the discrete force-state norm under-resolves the
continuum yield calibration. It is a property of the discretization, not of either
correction, and is irrelevant to the point here, which is stability vs. divergence.
A finer `m` moves the plateau toward `m_yield`.)

Two things to read off the table:

1. The **norm fix changes the step-0 stress** from 1.14e9 to 2.69e8 Pa (it corrects
   the yield onset -- Correction 1 is real and matters), but on its own it does
   **not** stabilize the run.
2. The "**norm fix only**" row is exactly the "accept one, revert the other"
   proposal. It still diverges -- to NaN. Reverting the line-322 update restores a
   catastrophically unstable model whether or not the norm fix is kept. Both
   corrections are required: 577 for the correct yield point, 322 for a convergent
   integration.

Why this was not caught before: the shipped `peri/eps` examples are
projectile-impact decks that fracture immediately, so bonds that would diverge
under sustained plastic flow simply break first (and `s00` is small). Disabling
breaking and holding a deviatoric strain is what exposes the underlying
plastic-update instability.

---

## Consistency with PR #5045

The corrected forms are also what the BPM re-implementation (`bond_style bpm/peri`,
PR #5045) uses. Whatever is decided here should be applied identically in both, so
the legacy `peri/eps` and the new `bpm/peri` EPS model agree in the plastic regime;
deciding the two PRs separately would make them diverge there.

## To reproduce

- `in.peri_eps_hold` -- the deviatoric strain-and-hold deck (primary demonstration).
- `in.peri_eps_uniaxial` -- a slow uniaxial strain ramp (secondary; shows the same
  divergence in a stress-strain sweep).
- `plot_hold.py` -- reads the two logs and writes `peri_eps_hold.png`.

Build `develop` and the PR #5046 branch, run the same deck against each:

```
./lmp_develop  -in in.peri_eps_hold > log.hold.baseline.txt
./lmp_pr5046   -in in.peri_eps_hold > log.hold.fixed.txt
python3 plot_hold.py log.hold.baseline.txt log.hold.fixed.txt
```

peri_eps_hold

[peri_eps_demo.tar.gz](https://github.com/user-attachments/files/29114621/peri_eps_demo.tar.gz)

jtclemm:
Thanks, I'll take a look at the response. I'll probably still be reluctant to make the change (unless the AI demonstrates the form does follow the paper) just because I'd rather such decisions be made by someone who actually works with peridynamics. I don't have much interest in diving into the nitty gritty details of the particular formulation beyond comparing to the source paper.

stanmoore1:
> Thanks, I'll take a look at the response. I'll probably still be reluctant to make the change (unless the AI demonstrates the form does follow the paper) just because I'd rather such decisions be made by someone who actually works with peridynamics. I don't have much interest in diving into the nitty gritty details of the particular formulation beyond comparing to the source paper.

I wonder how many people actually use the PERI capability in LAMMPS vs a code dedicated solely to peridynamics? Seems like if no one can maintain PERI and it doesn't have many users, maybe it needs to be deprecated.

akohlmey:
> I wonder how many people actually use the PERI capability in LAMMPS vs a code dedicated solely to peridynamics? Seems like if no one can maintain PERI and it doesn't have many users, maybe it needs to be deprecated.

I see some occasional questions in the forum. I just did a search and hit this one:
https://matsci.org/t/errors-nan-in-thermo-killed-by-signal-9-was-encountered-when-using-peri-package-in-pd-lammps/58359

Which uses pair style peri/eps, so I should probably compile the bugfix PR code and see if that makes a difference (not likely given the thrust of the discussion, but worth checking out).

With #5045 and the associated significant speedup, it may become more attractive.

akohlmey:
> I'll probably still be reluctant to make the change (unless the AI demonstrates the form does follow the paper) just because I'd rather such decisions be made by someone who actually works with peridynamics.

I have a different non-expert take on the situation. I am convinced by the evidence provided, that there is a confirmed bug in pair style peri/eps. Now, while I cannot confirm that the fix is correct, I take the view that the change does not make it worse. It is still a good idea to take a note and have it validated in some form, especially since the same change has carried over to the BPM re-implementation. On the other hand, I am hesitant to add a warning to the docs (There is likely a bug in this feature, be we are not sure. Let us know if you can help us to fix it.)

jtclemm:
> I am convinced by the evidence provided, that there is a confirmed bug in pair style peri/eps. Now, while I cannot confirm that the fix is correct, I take the view that the change does not make it worse.

I have to disagree. I think band aids make models more opaque (straying from the literature reference) and it makes it less likely someone with expertise will dig into and check it. It's too easy to pass the buck. I argue someone needs to take the time to understand the model + solution since the appearance of fixing a bug (or making results look more physical/stable) is not a rigorous criterion for scientific computing. I'm not comfortable with outsourcing verification to AI.

For the lubrication poly model, I think there is evidence of enough issues that I would argue for deprecation (particularly with the totally broken support of fix deform). In this scenario, we can probably find an expert to review the change in the flow rule integration. 

Can you ask AI to render both the original expression and its new formulation in latex using the same notation as the source text? Then I could write Stewart or some other peridynamics expert for review.  Or at least it's easier for me to figure out the details if I don't have to mentally translate everything.

akohlmey:
> I have to disagree. I think band aids make models more opaque (straying from the literature reference) and it makes it less likely someone with expertise will dig into and check it. It's too easy to pass the buck. I argue someone needs to take the time to understand the model + solution since the appearance of fixing a bug (or making results look more physical/stable) is not a rigorous criterion for scientific computing. I'm not comfortable with outsourcing verification to AI.

Point taken. Thank you for being patient with me and convincing me that scientific rigor is required here.

> For the lubrication poly model, I think there is evidence of enough issues that I would argue for deprecation (particularly with the totally broken support of fix deform). In this scenario, we can probably find an expert to review the change in the flow rule integration.

When removing a feature like a pair style, I typically create an alias for it in the header for pair style deprecated and update the documentation here https://docs.lammps.org/Commands_removed.html I can make most of the corresponding changes in PR #5040 but it would help to have a few words to justify the removal (just saying "it was crap" is not looking good).

> Can you ask AI to render both the original expression and its new formulation in latex using the same notation as the source text? Then I could write Stewart or some other peridynamics expert for review. Or at least it's easier for me to figure out the details if I don't have to mentally translate everything.

Here you are:

[peri_eps_corrections.pdf](https://github.com/user-attachments/files/29144382/peri_eps_corrections.pdf)
[peri_eps_corrections.tex.gz](https://github.com/user-attachments/files/29144392/peri_eps_corrections.tex.gz)

jtclemm:
Thanks, the pdf is very helpful in figuring out the issue. A few observations.
1) The code does implement the definition from the text as I stubbornly insisted (the AI admits this in Section 6)
2) However, the point that I didn't appreciate (the big point of confusion) is the units of the influence function omega differ. The implementation in LAMMPS uses 1/r0 (units 1/L) whereas the paper assumes unitless (or it doesn't assume units at all). This then affects the units of \alpha between the manuscript and the code which further implies \Delta \Lambda's units differ. With LAMMPS's choice, \Delta \Lamda has units of L^6 F^{-1} (as Claude states in the pdf) whereas the paper requires \Delta \Lamda have units of L^7 F^{-1} such that e^dp_n+1 has units of L. 

So the old code IS implementing Eq. 45 but that Eq. does not technically apply b/c of lower level decisions made in the code (the definition of the influence function). Now that the origin of the bug is clear, we need to figure out how to solve it. Naively, I would think you'd just multiply a factor of L. But it looks like the proposed solution is slightly different. I'll dig into that later. 

Bottom line, I think this kind of explanation for discrepancies from the paper is a lot more understandable (and is necessary) rather than the inscrutable mix of variables that Claude wrote in the documentation:
" The update of the per-bond plastic deviatoric extension now uses the consistent radial-return form ``edp += (e_dev - edp)*(1 - sqrt(2*psi)/||t_d||)`` instead of the previous ``edp += rkNew*deltalambda``, which was dimensionally inconsistent and could diverge under sustained plastic flow."

jtclemm:
@akohlmey  I finished looking at the issue and my current opinion is that we should:
1) Accept the change removing the spurious factors of theta multiplying omega. This just looks like a mistake. 
2) Multiply the original expression for the integration of the plastic by the missing factor of (2/r0) that was hidden in the units of the influence function (not adopting Claude's full solution to use radial return)
3) Add a note with Claude's proposed full solution suggesting radial return should be considered in the future to explicitly avoid overshooting the yield surface. From my understanding, it's basically making sure the integration doesn't drift off the surface to ensure it can't cause issues in long simulations runs. Could note this as something to look out for in the doc page too.

Patches 1 and 2  would be explicitly correcting bugs in the implementation relative to the source report and should dramatically reduce the massive swings. While the radial return could be optimal, I do not want to be a judicator for peridynamics at the moment and think it'd be better to save that for another PR when there's a smaller reviewing burden (or when we get an expert to weigh in).

Does this seem reasonable to you? (and/or Claude AI)

Edit: so focus on the porting and fixing clear bugs in these 2 PRs, then worry about improving physics in another PR.

akohlmey:
> Does this seem reasonable to you? (and/or Claude AI)

@jtclemm 

Yes, and Claude didn't object either. I have requested to update the pull request along the lines of your suggestion and add comments (to both source code and documentation) and also then adapt the code in PR #5045 accordingly (and reset all test reference data that is changed by it). And finally, I asked to create a feature request issue, so we have a reminder of this this being only a temporary solution.

I will act on it in stages. This PR has the update now and is also synced with upstream.

akohlmey:
Some small additional comments from the AI:

1. **Direction of the factor.** The missing quantity is a *length* of order
   `r0/2`, not `2/r0`. In the code's units the literal increment
   `rkNew*deltalambda` is dimensionless while `edp` is a length, so the fix is to
   *multiply* the increment by `0.5*r0` (equivalently, divide out the spurious
   `2/r0` gain). Multiplying by `2/r0` would go the wrong way and square the
   blow-up. This matches your own dimensional reading (deltalambda is short by a
   factor of length) and the "missing factor is ~r0" line in the PDF. Implemented
   as:

   ```c++
   deviatorPlasticExtTemp[i][jj] = edpNp1 + rkNew * deltalambda * (0.5 * r0[i][jj]);
   ```

2. **Why this is the genuinely "minimal, not radial-return" version.** Dividing
   the literal increment by the *exact per-bond* factor `omega*(1 + m_i/m_j)`
   turns out to be algebraically identical to the radial return. So the only way
   to have a correction that is dimensionally consistent yet distinct from the
   radial return is to use the interior constant `0.5*r0` (the equal-weighted-
   volume value of that factor). That is what's committed: it is exact in the
   interior and leaves a bounded `O(1)` residual factor `(1 + m_i/m_j)/2` only at
   free surfaces. It removes the `~2/r0` (~10^3) runaway, so it "dramatically
   reduces the massive swings" as you expected, without claiming the surface-
   exact result.

The exact per-bond radial return is written out in a source comment and the doc
page as the deferred refinement, and I've opened a tracking issue for it (link
below) so it stays a visible TODO for an expert.

**Consistency with #5045 (bpm/peri).** Worth flagging explicitly: the new
`bond_style bpm/peri` EPS model already uses the exact extension-space radial
return, because it stores the plastic extension *per bond* (the legacy code
stores it per node/bond, which is why the conservative interior patch was the
natural choice there). So after this change the two implementations agree in the
interior and differ only at free surfaces, by the bounded factor above, until
the tracked issue is resolved (which would bring legacy peri/eps onto the same
exact form). I've trimmed the bpm/peri comment to point at the PERI derivation
rather than duplicate it. In practice that surface difference sits below the
peridynamic surface effect and coarse-lattice discretization error that the
model already carries, so co-existing with it for now seems fine.

jtclemm:
> Yes, and Claude didn't object either. I have requested to update the pull request along the lines of your suggestion and add comments

Great, thank you. I'll do a double check and maybe tweak some language then can move onto the other PR.

Oh, and the unit test one too.

jtclemm (approved):

PR #5048, opened 2026-06-20
GRAPHICS: color dump image bonds by a per-bond compute value

akohlmey:
**Summary**

Extend the `bond` keyword of the `dump image` command so it can color bonds by
a per-bond value taken from a `/local` compute (e.g. `compute bond/local`), in
addition to the existing `atom`, `type`, and `none` settings. A new
`dump_modify bmap` keyword sets the bond color map, exactly mirroring `amap`
(atoms) and `gmap` (grid cells).

For example, `dump ... image ... bond c_blocal 0.3` together with
`dump_modify ... bmap ...` colors each bond by the corresponding value of a
`compute bond/local` (bond length, energy, force, ...).

Also included is a modification to fix graphics/labels that allows it to select whether to display a colormap for atoms, bonds, or a grid.

**Related Issue(s)**

None.

**Author(s)**

**Licensing**

The C++ changes to `dump_image.{cpp,h}`, the documentation updates, and the
example inputs in this pull request were generated by Claude (Opus 4.8) under
the direction and review of the human author, who specified the design,
validated the results, and takes responsibility for the contribution. This is
also reflected by the `Co-Authored-By` trailer on the commit.

**Backward Compatibility**

Fully backward compatible. The `bond` keyword merely gains an additional
accepted value (`c_ID`/`c_ID[N]`); all existing inputs behave exactly as
before, and `bmap` is a new, optional `dump_modify` keyword.

**Implementation Notes**

- A third `Image` color map (index 2) is added for bonds; `dump_modify bmap`
  reuses the existing `amap`/`gmap` parser path with `map_reset(2, ...)`.
- The per-bond compute is consumed the same way `dump local` does: `local_flag`
  check, `compute_local()` + `INVOKED_LOCAL`, reading
  `vector_local`/`array_local[col]`/`size_local_rows`, with `ArgInfo` parsing
  the optional `c_ID[N]` column.
- `compute bond/local` and the dump image bond loop iterate the bond topology
  identically (local atoms, then bond slots, with the same `newton_bond` and
  deleted-bond skips), so the n-th drawn bond is the n-th compute row. This
  requires the compute to select the same bonds as the dump (normally group
  `all`); an inline bounds guard plus a row-count consistency check stop with a
  clear error if they diverge, rather than mis-coloring. The per-bond values are
  local to each rank, so no extra MPI communication is needed.
- Verified with the CMake build and on 1 and 4 MPI ranks: a controlled two-bond
  test confirms the correct value-to-color mapping (short bond = color map min,
  long bond = max; one color per bond, no off-by-one). The new BPM fracture
  example runs cleanly through bond breakage (BPM removes broken bonds from the
  topology, so the compute and the dump stay in sync). Remaining cross-rank
  image differences match those of the pre-existing `bond type` coloring
  (render/merge tie-breaking, not a coloring difference).

**Post Submission Checklist**

The new `examples/GRAPHICS/in.bpm-fracture` colors the bonds of a fracturing BPM
disk by their force using a matplotlib-style "plasma" color map (built from
custom `map1..map5` colors), so the near-critical, highest-force bonds glow
yellow at the radiating crack tips.

---

image

[bpm-fracture.webm](https://github.com/user-attachments/assets/c63f6bc2-b35d-40cb-bd14-8fc8e1ac9ad3)


akohlmey:
@jtclemm Is this what you were looking for?

jtclemm:
> @jtclemm Is this what you were looking for?

Yes! That is what I was referring to when I floated the idea. Let me look at the code now.

examples/GRAPHICS/README
```diff
@@ -16,6 +16,10 @@ Here is a list of the inputs and which graphics features they demonstrate
 - in.breakable:             using autobond, color-by-property, adding objects and text
                             https://www.youtube.com/watch?v=f4hfPs7aCmI
 
+- in.bpm-fracture:          brittle disk (bonded particle model) shattered by a
```
jtclemm, line 19: Just to confirm, is this a duplicate of the example in the BPM folder?

Not a big deal, but just wondering if there's a difference/need to have two copies.

jtclemm (approved):

examples/GRAPHICS/README
akohlmey: Yes, it uses the same basic input. But the purpose is different: this one shows how to do the visualization. I just see that I forgot to add a few features to that input, like adding a color scale.

jtclemm: That's fair, I just wanted to make sure that was the intention.

akohlmey:
> > @jtclemm Is this what you were looking for?
> 
> Yes! That is what I was referring to when I floated the idea. Let me look at the code now.

FWIW, you can prototype and test the dump command for this with the new version of LAMMPS-GUI from the Atoms/Bonds menu. https://github.com/akohlmey/lammps-gui/releases/tag/v3.0.0
https://lammps-gui.lammps.org/visualization.html#atoms-bonds-settings

PR #5051, opened 2026-06-24
KOKKOS: fuse device atom sort into two generic kernels

stanmoore1:
**Summary**

KOKKOS: fuse device atom sort into two generic kernels

**Related Issue(s)**

None

**Author(s)**

**Licensing**

I used Claude Code (Opus 4.8).

**Backward Compatibility**

Yes

**Implementation Notes**

Description from Claude:

## Summary

Device-side atom sorting was dominated by kernel-launch overhead. `AtomKokkos::sort_device()` builds the permutation vector once, but then each atom style's `sort_kokkos()` called `Kokkos::BinSort::sort()` **once per per-atom array** — 6 calls for `atomic`, 30+ for `full`/`molecular` — and each of those internally launches a permute kernel plus a copy-back, with a fence on either side. On GPUs this is many tiny launches per sort event.

This PR replaces the per-style implementations with a single generic, mask-driven path in `AtomVecKokkos`, modeled on the existing comm/exchange packing kernels:

- **`AtomVecKokkos_PackSortFunctor`** — one kernel gathers *all* per-atom arrays into a scratch buffer in sorted order, indexed by the BinSort permutation vector (`get_permute_vector()`), with each field gated by an `if (_datamask & X_MASK)`-style bit test.
- **`AtomVecKokkos_UnpackSortFunctor`** — one kernel copies them back in place.
- **`AtomVecKokkos::sort_kokkos()`** — the base driver, now a concrete method instead of pure-virtual.

So the per-sort kernel count is now fixed (permutation build + gather + copy-back), independent of atom style, versus `O(#arrays)` before.

## Key idea

The set of arrays a spatial sort must permute is exactly the *persistent* per-atom set that travels when an atom migrates ranks during `exchange`. So the sort reuses `datamask_exchange` / `size_exchange` and the same field layout as `pack/unpack_exchange`, just indexed by the permutation instead of a sendlist. This was verified against `dpd`, whose `fields_exchange` (`dpdTheta, uCond…uCGnew`) matches the old sort set and correctly excludes the grown-but-recomputed `rho`/`duChem`.

## Changes

- New generic sort functors + base `sort_kokkos()` driver and a reusable `k_buf_sort` scratch buffer in `atom_vec_kokkos.{cpp,h}`.
- Removed the now-redundant per-style `sort_kokkos` overrides for `atomic, charge, bond, angle, molecular, full, dipole, sphere, dpd, spin` (they inherit the base).
- `hybrid` and `ellipsoid` keep their overrides and the existing legacy host-sort path (both already fall back to host sorting).

## Validation (KOKKOS Serial build)

- All edited translation units compile; full `lmp` builds.
- **Atomic LJ melt** (`DEFAULT=1` fast path) and **micelle bonded system** (`DEFAULT=0`, exercising the variable-length bond + special-neighbor topology copy): device-sort-on vs sort-off agree to floating-point round-off (summation-order only — bit-identical at step 0, no divergence/NaN over the run). The bonded run completes with no "bond atoms missing" error, confirming the topology/special arrays are permuted correctly.
- Confirmed the `sort_device()` path is actually used (no legacy-fallback warning).

## Follow-up (not in this PR)

Fixes with per-atom arrays (`fix_neigh_history_kokkos`, `fix_qeq_reaxff_kokkos`, …) still call `Sorter.sort()` per array. They remain correct; fusing each fix's arrays via the same permutation vector is a natural next step.


stanmoore1:
I tested this on H100 for the `in.chain` benchmark and it is actually slower.

PR #4690, opened 2025-08-13
Extend fix property/atom/kk with support for ivector, iarray and darray

rbberger:
**Summary**

Starts implementing the extensions needed for fix `property/atom/kk` that are needed for #4673

**Related Issue(s)**

**Author(s)**

@rbberger 

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Backward Compatibility**

**Implementation Notes**

**Post Submission Checklist**

- [ ] The feature or features in this pull request is complete
- [ ] Licensing information is complete
- [ ] Corresponding author information is complete
- [ ] The source code follows the LAMMPS formatting guidelines
- [ ] Suitable new documentation files and/or updates to the existing docs are included
- [ ] The added/updated documentation is integrated and tested with the documentation build system
- [ ] The feature has been verified to work with the conventional build system
- [ ] The feature has been verified to work with the CMake based build system
- [ ] Suitable tests have been added to the unittest tree.
- [ ] A package specific README file has been included or updated
- [ ] One or more example input decks are included

**Further Information, Files, and Links**

PR #5052, opened 2026-06-24
KOKKOS: extend `fix property/atom/kk` to `ivector/iarray/darray` custom types

stanmoore1:
**Summary**

KOKKOS: extend `fix property/atom/kk` to `ivector/iarray/darray` custom types

**Related Issue(s)**

None

**Author(s)**

**Licensing**

I used Claude Code (Opus 4.8)

**Backward Compatibility**

Yes

**Implementation Notes**

Description from Claude:

PR #4690 added property/atom/kk but only made the DVEC custom type
device-resident (a single contiguous 2D view k_dvector). This extends
Kokkos support to the remaining custom types, choosing storage by shape:

  - ivector: width-1, so a single contiguous 2D view k_ivector, mirroring
    k_dvector (dvector must stay contiguous because fix rx/kk indexes
    dvector(ispecies,i) across many properties in device kernels).
  - iarray/darray: each property has its own column count, so the set is
    ragged and cannot be one rectangular view. These use the SPARTA
    "views of views" pattern -- an outer DualView<struct{inner DualView}*>
    that holds a dynamically-sized, device-accessible list of inner views.

This avoids a std::vector<DualView> container entirely. The outer views are
resized with Kokkos::view_alloc(Kokkos::SequentialHostInit) because the
struct elements wrap a DualView (non-trivial ctor + atomic refcount), so the
default parallel host init would race on the inner reference counts.

AtomKokkos::add_custom/remove_custom and ~AtomKokkos manage the new views;
remove_custom no longer memory->destroy()s the now-Kokkos-managed data (which
would free pointers aliasing into a view), and the base Atom destructor gets
an "if (ivector)" guard mirroring the existing dvector one.

Verified (KOKKOS Serial + MISC + DPD-REACT): all four custom types
round-trip under -sf kk matching the host baseline, survive atom migration
and write_restart/read_restart, and are valgrind-clean (no invalid frees,
0 bytes lost).


src/atom.cpp
```diff
@@ -321,7 +321,8 @@ Atom::~Atom()
 
   for (int i = 0; i < nivector; i++) {
     delete[] ivname[i];
-    memory->destroy(ivector[i]);
+    if (ivector) // (needed for Kokkos)
```
akohlmey: @stanmoore1 I don't understand this change. The `Memory::destroy()` methods are supposed to be safe to use on null pointers. That is why we try to initialize all class member pointers in the initializer to `nullptr` (see #5058 for a long list of not yet properly initialized cases) so that we can safely use either `delete[]` or `memory->destroy()` before (re-)allocating or in the destructor (even when the constructor aborts somewhere).

src/KOKKOS/atom_kokkos.cpp
```diff
@@ -439,8 +473,14 @@ int AtomKokkos::add_custom(const char *name, int flag, int cols, int ghost)
...
 {
+  // the per-atom data is Kokkos-managed (k_ivector/k_dvector are contiguous, and
+  // k_iarray/k_darray are views-of-views), so do NOT memory->destroy() it here --
+  // that would free pointers that alias into a Kokkos view.  For ivector/dvector
+  // the data lives in a shared contiguous view and cannot be freed per index, so
+  // just drop the legacy alias; for iarray/darray we free both the row-pointer
+  // array and the inner DualView's host+device storage.
+
   if (flag == 0 && cols == 0) {
-    memory->destroy(ivector[index]);
```
akohlmey, line 444: @stanmoore1 see my other comment. It should be safe to use `memory->destroy()` here if the pointer is properly initialized to nullptr. Omitting this can lead to memory leaks under unusual circumstances.

src/atom.cpp
stanmoore1: Fixed.

akohlmey (approved):

PR #4992, opened 2026-05-25
Add ILVES constraint solver

akohlmey:
**Summary**

This pull request adds a new fix ilves command that implements the ILVES constraint solver.

**Related Issue(s)**

Implements and thus fixes #4962 
Closes #4998 since we can include either this or the other PR.

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Artificial Intelligence (AI) Tools Usage**

The code in this pull request has been predominantly implemented by Claude Code Opus 4.7 perusing fix shake, fix rattle, fix restrain, and the GROMACS implementation of ILVES as references.

**Backward Compatibility**

N/A

**Implementation Notes**

The main differences to fix shake are that this solver supports chains of connected bonds and has a workaround for (near) linear angles. However, it currently requires some replicated global topology storage and the solver can be very slow for large clusters of connected bonds and thus its parallel efficiency may be poor. Thus for systems that are compatible with LAMMPS SHAKE implementation, it is usually faster to stick with fix shake. For as long as there is no overlap between atoms involved in constraints, it is also possible to use both fix ilves and fix shake at the same time.

**Post Submission Checklist**

- [ ] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [x] The source code follows the LAMMPS formatting guidelines
- [x] Suitable new documentation files and/or updates to the existing docs are included
- [x] The added/updated documentation is integrated and tested with the documentation build system
- [x] The feature has been verified to work with the conventional build system
- [x] The feature has been verified to work with the CMake based build system
- [x] Suitable tests have been added to the unittest tree.
- [x] One or more example input decks are included

Issue #4962, opened 2026-04-26
[Feature Request] New constraint solver


pablogr:
**Summary**

Proposal to include a novel parallel constraint solver into LAMMPS

**Detailed Description**

My name is Pablo García-Risueño, and I am writing on behalf of a research project focused on developing novel methods for constrained molecular dynamics. We have recently introduced a new algorithm that, in most cases, outperforms the current state of the art in terms of efficiency and/or accuracy. Moreover, it allows significant increases in the time step of the simulations. The article where we present the algorithm is available at the following link (JCTC):

https://pubs.acs.org/doi/10.1021/acs.jctc.5c01376

The article linked above is based on a parallel implementation within the GROMACS package that supports both shared and distributed memory parallelism for CPUs. We are confident that this parallel implementation can be naturally migrated to LAMMPS, making it possible to efficiently impose connected constraints on arbitrarily large sets of atoms (for example, C–C bonds in a protein backbone), overcoming a current limitation of LAMMPS (1).

We are writing to ask whether you might be interested in collaborating in the inclusion of this new algorithm in LAMMPS. We do not have available manpower at the moment for that, but we would be eager to support LAMMPS developers with any related implementation efforts.

Thank you very much for your time and consideration. With best regards.

(1) In LAMMPS, only small clusters of atoms can be constrained. This is sothe constraint calculation for a cluster can be performed by a single processor, to enable good parallel performance. A cluster is defined as a central atom connected to others in the cluster by constrained bonds. LAMMPS allows for the following kinds of clusters to be constrained: one central atom bonded to 1 or 2 or 3 atoms, or one central atom bonded to 2 others andthe angle between the three atoms alsoconstrained. This means water molecules or CH2 or CH3 groups may be constrained, but not all the C-C backbone bonds of a long polymer chain.

**Further Information, Files, and Links**

https://github.com/LorienLV/_PAPER_ILVES

akohlmey:
Revised implementation in PR #5054 just got merged

PR #5054, opened 2026-06-25
Add fix ilves: ILVES bond and angle constraint solver (port of the GROMACS implementation)

akohlmey:
**Summary**

This adds `fix ilves`, a bond-length and angle constraint solver implementing the ILVES-M algorithm of Lopez-Villellas et al. (J. Chem. Theory Comput. 21, 8711, 2025). Unlike `fix shake`, ILVES solves the whole local constraint system as a single sparse direct (LU) solve per Newton iteration, so it handles arbitrarily large connected clusters (e.g. a whole protein backbone or long chains for connected atoms) directly. The numerical core is a largely verbatim port of the reference ILVES implementation the algorithm authors implemented in GROMACS; the LAMMPS interface (time-integration coupling, distributed solve, near-linear-angle handling, minimization) follows `fix shake`.

This is the fourth and most successful attempt. It can constrain clusters with chains of ~10,000 atoms (see image below containing 5 such chains) each in parallel with just the default settings and without excessive replicated memory use.

The completed feature set so far:

- Bond-length constraints (`b`/`t`/`m` selectors as in `fix shake`), solved by exact Newton iteration (one sparse direct solve of the whole local system per iteration).
- Angle constraints (`a` selector) via an A-C "virtual bond" (law of cosines), as `fix shake`.
- Distributed MPI solve (block-Jacobi over ranks); results are independent of rank count to the solver tolerance.
- r-RESPA (`run_style respa`) supported, with level-dependent timestep coupling as `fix shake`.
- degrees of freedom reported to temperature computes; constraint contribution to the pressure virial (on by default, like `fix shake`).
- Tolerance-based convergence (default like `fix shake`) or `mode fixed` for a fixed iteration count; per-type statistics output and optional per-atom constraint-force storage (like `fix shake`).
- Type labels: the `b`/`a`/`t` type selectors accept symbolic type labels from a `read_data` labelmap, and the per-type statistics are reported by label, as in `fix shake` (whose statistics output gained the same, for consistency).
- Near-linear angles handling: (`linearangle error|skip|restrain <threshold>`, default `error` 175 deg), since the A-C constraint becomes rank-deficient near 180 deg.
- Energy minimization support: constraints are replaced by stiff harmonic restraints (`kbond`), as in `fix shake`.
- Only the exact-Newton (asymmetric) solver is provided. A symmetric quasi-Newton variant was prototyped and dropped: it factors once per step but consistently needed more iterations, making it marginally slower overall for typical systems. Simplifying the resulting code for just ILVES-M resulted in a measurable speedup
- Benchmarks for typical systems show the overall time spent in the fix is too little versus the force computation to warrant implementing SIMD or OpenMP support (Amdahl's Law!)
- LAMMPS only supports double precision so the "real" alias from the GROMACS' version was dropped and without needing to support SIMD and threading some code became dead and was removed or simplified

**Related Issue(s)**

Addresses #4962. Supersedes the earlier attempts #4992 and #4998 and a first attempt that was not completed.

**Author(s)**

Axel Kohlmeyer (Temple University), corresponding author, akohlmey@gmail.com. The ILVES algorithm and the ported solver code are the work of L. Lopez-Villellas, C. C. K. Mikkelsen, P. Garcia-Risueno and co-authors (cited in the source and the documentation), reused with their encouragement from their GROMACS contribution.

**Licensing**

By submitting this pull request, I agree that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1). The ported solver files originate from the GROMACS ILVES implementation, which is distributed under LGPL v2.1; the ported files retain that attribution in their headers.

This pull request was prepared with the assistance of an AI coding tool (Claude Code, Anthropic, model Claude Opus 4.8) working under the direction, review, and testing of the author. The LAMMPS interface code -- the `fix ilves` class, the topology/comm/force coupling, the near-linear-angle handling, the minimization support, the tests, and the documentation -- was largely AI-generated and then reviewed, debugged, and corrected by the author. The numerical solver core in the `LAMMPS_NS::ILVES` namespace (the sparse direct solver, the CSR graph, the memory pool, the molecule topology, and the ILVES base and asymmetric-variant classes) began as a near-verbatim, non-AI port of the original GROMACS ILVES source. It was then adapted to LAMMPS and simplified with AI assistance: converted to double precision, stripped of the SIMD and multi-threading machinery this serial port does not use, and -- since for a single process the GROMACS Schur-complement partitioning reduces to one block -- collapsed from the partitioned Schur solver to a plain single-matrix sparse LU factorization. The underlying numerical algorithm (minimal-degree reordering, fill-in, LU factor/solve, Jacobian assembly) is unchanged from the reference.

**Backward Compatibility**

Fully backward compatible. This adds a new fix and new example/test files.

**Implementation Notes**

- Force coupling matches `fix shake`: predict the unconstrained positions in `post_force`, solve the constraints, and convert the multipliers to forces on `atom->f` so Verlet integration produces the constrained positions.
- The distributed solve drives a global Newton loop uniformly on all ranks (so ranks with no local constraints still join the collectives), reverse-summing per-atom increments to owners and forward-communicating predicted positions to ghosts each iteration.
- Verified against `fix shake`: identical steady-state energy/temperature to integrator  noise, rigid bonds/angles to the solver tolerance, rank-independent dynamics (np=1 vs 4  agree to several digits, set by the solver tolerance). Velocities are projected onto the constraint manifold once at run start (as `fix shake`'s `correct_velocities`), with no  er-step velocity projection -- SHAKE-style coupling, not `fix rattle`'s. A solvated peptide run with `fix ilves` on the protein and `fix shake` on the water matches the all-`fix shake` reference.
- Near-linear angles are characterized empirically (a tumbling-triatomic test): the A-C constraint degrades past ~165-170 deg and loses atoms near 179 deg, which sets the `linearangle` default threshold of 175 deg.
- Minimization replaces constraints by stiff harmonic bonds; with `bond_style zero` (no force-field bond force) the constrained bonds are pulled to their target by the restraint alone, in serial and on 4 ranks. The `kbond` defaults are expressed as multiples of the Boltzmann constant so they scale with the unit system.
- An example adapted from the SARS-CoV-2 main protease (Mpro) system of the ILVES publication is included (`in.covid`): `fix ilves` constrains all 4697 protein bonds as a single connected cluster -- exactly the regime the single sparse solve targets -- while `fix shake` keeps the water rigid. The force field is an approximate CHARMM36 conversion (GROMACS -> ParmEd -> VMD/topotools -> LAMMPS, with CMAP omitted for simplicity).
- New force-style regression tests (bond and angle) and docs; builds and runs under both the conventional and CMake build systems.

**Post Submission Checklist**

ILVES: https://doi.org/10.1021/acs.jctc.5c01376

----
image


akohlmey:
@sjplimp @jtclemm @ndtrung81 I feel this is yet another marked improvement over the previous edition(s) and finally at a level where short of the known performance deficits for extreme use cases, it is production level code. A lot of it is owed to importing large chunks from the GROMACS implementation and then retooling it for use with LAMMPS.

akohlmey:
@sjplimp @jtclemm @ndtrung81 Given that this now behaves as expected for all "normal" *and* extreme cases and that the only known deficiency is the unexpected scaling behavior for very extreme cases (which are *not* covered by the ILVES algorithm paper), I've flagged this now as complete and ready for review and merge. The performance issues for extremely large connected clusters can be handled in a follow-up pull request.

doc/src/fix_ilves.rst
```diff
@@ -0,0 +1,319 @@
...
+number of Newton iterations per step, and the frequency of statistics output
+(0 turns it off).
+
+Then one or more groups of selectors (``b``, ``a``, ``t``, ``m`` lists) pick
+which bonds and angles get constrained.  A bond is constrained when **both** of
+its atoms are in the fix group AND at least one of the selectors matches:
+
+* the bond type is in the *b* list, or
+* either atom type is in the *t* list, or
+* either atom mass is within 0.1 mass units of any value in the *m* list.
```
stanmoore1: within 0.1 mass units seems arbitrary, shouldn't this follow fix shake: `All bonds connected to atoms of the specified masses will be constrained (within a fudge factor of MASSDELTA specified in src/RIGID/fix_shake.cpp).`

stanmoore1 (approved):
I approve, but had one comment

doc/src/fix_ilves.rst
akohlmey: This is *supposed* to be *exactly* the same as with fix shake. If it isn't it is a bug in fix ilves and I need to fix it. Will make some empirical tests this afternoon.

akohlmey: It could be that it is only a problem of the documentation...

akohlmey: Confirmed. The difference is only in the wording of the documentation. The 0.1 value is the exact MASSDELTA fudge factor and can be changed in fix ilves the same way as in fix shake at compile time. Will update the fix ilves docs.

ndtrung81 (approved):
Given the recent feedback from the ILVES authors that "it now appears identical to the GROMACS version (except for the multi-threading layer)" I think the PR is ready to merge.

akohlmey:
> Given the recent feedback from the ILVES authors that "it now appears identical to the GROMACS version (except for the multi-threading layer)" I think the PR is ready to merge.

Working on the slow convergence issue right now and the suggested reason for it is already confirmed. Now it becomes a question of implementing a proper solution.

PR #5056, opened 2026-06-25
Add OPENMP and KOKKOS package variants of the pair_style dispersion/d3

MarcD3:
**Summary**

This pull request adds OpenMP and Kokkos accelerated code to the OPENMP and KOKKOS packages. The existing pair style in the EXTRA-PAIR package is revised minorly for performance.

**Related Issue(s)**

None

**Author(s)**

Marc L. Descoteaux (marc.descoteaux.0@gmail.com)
Yizhong R. Hu
Ulrik Unneberg
William C. Witt
Affiliation: Harvard University

**Licensing**

AI tools were used to revise and debug the code added in this pull request. The Kokkos pair style was heavily revised from a human written version by the AI tools, before being re-examined and revised by a human developer. The AI tool used was the Codex VSCode extension with the Codex-5.2, Codex-5.3 and GPT 5.4 models.

**Backward Compatibility**

No changes in this pull request should cause backwards compatibility issues.

**Implementation Notes**

In the non-accelerated pair_style, we found in testing that the use of std::pow() was somewhat slow and not required. We rearranged the code to try to avoid using this function when it is known that the exponent is an integer or similar.

The code passes the existing manybody test. Some of the many variants were tested and the outputs compared, though not yet exhaustively.

**Post Submission Checklist**

None

### DIFF
--- a/src/EXTRA-PAIR/pair_dispersion_d3.h
+++ b/src/EXTRA-PAIR/pair_dispersion_d3.h
@@ -21,8 +21,33 @@ PairStyle(dispersion/d3,PairDispersionD3);
 #define LMP_PAIR_DISPERSION_D3_H
 
 #include "pair.h"
-#include <cmath>
-#include <type_traits>
+
+// global ad hoc parameters of the D3 model, shared by the plain pair style and
+// its accelerated variants.  They live in a named namespace, not at global
+// scope, because this is a style header and gets included all over the place.
+
+namespace LAMMPS_NS::DispersionD3 {
+
+  static constexpr double K1 = 16.0;
+
+  /*  reasonable choices for k3 are between 3 and 5 :
+      this gives smooth curves with maxima around the integer values
+      k3=3 give for CN=0 a slightly smaller value than computed
+      for the free atom. This also yields to larger CN for atoms
+      in larger molecules but with the same chemical environment
+      which is physically not right.
+      values >5 might lead to bumps in the potential.
+  */
+
+  static constexpr double K3 = -4.0;
+
+  static constexpr double AUTOANG = 0.52917725;     // atomic units (Bohr) to Angstrom
+  static constexpr double AUTOEV = 27.21140795;     // atomic units (Hartree) to eV
+
+  // conversion factor for the tabulated C6 reference values
+  static constexpr double AUTOANG3 = AUTOANG * AUTOANG * AUTOANG;
+  static constexpr double AUTOANG6 = AUTOANG3 * AUTOANG3;
+}    // namespace LAMMPS_NS::DispersionD3
 
 namespace LAMMPS_NS {
 class PairDispersionD3 : public Pair {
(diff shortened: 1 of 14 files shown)


src/EXTRA-PAIR/pair_dispersion_d3.h
```diff
@@ -21,8 +21,56 @@ PairStyle(dispersion/d3,PairDispersionD3);
 #define LMP_PAIR_DISPERSION_D3_H
 
 #include "pair.h"
+#include <cmath>
+#include <type_traits>
 
 namespace LAMMPS_NS {
+template <typename T> inline T pow2(T x)
```
akohlmey: These kinds of generic functions *must* *NOT* be in a "style header". It almost guarantees namespace clashes. For the non-KOKKOS versions, they are not needed, anyway because you can import `MathSpecial::square()`, and `MathSpecial::powint()` and perhaps add the automatic (integer given as double gate) to `math_special.h` as MathSpecial::powauto() (using only doubles). These need to be imported only in the .cpp file and thus are not a problem. For KOKKOS, you could add this templated and thus multi-precision version to `math_special_kokkos.h` but follow the same naming scheme as in the non-KOKKOS version. I think that adding pow4()/pow6()/pow8() is excessive powint() will do well enough. I would not accept anything else without extensive benchmarks showing a significant speedup.

src/OPENMP/pair_dispersion_d3_omp.cpp
```diff
@@ -0,0 +1,635 @@
...
+static constexpr double AUTOEV = 27.21140795;    // atomic units (Hartree) to eV
+
+/* ---------------------------------------------------------------------- */
+
+PairDispersionD3OMP::PairDispersionD3OMP(LAMMPS *lmp) :
+  PairDispersionD3(lmp), ThrOMP(lmp, THR_PAIR)
+{
+  suffix_flag |= Suffix::OMP;
+}
+
```
akohlmey, line 60: There is lots of excessive whitespace here, please remove.

src/OPENMP/pair_dispersion_d3_omp.cpp
```diff
@@ -0,0 +1,635 @@
...
+  const auto * _noalias const x = (dbl3_t *) atom->x[0];
+  const int * _noalias const type = atom->type;
+  const int nlocal = atom->nlocal;
+  const int * _noalias const ilist = list->ilist;
+  const int * _noalias const numneigh = list->numneigh;
+  const int * const * const firstneigh = list->firstneigh;
+
+  // Thread-local cn array to avoid race conditions
+  double *thr_cn = new double[atom->nmax]();  // Initialize to zero
```
akohlmey: Wouldn't it be better to use `auto thr_cn = std::vector<double>(atom->nmax);` here so that don't have to use (and not forget!) delete[]?

akohlmey (changes requested):
@stanmoore1 has to review the KOKKOS package changes.

stanmoore1:
@MarcD3 I had Claude review this PR, and it found and fixed several issues. I submitted a PR to your fork here: https://github.com/MarcD3/lammps-d3/pull/1. I think that will get this PR a lot closer to ready to merge. Let me know if you have any questions.

akohlmey:
@MarcD3 Please note that this PR now has merge conflicts with upstream that need to be resolved before that can be considered for merging.

PR #5057, opened 2026-06-25
Fix a batch of Coverity Scan static analysis defects

akohlmey:
**Summary**

This pull request addresses a batch of defects reported by Coverity Scan static analysis. It contains a set of small, localized changes (each in its own commit), plus a Coverity modeling file and consolidated export tooling under `tools/coverity/`. The fixes were selected from a full triage of the open Coverity reports as the ones that are genuine and confidently fixable

**Author(s)**

Axel Kohlmeyer (Temple University), akohlmey@gmail.com (corresponding author)

**Licensing**

The changes in this pull request were developed with the assistance of an AI tool (Anthropic Claude, Opus 4.8) operating under the direction and review of the author. The tool was used to triage the Coverity Scan defect reports, distinguish genuine issues from false positives, and draft the corresponding fixes; the author reviewed and verified every change. The individual commits carry a `Co-Authored-By` trailer reflecting this.

**Implementation Notes**

Tooling / infrastructure:
- `tools/coverity/coverity_model.cpp`: a Coverity modeling file for the custom `Memory` allocator (models `smalloc`/`srealloc`/`sfree` as alloc/realloc/free).
- `tools/coverity/` export scripts + READMEs: local helper tooling to export the defect grid and per-defect event traces from the Coverity Scan web interface (the open-source tier offers no direct export).

Bug fixes (one commit each):
- GRANSURF (`pair_surf_granular`, `fix_surface_global`, `fix_surface_local`): guard a negative contact-map index; make a non-exhaustive triangle corner-flag branch exhaustive; check the STL reader return value before using it as a size.
- `fix gemc`: free the three `RanPark` RNGs in the destructor (leak) and initialize the pointer members.
- RIGID (`fix_rigid`, `fix_rigid_small`): convert the `readfile` `FILE*` to `SafeFilePtr` (use-after-free on the error path), fill a `molecule` branch gap, null the bonus pointers in `set_xv`, and avoid dereferencing end iterators in `rendezvous_body`.
- `temper` / `temper/npt` / `temper/grem`: validate the broadcast swap-partner index before using it to index the world arrays.
- `read_restart` / `variable`: fix unterminated-string reads (use `memcmp` on the fixed-length magic string; bound and null-terminate the `fread` buffer).
- `neighbor`: give the base `create_kokkos_list` an implementation that raises an  internal error instead of silently doing nothing (clears an uninitialized-list false positive while encoding the invariant).
- `mliap_model`: use `SafeFilePtr` for the coefficient file (RAII close on every path; removes the use-after-close and the exception-mode leak).
- `replicate`: guard the `newtag` map lookups against a missing atom.
- `fix_phonon`: initialize and range-check the Gauss-Jordan pivot indices (also yields a graceful singular-matrix error).
- `reader_native`: reject a negative per-chunk element count when skipping a binary dump frame.

Verification: builds cleanly with the CMake build system (`gcc` + `most` presets); `make check-whitespace` and `make check-permissions` pass.

**Post Submission Checklist**


src/PHONON/fix_phonon.cpp
```diff
@@ -810,6 +811,7 @@ void FixPhonon::GaussJordan(int n, std::complex<double> *Mat)
         }
       }
     }
+    if (icol < 0) error->one(FLERR,"Singular matrix in complex GaussJordan!");
```
stanmoore1: I don't think the exclamation point is needed here.

stanmoore1 (approved):
I approve, but had one comment.

src/PHONON/fix_phonon.cpp
akohlmey: > I don't think the exclamation point is needed here.

Looks like Claude was adapting to the general style of the source file which is "freckled with exclamation points" all over and some other things that we usually don't do in LAMMPS. I'm going to do a general exorcism.

jtclemm: > I'm going to do a general exorcism.

I'm going to vote that we start adding more emojis to error messages in LAMMPS to make it more exciting when simulations crash.

"Non-numeric positions - simulation unstable :("

akohlmey: > adding more emojis to error messages in LAMMPS

Nice idea!

But on the serious side, this only makes sense if we allow UTF-8 characters and that would open a can of worm that I prefer to leave closed until we have time to do a general audit of the core of LAMMPS to be able to handle this cleanly. We avoid a lot of trouble by sticking to (7-bit) ASCII and rather transparently transform UTF-8 lookalikes to their ASCII equivalents. BTW: this also gets us on the good side of software secrurity people(!).

akohlmey: > "Non-numeric positions - simulation unstable :("

Wouldn't it be better to have: "Non-numeric positions - simulation unstable :exploding_head:"

Or use :facepalm: when people make impossible choices.

jtclemm: > Wouldn't it be better to have: "Non-numeric positions - simulation unstable 🤯"
> 
> Or use 🤦 when people make impossible choices.

Could we have a randomly generated emoji to keep it fresh?

akohlmey: If you are going down this route you could go all the way and also do the same as GROMACS and Gaussian do and add a random "fortune" quote at the end of a run. But then again, we should aim for something different, so perhaps some ASCII-art images?
Curious to hear what @sjplimp thinks about this kind of stuff.

src/PHONON/fix_phonon.cpp
```diff
@@ -72,14 +72,14 @@ FixPhonon::FixPhonon(LAMMPS *lmp,  int narg, char **arg) : Fix(lmp, narg, arg)
 
   if (narg < 8) error->all(FLERR,"Illegal fix phonon command: number of arguments < 8");
 
-  nevery = utils::inumeric(FLERR, arg[3],false,lmp);   // Calculate this fix every n steps!
+  nevery = utils::inumeric(FLERR, arg[3],false,lmp);   // Calculate this fix every n step!
```
stanmoore1: I think the wrong char got removed here.

PR #5065, opened 2026-07-01
Improve accuracy of virial stress computation in fixes rigid and rigid/small by isolating constraint force contributions

l09rin:
**Summary**

This PR improves the computation of the virial in `fix rigid` and `fix rigid/small` by:

* excluding langevin thermostat forces (when present) and isolating the contributions from rigid constraint forces and gravity;
* computing the constraint virial by using particle positions relative to the center of mass of each rigid body;
* replacing the approximate evaluation of total forces with a consistent formulation from rigid body dynamics.

In the current implementation, constraint forces effectively include contributions from Langevin thermostat forces and gravity (when present), and absolute atom positions are used to compute the virial. Moreover, total forces on atoms are obtained approximately via numerical evaluation of particle accelerations.
This leads to the following issues:
* thermostat forces introduce stochastic noise in the virial, despite averaging to zero over time;
* the constraint virial depends on absolute positions, making it sensitive to rigid translations of the whole configuration;
* force evaluation is only approximate, introducing numerical errors depending on the integration timestep.

The changes included in this PR, detailed in the implementation notes below, should lead to more accurate and less noisy virial/pressure estimates by:
* removing spurious contributions from stochastic thermostat forces;
* improving consistency between rigid body dynamics and atom's force/acceleration evaluation;
* making constraint virial translationally invariant to further reduce statistical noise.

**Author(s)**

Giovanni Del Monte (University of Barcelona), giovanni.delmonte90@gmail.com.

**Licensing**

**Implementation Notes**

Key changes:

1. *Separation of force contributions*. Constraint forces are evaluated from rigid body accelerations and angular velocity. To do so, Langevin thermostat forces and gravity are subtracted from the total force acting on the center of mass `fcm`. The eventual virial contribution from gravity is evaluated separately, while thermostat forces are always excluded from the virial. This improves physical interpretability and removes spurious contributions, consistently with the behavior of `fix langevin`, if I didn't get it wrong.

2. *Constraint virial computed relative to center of mass*. The virial contribution of constraint forces is computed using atom positions relative to the center of mass of each rigid body, instead of absolute positions. This is physically consistent, since constraint forces sum to zero within each rigid body, and ensures translational invariance reducing statistical noise.

3. *Consistent computation of forces and accelerations*. In the previous implementation, constraint forces were inferred from particle accelerations computed approximately from velocity differences between current timestep `t` and `t+dt/2`. In this PR I propose to compute accelerations directly from rigid body forces and angular velocity, assuming they are evaluated at the current timestep; the contribution from angular acceleration is not included, as it does not contribute to the virial.
This improves numerical accuracy and ensures consistency between rigid body and atoms dynamics in the evaluation of virial.

**Post Submission Checklist**


akohlmey:
@l09rin Thanks for your submission, but this currently segfaults when running unit tests. Please also note that there is an upcoming KOKKOS port for fix rigid/small very close to be merged and also there is an OPENMP version of fix rigid/small. So any changes to the base class will have to be translated to the accelerated versions.

Please also note that LAMMPS uses velocity verlet so velocities are updated in two half steps, once with the old forces and once with the new forces after the position update.

l09rin:
Thank you @akohlmey, I am working on the failing tests first.
In the meantime, since in this version the calculation is based on forces, positions and velocities at the current timestep, would you and @sjplimp think that it is suitable to calculate the virial just once per timestep, only in the function set_v()? This is called by setup(), final_integrate() and the other methods changing velocities, then should be safe.

l09rin:
I moved the calculation of the virial totally within the method set_v(), based on rigid body quantities at the actual timestep. Segfault errors now are solved, but some tests still fail, I think due to the discrepancy between virial values that before were calculated by a numerically estimated acceleration.
I tested it with a specific system (using fix rigid/small) of 2D dumbbells each constituted of two rigidly bonded point particles. I checked that the static part of the virial (due to pair potential interactions and the part of constraint forces balancing forces) is identical to 12 digits to the value calculated with another simulation code on the basis of the rigid bodies (calculated by using centers of mass and total forces), as it should be [Computer Physics Reports 4.6 (1986): 346-392]. Also, the sum of the kinetic part of the virial and the part of the constraints counteracting the centrifugal acceleration in each rigid body is consistent with a value of NkT/V, where N is intended as number of rigid molecules.
I calculated on a specific configuration the relative difference between the values of the constraint forces as calculated using the procedure implemented in this pull request and the previous one, and it spanned in a range between 10^-3 and 10^-6, approximately. How were virial values calculated in the test reference configurations? With the algorithm used by fix rigid/small or they are exact (calculated on the basis of molecules instead of atoms)? Probably this discrepancy is what makes the tests to fail.

l09rin:
I accidentally deleted a line before the before-last commit, sorry.

Current status of failing tests on my machine:

- Test n. 21: a segfault occurs in the allocation of the random number generator, but I don't know if this is related to the changes contained in this PR, since it fails also in the develop branch;

- Tests n. 613-627: all of the tests `FixTimestep:rigid_*` fail at the check points EXPECT_STRESS, except for the nph and npt versions, also complaining for EXCEPT_POSITIONS and EXCEPT_VELOCITIES, due to the fact that the computed stress is used in the integration algorithm, I guess?

Before applying the changes to the omp/kokkos derived classes I would wait to check their correctness for the base classes of `rigid` and `rigid/small`. Do you have any hint on how to proceed?

l09rin:
Regarding last commit, when using MPI probably langextra values for ghost bodies need to be communicated, for the calculation of the virial. Also, sometimes a segfault error occurs when running on more than 1 processor, happening when at a certain point it tries to access to a body index overflowing langextra array length. This means that at a certain point nghost_body is updated, while langextra remains stale, probably, but I still didn't understand where exactly.

l09rin:
In the last commit I solved the issue for the communication of `langextra` for ghost bodies. I noticed that in the method `compute_forces_and_torques()` forces and torques are reverse-communicated before adding the contribution from langevin thermostat and gravity, then I just made the array `langextra` of length `nlocal_body` (as it is only used for local rigid bodies), and subtract its contribution from `fcm` only for local bodies, as ghost ones do not contain that contribution.

This solution works, but it is not particularly elegant, I guess, as now `fcm` and `tcm` contain different contributions for local and ghost bodies. Maybe this can bring confusion in case of further changes in the future, too. Then, we may probably move the incorporation of langevin thermostat in the integration steps, instead of leaving it in the method for calculation of forces, as is for the `fix langevin`. This would avoid the need of subtracting it to calculate the constraints virial and keep homogeneous the `fcm` and `tcm` quantities for local and ghost bodies. What do you think?

sjplimp:
@l09rin Thanks for contributing this PR.
Aside from excluding Langevin thermostat forces from the virial (which I agree is a good thing to do),
can you explain "how much" virial accuracy is gained by this versus the current implementation?

@ndtrung81 Can you look over these changes in the virial computation to see what you think and make any comments?

l09rin:
Hi @sjplimp, thank you, and sorry for the late reply, I was out of office in august.
In a couple of tests that I performed on a system of dumbbell particles the relative difference between the values calculated with the two different implementations were broadly between $10^{-5}$ and $10^{-8}$. If you need a more quantitative answer, I will give you after 15 sep., when I will have the time to work again on the PR.
In any case, I need to recheck again, because I realized later that even in the previous implementation there was an inconsistency, because the virial data communicated to other processors did not contain the langevin thermostat, then the calculation was not correct for atoms belonging to ghost bodies. This may be still in place in the last update, even though I do not expect it to affect the calculation that much.

PR #5072, opened 2026-07-03
Coverity Scan triage part 1: bug fixes

akohlmey:
**Summary**

This is the first of three stacked pull requests resulting from a complete triage of all 1809 outstanding defect reports in the new Coverity Scan project. It contains the fixes for defects classified as real bugs: wrong results, crashes, or undefined behavior reachable with valid or plausible inputs. One commit per bug or bug family; the commit messages list the Coverity CIDs and explain the failure modes. Highlights:

- AMOEBA `kmpole()` corrupted the sorted 1-3 neighbor list for most atom-ID orderings (wrong multipole frames/forces)
- the TIP4P OpenMP pair styles wrote per-atom energies to invalid indices when `compute pe/atom` was used without any virial request
- `Pair::init_tables()` produced garbage Coulomb tables for `pair lj/cut/coul/esp` under `run_style respa`
- pair `edpd` applied uninitialized heat flux to neighbor atoms whenever the temperature cutoff is smaller than the force cutoff, and clamped the wrong random number for the random heat flux
- restart-gap fixes: settings parsed by `settings()`/`coeff()` but never stored in the restart file (gran/hooke/history `limit_damping`, four of six `dsmc` settings, the oxDNA/oxRNA `seqav/seqdep` choice, the entire `eff/cut` ECP parameter state)
- region cylinder produced garbage granular wall contacts at cap rims with an open side face; region cone had three copy-paste errors in its hi-boundary parsing
- `fix gemc` could lose or duplicate atoms in triclinic exchanges (fractional coordinates read before being broadcast)
- assorted crash/overflow guards: `variables[-1]` accesses from the `next` command and `py_` typos, MLIAP size checks running after the truncation they guard, `NPair::coord2bin_big()` computing its 64-bit result with 32-bit multiplies, several KOKKOS ports that had dropped error checks or narrowed 64-bit arithmetic present in their CPU counterparts

Also included is a mechanical follow-up to #5058 initializing class members in about 20 files that the earlier batch missed.

**Related Issue(s)**

Follow-up to the merged #5057 and #5058. The remaining Coverity findings are addressed in the stacked pull requests #5073 (hardening) and #5074 (regression tests) based on this branch.

**Author(s)**

**Licensing**

The Coverity report triage and the code changes in this pull request were generated with Claude Code (Claude Fable 5) under the direction of and with review, verification, and testing by Axel Kohlmeyer.

**Backward Compatibility**

- The restart file format grows for pair styles gran/hooke/history (and hooke/hertz siblings), dsmc, oxdna/hbond, oxdna/stk, oxrna2/stk, and eff/cut; as usual, restart files are only compatible within the same LAMMPS version.
- `region cone` inputs that relied on the broken hi-boundary parsing (`EDGE` in the lo argument affecting hi, `hi INF` on a y-axis cone) now behave as documented.
- `dynamical_matrix` and `third_order` now always open their documented default output files; previously the former silently discarded all output and the latter crashed without the `file` keyword.
- `fix adapt` now rejects scaling diameter or charge through a value of exactly 0.0 (previously a division by zero destroyed the atom data irrecoverably).

**Implementation Notes**

Every fix was verified against the affected code path; the branch passes the full unit test suite locally (pair/bond/angle/dihedral/improper style suites, FixTimestep, Dump, DEM/granular, region and variable command tests) plus a granular restart round-trip check. Details on the triage classification (99 real defects out of 1809 reports, dismissal rules for the ~80% false positives) are recorded in the commit messages and in the triage records kept locally.

**Post Submission Checklist**


athomps (approved):
Coverity + Claude Code + Axel, what a great combination.

PR #5073, opened 2026-07-03
Coverity Scan triage part 2: defensive hardening

akohlmey:
**Summary**

Second of three stacked pull requests from the Coverity Scan triage (based on the bug-fix branch of the first PR; only the last six commits are new here). This one contains defensive hardening for findings that are safe today by invariant but cheap to guard, in four thematic commits:

- validation of size and index fields read from binary restart files and from fix bond/react map files (corrupt or hand-edited files now produce clear errors instead of allocation failures, heap overflows, or stalled read loops); guards on broadcast data are placed after the broadcast so all MPI ranks take the error path collectively
- validation of user parameters used as divisors or lookups (pair exp6/rx site species, pair ldd coefficients, MEAM rho0, RHEO equilibrium density, fix pour diameters, pair gauss mixing with b = 0)
- initialization and error reporting for geometrically guaranteed searches (rigid-body bonus pointers, GRANSURF shared-edge connectivity, BOP neighbor coordination searches) and for degenerate all-faces-open regions
- an assortment of exhaustive-chain else branches, overflow-safe rewrites (the hyper fixes' overflow guards could never trigger), socket input validation for fix ipi, SafeFilePtr for fix qeq/comb, and consistency fixes between base and accelerator variants

It also adds a corrupted-restart regression test which immediately exposed (and the same commit fixes) a heap corruption in `Group::read_restart()`: the group names were deleted but not nulled, so failing to read a restart file in library mode (e.g. from Python, where errors are exceptions) led to a double free when the LAMMPS instance was destroyed or reused.

**Related Issue(s)**

Based on #5072; the stacked pull request #5074 adds the remaining regression tests.

**Author(s)**

**Licensing**

The Coverity report triage and the code changes in this pull request were generated with Claude Code (Claude Fable 5) under the direction of and with review, verification, and testing by Axel Kohlmeyer.

**Backward Compatibility**

No changes for valid inputs: the full style test battery passes with none of the new guards triggering. Previously silently accepted invalid inputs (corrupt restart files, out-of-range cvff multiplicities, zero pour diameters, etc.) now produce errors.

**Implementation Notes**

The restart guards were exercised against valid restart files of all bonded hybrid and LEPTON/gaussian-family styles (no false positives) and against deliberately corrupted and truncated files in the new test. Local validation: full unit test suite green.

**Post Submission Checklist**


athomps:
This all looks good to me. I will approve when requested.

athomps (approved):
I approve.

PR #5074, opened 2026-07-03
Coverity Scan triage part 3: regression tests

akohlmey:
**Summary**

Third of three stacked pull requests from the Coverity Scan triage (based on the bug-fix and hardening branches; only the last four commits are new here). It adds regression tests designed to catch the *classes* of bugs found in the triage, not just the instances:

- `test_restart_settings` (commands): write/read/write restart round trips must be byte-identical (with the instance counter in internal fix IDs masked), verifying the restart I/O reads back exactly what it writes for the styles whose settings were added to the restart format; the granular case additionally verifies the restored `limit_damping` flag behaviorally, since a setting missing from both writes is invisible to a round-trip comparison
- `test_input_validation` (commands): clean errors for the newly validated inputs (cvff multiplicity, fix vector nmax, indexed atom-style variables in fix ave/grid, empty RANN potential files that previously hung forever) and a no-crash check for fix store/state on a proc without atoms
- `eatom_only`/`eatom_only_omp` subtests in the pair style tester: direct pair invocations with per-atom energy as the only tally request, compared per-atom against a full-virial reference. The regular test runs cannot reach this code path because their thermo output schedules the virial on every sampled step; with the TIP4P fixes reverted these subtests catch the bug for all six affected pair style configurations, and all 262 pair style tests pass with the fixes. Every pair style YAML gains this check without regenerating any reference data.
- a DEM regression test bouncing a sphere off the cap rim of an open-side region cylinder through fix wall/gran/region; it fails against the recorded trajectory when the region cylinder fix is reverted

**Related Issue(s)**

Based on #5072 and #5073.

**Author(s)**

**Licensing**

The tests in this pull request were generated with Claude Code (Claude Fable 5) under the direction of and with review, verification, and testing by Axel Kohlmeyer.

**Backward Compatibility**

Test-only changes; no impact on simulation behavior.

**Implementation Notes**

Every test was validated in both directions: green on the fixed code and failing (or catching a segfault) with the corresponding fix reverted. Notable implementation constraints are documented in the code: dynamics-based runs cannot produce the energy-without-virial tally combination, a sum rule over per-atom energies cannot detect misdistribution to stale in-range indices, and in-process restart comparisons must mask the `Pair::instance_me` counter embedded in internal fix IDs.

**Post Submission Checklist**


akohlmey:
@copilot resolve the merge conflicts in this pull request

akohlmey:
The BIGBIG unittest failure (ManybodyPairStyle:rann, subprocess abort) is fixed by 639b07387a: the eatom_only reference invocation requested a per-atom virial, which pair rann legitimately rejects with an error, and the escaping exception left the gtest stdout capturer active, aborting the test program at the next captured section. The reference now requests only the global virial in the mode the integrator would pick for the style (fdotr, or pair-wise for no_virial_fdotr_compute styles), and the capture section is exception-safe so a throwing style can only fail its own test. Verified locally: manybody, molecular, and atomic pair suites all pass, and the TIP4P negative control (reverting the fix from #5072) still catches the original bug.

athomps (approved):
`It adds regression tests designed to catch the classes of bugs found in the triage, not just the instances:`
Awesome!

PR #5077, opened 2026-07-07
Improve and compress figures in the documentation

akohlmey:
**Summary**

This pull request replaces several clunky or hard to understand figures in the 
documentation with recreations using either pure vector graphics or a raytraced
geometry overlayed with vector graphics.  Also several figures with large storage
requirements were compressed, in most cases lossless through using `optipng`.

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

The vector graphics parts were created by Claude Opus 4.8 under close supervision. The ray traced graphics were manually created using custom VMD scripting and then postprocessed by Claude Opus 4.8.

**Backward Compatibility**

N/A

**Implementation Notes**

N/A

**Post Submission Checklist**


jtclemm (approved):
These all look great, I vote to replace the old ones.

If I had to nitpick, the only complaint I might have is that the red distance vector in improper_distance.png is off center from the atom.

akohlmey:
> If I had to nitpick, the only complaint I might have is that the red distance vector in improper_distance.png is off center from the atom.

That was a deliberate choice. If you center the arrow under that atom, it looks as if the distance between the plane and the surface of the sphere not the center. A more elaborate fix would require to make the atom transparent so that you see the tip of the arrow.
I can look into creating such a version, but my general experience with transparency in such figures is that one has to be very careful and it is often not helping. That is why I chose the off-center placement to both show the exact distance and avoid overlapping with the label etc.

PR #5076, opened 2026-07-05
Collected small changes and fixes

akohlmey:
**Summary**

This pull request combines multiple small changes and fixes that do not warrant a pull request on their own

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Artificial Intelligence (AI) Tools Usage**

The updated hydrogen bond image was created with the help of Claude Code using Opus 4.8.

**Backward Compatibility**

N/A

**Implementation Notes**

The following changes are included:
- revert PLUMED download to version 2.9.5 since 2.10.x has problems with our Windows cross-compiler scripts
- recreate the clunky hydrogen bond diagram for compute hbond/local to be rather spiffy
- add python script to rebalance tables in .rst file
- apply RAII principle for local allocations in a few more cases in the GRANSURF package
- apply static code analysis suggested linting and modernizing changes
- fix bug in dump image where per bond transparency was not applied correctly
- fix incorrect use of std::clamp() when processing lights info from a JSON file
- fix global energy accounting bugs and inconsistencies in pppm/tip4p with slab correction enabled
- generalize coding agent instructions to have a lean initial file and load additional files on demand, also share as much information between Claude code and Copilot.
- remove Coverity Scan workflow. We run this elsewhere now.

**Post Submission Checklist**

- [ ] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [x] The source code follows the LAMMPS formatting guidelines
- [x] Suitable new documentation files and/or updates to the existing docs are included
- [x] The added/updated documentation is integrated and tested with the documentation build system
- [x] The feature has been verified to work with the conventional build system
- [x] The feature has been verified to work with the CMake based build system
- [ ] Suitable tests have been added to the unittest tree.

PR #5078, opened 2026-07-07
KOKKOS versions of the TIP4P pair styles and pppm/tip4p

akohlmey:
**Summary**

This pull request adds KOKKOS versions of the four TIP4P pair styles (`tip4p/cut`, `tip4p/long`, `lj/cut/tip4p/cut`, `lj/cut/tip4p/long`) and of the `pppm/tip4p` kspace style. The pair styles are built on a shared, header-only base class template `PairTIP4PKokkos<DeviceType, PairCPUBase>` that owns the common TIP4P device machinery (M-site pre-kernel with tag-based H lookup, force redistribution, tally functions, long-range Coulomb tables); each concrete style only implements its own interaction kernel. `PPPMTIP4PKokkos` derives from `PPPMKokkos` and overrides the five virtual hooks that must operate on the M-site positions.

Validating the port against the CPU styles uncovered two bugs in the *CPU* slab correction of `PPPMTIP4P` (also inherited by `pppm/tip4p/omp`, one of them also present in `pppm/disp/tip4p`), which are fixed here as well.

Limitations (clean errors at init): requires newton on, orthogonal boxes, and `kspace_modify diff ik` (the latter inherited from `PPPMKokkos`).

**Related Issue(s)**

None.

**Author(s)**

Axel Kohlmeyer (Temple University), akohlmey at gmail.com

**Licensing**

The code and modifications in this pull request were developed with substantial assistance from Anthropic's Claude (via Claude Code using Opus 4.8 and Fable 5): the KOKKOS port, the code review, and the fixes derived from it were largely AI-generated under direction of the author. All changes were reviewed, tested, and validated by the (human) author.

**Backward Compatibility**

No input script syntax changes. However, the two CPU-side bugfixes change numerical outputs for existing inputs that use `kspace_modify slab` with `pppm/tip4p` (or `pppm/tip4p/omp`, `pppm/disp/tip4p`):

- the slab correction energy was silently dropped from the reported long-range energy (accumulated into `energy_1`, which plain PPPM never reads); slab runs now report a correspondingly different total energy. Forces were and are unaffected.
- per-atom energies from the slab correction are now evaluated at the M site and redistributed onto O and the two H atoms, so that the per-atom energies sum to the global energy. For non-neutral slab systems the `dipole_r2` term and the force prefactor are now also evaluated at the M site (for neutral systems, forces and global energies are unchanged by this).

**Implementation Notes**

The branch went through two rounds of a thorough multi-agent code review, with a focus on single/mixed precision support and consistency with recent KOKKOS package conventions. Review findings and their resolution:

- *Missing TIP4P hydrogen caused out-of-bounds device writes.* The M-site pre-kernel records a missing H atom (communication cutoff too small) as a `-1` sentinel, but the compute kernels never checked it and would issue `Kokkos::atomic_add()` with index `-1`. The kernels now check the sentinel at the point of use — the same semantics as the lazy `find_M()` lookup in the CPU styles, so ghost O atoms whose interactions are never needed cannot trigger spurious aborts — and stop with the CPU styles' "TIP4P hydrogen is missing" error. Verified on 4 ranks that CPU and KOKKOS abort identically.
- *CPU `PPPMTIP4P::slabcorr()` accumulated into the dead `energy_1` variable* (apparent copy-paste from `pppm_disp_tip4p.cpp`, where `energy_1` is the correct target). Demonstrated with a slab system where `lj/cut/tip4p/long` with `qdist = 0` must match `lj/cut/coul/long` + plain `pppm` exactly; fixed to accumulate into `energy`.
- *CPU slab correction mixed raw O coordinates with M-site coordinates.* The global dipole used the M site, but `dipole_r2`, the per-atom energy loop, and the force prefactor used the raw O position, so per-atom energies did not sum to the global slab energy. Fixed in `pppm_tip4p.cpp` and `pppm_disp_tip4p.cpp` (same code there), following the per-atom redistribution convention of `fieldforce_peratom()`; the KOKKOS slab correction kernels were aligned with the corrected baseline.
- *Stale `qscale` in `PPPMKokkos::compute()` (pre-existing in plain `pppm/kk`).* The member was assigned only after `fieldforce()` returned, so when `fix adapt` ramps the kspace scale factor the interpolated grid forces used the previous step's value (the CPU computes the factor fresh inside `fieldforce_ik()`). The assignment was moved before `fieldforce()`; a 5-step `fix adapt` kspace ramp now matches the CPU run to 13 significant digits.
- *Precision conventions.* Bare double-precision literals in the interaction kernels and tally functions silently promoted parts of the hot loop to double in single/mixed precision builds; the O/H force-split factors `1-alpha` and `alpha/2` are now precomputed on the host in double and narrowed once to `KK_FLOAT`. The explicitly tallied TIP4P virial is computed from absolute positions, so `v[6]` and the intermediates now use `KK_ACC_FLOAT`: the large i- and j-side contributions must cancel in the accumulation precision to avoid catastrophic cancellation in mixed precision builds. All translation units compile cleanly in double/double, single/double, and single/single configurations.
- *Efficiency.* The per-type-pair LJ coefficient views are copied to the device only when `init_one()` has run (setup, or `fix adapt` via `Pair::reinit()`) instead of on every `compute()` call; when no energy or virial is requested the kernels launch as `parallel_for` instead of a `parallel_reduce` over an always-zero `EV_FLOAT`.
- *Deduplication.* The 8 copies of the M-site force redistribution + virial + tally-list block were folded into `PairTIP4PKokkos::apply_site_force()` (the interaction sign is folded into the force-prefactor argument, which keeps every floating-point operation association unchanged — double precision results are bit-identical). The Coulomb interpolation tables, `init_tables()`, and the analytic-erfc/table force+energy branch, previously duplicated verbatim between the two long-range styles, moved into the shared base, as did the standard pairwise LJ `ev_tally()`. The device `closest_image()` (previously duplicated between the pair base and `pppm/tip4p/kk`) is now `AtomKokkos::closest_image_kokkos()` next to `map_kokkos()`; `fix shake/kk` still carries its own copy and could adopt it in a follow-up. The TIP4P parameter block that `PPPMKokkos::init()` had copied verbatim from `PPPM::init()` is now a shared protected `PPPM::init_tip4p()`.
- *Known remaining duplication (deliberate).* `PPPMTIP4PKokkos` duplicates the base `particle_map`/`make_rho`/`fieldforce_peratom` kernels with only the position view replaced by the precomputed M-site positions. Eliminating this would require the base `PPPMKokkos` kernels to read a bindable position view, i.e. touching the hot production kernels of plain `pppm/kk`; deferred for discussion.
- *Test coverage gap (for a follow-up).* The kspace force-style test harness does not compare the global `elong` scalar, which is why the dropped slab-correction energy went unnoticed; `kspace-pppm_tip4p_slab.yaml` passes both before and after that fix.

Correctness verification: CPU vs KOKKOS agreement to 13 significant digits for all four pair styles with the real TIP4P/2005 geometry (energies, pressure, per-atom energy and stress sums; 1 and 4 MPI ranks; including slab-corrected `pppm/tip4p` runs and a 20-step NVE trajectory); after each refactoring pass all double-precision results were verified bit-identical against the pre-refactor state. The TIP4P force-style and kspace test fixtures pass, including via the `kokkos_serial` fixture ladder from #5035.

**Post Submission Checklist**

The two CPU-side slab correction bugfix commits are also part of the `collected-small-changes` branch used by PR #5076; git will reconcile them regardless of which branch is merged first.


stanmoore1 (approved):
I approve

stanmoore1:
@akohlmey has this been tested on a GPU yet?

akohlmey:
> @akohlmey has this been tested on a GPU yet?

Only with HIP on my desktop with APU. I don't expect any problems, but you never know...

PR #5082, opened 2026-07-09
Add YAML-based test facility for dump image and GRAPHICS package rendering

akohlmey:
**Summary**

This pull request adds a new test facility in `unittest/graphics` for the `dump image` command and the rendering features of the GRAPHICS package, plus an initial set of 79 tests and one small rendering bugfix found while writing them.

Each test YAML file describes a complete scene (system setup plus the `dump image` and `dump_modify` commands under test) and stores sampled reference pixel data from the rendered image directly in the YAML file: per-row and per-column mean RGB projections (full-image coverage) and mean RGB values of small pixel blocks on a uniform grid (failure localization).  No reference image files need to be maintained.  The test driver doubles as the reference generator (`-u`/`-g`, analogous to the force-style tests) and keeps the rendered image as `<name>.failed.ppm` when a comparison fails, for visual inspection.

The initial test set covers: atom coloring by all attribute types incl. computes and color maps in all four styles, bond coloring incl. per-bond values from `compute bond/local`, custom colors, per-type diameters and transparency (atoms, bonds, box/axes/subbox lines), background gradients, lights, camera keywords (view/center/zoom/up), FSAA and seeded SSAO, autobond, box/axes/subbox drawing incl. triclinic boxes, ellipsoid and body particle rendering, region drawing in all four draw styles across six region geometries incl. open faces, union/intersect regions, and dynamic regions (move/rotate keywords and variable-driven radii, rendered after a 100-step run).

**Related Issue(s)**

None.  Part of the pre-release effort to expand test coverage.

**Author(s)**

**Licensing**

The test driver code, the test scene inputs, and this pull request were developed with the assistance of Claude (Anthropic) under human direction, with all code, scenes, and findings reviewed by the author.  The pixel-sampling scheme was validated in a systematic experiment (see Implementation Notes) before the harness was implemented.

**Backward Compatibility**

No changes to input syntax or behavior, with one exception: the included bugfix changes the rendering of bonds colored by atom when the two bonded atoms have different `dump_modify atrans` transparency values (previously the second bond half incorrectly used the transparency of the first atom).

**Implementation Notes**

*Sampling scheme validation:* before implementing the harness, candidate schemes (uniform 3x3 pixel blocks, gradient-selected feature blocks, coarse average-pooled downsamples, row/column mean projections) were evaluated against 13 deliberately introduced rendering regressions (subtle color changes, 1% zoom drift, missing objects, thin-line diameter changes, label text changes, SSAO seed changes) and legitimate variation (reruns, compilers, optimization levels, MPI rank counts).  The combination of row/column projections (tolerance 0.5/255) with a uniform block grid (tolerance 2.0/255) detected all 13 regressions, including a change affecting only 21 pixels; simpler block-only schemes missed several.

*Determinism:* serial rendering is bit-deterministic and was found to be bit-identical between GCC -O2, Clang -O2, and GCC -O3 -march=native builds on x86-64 Linux, including FSAA and seeded SSAO.  The only observed toolchain difference was +-1/255 on ~70 pixels where a computed value feeds a rendered text label (FMA contraction).  The tolerances leave ample headroom and are per-test adjustable.  Rendered images legitimately depend on the MPI rank count (depth-tie compositing at object silhouettes; SSAO partitions its random stream by rank), so every test records the rank count of its reference and is skipped for any other; CMake registers tests with nprocs > 1 through the MPI launcher (one such test exercises the parallel image compositing).

*Bugfix:* when bonds are colored by atom, each bond is drawn as two half cylinders.  The second half was drawn with the color of its atom but the transparency of the *other* atom.  Fixed to match the (correct) autobond drawing code; the `dump-modify-atrans-mixed` test is the regression test and fails when the fix is reverted.  The same fix is also included in #5076 via cherry-pick; whichever pull request is merged first turns the commit in the other into a no-op.

*Interesting cases documented in unittest/graphics/README.md:* `autobond` finds nothing on molecular systems with default `special_bonds 0 0 0` (bonded pairs are excluded from the pair neighbor list); `btrans` only applies when bonds are not colored by atom; continuous color maps require literal min/max first/last entries and discrete maps a final catch-all bin.

A second round is planned to cover the graphics objects provided by fix and compute styles (fix graphics/arrows, graphics/labels, graphics/objects, graphics/isosurface, etc., compute chunk/atom and hbond/local, fix indent, the wall fix family) using the same facility.

**Post Submission Checklist**

- [x] The documentation in Developer_unittest.rst describes the new test facility, with figures illustrating the pixel sampling scheme and how a rendering change is detected.

All 79 tests pass serially and with `ctest -j 8` (about 4 seconds wall clock).  Reference data was generated on Fedora 43 / GCC 15 / x86-64.


unittest/graphics/tests/dump-image-acolor-compute.yaml
```diff
@@ -0,0 +1,606 @@
...
+  group peptide type <= 12
+  compute ke all ke/atom
+  dump viz peptide image 1 ${imagefile} c_ke type size 200 200 zoom 2.15 view 80 30 center s 0.25 0.5 0.5 box no 0.0 shiny 0.2 bond atom 0.3
+  dump_modify viz backcolor gray
+run_commands: |
+  run 0
+image_size: 200 200
+sampling: 3 16
+row_means: |
+  128.00 128.00 128.00
```
athomps, line 35: This is a good thing.

athomps (approved):
This is a nice simple way to validate images efficiently.

akohlmey:
> This is a nice simple way to validate images efficiently.

Thanks. I had to talk the AI into doing it. :grimacing:

PR #5083, opened 2026-07-09
Add force-styles testers for minimizer styles and compute/fix output data

akohlmey:
**Summary**

This pull request adds two new YAML-driven test programs to `unittest/force-styles`, closing two gaps in the unit test infrastructure: minimizer styles had no tests at all, and compute styles had no YAML-based test facility.  98 tests are included (12 minimizer, 86 output-data tests).

- `test_min_style` tests `min_style` and `min_modify` settings: it sets up the fourmol molecular test system (same pattern as the fix-timestep tester), applies the minimizer selection and optional fixes from the post_commands block of the test YAML, and runs a minimization with a fixed iteration budget (100 iterations, zero tolerances) so that the reference data is deterministic.  Compared are the potential energy before and after minimization and all six box dimensions and tilt factors (so `fix box/relax` changes are captured and an unchanged box is verified everywhere else); the total force norm is checked as an upper bound (10x the reference) and the energy must have decreased.  Tests: cg with quadratic/backtrack/forcezero line searches and a dmax setting, sd, hftn, quickmin, fire (default, verlet integrator, abcfire mode), and fix box/relax iso/aniso.

- `test_output_style` tests the *output data* of computes and fixes: the test YAML defines a compute or fix with ID `test` (plus helpers like groups, chunks, or feeder computes) in post_commands; the driver runs 10 MD steps and then introspects the style's output flags and compares everything it provides - global scalar, global vector, global array (including variable-size per-chunk arrays), per-atom vector/array, and local vector/array.  The driver adds its own nve integrator (unless the test defines a time-integrating fix, e.g. fix rigid/small) so history-dependent styles report real data.  Tests cover the chunk family, temperature/pressure/energy variants, connectivity computes, local bonded data, profile computes (adf, stress/mop, stress/cartesian), the TALLY package, rigid-body computes, lattice analysis and diffraction on a metal system (ackland, centro, cna, entropy, hexorder, orientorder, sna/atom, voronoi, xrd, saed), aspherical particle computes, and reporting fixes (ave/time, ave/atom, ave/chunk, ave/histo(+weight), ave/correlate, numdiff, numdiff/virial).

Both testers reuse the shared force-styles YAML infrastructure and double as their own reference generators (`-u`/`-g`).  The shared test configuration gained three additive keys (`global_array`, `peratom_data`, `local_data`); all existing force-styles tests are unaffected.

**Related Issue(s)**

None.  Part of the pre-release effort to expand test coverage; same groundwork approach as #5082.

**Author(s)**

**Licensing**

The test driver code, the test YAML files, and this pull request were developed with the assistance of Claude (Anthropic) under human direction, with all code and test definitions reviewed by the author.

**Backward Compatibility**

No changes to LAMMPS itself; test infrastructure only.  The new YAML configuration keys are additive and ignored by the existing testers.

**Implementation Notes**

*Determinism:* the minimizer tests use a fixed iteration budget instead of convergence tolerances, since line searches are branchy and stopping-criterion timing would make references fragile.  All 12 minimizer variants produce distinct trajectories from the same start (verified), with identical initial energies as a built-in consistency check.  The Windows CI runs showed that the state after a fixed-iteration minimization is nevertheless not reproducible across compilers and math libraries: the branchy line searches amplify last-bit force differences into different but equally valid descent paths (positions and force norm off by up to ~90%, while the final energies still agree to ~1e-4 since all paths land in the same basin).  The minimizer checks are therefore restricted to path-stable observables: energies and box dimensions with epsilon 1e-3, the force norm as an upper bound, and no per-atom position references.  Corrupted-reference checks confirm broken minimizers are still detected.  The output tests read all data right after a 10-step run, which matches the tally state of the final thermo output.

*Verification:* behavioral-change detection was verified for both testers by pairing reference data with a modified scene (e.g. min_style cg references against an sd run: fails; compute group change: fails).  Reference data was additionally audited for degenerate (all-zero) values; one test (hbond/local) was re-tuned as a result.

*Known limitations (documented in the driver):* computes requiring per-atom energy or virial tallies during the run (pe/atom, stress/atom, centroid/stress/atom, heat/flux) are not supported yet and need follow-up driver support; compute angle/bond/dihedral/improper require hybrid bonded styles and were left out; fix vector needs variable-size global vector support; spin minimizers will be covered together with the planned SPIN test expansion.  The initial comparison epsilons (1e-3 for minimizers, 1e-8 for output data) may need per-test adjustment once CI exposes the tests to other platforms - they are set per YAML file.

*Documentation and validation:* both testers are documented in the force-styles section of `Developer_unittest.rst` (new subsections plus updates to the configuration key table).  The JSON schema `tools/json/force-style-test-schema.json` was extended for the new reference keys and testers, and also catches up with previously missing keys (`timestep`, the SPIN magnetic reference data) - all 714 reference files in `unittest/force-styles/tests` now validate with `check-jsonschema`.

**Post Submission Checklist**

All 98 tests pass serially and with `ctest -j 8` (about 25 seconds wall clock).  Reference data was generated on Fedora 43 / GCC 15 / x86-64.


athomps (approved):
Yes, minimization is notoriously non-deterministic, sensitive to precision, etc.  Also, hard to distinguish errors in logic from peculiarities of the descent path.  I am glad you were able to find an error threshold that seems to work well in practice.

Issue #5071, opened 2026-07-03
LAMMPS Testing Status


akohlmey:
This issue keeps track of the LAMMPS testing status for all test that are being run *after* a pull request has been merged.

Current status of the automated LAMMPS test runs and manual builds. Full details on the [test status website](https://lammps.github.io/lammps-test-results/).

This status table is updated in place (no notifications); a comment is posted only when new failures appear or failures are fixed - subscribe to this issue to be notified about regressions.

| Suite | Tests | Passed | Failed | Errors | Timeouts | Runtest | Skipped | Changes |
|---|---:|---:|---:|---:|---:|---:|---:|---|
| :x: Full Regression: serial | 688 | 449 | 96 | 1 | 0 | 118 | 24 | **+1 new**, 1 fixed |
| :x: Full Regression: parallel | 688 | 438 | 138 | 0 | 4 | 77 | 31 | 1 fixed, 1 newly out of time |
| :x: Full Regression: openmp (MPI+OpenMP) | 688 | 274 | 290 | 6 | 5 | 75 | 38 | 1 fixed |
| :x: Full Regression: kokkos (KOKKOS/OpenMP) | 540 | 181 | 209 | 66 | 4 | 54 | 26 |  |
| :x: KOKKOS/OpenMP Regression | 395 | 127 | 165 | 40 | 0 | 48 | 15 |  |
| :white_check_mark: Example Input Check | 719 | 0 | 0 | 0 | 0 | 613 | 106 |  |
| :white_check_mark: Unit Tests: kokkos-openmp-double | 652 | 652 | 0 | 0 | 0 | 0 | 0 |  |
| :white_check_mark: Unit Tests: kokkos-openmp-mixed | 652 | 652 | 0 | 0 | 0 | 0 | 0 |  |
| :white_check_mark: Unit Tests: kokkos-openmp-single | 652 | 652 | 0 | 0 | 0 | 0 | 0 |  |
| :white_check_mark: Unit Tests: kokkos-serial-double | 652 | 652 | 0 | 0 | 0 | 0 | 0 |  |
| :white_check_mark: Unit Tests: kokkos-serial-mixed | 652 | 652 | 0 | 0 | 0 | 0 | 0 |  |
| :white_check_mark: Unit Tests: kokkos-serial-single | 652 | 652 | 0 | 0 | 0 | 0 | 0 |  |
[...]

PR #5084, opened 2026-07-10
Force-style test additions: MOFFF, FEP, SPH, DPD-MESO, SPIN packages

akohlmey:
**Summary**

First batch of the force-style test YAML backfill (stage 1 of the unittest coverage plan): 22 new tests for previously untested pair, bond, angle, and kspace styles, plus fixes for five distinct bugs that the new tests uncovered.

New tests (verified against the freshly generated references, all suffix variants):

- MOFFF: pair `buck6d/coul/gauss/dsf`, `buck6d/coul/gauss/long`, angle `cosine/buck6d`
- FEP: pair `lj/cut/soft/gapsys`
- EXTRA-MOLECULE: pair `hbond/dreiding/lj/angleoffset`, `hbond/dreiding/morse/angleoffset`
- MISC: bond `special`
- MANYBODY: pair `local/density`
- SPIN/KSPACE: pair `spin/dipole/long`, kspace `ewald/dipole/spin`, `pppm/dipole/spin`
- SPH: `sph/heatconduction`, `sph/idealgas`, `sph/lj`, `sph/rhosum`, `sph/taitwater`, `sph/taitwater/morris` with a new `atom_style sph` test system (in.sph/data.sph)
- DPD-SMOOTH: `sdpd/taitwater/isothermal`
- DPD-MESO: `edpd`, `tdpd`, and `mdpd`+`mdpd/rhosum` (hybrid/overlay) with new edpd/mdpd/tdpd test systems

`check_tests.py`: pair 216 -> 234 of 323, bond 19 -> 20, angle 26 -> 27, kspace 21 -> 23; total missing 334 -> 312 of 746.

**Bug fixes (all found by the new tests)**

1. KOKKOS `buck6d/coul/gauss/dsf`: special_bonds scaled the entire damped Coulomb term instead of subtracting only the bare `1/r` part (up to 100% force error on molecular systems with the default special_bonds settings). Both KOKKOS buck6d variants also used the low-accuracy Abramowitz-Stegun erfc polynomial where the CPU versions use `MathSpecial::my_erfcx()`; they now use `Kokkos::erf()` (agrees with the CPU expression to 9e-16).
2. `pppm/dipole/spin`: wrong per-grid-point counts for the inherited grid comm pack/unpack functions (3/9/18 instead of 4/12/25) silently corrupted communicated ghost grid data in parallel; the per-atom energy/virial path called forward_comm on the never-allocated base class `gc` grid and segfaulted. The spin solvers also left the charge-channel accumulators (`q2` etc.) and `density_brick` uninitialized, making the chosen PPPM grid and the charge structure factor depend on uninitialized memory (nondeterministic forces up to NaN, verified with valgrind). Three dipole solvers also called the collective `error->all()` only on rank 0 (parallel hang).
3. `sph/lj`: the per-pair long-range pressure correction accumulated into the per-atom EOS pressure of atom i inside the neighbor loop, so forces depended on the number and order of neighbor list entries (completely different results for newton on vs. off, even in serial). The GPU kernels replicated the defect and are fixed identically.
4. `sdpd/taitwater/isothermal`: the RNG seed included `atom->nlocal` at `pair_style` time, so the random stream depended on whether the pair style is defined before or after the atoms exist.
5. DPD-MESO data file and restart support: `edpd`/`tdpd` set `writedata` but had no `write_data()` implementations (write_data produced an unreadable empty section), `mdpd` omitted two mandatory columns, and reading an `edpd` restart crashed (uninitialized/unsaved `power_flag`/`kappa_flag`, unallocated optional tables). Note: this changes the (version-specific) restart format for `edpd`.

**Infrastructure**

- new `skip_tests` token `nofdotr` for pair styles whose dissipative/random forces are non-central and cannot reproduce the fdotr virial through `ev_tally()` (used by the sdpd test)
- new ctest glob blocks for the `sph-pair-*` and `meso-pair-*` test file prefixes
- not testable with this harness: `pppm/rk` (requires the multi-partition `verlet/split/rk` integrator); `dsmc` (stochastic, no forces); `agni` (force-only ML potential whose regression kernel sum amplifies last-ulp exp() differences by many orders of magnitude, so no portable reference tolerance exists). DRUDE-dependent styles (`thole`, `lj/cut/thole/long`, `coul/tt`) are deferred to the DRUDE test batch.

**Author(s)**

Axel Kohlmeyer (Temple U)

**Licensing**

**Backward Compatibility**

The edpd restart file format changes (restart files are version-specific in LAMMPS). The sph/lj force fix changes trajectories of all sph/lj runs; the previous forces depended on neighbor list order and were not reproducible between newton settings.

**Implementation Notes**

All new reference data was generated on Fedora 43 / GCC and verified with repeated runs, including the KOKKOS (OpenMP) suffix runs. The full force-styles suite (733 tests) passes locally.

**AI Tools Usage**

The test files, the bug analyses, and the bug fixes in this PR were developed with the help of Claude Code (Fable 5); all changes were verified by building and running the affected tests and tools, including valgrind verification for the memory-related fixes.


akohlmey:
@athomps this is the next step in expanding the test coverage.
This will also re-enable Trung's full regression tests on a regular basis.
The new testing dashboard should make checking the results easier.
It also create a proper notification for a new failed post-merge unittest:
https://github.com/lammps/lammps/issues/5071

All the bits start to fall into their places... :grinning:

sjplimp (approved):

PR #5086, opened 2026-07-11
Add support for neigh/cluster for the KOKKOS package

ndtrung81:
**Summary**

This PR adds an option for KOKKOS to build neighbor lists for clusters of atoms and compute pair forces with this data layout.

**Related Issue(s)**

None

**Author(s)**

Trung Nguyen (U Chicago)

**Licensing**

By submitting this pull request, I confirm that AI tools were used to generate parts of the code and modifications in this pull request: the implementation, documentation, tests, and verification runs were produced with Claude Code (Anthropic) Opus 4.8 and Fable 5 under the author's direction and review. 

**Backward Compatibility**

The changes in this PR should maintain backward compatibility.

**Implementation Notes**

Summary of implementation and validation is included in the conversation below.

**Post Submission Checklist**


ndtrung81:
# LAMMPS KOKKOS Neighbor List & Pair Force Analysis
inspired by GROMACS-style cluster-cluster scheme for high neighbor counts. 

The lengthy document below is for future reference and keeping track of the *thought* process of the AI coding agent. A more concise version of summary will be created later.

---

## 1. Current Architecture

### 1.1 Neighbor List Memory Layout

`NeighListKokkos` (`src/KOKKOS/neigh_list_kokkos.h:77`) stores neighbors in a 2D Kokkos View:

```
d_neighbors[nmax_atoms][maxneighs]   — LayoutLeft (column-major on CUDA)
d_numneigh[nmax_atoms]
d_ilist[nmax_atoms]
```

The critical problem: **LayoutLeft** means atom `i`'s neighbors are stored at
`base + i + jj*stride`. Iterating over `jj = 0..699` for atom `i` means striding through
memory at intervals of `nmax_atoms * sizeof(int)`. This is cache-poisoning for 500–700
neighbors.

There is a `d_neighbors_transpose` with LayoutRight (row-major), enabled via
`lmp->kokkos->neigh_transpose`. This helps somewhat, but the problem goes deeper than
layout alone.

### 1.2 Force Kernel Structure (`pair_kokkos.h:293–362, 895–983`)

```
League size = num_I_atoms        (one team per atom i)
Team size   = AUTO (32–128)
Vector size = lane count

for each team:
  i = d_ilist[league_rank]
  xtmp = x(i, 0..2)              # 3 loads
  parallel_reduce(ThreadVectorRange, jnum):
    j = neighbors_i(jj)          # strided load
    delx = xtmp - c.x(j, 0)      # random j → scatter read
    ...
    a_f(j, 0) -= fx              # atomic scatter write
```

With 600 neighbors per atom, each of the 32 warp threads requests a **different** `j` atom's
coordinates, which are scattered in memory — effectively 32 separate cache-line fetches per
neighbor iteration. On A100/H100 with 128-byte L2 lines, ~96 bytes are wasted per 32 bytes
used.

### 1.3 Neighbor List Building (`npair_kokkos.cpp:244–829`)

The GPU build kernel (`build_ItemGPU`) uses a bin-based scheme with shared memory staging:

- `BINS_PER_BLOCK = 2`, one team per pair of bins
- Loads neighbor bin atoms into shared memory (`5 * atoms_per_bin` doubles: x, y, z, type, tag)
- Inner loop with `#pragma unroll 4/8` to hide latency

This part is reasonably efficient — the problem is the **resulting flat neighbor list**, not
its construction.

### 1.4 Position and Force Data Layout

**Positions** (`pair_lj_cut_kokkos.h:80–82`):
```cpp
typename AT::t_kkfloat_1d_3_lr_randomread x;   // LayoutRight: x(i,0..2) contiguous
```

**Forces** (`pair_lj_cut_kokkos.h:82`):
```cpp
typename AT::t_kkacc_1d_3 f;   // LayoutLeft by default: f(0,0..2) NOT contiguous per atom
```

---

## 2. Why This Breaks at 500–700 Neighbors

Three compounding effects:

### 2.1 Scattered Coordinate Reads

`c.x(j, 0)` for 32 warp threads with random `j` → 32 L2/DRAM transactions instead of 1.
At 600 neighbors, the inner loop executes ~600 iterations per atom and almost none of the
`j`-atom data is hot in L1/L2 by the time it's needed.

### 2.2 Scattered Atomic Force Writes

`a_f(j, 0) -= fx` is similarly scattered across random `j` atoms → serialized atomic
operations, high retry rates, and bus saturation.

### 2.3 Work Imbalance

One team per atom; atoms near surfaces or in dense regions have 10× more work than bulk
atoms in dilute regions. The `ThreadVectorRange` reduction over `jnum` can't rebalance
across teams, so the warp with 700 neighbors stalls the scheduler while lighter warps sit
idle.

**Net effect**: At 500+ neighbors, memory bandwidth is the hard wall. Each LJ pair is ~10
FLOPs but requires loading ~30 bytes of data that cannot be reused across warps with the
current one-team-per-atom mapping.

---

## 3. What GROMACS Does (NBNXM Cluster Scheme)

GROMACS groups atoms into **i-clusters** of 4 atoms (CUDA) or 8 atoms (SIMD AVX-512). The
neighbor list becomes a **cluster-pair list**:

```
i-cluster 0 → [j-cluster 5, j-cluster 12, j-cluster 13, ...]
i-cluster 1 → [j-cluster 5, j-cluster 7, ...]
```

Coordinates are packed in **SOA (Structure of Arrays) within each cluster**:

```
x_packed = [x0,x1,x2,x3, x4,x5,x6,x7, ...]   # all x-coords of cluster 0, then cluster 1
y_packed = [y0,y1,y2,y3, y4,y5,y6,y7, ...]
z_packed = ...
```

One warp processes **one (i-cluster, j-cluster) pair = 4×4 = 16 interactions**:

1. Load 4 i-atom coords → **4 contiguous reads, broadcast to warp** (48 bytes, fits in
   one cache line)
2. Load 4 j-atom coords → **4 contiguous reads, broadcast to warp**
3. Compute all 16 force pairs **entirely in registers**
4. No j-scatter problem: all 32 threads work on the same 16 atom pairs

**Cache reuse**: each cluster's 4×3 = 12 doubles (96 bytes, < 1 cache line) is loaded once
and reused for all `Cj` j-cluster interactions for that i-cluster.

---

## 4. Concrete Proposal for LAMMPS KOKKOS

### 4.1 Step 1 — Extend `NeighListKokkos` (`neigh_list_kokkos.h:69`)

Add the following cluster-specific fields alongside the existing `d_neighbors`:

```cpp
// Cluster-based neighbor list (opt-in, parallel to existing flat list)
int cluster_size;              // 4 (CUDA) or 8 (SIMD); default 4
int num_clusters;              // ceil(nlocal / cluster_size)

AT::t_int_1d   d_cluster_atom_offset;  // first atom index of cluster ci  [num_clusters]
AT::t_int_1d   d_cluster_numneigh;     // number of j-clusters for ci     [num_clusters]
AT::t_int_2d   d_cluster_jlist;        // j-cluster IDs                   [num_clusters][max_jclusters]
AT::t_int_1d   d_cluster_excl_mask;    // exclusion bitmask for self-pairs (optional)

// Packed SOA coordinates in cluster order
AT::t_kkfloat_1d  d_xc, d_yc, d_zc;   // size: num_clusters * cluster_size
AT::t_int_1d      d_typec;             // atom types in cluster order

void grow_clusters(int nclusters);
void pack_clusters();
```

**Memory sizing**: With cluster_size=4 and N=100k atoms:
- `d_cluster_jlist`: ~25k clusters × ~150 j-clusters = 3.75M integers ≈ 15 MB (acceptable)
- `d_xc/yc/zc`: 100k floats × 3 = 1.2 MB (tiny)

### 4.2 Step 2 — Post-Build Clustering Kernel (`npair_kokkos.cpp`)

After the existing `build_ItemGPU` populates `d_neighbors`, add a second pass in a new
`NPairKokkosCluster` class:

```cpp
template<class DeviceType>
struct ClusterBuildFunctor {
  // inputs: d_neighbors, d_numneigh, x, type
  // outputs: d_cluster_jlist, d_cluster_numneigh, d_xc, d_yc, d_zc, d_typec

  KOKKOS_FUNCTION
  void operator()(const int ci) const {
    int first_i = ci * cluster_size;

    // Pack i-cluster coordinates (SOA)
    for (int ki = 0; ki < cluster_size; ki++) {
      int i_atom = first_i + ki;
      d_xc(ci * cluster_size + ki) = x(i_atom, 0);
      d_yc(ci * cluster_size + ki) = x(i_atom, 1);
      d_zc(ci * cluster_size + ki) = x(i_atom, 2);
      d_typec(ci * cluster_size + ki) = type(i_atom);
    }

    // Collect unique j-clusters from all atoms in this i-cluster
    // Use a small per-thread hash set in registers (max ~256 j-clusters in practice)
    int jclusters[256];
    int njc = 0;

    for (int ki = 0; ki < cluster_size; ki++) {
      int i_atom = first_i + ki;
      int jnum = d_numneigh(i_atom);
      for (int jj = 0; jj < jnum; jj++) {
        int j_atom = d_neighbors(i_atom, jj);
        int cj = j_atom / cluster_size;
        // Insert cj if not already in jclusters[]
        bool found = false;
        for (int k = 0; k < njc; k++) {
          if (jclusters[k] == cj) { found = true; break; }
        }
        if (!found) jclusters[njc++] = cj;
      }
    }

    d_cluster_numneigh(ci) = njc;
    for (int k = 0; k < njc; k++)
      d_cluster_jlist(ci, k) = jclusters[k];
  }
};
```

**Notes**:
- The deduplication hash set can be done more efficiently with a bitmask if cluster IDs are
  bounded, or with shared memory in a team-policy variant.
- At 600 neighbors and cluster_size=4, `njc` ≈ 150 j-clusters per i-cluster on average —
  well within a register array.
- This kernel runs once per neighbor list rebuild, so its cost is amortized.

### 4.3 Step 3 — Cluster Force Kernel

Replace (or supplement) the existing `pair_compute` loop in `pair_kokkos.h`:

```cpp
// One warp per i-cluster
TeamPolicy: league_size = num_clusters, team_size = 32

KOKKOS_FUNCTION
void operator()(const member_type& team, EV_FLOAT& ev) const {

  int ci = team.league_rank();
  int first_i = d_cluster_atom_offset(ci);

  // Load i-cluster coords into shared memory (48 bytes)
  // All 32 threads cooperate; only first 4 do useful work, rest are idle here
  ScratchView<KK_FLOAT*> s_xi(team.team_scratch(0), cluster_size);
  ScratchView<KK_FLOAT*> s_yi(team.team_scratch(0), cluster_size);
  ScratchView<KK_FLOAT*> s_zi(team.team_scratch(0), cluster_size);

  Kokkos::single(Kokkos::PerTeam(team), [&]() {
    for (int ki = 0; ki < cluster_size; ki++) {
      s_xi(ki) = d_xc(first_i + ki);
      s_yi(ki) = d_yc(first_i + ki);
      s_zi(ki) = d_zc(first_i + ki);
    }
  });
  team.team_barrier();

  int cj_count = d_cluster_numneigh(ci);

  // Iterate over j-clusters
  for (int cj_idx = 0; cj_idx < cj_count; cj_idx++) {
    int cj = d_cluster_jlist(ci, cj_idx);
    int first_j = cj * cluster_size;

    // Load j-cluster coords into shared memory
    ScratchView<KK_FLOAT*> s_xj(team.team_scratch(0), cluster_size);
    // ... similarly s_yj, s_zj

    Kokkos::single(Kokkos::PerTeam(team), [&]() {
      for (int kj = 0; kj < cluster_size; kj++) {
        s_xj(kj) = d_xc(first_j + kj);
        // ...
      }
    });
    team.team_barrier();

    // Distribute 4×4 = 16 pairs across 32 threads (2 pairs per thread)
    Kokkos::parallel_for(Kokkos::TeamThreadRange(team, cluster_size * cluster_size),
      [&](const int pair_idx) {

        int ki = pair_idx / cluster_size;
        int kj = pair_idx % cluster_size;
        int i_atom = first_i + ki;
        int j_atom = first_j + kj;

        KK_FLOAT delx = s_xi(ki) - s_xj(kj);
        KK_FLOAT dely = s_yi(ki) - s_yj(kj);
        KK_FLOAT delz = s_zi(ki) - s_zj(kj);
        KK_FLOAT rsq  = delx*delx + dely*dely + delz*delz;

        int itype = d_typec(first_i + ki);
        int jtype = d_typec(first_j + kj);

        if (rsq < d_cutsq(itype, jtype)) {
          KK_FLOAT fpair = compute_fpair(rsq, i_atom, j_atom, itype, jtype);
          KK_FLOAT fx = delx * fpair;
          KK_FLOAT fy = dely * fpair;
          KK_FLOAT fz = delz * fpair;

          Kokkos::atomic_add(&f(i_atom, 0),  fx);
          Kokkos::atomic_add(&f(i_atom, 1),  fy);
          Kokkos::atomic_add(&f(i_atom, 2),  fz);
          Kokkos::atomic_add(&f(j_atom, 0), -fx);
          Kokkos::atomic_add(&f(j_atom, 1), -fy);
          Kokkos::atomic_add(&f(j_atom, 2), -fz);
        }
    });
    team.team_barrier();
  }
}
```

**Key properties:**

- All 32 warp threads access the **same 4 i-coords and 4 j-coords** from shared memory →
  zero main-memory coordinate traffic during force computation
- `d_cluster_jlist(ci, cj_idx)` accesses are sequential per i-cluster → coalesced L2 reads
- With cluster_size=4, force writes target only 4 unique j-atoms per j-cluster → far less
  atomic contention vs. 600 random atoms

---

## 5. Memory Traffic Analysis

| Access type | Current (500 nbrs/atom) | Cluster scheme (C=4) |
|---|---|---|
| Coord reads per atom | 500 × 12 B = 6 KB | 125 j-clusters × 48 B = 6 KB |
| L2 transactions (scatter factor) | ~500 (fully scattered) | ~125 (sequential cluster loads) |
| Cache line efficiency | ~15–20% | ~90%+ |
| Force write atomics | 500 per atom (scattered) | 4 × 125 = 500 (spatially local) |
| Warp divergence | High (cutoff check varies per j) | Low (all 16 pairs same control flow) |

Total bytes transferred is the same; **effective bandwidth utilization** goes from ~15% to
~90% because each cache line is fully used instead of partially.

---

## 6. Implementation Challenges

### 6.1 Ghost Atoms (Hardest)

Clusters must span periodic boundaries correctly. Ghost atoms need to be pre-packed into
cluster order. The clustering pass must happen **after** ghost exchange but **before** the
force kernel. This ties into the existing `forward_comm` / `exchange` ordering in
`Verlet::run()`.

Options:
- Cluster only local atoms; ghost atoms remain in flat list and are accessed individually
  (simplest, partial benefit)
- Pre-cluster ghost atoms using a separate ghost-cluster map (full benefit, more complex)

### 6.2 Half-List Newton 3rd Law

The 4×4 block may double-count pairs if both `i_atom < j_atom` and `j_atom < i_atom`
cluster pairs appear in the list. Need an **exclusion bitmask** (16 bits per cluster-pair)
to mask out already-counted interactions when `ci > cj`.

### 6.3 Mixed Species / Per-Type Cutoffs

When per-type cutoffs vary significantly, some of the 16 pairs in a 4×4 block will be
inactive. The bitmask handles correctness but reduces the "uniform control flow" benefit
for polydisperse systems. For monatomic or two-component systems this is not an issue.

### 6.4 Force Array Layout

The current `f(i, 0)` LayoutLeft layout means `f(i,0)`, `f(i,1)`, `f(i,2)` are NOT
contiguous for any atom `i`. Switching to **LayoutRight** (SOA: all `f_x` values first,
then all `f_y`, then all `f_z`) would improve coalescing for atomic adds from cluster
threads writing to 4 nearby atoms.

### 6.5 Integration with NPair Framework

The cleanest path is a new `NPairKokkosCluster` class that:
1. Calls the existing `NPairKokkos::build()` to get the flat `d_neighbors`
2. Runs `ClusterBuildFunctor` as a post-pass to populate `d_cluster_jlist` etc.

This avoids touching the existing half/full/trim npair class hierarchy and allows fallback
to the flat list for small systems where cluster overhead is not justified.

---

## 7. Expected Performance

**Current bottleneck profile** (500–700 neighbors/atom, flat list):
- ~50% warp efficiency (scattered access, variable `jnum` imbalance)
- ~20–30% L2 cache utilization (random `j` atoms, cold data)
- Compute-to-memory ratio: heavily memory-bound

**With cluster scheme** (C=4):
- ~95% warp efficiency (all threads do identical work per cluster-pair)
- ~90% L2 cache line utilization (sequential cluster reads, shared-mem reuse)
- Effective memory bandwidth: 3–4× improvement
- **Expected end-to-end force kernel speedup: 1.8–2.5× for 500–700 neighbor regime**

**Low neighbor count regime** (<50 neighbors):
- Cluster overhead per i-cluster is small (~150 ns setup), amortized over fewer pairs
- Speedup is smaller (~1.1–1.3×) but the scheme remains correct
- Could auto-select flat vs. cluster kernel based on average `jnum` at runtime

---

## 8. Phased Implementation Roadmap

NOTE (2026-07-11): this table is the ORIGINAL pre-implementation plan; its
phase numbers do NOT match the session-log phase names (sections 10-12,
where Phase 0 = Morton sort, Phase 2 = cluster kernel, Phase 3 =
cluster-level Newton). Status added below; do not schedule work off the
stale rows.

| Phase | Work | Files | Status |
|---|---|---|---|
| 1 | Add cluster fields to `NeighListKokkos` | `neigh_list_kokkos.h` | DONE |
| 2 | Implement `ClusterBuildFunctor` | `npair_kokkos.cpp` | DONE |
| 3 | Implement cluster force kernel | `pair_kokkos.h` | DONE |
| 4 | Ghost atom cluster packing | `comm_kokkos.cpp` | SUPERSEDED: locals-first Morton order + per-step packed gather into `d_xcl` gives coalesced j-reads without touching comm; revisit only if the gather shows up in an A100 profile |
| 5 | Newton 3rd law exclusion masks | `pair_kokkos.h`, `npair_kokkos.cpp` | DONE = session-log Phase 3 (ownership zig-zag + i<j self-tile masks + exclusion bits), validated in 12.4 |
| 6 | Validation vs. flat list | test suite | DONE (4-deck flat-vs-cluster + 500-step melt, 12.4) |
| 7 | Tuning: cluster_size, force layout | various | OPEN -- do on A100, not GP100 |

Current next steps (ranked, from measurements in 12.4): (1) single/mixed
precision; (2) builder consumes bins directly instead of the full flat
list (371 ms vs 137 ms per rebuild); (3) re-profile on A100 before any
shape/team-size tuning.

---

## 9. Key File References

| File | Relevant Lines | Content |
|---|---|---|
| `src/KOKKOS/neigh_list_kokkos.h` | 69–123 | `NeighListKokkos` data structure, View types |
| `src/KOKKOS/npair_kokkos.cpp` | 244–312 | Kernel launch configuration |
| `src/KOKKOS/npair_kokkos.cpp` | 600–829 | `build_ItemGPU` shared-memory bin kernel |
| `src/KOKKOS/pair_kokkos.h` | 293–362 | Per-atom force kernel inner loop |
| `src/KOKKOS/pair_kokkos.h` | 895–983 | Force kernel launch (TeamPolicy setup) |
| `src/KOKKOS/pair_lj_cut_kokkos.h` | 80–82 | Position/force View declarations |
| `src/KOKKOS/kokkos_type.h` | 1274–1286 | `t_neighbors_2d` and related typedefs |
| `src/neigh_list.h` | 14–119 | Base `NeighList` class |

---

## 10. Session Findings (2026-07-10): Correctness Fixes and True Profile

### 10.1 True GPU time breakdown for the melt benchmark

The LAMMPS section timers ("Pair 0.5%, Bond 60%") are misleading on GPU: kernel
launches are asynchronous and the first blocking operation in the bonded path
(`k_warning_flag.sync_host()` in `dihedral_opls_kokkos.cpp:130`, executed every
step) absorbs the pair kernel's wait time into the Bond section. `timer sync`
does not help (it only adds MPI barriers, no GPU fences).

nsys profile of `bench/tuning/in.melt` (278,704 atoms, lj/cut/coul/long 12 12 +
PPPM, 755 neighs/atom half list, GP100, double precision, binsize 8.0):

| GPU work (per step, amortized)     | time    | share |
|------------------------------------|---------|-------|
| Pair force kernel (flat, half)     | 32.9 ms | 58.8% |
| Neighbor build (137 ms per build)  |  6.8 ms | 12.4% |
| PPPM (two dominant kernels)        |  7.8 ms | 14.3% |
| Dihedral + Angle + Bond            |  1.9 ms |  3.5% |

Bonded kernels are cheap; the pair kernel is the wall, exactly the regime the
cluster scheme targets. A large block of `ViewCopy` kernels (~311 ms of int**
copies) turned out to be one-time setup cost (read_data/topology growth), not
loop time.

### 10.2 Cluster-kernel correctness bugs found and fixed

The "validated bit-for-bit" state did not hold for the current tree. Re-running
`bench/in.cluster_validate_special` gave E_pair -9.92 vs -0.634 (flat). Three
root causes, all stemming from the kernel treating "within cutoff" as "in the
flat list":

1. **Half lists double count.** A pair stored only at i is also computed from
   j's tile (geometric check passes), and the duplicate side lacks the sbmask
   bits, so bonded 1-2 pairs got full LJ at bond distance (melt: step-0
   pressure 3e8 instead of 759).
2. **Deleted special pairs reappear.** With `special_lj 0.0` and no kspace,
   1-2 pairs are removed from the flat list entirely; the geometric check
   computed them at full strength (sm defaults to 0 = factor 1.0).
3. **Coul EV path was a stub** delegating to the NoCoul EV kernel: on every
   energy/virial step (including setup), coul styles computed LJ-only forces.

**Fix: per-cluster-pair presence mask** (`d_cluster_pres(ci, cj_idx)`, bit p =
pair p is a flat-list entry), built by the ClusterBuildFunctor with one shared
mem atomic OR per entry. Tile membership now mirrors the flat list exactly
(half-list storage, special deletions, ghost ownership, self pairs). Phase D
j-updates are gated on the presence bit too, skipping atomic adds of zero.
Also implemented the real Coul EV cluster kernel (evdwl/ecoul split, flat-list
energy scale semantics incl. the newton-off ghost-j 0.5 rule), and fixed the
NoCoul EV global-energy scale for that same corner.

The dedup-hash is now runtime-sized (`cluster_hash_sh`, persisted in
`NeighListKokkos`, doubling 1024 -> 2048 on saturation; 2048 is the 48 KB
per-block scratch ceiling). The melt at 755 neighs/atom saturates 1024 slots.
The cluster-list rebuild trigger moved from a function-local static to
`NeighListKokkos::cluster_built_step` (robust across run resets / multiple
pair styles).

All four validation decks now match the flat kernel digit-for-digit in thermo
output: special half+newton, special full, plain-LJ half+newton, plain-LJ full.

### 10.3 Structural concern for half lists at high neighbor counts

With a half list at 755 neighs/atom, the union of 8 i-atoms' lists (~6000
entries) spreads over >1024 unique j-clusters, i.e. average tile occupancy is
only ~12-18%. The cluster list (4 ints per tile) then approaches the flat
list in size and the traffic advantage largely evaporates. The dense-tile
regime that makes GROMACS fast requires cluster-level Newton over a full-style
list: emit each unordered (ci, cj) tile once (cj/2 >= ci rule), compute all
present pairs, apply forces to both sides (i in registers, j atomically), with
an i<j mask inside self-overlap tiles. That keeps interior tiles ~100% dense
and halves the tile count. This is the recommended Phase 3 design.

### 10.4 Melt benchmark results after the fix (GP100, double, 500 steps)

Thermo matches the flat run to 5-6 digits over 500 steps (FP summation-order
drift only). Wall clock:

| config                            | Loop time | pair kernel | notes |
|-----------------------------------|-----------|-------------|-------|
| flat, half + newton on            |  25.93 s  | 32.9 ms/step | baseline (best) |
| cluster, half + newton on         |  31.24 s  | 36.3 ms/step | +10% kernel, +~5 ms/step cluster build |
| flat, full + newton off           |  41.06 s  |     —       | half+newton clearly wins here |
| cluster, full + newton off        |  aborts   |     —       | >2048 unique j-clusters/i-cluster |

Reading:
- The cluster pair kernel reaches near parity with flat despite only ~15%
  tile occupancy — coalesced coordinate loads nearly cancel the ~6-8x pair
  compute amplification even in double precision. Denser tiles (Phase 3
  cluster-level Newton) and/or mixed precision should tip it to a win.
- The cluster build post-pass costs ~96 ms per reneighbor (~70% of the flat
  build itself). Root cause: one 32-thread warp per team with 16-32 KB scratch
  -> 2-4 teams/SM -> latency-bound. TEAM_SIZE raised 32 -> 128 (4 warps share
  the hash) to hide latency; result pending.
- Capacity: full lists at 1510 neighs/atom exceed HASH_MAX=2048 unique
  j-clusters per i-cluster. Needs CJ=8 tiles or a global-memory fallback if
  full-list support at this cutoff matters.

### 10.5 Builder TEAM_SIZE 32 -> 128 (final state of this session)

Raising ClusterBuildFunctor TEAM_SIZE to 128 (4 warps cooperating on one
i-cluster's shared-mem hash) cut the cluster build post-pass from 96 to 40 ms
per reneighbor (2.4x) with identical results (all validation decks still match
digit-for-digit). Melt totals, 500 steps GP100 double:

| config                     | Loop time |
|----------------------------|-----------|
| flat, half + newton        |  25.93 s  |
| cluster (TEAM_SIZE=32)     |  31.24 s  |
| cluster (TEAM_SIZE=128)    |  29.28 s  |

Remaining gap vs flat = +3.1 ms/step pair kernel (36.0 vs 32.9) + ~2 ms/step
amortized cluster build. Next steps ranked:
1. **Phase 3 cluster-level Newton** (dense tiles; see 10.3) -- the pair kernel
   is at parity with ~15% occupancy, so ~3-4x more present pairs per tile is
   the big lever.
2. **Mixed/single precision runs** -- the flop amplification the sparse tiles
   cause is twice as cheap there; the A100 mixed-precision estimates in the
   memory notes came from that regime.
3. Cluster-build capacity for full lists (CJ=8 tiles or global-mem hash
   fallback past 2048 slots).
4. PPPM (7.8 ms/step) and the flat neighbor build (137 ms/build) are the next
   biggest non-pair items on this benchmark.

## 11. Session Findings (2026-07-10, later): j-Side Cluster Ordering Rerun + Profile

### 11.1 What was tested

The working-tree change on top of the committed presence-mask state: a
dedicated j-side cluster ordering. At each reneighbor, all nall atoms
(locals + ghosts) are Morton-sorted on a fine grid (cell ~ cutneigh/8,
capped at 1024 cells/axis) via `Kokkos::Experimental::sort_by_key`;
`d_cl2atom` / `d_atom2cl` map between atom index and cluster slot. The
ClusterBuildFunctor maps j through `d_atom2cl` so j-clusters are groups of
4 spatially compact slots (ghosts included -- comm order had left ghosts
spatially scattered). Every step, a gather kernel packs `d_xcl` /
`d_typecl` in cluster order so tiles read coalesced coordinates; the force
kernels take j-coords from the packed arrays and route force/charge
accesses through `s_jatom = d_cl2atom(jcl)`.

### 11.2 Two bugs fixed before the rerun

1. **`sort_by_key` extents crash at the first mid-run reneighbor.**
   `grow_cluster_order` keeps `d_cl2atom` at its high-water extent, but the
   keys view is allocated at the current nall; when nghost shrank on a
   rebuild (502592 -> 502570), Kokkos aborted with "values and keys extents
   must be the same". Fix: sort a `Kokkos::subview(cl2atom, {0, nall})`.
   Note: both validation decks are 1-step (`every 100 check no`) and never
   reneighbor mid-run, so they cannot catch this class of bug; the 500-step
   melt run is the effective regression test for the resort path.
2. **Builder retry every rebuild.** On jlist overflow the code set
   `max_jclusters` to the exact needed count; per-rebuild jitter then forced
   a full ~50 ms retry on subsequent rebuilds during equilibration-driven
   growth. Fix: grow with +12.5% headroom. Builder launches for the 100-step
   profile dropped 9 -> 7; steady state is one ~52 ms launch per rebuild.

Validation after both fixes: all four decks (plain LJ and special_bonds,
half+newton and full+newtoff) match the flat kernel digit-for-digit;
500-step melt thermo matches flat to 6 digits.

### 11.3 Measured occupancy (new instrumentation)

`LMP_CLUSTER_STATS=1` env var enables a per-rebuild reduction printing tile
count and present-pair popcount. Melt, half+newton:

```
CLUSTER_STATS: iclusters 34838 tiles 14147684 pairs 210473093 occ 0.4649 avg 406.1 max_jc 1101
CLUSTER_STATS: iclusters 34838 tiles 14833709 pairs 210477508 occ 0.4434 avg 425.8 max_jc 1390
```

**Occupancy is 44-46%, not the 12-18% previously estimated.** For a HALF
list this is close to the ceiling: half-list storage holds ~half of each
tile's geometric pairs, so ~50% is the structural maximum (boundary tiles
cost a bit more). The tiles are already essentially as dense as a half
list permits.

### 11.4 Profile (nsys, 100 steps, GP100 double)

| GPU work                          | this session | prior (ts128) | flat  |
|-----------------------------------|--------------|---------------|-------|
| pair kernel (ms/step)             | 36.8         | 36.3          | 32.9  |
| flat neigh build (ms/build)       | 137          | 137           | 137   |
| ClusterBuild (ms/rebuild, steady) | ~52          | ~40-100       | --    |
| cluster gather (ms/step)          | 0.13         | --            | --    |
| Morton sort_by_key (ms/rebuild)   | ~0.1         | --            | --    |

500-step melt wall clock: flat 25.93 s | cluster ts128 29.28 s | cluster
+ j-ordering 29.07 s. The j-side ordering is correctness-clean and nearly
free, but bought only ~0.2 s.

### 11.5 Reading: per-tile cost is occupancy-insensitive; tiles are the lever

A warp executes the union of its 32 lanes' control flow, so a tile costs
roughly the same whether 15 or 32 pairs are present: kernel time scales
with TILE COUNT, not with present pairs. The j-ordering did not reduce
tile count materially (kernel 36.3 -> 36.8 ms/step), consistent with
occupancy already being near the half-list ceiling before the change.

Ceiling arithmetic: 210.5M pairs / 32 = 6.58M tiles at occ 1.0 vs 14.1M
measured -- a 2.15x kernel-time reduction is the maximum available, and it
is exactly what cluster-level Newton over a full-style list targets: each
unordered (ci,cj) tile emitted once holds BOTH directions' pairs (occ ~90%
interior), tile count ~7M, same total force computes and j-atomics as the
half list. Projected pair kernel ~19-22 ms/step vs flat 32.9 -- the first
configuration where cluster beats flat in double precision.

### 11.6 Next steps (ranked)

1. **Phase 3 proper: cluster-level Newton over a full-style list** (emit
   unordered tiles once, cj/2 >= ci; i<j mask in self-overlap tiles; +f in
   registers / -f atomic). Projected ~1.5-1.7x vs flat on the pair kernel.
   The j-side ordering infrastructure from this session (cl2atom/atom2cl,
   packed gather) is a prerequisite and is now in place and validated.
2. **Mixed/single precision** on top -- halves the per-tile compute cost.
3. i-side fine ordering (i-clusters from the same Morton order restricted
   to locals) -- likely small now that measured occupancy is near ceiling,
   but folds naturally into the Phase 3 builder rewrite.
4. ClusterBuild post-pass (52 ms/rebuild = ~2 ms/step amortized) and the
   flat build (137 ms) matter more once the pair kernel drops under
   ~20 ms/step.

## 12. Session Findings (2026-07-11): Phase 3 Cluster-Level Newton -- Implemented, Validation In Progress

### 12.1 What was implemented (working tree, uncommitted)

Cluster-level Newton over a FULL-style flat list, replacing the per-side
half-list tiles whenever the pair style runs `neigh half`/`HALFTHREAD` +
`newton on` + `neigh/cluster yes`. Each unordered (ci,cj) tile is emitted
once and holds BOTH directions' pairs (~2x occupancy, ~half the tiles);
the per-pair force/EV math and the Phase D `-f` atomic scatter were already
pair-exact and are unchanged.

Pieces:

- **Locals-first cluster order** (`build_cluster_order`): bit 30 above the
  30-bit Morton key sorts all ghosts after all locals, so slots
  `[0, nlocal)` are exactly the locals. i-clusters in Newton mode are
  cluster-order slot blocks of 8 (this also delivers the old "i-side fine
  ordering" idea for free), and every local-ghost pair is guaranteed to be
  seen from the local side.
- **Builder ownership filter** (`ClusterBuildFunctor`, `newton_cluster=1`,
  consumes a FULL flat list where every pair appears from both sides):
  - local j: same 8-block -> keep upper triangle (`jcl > ii`); different
    blocks -> **parity zig-zag** (even block sum -> lower block keeps, odd
    -> higher). NOT the naive `jblock >= ci` rule: that would make
    Morton-early i-clusters keep their whole interaction ball (~2x the
    measured half-list max of 1390 unique j-clusters), overflowing the
    2048-slot shared-mem dedup hash. Zig-zag keeps ~half the ball per
    i-cluster and load-balances tile counts across teams.
  - ghost j: the exact "above and to the right" (z,y,x) lexicographic rule
    from the flat half+newton build (npair_kokkos.cpp build_item), applied
    on `d_xcl` coords. THIS IS LOAD-BEARING: with a full list, a
    local-ghost pair is listed on both owning ranks (and both PBC images
    in serial); an index- or slot-based rule cannot be globally
    consistent, and getting it wrong double-counts every cross-boundary
    pair after reverse comm.
- **Force kernels**: only change is the i-atom source in Newton mode --
  `i = d_cl2atom(slot)` with coords/types read from packed `d_xcl`/
  `d_typecl` (coalesced) instead of `d_ilist` + scattered `c.x`.
- **Plumbing**: `pair lj/cut/kk` and `lj/cut/coul/long/kk` (melt deck
  style) morph their half request to `enable_full()` in `init_style` when
  cluster+newton+GPU is active. Dispatch detects the morph via a new
  `NeighRequest::get_full()` accessor and falls back to the legacy
  per-side path for styles that did not opt in (eam etc. stay correct).
  Gather now runs BEFORE the cluster build on rebuild steps (the ghost
  rule reads `d_xcl`). Mode is stamped in the list
  (`cluster_newton_built`) alongside `cluster_built_step`.
  Newton mode fatals on `inum != nlocal` and on triclinic (ownership rule
  not implemented for tri tag-based ordering yet).

### 12.2 Bugs hit during bring-up

1. **`neighbor->requests[i]` is deleted (nulled) after `Neighbor::init()`**
   (neighbor.cpp ~line 690); reading it at compute time segfaults on the
   first setup step. The persistent, index-aligned copies live in
   `neighbor->old_requests[]` -- use those (core does the same, e.g.
   `old_requests[i]->full` scans). Fixed in the dispatch.
2. **Stale-binary trap:** a background `make -C build-kokkos -j8` ran from
   `bench/` cwd, failed instantly ("No such file or directory"), and the
   `| tail` pipeline masked the nonzero exit -- the "fixed" binary was the
   old one and reproduced the identical crash. Use absolute `-C` paths and
   check `$?` explicitly.
3. **Vacuous validation diff:** the thermo header line starts with
   whitespace, so `grep "^Step"` matched nothing and every flat-vs-cluster
   diff printed IDENTICAL trivially. The script now extracts
   `/^ *Step /,/^Loop time/` and fails loudly when a log has no thermo.

### 12.3 Status at break (2026-07-11)

- Everything above is implemented; full+newtoff legacy decks passed with
  real (fixed) diffs: flat vs cluster thermo identical.
- `make -C /home/ndtrung/Codes/lammps-github/build-kokkos -j8` with the
  `old_requests` fix was still running at the time of writing
  (~15-20 min; pair_kokkos.h touches all KOKKOS pair styles).
- **Pending:** (1) rerun `scratchpad/validate_newton.sh` (4 decks:
  plain/special x half+newton/full+newtoff, flat vs cluster); (2)
  500-step melt on GP100 with `LMP_CLUSTER_STATS=1` -- expect ~7M tiles at
  ~90% occ (stats line now prints `newton 1`) and pair kernel ~19-22
  ms/step vs flat 32.9; watch for shared-mem hash saturation (fatal msg)
  -- fallback plan is level-1 (global-mem) scratch for hash > 2048 slots.
- Build protocol note: user asked for `make -j8` (not `cmake --build -j14`)
  for all future rebuilds.

### 12.4 Phase 3 results (2026-07-11): validated correct, but a perf regression on GP100

**Validation: 4/4 decks pass.** {plain, special} x {half+newton-on,
full+newton-off}, flat vs cluster, thermo bit-identical
(scratchpad/validate_newton.sh, recreated). The 500-step melt also runs
clean through mid-run resorts; thermo agrees exactly to step 200 and
diverges only at ULP level by step 500 (FP summation order).

**Profile (GP100, double, in.melt.sync 500 steps / nsys 100 steps):**

| metric                     | flat half+newton | cluster+Newton (P3) |
|----------------------------|------------------|---------------------|
| Loop time (500 steps)      | 25.92 s          | 31.12 s             |
| pair kernel /step          | 32.92 ms         | 35.29 ms (+7%)      |
| build /rebuild             | 137 ms (half)    | 286 ms (full flat) + 85 ms (ClusterBuild) = 371 ms |

CLUSTER_STATS: newton 1, 34838 iclusters, 13.02M tiles, 210.5M pairs,
occ 0.5051, max_jc 1162 (no hash saturation; 2048-slot headroom fine).

**Why the 11.5 projection (~7M tiles, ~90% occ, 19-22 ms/step) was wrong.**
It assumed half-mode tiles were direction-split duplicates -- that a given
unordered cluster pair (ci,cj) appeared as two ~half-empty tiles that
Newton would merge. Measured: half mode had 14.6M tiles at 44-46% occ,
Newton mode 13.0M at 50.5% -- only 11% fewer tiles. The coordinate-based
half rule plus Morton coherence already assigned nearly ALL pairs of a
given cluster pair to one side, so there was almost nothing to merge.
**The ~50% occupancy is the intrinsic geometric density of an 8x4 tile at
this cutoff (16 A) and density (~0.1/A^3)** -- of the 32 slot pairs
between two "touching" clusters, only ~16 are within cutoff. That ceiling,
not direction duplication, is the binding constraint; no ownership-rule
cleverness recovers it.

Additional cost drivers vs flat: 2x lane waste at 50% occ is not paid
back by coalesced d_xcl reads on GP100, and Newton mode forces the
FULL flat build (286 ms vs 137 ms half) as builder input.

**Verdict:** Phase 3 is correct and stays in the tree behind
`neigh/cluster yes`, but in double precision on GP100 it loses to the
flat kernel on every axis. Re-evaluate on A100 (target hardware; atomics
and coalescing economics differ). Next lever per plan: mixed/single
precision -- it halves per-lane compute cost, which also halves the price
of the wasted lanes, and is where GROMACS-style cluster kernels actually
win. Longer term the cluster builder must consume bins directly instead
of a full flat list (the 286+85 ms build chain is 2.7x the half build).

PR #5092, opened 2026-07-14
New chunk-based compute for coarse graining granular fields

jtclemm:
**Summary**

A new compute for coarse graining continuum fields has been added to the GRANULAR package based on the formulation by Goldhirsch. The implementation is quite similar to fix ave/chunk except the fix is uses a smoothing kernel to weight contributions from atoms and it does not perform the actual averaging (needs to be handed off to fix ave/time). Thus, the cutoff of the smoothing kernel can be larger than a chunk size such that atoms outside of a chunk can contribute to that chunk's value (when the kernel cutoff is > than the chunk size). The size/position of chunks therefore corresponds more to the position/resolution of the resulting coarse-grained field.

The fix only works with bin-based chunks (e.g. not spherical, cylindrical...).

**Author(s)**

Joel Clemmer (SNL)

**Licensing**

Github copilot was used to scan code for possible bugs/errors and draft the port from the fix to compute implementation (a lot of replacing).

**Backward Compatibility**

NA

**Implementation Notes**

NA


sjplimp:
@jtclemm A couple of conceptual Qs
(1) If this is averaging per-atom quantities over spatial bins (chunks), I'm not clear why not do this:  Implement the per-atom calculations in a compute (e.g. compute continuum/atom), and then use the existing fix ave/chunk or fix ave/grid to do the time averaging.  This seems more consistent with the way the other "averaging" fixes work.  Fix ave/grid would allow the same compute to work for much large systems, since it doesn't produce "global" output.
(2) You mention cutoffs beyond the chunk (bin) size.  Does that require communication?  Is so, that might be something fix ave/grid or the Grid classes do naturally.

jtclemm:
> (1) If this is averaging per-atom quantities over spatial bins (chunks), I'm not clear why not do this: Implement the per-atom calculations in a compute (e.g. compute continuum/atom), and then use the existing fix ave/chunk or fix ave/grid to do the time averaging. This seems more consistent with the way the other "averaging" fixes work. Fix ave/grid would allow the same compute to work for much large systems, since it doesn't produce "global" output. 

That could certainly be an option: have `compute continuum/chunk` compute per-chunk values and pass them along to `fix ave/chunk` for averaging/output. I did not go that route only because I do not know of another compute that operates this way - I didn't think `fix ave/chunk` is setup to grab data from other computes. 

Edit: I see now also `continuum/chunk` is probably a bad name since it hides the averaging part. Maybe something like `ave/chunk/continuum` or `ave/chunk/granular` would be better.

> (2) You mention cutoffs beyond the chunk (bin) size. Does that require communication? Is so, that might be something fix ave/grid or the Grid classes do naturally.

As in `fix ave/chunk`, this does require reduce all calls to sum up arrays of length `Nchunks`. It could very well be the case that the Grid class would be a better foundation for this fix, but admittedly I'm still not very familiar with it. I don't know how/whether grids could handle a kernel with a cutoff larger than a processor domain size.

jtclemm:
Converting to draft. realized chunk data can be fed to fix ave/time. This might be a simple way of removing duplicate code for handling temporal averaging.

jtclemm:
@sjplimp I finished converting this from a fix to a compute. Ends up I didn't have to make the corresponding compute property/chunk like we discussed since it conveniently already exists. Adding a grid version might be a nice follow up at a later date.

doc/src/compute_continuum_chunk.rst
```diff
@@ -0,0 +1,280 @@
...
+<Howto_chunk>` for details.
+
+This compute is only compatible with the binning styles of
+:doc:`compute chunk/atom <compute_chunk_atom>` (*bin/1d*, *bin/2d*, or
+*bin/3d*).  If binning is not performed along one of the box dimensions,
+all outputs are normalized by the box length in that dimension.  For
+example, if a 3d system is binned only along *z*, the reported fields are
+normalized by *Lx* and *Ly*.
+
+The available fields include scalar, vector, and tensor quantities.  A
```
sjplimp: Here and below you call these fields, but they are called values in the syntax section above.  Maybe you want to call them field1, field2, etc there.

doc/src/compute_continuum_chunk.rst
```diff
@@ -0,0 +1,280 @@
...
+vector field such as the velocity may be requested as an individual
+component such as *velocity/x* or with a wildcard such as
+*velocity/\**.  This expands to all components (two in 2d, three in
+3d).  Tensor fields such as the stress use component names like
+*stress/xy*.  Wildcards such as *stress/\** expand to all components
+(four in 2d, nine in 3d).  Tensors are not assumed to be symmetric.
```
sjplimp: Is this nine in 3d, instead of 6, b/c there are not symmetric 3x3 tensors ?  If so, would help to clarify that xy is a different value than yx.

doc/src/compute_continuum_chunk.rst
```diff
@@ -0,0 +1,280 @@
...
+
+Only atoms in the specified group contribute to the calculations.  In
+pair sums, both atoms must be in the group.  The referenced
+:doc:`compute chunk/atom <compute_chunk_atom>` command defines its own
+group and optional region.  Atoms with chunk ID = 0 are not assigned to
```
sjplimp, line 102: Is it possible for binning chunks to have atoms with chunk ID = 0 ?

doc/src/compute_continuum_chunk.rst
```diff
@@ -0,0 +1,280 @@
...
+any chunk and do not contribute.
+
```
sjplimp, line 104: I think in this sub-section, before you describe all the field options, you need to document what cutoff and width mean.  In particular that cutoff can be smaller or larger than bin size.  Also explain what happens when cutoffs are long.  E.g. does the comm cutoff need to be specified?  Or does that incur large computational costs ?
Also you might want to motivate for the user, what values to use for cutoff/width.

doc/src/compute_continuum_chunk.rst
```diff
@@ -0,0 +1,280 @@
...
+where :math:`\grad_{ab} v` is the :math:`ab` component of the
+*velocity/grad* field.
+
+The *temperature* field is a local granular temperature:
+
+.. math::
+
+   \frac{1}{2} \sum_i m_i (v_i - v_\mathrm{chunk})^2
+   W(\vec{r}_\mathrm{chunk} - \vec{r}_i)
+
```
sjplimp, line 222: I suggest putting the optional keywords in a sub-section of their own.
I.e. add a dashed line between fields and optional.

doc/src/compute_continuum_chunk.rst
```diff
@@ -0,0 +1,280 @@
...
+*boundary/atom* keyword designates a group of atoms as a boundary.
+Those atoms are removed from the atom and pair sums above, except for the
+*boundary/force* contribution.  The *boundary/fix* keyword applies the
+analogous correction for boundaries created by
+:doc:`fix wall/gran <fix_wall_gran>`.
+
+Output info
+""""""""""
+
+This compute calculates a global array where the number of rows is the
```
sjplimp, line 237: This is a global array even if the # of columns = 1 ?  Which is fine, just want to clarity.

doc/src/compute_continuum_chunk.rst
```diff
@@ -0,0 +1,280 @@
...
+Description
+"""""""""""
+
+.. versionadded:: TBD
+
+Define a computation that calculates coarse-grained continuum fields for
+chunks of atoms on demand using the construction in
+:ref:`(Goldhirsch) <_compute_continuum_chunk_goldhirsch>`.  The fields
+are evaluated at the chunk centers with a truncated Gaussian kernel.
+This compute does no time averaging and writes no files itself.  To time
```
sjplimp: output files

doc/src/compute_continuum_chunk.rst
```diff
@@ -0,0 +1,280 @@
...
+average or write the per-chunk data, use :doc:`fix ave/time
```
sjplimp: write out

doc/src/compute_continuum_chunk.rst
```diff
@@ -0,0 +1,280 @@
...
+   compute cont2 flow continuum/chunk cc2 2.0 1.0 volume/fraction temperature boundary/force/* boundary/fix
+
+Description
+"""""""""""
+
+.. versionadded:: TBD
+
+Define a computation that calculates coarse-grained continuum fields for
+chunks of atoms on demand using the construction in
+:ref:`(Goldhirsch) <_compute_continuum_chunk_goldhirsch>`.  The fields
```
sjplimp: Maybe add one sentence as to what these fields are useful for?  From a user perspective.

doc/src/compute_continuum_chunk.rst
```diff
@@ -0,0 +1,280 @@
...
+
+These values can be accessed by any command that uses global arrays from
+a compute as input.  See :doc:`Howto output <Howto_output>` for an
+overview of output options.  The array values are intensive.  Units
+depend on the requested fields.
+
+Restrictions
+"""""""""""
+
+This compute is part of the GRANULAR package.  It is only enabled if
```
sjplimp, line 254: Is putting this in the GRANULAR package b/c these fields are only useful in a granular context?
Which is fine, just wondering if these (or similar fields which could be added) have broader applicability,
e.g. a stress-field in a solid.

sjplimp:
@jtclemm I also think adding a couple example scripts would be good.  If you do that I will approve, and possibly this could be included in the stable release (tomorrow deadline?)

doc/src/compute_continuum_chunk.rst
jtclemm: There are some scenarios described on the compute chunk/atom doc page, but also atoms not in the group.

doc/src/compute_continuum_chunk.rst
jtclemm: Good catch, I added that. 

All processors store and array of length nchunk x nvalues. Each processor then loops through all owned atoms and adds their contributions to all chunks within cutoff regardless of where the chunk center is located. This is all summed by a reduce_all() call. So no, there's no need to worry about comm cutoffs.

doc/src/compute_continuum_chunk.rst
jtclemm: Yes, this is always an array. I added clarification.

doc/src/compute_continuum_chunk.rst
jtclemm: Good question. As far as I know, the only usecase of this method has been the discrete element method. There's no reason it couldn't be applied to other methods like peridynamics, but those methods often have their own methods of representing such fields. 

Code wise, there is a hardcoded requirement that atoms have the radius property. So that practically restricts it to pair granular. 

I'm not opposed to trying to generalize it, but my immediate thought is that at the present it makes sense to associate it with granular. Maybe something to keep in mind in case we find another application or if it's requested?

jtclemm:
> @jtclemm I also think adding a couple example scripts would be good. If you do that I will approve, and possibly this could be included in the stable release (tomorrow deadline?)

Thanks for all the feedback! I added all the changes and replied to the above comments where there were questions. I don't have an example immediately handy, but I can see what I can do. Probably nothing by tomorrow though.

sjplimp:
thanks - I'll approve it now.  Axel can decide whether to add to stable.  You could add example input scripts later.
I think you should edit the Summary at the top of this page to reflect compute (not fix) and no time averaging.

sjplimp (approved):

jtclemm:
> thanks - I'll approve it now. Axel can decide whether to add to stable. You could add example input scripts later. I think you should edit the Summary at the top of this page to reflect compute (not fix) and no time averaging.

Added a little 2D chute flow example that runs pretty quick. I'll edit the summary too, thanks for noting.

No big deal in my opinion whether or not it makes the stable release.

PR #5096, opened 2026-07-17

### LINKED ISSUE #5095: press/langevin documentation page mentions modulus but does not appear to be supported
The `press/langevin` docs page https://docs.lammps.org/fix_press_langevin.html mentions a `modulus` keyword, but there's no information on it, and using it seems to indicate that it is not in fact part of `press/langevin` 
```
ERROR: Unknown fix press/langevin keyword: modulus (src/lammps_cmake_symmetrix/src/fix_press_langevin.cpp:249)
```
(perhaps it was cut and pasted from `press/berendsen`?)

akohlmey:
**Summary**

This pull request contains various small changes, primarily related to updating and modernizing the [example graphics movies](https://www.lammps.org/examples/) on the LAMMPS homepage. Checking on the GRANSURF package examples exposed a few bugs that should be addressed here.

**Related Issue(s)**

Fixes #5095 
Addresses [this problem](https://matsci.org/t/change-in-variable-behavior-between-2024-2026/67350/) reported on MatSci.org

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

N/A

**Implementation Notes**

- remove stray *modulus* keyword from command summary of `fix press/langevin`
- check memory pool storage in `fix surface/local` for overflows and error out instead of segfaulting
- drop degenerate triangles when importing from STL files
- protect surface connectivity information from being currupted by having multiple identical endpoint
- fix a copy-n-paste bug in `fix surface/local`
- correctly compute and set the buffer sizes for border and exchange communication in `fix surface/local`
- missing connected surface atoms are not an error if the missing atom would be a ghost. Print a warning instead.
- update the visualization settings for several GRANSURF examples and set communication cutoff to avoid communication problems in parallel runs
- add note to granular surfaces howto about remeshing STL files created for 3d-printing
- add a `dump_modify gtrans` keyword so that `dump image` can also render grids as transparent
- re-enable self-referentials when defining string style variables
- update several input examples to make use of LAMMPS' improved graphics capabilities
- adjust unittest YAML file epsilon in a few cases so more tests pass on ARM64

**Post Submission Checklist**


akohlmey:
@jtclemm I could use a little help to improve the examples/GRAPHICS/in.mixer-drum example a little bit. I've used one of the gransurf example inputs as a template and then imported the downloaded STL file and set up the simulation from there.
- in the process, I ran into multiple issues which are now fixed in this pull request
- I found that re-meshing the STL file is a big help to reduce the neighbor list and thus speed up the run substantially, thus I added a note to the manual
- I am trying to recreate a demo that I did many years back for the `create_atoms mesh` command: https://www.lammps.org/examples/mixer_drum/
- the remaining issue is that with the given parameters, particles get pushed through the surface or get stuck in the surface somehow. Since I have no clue about the science of those models, I am at a loss here.

The run functions OK as a visual demo, but it would be cool if the physics would be meaningful, too. Thus, any suggestions or direct improvements on the input file are much appreciated.

jtclemm:
Hi Axel, thanks for finding those issues and fixing them. The walls still strike me as being in a bit of a beta-phase since there's a lot of scenarios still to test. THe drum is a good stress test (and also a nice visual!).

Regarding particles getting stuck in the walls, this is most likely due to the wall not being stiff enough relative to gravity (although it is also possible that there is some edge case where triangles are arranged in some position that isn't fully atom-tight). I tried halving gravity and doubling the wall stiffness as a quick go, and it seems to have worked. 

Still running it longer to confirm though.

Edit: Yes, that seems to have done the trick. One particle may have still gotten stuck, but hard to tell with the few frames I printed out. Could drop the gravity a bit more.

src/COLLOID/pair_colloid.cpp
```diff
@@ -550,3 +551,16 @@ double PairColloid::single(int /*i*/, int /*j*/, int itype, int jtype, double rs
 
   return factor_lj*phi;
 }
+
+/* ---------------------------------------------------------------------- */
+
+void *PairColloid::extract(const char *str, int &dim)
```
jtclemm, line 557: Does this addition need to be reflected in the doc files? For fix adapt maybe?

src/GRANSURF/fix_surface.cpp
```diff
@@ -348,69 +349,86 @@ void FixSurface::extract_from_stlfile(char *filename, int stype, int smol,
...
 
   for (int itri_new = 0; itri_new < ntris_new; itri_new++) {
-    int itri = itri_new + ntris_old;
+    double *corners = stltris[itri_new];
+
+    double edge1[3], edge2[3], cross[3];
+    MathExtra::sub3(&corners[3], &corners[0], edge1);
+    MathExtra::sub3(&corners[6], &corners[0], edge2);
+    MathExtra::cross3(edge1, edge2, cross);
+    if (cross[0] == 0.0 && cross[1] == 0.0 && cross[2] == 0.0) {
```
jtclemm, line 372: I assume this is checking to see if the triangle has no area. Does it make sense to check if `fabs(cross[x]) < EPSILON` instead of an exact zero check? Not sure it makes a difference, but I'm always superstitious about `== 0`.

src/COLLOID/pair_colloid.cpp
akohlmey: Good point. Probably, yes. I'll check.
This is one of the things where it would be good to have a script checking for it.

src/GRANSURF/fix_surface.cpp
akohlmey: That is ultimately a question about how robust the force computation in fix surface/* is. The immediate necessity here is that there are no identities because they mess up the bookkeeping. Beyond that, it may be smarter to skip very tiny triangles, but I would expect that to be taken care of by re-meshing, which I strongly recommend based on my experience. STL files from e.g. Thingiverse can have very narrow and long triangles in addition to zero area ones and those mess up the computational efficiency a lot because they trigger long neighbor lists.

Examples like the mixer drum are also challenges for load balancing, but I found a lot of time is spent communicating and building neighbor lists, so this could also be a case to explore alternative options, e.g. multi cutoff.

jtclemm: The multi neighbor style does work with the STLs and I have found it can substantially improve performance in relevant scenarios.

jtclemm (approved):

PR #5106, opened 2026-07-25
Add VTKWriter class for unified VTK output without requiring the VTK libraries

akohlmey:
**Summary**

This adds a self-contained VTKWriter class that supports only writing VTK output files (legacy and XML in ASCII and binary with optional zlib compression), `fix saed/vtk`, `dump vtk` and `dump grid/vtk` are adapted to use the new class now replacing custom code and interfacing to VTK.

There are multiple benefits:
- no more dependency on the VTK development package, which is huge and complex to build from source and rather wasteful for using only a very small subset of its functionality
- the VTKWriter class is fully portable and thus dump vtk become generally available and moves from its own VTK package to the EXTRA-DUMP package
- `fix saed/vtk` and `dump grid/vtk` gain support for missing legacy/XML support and (compressed) binary support.

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

The bulk of the code was drafted by Claude Code using Opus 5.

**Backward Compatibility**

Yes

**Implementation Notes**

The implementation corrects minor issues in the ad hoc versions and skips data that is not used or required and thus files should overall shrink. Especially compressed binary output should be smaller. This version retains outputting single precision floating point, but has a built-in check for output where precision may be impacted. A future extension to support double precision floating point data should be straightforward.

**Post Submission Checklist**


akohlmey:
@rbberger @sjplimp The code currently outputs only coordinate data in single precision and sticks with double precision for other data. This more-or-less follows the dump vtk example, but seems rather arbitrary and in my personal opinion, that data doesn't need that much precision either. I think having all floating point data in single precision would be more meaningful and double precision support could easily by a dump_modify option. Adding this to the implementation is very little effort. But I didn't want to change behavior without asking. The current implementation reproduces dump vtk.

As a bonus, some heuristics are added for larges systems/coordinates that print a warning if the absolute precision drops too much. My intention was to wait and see, if somebody comes across this limit and asks for a change. Better than just silently truncating data.

sjplimp (approved):

sjplimp:
@akohlmey Not clear on the coord precision comment.  Seems like we should just output coords (and other data) the same as dump atom or dump custom does.  Is there something unique about the way VTK files are typically formatted ?

akohlmey:
> @akohlmey Not clear on the coord precision comment. Seems like we should just output coords (and other data) the same as dump atom or dump custom does. Is there something unique about the way VTK files are typically formatted ?

Not much, but we have 4 different encodings possible: legacy and XML, and for each there is binary (as Base64 text) and text.

In the existing output options, there were either some of the permutations missing or they were already inconsistent. So whatever is done, there is not 100% backward compatibility possible. But I think that a consistent output would be an improvement (and waste less electrons or magnetic spins).

If we want somewhat mimic dump atom/custom we should use single precision (i.e. 8 characters in text and 4 bytes binary) by default and double precision (i.e. 16 and 8) on request with a dump_modify option. That would be my personal preference.

sjplimp:
@akohlmey I agree with the your last (personal preference) option.  Doing that in follow-up PR is fine.

PR #5024, opened 2026-06-06
Add classical constant-potential fix uvt

Lily200202:
**Summary**

This pull request adds a new `fix uvt` style for classical constant-potential dynamics in LAMMPS.

The implementation introduces a single extended generalized coordinate for the electron number, `Ne`, coupled to the existing Nosé-Hoover thermostat machinery. The driving force on `Ne` is

$$
\mu - \frac{dE}{dN}
$$

where `mu` is the imposed chemical potential and `dE/dN` is supplied externally.

The main design goal is to keep the LAMMPS-side implementation generic. `fix uvt` acts as a consumer of `dE/dN`, while the actual source of `dE/dN` can come from an equal-style variable, a compute, or a fix.

This pull request includes:

- a new `fix uvt` style
- documentation for the new fix
- a toy-model example with an analytic equilibrium
- a restart regression example
- a command-level unit test
- a companion backend/provider pull request for DeepMD-kit

**Related Issue(s)**

None currently referenced.

**Author(s)**

Li Fu  
Peking University, Beijing 100871, People’s Republic of China  
Corresponding e-mail: [fuli200202@163.com](mailto:fuli200202@163.com)

**Licensing**

By submitting this pull request, I agree that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Artificial Intelligence (AI) Tools Usage**

By submitting this pull request, I confirm that I used AI tools to help draft and revise parts of the implementation, documentation, tests, and validation workflow.

In particular, OpenAI Codex / ChatGPT were used to assist with code drafting, documentation wording, test scaffolding, and iterative implementation review. Final design choices, validation decisions, and repository integration were reviewed and approved by the human author.

**Backward Compatibility**

This pull request does not change the behavior of existing `fix nvt`, `fix npt`, or `fix nph` inputs.

It adds a new `fix uvt` style and related documentation, examples, and tests. Existing input scripts are expected to remain backward compatible.

**Implementation Notes**

`fix uvt` extends the Nosé-Hoover NVT framework with one additional thermostatted quadratic degree of freedom corresponding to the electron-number coordinate `Ne`.

The implementation includes:

- thermostat-potentiostat coupling through the existing Nosé-Hoover chain machinery
- restart support for the electron-number state
- a generic consumer-side interface for `dE/dN`

The `dE/dN` input may be supplied by:

- an equal-style variable: `v_name`
[...]

PR #5107, opened 2026-07-26
Add constant-potential UVT and PIMD/UVT fixes

Lily200202:
**Summary**

This pull request supersedes #5024 and #5061. It brings the classical and path-integral constant-potential implementations into a unified constant-potential framework with two related fix styles sharing a common electronic-response input convention:

- `fix uvt` adds classical constant-potential dynamics through a single extended electronic coordinate, `Ne`.
- `fix pimd/uvt` adds the path-integral constant-potential extension within the PIMD NVE/NVT hierarchy, with a single electronic coordinate shared across bead partitions and synchronized during integration.

Both fixes use the same user-facing `dE/dN` input convention: an equal-style variable, a global compute, or a global fix, optionally with a vector index. The driving force on the electronic coordinate is based on `mu - dE/dN`.

Compared with the earlier standalone `fix tprpmd` implementation in #5061, the PIMD constant-potential implementation is now exposed as `fix pimd/uvt` and reuses the PIMD hierarchy instead of duplicating normal-mode propagation and inter-partition bead communication logic.

**Related Issue(s)**

Supersedes #5024 and #5061.

**Author(s)**

Li Fu  
Peking University, Beijing 100871, People's Republic of China  
Corresponding e-mail: fuli200202@163.com

**Licensing**

By submitting this pull request, I confirm that I used AI tools to assist with parts of the implementation, documentation, test scaffolding, debugging, and validation workflow for this contribution.

OpenAI ChatGPT / Codex was used to help draft and revise code, documentation text, tests, and validation steps. The algorithmic design, physical model, validation decisions, and final repository integration were reviewed and approved by the human author.

**Backward Compatibility**

This pull request adds new upstream-facing fix styles and does not change the behavior of existing input scripts for `fix nvt`, `fix npt`, `fix nph`, or `fix pimd/langevin`.

The previous standalone `fix tprpmd` proposal from #5061 is not retained as a public style in this PR. The upstream-facing path-integral constant-potential interface is now `fix pimd/uvt`, reflecting that the implementation is an extension of the PIMD fix hierarchy.

The deprecated `fix pimd/nvt/bosonic` style entry is retained as a stub that directs users to `fix pimd/langevin/bosonic`.

**Implementation Notes**

As part of this work, the PIMD NVE/NVT implementation was refactored into a reusable hierarchy. This refactor provides shared hooks for normal-mode propagation, inter-partition bead communication, restart/vector output, and thermostat integration, which are then extended by `fix pimd/uvt`.

`fix uvt` is implemented in the `EXTRA-FIX` package as a `FixNH`-derived integrator. It introduces one additional electronic degree of freedom, `Ne`, coupled to the Nose-Hoover thermostat dynamics, and supports restart and global-vector output of the electronic state.

`fix pimd/uvt` is implemented in the `REPLICA` package. It derives from `FixPIMDNVT` and reuses the shared PIMD machinery for normal-mode propagation, inter-partition bead communication, restart/vector hooks, and thermostat integration. Each bead partition evaluates its local `dE/dN` source; the result is averaged across partitions before the shared electronic coordinate is updated and broadcast.

The submitted tests cover classical UVT behavior, PIMD/NVT hierarchy behavior, PIMD/UVT serial behavior, restart restoration, electronic-state extraction, input rejection for legacy/non-supported syntax, and MPI partitioned PIMD/UVT runs. The implementation was verified with CMake builds and targeted serial/MPI unit tests for `fix uvt`, `fix pimd/nvt`, and `fix pimd/uvt`.

**Post Submission Checklist**

This PR supersedes the earlier classical-only `fix uvt` PR (#5024) and the earlier standalone `fix tprpmd` PR (#5061). The current implementation combines both layers into one unified constant-potential framework and addresses the prior review request to avoid duplicating PIMD propagation logic by reusing the PIMD hierarchy.

Additional long-form toy-model and real-system validation was performed outside the upstream PR tree.


src/REPLICA/fix_pimd_nve.cpp
```diff
@@ -0,0 +1,1127 @@
...
+      (strcmp(arg[i], "tchain") == 0) || (strcmp(arg[i], "tloop") == 0) ||
+      (strcmp(arg[i], "drag") == 0) || (strcmp(arg[i], "mu") == 0) ||
+      (strcmp(arg[i], "ne") == 0) || (strcmp(arg[i], "ne_velocity") == 0) ||
+      (strcmp(arg[i], "dedn") == 0) || (strcmp(arg[i], "tau") == 0) ||
+      (strcmp(arg[i], "seed") == 0) || (strcmp(arg[i], "PILE_L_temp") == 0) ||
+      (strcmp(arg[i], "barostat") == 0) || (strcmp(arg[i], "iso") == 0) ||
+      (strcmp(arg[i], "aniso") == 0) || (strcmp(arg[i], "taup") == 0) ||
+      (strcmp(arg[i], "fixedpoint") == 0)) {
+    error->all(FLERR, "Keyword {} is not supported by fix {}", arg[i], style);
+  }
```
Yi-FanLi: It seems several keywords, such as `ne` and `dedn` are for `uvt` rather than `nve`. Why are they here? Is it necessary to make a blacklist for these keywords in the `FixPIMDNVE` class?

src/REPLICA/fix_pimd_nve.cpp
```diff
@@ -0,0 +1,1127 @@
...
+  return mask;
+}
+
+void FixPIMDNVE::init()
+{
+  if (atom->map_style == Atom::MAP_NONE)
+    error->all(FLERR, "Fix {} requires an atom map, see atom_modify", style);
+
+  if (comm->nprocs != 1)
+    error->all(FLERR, "Fix {} currently requires one MPI rank per bead", style);
```
Yi-FanLi: This part looks inconsistent to me. Why do you need to restrict the `nprocs` to `1`? We definitely support multi-proc parallelization for each bead, right?

src/REPLICA/fix_pimd_nve.cpp
```diff
@@ -0,0 +1,1127 @@
...
+{
+  inter_replica_comm(ptr);
+  nmpimd_transform(normal_mode_transform_buffer(), ptr, M_xp2x[universe->iworld]);
+}
+
+void FixPIMDNVE::prepare_setup_normal_mode_coordinates()
+{
+  unmap_coordinates(atom->x, atom->image);
+  forward_normal_mode_transform(atom->x);
+}
```
Yi-FanLi: How do these two functions differ? `prepare_setup_normal_mode_coordinates` and `begin_normal_mode_coordinate_propagation`.

src/REPLICA/fix_pimd_nve.cpp
```diff
@@ -0,0 +1,1127 @@
...
+  }
+  MPI_Bcast(xcall, ntotal * 3, MPI_DOUBLE, 0, universe->uworld);
+}
+
+void FixPIMDNVE::b_step()
+{
+  apply_force_velocity_kick(false);
+}
+
+void FixPIMDNVE::apply_force_velocity_kick(bool restrict_group)
```
Yi-FanLi: The group check logic in the old implementation is direct and consistent with the design of the main body of LAMMPS. Why is `restrict_group` introduced? I didn't see the point.

src/REPLICA/fix_pimd_nve.cpp
```diff
@@ -0,0 +1,1127 @@
...
+  if (universe->iworld == 0) {
+    for (int i = 0; i < nlocal; i++) {
+      x[i][0] += dtv * v[i][0];
+      x[i][1] += dtv * v[i][1];
+      x[i][2] += dtv * v[i][2];
+    }
+  }
+}
+
+void FixPIMDNVE::a_step()
```
Yi-FanLi, line 642: The group check is missing from this function, making the behavior incorrect when one wants to run PIMD for a specific group of atoms. You may refer to the old implementation here: https://github.com/lammps/lammps/blob/2aea0937c2dead1bed206007dfdb8cba972cbdb0/src/REPLICA/fix_pimd_langevin.cpp#L875.

src/REPLICA/fix_pimd_nve.cpp
```diff
@@ -0,0 +1,1127 @@
...
+    if (restrict_group && !(mask[i] & groupbit)) continue;
+
+    double dtfm = dtf / mass[type[i]];
+    v[i][0] += dtfm * f[i][0];
+    v[i][1] += dtfm * f[i][1];
+    v[i][2] += dtfm * f[i][2];
+  }
+}
+
+void FixPIMDNVE::qc_step()
```
Yi-FanLi, line 626: Similar to https://github.com/lammps/lammps/pull/5107/changes#r3654336254.

Yi-FanLi (changes requested):
I started from reviewing the base class `FixPIMDNVE` and commented several issues. @Lily200202 Could you please address my comments before we move on to the remaining parts of this PR?

src/REPLICA/fix_pimd_nve.cpp
Lily200202: Thanks for pointing this out. I agree that the PIMD/NVE base class should not maintain a blacklist of keywords that belong to derived styles such as pimd/uvt. I removed this blacklist from FixPIMDNVE::parse_common_keyword(), so the base parser now only handles the common PIMD/NVE keywords. The NVT and UVT-specific keywords are handled by their respective parsers, and unsupported keywords fall through to the generic unknown-keyword error.

src/REPLICA/fix_pimd_nve.cpp
Lily200202: Thanks, you are right. The PIMD infrastructure already has a multi-rank-per-bead communication path, and this restriction was unnecessarily disabling it. I removed the one-rank-per-bead check from FixPIMDNVE::init().

I also added a regression test that runs fix pimd/nve with 2 bead partitions and 2 MPI ranks per bead (-partition 2x2), so the multi-rank-per-bead path is now covered in the unit tests.

src/REPLICA/fix_pimd_nve.cpp
Lily200202: These two functions were doing the same operation: unwrapping coordinates and transforming them to normal-mode coordinates. I removed prepare_setup_normal_mode_coordinates() and now reuse begin_normal_mode_coordinate_propagation() in setup(), including the corresponding NMPIMD setup path in fix pimd/langevin.

src/REPLICA/fix_pimd_nve.cpp
Lily200202: Thanks, I agree. The restrict_group argument was unnecessary and made the group handling less direct. I removed it and restored the standard LAMMPS-style group mask check inside apply_force_velocity_kick(). Both the PIMD NVE path and the PIMD Langevin path now use the same helper with consistent group handling.

src/REPLICA/fix_pimd_nve.cpp
Lily200202: Thanks, you are right. The coordinate propagation steps also need to respect the fix group. I added the standard group mask check to both qc_step() and a_step(), consistent with the old implementation and with the force kick path.

PR #5109, opened 2026-07-27
Add try_jump option to read_restart

Matthew-Whitlock:
**Summary**

This option allows users to attempt to read_restart when the presence of a restart file is unknown. Allows checkpoint/restart to run via a single input file that can be called repeatedly. The *try_jump* option allows you to jump past any simulation setup that you would do in the absence of a restart file.

**Author(s)**

Matthew Whitlock, Sandia National Laboratories.

**Licensing**

**Backward Compatibility**

No breaks.

**Implementation Notes**

There aren't any existing unit tests for jump functionality, so I chose an example file that uses read_restart and modified it to use this new feature. I manually confirmed correct behavior when no restart file was written by commenting out the write_restart command in the example's input file.

Is this sufficient for this small change, or do changes need to be made to support jumping to labels in the unit tests?

**Post Submission Checklist**

### DIFF
--- a/doc/src/read_restart.rst
+++ b/doc/src/read_restart.rst
@@ -8,9 +8,16 @@ Syntax
 
 .. code-block:: LAMMPS
 
-   read_restart file
+   read_restart file keyword value ...
 
 * file = name of binary restart file to read in
+* zero or more keyword/value pairs may be appended
+* keyword = *try_jump*
+
+  .. parsed-literal::
+
+      *try_jump* value = jump_location
+        jump_location = arg(s) passed to jump command if restart is successful
 
 Examples
 """"""""
@@ -19,6 +26,8 @@ Examples
 
    read_restart save.10000
    read_restart restart.*
+   read_restart restart.* try_jump newfile
+   read_restart restart.* try_jump "SELF runloop"
 
 Description
 """""""""""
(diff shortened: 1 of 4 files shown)


akohlmey:
@Matthew-Whitlock I am sorry, but there is no need for such a feature in LAMMPS.

We have a variable function `is_file()` that can be used in combination with the `if` command to check for the presence of a file and then act accordingly. That is far more generic and more flexible than what you are proposing.

Matthew-Whitlock:
@akohlmey That would definitely work for the example I gave, but I'm really thinking of the "restart" command use cases where the exact file name is unknown. It might be checkpoint.100.base, checkpoint.200.base, etc, but the file name for is_file is taken as a literal.

Would you instead support a "noerror" flag to do this without the jump? Users could then check if the box exists to know if recovery was successful.

akohlmey:
> It might be checkpoint.100.base, checkpoint.200.base, etc, but the file name for is_file is taken as a literal.

That case is already taken care of by read_restart which accepts wildcard filenames.
For most other cases, the handling of finding and naming the restart file is handled by the submit shell script that calls LAMMPS and then the restart file name is passed to the LAMMPS input as an index style variable with the -var / -v command line flag. Folks have been using this for many years.

PR #5110, opened 2026-07-27
Granular unittests

zawh4159:
**Summary**

This pull request updates the YAML-driven regression and verification test suite for granular / discrete element method (DEM) contact models under `unittest/granular`, replacing the previous 11-test suite with a 8-test suite, built in the same spirit as the `unittest/force-styles` harness.

| program | scenario | analytic model |
|---|---|---|
| `test_dem_01` | two-sphere head-on normal collision | `collision_restitution` |
| `test_dem_02` | two-sphere elastic Hertzian normal impact (peak force) | `hertz_normal_impact` |
| `test_dem_03` | sphere–wall elastic Hertzian normal impact (peak force) | `hertz_normal_impact` |
| `test_dem_04` | oblique impact on a wall (gross-sliding friction) | `oblique_impact` |
| `test_dem_05` | sphere sliding then rolling without slipping | `slip_cessation` |
| `test_dem_06` | spinning sphere impact (rebound with friction) | `spin_impact` |
| `test_dem_07` | rolling-resistance decay | `rolling_decay` |
| `test_dem_08` | cohesive DMT pull-off force | `pulloff_dmt` |

These particle-impact-level benchmarks follow the software-agnostic DEM benchmark of Mohajeri et al. (Powder Technology 447, 2024) — rolling resistance and cohesion — and the particle-impact benchmark of Chung & Ooi (Granular Matter 13:643–656, 2011) — normal, oblique, and spinning-sphere collisions.

Each test drives a small, analytically tractable system built entirely from a YAML file, runs it in segments, and checks the resulting trajectory both as a bit-for-bit regression against a stored per-atom reference (under `newton on` and `newton off`) and against a closed-form analytic model appropriate to the scenario. A given driver may cover several variants of one scenario — across contact models (`hooke`, `hooke/history`, `hertz`, `hertz/material`, `mindlin`, `mindlin/rescale`), dimensionality, or unit systems — since it runs every `dem0N-*.yaml` file registered under its number.

**Author(s)**

Zachary White (Sandia National Labs, University of Colorado Boulder), zwhite55@gmail.com
Emad Renfroe (Sandia National Labs, Georgia Tech)

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Artificial Intelligence (AI) Tools Usage**

[...]

PR #5102, opened 2026-07-23
Collected small changes and fixes

akohlmey:
**Summary**

This pull request combines multiple small changes and fixes. Mostly related to the documentation

**Related Issue(s)**

Closes #5101
Closes #5099
Closes #5098
Fixes #5104 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

By submitting this pull request, I agree, that my contribution will be included in LAMMPS and redistributed under either the GNU General Public License version 2 (GPL v2) or the GNU Lesser General Public License version 2.1 (LGPL v2.1).

**Artificial Intelligence (AI) Tools Usage**

The python script for checking the fix adapt and fix adapt/fep tables of adaptable
styles was created by Claude Fable 5. The bibliography consolidation was performed by Claude Opus 5. Both under the supervision and with directions from me.

Everything else was done without AI assistance.

**Backward Compatibility**

N/A

**Implementation Notes**

The following changes are included:
- do not use the OPENMP version of fix NEIGH_HISTORY with pair styles granular, tracker, granular/superellipsoid which do not have OPENMP versions
- fix bugs in `fix NEIGH_HISTORY/omp` that could lead to segfaults or data corruption
- fix bug in `fix pimd/nvt` where necessary memory allocation was skipped
- fix bug in `pair_style eam` allocated data was counted before it was allocated and available, which caused a segfault when LAMMPS threw and exception due to a broken EAM potential file.
- fix bug in `create_atoms` that would cause a segmentation fault when creating atoms from a molecule file containing dipoles. Refactor the relevant code to make copies instead of passing pointers between translation units.
- change `fix graphics/replica` so that it selects the atoms from the fix group on replica 0 across all replica
- replace `.. deprecated::` with `.. versionremoved::`
- consolidated citations and updated Bibliography.rst
- add tool to check extract'able style properties for `fix adapt` and `fix adapt/fep`
- update and correct the aforementioned tables
- remove a leftover reference to `ML-RANN`
- Include trivial PRs #5098, #5099, and #5101 with some minor edits
- Update some unittest YAML files to pass automated tests on ARM64
- Add commented out dump image visualization to powerlaw example in multi folder
- Add example visualization a PIMD simulation to the ``examples/GRAPHICS`` folder

**Post Submission Checklist**

- [x] The feature or features in this pull request is complete
- [x] Licensing information is complete
- [x] Corresponding author information is complete
- [x] The source code follows the LAMMPS formatting guidelines
[...]

PR #5113, opened 2026-07-28
Add more granular unittests

akohlmey:
**Summary**

This pull request tries to expand on PR #5110 to add more tests, particularly for legacy granular styles and to cover components that were missed or need more data points.

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U and Joel Clemmer, SNL

**Licensing**

Most of the added tests were added by Claude Code (Fable 5) based on multiple publications, the initial draft implementation of the DEM tests and collected insights from multiple bugfix and analysis sessions evaluating data from static code analysis.

**Backward Compatibility**

N/A

**Implementation Notes**

- 5d372a0863 -- fix division by zero (NaN forces) for undamped classic granular wall models
- 027c2f9272 -- harness robustness: safe reference regeneration (temp file + rename), tags round-trip fix
- 95f9efecdc -- restore the four analytic models dropped in the PR #5110 rewrite
- 5af22707d8 -- new tests: legacy-style variants, 76-degree energy-injection guard, zero-slip spin guard, terminal velocity under drag (`test_dem_09`)
- c7497e5fda -- documentation for the first batch
- 51404698e6 -- fix a pre-existing typo found by `make spelling`
- cdf871b5e6 -- new and restored analytic models: `hertz_peak`, `slip_transient`, `incline_rolling`, `wall_restitution`, `momentum_conservation`, `pulloff_jkr`, `freefall`, `stack_energy`
- b904660349 -- new tests: restitution/damping/tangential-velocity sweeps, per-quantity Hertz peaks, JKR pull-off, inclined-plane rolling, free fall + stacked compression (`test_dem_10`), region wall (`test_dem_11`), superellipsoid momentum conservation (`test_dem_12`)
- 101687e0a5 -- documentation for the second batch
- 6e7933f3c1 -- new analytic models: `twist_decay`, `twist_decay_marshall`, `heat_equilibration`
- a387eca3ad -- new tests: twisting friction (`test_dem_13`), granular heat conduction (`test_dem_14`), sticking-regime oblique impact
- 50a4f17894 -- documentation for the third batch
- 8eb6d68cd0 -- spelling false positives
- 6c09dddd2c -- add the parked `damping tsuji` restitution test, enabled by the sqrt(2) fix from PR #5116
- 10e220f099 -- new driver `test_dem_15` + `oblique_impact_pair` model: two-sphere oblique impact (Chung & Ooi Test 5)
- e248bb89a0 -- new tests completing sub-model coverage except `mdr`: `mindlin[_rescale]/force`, `mass_velocity` damping, tsuji wall
- 369f4d4f85 -- documentation for the fourth batch

**Post Submission Checklist**


akohlmey:
@jtclemm here is the first batch of expansions and fixes. Feel free to request changes here or on slack, I will feed them back to the coding agent. The goal of the next round it to extend coverage and integrate more tests suggested by the published literature.

akohlmey:
@jtclemm While preparing a restitution test for `damping tsuji` for this PR, it was found
that the model does not reproduce the requested restitution coefficient. For that reason the corresponding test was not added.

**Observation.**  For a Hertzian two-sphere impact with
`hertz/material ... damping tsuji` and a requested restitution coefficient
of e = 0.5, the measured rebound gives e = 0.616.  The deviation is
systematic:

| requested e | measured e |
|---|---|
| 0.2 | 0.343 |
| 0.5 | 0.616 |
| 0.7 | 0.778 |
| 0.9 | 0.928 |

The docs state for *tsuji* that "the damping coefficient specified as part
of the normal model is interpreted as a restitution coefficient e", so the
behavior contradicts the documentation, and the model always dissipates
less than requested.

**Diagnosis.**  Independently integrating the documented force law (m_eff
d'' = -K d^{3/2} - alpha sqrt(m_eff k_nd) d', with k_nd = F/d) reproduces
the LAMMPS results exactly, i.e. the code implements its formula correctly.
The problem is the alpha(e) polynomial in `GranSubModDampingTsuji::init()`
(gran_sub_mod_damping.cpp): the alpha actually required for a given e is
**larger by exactly sqrt(2)** across the entire range (numerically inverted:
ratios 1.4139 to 1.4169 for e = 0.2 to 0.9).  So the polynomial appears to
be a correct fit, but for a damping convention whose m*k product is twice
ours (e.g. single-particle mass where we use m_eff = m/2, or a
tangent-stiffness convention).

**Proposed changes.** 
1. Fix the mapping by adding `damp *= sqrt(2)` after the polynomial.
   This would restore e_in = e_out to better than 1 percent, but
   changes the damping model to be more dissipative
2. Change the documentation to describe what the parameter actually is.
   Not sure if that would require to change the name of the model.
3. Like 1. but using a more accurate representation for a (slightly)
   better match to the analytic model (probably not worth it).

akohlmey:
@jtclemm Here are suggested additional changes that require input from you and are probably best added in a followup pull request.

- test mdr normal model: needs more information to have a suitable reference to compare to
- tests for surface/global and surface/local comparing to a flat wall or a box (corner?): needs to merge pending pull requests that contain fixes for them (e.g. #5102) or cherry-pick fixes from them
- test (small) bulk systems similar to the force style tests: regression tests only. the older tests had a few. this could be used to confirm neighbor list settings, too.
- run the tests for (existing and future) accelerated versions

jtclemm:
> @jtclemm While preparing a restitution test for `damping tsuji` for this PR, it was found that the model does not reproduce the requested restitution coefficient. For that reason the corresponding test was not added.

Thanks Axel for finding this and letting me know. I (or @dsbolin) will take a look. I just uncovered another general (non-granular) bug with regions that span periodic boundaries w/ open ends. I will submit a short PR once I have a chance to patch both.

jtclemm:
Made some progress in investigating. Extracting the data from Tsuji's [original paper](https://doi.org/10.1016/0032-5910(92)88030-L), the sqrt(2) factor does improve the fit of the equation in LAMMPS relative to Tsuji's Fig. 3 data:
image

The equation in LAMMPS is a numerical fit from [Marshall 2009](https://doi.org/10.1016/j.jcp.2008.10.035), equation 53. 

At a quick glance, the two papers do have slightly different expressions they are solving (Eq. 24-27 for Tsuji and Eq. 46-48 for Marshall) which could lead to a factor of sqrt(2) mismatch. However, this could also be due to solving particle-to-wall impact vs. particle-to-particle impact. 

I'll look a bit more at their expressions (or AI could), but can you ask the AI to rerun the test for the particle-to-wall scenario and see if that works? 

Edit: there is in fact a factor of sqrt(2) in their different formulations lurking around. I think it's in the convention of whether you use a particle mass or an effective mass. I'll confirm.

akohlmey:
@jtclemm We ran the particle-to-wall scenario, plus a mass-ratio sweep and
`coeff_restitution` controls (same material parameters as before:
hertz/material, E = 7e10, nu = 0.33, d = 5 mm, rho = 2700):

| scenario | damping | e requested | e measured |
|---|---|---|---|
| two spheres, equal mass | tsuji | 0.5 | 0.61605 |
| two spheres, mass ratio 3 | tsuji | 0.5 | 0.61608 |
| sphere on plane wall | tsuji | 0.5 | 0.61608 |
| two spheres | tsuji | 0.9 | 0.92836 |
| sphere on plane wall | tsuji | 0.9 | 0.92837 |
| two spheres (control) | coeff_restitution | 0.5 | 0.49994 |
| sphere on wall (control) | coeff_restitution | 0.5 | 0.49996 |

So particle-to-wall does NOT behave differently: both geometries are
under-damped by the same amount (identical to five digits), and the result
is independent of the mass ratio.

That is actually the expected outcome given how the code is written, and it
narrows down where the sqrt(2) lives.  In LAMMPS the meff passed to the
damping prefactor is always the true inertial mass of the contact: the
reduced mass mi*mj/(mi+mj) in pair granular and the particle mass rmass[i]
in fix wall/gran.  Since the same mass governs the contact dynamics, the
dimensionless problem depends on alpha alone, and e must come out the same
for pair, wall, and any mass ratio -- which is what we measure.  The LAMMPS
meff plumbing is therefore self-consistent and correct.

The sqrt(2) must then sit entirely in the alpha(e) fit convention: Marshall
Eq. 53 corresponds to a formulation whose damping-prefactor mass is twice
the dynamic mass (e.g. the single-particle mass in the prefactor combined
with reduced-mass two-particle dynamics).  That agrees with your Fig. 3
extraction, where Tsuji's own data follows the self-consistent curve and
Marshall's fit sits sqrt(2) low.

Practical consequence: a uniform `damp *= sqrt(2)` (or a refitted
polynomial) corrects pair and wall contacts alike; no geometry-dependent
correction is needed and the meff code paths should stay as they are.

jtclemm:
> @jtclemm We ran the particle-to-wall scenario, plus a mass-ratio sweep and
> `coeff_restitution` controls (same material parameters as before:
> hertz/material, E = 7e10, nu = 0.33, d = 5 mm, rho = 2700):

Thanks!  A helpful sanity check.



> The sqrt(2) must then sit entirely in the alpha(e) fit convention: Marshall
> Eq. 53 corresponds to a formulation whose damping-prefactor mass is twice
> the dynamic mass (e.g. the single-particle mass in the prefactor combined
> with reduced-mass two-particle dynamics). That agrees with your Fig. 3
> extraction, where Tsuji's own data follows the self-consistent curve and
> Marshall's fit sits sqrt(2) low.

Good to know the diagnosis is all in agreement. Should be patched in the other PR.

jtclemm (approved):
I did not double check the formulae for the analytic tests, but even without verifying the result against the literature the new additions can serve as regression tests.

Maybe worth considering whether we want to flag tests that have not been fully human vetted. I suppose the 'generated' flag isn't quite appropriate though.

akohlmey:
> Maybe worth considering whether we want to flag tests that have not been fully human vetted. I suppose the 'generated' flag isn't quite appropriate though.

@jtclemm That is *very* easy to do. The tags are just arbitrary text. adding more does not change anything outside that you can use them with CTest to either select or exclude tests. For example we use the tag "unstable" to flag tests that fail on ARM64 and other non-x86 platforms due to differences in the (optimized) floating point math. Some tags have a meaning as the generic test programs will switch settings because of that. But adding new tags is no problem at all.

I can add "unvetted" for example

jtclemm:
> I can add "unvetted" for example

That would be a nice addition to track them. Can easily then remove it down the road as tests are checked.

akohlmey:
> > I can add "unvetted" for example
> 
> That would be a nice addition to track them. Can easily then remove it down the road as tests are checked.

done!

jtclemm:
Thanks!

PR #5116, opened 2026-07-29
Fixing small bugs

jtclemm:
**Summary**

Fixes bugs in two parts of LAMMPS

**Issue 1**

Regions that extend into periodic dimensions can return incorrect or even undefined forces if particles briefly extend beyond the boundary. Two scripts are attached to demonstrate the issue by moving a particle along a block region wall across a periodic boundary. The first script uses fix wall/region and fails with an error. The second script uses fix wall/gran and a nan force can be seen when the particle sits exactly on the boundary. When it moves further, forces have an aphysical tangential force until LAMMPS remaps the particle.

The nan issue originates from this block of code in `Region::surface()`
```
    // one of surface_int/ext() will return 0
    // so no need to worry about offset of contact indices
    ncontact = surface_exterior(xnear, cutoff) + surface_interior(xnear, cutoff);
```
Essentially, when exactly on the boundary this is not true and both surfaces can return 1. The interior contact overwrites the exterior contact and then LAMMPS tries to process an undefined second contact. This was fixed by breaking this into a two step process where the interior contact is only checked if no exterior contact exists.

The incorrect orientation was fixed by remapping the position of an atom within the box within `Region::match` and `Region::surface`. This is necessary as some (or maybe all?) region subclasses assume the region ends at the boundary when doing an exterior check.

**Issue 2**

As identified in #5113, there is a missing factor of sqrt(2) in the Tsuji damping force calculation in GRANULAR. This seems to be due to a mismatch in conventions between two literature papers in defining a variable alpha in terms of the particle mass vs. the effective mass. See the other PR for more details and confirmation in the patch. 

**Author(s)**

Joel Clemmer (SNL)

**Licensing**

AI detected that there was an issue in Tsuji damping in #5113. AI was not used in generating changes in this PR.

**Backward Compatibility**

The in.pour.box and in.pour.heat reference outputs were updated since the tsuji damping is  more dissipative with the changes in this pull request.

[in.chute_atomic.txt](https://github.com/user-attachments/files/30523112/in.chute_atomic.txt)
[in.chute_gran.txt](https://github.com/user-attachments/files/30523114/in.chute_gran.txt)


src/GRANULAR/gran_sub_mod_damping.cpp
```diff
@@ -136,10 +136,14 @@ void GranSubModDampingTsuji::init()
...
-  damp += -22.348 * cube(tmp) + 27.467 * powint(tmp, 4);
-  damp += -18.022 * powint(tmp, 5) + 4.8218 * powint(tmp, 6);
+  // Eq. 53 from Marshall 2009
+  double cor = gm->normal_model->get_damp();
+  damp = 1.2728 - 4.2783 * cor + 11.087 * square(cor);
+  damp += -22.348 * cube(cor) + 27.467 * powint(cor, 4);
+  damp += -18.022 * powint(cor, 5) + 4.8218 * powint(cor, 6);
+
+  // multiply sqrt(2) for effective mass convention (as in Tsuji 1992)
+  damp *= M_SQRT2;
```
akohlmey:
```suggestion
  damp *= MY_SQRT2;
```

Please use math constants from the MathConst namespace.
Windows doesn't by default include the M_* constants and we want
those to be consistent across platforms and not implementation dependent.

jtclemm: Ah sorry, I went with the autocomplete and didn't think. Fixed

akohlmey: No worries, that is what code review is for.

akohlmey (approved):

PR #5118, opened 2026-07-30
Add GFN-xTB QM/MM with PPPM electrostatics

njzjz:
## Summary

This pull request adds the CMake-only `QMMM-XTB` package and the `fix qmmm/xtb` command for self-consistent periodic GFN1-xTB and GFN2-xTB QM/MM simulations with PPPM electrostatics.

The implementation provides:

- GFN1-xTB and GFN2-xTB through an in-process libxtb interface;
- electrostatic embedding of MM point charges in the xTB SCC calculation;
- a PPPM source/sensor potential API for evaluating the MM potential at zero-charge QM atoms;
- direct-Ewald treatment of periodic QM images during SCC convergence;
- energy and force de-duplication between xTB, the real-space pair style, and PPPM;
- implicit TIP4P M-site coordinates and force redistribution through the native TIP4P KSpace interface;
- MPI support, with the xTB calculation performed once and its results broadcast to all ranks; and
- unit tests, documentation, and a runnable example.

Compatible libxtb 6.7 or newer Fortran modules are required. libxtb must be built with `INSTALL_MODULES=ON` using a Fortran compiler compatible with the one used to build LAMMPS.

## Variational periodic embedding

The SCC callback applies

```text
phi_shift = Delta phi_MM,PME + A_QM,image q_QM

E_shift = q_QM^T Delta phi_MM,PME
        + 1/2 q_QM^T A_QM,image q_QM
```

Consequently, the SCC Hamiltonian shift, periodic embedding energy, and explicit periodic forces are derivatives of the same variational expression. The converged QM charges are then used for the production PPPM calculation, with overlapping pair, reciprocal-space, and QM-image terms removed from the final energy and forces.

## TIP4P support

For an implicit TIP4P molecule in the MM region, the oxygen charge is evaluated at the virtual M site returned by the KSpace interface. Point-site gradients from xTB are redistributed to O/H/H using the same linear mapping as the selected LAMMPS TIP4P pair and KSpace styles.

Supported pair styles include `tip4p/long` and `lj/cut/tip4p/long`, including their standard accelerated suffixes. An implicit TIP4P molecule may not intersect the QM group.

## Current scope and limitations

- Three-dimensional periodic orthogonal simulation boxes
- `real` and `metal` units
- PPPM-compatible long-range Coulomb pair styles
- One `fix qmmm/xtb` instance
- Fixed simulation box
- No energy minimization or r-RESPA support
- Static QM group

### Point-charge MM environment

The first comparison used an H2 QM region with Na+ and Cl- MM point charges in a 20 Angstrom periodic box. Both programs used an 8 Angstrom cutoff, Ewald coefficient 0.5 1/Angstrom, direct-image `kmax = 8 8 8`, `ksqmax = 64`, a 128^3 PME/PPPM mesh with order 7, xTB accuracy `1e-3`, an electronic temperature of 300 K, and `mmhardness = 0`.

| Method | sander total (kcal/mol) | LAMMPS total (kcal/mol) | Maximum charge difference (e) | Maximum force difference (kcal/mol/Angstrom) |
| --- | ---: | ---: | ---: | ---: |
| GFN1-xTB | -695.9310725629383 | -695.9326464650694 | 4.49e-10 | 4.06e-6 |
| GFN2-xTB | -662.5485478165226 | -662.5501217389989 | 6.34e-10 | 3.08e-6 |

The approximately `1.574e-3 kcal/mol` raw energy offset follows from the different electrostatic conversion constants: `18.2223^2 = 332.05221729` in the reference calculation and `332.06371` in LAMMPS `real` units. Correcting this convention leaves residual energy differences of `1.38e-6 kcal/mol` for GFN1-xTB and `1.36e-6 kcal/mol` for GFN2-xTB.

### TIP4P water in the MM region

The second comparison used an H2 QM region and one TIP4P water molecule in the MM region. sander represents the system with six topology atoms, including an explicit extra point; LAMMPS uses five explicit atoms and generates the M site implicitly. All PME/PPPM and xTB settings were identical to those listed above.

| Method | sander total (kcal/mol) | LAMMPS total (kcal/mol) | Absolute energy difference (kcal/mol) | Maximum QM charge difference (e) | Maximum QM force difference (kcal/mol/Angstrom) |
| --- | ---: | ---: | ---: | ---: | ---: |
| GFN1-xTB | -649.9769223348 | -649.9768729235 | 4.94e-5 | 4.04e-11 | 3.15e-6 |
| GFN2-xTB | -616.5108234132 | -616.5107740189 | 4.94e-5 | 1.14e-10 | 1.95e-6 |

For the rigid TIP4P molecule, the physically relevant resultant force and torque agree as follows:

| Method | Maximum resultant-force difference (kcal/mol/Angstrom) | Torque difference (kcal/mol) |
| --- | ---: | ---: |
| GFN1-xTB | 2.39e-8 | 5.07e-8 |
| GFN2-xTB | 2.51e-8 | 5.51e-8 |

The individual O/H/H Cartesian forces differ because sander uses a fixed-distance local-frame extra point, whereas native LAMMPS implicit TIP4P uses a linear virtual-site coordinate. At the rigid equilibrium geometry both definitions generate the same M-site position, and the remaining per-atom difference is an internal constraint force: the total force and torque agree to approximately `1e-8`.

The GFN2-xTB TIP4P calculation was additionally run with one and two MPI ranks, producing identical results to machine precision.

## Testing

- PPPM source/sensor potential compared with a reciprocal-energy finite difference
- `pppm/tip4p` virtual M-site potential test
- Direct-Ewald QM-image response unit tests
- Single-precision FFT-aware finite-difference coverage
- CMake build and unit tests
- Legacy GNU Make compatibility build
- C++23, VLA, formatting, Linux, macOS, and Windows CI workflows

## Related issue

Design discussion: https://github.com/njzjz/lammps/issues/81

## Author

Jinzhe Zeng (`jinzhe.zeng@ustc.edu.cn`)

## Licensing

By submitting this pull request, I agree that this contribution may be included in LAMMPS and redistributed under either the GNU General Public License version 2 or the GNU Lesser General Public License version 2.1.

## Artificial intelligence tools

Codex was used for implementation, tests, documentation, and validation under the author's direction and review.

Coding agent: Codex
Codex version: codex-cli 0.144.6
Model: gpt-5.6-sol
Reasoning effort: xhigh

## Backward compatibility

This change is backward compatible. It adds a new optional package, a new fix style, and backward-compatible KSpace virtual methods; existing input scripts are unaffected.

## Post-submission checklist

- [x] The feature is complete
- [x] Licensing and corresponding-author information are included
- [x] Source code follows the LAMMPS formatting guidelines
- [x] Documentation is included and integrated into the documentation build
- [x] The CMake build has been verified
- [x] Unit tests have been added
- [x] A package README is included
- [x] A runnable example input is included

## Reference

GFN-xTB QM/MM with PME: https://pubs.acs.org/doi/10.1021/acs.jpcb.4c01466


src/QMMM-XTB/fix_qmmm_xtb.h
```diff
@@ -0,0 +1,101 @@
...
+// clang-format off
+FixStyle(qmmm/xtb,FixQMMMXTB);
+// clang-format on
+#else
+
+#ifndef LMP_FIX_QMMM_XTB_H
+#define LMP_FIX_QMMM_XTB_H
+
+#include "fix.h"
+#include "qmmm_xtb_ewald.h"
```
akohlmey: Package specific headers must not be included in the style header to avoid namespace conflicts. Is there anything that is actually used from that header?
It usually is sufficient to move to the .cpp file and add a forward declaration, if needed.

src/KSPACE/pppm.cpp
```diff
@@ -811,6 +811,84 @@ void PPPM::compute(int eflag, int vflag)
   if (slabflag == 1) slabcorr();
 }
 
+/* ----------------------------------------------------------------------
```
akohlmey: We generally prefer to not have changes to base classes for package specific needs. Instead you should create a derived class, PPPM/xtb, and add the functionality there.

src/KSPACE/pppm_tip4p.cpp
```diff
@@ -40,6 +40,25 @@ PPPMTIP4P::PPPMTIP4P(LAMMPS *lmp) : PPPM(lmp)
   tip4pflag = 1;
 }
 
+/* ----------------------------------------------------------------------
```
akohlmey: same as for PPPM, please create a derived class in your package folder.

cmake/CMakeLists.txt
```diff
@@ -336,6 +336,7 @@ set(STANDARD_PACKAGES
   PYTHON
   QEQ
   QMMM
+  QMMM-XTB
```
akohlmey, line 348: The package should also be added to the `all-on.cmake` and `all-off.cmake` CMake preset files

unittest/commands/CMakeLists.txt
```diff
@@ -54,6 +54,13 @@ target_compile_definitions(test_kspace_group_group_slab PRIVATE -DTEST_INPUT_FOL
 target_link_libraries(test_kspace_group_group_slab PRIVATE lammps GTest::GMock)
 add_test(NAME KSpaceGroupGroupSlab COMMAND test_kspace_group_group_slab)
 
+add_executable(test_kspace_group_potential test_kspace_group_potential.cpp)
```
akohlmey: Adding this test must be gated on the QMMM-XTB package being enabled.

src/kspace.cpp
```diff
@@ -150,6 +150,18 @@ KSpace::KSpace(LAMMPS *lmp) :
   dgcons[6][5] = 693.0 / 256.0;
 }
 
+/* ----------------------------------------------------------------------
```
akohlmey: Same as with PPPM please add such functions to a derived class

src/kspace.h
```diff
@@ -187,6 +193,14 @@ class KSpace : protected Pointers {
   virtual void compute(int, int) = 0;
   virtual void compute_group_group(int, int, int) {};
 
+  // Compute the reciprocal-space electrostatic potential generated by a
```
akohlmey: Same as with the others, we prefer to have no changes to the base class. The only acceptable change would be adding a `xtbflag` variable which is set to 0 in the constructor and then should be set to 1 in your derived classes, so you can check if the current kspace style is compatible. With the other flags, you can ensure that it is a PPPM and - if needed - a TIP4P compatible style and then perform a suitable dynamic cast to access the specific kspace style and its added functionality.

akohlmey (changes requested):
Your package should be as much encapsulated as possible and the impact to the rest of LAMMPS minimal. I've made suggestions to that effect. Any changes to the LAMMPS core need a sound technical motivation.

src/QMMM-XTB/fix_qmmm_xtb.h
njzjz: Done. I removed the package-specific headers from `fix_qmmm_xtb.h`. The header now uses forward declarations, while the concrete PPPM headers are included only by the implementation file.

https://github.com/lammps/lammps/blob/5d990e1fe05caea710a852d5da7b49edd7acb508/src/QMMM-XTB/fix_qmmm_xtb.h#L18-L29
https://github.com/lammps/lammps/blob/5d990e1fe05caea710a852d5da7b49edd7acb508/src/QMMM-XTB/fix_qmmm_xtb.cpp#L18-L28

src/KSPACE/pppm.cpp
njzjz: Done. The xTB-specific functionality now lives in the package-local `PPPMXTB` class, registered as `pppm/xtb`. The corresponding changes have been removed from the core PPPM implementation.

https://github.com/lammps/lammps/blob/5d990e1fe05caea710a852d5da7b49edd7acb508/src/QMMM-XTB/pppm_xtb.h#L9-L31
https://github.com/lammps/lammps/blob/5d990e1fe05caea710a852d5da7b49edd7acb508/src/QMMM-XTB/qmmm_xtb_pppm.cpp#L195-L211

src/KSPACE/pppm_tip4p.cpp
njzjz: Done. I added the package-local `PPPMTIP4PXTB` class, registered as `pppm/tip4p/xtb`, and removed the xTB-specific changes from the core `PPPMTIP4P` implementation.

https://github.com/lammps/lammps/blob/5d990e1fe05caea710a852d5da7b49edd7acb508/src/QMMM-XTB/pppm_tip4p_xtb.h#L9-L31
https://github.com/lammps/lammps/blob/5d990e1fe05caea710a852d5da7b49edd7acb508/src/QMMM-XTB/qmmm_xtb_pppm.cpp#L213-L257

cmake/CMakeLists.txt
njzjz: Done. `QMMM-XTB` has been added to both requested package presets. I also added it to `nolib.cmake` because it requires external xTB libraries.

https://github.com/lammps/lammps/blob/5d990e1fe05caea710a852d5da7b49edd7acb508/cmake/presets/all_on.cmake#L78-L85
https://github.com/lammps/lammps/blob/5d990e1fe05caea710a852d5da7b49edd7acb508/cmake/presets/all_off.cmake#L76-L83
https://github.com/lammps/lammps/blob/5d990e1fe05caea710a852d5da7b49edd7acb508/cmake/presets/nolib.cmake#L20-L27

unittest/commands/CMakeLists.txt
njzjz: Done. The group-potential test target and test registration are now created only when `PKG_QMMM-XTB` is enabled.

https://github.com/lammps/lammps/blob/5d990e1fe05caea710a852d5da7b49edd7acb508/unittest/commands/CMakeLists.txt#L57-L66

src/kspace.cpp
njzjz: Done. The group-potential mesh operations have been removed from the base `KSpace` class. They are now implemented by a package-local helper shared by the two derived PPPM styles.

https://github.com/lammps/lammps/blob/5d990e1fe05caea710a852d5da7b49edd7acb508/src/QMMM-XTB/qmmm_xtb_pppm.cpp#L29-L39
https://github.com/lammps/lammps/blob/5d990e1fe05caea710a852d5da7b49edd7acb508/src/QMMM-XTB/qmmm_xtb_pppm.cpp#L51-L193

src/kspace.h
njzjz: Done. The only xTB-specific state retained in the base `KSpace` class is `xtbflag`, initialized to zero. Both derived xTB styles set it to one. `fix qmmm/xtb` checks the compatibility flags and then dynamically casts to the appropriate concrete PPPM implementation.

https://github.com/lammps/lammps/blob/5d990e1fe05caea710a852d5da7b49edd7acb508/src/kspace.h#L77-L86
https://github.com/lammps/lammps/blob/5d990e1fe05caea710a852d5da7b49edd7acb508/src/kspace.cpp#L44-L47
https://github.com/lammps/lammps/blob/5d990e1fe05caea710a852d5da7b49edd7acb508/src/QMMM-XTB/fix_qmmm_xtb.cpp#L215-L225
https://github.com/lammps/lammps/blob/5d990e1fe05caea710a852d5da7b49edd7acb508/src/QMMM-XTB/qmmm_xtb_pppm.cpp#L195-L220

src/QMMM-XTB/README
```diff
@@ -0,0 +1,46 @@
+QMMM-XTB package
+================
+
+This CMake-only package provides `fix qmmm/xtb`, an in-process GFN1-xTB and
+GFN2-xTB QM/MM implementation with self-consistent periodic PPPM embedding.
+The SCC adapter uses libxtb's Fortran module interface because the public C
+API does not expose the atom-dependent potential callback needed during every
+SCC iteration.
+
+Requirements
```
akohlmey, line 10: The build instructions from this README should be included in doc/src/Build_extras.rst

njzjz: Added in d2289617db09632b688236b597b08b8f925ef7f2

akohlmey (approved):

src/QMMM-XTB/qmmm_xtb_ewald.h
```diff
@@ -0,0 +1,49 @@
...
+
+#ifndef LMP_QMMM_XTB_EWALD_H
+#define LMP_QMMM_XTB_EWALD_H
+
+#include <array>
+#include <vector>
+
+namespace LAMMPS_NS {
+
+// Small direct-Ewald operator used inside the xTB SCC loop.  PME remains the
```
stanmoore1: PME should be PPPM (this is just one of several places).

stanmoore1 (approved):
This looks fine to me, expect "PME" should be changed to "PPPM" in several places

PR #5120, opened 2026-08-01
[BUGFIX] Correct harmonic restoring force in fix spring/chunk

### LINKED ISSUE #5119: [BUG] fix spring/chunk applies a constant-magnitude force inconsistent with its harmonic potential
**Summary**

`fix spring/chunk` reports the documented harmonic spring energy
`0.5 * K * r^2`, but normalizes the COM displacement when computing the
force. Consequently, every displaced chunk receives a total restoring force
of magnitude `K`, independent of its displacement. The applied force is not
the negative gradient of the reported potential.

- LAMMPS `4 Jul 2026 - Development - patch_4Jul2026-376-g2cb47c3845`
- Commit `2cb47c3845dedf48f9916e35d912eede9b10c557` from the `develop` branch
- Ubuntu 22.04.5 LTS, Linux 6.8.0-136-generic, x86_64
- Reproduced with one MPI rank

**Expected Behavior**

The manual describes `K` as having force/distance units and gives the
per-chunk energy as

```text
U = 0.5 * K * r^2,
```

where `r = |Rm - R0m|`. Its negative gradient, distributed by atomic mass,
is

```text
Fi = -K * (Rm - R0m) * Mi / Mm.
```

For one atom, `K = 2`, and an x displacement of `0.25`, the expected values
are therefore `U = 0.0625` and `Fx = -0.5`.

Manual: https://docs.lammps.org/latest/fix_spring_chunk.html

**Actual Behavior**

The current implementation computes each force component with `dx/r`,
`dy/r`, or `dz/r`:

```c++
fcom[m][0] = k_spring*dx/r / masstotal[m];
fcom[m][1] = k_spring*dy/r / masstotal[m];
fcom[m][2] = k_spring*dz/r / masstotal[m];
esprings += 0.5*k_spring*r*r;
```

Thus the total force on a non-empty chunk is

```text
Fchunk = -K * (Rm - R0m) / r,
```

whose magnitude is always `K` for nonzero displacement. This force would
correspond to a linear potential `U = K*r`, not the reported quadratic
potential.

The reproducer gives:

| x displacement | Expected Fx | Actual Fx | Reported energy |
| ---: | ---: | ---: | ---: |
| 0.25 | -0.5 | -2.0 | 0.0625 |
| 0.50 | -1.0 | -2.0 | 0.25 |

As an independent force/energy check, a central finite difference around
`x = 0.25` with `h = 1e-4` gives:

```text
U(0.2501) = 0.06255001
U(0.2499) = 0.06245001
-[U(0.2501)-U(0.2499)]/(2h) = -0.5
applied Fx = -2.0
```

Thus the force stored by LAMMPS is demonstrably not the negative derivative of
the scalar energy returned by the same fix.

The `SMALL` clamp also makes the reported energy at exactly zero displacement
`1e-20` for `K = 2`, instead of zero.

This can affect dynamics and makes `fix_modify energy yes` inconsistent with
the forces used during minimization.

**Steps to Reproduce**

Run the following input with one MPI rank:

```bash
lmp -log none -in in.spring_chunk_force_energy_bug.txt
```

Complete minimal input:

```text
units lj
atom_style atomic
boundary f f f

region box block -2 2 -2 2 -2 2
create_box 1 box
create_atoms 1 single 0.0 0.0 0.0 units box
mass 1 1.0

pair_style zero 1.0
pair_coeff * *

compute chunks all chunk/atom type
compute centers all com/chunk chunks
compute total_force all reduce sum fx fy fz

fix integrate all nve
fix spring all spring/chunk 2.0 chunks centers

thermo 1
thermo_style custom step f_spring c_total_force[1]

# Store the initial COM.
run 0

# For K=2 and dx=0.25: expected U=0.0625 and Fx=-0.5.
displace_atoms all move 0.25 0.0 0.0 units box
run 0

# For K=2 and dx=0.50: expected U=0.25 and Fx=-1.0.
displace_atoms all move 0.25 0.0 0.0 units box
run 0
```

The three thermo rows are:

```text
Step  f_spring  c_total_force[1]
0     1e-20      0
0     0.0625    -2
0     0.25      -2
```

- Source lines at the tested commit:
  https://github.com/lammps/lammps/blob/2cb47c3845dedf48f9916e35d912eede9b10c557/src/fix_spring_chunk.cpp#L174-L188
- The related `fix spring` implementation retains the displacement factor and
  is consistent with its quadratic energy:
  https://github.com/lammps/lammps/blob/2cb47c3845dedf48f9916e35d912eede9b10c557/src/fix_spring.cpp#L184-L202

A minimal correction is to remove `/r` from the three `fcom` expressions and
compute the energy from the unclamped squared displacement.

[in.spring_chunk_force_energy_bug.txt](https://github.com/user-attachments/files/30616213/in.spring_chunk_force_energy_bug.txt)

csmonago:
**Summary**

Correct the restoring force applied by `fix spring/chunk` so that it is the
negative gradient of the documented and reported harmonic potential.

The previous implementation normalized the COM displacement by its magnitude,
which made the total force on every displaced chunk have constant magnitude
`K`. The corrected force is linear in the COM displacement.

**Related Issue(s)**

Fixes #5119.

**Author(s)**

Carlos Monago, UNED

Corresponding author: csmonago@fisfun.uned.es

**Licensing**

OpenAI Codex was used to analyze the force/energy inconsistency and draft the
source and regression-test changes. The contributor reviewed the analysis,
inspected the changes, and validated the behavior with a locally compiled
LAMMPS binary.

**Backward Compatibility**

Input syntax and documented behavior are unchanged. Numerical trajectories that
use `fix spring/chunk` will change because this corrects the force from a
constant-magnitude radial force to the documented harmonic force. This is an
intentional bug fix; simulations that relied on the previous undocumented
behavior will not reproduce their old trajectories.

**Detailed Description**

For COM displacement `d = Rm - R0m`, the fix reports

```text
U = 0.5 * K * |d|^2.
```

The old code applied

```text
Fchunk = -K * d/|d|,
```

while the consistent harmonic force is

```text
Fchunk = -K * d.
```

This pull request removes the displacement normalization from the three force
components. It also computes the energy from the unclamped squared displacement,
so the spring energy is exactly zero at the reference COM.

The regression test uses a one-atom, one-chunk system with `K=2`. At COM
displacements `0.25` and `0.50`, it checks the exact force/energy pairs
`(-0.5, 0.0625)` and `(-1.0, 0.25)`. It also checks the applied force against a
central finite difference of the fix energy. The existing
`fix-timestep-spring_chunk` reference was regenerated because it recorded the
buggy trajectory.

Verification performed:

- rebuilt the CMake `lmp` target successfully
- minimal one-atom reproducer gives `(Fx,U) = (0,0)`, `(-0.5,0.0625)`, and
  `(-1.0,0.25)` at displacements `0`, `0.25`, and `0.50`
- `ctest -R '^(FixSpringChunk|FixTimestep:spring_chunk)$'` passes both tests;
  the generic fix test includes Verlet and r-RESPA coverage
- LAMMPS `check-whitespace` and `check-permissions` pass
- `git diff --check` passes

No documentation behavior change is needed because the existing manual already
describes the harmonic force and quadratic energy implemented by this fix.

**Post Submission Checklist**

- [x] The bug fix is complete
- [x] Suitable tests have been added to the unittest tree


jtclemm (approved):
Thanks for spotting the error, reporting it, and fixing it!

PR #5124, opened 2026-08-05
Fix multiple bugs in the ML-POD package

### LINKED ISSUE #5114: [BUG] `pair_style pod` crashes (segfault or heap corruption) on systems with 1-2 atoms per neighbor-list candidate count
**Summary**

`pair_style pod` (ML-POD package) crashes the LAMMPS process for systems where at least one atom has a very small number of neighbor candidates in LAMMPS's own neighbor list (`numneigh[i]`) and/or the model has too many high-order body features and the number of neighbors is too small:

- If `numneigh[i] == 0` for some atom, LAMMPS crashes with a segmentation fault (null-pointer write) inside `PairPOD::lammpsNeighborList` ; exit code 139
- If `numneigh[i]` is small but nonzero (in my test, equal to 1, with the default POD body-order settings): LAMMPS aborts on heap corruption (or, with a less strict allocator, silently corrupts adjacent heap memory) due to a heap-buffer-overflow inside `EAPOD::peratomenergyforce2` ; exit code 134

I traced the failtures by rebuilding LAMMPS with AddressSanitizer. The crash is driven by the descriptor complexity (its body-order settings), and by LAMMPS's pre-cutoff-filtered neighbor candidate count.

- LAMMPS version: `22 Jul 2025 - Update 4`
- Package required: `ML-POD`
- OS: Linux, Debian 12, x86_64
- Compiler (used for the diagnostic rebuild): GCC 12.2.0
- Build: serial (`LAMMPS MPI STUBS`), single thread

**Expected Behavior**

`pair_style pod` should compute correctly for any physically sensible configuration, including a single isolated atom in a large periodic cell, or two atoms far enough apart that each has only one neighbor within the cutoff.

**Actual Behavior**

Instead, the LAMMPS process crashes (segmentation fault) or aborts (heap corruption) before anything diagnostic can be printed. 

1. **Null-pointer write** in `PairPOD::lammpsNeighborList`, triggered whenever some atom has zero neighbor-list candidates (`numneigh[i] == 0`), because the temp buffers (`tmpmem`/`tmpint`) are never allocated in that case (a pre-existing `nijmax = 0` initial value defeats the `if (nijmax < jnum)` reallocation guard when `jnum` is also `0`).
2. **Heap-buffer-overflow** in `EAPOD::peratomenergyforce2`, triggered whenever the allocated size of the `bdd` scratch buffer (`3*Nj*Mdesc` doubles, sized by the neighbor count) is smaller than the fixed-size `forcecoeff` region it is also used to store (`(nl2+nl3+nl4) + nelements*K3*nrbf3` doubles, which does not depend on the neighbor count).

**Steps to Reproduce**

A LAMMPS binary built with the `ML-POD` package. Some files are attached. Short version:

```bash
lmp -in case1_single_atom.in
# -> "Segmentation fault (core dumped)", exit code 139
# (1 Cu atom, 20x20x20 Angstrom periodic cell)

lmp -in case2_two_atoms.in
# -> "Aborted (core dumped): int_mallinfo(): unaligned fastbin chunk detected"
#    exit code 134
# (2 Cu atoms, 1.8 Angstrom apart, same cell)

lmp -in case3_three_atoms.in
# -> exit code 0
# control system, 3 Cu atoms
```

The potential used (`POD.descriptor` + `POD.model`, included) is a simple Cu potential (`rcut=3.5` Angstrom, default POD body-order settings: two-, three-, and four-body terms all enabled), fit on 10 rattled bulk-Cu reference configurations.

Also included, `case2_twobody.in` re-runs the identical 2-atom geometry against a variant of the same potential with three-/four-body terms disabled. This one runs cleanly (exit code 0), showing that the crash threshold depends on the potential's descriptor complexity, not on the atom count alone.

[pod_bug.tar.gz](https://github.com/user-attachments/files/30496965/pod_bug.tar.gz)

*Root cause 1*

`PairPOD::compute()` sizes a set of shared scratch buffers based on `jnum = numneigh[i]`, the LAMMPS neighbor-list candidate count for the atom currently being processed:

```cpp
int jnum = numneigh[i];

// allocate temporary memory
if (nijmax < jnum) {
  nijmax = MAX(nijmax, jnum);
  fastpodptr->free_temp_memory();
  fastpodptr->allocate_temp_memory(nijmax);
}

double *rij1 = &fastpodptr->tmpmem[0];
...
int *ti1 = &fastpodptr->tmpint[2*nijmax];
...
lammpsNeighborList(rij1, ai1, aj1, ti1, tj1, x, firstneigh, type, map,
                    numneigh, rcutsq, i);
```

`nijmax` is initialised to `0` in the constructor 
and only ever grows (`nijmax = MAX(nijmax, jnum)`) when a strictly
larger `jnum` is seen than any atom processed so far. For a genuinely
isolated atom (`jnum == 0`), the guard `if (nijmax < jnum)` evaluates
`0 < 0` on the very first call, so `allocate_temp_memory` is **never invoked**, and
`fastpodptr->tmpmem`/`fastpodptr->tmpint` remain `nullptr`.

`lammpsNeighborList` then unconditionally writes through `ti1` before
its neighbor loop even starts:

```cpp
void PairPOD::lammpsNeighborList(double *rij1, int *ai1, int *aj1,
                                  int *ti1, int *tj1, ...)
{
  nij = 0;
  int itype = map[atomtypes[gi]] + 1;
  ti1[nij] = itype;               // unconditional
  int m = numneigh[gi];
  for (int l = 0; l < m; l++) { ... }   // m == 0 here; never runs
}
```

With `ti1 == &tmpint[2*0] == nullptr`, this is a write through a null
pointer.

Output of AddressSanitizer:

```
==82609==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000
==82609==The signal is caused by a WRITE memory access.
    #0 ... LAMMPS_NS::PairPOD::lammpsNeighborList(...) pair_pod.cpp:367
    #1 ... LAMMPS_NS::PairPOD::compute(int, int) pair_pod.cpp:206
    #2 ... LAMMPS_NS::Verlet::setup(int) verlet.cpp:140
```

*Possible fix:* the reallocation guard should trigger on
the first call regardless of `jnum`'s value (e.g. initialise
`nijmax = -1` instead of `0`, or track allocation state with a
separate boolean), so a buffer of at least size 1 always exists
before the unconditional `ti1[0] = itype;`.

---

*Root cause 2*

`EAPOD::allocate_temp_memory(int Nj)`:

```cpp
void EAPOD::allocate_temp_memory(int Nj)
{
  estimate_temp_memory(Nj);
  memory->create(tmpmem, ndblmem, "tmpmem");
  memory->create(tmpint, nintmem, "tmpint");
  memory->create(bd, Mdesc, "bdd");
  memory->create(bdd, 3*Nj*Mdesc, "bdd");   // the buffer that overflows
  memory->create(pd, nClusters, "bdd");
  memory->create(pdd, 3*Nj*nClusters, "bdd");
}
```

`bdd` is allocated with `3*Nj*Mdesc` doubles, where `Nj` here is
`nijmax` (LAMMPS's neighbor-list candidate count for whichever
atom has needed the largest buffer so far; see root cause 1), which is
`>=` the true, cutoff-filtered neighbor count actually used inside the
descriptor math.

Inside `EAPOD::peratomenergyforce2`,
`cb = &bdd[0]` backs several logical sub-buffers, including one whose
required size does not depend on `Nj` at all:

```cpp
double *forcecoeff = &cb[(nl2 + nl3 + nl4)];      
std::fill(forcecoeff, forcecoeff + nelements * K3 * nrbf3, 0.0); 
```

`forcecoeff` needs `(nl2 + nl3 + nl4) + nelements * K3 * nrbf3`
doubles, quantity being fixed by the trained potential's body-order settings,
independent of how many neighbors any atom actually has. The
`3*Nj*Mdesc` formula happens to cover this whenever `Nj` is reasonably
large, but not when it is small.

Output of AddressSanitizer:

```
==82616==ERROR: AddressSanitizer: heap-buffer-overflow ... WRITE of size 8
    #0 ... std::fill<double*, double>(...)
    #3 ... LAMMPS_NS::EAPOD::peratomenergyforce2(...) eapod.cpp:1340
    #4 ... LAMMPS_NS::PairPOD::compute(int, int) pair_pod.cpp:208

0x61c000001740 is located 0 bytes to the right of 1728-byte region [0x61c000001080,0x61c000001740)
allocated by thread T0 here:
    #3 ... LAMMPS_NS::EAPOD::allocate_temp_memory(int) eapod.cpp:2632
```

This is a body-order-complexity threshol. Re-running the identical 2-atom geometry against a
potential with three-/four-body terms disabled (`threebody_number_radial_basis_functions=0`,
`fourbody_number_radial_basis_functions=0`, etc.) does not crash.

With `K3=1, nrbf3=0`, the fixed requirement drops to 8 doubles, well
within the 24 allocated.

The trigger is not simply "fewer than 3 atoms": a 4-, 5-, 6-, or 8-atom linear chain (same potential, same `rcut`) does
not crash even though its end atoms have only 1 true neighbor
within cutoff, because LAMMPS's neighbor list reports a larger raw
candidate count for those atoms (`numneigh[i]=3` in the 4-atom-chain
test) than the true, filtered count.
The precise condition, as far as I can determine, is:

```
3 * numneigh[i] * Mdesc  <  (nl2 + nl3 + nl4) + nelements * K3 * nrbf3
```

for any atom `i` processed during a `compute()` call.

*Possible fix:* size `bdd` (and likely `pdd`, which has the
analogous `3*Nj*nClusters` formula) as
`max(3*Nj*Mdesc, (nl2+nl3+nl4) + nelements*K3*nrbf3)`, or add an
explicit bounds check/assertion comparing the allocated size against
every fixed-size sub-buffer's requirement, so a similar mismatch fails
with a clear LAMMPS error in the future rather than an unexplained
crash.

akohlmey:
**Summary**

This pull request fixes multiple memory management issues in the ML-POD package.

**Related Issue(s)**

Fixes #5114 

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

The detailed error analysis and draft code changes were based on the report in issue #5114 and performed by Claude Code (using Fable 5). Changes were reviewed manually and tested on the bundled examples (using reference data bundled with LAMMPS from a repository of the ML-POD author) and are reproducing the original references to machine precision.

**Backward Compatibility**

Yes.

**Implementation Notes**

In addition to the reported segfault, multiple other, related memory management issues were resolved and memory leaks from element label storage were plugged. The updated code was made consistent across all functionality in the ML-POD package. The unit test is reproduced as well.

**Post Submission Checklist**

### DIFF
--- a/src/ML-POD/compute_pod_atom.cpp
+++ b/src/ML-POD/compute_pod_atom.cpp
@@ -53,8 +53,6 @@ ComputePODAtom::ComputePODAtom(LAMMPS *lmp, int narg, char **arg) :
 
   nmax = 0;
   nijmax = 0;
-  pod = nullptr;
-  elements = nullptr;
 
   size_peratom_cols = podptr->Mdesc * podptr->nClusters;
   peratom_flag = 1;
@@ -64,6 +62,10 @@ ComputePODAtom::ComputePODAtom(LAMMPS *lmp, int narg, char **arg) :
 
 ComputePODAtom::~ComputePODAtom()
 {
+  if (elements) {
+    for (int i = 0; i < atom->ntypes; i++) delete[] elements[i];
+    delete[] elements;
+  }
   memory->destroy(map);
   memory->destroy(pod);
   delete podptr;
@@ -130,23 +132,28 @@ void ComputePODAtom::compute_peratom()
   int Mdesc = podptr->Mdesc;
   double rcutsq = podptr->rcut*podptr->rcut;
 
+  // determine the maximum number of neighbor list candidates for all local atoms
+  // and allocate temporary memory accordingly.  a minimum of one guarantees that
+  // the buffers always exist, even if no atom has any neighbors at all.
+
+  int jnummax = 1;
+  for (int ii = 0; ii < inum; ii++) jnummax = MAX(jnummax, numneigh[ilist[ii]]);
+
+  if (nijmax < jnummax) {
+    nijmax = jnummax;
+    podptr->free_temp_memory();
+    podptr->allocate_temp_memory(nijmax);
+  }
+
+  rij = &podptr->tmpmem[0];
+  tmpmem = &podptr->tmpmem[3*nijmax];
+  ai = &podptr->tmpint[0];
+  aj = &podptr->tmpint[nijmax];
+  ti = &podptr->tmpint[2*nijmax];
+  tj = &podptr->tmpint[3*nijmax];
+
   for (int ii = 0; ii < inum; ii++) {
     int i = ilist[ii];
-    int jnum = numneigh[i];
-
-    // allocate temporary memory
-    if (nijmax < jnum) {
-      nijmax = MAX(nijmax, jnum);
-      podptr->free_temp_memory();
-      podptr->allocate_temp_memory(nijmax);
-    }
-
-    rij = &podptr->tmpmem[0];
-    tmpmem = &podptr->tmpmem[3*nijmax];
-    ai = &podptr->tmpint[0];
-    aj = &podptr->tmpint[nijmax];
-    ti = &podptr->tmpint[2*nijmax];
-    tj = &podptr->tmpint[3*nijmax];
 
     // get neighbor list for atom i
     lammpsNeighborList(x, firstneigh, atom->tag, type, numneigh, rcutsq, i);
--- a/src/ML-POD/compute_pod_global.cpp
+++ b/src/ML-POD/compute_pod_global.cpp
@@ -57,14 +57,16 @@ ComputePODGlobal::ComputePODGlobal(LAMMPS *lmp, int narg, char **arg) :
   cutmax = podptr->rcut;
 
   nijmax = 0;
-  pod = nullptr;
-  elements = nullptr;
 }
 
 /* ---------------------------------------------------------------------- */
 
 ComputePODGlobal::~ComputePODGlobal()
 {
+  if (elements) {
+    for (int i = 0; i < atom->ntypes; i++) delete[] elements[i];
+    delete[] elements;
+  }
   memory->destroy(map);
   memory->destroy(pod);
   delete podptr;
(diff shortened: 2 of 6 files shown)


exapde (approved):
Thank you Axel for fixing the bugs. The fixes look good to me.

sjplimp (approved):

PR #5128, opened 2026-08-13
DIFFRACTION: add compute xrd/fft, support triclinic cells, and fix several defects

stanmoore1:
**Summary**

Adds `compute xrd/fft`, which computes the same X-ray diffraction pattern as `compute xrd` using fast Fourier transforms instead of the direct sum, and fixes a number of defects found in the DIFFRACTION package while working on it.

The direct sum costs one sine and one cosine per (reciprocal lattice node, atom) pair, which is prohibitive for large cells. The sum is a type-weighted non-uniform Fourier transform, so it is evaluated here by spreading the atoms onto a uniform mesh with a Kaiser-Bessel window, taking one FFT per chemical element, and dividing out the analytic Fourier transform of the window. `compute xrd/fft` takes the same arguments as `compute xrd` and produces the same rows in the same order, so it is a drop-in replacement in existing input scripts.

On the shipped `examples/PACKAGES/diffraction/in.BulkNi` (32000 atoms, 90152 nodes) it reproduces the stored 2018 reference histogram to 3.5e-6 in 0.23 s, against 56.5 s for the direct sum. Scaling for face-centered cubic nickel with `2Theta 40 80 c 2 2 2`:

| atoms | nodes | `xrd` | `xrd/fft` | speedup |
|---|---|---|---|---|
| 864 | 2460 | 0.07 s | 0.01 s | 7x |
| 4000 | 11270 | 1.39 s | 0.02 s | 70x |
| 10976 | 30928 | 10.4 s | 0.07 s | 148x |
| 32000 | 90152 | 87.0 s | 0.47 s | 185x |
| 256000 | 719238 | ~1.5 h (extrapolated) | 5.6 s | ~1000x |

The remaining commits are independent of the new style and each stand on their own:

* **Cobalt selected the wrong scattering factors.** `XRDtypeList` listed `"Co"` twice, as the second and third of the three cobalt entries, and the lookup loop had no `break`, so the last match won. The coefficients of the third entry sum to 24 electrons at zero scattering angle and match the tabulated Co(3+) values, so `Co` selected the Co(3+) factors while neutral cobalt, whose coefficients sum to 27, was unreachable, and Co(3+) could not be requested at all. The third entry is now spelled `"Co3+"`.
* **Integer overflows.** `ComputeXRD::init()` walked the search box with a flattened index computed as a product of three `int`s, which overflows once `Knmax` reaches 645 and made the calculation abort with a rows mismatch. Both computes also counted nodes in an `int` and then allocated three `int`s per node without checking, and assigned the `bigint` returned by `group->count()` to an `int`, which is the divisor that normalizes the intensity.
* **The pattern did not follow the box.** The reciprocal lattice was built once, when the compute was defined. Expanding a 3.52 Angstrom face-centered cubic nickel cell by 5% left the strongest reflection at 44.59 degrees instead of moving it to 42.36, and cut its intensity from 4.81e6 to 1.66e6. It now follows the approach of the kspace styles: which nodes are explored stays fixed, since the number of rows of the output array cannot change, but the reciprocal lattice vectors are scaled with the cell.
* **Triclinic cells are supported** by `compute xrd` and `compute xrd/fft`.
* **`compute saed` used the wrong scattering factors above `Kmax` 4.** Two coefficient sets are tabulated per element, one fitted for sin(theta)/lambda below 2 and one for 2 to 6, but the set was chosen once from `Kmax` rather than per node, so a calculation with `Kmax` above 4 used the upper range everywhere.
* **Swapped references.** Each manual page carried the scattering factor references belonging to the other one.

**Related Issue(s)**

None.

**Author(s)**

Stan Moore (stanmoore1@gmail.com), corresponding author for the new files.

The DIFFRACTION package was created by Shawn P. Coleman (scolema3 at gmail.com), with Douglas Spearot, both then at the University of Arkansas. `compute xrd/fft` derives from their `compute xrd`, and its spreading kernel follows the Kaiser-Bessel non-uniform FFT in a `compute structure/factor/fft` prototype written by Stan Moore.

**Licensing**

The code, documentation, tests and example scripts in this pull request were generated with Claude Code (Anthropic) Opus 5, under human direction.

The reference values stored in the new unit tests were not written by hand: `compute-xrd-fft.yaml`, `compute-xrd-fft-default.yaml` and `compute-xrd-fft-triclinic.yaml` store the values produced by the direct sum, so that the tests assert agreement between the FFT path and `compute xrd` rather than agreement with previously recorded output of the FFT path itself.

**Backward Compatibility**

Input scripts are unaffected, but four changes alter results, in each case because the previous result was wrong:

1. `compute xrd` with type `Co` now uses the neutral cobalt scattering factors rather than those of Co(3+). `Co3+` is now a valid type name.
2. `compute saed` with `Kmax` above 4 now selects the scattering factor coefficients per node. The default `Kmax` of 1.7 never reached the affected range.
3. `compute xrd` and `compute saed` under `fix npt`, `fix deform` or `change_box` previously reported the diffraction angles of the box as it was when the compute was defined. `compute xrd` now rescales with the cell; `compute saed` warns, because `fix saed/vtk` derives the same reciprocal lattice independently for its grid headers and the two would otherwise disagree.
4. `compute xrd` now accepts, and ignores, the `order` and `oversample` keywords, which are used by `compute xrd/fft`. Its option loop enumerates the reciprocal lattice nodes and cannot be extended from the derived class, so the keywords are parsed there. This is the one part reviewers may want to see done differently.

**Implementation Notes**

`ComputeXRDFFT` derives from `ComputeXRD` and calls its constructor and `init()`, so the set and ordering of reciprocal lattice nodes is identical by construction rather than by reimplementation.

`compute xrd` samples reciprocal space at multiples of the reciprocal lattice vectors scaled by the `c` parameters. The phase of node index d is 2 pi times the projection of the position on the d-th such vector, and is periodic in that projection, so folding it into one period is exact rather than an approximation. That holds whether the period is longer than the simulation box (small `c` values) or shorter than it (large `c` values, as `in.BulkNi` uses), so one uniform mesh covers every case.

The mesh cannot be decomposed spatially, because when the period is shorter than the box two points far apart in space fold onto the same mesh cell while living on different MPI ranks. It is divided into slabs along z instead: every rank spreads onto a full copy and one `MPI_Reduce_scatter` produces the owned slab, which is also the layout `FFT3d` takes as input and output. With more ranks than mesh planes the trailing ranks own nothing and take no part in the transform, on a communicator holding only the ranks that do, since not every FFT backend accepts an empty block. The mesh therefore depends only on the settings of the compute, not on the number of ranks, and results are reproducible across rank counts.

For a triclinic cell the mesh is built on the rows of the inverse box matrix instead of on the coordinate axes, as the kspace styles do. Those rows are exactly zero off the diagonal for an orthogonal cell, so the same expression covers both cases and reproduces the orthogonal results bit for bit. The index of a node satisfies i = K.a_i/c_i, so the search is bounded by Kmax*|a_i|/c_i, which grows with the tilt the same way the kspace grid has to.

One transform is done per distinct element rather than per atom type, and each atom is spread once in total. The owning rank forms the intensity locally, so the reduction is one value per row.

Verification:

* Against an independent implementation written from the equations on the manual page: 2104 nodes for an orthogonal cell, agreeing to 5e-14 for `compute xrd` and 2e-11 for `compute xrd/fft`; 2236 nodes for a cell tilted in all three directions, agreeing to 4e-13 and 1e-11.
* Against `compute xrd` in the same run across `c` values spanning periods longer than, equal to and shorter than the box, anisotropic `c`, `manual` spacing, both `LP` settings, `2Theta` in degrees and radians, one and two elements, cubic, non-cubic and triclinic cells, periodic and non-periodic directions, crystalline and disordered configurations, a subgroup rather than `all`, and molecular dynamics with the compute invoked every step across several `run` commands.
* Reproducible across 1, 2, 3, 4, 8, 16, 24 and 32 MPI ranks, including cases where more ranks than mesh planes leave some ranks idle.
* By cell choice invariance: tilting xy by a full box length replaces the cell vector a2 by a2+a1, which generates the same lattice and moves no atom, and the histogrammed pattern of `compute xrd` is then unchanged bit for bit.
* The error falls geometrically with the `order` keyword. For a single atom the relative error of the intensity is 1.2e-6, 1.3e-8 and 1.2e-10 at order 7, 9 and 11. For N atoms the error of a strong reflection stays at that level while the relative error of the weak diffuse intensity between reflections grows as the square root of N, which the measurements confirm (1.15e-6 times sqrt(864) predicts 3.3e-5, measured 3.41e-5).
* AddressSanitizer reports no leaks and no memory errors attributable to these changes, checked against a control run containing no computes.
* Both FFT backends available here, the bundled KISS FFT and FFTW3, give identical results; a single precision build works and warns that it limits the accuracy of weak diffuse intensities.
* `examples/PACKAGES/diffraction/in.BulkNi` still reproduces its stored reference histogram exactly.

Not addressed here: `compute saed` still rejects triclinic cells, because `fix saed/vtk` writes its output on an orthogonal image grid which cannot represent an oblique reciprocal lattice, so that needs a different output format rather than only a change to the compute. `fix saed/vtk` has no test coverage at all. `compute xrd` and `compute saed` remain largely duplicated, and a shared non-uniform FFT helper would make a `compute saed/fft` nearly free.

**Post Submission Checklist**

`compute xrd/fft` needs the FFT wrappers of the KSPACE package. Following the precedent of `fix phonon`, a new `src/DIFFRACTION/Install.sh` and `cmake/Modules/Packages/DIFFRACTION.cmake` install it only when KSPACE is present, so DIFFRACTION continues to build on its own. Both directions were tested with CMake and with the traditional make build.

`make html` and `make spelling` complete without new warnings, and `make check` in `doc` passes.


sjplimp (changes requested):
@stanmoore1 Looks like a nice idea for perf improvement.  Have you asked Shawn or Doug to take a look at it ?  Last I knew Doug was at U Florida.

stanmoore1:
@scolema3 do you have any feedback on this pull request? Thank you

PR #5130, opened 2026-08-13

akohlmey:
**Summary**

This pull request contains multiple small changes and fixes

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

**Backward Compatibility**

A few cases, especially for vector and atom style variables, that were flagged as an error are now accepted and return a meaningful result. On the other hand, some other cases that would silently result in a return value of `Inf` now abort with an error message.
All existing unit tests pass. New tests have been added.

**Implementation Notes**

- consistently return 1.0 for any variable expression where a number is taken to the power of 0
- abort with an error when taking 0.0 to a negative power in variable expressions
- silence compiler warnings
- avoid std::move() of a trivially copyable object
- don't use std::move() inside a loop using/modifying the same object
- don't always update the libAtoms git repo to the latest version but pick a tested commit hash
- update the RuNNer library release tag to version 2.0.4
- don't always attempt to build the ML-HDNNP library
- avoid segfaults when accessing fix ave/time data early; return 0.0 instead
- make sure that fix vector data is always initialized to 0 after allocation so it doesn't contain garbage when accessed before data has been added
- address multiple minor issues reported by coverity scan

**Post Submission Checklist**


stanmoore1 (approved):
I approve

Issue #5047, opened 2026-06-18
[Feature Request] Portable data-file restart for BPM internal state via a recovery-proxy fix


akohlmey:
**Summary**

BPM-package simulations can currently only be restarted from binary restart files, because the per-bond reference and history state managed by `FixBondHistory` (the reference length `r0`, plus the evolving VES/EPS/rotational history) is never written to data files. Since data files are portable across LAMMPS versions while binary restarts are not, I would like a portable, text-based way to resume a BPM simulation: persist this internal state as additional data-file *section(s)* recovered through a new, generic `fix bpm/restore` proxy.

**Detailed Description**

This came up while implementing PR #5045 (re-implementing the PERI peridynamics models on the BPM framework). For those models -- and for BPM bond styles in general -- almost all of the state needed to resume a run is already portable:

- atoms / velocities / masses / bond topology -> standard data-file sections;
- per-atom fields such as `vfrac` / `smin` / `lambda` -> `fix property/atom` data-file sections (already supported);
- derived quantities (`wvolume`, `vinter`, ...) are recomputed at setup.

The one piece that is *not* portable is the **per-bond reference state** held by `FixBondHistory`: the reference length `r0` (frozen at the undeformed configuration, so it cannot be reconstructed from the deformed positions stored in a data file) and the evolving per-bond history of the VES/EPS/rotational models. Without it, re-creating the bonds would re-baseline `r0` to the current (deformed) length and silently discard all accumulated deformation.

Proposed approach -- a generic recovery proxy fix, conceptually the data-file analog of how a binary restart hands per-fix data blocks back to the owning fix:

- A new, user-declared `fix bpm/restore` (name TBD) that owns one or more BPM data-file sections via the existing generic `Fix` data-file section API (`write_data_header` / `write_data_section*` / `read_data_section`, gated by `wd_header` / `wd_section`). This is the same additive, format-preserving mechanism already used by `fix cmap` (a per-crossterm section keyed by atom tags) and `fix property/atom`: it *adds a section*, it does not change the data-file format.
- On `write_data`, the proxy asks each registered producer to serialize its state and emits a self-describing section (e.g. a `Bond History` section: one tag-ordered row per bond, `atom1 atom2 h0 ... h{n-1}`, with the column count recorded so a reader can validate it).
[...]

PR #5131, opened 2026-08-14
Add read/write reference file ability to restore BPM internal bond history data 

zawh4159:
**Summary**

This feature adds an ability to read and/or write separate reference files from an optional setting(s) in the BondBPM class. This can be used to restore the per-bond reference state data held by `FixBondHistory.` Two optional settings are added `write/reference` and `read/reference`

`write/reference`
Similar to `store/local` bond data is transferred to an internal fix that can be accessed by other LAMMPS commands. Reference files can then be written using standard `dump local` command.

`read/reference`
Reads a text file in that same format at bond-style construction time and restores the reference state from it.

**Related Issue(s)**

This addresses a commonly requested feature [#5047](https://github.com/lammps/lammps/issues/5047).

**Author(s)**

Zachary White (Sandia National Labs, CU Boulder) zwhite55@gmail.com

**Licensing**

Claude (Anthropic) was used to help write and edit the documentation (the `.rst` files) included in this pull request. All code and modifications to `bond_bpm.cpp`/`bond_bpm.h` were written by the author without AI assistance.

**Backward Compatibility**

Both keywords are opt-in and off by default (`reference_flag`/`id_fix_write_ref`) are unset unless the keyword is given), so existing input scripts and their behavior are unaffected. No existing keyword syntax, restart format, or data structure was changed.

**Implementation Notes**

*Writing:* `write/reference` reuses the existing `FixStoreLocal` machinery already used by `store/local` (broken-bond output), but as a second, independent internal fix (`fix_write_ref`) sized to `nhistory + 2` columns (the two atom IDs plus every history value for that substyle). `BondBPM::write_reference()` is called from `pre_compute()` every `write_ref_freq` steps and, for each locally-owned bond, packs `(tag[i], tag[j], history[0..nhistory))` into the fix via `add_data()`. The user then dumps that fix's columns with a standard `dump local` command; the resulting file is an ordinary dump-local file, so all of LAMMPS's existing tooling for that format applies. Because the reference-file reader locates data using the `ITEM: NUMBER OF ENTRIES` / `ITEM: ENTRIES` header lines that `dump local` produces, `dump_modify ... header no` must not be used on this dump.

*Reading:* `BondBPM::read_reference()` runs on proc 0 (via `TableFileReader`/`ValueTokenizer`) during `settings()`, before the simulation begins. It scans for the `ITEM: NUMBER OF ENTRIES` keyword to get the bond count, then the following `ITEM: ENTRIES` line to infer the number of history columns (`nbonddata`) present in the file, and finally reads that many data lines into two flat arrays: `bListdata` (atom ID pairs) and `bHistdata` (history values). This is broadcast to all processors via `MPI_Bcast`.

*Restoring:* restoration into `FixBondHistory` happens in `BondBPM::restore_data()`, which is called once from `pre_compute()` (guarded by `restore_flag`) in place of the substyle's own `store_data()`. It builds a hash map keyed on `min(tag_i,tag_j)*natoms + max(tag_i,tag_j)` to look up each local bond's row in the file data in O(1), then overwrites both `fix_bond_history`'s per-atom storage and the `bondstore` array with the file's values, bypassing whatever the substyle would otherwise compute from the bonds' current geometry. This is what makes it possible to restore a reference state after a `data file` read (which has no history at all) or to override the state computed from the current configuration with an externally supplied one.

*Column layout:* the reference-file column order after the two atom IDs is substyle- and setting-dependent (e.g. `bpm/rotational` has different column counts under `damping derivative` vs. `damping dem`; `bpm/peri` depends on which constitutive model, `pmb`/`lps`/`eps`/`ves`, is active, and on the largest model in use when bond types are mixed). Two error checks guard against mismatches: the number of history columns in the file must equal `nhistory` for the active substyle/settings, and the number of file entries must equal `atom->nbonds` exactly. `Howto_bpm.rst` documents the full per-substyle column table so a reference file can be constructed or interpreted correctly outside of LAMMPS.

**Post Submission Checklist**


jtclemm (approved):
Thanks for adding this feature @zawh4159 !

I made a few changes, mostly precautionary guards on memory and renaming write/reference -> write/history to be a bit more general. The only other improvement I can think of is adding an example. But that might make more sense to add later (maybe a viscoelastic test where you want to control bond history?).

PR #5132, opened 2026-08-14
Add clamp for Kokkos 

### LINKED ISSUE #5145: [BUG] KOKKOS neighbor binning can access bincount out of bounds when an atom remains multiple periods outside a periodic box
### Summary

A dynamic CUDA/KOKKOS run can terminate with `cudaErrorIllegalAddress` in `NBinKokkos<Cuda>::bin_atoms()` when an owned atom moves by more than one periodic box length before neighbor rebuilding.

There appear to be two related fail-safety issues:

1. `DomainPBCFunctor` applies at most one periodic correction per dimension, so an atom that is multiple periods outside the box can remain outside after `DomainKokkos::pbc()`.
2. `NBinKokkos::binatomsItem()` uses the result of `coord2bin()` to index `bincount[ibin]` without checking `0 <= ibin < mbins`.

The result is an illegal CUDA memory access instead of a controlled LAMMPS error. I am reporting the memory-safety/fail-closed behavior, not claiming that the underlying high-velocity trajectory is physically valid.

### Environment

- LAMMPS: development commit `a5040aae` (`4 Jul 2026 - Development`, locally modified only for diagnosis)
- Kokkos: `5.1.99`
- Kokkos backend: CUDA + Serial, double precision, legacy view layout
- Architecture: `AMPERE80`
- CUDA toolkit: `12.6.85`
- NVIDIA driver: `535.129.03`
- GPU: NVIDIA A800-SXM4-80GB
- Compiler: GCC 11.2, C++20
- One MPI rank, one OpenMP thread, one GPU

The pair style in the reproducer is an external KOKKOS pair style, but the crash occurs during LAMMPS neighbor binning. A non-KOKKOS control using the same potential and coordinates does not use `NBinKokkos` and completes the bounded test.

### Reproducer characteristics

The smallest case observed contains 60,480 atoms in a periodic triclinic box. The relevant input is:

```lammps
units metal
atom_style atomic
boundary p p p
atom_modify map yes
read_data system.data

mass 1 15.999
mass 2 14.007
mass 3 12.011
mass 4 1.008

pair_style deepmd/kk model.pt2
pair_coeff * * O N C H
neighbor 2.0 bin
neigh_modify every 10 delay 0 check yes
comm_modify cutoff 8.0

timestep 0.0002
thermo 100
thermo_style custom step atoms temp pe ke etotal press vol
thermo_modify flush yes lost error

run 0
velocity all create 4000.0 4928459 mom yes rot yes dist gaussian
fix int all nve
fix bath all langevin 4000.0 4000.0 0.1 5928459 zero yes
run 5000
```

Launch command:

```text
lmp -k on g 1 -sf kk -in in.lmp
```

The last completed thermo record is step 3900. Before step 4000, CUDA reports:

```text
cudaDeviceSynchronize() error(cudaErrorIllegalAddress)
NBinKokkos<Kokkos::Cuda>::bin_atoms()
NeighborKokkos::build_kokkos<Kokkos::Cuda>()
VerletKokkos::run()
```

A stationary control with forced rebuilding every step completes 100 neighbor builds at 60,480, 282,240, 399,168, and 504,000 atoms. Setting `nbin/atoms/per/bin 128` does not change the dynamic failure window, so ordinary per-bin occupancy does not explain the error.

### Diagnostic guard result

I added a local diagnostic check before indexing `bincount`:

```cpp
const int ibin = coord2bin(x(i,0), x(i,1), x(i,2));
if (ibin < 0 || ibin >= mbins) {
  // record atom index, ibin and position, then report on host
  return;
}
```

With this check, a deterministic 504,000-atom reproducer reports:

```text
atom index: 38356
ibin: -4461
valid bin range: [0, 21295)
position: [122.12335419024613, 80.8188280020149, -61.73705820551988]
mbinlo: [-3, -2, -2]
mbin: [39, 21, 26]
```

The periodic box z length is `182.796`. The coordinate is finite but remains outside the valid periodic/bin extent.

### Relevant source behavior

In `src/KOKKOS/domain_kokkos.cpp`, `DomainPBCFunctor::operator()` uses one-shot conditions:

```cpp
if (x(i,2) < lo[2]) x(i,2) += period[2];
if (x(i,2) >= hi[2]) x(i,2) -= period[2];
```

By contrast, `DomainKokkos::remap_all()` uses `while` loops and explicitly handles points no matter how far away they are.

In `src/KOKKOS/nbin_kokkos.cpp`, the unmodified path computes `ibin` and immediately accesses:

```cpp
Kokkos::atomic_fetch_add(&bincount[ibin], 1);
```

without an index check. The CPU binning path also checks for non-finite positions before converting coordinates to bins.

### Expected behavior

Even if the simulation is physically unstable, LAMMPS should fail with a controlled diagnostic rather than access GPU memory out of bounds.

Possible fixes/questions:

1. Should the KOKKOS PBC functor use repeated wrapping, matching `remap_all()`, for atoms more than one period outside?
2. Should `NBinKokkos::binatomsItem()` fail closed when `ibin` is outside `[0, mbins)`?
3. Would a device-side finite-coordinate check matching the CPU path also be appropriate?

I can prepare a focused patch and a smaller standalone reproducer if the maintainers agree on the intended behavior.

### Control without KOKKOS

Using the same model with unsuffixed `pair_style deepmd` and no `-k/-sf kk`, LAMMPS selects:

```text
pair build: full/bin/atomonly
bin: standard
```

A 60,480-atom run completes 35,000 steps / 7 ps at 0.2 fs under 4000 K Langevin NVT, retains all atoms, and reports 1,588 neighbor builds with zero dangerous builds. This supports the diagnosis that the illegal memory access is specific to the KOKKOS neighbor/PBC path. It does not imply that every unconstrained NVE trajectory is physically stable.

wcwitt:
**Summary**

I believe this fixes a small bug. I haven't taken the time to understand all possible implications, but I'm reasonably confident in the fix since the strategy is to replicate the logic of the [non-Kokkos version](https://github.com/lammps/lammps/blob/0e2cc2a387b81ea0177228e24b28d511b0f75c18/src/neighbor.cpp#L2453-L2468).

Codex says:
> The Kokkos implementation omitted a clamp present in the non-Kokkos code since 2019, allowing sufficiently large cell deformations to turn a negative neighbor-rebuild threshold positive when squared, incorrectly skip a required rebuild, and calculate energies using a stale neighbor list.

**Related Issue(s)**

It fixes the problem [reported here](https://matsci.org/t/internal-lammps-positions-values-change-between-run-n-and-subsequent-run-0/67415) (confirmed with the author of the thread).

Fixes https://github.com/lammps/lammps/issues/5145

**Author(s)**

Chuck Witt, Harvard University, Stan Moore (SNL)

**Licensing**

I found the explanation with help from OpenAI's Codex. I also asked it to implement the fix, although fairly simple.

Claude Code was used to further restore parity between non-Kokkos and KOKKOS neigh list code, and fix bugs, as well as for testing.

**Backward Compatibility**

Likely backwards incompatible, since it is a bugfix.

**Implementation Notes**

We confirmed it fixes some unexpected behavior [recorded here](https://matsci.org/t/internal-lammps-positions-values-change-between-run-n-and-subsequent-run-0/67415) (confirmed with the author of the thread).

**Post Submission Checklist**


stanmoore1:
@wcwitt good catch, this is a bug for sure. I asked Claude to search for any other non-parity in KOKKOS vs non-Kokkos neighbor code, I have some more changes coming soon, will push here.

wcwitt:
Okay thanks! I don't know this part of LAMMPS well, but it felt right.

bernstei:
To what extent are there tests for this neighbor list rebuild logic?  

I've been thinking about this for other implementations, and in particular, if there are periodic image interactions with neighbors that are interacting through > 1 lattice vector shifts, I don't see how the current logic will catch those - using the corners seems to implicitly assume that there's never more than 1 * cell vector relative displacement.  And modern MLIPs can have very long effective cutoffs, which could lead to such long range interactions. It's subtle, so I'm not saying I'm sure it's wrong, by any means, but I'm worried.

stanmoore1 (approved):

PR #5135, opened 2026-08-16
Continue modernizing internal API accesses and avoid depending on knowledge of core class internals

akohlmey:
**Summary**

This continues the work to use modernized accessors and APIs and the utils namespace functions and parsing tools instead of legacy functions that have insufficient error checks or unexpected behavior or require complex code.

**Related Issue(s)**

N/A

**Author(s)**

Axel Kohlmeyer, Temple U

**Licensing**

Large parts of the changes in this pull request were implemented with the assistance of Claude Code, working from a detailed, human-reviewed plan and with all changes reviewed, built, and tested locally.  The initial commits adjusting the `Modify::delete_fix()`/`delete_compute()` semantics and many changes to introduce clang-format processing, use improved error messages and correct style names were written by hand.

**Backward Compatibility**

All unit tests and regression tests pass as before. Tests with pair style ldd needed an update of the reference because of the parsing now (correctly) parsing the parameters into double precision values instead of single precision

**Implementation Notes**

- make `Modify::delete_compute()` and `Modify::delete_fix()` safe to use in a class's destructor when the list of computes and fixes has already been purged in the Modify class destructor
- replace external accesses to `Modify::nfix/ncompute/fix[]/compute[]/fmask[]` and calls to `Modify::find_fix()/find_compute()` with the `get_fix_by_id()`, `get_fix_by_style()`, `get_fix_list()` accessors (and compute equivalents); a small `set_fix_mask()/clear_fix_mask()` API was added for the few fixes that need to toggle bits in their stored fix mask
- replace use of `atoi()`, `atof()`, `std::stoi()`, `std::stod()`, `sscanf()` in favor of `utils::*numeric()` and `ValueTokenizer`, converting several simple custom file readers to `TextFileReader`/`PotentialFileReader` in the process (LCBOP, ORIENT, BOCS, AMOEBA bitorsion, bond/react map files, NEB headers, -reorder file)
- replace use of `sprintf()` with `fmt::format()` or `std::to_string()`
- apply a consistent convention to `get_fix_by_style()` patterns: keep the leading `^` anchor, drop the trailing `$` where the style name is an unambiguous prefix so that future suffixed (accelerated) variants are matched, keep `$` where a prefix match could return an unrelated style (e.g. `^drude$`, since the `drude/transform` fixes are not derived from `FixDrude`)

Several small latent bugs found during the conversion are fixed as part of the respective commits (an unchecked negative compute index in fix gemc, a memory leak on re-init in fix nve/spin and in the BPM bond style destructors, a leaked malloc buffer in fix shardlow/kk, uninitialized reads in the -reorder and fix ehex error paths, an out-of-bounds access for malformed KIM index ranges, an uncaught exception for non-numeric variable values passed to Python functions, and two no-op sscanf calls on section titles in fix amoeba/pitorsion).  File readers converted to the reader classes were verified to reproduce the reference logs of the bundled examples exactly (amoeba ubiquitin, bocs LD, tiny_nylon, neb, orient/eco) and the lcbop force-style test.

:warning: DRUDE package: While applying the style match conventions it was noticed that `fix_tgnh_drude.cpp` is a modified copy of `fix_nh.cpp` where most error messages and several comments were left unchanged from the original, so problems were reported for "fix nvt/npt/nph" instead of the actual tgnvt/tgnpt/drude styles.  The file was audited against the evolution of `fix_nh.cpp` since it was forked (November 2020): all applicable fix_nh bugfixes since then had already been carried over, but the audit uncovered and fixes an inherited **out-of-bounds read**: the `temp` keyword parsing still used the fix nh bounds check for 4 arguments although the tgnh variant reads 6, so a too short fix command read past the end of the argument list.

In addition, the detection of incompatible constraint fixes (shake/rattle/ilves) was made consistent across `fix ehex`, `fix filter/corotate`, `fix tfmc`, `fix temp/csld`, `fix baoab`, and `fix nve/limit`, and now also covers the new `fix ilves` constraint fix.

21 per-atom computes in the MACHDYN and DPD-MESO packages compared the compute style against hard-coded strings that did not match the registered style names (underscores instead of slashes, copy-paste names from other computes), so their duplicate instance warnings were unreachable dead code; the names were corrected and the checks converted to the anchored `get_compute_by_style()` pattern, which means inputs defining more than one instance of these computes will now start to see the warning.

Processing of the GitHub Copilot review (threads are resolved after processing):

- *Modify::delete_fix()/delete_compute() early return changes unfix/uncompute semantics*: **valid.**  The lenient API is intentional (destructor cleanup), so a strict check for a valid ID was added to the unfix/uncompute command handling in the Input class instead, which restores the previous behavior and reports the error with a pointer to the failing argument of the input line.
- *fix nve/spin caches stale fix pointers across runs*: **partially valid.**  The claims for the precession/langevin fixes were incorrect (those loops are bounded by counts that are reset to zero), but the setforce/spin pointer use is not count-guarded and could indeed dereference a dangling pointer after that fix was deleted; this staleness pre-dates this pull request and is now fixed by resetting the cached flags and pointer in init().
- *error->one() instead of error->all() for collective checks in fix filter/corotate and fix ehex*: **valid,** fixed (partially a pre-existing issue inherited from the original fix shake check).
- Two additional comments on fix nve/spin were *suppressed as low confidence* by Copilot but checked anyway: the setforce/spin one was in fact the **most severe finding of the review** (a genuine pre-existing use-after-free, since that pointer use is not count-guarded), the langevin one is runtime-harmless (count-bounded loop).  Both are covered by the flag/pointer reset applied for the inline comment above.

**Post Submission Checklist**


src/DPD-MESO/compute_edpd_temp_atom.cpp
```diff
@@ -50,11 +53,8 @@ ComputeEDPDTempAtom::~ComputeEDPDTempAtom()
 void ComputeEDPDTempAtom::init()
 {
 
-  int count = 0;
-  for (int i = 0; i < modify->ncompute; i++)
-    if (strcmp(modify->compute[i]->style,"temp_vector/atom") == 0) count++;
```
jtclemm, line 55: Just to confirm, was the temp_vector/atom style renamed?

src/AMOEBA/fix_amoeba_bitorsion.cpp
```diff
@@ -727,23 +729,25 @@ double FixAmoebaBiTorsion::compute_scalar()
...
 
-    (void) fgets(line,MAXLINE,fp);
-    (void) fgets(line,MAXLINE,fp);
-    eof = fgets(line,MAXLINE,fp);
-    if (eof == nullptr)
-      error->one(FLERR, Error::NOLASTLINE, "Unexpected end of fix amoeba/bitorsion file");
+    // skip title line, then read number of types from "N bitorsion types" line
 
-    sscanf(line,"%d",&nbitypes);
+    reader = new TextFileReader(fp, "fix amoeba/bitorsion");
```
rbberger:
```c++
std::unique_ptr<TextFileReader> reader;

reader = std::make_unique<TextFileReader>(fp, "fix amoeba/bitorsion");
```

src/DRUDE/fix_tgnh_drude.cpp
```diff
@@ -14,6 +14,11 @@
 
 /* -------------------------------------------------------------------------------------
    Contributing authors: Mark Stevens (SNL), Aidan Thompson (SNL), Zheng Gong (ENS Lyon)
+
+   This is a modified copy of fix_nh.cpp (by Mark Stevens and Aidan Thompson) that
+   implements temperature-grouped Nose-Hoover thermostatting/barostatting for
+   Drude polarizable models.  Bugfixes and updates applied to fix_nh.cpp and its
+   derived classes may have to be applied to this file and its derived classes, too.
```
jtclemm, line 21: Seems a little vague of a warning. Is there a particular date from which bugs have to be checked? Or is there strong reason to believe there are bugs?

src/DPD-MESO/compute_edpd_temp_atom.cpp
akohlmey: This must have been done *before* the package was included in LAMMPS.

src/DRUDE/fix_tgnh_drude.cpp
akohlmey: I have done what I can, everything beyond that needs to be done by the people that have contributed the package. It *looks* as if all fixes from upstream have been included as well, so the changes here are mostly cosmetic. However, I felt that adding a comment indicating the heritage of this file and that it replicates (to some degree) fix_nh.cpp was needed, since it was just copied without.

src/AMOEBA/fix_amoeba_bitorsion.cpp
akohlmey: done. thanks for reminding me. I keep forgetting this pattern for no reason.

src/DRUDE/fix_tgnh_drude.cpp
akohlmey: > Is there a particular date from which bugs have to be checked? Or is there strong reason to believe there are bugs?

Just to be sure, I have checked all changes to fix_nh.cpp since this file was copied and verified if the equivalent changes are applicable and present here. This is now up-to-date as far as I can tell.

jtclemm: Got it, I suppose that's as good as it'll get.

jtclemm (approved):

PR #5141, opened 2026-08-19
Add KOKKOS Support for CG-DNA (oxDNA) Package, plus some minor bug fixes for bonus data KOKKOS

lrussell676:
**Summary**

Primarily, this PR covers KOKKOS support for oxDNA and all of it's variants. In the process, I did encounter some very minor bugs (oversights more like) with some of the more fundamental parent classes however. Some performance optimisation is included too. These are as follows:

    - src/KOKKOS/atom_vec_ellipsoid_kokkos.cpp | offset required "atomKK->avecKK->" in order to set the correct buffer values when using atom_vec_hybrid_kokkos.
    
    - atom_vec_kokkos/comm_kokkos.cpp | added in a new fused kernel for reverse comm when only running on 1 GPU. This, in our testing, gave ~5% speedup.
    
    - bond_hybrid_kokkos.cpp | just needed an extra if statement trigger and a memoryKK->create_kokkos for the k_bondlist. 

    - comm_tiled_kokkos.cpp | this segfaults if trying to use with KOKKOS atom styles that have bonus data (we only ever worked on comm_kokkos.cpp for ellipsoidKK in PR https://github.com/lammps/lammps/pull/4865). So an error message is now added in here.

    - fix_langevin_kokkos/fix_nve_asphere_kokkos.cpp | some refactoring and use of Kokkos:fma, purely for performance and is a little faster now.

    - kokkos_type.h | fma gets routed through Kokkos::fma. This works, but perhaps it would be preferred to replace fma with Kokkos::fma in my kernels instead? I also added in some new View types here, although in the end I didn't use some of these (ttransform_kkfloat_5d, ttransform_kkfloat_1d_4x4, and kkfloat_1d_4x4 are unused). These are safe to keep for future, or can be removed.

    - math_extra_kokkos.h | added float & double versions for some of these since quat in bonus struct is still double (i,e, - not just KK_FLOAT). In a few places, I static_cast these quats locally to floats in single/mixed prec. So we end up needing both float and double versions.

    - atom_masks.h | added a new CG_DNA_MASK

    - compute_bond.cpp | I don't remember doing this, but it looks like I needed this to handle hybrid/kk at some point.

    - KOKKOS.cmake | Registers our internal fixes if PKG_CG-DNA

**Related Issue(s)**

N/A

**Author(s)**

Lewis M Russell, Stan Moore, Oliver Henrich

**Licensing**

AI (GH Copilot) has been used in places, though for the most part this has been to port/debug maths between vanilla oxDNA interactions over to KOKKOS. There aren't any parts of this PR that have blindly been prompted and taken for granted. 

**Backward Compatibility**

Yes, this is 100% backwards compatible. 

**Implementation Notes**

I have a test_KOKKOS.sh in examples/PACKAGES/cgdna/examples that nicely tests for numerical correctness vs vanilla. We are also in the process of benchmarking far larger system sizes (not shown here), and we haven't encountered any problems in doing so. 

**Post Submission Checklist**

N/A


src/CG-DNA/bond_oxdna3_fene.cpp
```diff
@@ -24,13 +24,24 @@
...
+   by not calling BondOxdna3FENE::coeff directly. We can also avoid
+   code duplication of coeff within KOKKOS using this approach.
+
+   coeff_oxdna3_common() lives on the BondOxdnaFene base class, which means
+   it can be called from both the vanilla and KOKKOS versions without
+   duplicating the file parsing logic or requiring KOKKOS to inherit from
+   BondOxdna3Fene.
 ------------------------------------------------------------------------- */
-void BondOxdna3Fene::coeff(int narg, char **arg)
+void BondOxdnaFene::coeff_oxdna3_common(int narg, char **arg)
```
akohlmey: It is a *very* bad idea to place the implementation of a function into source file for a *different* class.

src/CG-DNA/bond_oxdna3_fene.cpp
```diff
@@ -24,13 +24,24 @@
 #include "math_special.h"
 #include "potential_file_reader.h"
 
+#include <cmath>
```
akohlmey: Is there anything in *this* header that uses math functions?
At least there is nothing added in this pull requests.
For a large program like LAMMPS you want the *minimum* number of includes in the header (fewer namespace conflicts, faster compilation due to fewer dependencies). And also follow the Include-what-you-use principle in the .cpp file

src/CG-DNA/bond_oxdna_fene.h
```diff
@@ -41,10 +41,13 @@ class BondOxdnaFene : public Bond {
   double single(int, double, int, int, double &) override;
 
  protected:
+  // Shared oxDNA3 coeff parser (called by KOKKOS too), defined in bond_oxdna3_fene.cpp
```
akohlmey: See my other comment about this in the .cpp file.

src/CG-DNA/pair_oxdna2_coaxstk.cpp
```diff
@@ -543,13 +543,13 @@ void PairOxdna2Coaxstk::compute(int eflag, int vflag)
         deltb[2] -= t6dir[2] * tpair;
       }
 
-      // increment torques
+              // increment torques
```
akohlmey: This looks like a misplaced indentation.

src/CG-DNA/pair_oxdna2_coaxstk.h
```diff
@@ -44,6 +44,9 @@ class PairOxdna2Coaxstk : public Pair {
   void *extract(const char *, int &) override;
 
  protected:
+  // Shared oxDNA3 setup and coeff parser (used by KOKKOS), defined in pair_oxdna3_coaxstk.cpp
```
akohlmey: Again, this is a very bad idea to mix up declaration and definition between classes and their source files.

src/CG-DNA/pair_oxdna3_coaxstk.cpp
```diff
@@ -22,33 +22,60 @@ using namespace LAMMPS_NS;
...
 
-PairOxdna3Coaxstk::PairOxdna3Coaxstk(LAMMPS *lmp) : PairOxdna2Coaxstk(lmp)
+void PairOxdna2Coaxstk::init_eta_cxst_oxdna3()
 {
+   // sequence-specific coaxial stacking strength
+   // A:0 C:1 G:2 T:3, 3'- [i] X [j] -5'
 
-  // sequence-specific coaxial stacking strength
-  // A:0 C:1 G:2 T:3, 3'- [i] X [j] -5'
+   eta_cxst[0][0] = 1.1217958408368172;
```
akohlmey: There seems to be misplaced indentation here.

src/CG-DNA/pair_oxdna_coaxstk.cpp
```diff
@@ -581,27 +581,6 @@ void PairOxdnaCoaxstk::compute(int eflag, int vflag)
       // Full cosphi3 and cosphi4 (=cosphi3) contribution to the torque
       if (cosphi3 != 0.0) {
 
-        gamma = dx_cbk_oxdna1 - dx_cstk_oxdna1;
```
akohlmey, line 584: Was this dead code?

src/compute_bond.cpp
```diff
@@ -33,9 +33,14 @@ ComputeBond::ComputeBond(LAMMPS *lmp, int narg, char **arg) :
   peflag = 1;
   timeflag = 1;
 
-  // check if bond style hybrid exists
+  // check if bond style hybrid with and without suffix exists
 
-  bond = dynamic_cast<BondHybrid *>(force->bond_match("hybrid"));
+  if (utils::strmatch(force->bond_style, "^hybrid"))
```
akohlmey: This is the wrong place and the wrong way to address the issue of suffixes. You will have to repeat this all over the place. Instead you should augment `Force::bond_match()` to have an interface similar to `Force::pair_match()`.

akohlmey (commented):
Here are some preliminary issues that stood out while having a casual look at the changes. There is likely more.

src/CG-DNA/pair_oxdna2_coaxstk.cpp
lrussell676: I swear I've had to fix that exact indentation in that exact line about 3-4 times now - perhaps I overlooked it during a merge. Sorted now.

src/CG-DNA/pair_oxdna_coaxstk.cpp
lrussell676: Yes, I noticed this was apples for apples repeated calculation from the previous cosphi3 block. So we can just go straight to tpair = ..... at this point.

src/CG-DNA/pair_oxdna3_coaxstk.cpp
lrussell676: Yes, indeed. Now sorted.

src/CG-DNA/bond_oxdna3_fene.cpp
ohenrich: I couldn't agree more. But we are facing some of KOKKOS' idiosyncrasies.

The overall goal is to use the existing init(), coeff(), etc routines from vanilla as much as possible to run on the host where possible and to minimise code duplication (which is already horrific due to some design principles that don't work on the GPU).

We will think about alternatives, but it would be good to check suggestions before implementing them. We are open for any solution.

lrussell676: Yes.... We never really came to a 'nice' solution here. 

That's a crude example of what we are after inheritance wise, taking our oxdna fene bond as an example.

I originally had a different approach with forward declarations and friends. We didn't really like this either. This commit looks like it summarises what I had: https://github.com/lammps/lammps/pull/5141/changes/82c47c8e40d3213722ab16cda89eebe8c6a8a79c#diff-a20b8279444d78dad5f837d93639db3ac441a7ccce5bcc0fe680563701777f06

I remember first going for the obvious virtual inheritance approach. But I quickly ran into a ton of compiling errors with this, which boiled down to multiple inheritance issues between KOKKOS vs vanilla sharing oxDNA1 vanilla as a parent. Due to the caveat of virtual functions being a problem on GPUs, I went with templating instead for the KOKKOS compute routines across oxDNA model variants. Which wasn't a problem for oxDNA1/2 and oxRNA2, but became this whole headache with oxDNA3 as it has a different coeff routine.

Honestly, I remember asking GH Copolit about this and tried searching for what best to do here. Nothing shined.

I'm open to any suggestions you have for this. We really want to avoid code duplication wherever possible.

ohenrich: To add to Lewis explanation: The general idea is to use from vanilla whatever is possible in terms of `init()`, `coeff()`, etc routines and run it on the host. This leads to this rather cumbersome inheritance structure, which is not a problem on the CPU, but is on the GPU. Suggestions?

src/CG-DNA/bond_oxdna3_fene.cpp
lrussell676: No, you're right. I reckon a lot of these have creeped in from WIP development and I've not remembered to remove them.

I've just pushed IWYU updates. A lot of the "error: add" results where false positives, though "error: superfluous" caught a bunch of unnecessary includes (though again, some of these where false positives, but not too many).

So I've removed a good chunk of these, and in places moved some .h includes into .cpp where they actually get used.

I'm not saying I've caught everything, but this is a cleanup nonetheless.

src/CG-DNA/bond_oxdna3_fene.cpp
akohlmey: The cleanest way that I can think of to share functionality is add one more inheritance layer (say `PairCGDNA` and `BondCGDNA`) that goes between `Pair` or `Bond` and the actual classes. The common class declares flags that indicate the variant in use and the corresponding constructors set those flags as needed.
Then you can either inherit the common functionality completely, if nothing else is needed, or override, say, `coeffs()` and then inside either run `PairCGDNA::coeffs()` first and then process any additional arguments, or do it the other way around.

This can take some inspiration from `fix_nh.h/.cpp`. In fact, it should be easier, since you don't need to parse arguments from the constructor.

Let me also loop in @jtclemm who has implemented even more complex schemes with shared initialization.

src/compute_bond.cpp
lrussell676: Ok, I think I've understood you correctly. I've augmented bond_match to exactly the same logic as pair_match.

Only difference is that for bond match I have "int = 1" default so that I didn't have to change all its function calls. bond_match appears to be used far less than pair_match however, so I can make the arg explicit if preferred?  grep shows the changes would be pretty minor:

```
(base) lewis@lewis-B550-AORUS-ELITE-AX-V2:/media/lewis/PhD/GH_lammps/src$ grep -d skip "bond_match" ./*
./compute_bond.cpp:  bond = dynamic_cast<BondHybrid *>(force->bond_match("^hybrid", 0));
./compute_bond.cpp:  bond = dynamic_cast<BondHybrid *>(force->bond_match("^hybrid", 0));
./fix_adapt.cpp:        ad->bond = force->bond_match(fmt::format("{}/{}",bstyle,lmp->suffix));
./fix_adapt.cpp:      if (ad->bond == nullptr) ad->bond = force->bond_match(bstyle);
./force.cpp:Bond *Force::bond_match(const std::string &word, int exact, int nsub)
./force.h:  Bond *bond_match(const std::string &, int = 1, int nsub = 0);
(base) lewis@lewis-B550-AORUS-ELITE-AX-V2:/media/lewis/PhD/GH_lammps/src$ grep -d skip "bond_match" ./*/*
./KOKKOS/neigh_bond_kokkos.cpp:  if (force->bond && force->bond_match("quartic")) bond_off = 1;
./PLUGIN/plugin.cpp:    if ((lmp->force->bond_style != nullptr) && (lmp->force->bond_match(name) != nullptr))
```

lrussell676: Actually, I just went ahead and put in the explicit "exact" arg into those files grep highlighted above - bond_match is now identical logic wise to pair_match. Can't imagine this would have any unintended consequences?

src/KOKKOS/atom_vec_oxdna_kokkos.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+/* ----------------------------------------------------------------------
+   process field strings to initialize data structs for all other methods
+------------------------------------------------------------------------- */
+
+void AtomVecOxdnaKokkos::init()
+{
+  AtomVecOxdna::init();
+
+  if (strcmp(update->unit_style, "lj") != 0)
```
lrussell676, line 42: Tying into your comment on my bond_match @akohlmey , is this strcmp here ok?

I gave it a test my end, changing from "units lj" to "units real" and it triggers as I intended.

akohlmey: > is this strcmp here ok?

What is your concern?

lrussell676: It's okay thanks - I was just panicking and overthinking for nothing sorry.... I confused myself with suffix handling after dealing with the bond_match. Of course, there is no suffix here on the units, and I was seeing things that don't exist.

src/CG-DNA/bond_oxdna3_fene.cpp
ohenrich: Sounds also a bit nasty as this means we would have to split some functionality (`coeff()`) from other functionality (`compute()`) that belongs to the same class. It would be quite messy as we have 6 pair classes and 3 model variants, so quite a few combinations. But we are considering this. 

@jtclemm Anything more complicated you could point us to for inspiration?

akohlmey: @ohenrich what you have right now is *ultra-nasty*. But I don't understand your concern, you just put all shared flags with the shared code and then use some setting, sounds like a bitmap would be the right choice here, to select which section of the shared function is run. Everything else is set to a safe dummy value. I mentioned fix_nh.cpp as example: it contains the code for fix nvt, fix npt, fix nph, where fix npt is the combination of fix nvt and fix nph. I also don't understand why you need to mix coeff() and compute() those serve very different purposes. It would be a bad idea to change the logic of what LAMMPS uses each of these functions for. That would be a maintenance nightmare.

Please also keep in mind that avoiding replication at all cost is just as bad as replicating everything. In most cases there is a reasonable middle ground.

jtclemm: Without fully understanding the layout of the oxDNA classes (nor having worked much with the restrictions involving kokkos), some examples I can think of that were used to deal with shared initialization problems are:

1) Fix deform and fix deform/pressure. Fix deform/pressure was a subclass requiring fix deform to work alone and unchanged. This created issues with the order of constructors and parsing arguments which were solved with various tricks. A lot of duplicated functions from the parent class were also extracted into their own methods so both could call them when needed to avoid code duplication too.
2) Fix surface/local and fix surface/global, which are thematically similar with a lot of duplicate intermediate methods but otherwise had very little in common. Here, they both inherit from a parent class but also make use of a helper file surf_extra.cpp.

But these were much less entangled then the diagram you have above!

akohlmey: > But these were much less entangled then the diagram you have above!

But on the other hand, there are no worries about the order of constructors because the initialization for pair style is done in settings() and coeff() and not the constructor.

akohlmey: @ohenrich @lrussell676 another approach to reduce redundant code is to have a templated style with functors. See PR #5020 for an example where this is applied to many pairwise additive potential pair styles which replicate a lot of the boilerplate code and differ only in a few places.

lrussell676: Thanks for the examples. I'm going to need to spend quite some time dissecting this. It is pretty nasty for our case for sure, but I'm sure we'll figure out a reasonable solution :)

Templating doesn't immediately jump out at me. It's great for GPU "virtual function " workaround, but I'm missing how this would help with "up-to-vanilla/parent, along to vanilla/child" that we currently have with oxDNA/kk?

lrussell676: Our main problem is the "along, up, along" idea with KOKKOS vs vanilla. It makes branchpoints between multiple classes a little bit insane.

Worst case, we admit code dup and put a big bold comment on these sections - a note to developers that if they edit "X" area they need to copy and paste to the associated KOKKOS area of the code?

My suggestion here is horrible though, and it should be treated as a last resort.

akohlmey: > Our main problem is the "along, up, along" idea with KOKKOS vs vanilla. It makes branchpoints between multiple classes a little bit insane.

@lrussell676 It looks to me that the biggest problem we have here is a communication problem. You are presenting us with a solution where many details don't add up. I am still confused about what can be the motivation for defining a class member function in the body of a *different* class. Furthermore, you are discussing this with a "shortcut notation" that may make sense to you, but is just further increasing *my* confusion. Which leads to the problem that it is not at all clear from the outside what are the problems that you are trying to solve that cannot be solved by a straightforward, logical inheritance tree. I have the suspicion that you may be 'suffering' from a case of 'premature optimization syndrome'. You would not be the first.

Thus, my suggestion at this point is to take a step back and first try to lay out in as simple a way as possible the flow of control in pseudocode. That will a) help *you* to be more clear in your mind about the problem you are trying to solve and b) allow *us* to better understand the nomenclature and logic that you are using.

I understand that this may be frustrating and annoying from your perspective, but let me assure you that I have been at the receiving side of this kind of treatment many times before. I have learned the lesson that the job of a maintainer is mostly to say "no" at the right time and be stubborn about it. The logic behind this is, that if a problem cannot be explained in terms that are simple enough for a person on the outside to understand, it either means that the problem is not yet understood well enough, or the explanation needs to be better (and the implementation of a code and how easy it is to follow it is part of the explanation). In almost all cases, the stubbornness of the maintainer has resulted in a much better solution. The important part right now is not to get lost in details.

lrussell676: @jtclemm

"A lot of duplicated functions from the parent class were also extracted into their own methods so both could call them when needed to avoid code duplication too."

I'll look into this next week.

ohenrich: Thanks for all your input. We had a think and are somewhat inclined to go for code duplication as the extent would be rather limited.

This would have the advantage of a bog-standard inheritance structure very similar to vanilla. Moreover, if in the future anything changes at the top (e.g. BondFeneKokkos not hanging off BondFene but being a child of a BondKokkos class) we are already set up for this further down in our inheritance structure.

lrussell676: @akohlmey seconding Oliver here, and thanks for the info. Oliver and I discussed this today and we are all on the same page that there is a middle ground between avoiding replication vs replicating everything.

In our case, the main problem was just the coeff() routines in oxDNA3. We decided code dupping was the cleanest approach for this. It wasn't as bad as I thought, and I've made the differences very clear from my comments (hopefully). I've now refactored our bond_ox[d/r]na*_fene[_kokkos] classes which are in my second last commit (last commit was just a minor debug from this). Do you think this acceptable now?

If so, it will be the same idea for all our pair styles. Some of these will have a tiny bit more code dup from their vanilla counterparts, but these are very limited (a couple of lines in the constructors and in only a few classes).

In the end, this should be pretty easy to follow and maintain. Our KOKKOSes are (will be) only ever a child of oxDNA(1) and hence inheritance-wise both vanilla and KOKKOS then go off their own inheritance trees in a clean and typical fashion for all our different oxDNA flavours/variants. In other words, with these fairly isolated code dups, we remove the need for any messy access to oxDNA3-vanilla from oxDNA3-KOKKOS.

lrussell676: I've went ahead and sorted this now across all our oxdna bond/pair styles now @ohenrich @akohlmey . These are my commits titled "refactor classes - [stk/hbond/etc]. Can we mark this comment/issue resolved?

lrussell676:
Couple of comments/notes regarding merging with develop today ([946113b](https://github.com/lammps/lammps/pull/5141/commits/946113bc70becfa4f824ef55ceb0cd1abc1362f4)) . This was pretty straightforward, aside from: fix_langevin_kokkos, fix_nve_asphere_kokkos, and math_extra_kokkos. Main thing was that these now have silent-conversion (static_casts) applied in develop, but I had refactored these in the interest of some performance boost. My edits to these files would most definitely benefit from a second pair of eyes as the conflicts where a tad messy to sort through.

I applied Kokkos::fma (fused-multipy-add) where possible; noticed some prefactors that I could take out of loops; cut-down on global memory access and assigned to locals. Regarding quaterions/shape, we have to carry these around as doubles (since the KOKKOS DualView for bonus data takes the vanilla struct and is forced into double) - as a bit of a workaround, in the heavier quat calculations I assign quat/shape to local KK_FLOATs with static_cast then promote back to double when writing back to the global quat/shape.

As for the upstream removal of "examples/PACKAGES/cgdna/examples/[lj/real]_units/oxDNA3/unique_bp/log.30Mar2026.dsring2.g++.8" I've kept these in here for now; this is just convenience-sake my end for running our test[_KOKKOS].sh scripts. We'll be moving/rejigging these anyway as per @akohlmey comments.

